PR #584, opened 2019-02-05
Remove DB env var

guicassolato:
**What this PR does / why we need it**

This PR removes any reference to the environment variable `DB` in porta, making it depend always and only on `DATABASE_URL` instead.

Safe defaults remain in place for mysql database (all environments).

Deployments that depend on `config/examples/database.yml`, in test environment, will automatically have the value of the `TEST_ENV_NUMBER` env var appended to the value of `DATABASE_URL` (thus modifying the  name of the database)

**Which issue(s) this PR fixes**

Without this, while using openshift templates, either
1) `DB` (environment variable) needs to come from a deployment parameter just like, independently, other database connection details (username, password); _OR_
2) multiple versions of the templates would have to be maintained (one per database option for porta), even for the 'ha' version (external database)

PR #894, opened 2019-06-14
Update Makefile to use DATABASE_URL instead of DB

Martouta:
Update the Makefile to use `DATABASE_URL` instead of `DB`.
Needed since https://github.com/3scale/porta/pull/584

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -166,11 +166,11 @@ bundle: gemfiles/prod/Gemfile Gemfile
 
 oracle-db-setup: ## Creates databases in Oracle
 oracle-db-setup: oracle-database
-	MASTER_PASSWORD=p USER_PASSWORD=p ORACLE_SYSTEM_PASSWORD=threescalepass NLS_LANG='AMERICAN_AMERICA.UTF8' DB=oracle bundle exec rake db:drop db:create db:setup
+	MASTER_PASSWORD=p USER_PASSWORD=p ORACLE_SYSTEM_PASSWORD=threescalepass NLS_LANG='AMERICAN_AMERICA.UTF8' DATABASE_URL="oracle-enhanced://rails:railspass@127.0.0.1:1521/systempdb" bundle exec rake db:drop db:create db:setup
 
 schema: ## Runs db schema migrations. Run this when you have changes to your database schema that you have added as new migrations.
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -166,11 +166,11 @@ bundle: gemfiles/prod/Gemfile Gemfile
 
 oracle-db-setup: ## Creates databases in Oracle
 oracle-db-setup: oracle-database
-	MASTER_PASSWORD=p USER_PASSWORD=p ORACLE_SYSTEM_PASSWORD=threescalepass NLS_LANG='AMERICAN_AMERICA.UTF8' DB=oracle bundle exec rake db:drop db:create db:setup
+	MASTER_PASSWORD=p USER_PASSWORD=p ORACLE_SYSTEM_PASSWORD=threescalepass NLS_LANG='AMERICAN_AMERICA.UTF8' DATABASE_URL="oracle-enhanced://rails:railspass@127.0.0.1:1521/systempdb" bundle exec rake db:drop db:create db:setup
 
 schema: ## Runs db schema migrations. Run this when you have changes to your database schema that you have added as new migrations.
 	bundle exec rake db:migrate db:schema:dump
-	MASTER_PASSWORD=p USER_PASSWORD=p ORACLE_SYSTEM_PASSWORD=threescalepass NLS_LANG='AMERICAN_AMERICA.UTF8' DB=oracle bundle exec rake db:migrate db:schema:dump
+	MASTER_PASSWORD=p USER_PASSWORD=p ORACLE_SYSTEM_PASSWORD=threescalepass NLS_LANG='AMERICAN_AMERICA.UTF8' DATABASE_URL="oracle-enhanced://rails:railspass@127.0.0.1:1521/systempdb" bundle exec rake db:migrate db:schema:dump
```
hallelujah: Missing postgresql by the way I think

Martouta: @hallelujah What do you mean? I didn't find any other `DB=` in this file. But if you mean that the makefile is missing steps for postgres, then I agree 😄 On the other hand, if no one has noticed of this so far, I wouldn't discard that no one uses this steps of the makefile and then i would directly remove them 😂

hallelujah: Yes it is missing postgres
Nobody uses it because this Makefile needs to be reworked on.

PR #898, opened 2019-06-14
Include Postgres schema dumping into make schema

duduribeiro:

After making some change in the database we need to regenerate the
schema for Postgres. This changes the `make schema` command to
accept a env var `POSTGRES_DATABASE_URL` and uses it to migrate
the database and updates the schema after the migration

**Verification steps** 
Start your postgres database and create the data.

to simulate a non-executed migration in postgres you can run:
`DATABASE_URL="YOUR_PG_URL" bundle exec rake db:rollback db:schema:dump`

the db/postgres_schema.rb should have been changed with some columns being removed.

Now you can run `POSTGRES_DATABASE_URL="YOUR_PG_URL" make schema` (it will run the migration and redump the schema) and your
db/postgres_schema.rb should have no difference (it will be the same as git since we regenerated the schema)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ C:\\nppdf32Log\\debuglog.txt
 .zeus.sock
 .deploy
 .s3cfg
+.tool-versions
 
 # Bundler Stuff
 .bundle
--- a/Makefile
+++ b/Makefile
@@ -169,8 +169,11 @@ oracle-db-setup: oracle-database
 	MASTER_PASSWORD=p USER_PASSWORD=p ORACLE_SYSTEM_PASSWORD=threescalepass NLS_LANG='AMERICAN_AMERICA.UTF8' DATABASE_URL="oracle-enhanced://rails:railspass@127.0.0.1:1521/systempdb" bundle exec rake db:drop db:create db:setup
 
 schema: ## Runs db schema migrations. Run this when you have changes to your database schema that you have added as new migrations.
+schema: POSTGRES_DATABASE_URL ?= "postgresql://postgres:@localhost:5433/3scale_system_development"
+schema:
 	bundle exec rake db:migrate db:schema:dump
 	MASTER_PASSWORD=p USER_PASSWORD=p ORACLE_SYSTEM_PASSWORD=threescalepass NLS_LANG='AMERICAN_AMERICA.UTF8' DATABASE_URL="oracle-enhanced://rails:railspass@127.0.0.1:1521/systempdb" bundle exec rake db:migrate db:schema:dump
+	DATABASE_URL=$(POSTGRES_DATABASE_URL) bundle exec rake db:migrate
 
 oracle-database: ## Starts Oracle database container
 oracle-database: ORACLE_DATA_DIR ?= $(HOME)


.gitignore
```diff
@@ -26,6 +26,7 @@ C:\\nppdf32Log\\debuglog.txt
 .zeus.sock
 .deploy
 .s3cfg
+.tool-versions
```
duduribeiro: This is for https://asdf-vm.com

Makefile
```diff
@@ -187,6 +180,29 @@ oracle-database:
...
+	MASTER_PASSWORD=p USER_PASSWORD=p ORACLE_SYSTEM_PASSWORD=threescalepass NLS_LANG='AMERICAN_AMERICA.UTF8' DATABASE_URL="oracle-enhanced://rails:railspass@127.0.0.1:1521/systempdb" bundle exec rake db:drop db:create db:setup
+
+## Postgres database containerization
+postgres-database: ## Starts Postgres database container
+postgres-database:
+	[ "$(shell docker inspect -f '{{.State.Running}}' postgres-database 2>/dev/null)" = "true" ] || docker start postgres-database || docker run \
+		-p 5432:5432 \
+		--name postgres-database \
+		-e POSTGRES_USER=postgres \
+		-e POSTGRES_DB=threescale \
```
Martouta: I think we don't need the password 🤔

Makefile
```diff
@@ -187,6 +180,29 @@ oracle-database:
...
+		circleci/postgres:10.5-alpine
+
+postgres-db-setup: ## Creates databases in Postgres
+postgres-db-setup: postgres-database
+	DATABASE_URL="postgresql://postgres:@127.0.0.1:5433/threescale" bundle exec rake db:drop db:create db:setup
```
Martouta: Mine works with: `DATABASE_URL="postgresql://postgres:@localhost:5433/circleci" bundle exec rake db:drop db:create db:setup` but if you tested this version and it works, good too 👍

Makefile
```diff
@@ -187,6 +180,29 @@ oracle-database:
...
+
+schema: ## Runs db schema migrations. Run this when you have changes to your database schema that you have added as new migrations.
```
Martouta: Probably this should do it also for mysql, so we do it for all DBs at once

Makefile
```diff
@@ -187,6 +180,29 @@ oracle-database:
 		-v $(PWD)/script/oracle:/opt/oracle/scripts/setup \
 		quay.io/3scale/oracle:12.2.0.1-ee
 
+oracle-db-setup: ## Creates databases in Oracle
+oracle-db-setup: oracle-database
+	MASTER_PASSWORD=p USER_PASSWORD=p ORACLE_SYSTEM_PASSWORD=threescalepass NLS_LANG='AMERICAN_AMERICA.UTF8' DATABASE_URL="oracle-enhanced://rails:railspass@127.0.0.1:1521/systempdb" bundle exec rake db:drop db:create db:setup
+
+## Postgres database containerization
+postgres-database: ## Starts Postgres database container
```
hallelujah: I wonder, most of the OS have postgres shipped with a package manager.
So is this really needed?

Makefile
```diff
@@ -187,6 +180,29 @@ oracle-database:
...
+		-e POSTGRES_DB=threescale \
+		circleci/postgres:10.5-alpine
+
+postgres-db-setup: ## Creates databases in Postgres
+postgres-db-setup: postgres-database
+	DATABASE_URL="postgresql://postgres:@127.0.0.1:5433/threescale" bundle exec rake db:drop db:create db:setup
+
+schema: ## Runs db schema migrations. Run this when you have changes to your database schema that you have added as new migrations.
+	bundle exec rake db:migrate db:schema:dump
+	MASTER_PASSWORD=p USER_PASSWORD=p ORACLE_SYSTEM_PASSWORD=threescalepass NLS_LANG='AMERICAN_AMERICA.UTF8' DATABASE_URL="oracle-enhanced://rails:railspass@127.0.0.1:1521/systempdb" bundle exec rake db:migrate db:schema:dump
```
Martouta: Do we need the `db:schema:dump` after the `db:migrate` ? 🤔

Martouta: Does this work? In my case it requires me to send all these env vars: `DATABASE_URL="oracle-enhanced://rails:railspass@127.0.0.1:1521/systempdb" ORACLE_SYSTEM_PASSWORD=threescalepass NLS_LANG=AMERICAN_AMERICA.UTF8`

Makefile
duduribeiro: Just added if the person didn't want to install it  😄. I can remove it also and accept the DATABASE_URL as argument, so the user can pass it using any installation method used. Wdyt @hallelujah and @Martouta ?

Makefile
duduribeiro: good catch!. not necessary, tested here. thanks ☕️

Makefile
Martouta: Good for me 👍 and we follow the same approach for all DBs 😄 but i am opened to opinions about this 😄

Makefile
```diff
@@ -171,6 +171,7 @@ oracle-db-setup: oracle-database
 schema: ## Runs db schema migrations. Run this when you have changes to your database schema that you have added as new migrations.
 	bundle exec rake db:migrate db:schema:dump
 	MASTER_PASSWORD=p USER_PASSWORD=p ORACLE_SYSTEM_PASSWORD=threescalepass NLS_LANG='AMERICAN_AMERICA.UTF8' DATABASE_URL="oracle-enhanced://rails:railspass@127.0.0.1:1521/systempdb" bundle exec rake db:migrate db:schema:dump
+	DATABASE_URL=$(POSTGRES_DATABASE_URL) bundle exec rake db:migrate
```
hallelujah: We could have default one wdyt?

Martouta: I agree with @hallelujah , it should work without params 😄

duduribeiro: Agreed!. Just thinking which one can be the default 🤔 . any suggestion?

Martouta: `postgresql://postgres:@localhost:5433/circleci`

duduribeiro: Just don't know about `circleci` as the dbname.. This relies that the user setup their database naming `circleci`. Maybe `threescale`?. WDYT?

any suggestion @hallelujah ?

hallelujah: Well I'd say the same as mysql :) `3scale_system_development`

duduribeiro:
@Martouta and @hallelujah, updated the code

thanks

duduribeiro:
also need review here @3scale/system 😄 

thanks

PR #903, opened 2019-06-14
Does not send request to segment if disabled

duduribeiro:

3scale on-prem tries to send requests to segment.io and fails when it
is not reachable. This commit creates a section to toggle segment at
features.yml and it avoids sending requests to segment.io if it is
disabled.

Closes [THREESCALE-2572](https://issues.jboss.org/browse/THREESCALE-2572)

### DIFF
--- a/app/lib/three_scale/analytics/user_tracking.rb
+++ b/app/lib/three_scale/analytics/user_tracking.rb
@@ -14,7 +14,7 @@ module ThreeScale
 
       class TrackingAdapter
         DELEGATED_METHODS = %i(flush track identify group).freeze
-        delegate(*DELEGATED_METHODS, to: :@adapter)
+        delegate(*DELEGATED_METHODS, to: :adapter)
 
         class NullAdapter
           def with_options(*)
@@ -31,11 +31,19 @@ module ThreeScale
         end
 
         def segment_configured?
-          defined?(::Segment)
+          config.enabled && defined?(::Segment)
         end
 
         def initialize(config = {})
-          @adapter = segment_configured? ? ::Segment::Analytics.new(config) : NullAdapter.new
+          @config = config
+        end
+
+        private
+
+        attr_reader :config
+
+        def adapter
+          @adapter ||= segment_configured? ? ::Segment::Analytics.new(config) : NullAdapter.new
         end
       end
 
--- a/test/unit/three_scale/analytics/user_tracking_test.rb
+++ b/test/unit/three_scale/analytics/user_tracking_test.rb
@@ -8,125 +8,178 @@ class ThreeScale::Analytics::UserTrackingTest < ActiveSupport::TestCase
     @user = FactoryBot.build_stubbed(:user, account: account, email: 'foo@example.net')
   end
 
-  def test_track_uses_segment
-    UserTracking::Segment.expects(:track).once
-    UserTracking.new(@user).track('foo')
+  def teardown
+    clean_segment_adapter
   end
 
-  def test_identify_uses_segment
-    UserTracking::Segment.expects(:identify).once
-    UserTracking.new(@user).identify(trait: 'value')
-  end
+  class SegmentEnabled < UserTrackingTest
+    def setup
+      super
+      System::Application.config.three_scale.segment.stubs(enabled: true)
+    end
 
-  def test_flush_uses_segment
-    UserTracking::Segment.expects(:flush).with().once
-    UserTracking.new(@user).flush
-  end
+    test 'track uses segment' do
+      UserTracking::Segment.expects(:track).once
+      UserTracking.new(@user).track('foo')
+    end
 
-  def test_group_uses_segment
-    UserTracking::Segment.expects(:group).once
-    UserTracking.new(@user).group
-  end
+    test 'identify uses segment' do
+      UserTracking::Segment.expects(:identify).once
+      UserTracking.new(@user).identify(trait: 'value')
+    end
 
-  def test_experiment
-    tracking = UserTracking.new(@user)
+    test 'flush uses segment' do
+      UserTracking::Segment.expects(:flush).once
+      UserTracking.new(@user).flush
+    end
 
-    UserTracking::Segment.expects(:identify)
-        .with(has_entry(traits: has_entry('Experiment: Some Name' => 'variant')))
-    UserTracking::Segment.expects(:track)
-        .with(has_entry(properties: { variation: 'variant'}))
+    test 'group uses segment' do
+      UserTracking::Segment.expects(:group).once
+      UserTracking.new(@user).group
+    end
 
-    tracking.experiment('Some Name', 'variant')
-  end
+    test 'experiment tracks and identify' do
+      tracking = UserTracking.new(@user)
 
-  def test_with_options
-    tracking = UserTracking.new(@user)
+      UserTracking::Segment.expects(:identify)
+          .with(has_entry(traits: has_entry('Experiment: Some Name' => 'variant')))
+      UserTracking::Segment.expects(:track)
+          .with(has_entry(properties: { variation: 'variant'}))
 
-    timestamp = 1.hour.from_now
+      tracking.experiment('Some Name', 'variant')
+    end
 
-    seq = sequence('segment')
+    test 'with options' do
+      tracking = UserTracking.new(@user)
 
-    UserTracking::Segment.expects(:identify).in_sequence(seq).with(has_entry(timestamp: timestamp))
+      timestamp = 1.hour.from_now
 
-    UserTracking::Segment.expects(:identify).in_sequence(seq).with(Not(has_key(:timestamp)))
+      seq = sequence('segment')
 
-    tracking.with_segment_options(timestamp: timestamp) do
+      UserTracking::Segment.expects(:identify).in_sequence(seq).with(has_entry(timestamp: timestamp))
+
+      UserTracking::Segment.expects(:identify).in_sequence(seq).with(Not(has_key(:timestamp)))
+
+      tracking.with_segment_options(timestamp: timestamp) do
+        tracking.identify
+      end
+
       tracking.identify
     end
 
-    tracking.identify
+    test 'basic traits' do
+      traits = UserTracking.new(nil).basic_traits
+      assert_equal Hash.new, traits
+
+      traits = UserTracking.new(@user).basic_traits
+
+      assert_equal traits.slice(:email), email: @user.email
+    end
+
+    test 'identify' do
+      tracking = UserTracking.new(nil)
+      refute tracking.identify
+
+      tracking = UserTracking.new(@user)
+      assert tracking.identify
+
+      @user.account.provider = false
+      refute tracking.identify
+    end
+
+    test 'track' do
+      tracking = UserTracking.new(nil)
+      refute tracking.track('Event')
+
+      tracking = UserTracking.new(@user)
+      assert tracking.track('Event')
+
+      @user.account.provider = false
+      refute tracking.track('Event')
+    end
+
+    test 'group' do
+      tracking = UserTracking.new(nil)
+      refute tracking.group
+
+      tracking = UserTracking.new(@user)
+      assert tracking.group
+
+      @user.account.provider = false
+      refute tracking.group
+    end
+
+    test 'group traits' do
+      tracking = UserTracking.new(nil)
+      assert_equal Hash.new, tracking.group_traits
+
+      tracking = UserTracking.new(@user)
+      @user.account.stubs(bought_plan: Plan.new(name: 'fooplan', cost_per_month: 42), state: 'suspended')
+
+      assert_equal({name: @user.account.name,
+                    plan: 'fooplan',
+                    monthly_spend: 42.0,
+                    license_mrr: 42.0,
+                    state: 'suspended',
+                   }, tracking.group_traits)
+    end
+
+    test 'can_send?' do
+      refute UserTracking.new(nil).can_send?
+      assert UserTracking.new(@user).can_send?
+    end
+
+    test 'only track providers' do
+      @user.account.provider = false
+      refute UserTracking.new(@user).can_send?
+    end
+
+    test 'do not track impersonation admins' do
+      @user.username = ThreeScale.config.impersonation_admin['username']
+      refute UserTracking.new(@user).can_send?
+    end
   end
 
-  def test_basic_traits
-    traits = UserTracking.new(nil).basic_traits
-    assert_equal Hash.new, traits
+  class SegmentDisabled < UserTrackingTest
+    def setup
+      super
+      System::Application.config.three_scale.segment.stubs(enabled: false)
+    end
 
-    traits = UserTracking.new(@user).basic_traits
+    test 'does not track' do
+      expect_not_contact_segment(:track)
 
-    assert_equal traits.slice(:email), email: @user.email
+      UserTracking.new(@user).track('foo')
+    end
+
+    test 'does not flush' do
+      expect_not_contact_segment(:flush)
+
+      UserTracking.new(@user).flush
+    end
+
+    test 'does not identify' do
+      expect_not_contact_segment(:identify)
+
+      UserTracking.new(@user).identify
+    end
+
+    test 'does not group' do
+      expect_not_contact_segment(:group)
+
+      UserTracking.new(@user).group
+    end
+
+    private
+
+    def expect_not_contact_segment(action_name)
+      UserTracking::TrackingAdapter::NullAdapter.any_instance.expects(action_name)
+    end
   end
 
-  def test_identify
-    tracking = UserTracking.new(nil)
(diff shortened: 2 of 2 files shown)


duduribeiro:
My first thought was to change https://github.com/3scale/porta/blob/df282667bac430e122538c7a10518565469a80e8/app/lib/three_scale/analytics.rb#L27 to check directly there (allowing future implementations of trackers to be avoided to send request), but if following this way, we will need to change https://github.com/3scale/porta/blob/df282667bac430e122538c7a10518565469a80e8/app/lib/three_scale/analytics/salesforce.rb#L5 to use ThreeScale::Analytics instead of the ThreeScale::Analytics::UserTracking implementation directly. what do you folks think? 🤔

test/unit/three_scale/analytics/user_tracking_test.rb
```diff
@@ -8,6 +8,32 @@ def setup
     @user = FactoryBot.build_stubbed(:user, account: account, email: 'foo@example.net')
   end
 
+  context 'On-Premises' do
```
duduribeiro: To avoid this context stuff, I thought on use 2 classes in the same file (what I honestly prefer 😄)

eg:

`test/unit/three_scale/analytics/user_tracking_test.rb`
```ruby
class ThreeScale::Analytics::UserTrackingOnPremisesTest < ActiveSupport::TestCase
...
end

class ThreeScale::Analytics::UserTrackingNotOnPremisesTest < ActiveSupport::TestCase
...
end
```

what do you folks prefer?

Martouta: I also prefer the 2 classes 👍 `NotOnPremises` should be `SaaS` 😂

hallelujah: We could call it with SegmentDisabled?
Better than on-prem or SaaS imo

Martouta: Yeah, I wanted to explain him about the whole features thing that we have but we got busy. I will write it before the 3pm sync though 👍

duduribeiro:
@Martouta / @hallelujah updated! thanks

app/lib/three_scale/analytics/user_tracking.rb
```diff
@@ -211,7 +211,7 @@ def cache_key
       end
 
       def can_send?
-        user_id && @account.try!(:provider?) && user_type != 'impersonation_admin'
+        !ThreeScale.config.onpremises && user_id && @account.try!(:provider?) && user_type != 'impersonation_admin'
```
hallelujah: Can we know if Segment is properly configured instead of using this (do not use anymore `ThreeScale.config.onpremises`) ?

duduribeiro: @hallelujah maybe pinging segment only once to check if it is enabled?

Martouta: no, it is easier 😄 but i didn't have time to explain yet 😂

Martouta: I will explain in the sync @duduribeiro

test/unit/three_scale/analytics/user_tracking_test.rb
```diff
@@ -8,125 +8,155 @@ def setup
...
+    def test_does_not_track
+      ThreeScale::Analytics::UserTracking::TrackingAdapter::NullAdapter.any_instance.expects(:track).once
+      # ThreeScale::Analytics::UserTracking::TrackingAdapter::NullAdapter.any_instance.expects(:track).once
+      UserTracking.new(@user).track('foo')
+    end
 
-  def test_do_not_track_impersonation_admins
-    @user.username = ThreeScale.config.impersonation_admin['username']
-    refute UserTracking.new(@user).can_send?
+    # def test_does_not_identify
```
hallelujah: What about those tests? Are they needed? If so will you fix them or do you think it is unneeded then remove>

duduribeiro: Me & @Martouta are playing with some tests.. We will fix it 😄 ..

duduribeiro: will add a wip tag on the PR

duduribeiro:
hey folks @3scale/system 

we changed it to check a configuration instead of checking if it is on-Prem. Can you folks do another round of review?

thanks

app/lib/three_scale/analytics/user_tracking.rb
```diff
@@ -31,11 +31,19 @@ def method_missing(method_sym, *args, &block)
...
-          @adapter = segment_configured? ? ::Segment::Analytics.new(config) : NullAdapter.new
+          @config = config
+        end
+
+        private
+
+        attr_reader :config
+
+        def adapter
+          segment_configured? ? ::Segment::Analytics.new(config) : NullAdapter.new
```
Martouta: I removed the memoization of this to make the tests pass but I am not sure if it is a good idea or it will make it considerably slower 🤔 the actual root cause of being hard to test and working without this memoization is this line:
https://github.com/3scale/porta/blob/06948bbd7118be15b77e561e269fa6b4878fb2bf/app/lib/three_scale/analytics/user_tracking.rb#L43

Any reviewer should give an opinion and ideas 😝

app/lib/three_scale/analytics/user_tracking.rb
```diff
@@ -31,11 +31,19 @@ def method_missing(method_sym, *args, &block)
         end
 
         def segment_configured?
-          defined?(::Segment)
+          Features::SegmentConfig.enabled? && defined?(::Segment)
```
Martouta: As CodeClimate says, this can become a class method instead of being an instance method 😄

app/lib/three_scale/analytics/user_tracking.rb
```diff
@@ -31,11 +31,19 @@ def method_missing(method_sym, *args, &block)
         end
 
         def segment_configured?
```
Martouta: As CodeClimate says, this can become a class method instead of being an instance method 😄 Anyway, not your fault 👍

Martouta:
@duduribeiro You need to update also the `features.yml` from [the deploy repository](https://github.com/3scale/deploy/blob/master/system/data/production/features.yml) because that one is the one we use in production 😄

app/lib/three_scale/analytics/user_tracking.rb
hallelujah: Why are the test failing?

app/lib/three_scale/analytics/user_tracking.rb
```diff
@@ -30,12 +30,20 @@ def method_missing(method_sym, *args, &block)
...
-          @adapter = segment_configured? ? ::Segment::Analytics.new(config) : NullAdapter.new
+          @config = config
+        end
+
+        private
+
+        attr_reader :config
+
+        def adapter
+          @adapter ||= TrackingAdapter.segment_configured? ? ::Segment::Analytics.new(config) : NullAdapter.new
```
duduribeiro: @Martouta @hallelujah Since the only place that instantiate the is the constant assignment (https://github.com/3scale/porta/pull/903/files#diff-d32b184052c26bb111d811f12f0a2e9aR51), I think we do not need the adapter to be a class method. Making the @adapter an instance variable we can memoize it, since `Segment` will only have one instance of `TrackingAdapter`.

What do you folks think? 🤔

Martouta: @duduribeiro I agree that `adapter` needs to be an instance method because it relies on `config` 😄

app/lib/three_scale/analytics/user_tracking.rb
```diff
@@ -231,7 +239,7 @@ def identified?
 
       def segment_client(options = {})
         merged_options = segment_options.deep_merge(options)
-        Segment.with_options(merged_options) { |segment| return segment }
+        Segment.with_options(merged_options) { |segment| segment }
```
hallelujah: Hmmm wait, `return` will make the calling method return so careful with that!

config/examples/features.yml
```diff
@@ -6,6 +6,8 @@ features: &default
     contract_unpaid_time: 183
     disabled_for_app_plans:
     - '%enterprise%'
+  segment:
+    enabled: true
```
hallelujah: I'd not enable it for examples.

test/unit/three_scale/analytics/user_tracking_test.rb
```diff
@@ -1,4 +1,5 @@
 require 'test_helper'
+require 'segment/analytics'
```
hallelujah: Why do we need this require?

app/lib/three_scale/analytics/user_tracking.rb
Martouta: This definitely requires a test 😝

config/examples/features.yml
Martouta: I agree with @hallelujah 😄

test/unit/three_scale/analytics/user_tracking_test.rb
duduribeiro: Since here (https://github.com/3scale/porta/pull/903/files#diff-757942a1b9ceda38bc0b76acea39c446R175) we expected that `Segment::Analytics` never receives the action method, we need to require it, otherwise it fails since it doesn't have it loaded yet.

Martouta: It works for me 🤔 
image

duduribeiro: yeah.. worked for me.. I don't know why it failed the first time haha.. thanks @Martouta

app/lib/three_scale/analytics/user_tracking.rb
guicassolato: Better using `self.class` than `TrackingAdapter`, or you can
```ruby
delegate :segment_configured?, to: :class
```
and call it in the instance.

test/unit/three_scale/analytics/user_tracking_test.rb
```diff
@@ -8,125 +8,176 @@ def setup
...
-  def test_only_track_providers
-    @user.account.provider = false
-    refute UserTracking.new(@user).can_send?
+  class SegmentDisabled < UserTrackingTest
+    def setup
+      super
+      Features::SegmentConfig.stubs(enabled?: false)
+    end
+
+    def test_does_not_track
```
guicassolato: You can use the DSL-ish form if you prefer

```ruby
test 'it does not track' do
  ...
end
```

app/lib/three_scale/analytics/user_tracking.rb
Martouta: I agree so we can override it by a subclass if ever needed 😄

test/unit/three_scale/analytics/user_tracking_test.rb
```diff
@@ -8,125 +8,176 @@ def setup
...
+      expect_not_contact_segment(:group)
+
+      UserTracking.new(@user).group
+    end
+
+    private
+
+    def expect_not_contact_segment(action_name)
+      UserTracking::TrackingAdapter::NullAdapter.any_instance.expects(action_name).once
+      Segment::Analytics.any_instance.expects(action_name).never
```
guicassolato: We all do the same despite knowing that `expects(...).never` tends to lead to outdated tests. However, it makes me think whether we should add a log message when `UserTracking` methods are used and segment is disabled, and then assert based on that log message instead of using `never`. WDYT?

test/unit/three_scale/analytics/user_tracking_test.rb
```diff
@@ -8,125 +8,176 @@ def setup
...
 
-  def test_track_uses_segment
-    UserTracking::Segment.expects(:track).once
-    UserTracking.new(@user).track('foo')
-  end
+  class SegmentEnabled < UserTrackingTest
+    def setup
+      super
+      Features::SegmentConfig.stubs(enabled?: true)
+      clean_segment_adapter
```
Martouta: Imo this line should be a teardown of `UserTrackingTest` 🤔 because:

1. We already expect it to be clean before starting, the thing is that we want to clean it at the end too so we don't unexpectedly affect other tests.
2. In `UserTrackingTest` because it affects both subclasses

test/unit/three_scale/analytics/user_tracking_test.rb
```diff
@@ -8,125 +8,176 @@ def setup
...
+    end
   end
 
-  def test_do_not_track_impersonation_admins
-    @user.username = ThreeScale.config.impersonation_admin['username']
-    refute UserTracking.new(@user).can_send?
+  private
+
+  def clean_segment_adapter
+    ThreeScale::Analytics::UserTracking::Segment.instance_variable_set(:@adapter, nil)
```
Martouta: Accessing to private instances from another class is not ideal. Imo acceptable that its own unit test knows a lot about the class but still risky... And knowing that the root cause is the constant `Segment = ...` I am thinking if we can do something else 🤔 probably make that constant actually be another memoized class method 🤔 I will check it and write here again 👍

Martouta:
Ok my bad, as Segment is an old feature we have, we have it outside `features.yml`, in its own `segment.yml`, and if it is enabled or not is already in its own config as well:
https://github.com/3scale/porta/blob/c68e02f2a09ffdfc0181c9f945fced3bca6cf807/config/examples/segment.yml#L2

app/lib/three_scale/analytics/user_tracking.rb
```diff
@@ -26,16 +27,25 @@ def respond_to_missing?(method_sym, include_all)
           end
 
           def method_missing(method_sym, *args, &block)
+            Rails.logger.info("#{method_sym} call avoided. Segment disabled")
```
duduribeiro: This was made to follow @guicassolato suggestion (https://github.com/3scale/porta/pull/903#discussion_r295384203)

guicassolato: Probably only when one of the delegated methods is invoked, no?

app/lib/three_scale/analytics/user_tracking.rb
Martouta: Response by Carlos (in case that anyone reads this in the future):

> let supose that the first test that runs was the one with segment enabled. The `Segment` constant assigned to the class will be the `::Segment::Analytics.new(config)` adapter
> and we can’t test the `non segment` after it.. because the constant was already assigned

app/lib/three_scale/analytics/user_tracking.rb
Martouta: @guicassolato Then why not to test that the method is invoked in the proper class and that's it?
because that is actually what we want. Like:
`NullAdapter.any_instance.expects(action_name).once`

guicassolato: There are 3 different things:
1. `UserTracking::TrackingAdapter.any_instance.expects(action_name).never` – weak but it doesn't really hurt having it. We can keep it in the test.
2. `UserTracking::TrackingAdapter::NullAdapter.any_instance.expects(action_name)` – you are right, @Martouta! This is better than testing the log.
3. Logging message in case the method is invoked when Segment integration is disabled – about this... now I'm thinking... we don't really prevent the method from being called, right? It's just a no-op when called on a `NullAdapter` instance. If this is correct, then better without the logs; otherwise, we will have a bunch of useless information dumped to stdout. On the other hand, if we DO prevent the integration from being used in the first place and those method should not even be invoked, then the logs can help to realize that something is wrong.

Martouta: Like this I mean @guicassolato @duduribeiro : 

image

image

duduribeiro: Makes total sense. Awesome 💯

app/lib/three_scale/analytics/user_tracking.rb
```diff
@@ -26,16 +27,26 @@ def respond_to_missing?(method_sym, include_all)
...
+            return super unless DELEGATED_METHODS.include?(method_sym)
+
+            Rails.logger.info("#{method_sym} call avoided. Segment disabled")
           end
         end
 
-        def segment_configured?
-          defined?(::Segment)
+        def self.segment_configured?
+          System::Application.config.three_scale.segment.enabled && defined?(::Segment)
```
Martouta: I think this should be from the same config that it has in the instance, like this:

```ruby
def segment_configured?
  config.enabled && defined?(::Segment)
end
```

image

image

duduribeiro: 💯

app/lib/three_scale/analytics/user_tracking.rb
duduribeiro: I added the test for NullAdapter expecting something. And I totally agree the expects().never is a weak test because we can easily forget about it.

PR #905, opened 2019-06-17
Fix Master Header's height

josemigallas:
`calc` is used with wrong syntax so it is not compiled properly.

### DIFF
--- a/app/assets/stylesheets/provider/_header.scss
+++ b/app/assets/stylesheets/provider/_header.scss
@@ -1,11 +1,12 @@
+$header-height: line-height-times(63 / 20);
 $master-ribbon-height: line-height-times(1);
 
 .Header {
 
-  height: 76px; // HACK: hardcoded value of PF4 Header hight.
+  height: $header-height; 
 
   &--master {
-    height: calc(76px + $master-ribbon-height); // HACK: hardcoded value of PF4 Header hight.
+    height: $header-height + $master-ribbon-height;
     border-top: $master-ribbon-height solid $header-border-color;
 
     &:before {


app/assets/stylesheets/provider/_header.scss
```diff
@@ -5,7 +5,7 @@ $master-ribbon-height: line-height-times(1);
   height: 76px; // HACK: hardcoded value of PF4 Header hight.
 
   &--master {
-    height: calc(76px + $master-ribbon-height); // HACK: hardcoded value of PF4 Header hight.
+    height: calc(76px + #{$master-ribbon-height}); // HACK: hardcoded value of PF4 Header hight.
```
thomasmaas: so the problem is using px as that won't scale the height

thomasmaas: that's why we use `height: line-height-times(63 / 20);` in currently deployed css

josemigallas: Do you mean we use that instead of the whole `calc`? But it's be better to have the height as a constant.

thomasmaas: - I don't think calc is needed as that would move the operation to the browser. Just sass arithmetics will do `+`
- Heights are defined in `line-height-times` so let's stick with that.

$ribbon-height: line-height-times(…)
$master-ribbon-height: $ribbon-height + line-height-times(…)

josemigallas: OK, header's height can be in line-height-times but the exact PF value is 76px.

app/assets/stylesheets/provider/_header.scss
```diff
@@ -1,11 +1,12 @@
+$header-hight: line-height-times(89 / 25); // HACK: hardcoded value of PF4 Header hight (76px)
```
thomasmaas: height

thomasmaas: And please lets keep the height exactly as it is now.

thomasmaas: And please lets keep the height exactly as it is now.

PR #813, opened 2019-05-20
Fix empty or corrupted tenant_id

Martouta:
Closes [THREESCALE-2571 - Fix users without tenant_id set](https://issues.jboss.org/browse/THREESCALE-2571)

The code that we currently have to reset the `tenant_id` updates all records of the DB ([here](https://github.com/3scale/porta/blob/f591474f5e7b00b99ac41b026936616c237ca309/lib/tasks/multitenant/triggers.rake#L41-L115)), not only the ones that are NULL. Besides, it does not rely on indexes. This is not ideal. Just in case, I tried executing it in preview with ops supervision, and it did not went well indeed. I am not really considering refactoring though:
1. That would be a lot of work in comparison to what we need to close the issue
2. It's technically out of scope.
3. There is already a refactoring in progress in [this old PR](https://github.com/3scale/porta/pull/598) which should have its own Jira issue opened.

To sum up, I will just do a simple task from this to update only the records that we currently need (those that have `tenant_id` as `nil` or that were updated on 2019-01-24). For all relevant tables affected by this day.

This code is meant to be a one time execution, and it is optimised for it. The conclusions are in https://github.com/3scale/infrastructure/issues/693
We will execute it with these params:
```
RAILS_ENV=production FILE=corrupted_accounts bundle exec rake 'multitenant:fix_corrupted_tenant_id_accounts[1, 1]'
RAILS_ENV=production FILE=corrupted_accounts bundle exec rake 'multitenant:fix_corrupted_tenant_id_for_table_associated_to_account[User, 01/24/2019 08:00 UTC, 01/24/2019 16:30 UTC, 1, 1]'
RAILS_ENV=production FILE=corrupted_accounts bundle exec rake 'multitenant:fix_corrupted_tenant_id_for_table_associated_to_account[Settings, 01/24/2019 08:00 UTC, 01/24/2019 16:30 UTC, 100, 1]'
RAILS_ENV=production FILE=corrupted_accounts bundle exec rake 'multitenant:fix_corrupted_tenant_id_for_table_associated_to_account[Profile, 01/24/2019 08:00 UTC, 01/24/2019 16:30 UTC, 100, 1]'
RAILS_ENV=production FILE=corrupted_accounts bundle exec rake 'multitenant:fix_corrupted_tenant_id_for_table_associated_to_account[PaymentDetail, 01/24/2019 08:00 UTC, 01/24/2019 16:30 UTC, 100, 1]'
[...]

PR #906, opened 2019-06-17
[317-custom-http-response] usage limit exceeded error

macejmic:

Custom HTTP status and header for Usage limit exceeded error

https://issues.jboss.org/browse/THREESCALE-317

### DIFF
--- a/app/models/proxy.rb
+++ b/app/models/proxy.rb
@@ -15,7 +15,7 @@ class Proxy < ApplicationRecord
   has_one :oidc_configuration, dependent: :delete, inverse_of: :oidc_configurable, as: :oidc_configurable
   accepts_nested_attributes_for :oidc_configuration
 
-  validates :api_backend, :error_status_no_match, :error_status_auth_missing, :error_status_auth_failed, presence: true
+  validates :api_backend, :error_status_no_match, :error_status_auth_missing, :error_status_auth_failed, :error_status_limits_exceeded, presence: true
 
   uri_pattern = URI::DEFAULT_PARSER.pattern
 
@@ -51,7 +51,7 @@ class Proxy < ApplicationRecord
 
   validates :credentials_location, inclusion: { in: %w[headers query authorization], allow_nil: false }
 
-  validates :error_status_no_match, :error_status_auth_missing, :error_status_auth_failed,
+  validates :error_status_no_match, :error_status_auth_missing, :error_status_auth_failed, :error_status_limits_exceeded,
                             numericality: { greater_than_or_equal_to: 200, less_than: 600 }
 
   validates :auth_app_id, :auth_app_key, :auth_user_key, :secret_token,
@@ -59,16 +59,16 @@ class Proxy < ApplicationRecord
 
   validates :error_headers_auth_failed,
                       :error_headers_auth_missing, :error_headers_no_match,
-                      :error_auth_failed,
-                      :error_auth_missing, :error_no_match,
+                      :error_auth_failed, :error_auth_missing, :error_no_match,
+                      :error_headers_limits_exceeded, :error_limits_exceeded,
                       format: { with: HTTP_HEADER }
 
   validates :api_test_path, length: { maximum: 8192 }
   validates :endpoint, :api_backend, :auth_app_key, :auth_app_id, :auth_user_key,
             :oidc_issuer_endpoint, :oidc_issuer_type,
             :credentials_location, :error_auth_failed, :error_auth_missing, :authentication_method,
-            :error_headers_auth_failed, :error_headers_auth_missing, :error_no_match,
-            :error_headers_no_match, :secret_token, :hostname_rewrite, :sandbox_endpoint,
+            :error_headers_auth_failed, :error_headers_auth_missing, :error_headers_limits_exceeded, :error_limits_exceeded,
+            :error_no_match, :error_headers_no_match, :secret_token, :hostname_rewrite, :sandbox_endpoint,
             length: { maximum: 255 }
 
   validates :oidc_issuer_type, inclusion: { in: OIDC_ISSUER_TYPES.keys.map(&:to_s), allow_blank: true }, presence: { if: ->(proxy) { proxy.oidc_issuer_endpoint.present? } }
--- a/db/migrate/20190617100508_add_limit_exceeded_error_to_proxies.rb
+++ b/db/migrate/20190617100508_add_limit_exceeded_error_to_proxies.rb
@@ -1,0 +1,18 @@
+class AddLimitExceededErrorToProxies < ActiveRecord::Migration
+  def up
+    add_column :proxies, :error_headers_limits_exceeded, :string
+    change_column_default :proxies, :error_headers_limits_exceeded, 'text/plain; charset=us-ascii'
+
+    add_column :proxies, :error_status_limits_exceeded, :integer
+    change_column_default :proxies, :error_status_limits_exceeded, 429
+
+    add_column :proxies, :error_limits_exceeded, :string
+    change_column_default :proxies, :error_limits_exceeded, 'Usage limit exceeded'
+  end
+
+  def down
+    remove_column :proxies, :error_headers_limits_exceeded
+    remove_column :proxies, :error_status_limits_exceeded
+    remove_column :proxies, :error_limits_exceeded
+  end
+end
--- a/db/migrate/20190620092211_backfill_add_limit_exceeded_error_to_proxies.rb
+++ b/db/migrate/20190620092211_backfill_add_limit_exceeded_error_to_proxies.rb
@@ -1,0 +1,17 @@
+class BackfillAddLimitExceededErrorToProxies < ActiveRecord::Migration
+  disable_ddl_transaction!
+
+  def up
+    Proxy.select(:id).find_in_batches(batch_size: 100) do |batch|
+      batch.each { |proxy| proxy.update_column(:error_headers_limits_exceeded, 'text/plain; charset=us-ascii') }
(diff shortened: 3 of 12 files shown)


db/migrate/20190617100508_add_limit_exceeded_error_to_proxies.rb
```diff
@@ -0,0 +1,18 @@
+class AddLimitExceededErrorToProxies < ActiveRecord::Migration
+  def up
```
Martouta: Why not like this?

```ruby
class AddLimitExceededErrorToProxies < ActiveRecord::Migration
  def change
    safety_assured { add_column :proxies, :error_headers_limits_exceeded, :string, null: false, default: 'text/plain; charset=us-ascii' }
    safety_assured { add_column :proxies, :error_status_limits_exceeded, :integer, null: false, default: 429 }
    safety_assured { add_column :proxies, :error_limits_exceeded, :string, null: false, default: 'Usage limit exceeded' }
  end
end
```

I have tried it and it works:
image

Either way we need to be careful because:
1. It affects all current records, so we need to be careful in the way of doing in SaaS
2. WE need to run the migration for our 3 DBs 😄

app/models/proxy.rb
```diff
@@ -51,23 +51,23 @@ class Proxy < ApplicationRecord
...
+                      :error_auth_failed, :error_auth_missing, :error_no_match,
+                      :error_headers_limits_exceeded, :error_limits_exceeded,
                       format: { with: HTTP_HEADER }
 
   validates :api_test_path, length: { maximum: 8192 }
   validates :endpoint, :api_backend, :auth_app_key, :auth_app_id, :auth_user_key,
             :oidc_issuer_endpoint, :oidc_issuer_type,
             :credentials_location, :error_auth_failed, :error_auth_missing, :authentication_method,
-            :error_headers_auth_failed, :error_headers_auth_missing, :error_no_match,
+            :error_headers_auth_failed, :error_headers_auth_missing, :error_headers_limits_exceeded, :error_no_match,
```
Martouta: you will need also `error_limits_exceeded` here to make the tests pass 😄

macejmic: true.

db/migrate/20190617100508_add_limit_exceeded_error_to_proxies.rb
macejmic: ok, that could wok :) thanks!

test/functional/api/integrations_controller_test.rb
```diff
@@ -215,6 +215,9 @@ def setup
         "error_auth_missing"=>"ooooh Authentication parameters missing",
         "error_status_no_match"=>504,
         "error_headers_no_match"=>"text/html; charset=us-ascii",
+        "error_status_limits_exceeded"=>499,
+        "error_headers_limits_exceeded"=>"text/html; charset=us-ascii",
+        "error_limits_exceeded"=>"Limit exceeeeeded",
```
guicassolato: We're trying not to depend only on functional tests as much as possible. Maybe add something specific to these new attributes in integration tests similar to https://github.com/3scale/porta/blob/06948bbd7118be15b77e561e269fa6b4878fb2bf/test/integration/proxy_example_test.rb#L599?

macejmic: ok, makes sense.

macejmic: added test for lua generator..

db/migrate/20190617100508_add_limit_exceeded_error_to_proxies.rb
hallelujah: Well this is explicitely saying that even you are doing a dangerous migration, you ack that and want to run it in production.
We should not do that but instead run the migrations with `SAFETY_ASSURED=1` environment when running them

db/migrate/20190617100508_add_limit_exceeded_error_to_proxies.rb
```diff
@@ -0,0 +1,7 @@
+class AddLimitExceededErrorToProxies < ActiveRecord::Migration
+  def change
+    safety_assured { add_column :proxies, :error_headers_limits_exceeded, :string, default: 'text/plain; charset=us-ascii' }
```
hallelujah: `safety_assured` is only needed when you are sure everything was assured before that it will not be dangerous anymore

db/migrate/20190617100508_add_limit_exceeded_error_to_proxies.rb
Martouta: I said that because I was expecting letting ops doing it in saas and then let's for on-prem actually run the migrations, but something else that we can do is actually do it non-blocking ourselves, and by that I mean doing this in the migration(s):
1. A migration that adds the column but at first with all records to null
2. In that migration, make the validation for not null and adding the default value to all records but one by one in batches and with a waiting time between batches... maybe ops should validate that this is a nice idea.

db/migrate/20190617100508_add_limit_exceeded_error_to_proxies.rb
Martouta: we should have this conversation in 2 threads 😕 
let's talk about it either here or in https://github.com/3scale/porta/pull/906#discussion_r295167579

db/migrate/20190617100508_add_limit_exceeded_error_to_proxies.rb
hallelujah: Yes do it non blocking, NULL with default.

The correct way to do things is:

1. writing migration normally, without `safety_assured`
2. run the migration on development environment
3. if it detects a dangerous migration, it will raise an error

In case of dangerous migration detected, do the appropriate measure described https://github.com/ankane/strong_migrations#strong-migrations

db/migrate/20190617100508_add_limit_exceeded_error_to_proxies.rb
```diff
@@ -0,0 +1,7 @@
+class AddLimitExceededErrorToProxies < ActiveRecord::Migration
+  def change
+    add_column :proxies, :error_headers_limits_exceeded, :string, default: 'text/plain; charset=us-ascii'
```
Martouta: This is still blocking 😞 
image

I propose to do as it says there but the 2nd migration to be:
```ruby
class BackfillAddLimitExceededErrorToProxies < ActiveRecord::Migration
  def up
    Proxy.find_in_batches(batch_size: 100) do |batch|
      batch.each { |proxy| proxy.update_column(error_headers_limits_exceeded: 'text/plain; charset=us-ascii') }
      sleep(1) # So we let customers do requests to the DB instead of blocking it for long. This is because it needs to work for all DBs
    end
  end

  def down
    Proxy.find_in_batches(batch_size: 100) do |batch|
      batch.each { |proxy| proxy.update_column(error_headers_limits_exceeded: nil) }
      sleep(1) # So we let customers do requests to the DB instead of blocking it for long. This is because it needs to work for all DBs
    end
  end
end
```

macejmic: why find in batches and not `Proxy.update_all` ? without loading everything..

Martouta: Because when I did [the PR of the rake tasks to fix empty or corrupted tenant_id in the DB](https://github.com/3scale/porta/pull/813), I made some tests with operations in preview, and doing `update_all` for literally all the records of a table means it takes very long and in that time the requests of customers are waiting because these are locked... 😞 so we changes it to be like this (updating one by one so customer requests can take a spot in between those) and with a waiting time of 1 second without doing anything so that time is only for customers.
But you should probably speak with operations to see if that is ideal for this case too or if they have a better idea in mind 😄

db/migrate/20190620092211_backfill_add_limit_exceeded_error_to_proxies.rb
```diff
@@ -0,0 +1,17 @@
+class BackfillAddLimitExceededErrorToProxies < ActiveRecord::Migration
+  disable_ddl_transaction!
+
+  def up
+    Proxy.select(:id).find_in_batches(batch_size: 100) do |batch|
+      batch.each { |proxy| proxy.update_column(:error_headers_limits_exceeded, 'text/plain; charset=us-ascii') }
+      sleep(1)
```
hallelujah: In SaaS we have around 33k proxies so it will last ~6 minutes
Can we lower that to 0.5 ?

db/schema.rb
```diff
@@ -1051,38 +1051,44 @@
...
-    t.text     "policies_config",              limit: 65535
-    t.string   "oidc_issuer_type",             limit: 255,   default: "keycloak"
+    t.boolean  "apicast_configuration_driven",                default: true,                                null: false
+    t.string   "oidc_issuer_endpoint",          limit: 255
+    t.integer  "lock_version",                  limit: 8,     default: 0,                                   null: false
+    t.string   "authentication_method",         limit: 255
+    t.text     "policies_config",               limit: 65535
+    t.string   "error_headers_limit_exceeded",  limit: 255,   default: "text/plain; charset=us-ascii",      null: false
+    t.string   "error_status_limit_exceeded",   limit: 255,   default: "429",                               null: false
+    t.string   "error_limit_exceeded",          limit: 255,   default: "Usage limit exceeded",              null: false
```
Martouta: lol a `error_limit_exceed` here and a `error_limits_exceed` in line 1091 😂

Martouta: Fixed locally 😄 I will force push 😉

PR #908, opened 2019-06-17
Local image build on a container

hallelujah:
Just a re-birth of the Dockerfile

It needs polishing

/cc @duduribeiro


Dockerfile
```diff
@@ -65,3 +53,6 @@ RUN if [ "${DB}" = "oracle" ]; then unzip /opt/oracle/instantclient-basiclite-li
  && (cd /opt/oracle/instantclient_12_2/ && ln -s libclntsh.so.12.1 libclntsh.so) \
  && rm -rf /opt/system/vendor/oracle \
  && rm -rf /opt/oracle/*.zip; fi
+
+USER 1001
+RUN bash -c "bundle install && npm install"
```
hallelujah: Probably installing the gem as root but meaning it should know scl-enable stuffs ... which IMO is not the right thing to do.

Dockerfile
```diff
@@ -1,31 +1,21 @@
-FROM quay.io/3scale/docker:ci-2.3.1-3
+FROM quay.io/3scale/system-builder:ruby24
```
hallelujah: Yes we are moving from 2.3 but latest is not yet 2.4 :/

Dockerfile
```diff
@@ -1,31 +1,21 @@
...
- && gem install bundler --version ${BUNDLER_VERSION} --no-document \
- && sed --in-place "s/databases 16/databases 32/" /etc/redis/redis.conf \
- && echo 'dns_nameservers 8.8.8.8 8.8.4.4' >> /etc/squid3/squid.conf \
- && cd /tmp && curl -o sphinxsearch.deb -J -L -O https://github.com/sphinxsearch/sphinx/releases/download/${SPHINX_VERSION}-release/sphinxsearch_${SPHINX_VERSION}-release-1.precise_amd64.deb \
- && curl -o unixODBC.deb -J -L -O https://github.com/3scale/unixODBC/releases/download/2.3.6-ubuntu-12.04/unixODBC_2.3.6_amd64.deb \
- && apt-install libodbc1 \
- && dpkg --install sphinxsearch.deb \
- && dpkg --install unixODBC.deb \
- && apt-get autoremove -y \
- && rm -f sphinxsearch_${SPHINX_VERSION}.deb unixODBC.deb
```
hallelujah: Sphinx is already in the system-builder

docker-compose.yml
```diff
@@ -1,44 +1,96 @@
...
+
+version: '3'
 services:
-  keycloak:
-    image: jboss/keycloak:3.0.0.Final
-    environment:
-      KEYCLOAK_USER: keycloak
-      KEYCLOAK_PASSWORD: development
-      KEYCLOAK_LOGLEVEL: DEBUG
+  system:
```
duduribeiro: @hallelujah the dependencies are already starting with the compose. I'm only having problem running the built image for system (the bundle install commands run perfectly but don't install nothing on vendor/bundle). Will look into it tomorrow 🍻

Dockerfile
```diff
@@ -1,67 +1,77 @@
-FROM quay.io/3scale/docker:ci-2.3.1-3
+FROM centos/ruby-24-centos7
```
hallelujah: Base it on system-builder

Dockerfile
```diff
@@ -1,67 +1,77 @@
...
-deb http://archive.ubuntu.com/ubuntu precise-updates main restricted universe multiverse\n\
-deb http://archive.ubuntu.com/ubuntu precise-backports main restricted universe multiverse\n\
-deb http://archive.ubuntu.com/ubuntu precise-security main restricted universe multiverse" > /etc/apt/sources.list
+ARG BUNDLER_ENV
+ENV BUNDLER_ENV="${BUNDLER_ENV}" \
+    TZ=:/etc/localtime \
+    BUNDLE_GEMFILE=Gemfile \
+    BUNDLE_JOBS=5 \
+    BUNDLE_WITHOUT=development:test \
+    NODEJS_SCL=rh-nodejs8
```
hallelujah: Now I think system-builder is already setting some of those environments so no need of them

Dockerfile
hallelujah: https://github.com/3scale/system-builder/blob/master/Dockerfile

Dockerfile
```diff
@@ -1,62 +1,40 @@
...
+ARG USER_PASSWORD=p
+
+ENV DISABLE_SPRING="true" \
+    ORACLE_SYSTEM_PASSWORD="threescalepass" \
+    NLS_LANG="AMERICAN_AMERICA.UTF8" \
+    TZ="UTC" \
+    MASTER_PASSWORD="${MASTER_PASSWORD}" \
+    USER_PASSWORD="${USER_PASSWORD}" \
+    LC_ALL="en_US.UTF-8" \
+    PATH="./node_modules/.bin:$PATH:/usr/local/nginx/sbin/" \
```
hallelujah: I wonder why we need nginx?

Dockerfile
```diff
@@ -65,3 +43,12 @@ RUN if [ "${DB}" = "oracle" ]; then unzip /opt/oracle/instantclient-basiclite-li
  && (cd /opt/oracle/instantclient_12_2/ && ln -s libclntsh.so.12.1 libclntsh.so) \
  && rm -rf /opt/system/vendor/oracle \
  && rm -rf /opt/oracle/*.zip; fi
+
+USER root
+RUN mkdir -p  /opt/system/tmp/sessions/ \
```
hallelujah: `rake tmp:create` should do the job right?

duduribeiro: @hallelujah the problem is that the gems are not installed for root and we would need to install it again. So it fails.

Dockerfile
duduribeiro: 💯 .. I think it is fine to remove.. will do it

Dockerfile
duduribeiro: awesome!

duduribeiro:
@hallelujah updated it with the last comments 🍻

docker-compose.yml
```diff
@@ -1,44 +1,99 @@
...
+      - CONFIG_QUEUES_MASTER_NAME=backend_redis:6379
+      - CONFIG_WORKERS_LOG_FILE=/dev/stdout
+      - CONFIG_INTERNAL_API_USER=system_app
+      - CONFIG_INTERNAL_API_PASSWORD=password
+      - RACK_ENV=production
+
   redis:
-    image: redis:3.2.5-alpine
-    container_name: redis
+    image: redis
```
hallelujah: you can use the tag `alpine` as it is smaller

docker-compose.yml
```diff
@@ -1,44 +1,99 @@
...
+  links:
+    - apisonator
+    - mailhog
+    - memcached
+    - mysql
+    - redis
+  environment:
+    - DATABASE_URL=mysql2://root@mysql/3scale_system_development
+    - REDIS_URL=redis://redis/1
+    - MEMCACHE_SERVERS=memcached:11211
```
hallelujah: I am not convinced of using the environment here.

docker-compose.yml
```diff
@@ -1,44 +1,99 @@
...
+
+  apisonator:
+    image: quay.io/3scale/apisonator:latest
+    container_name: apisonator_compose
+    ports:
+      - "3001:3001"
+    command: "3scale_backend start -p 3001"
     depends_on:
-      - redis
+      - backend_redis
```
hallelujah: We could use the same redis as system to save resources

docker-compose.yml
```diff
@@ -1,44 +1,99 @@
...
     ports:
       - "6379:6379"
-  apicast:
-    image: quay.io/3scale/apicast:3.3-stable
-    env_file: .env.apicast
-    depends_on:
-      - backend-listener
+
+  backend_redis:
+    image: redis
```
hallelujah: Same `redis:alpine`

Gemfile.base
```diff
@@ -88,6 +88,7 @@ gem '3scale_client', '~> 2.6.1', require: false
 gem 'analytics-ruby', require: false
 
 group :development, :test do
+  gem 'activerecord-nulldb-adapter'
```
duduribeiro: this was necessary to precompile the assets in the docker build phase. Since when we are building the image, the database does not exist and webpacker:compile needs the database up,

INSTALL.md
```diff
@@ -36,6 +36,22 @@ make bash
...
 All the source and dependencies for this project will be in place, making it possible to run porta and the tests from inside the container. See [Run Porta](#run-porta)
 
+#### Running the application
+
+If you want to run the application using docker, you can use the docker compose file to start all dependencies.
+
+In the first time, you need to setup the database. To configure it, run: (the Master and User password will be `p`)
+
+```
+MASTER_PASSWORD=p USER_PASSWORD=p make dev-setup
```
guicassolato: Do we need to pass `MASTER_PASSWORD` and `USER_PASSWORD` again if the `run` target will set them to these same default values?

docker-compose.yml
```diff
@@ -1,44 +1,99 @@
...
+    image: quay.io/3scale/apisonator:latest
+    container_name: apisonator_compose
+    ports:
+      - "3001:3001"
+    command: "3scale_backend start -p 3001"
     depends_on:
       - redis
+    environment:
+      - CONFIG_REDIS_PROXY=redis:6379/2
+      - CONFIG_QUEUES_MASTER_NAME=redis:6379/2
```
guicassolato: This will make apisonator to share the same db as porta message bus.

In `system_environment` you can also specify `MESSAGE_BUS_REDIS_URL`. If missing, porta will assume `REDIS_URL`'s db +1 for the message bus (used by proxy deployments and communication with zync).

https://github.com/3scale/porta/blob/5efe0aed74c429f17575958f86d842b09e8a0a11/app/lib/three_scale/message_bus_config.rb#L44

To fix this you have 3 options:
1. Set a different redis db for apisonator, other than 1 or 2 (both taken by porta already).
2. Set a specific redis db to porta message bus other than 1 or 2 (respectively taken by porta for sidekiq and by apisonator). Do it for example with `MESSAGE_BUS_REDIS_URL=redis://redis:6379/8` in `system_environment`.
3. Set in `system_environment` both `REDIS_NAMESPACE=sidekiq` and `MESSAGE_BUS_REDIS_NAMESPACE=mbus` so sidekiq and message bus can both share the same db without risk of key collision. (The values can actually be whatever you want. These are just examples.)

INSTALL.md
duduribeiro: it is not needed. Just added here in the docs as an example that it can be overwritten.

Dockerfile
```diff
@@ -1,62 +1,41 @@
...
-VOLUME [ "/opt/system/tmp/cache/", \
-         "/opt/system/vendor/bundle", \
-         "/opt/system/node_modules", \
-         "/opt/system/public/assets", \
-         "/opt/system/public/packs-test", \
-         "/home/ruby/.luarocks" ]
-
 ADD . ./
 ADD config/examples/*.yml config/
+ADD config/docker/webpacker.yml config/
```
duduribeiro: @guicassolato I needed to add this custom config file because I couldn't add a ruby code (like <%= ENV[xxx] %>) in `config/webpacker.yml` because webpacker doesn't eval it when loading: https://github.com/rails/webpacker/issues/1615 and looks like it will not support: https://github.com/rails/webpacker/pull/1801

duduribeiro:
@guicassolato can you re-review? 

thanks 🍻

Dockerfile
guicassolato: This will affect builds for OpenShift as well:
https://github.com/3scale/porta/blob/e00fefee7dd2def43140dc325a784ba59aa66654/openshift/system/Dockerfile#L55
https://github.com/3scale/porta/blob/e00fefee7dd2def43140dc325a784ba59aa66654/openshift/system/Dockerfile.on_prem#L54

INSTALL.md
```diff
@@ -36,6 +36,22 @@ make bash
 
 All the source and dependencies for this project will be in place, making it possible to run porta and the tests from inside the container. See [Run Porta](#run-porta)
 
+#### Running the application
```
josemigallas: I think this should be a title-5:
```suggestion
##### Running the application
```

INSTALL.md
```diff
@@ -36,6 +36,22 @@ make bash
...
+In the first time, you need to setup the database. To configure it, run: (the Master and User password will be `p`)
+
+```
+MASTER_PASSWORD=p USER_PASSWORD=p make dev-setup
+```
+
+and to start the application and all dependencies:
+
+```
+make up
```
josemigallas: Is it `run` here instead?
```
$ make up
make: *** No rule to make target `up'.  Stop.
```

INSTALL.md
josemigallas: About the name of the task, `dev-setup` implies it is the setup of a development env. But the point of our Dockerfile is to have a production app running, no?

INSTALL.md
```diff
@@ -36,6 +36,22 @@ make bash
...
+```
+
```
josemigallas: Once it's running, how do you open the app?

josemigallas:
When exiting the app container, there are 5 containers left running. It would be nice if they were all shut down at the same time.

Makefile
```diff
@@ -124,16 +117,28 @@ clean-tmp: ## Removes temporary files
...
+run: MASTER_PASSWORD ?= "p"
+run: USER_PASSWORD ?= "p"
+run:
 	@echo "======= Run ======="
 	@echo
-	$(DOCKER_COMPOSE) run --rm --name $(PROJECT)-build-run $(DOCKER_ENV) build bash -c "$(CMD)"
+	@docker-compose run -e MASTER_PASSWORD=$(MASTER_PASSWORD) -e USER_PASSWORD=$(USER_PASSWORD) --rm system $(CMD)
+
+dev-setup: ## Makes the initial setup for the application ##
+dev-setup: CMD=rake db:setup
```
josemigallas: How can I undo this?

INSTALL.md
duduribeiro: yeah!.. you are right. I renamed the `up` to `dev-run`.

Will change it. thanks!

INSTALL.md
duduribeiro: @josemigallas the Dockerfile itself yes. But this commit also includes the `docker-compose` file that is used for development. So the `dev-*` tasks are the one used for running locally

INSTALL.md
duduribeiro: @josemigallas fixed it in the doc. Also, renamed it to `dev-start` like we discussed.

INSTALL.md
duduribeiro: Like locally. For example, I have a dnsmasq rule with example.com and I access this way:

http://master-account.example.com.lvh.me:3000

Gemfile.base
hallelujah: Uhhhh that is weird
Probably worth a PR in webpacker
And if not then we need to fix our code

INSTALL.md
hallelujah: Lvh.me or nip.io should be documented for the sake of new users in the community
I think it is already

hallelujah: About this instruction, can’t all the targets be chained in one and only target? `make all` which can be the default one

Makefile
hallelujah: @josemigallas what do you mean by “undo” and what are your expectations?

PR #911, opened 2019-06-17
[wip] Fixes liquid tag do_not_send within control flow blocks

guicassolato:

It injects the fallback for unknown liquid tags of the custom-defined `Liquid::Tags::Email` into its superclass `Liquid::Block`.

Even though those email tags (`header`, `bcc`, `do_not_send`, etc) are meant to be used only within the scope of an `email...endemail` block, missing them in the uper level implicates control flow blocks not allowed to be used within the email block. Furthermore, having the email tags known in `Liquid::Block` is currently a no-op if used other than within an `email...endemail` block.

### Examples

Without this PR, the following liquid template fails:

```
{% email %}
  {% if true %}
    {% do_not_send %}
  {% endif %}
{% endemail %}
```

while the following succeeds:

```
{% if true %}
  {% email %}
    {% do_not_send %}
  {% endemail %}
{% endif %}
```

With this PR, all the above succeeds and the following ones are no
no-op:

```
{% if true %}
  {% do_not_send %}
{% endif %}
```

```
{% do_not_send %}
```

Closes [THREESCALE-2226](https://issues.jboss.org/browse/THREESCALE-2226)

**Verification steps** 

1. Choose an event type that would result in an email being sent to a buyer (e.g. charging an invoice).
2. Trigger the event and confirms that an email was sent to the buyer.
3. Edit the corresponding email template of the event chosen adding the following to the beginning of the template:
```
{% email %}
  {% if true %}
    {% do_not_send %}
  {% endif %}
{% endemail %}
```
4. Trigger again the corresponding event that previously resulted in the email being sent.
5. Confirm that this time the email was intercepted and not sent.

### DIFF
--- a/features/old/emails.feature
+++ b/features/old/emails.feature
@@ -45,3 +45,26 @@ Scenario: Disable 'Waiting list confirmation' notification
 
     When I act as "Kirill"
     Then I should receive no email with subject "Waiting list confirmation"
+
+  Scenario: Conditionally disable 'Waiting list confirmation' notification
+      Given a provider "foo.example.com" with default plans
+      And provider "foo.example.com" requires accounts to be approved
+      And provider "foo.example.com" has multiple applications enabled
+      And provider "foo.example.com" has email template "account_confirmed"
+      """
+      {% email %}{% if true %}{% do_not_send %}{% endif %}{% endemail %}
+      Dear More Things,
+
+      This email is to let you know that you own even more things.
+      """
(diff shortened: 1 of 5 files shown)


duduribeiro:
💯 It looks good for me

guicassolato:
This is actually not needed at all. Closing in favor of https://github.com/3scale/porta/pull/928.

PR #918, opened 2019-06-18
[npm] lock updated

didierofrivia:


thomasmaas:
Is this this still needed @didierofrivia ?

josemigallas:
I recommend you try again @didierofrivia but removing node_modules folder and cleaning cache `npm cache clean --force`. It happens that package-lock gets corrupted when simply running `npm i`.

PR #922, opened 2019-06-18
[Cukes] 2 minor improvements from 2.6: A scenario doesn't need JS + fix warning

Martouta:
Originally this PR was opened to investigate and test for the `Net::ReadTimeout` randomly appearing in CircleCI for the provider login, but since that has been fixed in https://github.com/3scale/porta/pull/930, this PR now only contains 2 minor improvements that I encountered when I was dealing with that error:
1. One of the scenarios of the feature _service contracts_ doesn't need JS 😄 
2. Fix deprecation warning: Ast::Table -> MultilineArgument::DataTable

### DIFF
--- a/features/old/accounts/service_contracts.feature
+++ b/features/old/accounts/service_contracts.feature
@@ -21,7 +21,6 @@ Feature: Account service plans management
     Given a buyer "bob" signed up to provider "foo.example.com"
     Given current domain is the admin domain of provider "foo.example.com"
 
-  @javascript
   Scenario: Link to service contracts on account page in enterprise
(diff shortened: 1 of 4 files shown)


features/old/accounts/service_contracts.feature
```diff
@@ -21,7 +21,6 @@ Feature: Account service plans management
     Given a buyer "bob" signed up to provider "foo.example.com"
     Given current domain is the admin domain of provider "foo.example.com"
 
-  @javascript
```
duduribeiro: 💯

PR #923, opened 2019-06-18
Fix ContextSelector position / alignment

josemigallas:
Moves `ContextSelector`'s dropdown menu down and also aligns icons vertically.

**Provider**
context

**Master**
Screen Shot 2019-06-18 at 17 56 43

**Help menu**
Screen Shot 2019-06-18 at 18 09 06

### DIFF
--- a/app/assets/stylesheets/provider/_pop_navigation.scss
+++ b/app/assets/stylesheets/provider/_pop_navigation.scss
@@ -1,3 +1,5 @@
+$link-padding: line-height-times(1 / 2);
+
 .PopNavigation {
   color: $content-background;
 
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/provider/_pop_navigation.scss
```diff
@@ -35,7 +37,8 @@
     border: $border-width solid $border-color;
     border-radius: $border-radius;
     display: none;
-    margin: line-height-times(1 / 2, true) 0 0 0;
+    margin-top: line-height-times(3 / 2, true);
+    margin-left: calc(-#{$link-padding} + -#{$border-width});
```
thomasmaas:
```suggestion
    margin-left: -calc(#{$link-padding} + #{$border-width});
```

thomasmaas: @jmprusi doesn’t that work too and make it a bit more readable?

josemigallas: It won't work with a preceding -, can't find it in the docs why but the negative has to come from within.

PR #926, opened 2019-06-19
Fix flash message spacing

josemigallas:
**Note**
Due to scrollbar not being overlayed, the flash message is not aligned with the header icons anymore.

Screen Shot 2019-06-19 at 12 50 47

### DIFF
--- a/app/assets/stylesheets/provider/_flash.scss
+++ b/app/assets/stylesheets/provider/_flash.scss
@@ -1,7 +1,7 @@
 #flashWrapper {
   position: absolute;
-  right: 0;
-  top: line-height-times(0.5);
+  right: line-height-times(1);
+  top: line-height-times(1);
   text-align: right;
   @include animation(fadeInLeft 0.5s ease-in-out);
 }


app/assets/stylesheets/provider/_flash.scss
```diff
@@ -1,7 +1,7 @@
 #flashWrapper {
   position: absolute;
-  right: 0;
-  top: line-height-times(0.5);
+  right: line-height-times(1);
+  top: line-height-times(1);
```
thomasmaas: this might be a problem on other pages than the dashboard where it might overlap header i think

josemigallas: Since it's positioned relative to "wrapper", I don't think that won't be an issue. This is an example in Analytics:
Screen Shot 2019-06-20 at 15 13 52

thomasmaas: Screenshot 2019-06-20 15 27 41

I think it wouldn't fit here for example

josemigallas: So the problem is not overlapping the header but the horizontal nav? It can certainly be half-line, no problem about that. I thought you meant something else.

This is how it looks with 1 line top and right, anyways:
Screen Shot 2019-06-20 at 15 35 11

PR #929, opened 2019-06-20
Jest circleci

didierofrivia:

After upgrading webpack and babel, jest tests are  costly.

fixes https://issues.jboss.org/browse/THREESCALE-2833

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -565,6 +565,7 @@ jobs:
       - *upload-coverage
 
   jest:
+    resource_class: medium+
     parameters:
       executor:
         type: string
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -578,7 +578,7 @@ jobs:
       - run:
           name: Jest specs
           command: |
-            npm run-script jest --maxWorkers=2
+            npm run-script jest --maxWorkers=3
       - *upload-coverage
 
   unit:


.circleci/config.yml
```diff
@@ -565,6 +565,7 @@ jobs:
       - *upload-coverage
 
   jest:
+    resource_class: medium+
```
josemigallas: Do we have budget approval for this?

didierofrivia: @thomasmaas @gsaslis what do you reckon?

thomasmaas: Let’s just do it, we can always undo it

josemigallas: Apparently a support ticket has to opened, in order for this to work.

josemigallas: It LGTM to merge but let's be sure it won't be for nothing.

gsaslis: yup just go ahead. we already have the option to use different resource classes enabled. (and we do use it already in porta ci). 

btw, this won't necessarily cost _that much more_ because if jest takes advantage of the extra cores and the test finish "faster" (and billing is per minute), it won't be "double" the cost. 

tl;dr go go go go go go go go go

Martouta:
Please @didierofrivia rebase from master 😄

josemigallas:
I rebased onto master, could you guys approve again? @gsaslis @hallelujah

PR #931, opened 2019-06-20
Rake task to issue domain changed events to all providers

guicassolato:

It implements a set of rake tasks to issue domain changed events of
providers and proxies. This will make Zync to resync Openshift routes
eventually managed.

Related to [THREESCALE-2841](https://issues.jboss.org/browse/THREESCALE-2841)

**Verification steps** 

Provider domains:
```
bundle exec rake zync:resync:provider_domains
```

Proxy domains:
```
bundle exec rake zync:resync:proxy_domains
```

Both provider and proxy domains:
```
bundle exec rake zync:resync:domains
```

### DIFF
--- a/lib/tasks/zync.rake
+++ b/lib/tasks/zync.rake
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+namespace :zync do
+  namespace :resync do
+    def each_with_progress(scope)
+      total_count = scope.count
+      batch_size = 100
+      index = 0
+
+      progress = -> do
+        break unless (index % batch_size) == 0
+        percent = (index / total_count.to_f) * 100.0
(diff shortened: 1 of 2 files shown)


lib/tasks/zync.rake
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+namespace :zync do
+  namespace :resync do
+    def each_with_progress(task_desc, scope)
```
hallelujah: `task_desc` is not used anywhere?

guicassolato: Right, forgot that. I has a header message printed to STDOUT before.

guicassolato: Fixed

PR #933, opened 2019-06-20
Fix shim js error

damianpm:
Fix shim.js error

### DIFF
--- a/app/javascript/packs/LoginPage.js
+++ b/app/javascript/packs/LoginPage.js
@@ -1,3 +1,6 @@
+console.log('------------>>>>>>>>>')
+console.log(document.body)
+
 import '@babel/polyfill'
 import 'core-js/es7/object'
 
--- a/app/views/layouts/provider/login.html.slim
+++ b/app/views/layouts/provider/login.html.slim
@@ -17,4 +17,5 @@ html lang="en"
     # are split Provider/Buyer side
 
   body class="login-layout"
+    = javascript_pack_tag 'LoginPage'
     = yield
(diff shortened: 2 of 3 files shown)


app/views/layouts/provider/login.html.slim
```diff
@@ -8,6 +8,7 @@ html lang="en"
     = render 'provider/analytics'
     = stylesheet_link_tag 'loginPageFonts'
     = javascript_pack_tag 'pf4BaseStyles'
+    = javascript_pack_tag 'LoginPage'
```
josemigallas: I think this has to be loaded after the body is created, not in the head.

damianpm: But it will wait for `DOMContentLoaded`

hallelujah: Yes exactly otherwise it will raise the same issue:

image

hallelujah: And it is not the only page that should be touched, it is wherever shim.js is used in the react components.

Martouta:
I'd say point to master
because otherwise we will have to wait for the tests again in ER1 branch if we merge to that one
and no need to wait twice if we already wait for yours

app/javascript/packs/LoginPage.js
```diff
@@ -1,3 +1,6 @@
+console.log('------------>>>>>>>>>')
+console.log(document.body)
```
didierofrivia: not needed this `console.log` I'd say...

app/views/layouts/provider/login.html.slim
```diff
@@ -17,4 +17,5 @@ html lang="en"
     # are split Provider/Buyer side
 
   body class="login-layout"
+    = javascript_pack_tag 'LoginPage'
```
didierofrivia: This is ok since there's just one layout - page affected, either way you should need to yield something like `:javascript_login` and se `content_for` helper in the views

app/javascript/packs/LoginPage.js
didierofrivia: Well, just for debugging, you might want to ditch this commit afterwards

PR #941, opened 2019-06-21
[WIP] Remove old "deleted object" entries

Martouta:
Closes [THREESCALE-2085 - Use the Janitor to regularly clean DeletedObject a while after the owner is destroyed](https://issues.jboss.org/browse/THREESCALE-2085)

Destroyes entries from "DeletedObject" once their owner and the event of the owner has been destroyed. Implemented only for Service owners because it is the only ones that we create so far 😄


test/unit/deleted_object_test.rb
```diff
@@ -16,4 +16,74 @@ class DeletedObjectTest < ActiveSupport::TestCase
     assert_same_elements metrics.map(&:id),   DeletedObject.metrics.pluck(:object_id)
     assert_same_elements contracts.map(&:id), DeletedObject.contracts.pluck(:object_id)
   end
+
+  test '.service_owner' do
+    metric = FactoryBot.create(:metric)
```
Martouta: I will do some refactoring in the tests of this class bcz there is a lot of repetition right now 😄

app/lib/event_store/services/service_deleted_event.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class EventStore::Services::ServiceDeletedEvent < EventStore::Event
+  default_scope -> { where(event_type: ::Services::ServiceDeletedEvent) }
+
+  scope :by_service_id, lambda { |service_id|
+    concat_query = if System::Database.oracle?
+                     "('%service_id: ' || #{service_id} || '%')"
```
guicassolato: Why does it have to be concatenated in the database?

app/models/deleted_object.rb
```diff
@@ -8,5 +8,10 @@ class DeletedObject < ApplicationRecord
     scope scoped_class.to_s.underscore.pluralize.to_sym, -> { where(object_type: scoped_class) }
   end
 
-  # TODO: cleanup regularly once the owner is destroyed too :) or this will grow uncontrollably
+  scope :service_owner, -> { where.has { owner_type == Service } }
```
guicassolato: How about `:owned_by_service`?

Martouta: Makes sense 👍

app/models/deleted_object.rb
```diff
@@ -8,5 +8,10 @@ class DeletedObject < ApplicationRecord
...
+  scope :service_owner_not_exists, -> { service_owner.joining { owner.of(Service).outer }.where('services.id IS NULL') }
```
guicassolato: It makes me think whether we need it always chained to the previous scope. I was thinking about something like `scope :missing_owner` that later could be chained `owned_by_service.missing_owner`. WDYT?

app/models/deleted_object.rb
```diff
@@ -8,5 +8,10 @@ class DeletedObject < ApplicationRecord
...
+  scope :service_owner_event_not_exists, lambda {
```
guicassolato: Following https://github.com/3scale/porta/pull/941/files#r298992964, this would become `missing_owner_event`.

app/lib/event_store/services/service_deleted_event.rb
Martouta: because it is inside a subquery @guicassolato and it is either this or load them all in memory and select one by one with Ruby afterwards.

app/models/deleted_object.rb
Martouta: @guicassolato That is what I tried first but I simply didn't find any way to make it work 🤷‍♀

Martouta: @guicassolato told me:

> i managed to solve for the first scope…

```ruby
scope :missing_owner, -> do
    associations = distinct(:owner_type).pluck(:owner_type).map(&:constantize).map do |association|
      -> { joining { owner.of(association).outer }.where("#{association.table_name}.id IS NULL") }
    end
    associations.inject(all) { |chain, association| chain.merge!(association) }
  end
```

> but still struggling with the events
> the problem is it could be any type of event of any type of object
> and `event_store_events` only holds a link to the associated object via an attribute inside the json data… when it does at all

Martouta: I said this:

> for the events i would also retrieve the kind of event and call to a class with that event type with the same name of scope for each and then:
> - if that scope exists (for example, let's call it `by_id`), then each of them adapts it to its own event type
> - if it doesn't exist is because it doesn't hold its id at all or because we didn't mind implementing it because we didn't need it, and in that case let's just make it as if it was a `none`

app/lib/event_store/services/service_deleted_event.rb
guicassolato: I don't get this. The if condition implies that the scope will always reduce to either

```ruby
->(service_id) { where("data LIKE ('%service_id: ' || #{service_id} || '%')") }
```
or
```ruby
->(service_id) { where("data LIKE CONCAT('%service_id: ',  #{service_id}, '%')") }
```

Either way, the actual SQL will never be built unless you provide the scope with a value for `service_id`, meaning that the scope could be simply:

```ruby
->(service_id) { where("data LIKE ('%service_id: #{service_id},%')") }
```

Notice the comma (`,`) added right before the last `%` to prevent `service_id: 10` from being selected when in fact we wanted only `service_id: 1`.

app/models/deleted_object.rb
guicassolato: For the events, I like your plan in https://github.com/3scale/porta/pull/941#discussion_r299449778.

app/lib/event_store/services/service_deleted_event.rb
Martouta: Right!! Good one!! 💪 🥇

app/models/deleted_object.rb
```diff
@@ -8,5 +8,76 @@ class DeletedObject < ApplicationRecord
     scope scoped_class.to_s.underscore.pluralize.to_sym, -> { where(object_type: scoped_class) }
   end
 
-  # TODO: cleanup regularly once the owner is destroyed too :) or this will grow uncontrollably
+  scope :owned_by_service, -> { where.has { owner_type == Service } }
+
+  scope :missing_owner, lambda {
+    associations = distinct(:owner_type).pluck(:owner_type).map(&:constantize).map do |association|
```
guicassolato: I was giving a second thought on this. This is not thread-safe.

Say we have _N_ distinct types of owners in `deleted_objects`. When we invoke the scope, `distinct(:owner_type).pluck(:owner_type)` will set the  array of types based on which we'll loop and add _N_ `JOIN` clauses to the scope. In between these two steps (fetching the types and fetching the records) however, a _N+1_-th distinct type of owner can be inserted into `deleted_objects`. That specific ongoing call to scope will miss the records whose `owner_type` value is equal to new type just inserted.

Whilst `missing_owner` is only used for fetching stale records within the scope of a `PurgeStaleObjectsWorker` job (or other low risk operations), perhaps it's all right to miss a few records every once in a while. For more critical operations on the other hand, we may need to reconsider this approach.

app/lib/event_store/services/service_deleted_event.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class EventStore::Services::ServiceDeletedEvent < EventStore::Event
+  default_scope -> { where(event_type: ::Services::ServiceDeletedEvent) }
+
+  scope :by_object_id, ->(service_id) { where("data LIKE ('%service_id: #{service_id},%')") }
```
guicassolato: @Martouta just reminded me that it's stored as a YAML, not as a JSON. So we need to replace `,%` with a `\n%`.

Martouta:
Closed for its inefficiency. Only searching in one event from the table to check the service id is already a table scan, and like this is toooooo much work for the DB, so we have thought in doing it differently and I will explain it in the Jira issue 😉

PR #944, opened 2019-06-25
Zync UnprocessableEntityError hook in a Sidekiq batch

guicassolato:

TLDR – Reverts 3scale/porta#846 + a couple fixes, including one after https://github.com/3scale/porta/pull/927 was merged

This PR modifies the way the hook on the `UnprocessableEntityError` exceptions in the Zync worker behaves.

Right now in master, whenever a zync job fails with a `UnprocessableEntityError`, we hook into the error to publish zync events for the possible dependent objects that are assumed to be missing in Zync and therefore make it to return a 422. Right after publishing those dependency zync events, it raises the error in the current job and all zync jobs concur to be processed in parallel by Sidekiq – the one that failed and the ones just created as "dependencies".

This PR makes the current job to wait for the dependency jobs to be performed to only then re-enqueue itself. So instead of making the job to concur with its dependencies right away, it waits until there are no missing dependencies in Zync and therefore no reason for another 422.

This was implemented using Sidekiq batches. When `UnprocessableEntityError` happens, the job opens a batch, add the dependency jobs to the batch and programs itself to be re-enqueued in the batch callback (`on_complete`).

**Verification steps** 
1. Quiet Sidekiq all processes
2. Create a new API in the Admin Portal
3. In the Sidekiq Web Console, delete the zync job meant to create the `Service` in zync (and keep the one to create the `Proxy`)
4. Redeploy `system-sidekiq`

Expected behavior:
- The Zync job to create the `Proxy` will get performed and fail with `UnprocessableEntityError` (although it shows up as a success in the logs)
- Since it’s not the “last_attempt”, the job will invoke `ZyncWorker#sync_dependencies` which creates a new zync job for the `Service` (deleted in step 3 above) in a batch
- The batch only contains that job for the `Service` and in its callback (`on_complete`) the first job (the one for the `Proxy`) is re-enqueued
- After the batch is finished (and therefore the `Service` is created in zync), the callback is invoked, the job for the `Proxy` enqueued and processed

Example:
Screen Shot 2019-06-25 at 6 09 11 PM

Contains commits of https://github.com/3scale/porta/pull/947

### DIFF
--- a/app/events/zync_event.rb
+++ b/app/events/zync_event.rb
@@ -53,6 +53,10 @@ class ZyncEvent < BaseEventStoreEvent
     end
   end
 
+  def create_dependencies
+    dependencies.map { |dependency| self.class.create(self, dependency) }
+  end
+
   def self.type_for(model)
     case model
     when Cinstance, ApplicationRelatedEvent then 'Application'
--- a/test/workers/zync_worker_test.rb
+++ b/test/workers/zync_worker_test.rb
@@ -44,9 +44,137 @@ class ZyncWorkerTest < ActiveSupport::TestCase
     ZyncWorker::MessageBusPublisher.expects(:enabled).returns(false)
 
     assert_difference(zync_events.method(:count), +2) do
+      worker.perform(event.event_id, event.data.with_indifferent_access)
+    end
+  end
+
+  class UnprocessableEntityRetryTest < ActiveSupport::TestCase
+    disable_transactional_fixtures!
+
+    attr_reader :worker, :event, :application
+
+    setup do
+      EventStore::Repository.stubs(raise_errors: true)
+      ZyncWorker::MessageBusPublisher.any_instance.stubs(enabled?: false)
+
+      @worker = ZyncWorker.new
+      @application = FactoryBot.create(:simple_cinstance).reload # reload to get tenant_id
+      FactoryBot.create(:simple_admin, account: application.provider_account) # for the access token
+      @event = ZyncEvent.create(RailsEventStore::Event.new, application)
+
+      worker.config.stubs(endpoint: 'http://example.com') # so it makes http request
+      worker.publisher.call(event) # so it is later available to find
+
+      stub_request(:put, 'http://example.com/tenant').to_return(status: 200)
+      stub_request(:put, 'http://example.com/notification').to_return(status: 422).with(body: event.data.to_json) # causes UnprocessableEntityError to be raised
+    end
+
+    test 'sync event dependencies when it fails' do
+      worker.expects(:sync_dependencies).with(event.event_id).returns(true)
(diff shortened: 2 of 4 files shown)


app/events/zync_event.rb
```diff
@@ -53,6 +53,10 @@ def dependencies
     end
   end
 
+  def create_dependencies
+    dependencies.map { |dependency| ZyncEvent.create(self, dependency) }
```
Martouta:
```suggestion
    dependencies.map { |dependency| self.class.create(self, dependency) }
```
Not very important, but just a small suggestion 😄

test/workers/zync_worker_test.rb
```diff
@@ -44,10 +44,99 @@ class ZyncWorkerTest < ActiveSupport::TestCase
...
+
+    test 'raises the error when there is no dependency' do
+      worker.expects(:sync_dependencies).with(event.event_id).returns([])
       assert_raises ZyncWorker::UnprocessableEntityError do
         worker.perform(event.event_id, event.data.with_indifferent_access)
       end
     end
+
+    test '#create_dependency_events' do
+      EventStore::Repository.expects(:find_event!).returns(event)
```
Martouta: I think that we could be more specific to avoid false positives:
```ruby
EventStore::Repository.expects(:find_event!).with(event.event_id).returns(event)
```

test/workers/zync_worker_test.rb
```diff
@@ -44,10 +44,99 @@ class ZyncWorkerTest < ActiveSupport::TestCase
...
+      dependency_events = event.create_dependencies
+
+      worker.expects(:create_dependency_events).with(event_id).returns(dependency_events)
+      worker.expects(:publish_dependency_events).with(dependency_events)
+
+      worker.sync_dependencies(event_id)
+    end
+
+    test '#publish_dependency_events enqueues the jobs' do
+      event_id = event.event_id
```
Martouta: This assignment is not used 😄

test/workers/zync_worker_test.rb
```diff
@@ -44,10 +44,140 @@ class ZyncWorkerTest < ActiveSupport::TestCase
...
+      klass.expects(:perform_async).with(event_id, anything, 1)
+      klass.new.on_complete(1, {'event_id' => event_id, 'manual_retry_count' => 0})
+    end
+
+    test 'number of attempts' do
+      event_id = event.event_id
+      retry_limit = worker.retry_limit
+
+      dependency_events = []
+      retry_limit.times { |i| dependency_events << event.create_dependencies }
```
Martouta:
```suggestion
      retry_limit.times { dependency_events << event.create_dependencies }
```
I know that this happened for a refactoring because I am looking commit by commit 😄

PR #859, opened 2019-06-06
Implements Vertical Nav in Patternfly 4

josemigallas:
**What this PR does / why we need it**:

Uses Patternfly 4 React to redesign the vertical navigation.

Changes all slim "sections" logic into a ruby helper.
Implements a Vertical Navigation component in React using Patternfly 4 styles.
Instead of using PF4-React it uses a "proxy" class that uses its same structure so that it could be easily swapped.

Look & Feel (ContextSelector icon takes a bit to load)
vertnav01

**Which issue(s) this PR fixes** 

[THREESCALE-2603: Add PF4 vertical navbar](https://issues.jboss.org/browse/THREESCALE-2603)

**Verification steps** 

Navigate to all possible sections and verify everything works as expected and items highlights

**NOTES**
Patternfly fonts are not being used so probably there are more things that could be removed from assets/patternfly. This is to be investigated.

PR #945, opened 2019-06-25
Feature/threescale 2883 reduce assets

josemigallas:
Removes some duplicated styles, clean up commented ones delete unused assets introduced in #859 

[THREESCALE-2883: Remove duplication of assets](https://issues.jboss.org/browse/THREESCALE-2883)

**Verification steps** 
It all have to look like in preview in Chrome, Firefox and IE11.

**NOTE FOR REVIEWERS**
Review changes in files located in `app/assets/stylesheets/provider/*`. The rest are just deletions.


app/assets/stylesheets/provider/_header.scss
```diff
@@ -1,11 +1,10 @@
 $header-height: line-height-times(63 / 20);
 $master-ribbon-height: line-height-times(1);
 
-.Header {
-
+header {
```
damianpm: I'd keep the `.Header` class and target it, because in the future we or any of our libraries may introduce more `<header>` tags. In fact, PF4/React introduces several of them, we already have some in the Login Page

josemigallas: OK! Didn't know that

PR #946, opened 2019-06-26
Adding @babel/polyfill to support Promise and other features in IE11

damianpm:
Adding @babel/polyfill to support Promise and other features in IE11

https://issues.jboss.org/browse/THREESCALE-2889


hallelujah:
Hmm probably checking where Promise is used and add it in the file is better though

damianpm:
> Hmm probably checking where Promise is used and add it in the file is better though

Now it's used in all stats pages. But we'll need this for any other PF4/React component. Then it has sense to add it at top level @hallelujah

PR #948, opened 2019-06-26
Classify ZyncWorker::UnprocessableEntityError as Bugsnag INFO level

guicassolato:

### DIFF
--- a/app/workers/zync_worker.rb
+++ b/app/workers/zync_worker.rb
@@ -54,6 +54,8 @@ class ZyncWorker
 
   class UnprocessableEntityError < InvalidResponseError; end
 
+  Bugsnag::Middleware::ClassifyError::INFO_CLASSES << UnprocessableEntityError.to_s
+
   module MessageBusInstrumentation
     def publish(channel, data, options = {})
       id = super


app/workers/zync_worker.rb
```diff
@@ -54,6 +54,8 @@ def find_model(name)
 
   class UnprocessableEntityError < InvalidResponseError; end
 
+  Bugsnag::Middleware::ClassifyError::INFO_CLASSES << UnprocessableEntityError.to_s
```
hallelujah: Hmm can't we do that in a config file?

guicassolato: Sorry, I was just following what we have in the file: https://github.com/3scale/porta/blob/795ec0d97ee5893da92e1744440227bb8fb953cf/app/workers/zync_worker.rb#L192

hallelujah: See https://docs.bugsnag.com/platforms/ruby/sidekiq/customizing-error-reports/#the-report-object

guicassolato:
We are merging this now, but better think about a more sophisticated way to configure report severities for classes in general.

PR #953, opened 2019-07-01
[Refactor] Remove unused Account::Gateway#payment_gateway_unconfigured?

Martouta:
Removes a method that was only used from tests and now it is not used from anywhere else:
image

I encountered this while working on another task 😄

### DIFF
--- a/test/unit/invoice_test.rb
+++ b/test/unit/invoice_test.rb
@@ -476,7 +476,8 @@ class InvoiceTest < ActiveSupport::TestCase
 
   test '#charge! failed if provider payment_gateway is unconfigured' do
     @buyer.expects(:charge!).never
-    @provider.stubs(:payment_gateway_unconfigured?).returns(true)
+    @provider.stubs(:payment_gateway_configured?).returns(false)
+    @billing.create_line_item!(name: 'Fake', cost: 1.233, description: 'really', quantity: 1)
     @invoice.update_attribute(:state, 'pending')
 
     refute @invoice.charge!, 'Invoice should not charge!'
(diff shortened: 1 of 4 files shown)


test/unit/invoice_test.rb
```diff
@@ -476,7 +476,8 @@ def setup_1964_and_2009_invoices
 
   test '#charge! failed if provider payment_gateway is unconfigured' do
     @buyer.expects(:charge!).never
-    @provider.stubs(:payment_gateway_unconfigured?).returns(true)
+    @provider.stubs(:payment_gateway_configured?).returns(false)
+    @billing.create_line_item!(name: 'Fake', cost: 1.233, description: 'really', quantity: 1)
```
Martouta: The 2 tests are now like this:

```ruby
  test '#charge! is successful' do
    @buyer.expects(:charge!).returns(true)
    @provider.stubs(:payment_gateway_configured?).returns(true)
    @billing.create_line_item!(name: 'Fake', cost: 1.233, description: 'really', quantity: 1)
    @invoice.update_attribute(:state, 'pending')

    assert @invoice.charge!, 'Invoice should charge!'
  end

  test '#charge! failed if provider payment_gateway is unconfigured' do
    @buyer.expects(:charge!).never
    @provider.stubs(:payment_gateway_configured?).returns(false)
    @billing.create_line_item!(name: 'Fake', cost: 1.233, description: 'really', quantity: 1)
    @invoice.update_attribute(:state, 'pending')

    refute @invoice.charge!, 'Invoice should not charge!'
  end
```

So now both tests contain the same and only the stub of `payment_gateway_configured?` changes. This is important because by default, without stubbing, `payment_gateway_configured?` is `true` but no matter if we stub it or not, without the `@billing.create_line_item!(name: 'Fake', cost: 1.233, description: 'really', quantity: 1)` is does not charge anyway 😂 And we want it not to charge for the config, not for the lack of items.

PR #954, opened 2019-07-02
Adding PF4/React components for Vertical Navbar

damianpm:

Add PF4/React components to vertical navigation

https://issues.jboss.org/browse/THREESCALE-2808

**Verification steps** 

Go to any page with vertical navigation
Everything should work the same way


app/javascript/src/Navigation/components/VerticalNav.jsx
```diff
@@ -1,14 +1,19 @@
...
+  NavItem,
+  NavList
+} from '@patternfly/react-core'
 import { createReactWrapper } from 'utilities/createReactWrapper'
 
 type Item = {
   id: string,
   title: string,
-  path: ?string,
-  target: ?string
```
didierofrivia: What was the `target` used for before?

spec/javascripts/Navigation/VerticalNav.spec.jsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react'
```
didierofrivia: Write some test about user interaction about main section and sub menu

app/javascript/src/Navigation/components/VerticalNav.jsx
damianpm: It was passed to `NavItem` in `PF4NavProxy.jsx`. Several props were passed because the components were crafted by us. But as we are using now the PF4 components, we don't need it anymore

josemigallas: This target is used for some links that are opened in a new tab. Therefore by removing it, it won't work. In other words, it can't be removed.

spec/javascripts/Navigation/VerticalNav.spec.jsx
```diff
@@ -0,0 +1,74 @@
...
+  expect(subLink.exists()).toBe(false)
+})
+
+it('should open/collapse a submenu', () => {
+  const propsSubmenu = {...props,
+    activeSection: 'Foo',
+    activeItem: null
+  }
+  const wrapper = mount(<VerticalNav {...propsSubmenu} />)
+  const link = wrapper.find('.pf-c-nav__item').at(2).find('a').at(0)
```
josemigallas: Make more natural assertions:
```suggestion
  const link = wrapper.find('NavExpandable').find('a').first()
```

spec/javascripts/Navigation/VerticalNav.spec.jsx
```diff
@@ -0,0 +1,74 @@
...
+  expect(wrapper.find('li.pf-c-nav__item').at(2).hasClass('pf-m-expanded')).toEqual(false)
```
josemigallas:
```suggestion
  expect(wrapper.find('NavExpandable').hasClass('pf-m-expanded')).toEqual(false)
```

PR #957, opened 2019-07-03
Adding basic PF4/React Signup Form

damianpm:

Adds PF4/React to Signup Page

https://issues.jboss.org/browse/THREESCALE-2938

**Verification steps** 

- Go to **Account Settings > Users > Invitations**
- Send an invitation
- Logout
- In the terminal, look for the link sent
- Go to the link

**Screenshots:**

signuo01

signup02


app/javascript/packs/SignUpPage.js
```diff
@@ -0,0 +1,11 @@
+import {SignupPageWrapper} from 'LoginPage'
+import {safeFromJsonString} from 'utilities/json-utils'
+
+document.addEventListener('DOMContentLoaded', () => {
+  const loginLayout = document.querySelector('.login-layout')
+  loginLayout.removeChild(document.getElementById('old-signup-page-wrapper'))
```
didierofrivia: In this specific case, the signup form, you might not need to do the progressive enhancement way we discussed in the past since it's an isolated case.

app/javascript/src/LoginPage/loginForms/SignupForm.jsx
```diff
@@ -0,0 +1,170 @@
...
+      isValidPassword: undefined,
+      isValidPasswordConfirmation: undefined
+    }
+  }
+
+  setIsValidUsername = () => this.setState({ isValidUsername: this.state.username !== '' })
+
+  handleTextInputUsername = (username: string) => this.setState({ username }, this.setIsValidUsername)
+
+  setIsValidEmail = () => this.setState({ isValidEmailAddress: this.state.emailAddress !== '' })
```
didierofrivia: I don't think this kind of validation is really useful this way.... Isn't PF4 providing a way to validate email? This applies to the rest of the validation in this form. You might want to use HTML5 validation instead.

app/javascript/src/LoginPage/loginForms/SignupForm.jsx
```diff
@@ -0,0 +1,170 @@
...
+  isValidPasswordConfirmation: ?boolean
+}
+type HandlersProps = {
+  handleTextInputUsername: (username: string) => void,
+  handleTextInputEmail: (emailAddress: string) => void,
+  handleTextInputPassword: (password: string) => void,
+  handleTextInputPasswordConfirmation: (passwordConfirmation: string) => void
+}
+
+const FormGroups = ({state, handlers}: {state: State, handlers: HandlersProps}) => {
```
didierofrivia: I'd keep each group separated, and not sharing the state, but feeding it props only.

Martouta:
These 2 CodeClimate issues seem easy to fix 🤔 
image

app/javascript/packs/SignUpPage.js
damianpm: It's needed for the cucumbers, or we'd need to add a lot of @javascript tags

app/javascript/src/LoginPage/loginForms/SignupForm.jsx
damianpm: No, PF4 is not providing any kind of validation.
We can add HTML validation easily

app/javascript/packs/SignUpPage.js
didierofrivia: That's not what I meant. Adding `@javascript` tags are a drawback when the feature not necessarily needs it. However, In this case, the register step might not be needed in a lot of features, do you know how many are?. You might want to create a new cuke for testing the new register (with `@javascript` tag) OR (in the case there are not many features using register) remove the old form and add the `@javascript` tag. We shouldn't aim to trick the cukes, it's more about understand the feature we are refactoring/creating and how it's been tested... then we could either adapt, indulge or write new tests.

app/javascript/src/LoginPage/loginForms/SignupForm.jsx
didierofrivia: Then all of those functions can be reduce to one: `required`

app/javascript/src/LoginPage/loginForms/SignupForm.jsx
```diff
@@ -0,0 +1,127 @@
...
+      isFilledEmailAddress: undefined,
+      isFilledFirstname: true,
+      isFilledLastname: true,
+      isFilledPassword: undefined,
+      isFilledPasswordConfirmation: undefined
+    }
+  }
+
+  handleTextInputUsername = (username: string) =>
+    this.setState({ username, isFilledUsername: username !== '' })
```
damianpm: @didierofrivia I think we can only check that input fields are not empty, to show the PF4 text hint, and let the browser do the rest of the validation.
I will create a task to add our own client side validation

app/javascript/packs/SignUpPage.js
```diff
@@ -0,0 +1,23 @@
+import '@babel/polyfill'
+import 'core-js/es7/object'
+
+import {SignupPageWrapper} from 'LoginPage'
+import {safeFromJsonString} from 'utilities/json-utils'
+
+const isIE11 = !!navigator.userAgent.match(/Trident\/7\./)
```
hallelujah: This could be a `navigatorUtil.isIE11` kinda function no?

app/javascript/src/LoginPage/loginForms/SignupForm.jsx
```diff
@@ -0,0 +1,127 @@
...
+}
+
+type State = {
+  username: string,
+  emailAddress: string,
+  firstname: string,
+  lastname: string,
+  password: string,
+  passwordConfirmation: string,
+  isFilledUsername: ?boolean,
```
didierofrivia: I honestly think this is not needed at all, you could validate each field when it's updated (and/or submit) that it's not empty and do something about it (i.e: show error message). This is something that just pollutes the state.

app/javascript/src/LoginPage/loginForms/SignupForm.jsx
didierofrivia: You could just use one fn to validate any field instead of having several and storing the state, maybe on submit

damianpm: The idea is to update the state using ´onChange´, so we can check each field as you type. This is the way it works on the Login Page.
We could also do the validation only on submit. We just need to agree on how to do it

PR #963, opened 2019-07-05
Allow inclusion of special chars in App ID/KEY as defined in RFC 6749

Martouta:
Closes [THREESCALE-1451](https://issues.jboss.org/browse/THREESCALE-1451)

### DIFF
--- a/app/models/application_key.rb
+++ b/app/models/application_key.rb
@@ -2,8 +2,6 @@ class ApplicationKey < ApplicationRecord
 
   KEYS_LIMIT = 5
 
-  APP_KEY_FORMAT = /[a-z\d][a-z\d\-\_]*[a-z\d]/i.freeze
-
 
   belongs_to :application, :class_name => 'Cinstance', :inverse_of => :application_keys
 
@@ -11,7 +9,7 @@ class ApplicationKey < ApplicationRecord
 
   validates :application, presence: true
   # letters, numbers, dash, cannot stat with dash, case insensitive
-  validates :value, format: { with: /\A#{APP_KEY_FORMAT}\Z/ },
+  validates :value, format: { with: /\A[\x20-\x7E]+\Z/ },
                     length: { within: 5..255 },
                     uniqueness: { scope: :application_id }
 
--- a/app/models/cinstance.rb
+++ b/app/models/cinstance.rb
@@ -110,7 +110,6 @@ class Cinstance < Contract
 
   validate :same_service, on: :update, if: :plan_id_changed?
 
-  APP_ID_FORMAT = /[\w-]+/.freeze
   # letter, number, underscore (_), hyphen-minus (-), dot (.), base64 format
   # In base64 encoding, the character set is [A-Z,a-z,0-9,and + /], if rest length is less than 4, fill of '=' character.
   # ^([A-Za-z0-9+/]{4})* means the String start with 0 time or more base64 group.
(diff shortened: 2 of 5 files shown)


app/models/cinstance.rb
```diff
@@ -110,16 +110,15 @@ def validate_plan_is_unique?
 
   validate :same_service, on: :update, if: :plan_id_changed?
 
-  APP_ID_FORMAT = /[\w-]+/.freeze
+  APP_ID_FORMAT = /[\x20-\x7E]+/
```
hallelujah: Not freezing it?

app/models/application_key.rb
```diff
@@ -2,7 +2,7 @@ class ApplicationKey < ApplicationRecord
 
   KEYS_LIMIT = 5
 
-  APP_KEY_FORMAT = /[a-z\d][a-z\d\-\_]*[a-z\d]/i.freeze
+  APP_KEY_FORMAT = /[\x20-\x7E]+/
```
hallelujah: Not freezing it?

app/models/cinstance.rb
```diff
@@ -110,16 +110,15 @@ def validate_plan_is_unique?
...
   # letter, number, underscore (_), hyphen-minus (-), dot (.), base64 format
   # In base64 encoding, the character set is [A-Z,a-z,0-9,and + /], if rest length is less than 4, fill of '=' character.
   # ^([A-Za-z0-9+/]{4})* means the String start with 0 time or more base64 group.
   # ([A-Za-z0-9+/]{4}|[A-Za-z0-9+/]{3}=|[A-Za-z0-9+/]{2}==) means the String must end of 3 forms in [A-Za-z0-9+/]{4} or [A-Za-z0-9+/]{3}= or [A-Za-z0-9+/]{2}==
   # matches also the non 64B case with (\A[\w\-\.]+\Z)
   USER_KEY_FORMAT = /(([\w\-\.]+)|([A-Za-z0-9+\/]{4})*([A-Za-z0-9+\/]{4}|[A-Za-z0-9+\/]{3}=|[A-Za-z0-9+\/]{2}==))/
 
-  # Arbitrary limit, increase don't broke anything.
-  validates :application_id, format: { with: /\A#{APP_ID_FORMAT}\Z/ }, length: { in: 4..140 }
+  validates :application_id, format: { with: /\A#{APP_ID_FORMAT}\Z/ }, length: { in: 4..255 }
```
hallelujah: Why do we have the regexp interpolated? Can't we just use 

```ruby
APP_ID_FORMAT = /\A[\x20-\x7E]\Z/
```

app/models/cinstance.rb
Martouta: On purpose 😄 
image

app/models/application_key.rb
Martouta: Same as https://github.com/3scale/porta/pull/963#discussion_r300677463

app/models/cinstance.rb
Martouta: sure 👍 good for me 👍 😉

app/models/cinstance.rb
Martouta: wow, interesting 🤔 
image

test/unit/cinstance_test.rb
```diff
@@ -1038,4 +1038,28 @@ def create_pending_cinstance
     assert_not_includes ids_cinstances_like_list, id_cinstance_with_another_plan
     assert_not_includes ids_cinstances_like_list, id_service_contract_with_enterprise_plan
   end
+
+  test 'App ID can include special characters as defined in the RFC 6749' do
+    generate_key_rfc_6749 = -> { ("\x20".."\x7e").to_a.shuffle.slice(0, 254).join }
```
hallelujah:
```suggestion
    generate_key_rfc_6749 = -> { 255.times.map { ("\x20".."\x7e").sample }.join }
```

Martouta: @hallelujah It isn't really the same 🤔 In the test I've done, it tested absolutely all the characters between `("\x20".."\x7e")` (because there are less than 100), and the only random thing is the order, but all of them appear 😄 And this test is not to test the validation of the length, that's the purpose of the next test in the file, but the purpose of this one is to test that these characters are all accepted in the validation.

hallelujah: OK but I did not understand the slice if you are sure that they are less than 255 chars.

Martouta: Right, I put it there while testing and forgot about it 👍 I'll remove it 😄 👍

PR #964, opened 2019-07-05
Creates a separate test database for seeds test

guicassolato:

It makes the seeds test to always run in its own database.

Closes [THREESCALE-2854](https://issues.jboss.org/browse/THREESCALE-2854)

For CircleCI only, an alternative could be not to run the seeds test among the others and create a separate job for it setting the `DATABASE_URL` of the test environment with a different and exclusive db, say by tweaking with `TEST_ENV_NUMBER`.

We could as well just clear the test database with `db:data:clear`, but I'm afraid that could mess with concurrent tests eventually pointing to the same db.

### DIFF
--- a/test/unit/db/seeds_test.rb
+++ b/test/unit/db/seeds_test.rb
@@ -3,7 +3,36 @@
 require 'test_helper'
 
 class SeedsTest < ActiveSupport::TestCase
+  setup do
+    with_fake_env('test_seed') do
+      System::Application.load_tasks
+      Rake::Task['db:create'].invoke
+      Rake::Task['db:schema:load'].invoke
+    end
+
+  end
+
+  attr_reader :env
+
+  teardown do
+    with_fake_env('test_seed') { Rake::Task['db:drop'].invoke }
+  end
+
   test 'the seeds do not fail' do
+    ActiveRecord::Base.establish_connection(:test_seed)
     assert Rails.application.load_seed
+    ActiveRecord::Base.establish_connection(:test)
+  end
+
+  private
+
+  def with_fake_env(new_env)
+    old_env = Rails.env
+    begin
+      Rails.env = ENV['RAILS_ENV'] = new_env
+      yield
+    ensure
+      Rails.env = ENV['RAILS_ENV'] = old_env
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


hallelujah:
Well the random error is probably because of Master been created and not deleted in precedent test.
So it would probably be better to fix that because it will cause another random error in another test

guicassolato:
@hallelujah, thanks for the tip. This is just a draft though. I wanted to see whether we can find an easy solution before digging too much (and for too long) into the actual root cause of the problem.

test/unit/db/seeds_test.rb
```diff
@@ -3,7 +3,30 @@
...
+  class SeedDatabaseConnection < ActiveRecord::Base
+    def self.abstract_class?
+      true
+    end
+  end
+
+  setup do
+    @config = ActiveRecord::Base.configurations['test_seed']
+    adapter = ActiveRecord::Tasks::DatabaseTasks.send(:class_for_adapter, config['adapter']).new(config)
+    creation_options = adapter.send(:creation_options)
```
guicassolato: This only works for MySQL.

https://github.com/rails/rails/blob/7f2327f65b4117384bec3547ccd5a99028df57e0/activerecord/lib/active_record/tasks/mysql_database_tasks.rb#L16
https://github.com/rails/rails/blob/7f2327f65b4117384bec3547ccd5a99028df57e0/activerecord/lib/active_record/tasks/postgresql_database_tasks.rb#L15
https://github.com/rsim/oracle-enhanced/blob/2df95002d0352a77a54d075a28fa4dcdabf2b700/lib/active_record/connection_adapters/oracle_enhanced/database_tasks.rb

guicassolato:
This ain't gonna work 😞. I think we cannot really avoid `establish_connection` from leaking to the other threads (e.g. https://circleci.com/gh/3scale/porta/71185#tests/containers/4).

Let's switch to the approach in https://github.com/3scale/porta/pull/964#issuecomment-508721750 to fix [THREESCALE-2854](https://issues.jboss.org/browse/THREESCALE-2854). Instead of a separate db for the seeds test, we need to identity the actual test that is not properly cleaning the db in the (after_)teardown, thus making `SeedsTest` to fail with _"Master can be only one"_.

Closing this.

PR #972, opened 2019-07-05
Clean db before seeds test

guicassolato:

It enforces the test database to be clean before the seeds test (so Master for sure does not exist)

Closes [THREESCALE-2854](https://issues.jboss.org/browse/THREESCALE-2854)

Alternative approach to https://github.com/3scale/porta/pull/964

### DIFF
--- a/test/unit/db/seeds_test.rb
+++ b/test/unit/db/seeds_test.rb
@@ -3,11 +3,6 @@
 require 'test_helper'
 
 class SeedsTest < ActiveSupport::TestCase
-  setup do
-    # some annoying test is failing to clear the db in the teardown making this one to randomly fail, so adding this to ensure the test database is "virgin"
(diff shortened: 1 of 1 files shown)


test/unit/db/seeds_test.rb
```diff
@@ -3,6 +3,13 @@
 require 'test_helper'
 
 class SeedsTest < ActiveSupport::TestCase
+  setup do
+    # some annoying test is failing to clear the db in the teardown making this one to randomly fail, so adding this to ensure the test database is "virgin"
+    DatabaseCleaner.clean
```
Martouta: We should remove this now 😄 With the `disable_transactional_fixtures!` added below should be enough

guicassolato: Done

PR #973, opened 2019-07-06
THREESCALE-2864 Re-enable the JS toggle for stripe

hallelujah:
It was removed by accident

Closes [THREESCALE-2864](https://issues.jboss.org/browse/THREESCALE-2864)

Just a quick revert copy/paste

### DIFF
--- a/app/views/payment_gateways/_stripe_elements.html.erb
+++ b/app/views/payment_gateways/_stripe_elements.html.erb
@@ -39,6 +39,14 @@
 </div>
 <% billing_address = current_account.billing_address %>
 <script type="text/javascript" charset="utf-8">
+  stripeForm = $( "#stripe-form" )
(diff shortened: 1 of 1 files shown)


josemigallas:
@hallelujah would be good to point to the relevant commit and PR next time. In this case the one that removed those lines.

Martouta:
I am reaaaaaally lost right now 😕 
Isn't this the exact same code that we currently have in here?:
https://github.com/3scale/porta/blob/a5fc806ec8e713e56cc9db5a14dc2924f0829e21/app/views/payment_gateways/_stripe.html.erb#L77-L84

Have you tested this and seen a difference? can you explain how or write a test for it?
Actually I think that we even currently have tests for this and are passing (or false positives otherwise 😱 ). And Gui and I checked on his account in production and it worked well 🤷‍♀

PR #975, opened 2019-07-08
Upload logo only allows images

Martouta:
Closes [THREESCALE-2633](https://issues.jboss.org/browse/THREESCALE-2633)

We have to keep this in mind too:
image


app/models/profile.rb
```diff
@@ -48,7 +48,8 @@ class Profile < ApplicationRecord
     :url => ':url_root/:account_id/:class/:attachment/:style/:basename.:extension'.freeze,
     :s3_permissions => 'public-read'.freeze,
     :default_url => '/assets/3scale-logo.png'.freeze
-  do_not_validate_attachment_file_type :logo
+  # do_not_validate_attachment_file_type :logo
+  validates_attachment_content_type :logo, content_type: /^image\/(png|gif|jpeg)/
```
thomasmaas: @Martouta more and more logo's are vector format these days so we should consider accepting svg. Not sure on the security implications.

Martouta: @thomasmaas Ok, I will check 👍

app/models/profile.rb
```diff
@@ -44,11 +44,11 @@ class Profile < ApplicationRecord
 
   # Profile has attached logo.
   has_attached_file :logo,
-    :styles => { large: '300x300>'.freeze, medium: '150x150>'.freeze, thumb: '100x100>'.freeze, invoice: ['200x50>'.freeze, :png].freeze }.freeze,
+    style: { large: '300x300>'.freeze, medium: '150x150>'.freeze, thumb: '100x100>'.freeze, invoice: ['200x50>'.freeze, :png].freeze }.freeze,
     :url => ':url_root/:account_id/:class/:attachment/:style/:basename.:extension'.freeze,
     :s3_permissions => 'public-read'.freeze,
     :default_url => '/assets/3scale-logo.png'.freeze
-  do_not_validate_attachment_file_type :logo
+  validates_attachment_content_type :logo, content_type: %r{^image\/(png|gif|jpeg|svg)}
```
Martouta: Not sure about the security of svg yet 🤔 I am checking this yet 👍

app/controllers/provider/admin/account/logos_controller.rb
```diff
@@ -1,38 +1,34 @@
...
-    respond_to do |format|
-      format.html do
-        flash[:notice] = @notice
-        redirect_to edit_provider_admin_account_logo_path
-      end
-
-      format.js
-    end
+    logo = params.require(:profile).require(:logo)
+    update_logo(logo: logo, success_message: 'Logo was successfully created.')
```
Martouta: I will probably add this message to use i18n

app/controllers/provider/admin/account/logos_controller.rb
```diff
@@ -1,38 +1,34 @@
...
   end
 
   def destroy
-    authorize! :delete, :logo
+    update_logo(logo: nil, success_message: 'Your logo was successfully deleted.')
```
Martouta: Same as https://github.com/3scale/porta/pull/975/files#r302962048

app/models/profile.rb
thomasmaas: any surer? if not let's not do it.

app/controllers/provider/admin/account/logos_controller.rb
```diff
@@ -1,38 +1,33 @@
...
+      flash[:notice] = t("#{action_name}_success", scope: 'logos_controller')
+    else
+      flash[:error] = profile.errors.full_messages.to_sentence
+    end
     redirect_to edit_provider_admin_account_logo_path
   end
 
-  private
+  def authorize
+    authorize! action_name, :logo
```
hallelujah: Hmm for the `destroy` action it would be `authorize 'destroy', :logo` instead of `authorize :delete, :logo` as it was https://github.com/3scale/porta/pull/975/files#diff-5512c93ae956a903f6808306318f5dc3L27

Martouta: @hallelujah 
When I made this change I checked in the abilities to add the "destroy" there next to the "delete", and what I found is the only permission we are giving is actually "manage".

So I think we should not complicate the code knowing that.

I also added the tests for the conditions under it should be forbidden.

app/models/profile.rb
Martouta: @hallelujah WDYT?

hallelujah: Let allow SVG, I do not see how users could exploit it.

test/integration/provider/admin/account/logos_controller_test.rb
```diff
@@ -0,0 +1,91 @@
...
+    assert_change(of: -> { provider.profile.reload.logo_file_name }, from: provider.profile.logo_file_name, to: nil) do
+      delete provider_admin_account_logo_path, {profile: {logo: logo_file}}
+    end
+
+    assert_redirected_to edit_provider_admin_account_logo_path
+    assert_equal 'Your logo was successfully deleted.', flash[:notice]
+    assert_nil flash[:error]
+  end
+
+  test 'destroy when it should fail' do
```
hallelujah: I do not get this test.
Can't you provide pre-condition on when it will fail? The stubs are strange to me

Martouta: @hallelujah The controller before this test was skipping validations and this tests means to ensure that this doesn't happen and if any validation makes it fail, then it should fail.
Right now there is no validation that would make it fail and that's why I've done it like this.

test/unit/cms/file_test.rb
```diff
@@ -39,6 +39,9 @@ class CMS::FileTest < ActiveSupport::TestCase
     file.attachment = Rails.root.join('test', 'fixtures', 'test.css').open
 
     assert_equal 'text/css', file.attachment_content_type
+
+    file.valid?
```
hallelujah:
```suggestion
    refute file.valid?`
```

Martouta: I don't mind doing this 👍 but this wasn't meant to be a validation originally... just called to get the errors 😄

test/integration/provider/admin/account/logos_controller_test.rb
Martouta: So what should I do about this? @hallelujah

app/models/profile.rb
Martouta: Something very weird is going on... I am testing the code in preview

Now it works well in preview (and it would be the same in production) for regular image files  (jpg, gif, png), the reason why it wasn't working before in preview but now yes is this last commit: https://github.com/3scale/porta/pull/975/commits/d81e3d9c4e8ebed2655ed63bd99eccc82f3ce1b6
What that does is make it work in preview/production but not locally in development (at least for me). I guess this happens because the version of the paperclip gem or something like that requires one or the other... something like that... 🤷‍♀ 

But now the weirdest part: If I test now in preview adding a SVG file... it fails, if I try locally (without the commit link I sent u, so I can test updating images locally) the SVG works well 🤷‍♀

I think I can fix the svg for preview and production anyway but it is a bit hard without being able to reproduce it locally and without being able to write a test that executes the same environment as preview for CircleCI 🤔 

I told Thomas and he said that he is not sure if worth the effort as we also have the possible security threats related to svg and that he would just do this pr without the svg part so we can merge and move on but up to me.

WDYT @hallelujah ?

Martouta: I've just pushed another commit removing the support for SVG 😄 It is in its own commit so if anyone disagrees, it is still possible to remove it before merging 😄

app/controllers/provider/admin/account/logos_controller.rb
```diff
@@ -1,38 +1,33 @@
...
-    authorize! :delete, :logo
+    update_logo(logo: nil)
+  end
 
-    profile.update_attribute(:logo, nil)
+  private
 
-    flash[:notice] = 'Your logo was successfully deleted.'
+  def update_logo(logo:)
+    if profile.update({logo: logo})
```
macejmic: why this change? `update` instead of `update_attribute` ?

Martouta: @macejmic because we have added validations in the model and they are skipped with `update_attribute` which is not what we want 😄 We want it to check the validations 😉

macejmic: i see. ok, makes sense. just didn't know what exactly does `update`.. needed to look to documentation 👯

test/integration/provider/admin/account/logos_controller_test.rb
macejmic: can you try to upload not valid file? should be the same as stubbing `valid?: false`.. right?

Martouta: @macejmic That's what I did for the `update` (see https://github.com/3scale/porta/pull/975/files#diff-cdaeffaabb0811cb7c29c744223afd65R28) but I can't do that for the `destroy` because it only takes `nil` and not the params (see https://github.com/3scale/porta/pull/975/files#diff-5512c93ae956a903f6808306318f5dc3R14)

macejmic: i got it now.. you're testing everything! 😺 i'm ok with that..

config/locales/en.yml
```diff
@@ -23,6 +23,9 @@ en:
     apicast_version_not_upgraded: 'The APIcast version for this service could not be upgraded'
     oidc_not_available_on_old_apicast: 'OpenID Connect is not supported for the old APIcast. Pick a different authentication option before reverting back.'
     edit_deployment_configuration: 'edit %{deployment} configuration'
+  logos_controller:
+    update_success: 'Logo was successfully created.'
```
thomasmaas:
```suggestion
    update_success: 'The logo was successfully uploaded.'
```

config/locales/en.yml
```diff
@@ -23,6 +23,9 @@ en:
...
+    destroy_success: 'Your logo was successfully deleted.'
```
thomasmaas:
```suggestion
    destroy_success: 'The logo was successfully deleted.'
```

test/integration/provider/admin/account/logos_controller_test.rb
```diff
@@ -0,0 +1,91 @@
...
+
+  attr_reader :provider
+
+  test 'update when it should work' do
+    assert_change(of: -> { provider.profile.reload.logo_file_name }, from: nil, to: 'hypnotoad.jpg') do
+      put provider_admin_account_logo_path, {profile: {logo: logo_file}}
+    end
+
+    assert_redirected_to edit_provider_admin_account_logo_path
+    assert_equal 'Logo was successfully created.', flash[:notice]
```
thomasmaas:
```suggestion
    assert_equal 'The logo was successfully uploaded.', flash[:notice]
```

test/integration/provider/admin/account/logos_controller_test.rb
```diff
@@ -0,0 +1,91 @@
...
+
+  test 'destroy when it should work' do
+    provider.profile.update_attribute(:logo, Rack::Test::UploadedFile.new(logo_file, 'image/jpeg', true))
+
+    assert_change(of: -> { provider.profile.reload.logo_file_name }, from: provider.profile.logo_file_name, to: nil) do
+      delete provider_admin_account_logo_path, {profile: {logo: logo_file}}
+    end
+
+    assert_redirected_to edit_provider_admin_account_logo_path
+    assert_equal 'Your logo was successfully deleted.', flash[:notice]
```
thomasmaas:
```suggestion
    assert_equal 'The logo was successfully deleted.', flash[:notice]
```

PR #976, opened 2019-07-09
Fixing PF4 BG image in firefox

damianpm:
In Firefox the BG image of PF4 Loginpage component is not rendering (in the signup form, with invitation email).

https://issues.jboss.org/browse/THREESCALE-2991

**Verification steps** 

- Go to **Account Settings > Users > Invitations**
- Send an invitation
- Logout
- In the terminal, look for the link sent
- Go to the link in Firefox

### DIFF
--- a/app/javascript/packs/LoginPage.js
+++ b/app/javascript/packs/LoginPage.js
@@ -12,10 +12,10 @@ import {safeFromJsonString} from 'utilities/json-utils'
 document.addEventListener('DOMContentLoaded', () => {
   const oldLoginWrapper = document.getElementById('old-login-page-wrapper')
   oldLoginWrapper.removeChild(document.getElementById('old-login-page'))
-  const loginPageContainer = document.getElementById('login-page-container')
+  const PFLoginPageContainer = document.getElementById('pf-login-page-container')
   if (isIE11) {
-    loginPageContainer.classList.add('isIe11', 'pf-c-page')
+    PFLoginPageContainer.classList.add('isIe11', 'pf-c-page')
   }
-  const loginPageProps = safeFromJsonString(loginPageContainer.dataset.loginProps)
-  LoginPageWrapper(loginPageProps, 'login-page-container')
+  const loginPageProps = safeFromJsonString(PFLoginPageContainer.dataset.loginProps)
+  LoginPageWrapper(loginPageProps, 'pf-login-page-container')
 })
--- a/app/javascript/packs/SignUpPage.js
+++ b/app/javascript/packs/SignUpPage.js
@@ -10,14 +10,16 @@ if (isIE11) {
 }
 
 document.addEventListener('DOMContentLoaded', () => {
-  const signupPageContainer = document.getElementById('signup-page-container')
+  const loginPageContainer = document.getElementById('login-page-container')
   if (isIE11) {
-    signupPageContainer.classList.add('isIe11', 'pf-c-page')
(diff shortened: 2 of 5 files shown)


app/views/provider/invitee_signups/show.html.erb
```diff
@@ -1,6 +1,16 @@
 <%= javascript_pack_tag 'SignUpPage' %>
 
-<div id="old-signup-page-wrapper">
+<%= tag("div", :id => "login-page-container", :"data-signup-props" => {:path => provider_invitee_signup_path(@invitation.token),
```
didierofrivia: is it really a login page? by the look of the properties seems to be a signup

app/javascript/packs/SignUpPage.js
```diff
@@ -10,14 +10,16 @@ if (isIE11) {
...
   if (isIE11) {
-    signupPageContainer.classList.add('isIe11', 'pf-c-page')
+    loginPageContainer.classList.add('isIe11', 'pf-c-page')
   }
 
-  const loginLayout = document.querySelector('.login-layout')
-  loginLayout.removeChild(document.getElementById('old-signup-page-wrapper'))
+  let oldLoginWrapper = document.getElementById('old-login-page-wrapper')
+  if (oldLoginWrapper.parentNode) {
+    oldLoginWrapper.parentNode.removeChild(oldLoginWrapper)
```
didierofrivia: `oldLoginWrapper.remove()` ?

app/javascript/packs/LoginPage.js
```diff
@@ -12,10 +12,10 @@ import {safeFromJsonString} from 'utilities/json-utils'
 document.addEventListener('DOMContentLoaded', () => {
   const oldLoginWrapper = document.getElementById('old-login-page-wrapper')
   oldLoginWrapper.removeChild(document.getElementById('old-login-page'))
-  const loginPageContainer = document.getElementById('login-page-container')
+  const PFLoginPageContainer = document.getElementById('pf-login-page-container')
```
damianpm: @didierofrivia This is the container for the `Patternfly LoginPage` Component. Inside the component goes the login, request password or signup form.
I renamed it to PFLoginPageContainer to make ir more clear

app/javascript/packs/SignUpPage.js
damianpm: Unfortunately, `remove()` is not supported in IE11

didierofrivia: But we are using the babel polyfill in the end, no?

app/javascript/packs/LoginPage.js
josemigallas: Why the change of name anyway?

app/views/provider/invitee_signups/show.html.erb
damianpm: @josemigallas the change in the name is because of this

app/javascript/packs/SignUpPage.js
damianpm: The problem is `'@babel/polyfill'` doesn't polyfills this specific feature. I've tried.
We should include another polyfill: https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/remove#Polyfill
But I think is too much, I wouldn't include it
@josemigallas @didierofrivia

josemigallas: Let's leave it like this then

PR #981, opened 2019-07-09
[WIP][Outdated]When deleting in parallel, try once to unstore credit card before destroying the buyer

Martouta:
Closes [THREESCALE-3637](https://issues.jboss.org/browse/THREESCALE-3637)
Remove all buyers (+ unstore credit cards) before the PaymentSetting

**VERY WIP**

### DIFF
--- a/app/models/account/payment_details.rb
+++ b/app/models/account/payment_details.rb
@@ -4,7 +4,7 @@ module Account::PaymentDetails
   extend ActiveSupport::Concern
 
   included do
-    has_one :payment_detail, -> { order id: :desc }, autosave: true
+    has_one :payment_detail, -> { order id: :desc }, autosave: true, dependent: :delete
 
     CREDIT_CARD_ATTRIBUTES = [
       :credit_card_auth_code,
--- a/app/workers/delete_object_hierarchy_worker.rb
+++ b/app/workers/delete_object_hierarchy_worker.rb
@@ -130,6 +130,8 @@ class DeleteObjectHierarchyWorker < ActiveJob::Base
         DeleteAccountHierarchyWorker
       when Service.name
         DeleteServiceHierarchyWorker
+      when PaymentGatewaySetting.name
+        DeletePaymentSettingHierarchyWorker
       else
         DeleteObjectHierarchyWorker
       end
--- a/app/workers/delete_payment_setting_hierarchy_worker.rb
+++ b/app/workers/delete_payment_setting_hierarchy_worker.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+class DeletePaymentSettingHierarchyWorker < DeleteObjectHierarchyWorker
+  alias payment_setting object
+
+  def delete_associations
+    destroy_users_of_provider(payment_setting.account) if called_from_provider_hierarchy?
+
+    super
+  end
+
+  private
+
+  def called_from_provider_hierarchy?
+    caller_worker_hierarchy.include?("Hierarchy-Account-#{payment_setting.account_id}")
+  end
+
+  def destroy_users_of_provider(provider)
+    provider.buyer_account_ids.each do |associated_object_id|
+      associated_object = Account.new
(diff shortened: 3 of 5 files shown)


app/workers/delete_object_hierarchy_worker.rb
```diff
@@ -53,7 +53,7 @@ def on_finish(method_name, options)
   def build_batch
     batch = Sidekiq::Batch.new
     batch.description = batch_description
-    batch_callbacks(batch) { batch.jobs { delete_associations(object) } }
+    batch_callbacks(batch) { batch.jobs { delete_associations } }
```
Martouta: because `object` is [an instance variable](https://github.com/3scale/porta/blob/de3b4578961b7b1c6ccf8cf3f315a84943aec7ef/app/workers/delete_object_hierarchy_worker.rb#L51).

app/workers/delete_object_hierarchy_worker.rb
```diff
@@ -74,7 +74,7 @@ def batch_callbacks(batch)
     end
   end
 
-  def delete_associations(object)
+  def delete_associations
```
Martouta: Same as https://github.com/3scale/porta/pull/981/files#r304674232

Martouta: because `object` is [an instance variable](https://github.com/3scale/porta/blob/de3b4578961b7b1c6ccf8cf3f315a84943aec7ef/app/workers/delete_object_hierarchy_worker.rb#L51).

app/models/account/gateway.rb
```diff
@@ -3,7 +3,7 @@ module Account::Gateway
 
   included do
     has_many :payment_transactions
-    has_one :payment_gateway_setting, dependent: :destroy, inverse_of: :account
+    has_one :payment_gateway_setting, dependent: :destroy, inverse_of: :account, class_name: 'PaymentGatewaySetting'
```
guicassolato: Why adding the `class_name`. `'PaymentGatewaySetting'` is what Rails will infer by default, no?

app/workers/delete_payment_setting_hierarchy_worker.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+class DeletePaymentSettingHierarchyWorker < DeleteObjectHierarchyWorker
+  alias payment_setting object
+
+  def delete_associations
+    destroy_users_of_provider(payment_setting.account) if called_from_hierarchy?
+
+    super
```
guicassolato: Does this mean that we will enqueue the deletion of each buyer account twice when a provider account is deleted?

Martouta: So far yes but I intend to change that by slightly changing the `DeleteAccountHierarchyWorker` :smile: But I am focused in making it work first :disappointed:

app/models/account/gateway.rb
Martouta: Because otherwise it is `nil` and doesn't go through `DeletePaymentSettingHierarchyWorker` but `DeleteObjectHierarchyWorker` instead.
That is in the tests so you can see it if you remove this change and run the new tests :smile:

app/models/account/payment_details.rb
```diff
@@ -4,7 +4,7 @@ module Account::PaymentDetails
   extend ActiveSupport::Concern
 
   included do
-    has_one :payment_detail, -> { order id: :desc }, autosave: true
+    has_one :payment_detail, -> { order id: :desc }, autosave: true, dependent: :delete
```
Martouta: Because:
1. This way it doesn't stay in the DB after removing the account
2. It does not get destroyed before the unstore.

This is a temporary step and **does not meant to be merged**, but it is there so it doesn't stop my tests to make things work first 😄

Martouta:
This is very outdates :smile: And had a broader goal :smile: so just opening a new one instead :wink:

PR #982, opened 2019-07-09
[DeleteObjectHierarchyWorker] Speeding up enqueueing job deletion

Martouta:
Closes [THREESCALE-2987 - Delete in background : New loading logic (more efficient)](https://issues.jboss.org/browse/THREESCALE-2987)

- Results of the tests done in preview with the code that we currently have in production: https://github.com/3scale/infrastructure/issues/746#issuecomment-509272493
- Results of the tests done in preview for the same data with the code of this task:
https://github.com/3scale/infrastructure/issues/746#issuecomment-509288005

### DIFF
--- a/app/workers/delete_object_hierarchy_worker.rb
+++ b/app/workers/delete_object_hierarchy_worker.rb
@@ -75,13 +75,9 @@ class DeleteObjectHierarchyWorker < ActiveJob::Base
   end
 
   def delete_associations(object)
-    object_associations = associations_to_destroy_for(object)
-    association_names = object_associations.map(&:name)
-    preloaded_object = association_names.any? ? object.class.eager_load(*association_names).find(object.id) : object
-
-    object_associations.each do |reflection|
-      worker = association_delete_worker(reflection)
-      delete_objects_of_association(preloaded_object, reflection.name, worker)
+    object.class.reflect_on_all_associations.each do |reflection|
+      next unless reflection.options[:dependent] == :destroy
+      ReflectionDestroyer.new(object, reflection, caller_worker_hierarchy).destroy_later
     end
   end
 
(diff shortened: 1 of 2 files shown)


app/workers/delete_object_hierarchy_worker.rb
```diff
@@ -91,25 +86,44 @@ def callback_options
...
-  end
+    attr_reader :main_object, :reflection, :caller_worker_hierarchy
 
-  def association_delete_worker(reflection)
-    delete_hierarchy_workers = Hash.new(DeleteObjectHierarchyWorker).merge('Account' => DeleteAccountHierarchyWorker)
-    delete_hierarchy_workers[reflection.options[:class_name]]
+    private
+
+    def destroy_has_many_association
+      main_object.public_send(reflection.name).ids.each do |associated_object_id|
```
hallelujah:
```suggestion
      main_object.public_send("#{reflection.name}_ids").each do |associated_object_id|
```

Martouta: @hallelujah that doesn't work
image

Martouta: ok, this works:
```ruby
main_object.public_send("#{reflection.name.to_s.singularize}_ids").each do |associated_object_id|
```

app/workers/delete_object_hierarchy_worker.rb
```diff
@@ -91,25 +86,44 @@ def callback_options
...
+      associated_object = main_object.public_send(reflection.name)
+      delete_associated_object_later(associated_object) if associated_object
+    end
+
+    def delete_associated_object_later(associated_object)
+      association_delete_worker.perform_later(associated_object, caller_worker_hierarchy)
+    end
+
+    def association_delete_worker
+      association_workers = Hash.new(DeleteObjectHierarchyWorker).merge('Account' => DeleteAccountHierarchyWorker)
```
hallelujah: That could be a simple `if` no?

app/workers/delete_object_hierarchy_worker.rb
```diff
@@ -91,25 +86,44 @@ def callback_options
...
+    attr_reader :main_object, :reflection, :caller_worker_hierarchy
 
-  def association_delete_worker(reflection)
-    delete_hierarchy_workers = Hash.new(DeleteObjectHierarchyWorker).merge('Account' => DeleteAccountHierarchyWorker)
-    delete_hierarchy_workers[reflection.options[:class_name]]
+    private
+
+    def destroy_has_many_association
+      main_object.public_send(reflection.name).ids.each do |associated_object_id|
+        associated_object = reflection.class_name.constantize.new
```
guicassolato: We have a a few `after_initialize` defined. Hopefully they won't be a problem here.

Martouta: good catch, i will check 🤔 👍

app/workers/delete_object_hierarchy_worker.rb
```diff
@@ -75,13 +75,8 @@ def batch_callbacks(batch)
...
 
   def delete_associations(object)
-    object_associations = associations_to_destroy_for(object)
-    association_names = object_associations.map(&:name)
-    preloaded_object = association_names.any? ? object.class.eager_load(*association_names).find(object.id) : object
-
-    object_associations.each do |reflection|
-      worker = association_delete_worker(reflection)
-      delete_objects_of_association(preloaded_object, reflection.name, worker)
+    object.class.reflect_on_all_associations.each do |reflection|
```
guicassolato: This will also return `belong_to` associations which will later be rejected because of https://github.com/3scale/porta/blob/d8dc41e0389b370b62a90126fcd14116cf742781/app/workers/delete_object_hierarchy_worker.rb#L97. Can we only created `ReflectionDestroyer` objects for the associations that really matter?

app/workers/delete_object_hierarchy_worker.rb
Martouta: @guicassolato no problem, we don't care about any value except for the ID which we update in the next line 👍

guicassolato: Try:

```ruby
> Account.reflect_on_all_associations.find { |a| a.name == :provided_access_tokens }.class_name
NoMethodError: undefined method `klass' for nil:NilClass
Did you mean?  class
```

app/workers/delete_object_hierarchy_worker.rb
```diff
@@ -91,25 +87,43 @@ def callback_options
...
+    attr_reader :main_object, :reflection, :caller_worker_hierarchy
 
-  def association_delete_worker(reflection)
-    delete_hierarchy_workers = Hash.new(DeleteObjectHierarchyWorker).merge('Account' => DeleteAccountHierarchyWorker)
-    delete_hierarchy_workers[reflection.options[:class_name]]
+    private
+
+    def destroy_has_many_association
+      main_object.public_send("#{reflection.name.to_s.singularize}_ids").each do |associated_object_id|
+        associated_object = reflection.class_name.constantize.new
```
guicassolato: Try:

```ruby
> Account.reflect_on_all_associations.find { |a| a.name == :provided_access_tokens }.class_name
NoMethodError: undefined method `klass' for nil:NilClass
Did you mean?  class
```

app/workers/delete_object_hierarchy_worker.rb
guicassolato: Update: `dependent: destroy` can be in `belongs_to` when used in conjunction with `has_one`. (It can also be when in conjunction with `has_many` although expressly recommended otherwise in the Rails documentation.)

For example:
```ruby
class Parent < ApplicationRecord
  has_one :child
end

class Child < ApplicationRecord
  belongs_to :parent, dependent: :destroy
end
```

Will deleting the `Child` cause to try to delete the `Parent` too?

Martouta: I don't know, but that is what `dependent: :destroy` means so it definitely should 🤔But I will check 👍

app/workers/delete_object_hierarchy_worker.rb
Martouta: @guicassolato u definitely found a bug but not related to this PR 😂 look at this:
image
and:
image

Because the association 'user' doesn't exist for accounts, it is 'users' 😄

Martouta: Fix in: https://github.com/3scale/porta/pull/984

PR #983, opened 2019-07-09
Moving IE11 detection to a utils file

damianpm:

Extract IE 11 detection method to a utils file

https://issues.jboss.org/browse/THREESCALE-2992

### DIFF
--- a/app/javascript/src/utilities/ie11Utils.js
+++ b/app/javascript/src/utilities/ie11Utils.js
@@ -1,3 +1,3 @@
-const isBrowserIE11 = () => !!window.navigator.userAgent.match(/Trident\/7\./)
+const isBrowserIE11 = (win = window) => !!win.navigator.userAgent.match(/Trident\/7\./)
 
 export { isBrowserIE11 }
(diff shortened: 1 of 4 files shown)


spec/javascripts/utilities/ie11Utils.spec.jsx
```diff
@@ -0,0 +1,14 @@
+import { isBrowserIE11 } from 'utilities/ie11Utils'
+
+describe('isBrowserIE11', () => {
+  it('should return false if browser is not IE11', () => {
```
josemigallas: I think the correct statement would be _it should return false if the user agent is other than Trident/7.0_.

damianpm: Updated!

app/javascript/src/utilities/ie11Utils.js
```diff
@@ -0,0 +1,3 @@
+const isBrowserIE11 = () => !!window.navigator.userAgent.match(/Trident\/7\./)
```
didierofrivia: use DI here for easier and isolated testing, like:

```
 const isBrowserIE11 = (win = window) => !!win.navigator.userAgent.match(/Trident\/7\./)
```

then in the test you could inject the value you want.

damianpm: Good idea, it's updated @didierofrivia !

PR #985, opened 2019-07-10
[WIP] Fix Links decoration

josemigallas:

Since the introduction to Patternfly4 React in the Vertical Navigation, links styles are overridden. These new styles are added inline and in a the last place (after all rails stylesheets have been loaded).

Therefore the best solution is to use `!important` os that our style has the priority. Eventually when we fully implement PF4 it will be as simple as removing this very line.

link

[THREESCALE-2994: links in dashboard and popup menu's shouldn't underline on hover](https://issues.jboss.org/browse/THREESCALE-2994)

**Verification steps** 

* In Vertical Navigation: titles, but not subtitles, should have an underline.
* In the rest of the UI: no link should have an underline.

### DIFF
--- a/app/assets/stylesheets/provider/_links.scss
+++ b/app/assets/stylesheets/provider/_links.scss
@@ -1,9 +1,10 @@
 p a {
-  text-decoration: underline;
   color: $font-color;
+  text-decoration: underline;
 
   &:hover {
     color: $link-color;
+    text-decoration: underline;
   }
 }
 
--- a/app/assets/stylesheets/provider/_menu.scss
+++ b/app/assets/stylesheets/provider/_menu.scss
@@ -55,6 +55,7 @@ ul.login-form-links {
 
       &:hover {
         color: $menu-hover-color;
+        text-decoration: underline;
       }
     }
   }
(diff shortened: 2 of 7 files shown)


app/assets/stylesheets/provider/_links.scss
```diff
@@ -8,7 +8,7 @@ p a {
 }
 
 a {
-  text-decoration: none;
+  text-decoration: none !important; // Overrides Patternfly 4 links decoration introduced by @patternfly/react
```
thomasmaas: doesn't this override it everywhere? Links in other places should probably be underlined. But that might be already in place with more specific style rules in our css.

josemigallas: I haven't looked very thoroughly, it might override some element when it shouldn't.

thomasmaas: for example  on dashboard should stay underlined so please check.

thomasmaas: I'd say a more specific override for vertical nav, horizontal dashboard nav & popup menus is safer

josemigallas: What about all the links in the main content? Those shouldn't have an underline either.

PR #986, opened 2019-07-10
Load widgets in Dashboard only when opening a Service

josemigallas:

Services charts are loaded even when they are not visible. The higher the number of services, the more time the UI takes to load.

This is just a first step to improve Dashboard's performance.

[THREESCALE-2916: Reduce dashboard loading time](https://issues.jboss.org/browse/THREESCALE-2916)

**Verification steps** 

* Services that are not open (charts not visible) should not send any XHR (see browser's network tab)
* When opening a service, requests should be sent and charts should load eventually.
* When a service is open by default (at loading time), the charts should also be loaded automatically.

**Performance results**
> With 28 services.
* Before (all widgets are loaded):
Screen Shot 2019-07-10 at 17 58 04

* After (only visible widgets load):
Screen Shot 2019-07-10 at 17 59 27

The background work is dramatically reduced, although it is all non-blocked time, keep in mind this had been recorded with only 28 services.


app/views/provider/admin/dashboards/_ajax_load.html.slim
```diff
@@ -1,4 +1,23 @@
...
+      dashboardWidget(widgetPath)
+    } else {
+      let value = window.localStorage['toggle:service_' + serviceId]
+      if (value && !JSON.parse(value)['is-closed']) {
+        dashboardWidget(widgetPath)
+      }
+
+      let container = document.getElementById('service_' + serviceId)
+      let title = container.querySelector('.DashboardSection-toggle')
+      title.addEventListener('click', function () {
```
Martouta: this should be inside an `if` verifying that `title` has actually found. Because when the connection with backend fails, this title doesn't exist and it will raise an error trying the `addEventListener` over an element that does not exist  😄 An example of this is locally in development 😄

josemigallas: Good catch, thanks!

features/old/accounts/bulk_operations.feature
```diff
@@ -1,4 +1,4 @@
-@javascript
+@javascript @ajax
```
hallelujah: I do not think we need that tag

features/percy/provider.feature
```diff
@@ -1,4 +1,4 @@
-@percy @javascript @selenium
```
hallelujah: Why did you remove `@percy` tag?
This is used against Percy

features/support/hooks.rb
```diff
@@ -97,7 +97,7 @@
 end
 
 AfterStep '@javascript', '@ajax' do
-  step 'wait for ajax'
+  wait_for_requests
```
hallelujah: I do not think it is wise to use the middleware everytime for `@javascript`
That will slow the tests as it needs to memoize the requests done

features/percy/provider.feature
Martouta: Right, my bad 👍 It was my mistake 👍

features/support/hooks.rb
Martouta: @hallelujah This is not every time for `@javascript` but for the `@ajax` tag.

The problem before this commit was that many cukes where failing because it was waiting for an ajax that was never being called (for the dashboard) so it was waiting forever... now it waits only if there are requests being done.

Although I would agree that adding the wait in those specific steps that need it instead of the whole scenario would be much faster, but then imo that would mean removing the `@ajax` tag completely and adapting it to that and that would be out of scope of this PR (because I don't think that this PR makes the current situation worse).

features/percy/provider.feature
josemigallas: @Martouta

features/old/accounts/bulk_operations.feature
josemigallas: @Martouta

Martouta: Same as https://github.com/3scale/porta/pull/986/files#r304489102 :)

Martouta: Anyway guys feel free to do whatever u consider with that commit, even completely remove it if u consider that's the best idea 👍

features/support/hooks.rb
Martouta: UPDATE: @hallelujah told me this does it for both tags, not for the combination. I wanted it only for `@ajax`, sorry 👍

features/step_definitions/dashboard_steps.rb
```diff
@@ -30,6 +30,7 @@ def top_traffic_for_name (name)
 
   assert hits.has_css? '.DashboardWidget-spinner'
   assert top_traffic.has_css? '.DashboardWidget-spinner'
+  wait_for_requests
```
josemigallas: @hallelujah for some reason this is not working anymore 🤔

features/old/accounts/service_contracts.feature
```diff
@@ -39,22 +39,19 @@ Feature: Account service plans management
      And I should see "Fancy API"
 
     When I follow "Subscribe to Fancy API"
-     And I wait until all animations have finished
```
hallelujah: Nice removal, this is not something we should indicate to the user.
It carries no useful information 🎖

app/javascript/src/Dashboard/ajax-widget.jsx
```diff
@@ -0,0 +1,34 @@
+// @flow
+
+import { widget } from 'Dashboard'
```
damianpm: I think that naming `widget` this method is wrong. The name suggest you'll get a widget but it only makes an ajax request.
I know you didn't added it in this PR, but we should change this in the future.
Just commenting

app/javascript/src/Dashboard/ajax-widget.jsx
```diff
@@ -0,0 +1,34 @@
...
+
+export function loadAudienceWidget (widgetPath: string) {
+  widget(widgetPath)
```
damianpm: Maybe you could

`import { widget as makeWidgetRequest } from 'Dashboard'`

Or something similar

josemigallas: I like that 👍

PR #989, opened 2019-07-11
Adding validation for PF4 forms

damianpm:

Extract PF4 Form validation to a common file

https://issues.jboss.org/browse/THREESCALE-2993

### DIFF
--- a/app/javascript/src/LoginPage/loginForms/Login3scaleForm.jsx
+++ b/app/javascript/src/LoginPage/loginForms/Login3scaleForm.jsx
@@ -4,6 +4,7 @@ import React from 'react'
 import type { Node } from 'react'
 
 import {HiddenInputs, FormGroup} from 'LoginPage'
+import { validateFormFields } from 'utilities/formValidation'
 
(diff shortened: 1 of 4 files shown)


didierofrivia:
@josemigallas @damianpm Yes, I don't think that kind of validation is useful nor reusable. There are libs that can aid with simple checks (ie: password don't match, email address invalid, etc..)and we can rely on proper server-side validation to display more specific errors (i.e: user taken)

damianpm:
@josemigallas @didierofrivia I know this is not reusable. But we have an issue, we are using PF4 react components and I was trying to adjust to that. We can look for a library to do the validation, just keep in mind we should adjust to PF4 components and styling

josemigallas:
@damianpm I don't understand, we should adjust to PF4 components and styling in what sense?

damianpm:
In that sense that there are lot of `React forms` packages, but we only want validation

damianpm:
Closing in favor of https://github.com/3scale/porta/pull/1005

PR #990, opened 2019-07-11
Fixes Header on small screen setup

josemigallas:

After introducing PF4 React in vert nav, the header is showing some weird styles due to Patternfly variables.

In order to fix this, it is required to use `!important` in our `_patternfly_ie11.scss`.

Before:
header

After:
header_after

[THREESCALE-3013: Header changes color when on small screen setup](https://issues.jboss.org/browse/THREESCALE-3013)

**Verification steps** 
* Go to a screen where there's a Vertical Navigation (e.g. Audience)
* Shrink browser and look at the Header's nav component. Background and margins should stay unchanged.

### DIFF
--- a/app/assets/stylesheets/provider/_patternfly_ie11.scss
+++ b/app/assets/stylesheets/provider/_patternfly_ie11.scss
@@ -282,8 +282,8 @@ button:-moz-focusring,
   -ms-grid-row-span: 1;
   grid-row: 2 / 3;
   min-width: 0;
-  padding-left: 0;
-  background-color: transparent;
+  padding-left: 0 !important; // Need to override styles introduced by PF React
+  background-color: transparent !important; // Need to override styles introduced by PF React
 }
 
 .pf-c-page__header-nav > .pf-c-nav {


hallelujah:
@josemigallas  It would be nice to have before/after screenshot ;)

app/assets/stylesheets/provider/_patternfly_ie11.scss
```diff
@@ -282,8 +282,8 @@ button:-moz-focusring,
   -ms-grid-row-span: 1;
   grid-row: 2 / 3;
   min-width: 0;
-  padding-left: 0;
-  background-color: transparent;
+  padding-left: 0 !important; // Need to override styles introduced by PF React
```
hallelujah: Or loading this after PF React won't do?
Or using more precise css rules?

damianpm: @josemigallas This file is called `_patternfly_ie11.scss`. I'd expect is loaded only when using IE11 but is imported in `app/assets/stylesheets/provider/_theme.scss`. We shouldn't load this file if unnecessary.
This style is only needed in IE11?

josemigallas: @hallelujah It's more complex than that. The problem is using PF react and rails. Rails styles are load first in a big bundle, after that React loads its own styles and put them inline. This means there is no way to load this style after PF react (that would be the ideal solution).

It would be possible to add these styles by React, in a different styles.scss file but I think leaving it here in `_patternfly_ie11.scss` makes it more maintainable. In the end the file itself is a hacky workaround and it will be removed when we drop IE11 support.

Adding a more precise rule could be possible (I guess) but I think this way the intent is clearer and again, it is a hack already.

@damianpm It is not. These contains all PF4 styles necessary for the Page, Header and Vertical Navigation, that are also compatible with IE11. This was discussed and introduced in #859. And it's the official Patternfly way according to [this wiki](https://github.com/patternfly/patternfly-next/wiki/IE11-Support).

It could have been named `patternfly_ie11_compatible.scss` to avoid confusion.

damianpm: OK, then we should rename it.
What about importing this file from webpack, after DOMContentLoaded, have you tried?

josemigallas: Styles will be broken if JS is disabled then. This file has to be there..

damianpm: Well if js is disabled everything will be broken

josemigallas: Well that's not entirely accurate. The relevant components (Page, Header, Nav) won't be broken and neither will be most of the content generated by rails, an example where nothing is broken:
Screen Shot 2019-07-12 at 12 03 55

damianpm: There is no context selector :(

josemigallas: Obviously a big amount of features won't work without JS but that doesn't mean _everything_ is broken. Lacking functionality should be implemented accordingly to progressive enhancement, if they are really necessary. Having JS disabled is an edge case but if we are able to minimize the damage...

PR #998, opened 2019-07-15
[Bugfix] Retry faster so the event still exists. Otherwise, raise the proper error

Martouta:
Closes [THREESCALE-2975](https://issues.jboss.org/browse/THREESCALE-2975)

The bug is because when it retries, the event doesn't exist anymore.
What this PR does:
1. If the event doesn't exist, raise `ActiveRecord::RecordNotFound` instead of crashing in `NoMethodError` for `nil`.
2. As @hallelujah realised, it [automatically retries with exponential time](https://github.com/mperham/sidekiq/wiki/Error-Handling#automatic-job-retry), and [we retry up to 25 times](https://github.com/3scale/porta/commit/1622917a37c08633962d9bfdabd30f7481cf3f8d#diff-0fcadbb087d877eb135040bb90798b38), so it takes over a week and the event is destroyed by then. So this PR makes the retry to go 3 times faster to take up to 6 days and some hours 😄

### DIFF
--- a/app/workers/zync_worker.rb
+++ b/app/workers/zync_worker.rb
@@ -13,6 +13,15 @@ class ZyncWorker
 
   sidekiq_options queue: :zync
 
+  # It is used as the delay, in seconds.
+  # The default variable is the same as sidekiq's default, as you can see in https://github.com/mperham/sidekiq/wiki/Error-Handling#automatic-job-retry
+  # the default would last up 20 days and some hours, and we need it to last less than 7 days for the event to still exist,
+  # so it goes 3 times faster to last maximum 6 days and some hours (1/3 = 7/21)
+  sidekiq_retry_in do |count, _exception|
+    default = (count ** 4) + 15 + (rand(30) * (count + 1))
+    default / 3
+  end
+
   def self.config
     Rails.configuration.zync
   end
@@ -215,9 +224,10 @@ class ZyncWorker
 
   def perform(event_id, notification, manual_retry_count = nil)
     return false unless valid?(event_id)
+    event = EventStore::Repository.find_event!(event_id)
 
     with_manual_retry_count(event_id, manual_retry_count) do
-      tenant, provider = update_tenant(event_id)
+      tenant, provider = update_tenant(event)
 
       publish_notification = -> { http_put(notification_url, notification, event_id) }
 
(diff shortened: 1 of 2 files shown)


app/workers/zync_worker.rb
```diff
@@ -13,6 +13,11 @@ class ZyncWorker
 
   sidekiq_options queue: :zync
 
+  # It is used as the delay, in seconds.
+  sidekiq_retry_in do |count, _exception|
+    10 * (count + 1) # (i.e. 10, 20, 30, 40, 50)
```
hallelujah: This is in seconds :/
Which means the last one will be retried in:

10 x (1 + 2 + 3 + ... + 25) = 10x(25x26)/2  = 3250 seconds which is less than one hour.

So if there is an issue in Zync side then it has to be solved in this period of time otherwise the job will be lost

Martouta: oh, I see 😞 Ok, I will change it 👍

app/workers/zync_worker.rb
```diff
@@ -13,6 +13,15 @@ class ZyncWorker
 
   sidekiq_options queue: :zync
 
+  # It is used as the delay, in seconds.
+  # The default variable is the same as sidekiq's default, as you can see in https://github.com/mperham/sidekiq/wiki/Error-Handling#automatic-job-retry
+  # the default would last up 20 days and some hours, and we need it to last less than 7 days for the event to still exist,
+  # so it goes 3 times faster to last maximum 6 days and some hours (1/3 = 7/21)
+  sidekiq_retry_in do |count, _exception|
```
guicassolato: Just a reminder that this has no effect over the `UnprocessableEntityError` errors now that we are not really raising them anymore. Those are retried immediately after the batch of dependencies completes.

app/workers/zync_worker.rb
```diff
@@ -13,6 +13,15 @@ class ZyncWorker
...
+    default = (count ** 4) + 15 + (rand(30) * (count + 1))
+    default / 3
```
guicassolato: Remember that this will truncate to the integer part of the quotient.

app/workers/zync_worker.rb
Martouta: That's a good thing, but what this PR means to fix is:
We have a spike in production that the retries are failing because the event doesn't exist anymore... that happens because it retries up to 25 times in a bit over 20 days.
The retry can be initially for any reason, even that Zync is down 😄 
So what this does is just splitting those retries to be 3 times faster so the event will still exist (and 1 week should be enough 😑 )

app/workers/zync_worker.rb
Martouta: Yeah, that's fine 😄

PR #999, opened 2019-07-15
[wip] npm dependencies check

didierofrivia:

We were having problems in the past with the npm dependencies in our different environments. This PR aims to solve that using `npm ci` in our CI process and check if all of them are installed correctly.

Do the test described in https://issues.jboss.org/browse/THREESCALE-2183


.circleci/config.yml
```diff
@@ -478,7 +478,8 @@ jobs:
           name: Install NPM dependencies
           command: |
             npm --version
-            npm install
+            npm ci
+            bundle exec rake npm:check
```
josemigallas: Shouldn't we run `npm ls` before `npm install`? And only install in case there are missing packages?

didierofrivia: `npm ci` removes the node_modules directory, so nothing is "installed". Thus, after ci we check the dependencies were installed correctly... does it makes sense?

josemigallas: What I mean is, instead of always running `npm ci`, we should first check if we're missing any dependencies, no? And only install when there are. This is at least what @hallelujah has been suggesting and I thought that's what we were doing.

**Besides that** and according to your comment, why do you think it is necessary to check for missing deps after `npm ci`? The expected behavior is to install everything, maybe it would make more sense to check for output errors from `npm ci`. Am I missing something here?

didierofrivia: True, it was talked before that better to check first before install. Will change the task now

hallelujah: Sadly `npm ci` does not install all dependencies so `npm ls || npm ci` will always run `npm ci` at the end?

didierofrivia: yeah, that's true... sadly it will ignore the dev deps... so it will fail... 
Will try to do a different approach..

josemigallas: The current behavior, theoretically, is that circle ci will only run the NPM task if `package-lock.json` has changed. Otherwise it loads node_modules from the cache.

Isn't that enough?

didierofrivia: What it was happening, is that sometimes it failed to install all dependencies correctly.. or the cached ones were somehow corrupted. That's why we needed the `npm ls`- you could see the issue attached in the description.
So, no, it's not enough. We need to make sure the dependencies are all there and healthy.

josemigallas: OK sorry I didn't read the issue first. I left you a comment in the JIRA, if the issue is missing peer dependencies then it has nothing to do with `npm ci` or `npm ls`.

didierofrivia: exactly, it's because the local `npm install` and endless package-lock.json modifications. The error generates in dev environment where the only way to spot the missing deps are in the possible warnings of the `npm install` output

package.json
```diff
@@ -7,7 +7,8 @@
   "license": "Apache-2.0",
   "scripts": {
     "clean": "rm -rf node_modules/ && npm cache clean --force",
-    "update-lock": "npm run clean && npm i --no-progress",
+    "update-dependencies": "npm run clean && npm i --no-progress",
+    "check-dependencies": "npm ls || npm run update-dependencies",
```
josemigallas: Maybe add `npm ls --depth=0` to reduce verbosity?

hallelujah: Does that mean it will check everything? Or just the direct dependencies and not the peer dependencies?
Can't we use `--silent` instead?

```
 npm ls --loglevel silent >/dev/null
```

package.json
```diff
@@ -7,7 +7,8 @@
   "license": "Apache-2.0",
   "scripts": {
     "clean": "rm -rf node_modules/ && npm cache clean --force",
-    "update-lock": "npm run clean && npm i --no-progress",
+    "update-dependencies": "npm run clean && npm i --no-progress",
+    "check-dependencies": "npm ls --depth=0 || npm run update-dependencies",
```
hallelujah:
```suggestion
    "check-dependencies": "npm ls --loglevel silent > /dev/null || npm run update-dependencies",
```

package.json
```diff
@@ -7,7 +7,8 @@
   "license": "Apache-2.0",
   "scripts": {
     "clean": "rm -rf node_modules/ && npm cache clean --force",
-    "update-lock": "npm run clean && npm i --no-progress",
+    "update-dependencies": "npm run clean && npm i --no-progress",
```
hallelujah: That will update the `package-lock.json` no?
So it is not a reproducible build on CI

didierofrivia: That won't update it as long as we keep it up-to-date. Sadly, there's no other way to do it in CI, since we are using dev dependencies to build test assets. The only way would be having locked versions in our package.json, and that would lead to a PITA when we want to upgrade.

package.json
didierofrivia: Thing is, I'd like to see the dependencies that were having troubles (in case it fails) in the logs, since we might not be able to reproduce it locally. So better to have it with depth 0.

package.json
hallelujah: Well the CI will update it right? The cache can be invalidated for many reasons, so it will update the package-lock.json

Hopefully peer dependencies will not mess up :)

package.json
hallelujah: But depth=0 is only checking for our dependencies not the peers ones ...

didierofrivia: Don't worry, the `--depth` will just modify the visual output, not the analysis itself ;)

package.json
didierofrivia: yeah, as I said, the only way to do it for our CI, is locking the version of the dependencies in package.json, since npm doesn't provide a method to install the exact same tree described in the lock (ci is lacking the dev dependencies)

PR #1001, opened 2019-07-16
Yet another update of the package-lock

josemigallas:
**Description**
I've detected a huge, huge amount of changes in the package-lock in master, this time versions.

There's also `npm@6.10.1` now.

**Steps to reproduce**
* run `npm run clean` (new task)
* run `npm i`
* see changes in package-lock.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,8 @@
   },
   "license": "Apache-2.0",
   "scripts": {
-    "clear": "rm -rf node_modules/",
+    "clean": "rm -rf node_modules/ && npm cache clean --force",
+    "update-lock": "npm run clean && npm i --no-progress",
     "flow:install": "flow-typed install",
     "lint": "flow && eslint .eslintrc app/javascript spec/javascripts",
     "test": "karma start --single-run",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -6,7 +6,7 @@
   },
   "license": "Apache-2.0",
   "scripts": {
-    "clear": "rm -rf node_modules/",
+    "clean": "rm -rf node_modules/ && rm package-lock.json && npm cache clean --force",
```
didierofrivia: The problem with deleting the lock is that we always will have changes (over time) because of our non-locked deps versions. I'd rather explicitly update them (manually or by dependantbot) than the corollary of executing this task and then `npm i`. However, we might want to have a task for automating this, then I'd name it differently:

* Removing modules and clearing the cache: `flush`, `clear`, `clean`
* Removing package-lock and running install: `update-lock`, `topup`

Bottom line: I would avoid having a task that will either just remove the lock or it will remove it with other operations that might create changes implicitly. I'll have a task with the sole purpose of what we get after deleting the lock and running the install clearly.

josemigallas: OK that makes perfect sense to me.

PR #1003, opened 2019-07-16
[apiap] Backend API model

hallelujah:
So it works with current code

Just wip to see the concept

### DIFF
--- a/app/lib/backend_api_logic/service_extension.rb
+++ b/app/lib/backend_api_logic/service_extension.rb
@@ -24,7 +24,7 @@ module BackendApiLogic
     private
 
     def create_first_backend_api_config!
-      backend_api = account.backend_apis.create!(
+      backend_api = account.backend_apis.build(
         system_name: system_name,
         name: "#{name} Backend API",
         description: "Backend API of #{name}",
--- a/app/lib/backend_api_logic/service_extension.rb
+++ b/app/lib/backend_api_logic/service_extension.rb
@@ -24,14 +24,15 @@ module BackendApiLogic
     private
 
     def create_first_backend_api_config!
-      return unless proxy && account
       backend_api = account.backend_apis.create!(
         system_name: system_name,
         name: "#{name} Backend API",
         description: "Backend API of #{name}",
-        private_endpoint: proxy['api_backend'].presence
+        private_endpoint: proxy['api_backend']
       )
-      backend_api_configs.create!(backend_api: backend_api, path: '')
+      constructor = new_record? ? :build : :create!
+      attrs = {backend_api: backend_api, path: ''}
+      backend_api_configs.public_send(constructor,attrs)
     end
   end
 end
--- a/app/lib/backend_api_logic/service_extension.rb
+++ b/app/lib/backend_api_logic/service_extension.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+module BackendApiLogic
+  module ServiceExtension
+    extend ActiveSupport::Concern
+
+    included do
+      has_many :backend_api_configs, inverse_of: :service, dependent: :destroy
+      has_many :backend_apis, through: :backend_api_configs, dependent: :destroy
+
+      alias_method :backend_api, :first_backend_api
+    end
+
+    def first_backend_api
+      @backend_api ||= find_or_create_first_backend_api!
+    end
+
+    def find_or_create_first_backend_api!
+      return unless proxy && account
+      config = backend_api_configs.first || create_first_backend_api_config!
+      config.backend_api
+    end
+
+    private
+
+    def create_first_backend_api_config!
+      return unless proxy && account
+      backend_api = account.backend_apis.create!(
+        system_name: system_name,
+        name: "#{name} Backend API",
+        description: "Backend API of #{name}",
+        private_endpoint: proxy['api_backend'].presence
+      )
+      backend_api_configs.create!(backend_api: backend_api, path: '')
+    end
+  end
+end
--- a/app/lib/backend_api_logic/service_extension.rb
+++ b/app/lib/backend_api_logic/service_extension.rb
@@ -16,7 +16,7 @@ module BackendApiLogic
     end
 
     def find_or_create_first_backend_api!
-      return unless proxy && account
+      return unless account
       config = backend_api_configs.first || create_first_backend_api_config!
       config.backend_api
     end
@@ -28,11 +28,11 @@ module BackendApiLogic
         system_name: system_name,
         name: "#{name} Backend API",
         description: "Backend API of #{name}",
-        private_endpoint: proxy['api_backend']
+        private_endpoint: proxy&.[]('api_backend')
       )
       constructor = new_record? ? :build : :create!
       attrs = {backend_api: backend_api, path: ''}
-      backend_api_configs.public_send(constructor,attrs)
+      backend_api_configs.public_send(constructor, attrs)
     end
   end
 end
--- a/app/validators/path_validator.rb
+++ b/app/validators/path_validator.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+class PathValidator < ActiveModel::EachValidator
+  def validate_each(record, attribute, value)
+    record.errors.add(attribute, :not_a_path) unless path?(value)
+  end
+
+  protected
+
+  def path?(value)
+    uri = URI.parse(value)
+    uri.path == value
+  rescue
+    false
+  end
+end
--- a/app/validators/path_validator.rb
+++ b/app/validators/path_validator.rb
@@ -10,7 +10,7 @@ class PathValidator < ActiveModel::EachValidator
   def path?(value)
     uri = URI.parse(value)
     uri.path == value
-  rescue
+  rescue URI::InvalidURIError
     false
   end
 end
--- a/app/validators/path_validator.rb
+++ b/app/validators/path_validator.rb
@@ -2,7 +2,7 @@
 
 class PathValidator < ActiveModel::EachValidator
   def validate_each(record, attribute, value)
-    record.errors.add(attribute, :not_a_path) unless path?(value)
+    record.errors.add(attribute, :not_path_format) unless path?(value)
   end
 
   protected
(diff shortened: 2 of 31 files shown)


app/models/backend_api_slug.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+class BackendApiSlug < ApplicationRecord
```
guicassolato: Hmm. We may want to later grow this to more than just a "slug", incorporating other attributes that belong to the association between `Service` and `BackendApi`. Also, we agreed that one backend API (and no more than one) used in a service is allowed to have an empty slug. So I think we should rename this model to something else.

hallelujah: Right for the naming I do not know what to put.
Slug would never be empty but will be `"/"`

hallelujah: I think `slug` should be `path` instead 
And we can use `BackendApiConfig` as name, what do you think @guicassolato ?

guicassolato: `BackendApiConfig` 👍

db/migrate/20190715144518_create_backend_apis.rb
```diff
@@ -0,0 +1,30 @@
+class CreateBackendApis < ActiveRecord::Migration
+  disable_ddl_transaction!
+
+  def change
+    create_table :backend_apis do |t|
```
Martouta:
```sql
CREATE TABLE `backend_apis` (`id` BIGINT(20) auto_increment PRIMARY KEY, `name` varchar(255) NOT NULL, `system_name` varchar(255) NOT NULL, `description` text, `private_endpoint` varchar(255), `account_id` bigint, `created_at` datetime NOT NULL, `updated_at` datetime NOT NULL, `tenant_id` bigint, UNIQUE INDEX `index_backend_apis_on_account_id_and_system_name`  (`account_id`, `system_name`) ) ENGINE=InnoDB
```

db/migrate/20190715144518_create_backend_apis.rb
```diff
@@ -0,0 +1,30 @@
...
+      t.string :private_endpoint
+
+      t.belongs_to :account, limit: 8
+
+      t.timestamps null: false
+      t.integer :tenant_id, limit: 8
+      t.index [:account_id, :system_name], unique: true
+    end
+
+    create_table :backend_api_configs do |t|
```
Martouta:
```sql
CREATE TABLE `backend_api_configs` (`id` BIGINT(20) auto_increment PRIMARY KEY, `path` varchar(255) DEFAULT '' NOT NULL, `service_id` bigint, `backend_api_id` bigint, `created_at` datetime NOT NULL, `updated_at` datetime NOT NULL, `tenant_id` bigint, UNIQUE INDEX `index_backend_api_configs_on_path_and_service_id`  (`path`, `service_id`) ) ENGINE=InnoDB
```

db/migrate/20190716110520_create_the_backend_apis_of_services.rb
```diff
@@ -0,0 +1,22 @@
+class CreateTheBackendApisOfServices < ActiveRecord::Migration
+  def change
+    say_with_time 'Migrating proxies api_backend to slugs...' do
+      reversible do |dir|
+        dir.up do
+          Service.transaction do
+            Service.find_each do |service|
```
Martouta:
```sql
SELECT  `services`.* FROM `services`  ORDER BY `services`.`id` ASC LIMIT 1000
```

db/migrate/20190716110520_create_the_backend_apis_of_services.rb
```diff
@@ -0,0 +1,22 @@
...
+              service.find_or_create_first_backend_api!
```
Martouta: It is inside a loop so I will add just one of them, with Service ID 1 and Account ID 1 (its provider):
```sql
SELECT  * FROM accounts WHERE id = 1 LIMIT 1
SELECT  * FROM backend_api_configs WHERE service_id = 1  ORDER BY id ASC LIMIT 1
SELECT  * FROM proxies WHERE service_id = 1 LIMIT 1
BEGIN
  SELECT  1 AS one FROM backend_apis WHERE (system_name = 'master_service' AND account_id = 1) LIMIT 1
  INSERT INTO `backend_apis` (`system_name`, `name`, `description`, `private_endpoint`, `account_id`, `created_at`, `updated_at`) VALUES ('master_service', 'Master Service Backend API', 'Backend API of Master Service', 'https://echo-api.3scale.net:443', 1, 'YYYY-MM-DD HH:MM:SS', 'YYYY-MM-DD HH:MM:SS')
COMMIT
BEGIN
  INSERT INTO `backend_api_configs` (`backend_api_id`, `service_id`, `created_at`, `updated_at`) VALUES (1, 1, 'YYYY-MM-DD HH:MM:SS', 'YYYY-MM-DD HH:MM:SS')
COMMIT
```

Martouta:
I've just rebased from master and fixed conflicts 👍

app/validators/path_validator.rb
```diff
@@ -0,0 +1,16 @@
...
+  def validate_each(record, attribute, value)
+    record.errors.add(attribute, :not_path_format) unless path?(value)
+  end
+
+  protected
+
+  def path?(value)
+    uri = URI.parse(value)
+    uri.path == value
+  rescue
```
Martouta: Why not `rescue URI::InvalidURIError` ?

db/migrate/20190715144518_create_backend_apis.rb
```diff
@@ -18,7 +18,7 @@ def up
     end
 
     create_table :backend_api_configs do |table|
-      table.string :path, null: false, default: ''
+      table.string :path, default: ''
```
Martouta: It is not ideal but otherwise it fails for Oracle 😑 
image

Because for some reason it tries to enter `null` 😕 And this way, if we don't bypass the rails validation, it is correct and inserts `''`.
Gui showed me that the `path` is not inside `changed_attributes`, probably because `''` is the default value, but I don't understand that the same thing happens for MySQL and it works.


And with this change, running  for postgres `bundle exec rake db:drop db:create db:setup` with the same `schema` as in `master` with the idea of running the migrations after it:
image

db/migrate/20190715144518_create_backend_apis.rb
```diff
@@ -0,0 +1,46 @@
...
+    create_table :backend_apis do |table|
+      table.string :name, null: false, limit: 511
+      table.string :system_name, unique: true, null: false
+      table.text :description
+      table.string :private_endpoint
+
+      table.belongs_to :account, limit: 8
+
+      table.timestamps null: false
+      table.integer :tenant_id, limit: 8
```
guicassolato: We're missing the table in https://github.com/3scale/porta/tree/master/lib/system/database/definitions

db/migrate/20190715144518_create_backend_apis.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+
+    create_table :backend_api_configs do |table|
+      table.string :path, default: ''
+
+      table.belongs_to :service, limit: 8
+      table.belongs_to :backend_api, limit: 8
+
+      table.timestamps null: false
+      table.integer :tenant_id, limit: 8
```
guicassolato: Same as https://github.com/3scale/porta/pull/1003/files#r308788591

app/lib/backend_api_logic/service_extension.rb
```diff
@@ -0,0 +1,38 @@
...
+    private
+
+    def create_first_backend_api_config!
+      backend_api = account.backend_apis.create!(
+        system_name: system_name,
+        name: "#{name} Backend API",
+        description: "Backend API of #{name}",
+        private_endpoint: proxy&.[]('api_backend')
+      )
+      constructor = new_record? ? :build : :create!
```
Martouta: This is very subjective and not a big deal but to me these last 3 lines would be clearer like this:
```ruby
config = backend_api_configs.build(backend_api: backend_api, path: '')
config.save! unless new_record?
```

guicassolato: The difference is what `create_first_backend_api_config!` would return, so if you really want to change it, it should be
```ruby
def create_first_backend_api_config!
  backend_api = account.backend_apis.create!(
    system_name: system_name,
    name: "#{name} Backend API",
    description: "Backend API of #{name}",
    private_endpoint: proxy&.[]('api_backend')
  )
  config = backend_api_configs.build(backend_api: backend_api, path: '')
  config.save! unless new_record?
  config
end
```

app/lib/backend_api_logic/service_extension.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    def find_or_create_first_backend_api!
+      return unless account
+      config = backend_api_configs.first || create_first_backend_api_config!
+      config.backend_api
+    end
+
+    private
+
+    def create_first_backend_api_config!
```
Martouta: I think that this does not have a test and that it dangerous 😞 
image

Martouta: wait, maybe it is tested in here:
```
test/unit/services/service_discovery/import_cluster_definitions_service_test.rb
96:      @service.backend_api.stubs(account: @account)
```
I will see 👍

guicassolato: `create_first_backend_api_config!` is private, but `find_or_create_first_backend_api!` should have a test for sure!

Martouta: ok...
image

so this means that there is at least 1 test that relies on this somehow:
```ruby
test 'unsupported backend base path' do
      @import_service.stubs(cluster_service: @cluster_service)

      @account.stubs(:provider_can_use?).with(:apicast_v1).returns(true)
      @account.stubs(:provider_can_use?).with(:apicast_v2).returns(true)
      @account.stubs(:provider_can_use?).with(:proxy_private_base_path).returns(false)
      @service.backend_api.stubs(account: @account)

      System::ErrorReporting.expects(:report_error).with(responds_with(:message, 'Could not save API backend URL'), any_parameters)
      @import_service.send(:import_cluster_service_endpoint_to, @service)
      refute @service.proxy.valid?
      assert_equal 'http://api.example.net:80', @service.proxy.reload.api_backend
    end
```

But imo we still need one test that tests this in isolation and a bit more in depth 😄

app/lib/backend_api_logic/service_extension.rb
Martouta: right 👍

PR #1004, opened 2019-07-16
Revert THREESCALE-2808-PF4-React-Navbar-Components

josemigallas:
This reverts commit 9b628aca43f3751ba7173050788055f9a1e553cd, reversing
changes made to b87d3d3e3c5c87970534f3ced174e48ff645390a.

It looks like styles from `@patternfly/react-core` break our current ones. They are included inline and although the only component being used is Nav, the whole page is affected.

There is no optimal solution for this. The only one is to import (again) our styles inside the React component so that they are include inline **but after** those included by PFs. And it's a terrible solution.

Instead of duplicating the import we might simply do it once in React however the whole page would break if JS happened to be disabled.

Without PF React it always look good, including when JS is disabled and in IE11. Hence my suggestion is to simply reverse this change.


damianpm:
@josemigallas In Safari I still see underlined links:
Screenshot 2019-07-16 at 16 57 32

josemigallas:
@damianpm That is out of the scope of this PR, since it is a reverse commit. It shouldn't contain any other changes.

However I made a different PR for it: https://github.com/3scale/porta/pull/1006

PR #1005, opened 2019-07-16
Adding validatejs to validate forms

damianpm:

Adds form validation to PF4/React forms using [Validate.js](https://github.com/ansman/validate.js)

https://issues.jboss.org/browse/THREESCALE-2993

**Pros and cons of Validate.js**

**Pros:**
- Small (5.05KB, minified and gzipped)
- Non-intrusive (It doesn't force any styling)
- It may validate a full form or single fields
- Active project and community
- MIT License

**Cons:**
- Depends on jQuery (but we already have jQuery, so maybe not a con)

----
validatejs

**Demo:**

form-validation


app/javascript/src/LoginPage/utils/formValidation.js
```diff
@@ -0,0 +1,96 @@
...
+    length: {minimum: 1}
+  },
+  email: {
+    presence: true,
+    email: true,
+    length: {minimum: 1}
+  },
+  password: {
+    presence: true,
+    length: {minimum: 1}
```
josemigallas: Maybe minimum 6 like in other forms?

app/javascript/src/LoginPage/loginForms/Login3scaleForm.jsx
```diff
@@ -33,50 +39,40 @@ class Login3scaleForm extends React.Component<Props, State> {
...
 
-  validateForm = (event: SyntheticEvent<HTMLInputElement>) => {
-    this.setIsValidUsername()
-    this.setIsValidPassword()
-    const isFormDisabled = this.state.username === '' ||
-      this.state.password === '' ||
-      this.state.isValidUsername === false ||
-      this.state.isValidPassword === false
-    if (isFormDisabled) {
+  validateForm = (event: SyntheticEvent<HTMLButtonElement>) => {
```
josemigallas: I think this is unnecessary, we validate already on each input change.

Also, the submit button should be disabled whenever there's an error.

app/javascript/src/LoginPage/loginForms/Login3scaleForm.jsx
```diff
@@ -33,50 +39,40 @@ class Login3scaleForm extends React.Component<Props, State> {
...
-  setIsValidPassword = () => {
-    this.setState({isValidPassword: this.state.password !== ''})
-  }
-
-  handleTextInputPassword = (password: string) => {
-    this.setState({ password }, this.setIsValidPassword)
+  handleInputChange = (value: string, event: SyntheticEvent<HTMLInputElement>) => {
+    const isValid = validateSingleField(event)
+    this.setState({
+      [namesToStateKeys[event.currentTarget.name].name]: value,
```
josemigallas: Could we get rid of `nameToStateKeys`? It's a bit convoluted.. and it's not used everywhere like I think it should be, for instance in the Type definition the state keys are hardcoded and also in the constructor.

If not, why not use `value` instead of `event.currentTarget.name`?

app/javascript/src/LoginPage/loginForms/Login3scaleForm.jsx
```diff
@@ -33,50 +39,40 @@ class Login3scaleForm extends React.Component<Props, State> {
...
-      this.state.password === '' ||
-      this.state.isValidUsername === false ||
-      this.state.isValidPassword === false
-    if (isFormDisabled) {
+  validateForm = (event: SyntheticEvent<HTMLButtonElement>) => {
+    const errors = validateAllFields(event.currentTarget.form)
+    if (errors) {
       event.preventDefault()
+      errors.forEach(
+        (error) => this.setState({[namesToStateKeys[error].isValid]: false})
```
josemigallas: I think errors is misnamed here. `error` is used like a name but it should be an error message. Maybe rename it to `invalidFields`? Or do `error => error.name ...`

spec/javascripts/LoginPage/Login3scaleForm.spec.jsx
```diff
@@ -28,17 +28,31 @@ it('should render HiddenInputs component', () => {
...
+    }
+  }
   const wrapper = mount(<Login3scaleForm {...props}/>)
-  jest.spyOn(wrapper.instance(), 'handleTextInputUsername')
-  wrapper.instance().handleTextInputUsername('foo')
-  expect(wrapper.instance().handleTextInputUsername).toHaveBeenCalled()
-  expect(wrapper.state().username).toEqual('foo')
+  jest.spyOn(wrapper.instance(), 'handleInputChange')
+  wrapper.instance().handleInputChange('Bob', event)
+  expect(wrapper.instance().handleInputChange).toHaveBeenCalled()
```
josemigallas: It's been called literally in the previous line. Instead of that, find the input and simulate a change there, then asset the spy.

app/javascript/src/LoginPage/loginForms/Login3scaleForm.jsx
```diff
@@ -33,31 +47,26 @@ class Login3scaleForm extends React.Component<Props, State> {
...
-      this.state.password === '' ||
-      this.state.isValidUsername === false ||
-      this.state.isValidPassword === false
-    if (isFormDisabled) {
+    const errors = validate(event.currentTarget.form, constraints)
+    if (errors) {
       event.preventDefault()
+      //$FlowFixMe: Needed due to a flow issue with Object values/keys: https://github.com/facebook/flow/issues/2221
+      for (const errorId in errors) {
+        this.setState({[namesToStateKeys[errorId]]: false})
```
didierofrivia: This might not be needed if you disable de form submit by default. You could also define your state like : 
```
{
username: '',
password: '',
validation: {
  username: undefined,
  password: undefined,
}
}
```
you'll be avoiding the declaration of https://github.com/3scale/porta/pull/1005/commits/36671e33b10614952825ac106c1e26e43af93773#diff-119386a465d8704baf0fcb20a45ed705R34

Then the check you can do is an `Array.every` iterating the validation keys and checking the values are `true`, then you avoid defining an extra step for checking the constrains.

What do you think?

josemigallas: I think having validations in a object is 👌

thomasmaas: disable by default through javascript, right? (otherwise no js will have no way to enable the form, i know hypothetical)

app/javascript/src/LoginPage/loginForms/Login3scaleForm.jsx
damianpm: But the user may press enter and submit the form, even by accident.

app/javascript/src/LoginPage/utils/formValidation.js
damianpm: In fact I don't know what is the minimum length for this specific form, because of that I leave it with just one.

app/javascript/src/LoginPage/loginForms/Login3scaleForm.jsx
damianpm: @thomasmaas are you OK with disabling the form by default?
Personally, I don't like it, I prefer disabling after the user has clicked on submit, because I think is good to give the user time to fill the form

josemigallas: IMHO having validation on submit is the most annoying thing in a form 🤣 I usually want to know what's wrong the earliest possible so that I don't waste time clicking submit.

app/javascript/src/LoginPage/loginForms/Login3scaleForm.jsx
josemigallas: What do you mean? Even if enter is pressed, the form will be validated already, there's no need to validate all fields again.

app/javascript/src/LoginPage/loginForms/RequestPassWordForm.jsx
```diff
@@ -17,56 +22,53 @@ type Props = {
...
 class RequestPasswordForm extends React.Component<Props, State> {
-  constructor (props: Props) {
-    super(props)
-    this.state = {
-      emailAddress: '',
-      isValidEmail: undefined
+  state = {
+    email: '',
+    validation: {
+      email: undefined
```
josemigallas: No need to initialize as `undefined`, it can be simply `validation: {}`. On the other hand, I think it would be good to initialize all validations to `false` (not only in this form but all of them).

app/javascript/src/LoginPage/loginForms/Login3scaleForm.jsx
damianpm: I mean if the user clicks on enter without filling the form. The validation happens only when you fill the form

app/javascript/src/LoginPage/loginForms/RequestPassWordForm.jsx
damianpm: I disagree, if we initialise all fields to false, the form will display error messages by default.  Like this:
errors

IMO error messages should be shown only if the user has done an actual error, not before.
WDYT @thomasmaas  ?

app/javascript/src/LoginPage/loginForms/SignupForm.jsx
```diff
@@ -22,17 +24,36 @@ type Props = {
...
-  isFilledUsername: ?boolean,
-  isFilledEmailAddress: ?boolean,
-  isFilledFirstname: boolean,
-  isFilledLastname: boolean,
-  isFilledPassword: ?boolean,
-  isFilledPasswordConfirmation: ?boolean
+  validation: {
+    username: ?boolean,
+    email: ?boolean,
+    firstName: boolean,
```
josemigallas: Why `firstName` and `lastName` are not _maybe_ types?

app/javascript/src/LoginPage/loginForms/SignupForm.jsx
```diff
@@ -53,68 +74,75 @@ const InputFormGroup = (props) => {
...
+    username: this.props.user.username,
+    email: this.props.user.email,
+    firstName: this.props.user.firstname,
+    lastName: this.props.user.lastname,
+    password: '',
+    passwordConfirmation: '',
+    validation: {
+      username: undefined,
+      email: undefined,
+      firstName: true,
```
josemigallas: I don't think this is correct, validation should be `undefined` just like the rest, unless the field is filled in by default?

app/javascript/src/LoginPage/loginForms/SignupForm.jsx
```diff
@@ -53,68 +74,75 @@ const InputFormGroup = (props) => {
...
+      validation
     })
+  }
 
-  handleTextInputPasswordConfirmation = (passwordConfirmation: string) =>
-    this.setState({
-      passwordConfirmation,
-      isFilledPasswordConfirmation: passwordConfirmation !== '' && passwordConfirmation === this.state.password
-    })
+  renameValidation = (namesToStateKeys: NamesToKeys, invalidFields: InvalidFields) => Object.keys(invalidFields)
```
josemigallas: What is this function for? What does it mean?

josemigallas: I think `invalidFields` and State's `validation` are exactly the same kind of map, they should share the same type.

app/javascript/src/LoginPage/loginForms/RequestPassWordForm.jsx
thomasmaas: Yes of course. Showing a form with error messages before user interaction would be super aggressive.

app/javascript/src/LoginPage/utils/formValidation.js
josemigallas: Can be 6 or 8, depending on the provider:
https://github.com/3scale/porta/blob/master/app%2Fmodels%2Fuser.rb#L80
https://github.com/3scale/porta/blob/master/app%2Fmodels%2Fuser.rb#L105-L106

I suggested using 6 as it is the "standard".

thomasmaas: Just check what our current minimum is at 3scale.net and stick with it. It should be exactly the same as the Ruby validations

app/javascript/src/LoginPage/utils/formValidation.js
```diff
@@ -0,0 +1,62 @@
...
+
+const validateAllFields = (form) => {
+  const errors = validate(form, constraints[form.id])
+
+  const fieldsWithErrors = Object.keys(errors).reduce((obj, item) => {
+    obj[item] = false
+    return obj
+  }, {})
+
+  return errors ? fieldsWithErrors : undefined
```
josemigallas: We could check for errors before the iteration.

spec/javascripts/LoginPage/Login3scaleForm.spec.jsx
```diff
@@ -27,18 +27,62 @@ it('should render HiddenInputs component', () => {
...
+describe('username', () => {
+  it('should set username and validation state to true', () => {
+    const event = {
+      currentTarget: {
+        name: 'username',
+        value: 'Bob',
+        type: 'text'
+      }
+    }
+    const wrapper = mount(<Login3scaleForm {...props}/>)
```
josemigallas: Would it be possible to use `shallow` instead? Using `mount` is a bit overkilling.

spec/javascripts/LoginPage/Login3scaleForm.spec.jsx
```diff
@@ -27,18 +27,62 @@ it('should render HiddenInputs component', () => {
...
+      currentTarget: {
+        name: 'username',
+        value: '',
+        type: 'text'
+      }
+    }
+    const wrapper = mount(<Login3scaleForm {...props}/>)
+    wrapper.find('input#session_username').props().onChange(event)
+    expect(wrapper.state().username).toEqual('')
+    expect(wrapper.state('validation').username).toEqual(false)
```
josemigallas: I know it's not a big deal but let's try to be consistent, use same kind of accessor:
* `state().username`, `state().validation.username`
* `state('username')`, `state('validation').username`

app/javascript/src/LoginPage/utils/formValidation.js
thomasmaas: ah sorry it depends on settings, so should ideally use those settings to provide the correct feedback.

app/javascript/src/LoginPage/loginForms/SignupForm.jsx
damianpm: This is needed because validate.js uses the name attribute to access the inputs.
In the other pages names are in the format `password`, or `email`, but in this page the name attrs are in the format `user[email]` , `user[password]`, etc.
For this reason I'm keeping the `namesToStateKeys` object only in this form.
That attributes comes from rails, we can't change them in the scope of this task

damianpm: 👍

spec/javascripts/LoginPage/Login3scaleForm.spec.jsx
damianpm: Nope, because all `wrapper.find(input)` fails

app/javascript/src/LoginPage/utils/formValidation.js
```diff
@@ -0,0 +1,64 @@
...
+    length: {minimum: 1}
+  },
+  email: {
+    presence: true,
+    email: true,
+    length: {minimum: 1}
+  },
+  password: {
+    presence: true,
+    length: {minimum: 6}
```
damianpm: Confirmed, minimum password length is 6.
Additionally, it will be needed to show error messages from server validation (e.g, password too short, username already token, etc.), but I'll make a new PR for that, because that implies updating some fields in the UI in all forms

damianpm:
Error messages from server were missing in Signup Page, I added them in last commit:

errors-list

app/javascript/src/LoginPage/utils/formValidation.js
```diff
@@ -0,0 +1,64 @@
...
+    'user[password]': constraintsTypes.password,
+    'user[password_confirmation]': constraintsTypes.password_confirmation
+  }
+}
+
+const validateAllFields = (form) => {
+  const errors = validate(form, constraints[form.id])
+
+  if (!errors) {
+    return undefined
```
didierofrivia: Avoid this construction, it's evil. Try to return the same type, and never undefined.

app/javascript/src/LoginPage/utils/formValidation.js
```diff
@@ -0,0 +1,64 @@
...
+  },
+  'signup_form': {
+    'user[username]': constraintsTypes.text,
+    'user[email]': constraintsTypes.email,
+    'user[password]': constraintsTypes.password,
+    'user[password_confirmation]': constraintsTypes.password_confirmation
+  }
+}
+
+const validateAllFields = (form) => {
```
didierofrivia: missing type

app/javascript/src/LoginPage/utils/formValidation.js
```diff
@@ -0,0 +1,64 @@
...
+    return undefined
+  } else {
+    return Object.keys(errors).reduce((obj, item) => {
+      obj[item] = false
+      return obj
+    }, {})
+  }
+}
+
+const validateSingleField = (event) => {
```
didierofrivia: here as well missing types.

app/javascript/src/LoginPage/utils/formValidation.js
```diff
@@ -0,0 +1,64 @@
...
+    'user[password_confirmation]': constraintsTypes.password_confirmation
+  }
+}
+
+const validateAllFields = (form) => {
+  const errors = validate(form, constraints[form.id])
+
+  if (!errors) {
+    return undefined
+  } else {
```
didierofrivia: You've defined a type for Errors, use it treat it accordingly... like an empty Array of Error is cool, if not, show them.

damianpm:
> Some refactoring needed: Functions should return always the same, there are fn missing types and the construction of those should be guided by the types defined.

Refactor done @didierofrivia !

app/javascript/src/LoginPage/loginForms/Login3scaleForm.jsx
```diff
@@ -51,7 +51,7 @@ class Login3scaleForm extends React.Component<Props, State> {
   validateForm = (event: SyntheticEvent<HTMLButtonElement>) => {
     const invalidFields = validateAllFields(event.currentTarget.form)
 
-    if (invalidFields) {
+    if (Object.keys(invalidFields).length > 0) {
```
didierofrivia: `if (Object.keys(invalidFields).length) {`

app/javascript/src/LoginPage/utils/formValidation.js
```diff
@@ -39,20 +41,22 @@ const constraints = {
   }
 }
 
-const validateAllFields = (form) => {
+const validateAllFields = (form: HTMLFormElement) => {
   const errors = validate(form, constraints[form.id])
```
didierofrivia: What does validate returns if not Object?

damianpm: Returns `undefined` when no errors found

app/javascript/src/LoginPage/utils/formValidation.js
```diff
@@ -0,0 +1,68 @@
...
+    'user[email]': constraintsTypes.email,
+    'user[password]': constraintsTypes.password,
+    'user[password_confirmation]': constraintsTypes.password_confirmation
+  }
+}
+
+const validateAllFields = (form: HTMLFormElement) => {
+  const errors = validate(form, constraints[form.id])
+
+  if (errors) {
```
didierofrivia: So we are using validate.js to validate the form, but then we are overriding its format with `false` ?
Isn't there a way to just tell validate to return that already?
if not, still having doubts, but you could simplify this with...

```
return !!errors ? Object.keys(errors).reduce((object, item) => { obj[item] = false; return obj }, {}) : {}
```
10 lines in 1 ;) 

PS: I'm still not sure if using validate.js and override its format...

app/javascript/src/LoginPage/loginForms/SignupForm.jsx
```diff
@@ -53,68 +67,68 @@ const InputFormGroup = (props) => {
...
+      validation
     })
+  }
 
-  handleTextInputPasswordConfirmation = (passwordConfirmation: string) =>
-    this.setState({
-      passwordConfirmation,
-      isFilledPasswordConfirmation: passwordConfirmation !== '' && passwordConfirmation === this.state.password
-    })
+  renameValidation = (namesToStateKeys: NamesToKeys, invalidFields: Validation) => Object.keys(invalidFields)
```
didierofrivia: Have you checked if validate.js can be assigned another format to avoid having this mapping function?

damianpm: Yes, I checked and it works with the `name` attribute: 
"This function will find all named inputs (inputs that specify the name attribute) and collect their values."
https://validatejs.org/#utilities-collect-form-values

The issue is for this page the name attribute comes in this format `user[username]`.  In the other forms is just `username`. And I don't want to add this names to the state `user[username]`, because it looks like object property accessors syntax

app/javascript/src/LoginPage/utils/formValidation.js
damianpm: I can try passing a custom validator

app/javascript/src/LoginPage/SignupPageWrapper.jsx
```diff
@@ -4,24 +4,40 @@ import React from 'react'
...
-    username: string
+    username: string,
+    errors: Errors
   },
   name: string,
   path: string
 }
 
+const ErrorMessages = ({errors}: {errors: Errors}) => {
+  return Object.entries(errors).map(
```
didierofrivia: You could avoid this if from the beginning return the desired format, apparently, an Array. Then the `errors` in the state would be an Array of Objects instead of an Object that it was parsed to store it and then it needs to be parse (again) for using it

app/javascript/src/LoginPage/loginForms/SignupForm.jsx
```diff
@@ -53,68 +54,85 @@ const InputFormGroup = (props) => {
...
+      validation
     })
+  }
 
-  handleTextInputPasswordConfirmation = (passwordConfirmation: string) =>
-    this.setState({
-      passwordConfirmation,
-      isFilledPasswordConfirmation: passwordConfirmation !== '' && passwordConfirmation === this.state.password
-    })
+  validateForm = (event: SyntheticEvent<HTMLInputElement>) => {
```
didierofrivia: This should be extracted to a common one, it's always the same for all the forms... "setState(...) unless invalidFields"...

app/javascript/src/LoginPage/loginForms/RequestPassWordForm.jsx
```diff
@@ -3,7 +3,12 @@
 import React from 'react'
```
didierofrivia: It should be named RequestPasswordForm.jsx`

app/javascript/src/LoginPage/loginForms/SignupForm.jsx
damianpm: I'm not sure if that's a good idea.
If we extract it, we'd need to use some kind of functional `setState` call. And that's  overkilling. 
If we didn't have a call to `setState` here, I'd say yes, but if we have to pass setState a function just to do the same it's currently doing I don't see the benefit.

didierofrivia: Ok, so let's see if how do you like this:

1. Have just one form component. You'll use it for those 3 forms.
2. This form will accept the "validators" and "inputs" list as props.  You'll compose the form and validate given those props injected.
3. Profit!

app/javascript/src/LoginPage/utils/formValidation.js
```diff
@@ -0,0 +1,65 @@
...
+  password_confirmation: {
+    presence: true,
+    length: {minimum: 1},
+    equality: {
+      attribute: 'user[password]'
+    }
+  }
+}
+
+const constraints = {
```
didierofrivia: I'd rather have separate objects in each form, create a `validate` instance and inject it in the form props. Then, when the time of validation comes, you do it there and not here with a lot of data that is not obvious which one is important.

Another note, we should get the validation rules from Rails in order to avoid duplication. This can be achieved in a different PR.

app/javascript/src/LoginPage/utils/formValidation.js
```diff
@@ -0,0 +1,65 @@
...
+// See: https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/validate.js_v0.x.x/flow_v0.25.x-/validate.js_v0.x.x.js
+// $FlowFixMe
+validate.formatters.customFormat = (errors) => {
+  return errors.reduce((obj, item) => {
+    obj[item.attribute] = false
+    return obj
+  }, {})
+}
+
+const validateAllFields = (form: HTMLFormElement) => validate(form, constraints[form.id], {format: 'customFormat'})
```
didierofrivia: With the approach mentioned above, these fn are not necessary

app/javascript/src/LoginPage/loginForms/SignupForm.jsx
didierofrivia: Another approach:

1. Keep these forms
2. Compose the validation object and create a validate instance, you could inject this with props
3. Then something like this:
```
(!!validate(...).length) && e.preventDefault(); setState(...)
``` 

OR

Just leave it to single input validation and disable the submit button while invalid/pristine

app/javascript/src/LoginPage/loginForms/SignupForm.jsx
```diff
@@ -20,101 +22,94 @@ type Props = {
...
+  'user[email]': string,
+  'user[first_name]': string,
+  'user[last_name]': string,
+  'user[password]': string,
+  'user[password_confirmation]': string,
+  validation: Validation
 }
 
+const formFieldsList = [
+  {isRequired: true, type: 'user[username]'},
```
didierofrivia: I'm sure this could be avoided with the validate instance crafted and injected in the component

app/javascript/src/LoginPage/loginForms/SignupForm.jsx
```diff
@@ -20,101 +22,94 @@ type Props = {
...
-  }
-  return (
-    <FormGroup isRequired={isRequired}
-      type={type}
-      labelIsValid={isValid}
-      inputProps={inputProps}
-    />
-  )
+type State = {
+  'user[username]': string,
```
didierofrivia: these strings could be extracted to constants and reuse in order to avoid bugs

app/javascript/src/LoginPage/loginForms/Login3scaleForm.jsx
```diff
@@ -18,73 +23,64 @@ type Props = {
...
+    validation: {}
   }
 
-  handleTextInputUsername = (username: string) => {
-    this.setState({ username }, this.setIsValidUsername)
-  }
-
-  setIsValidPassword = () => {
-    this.setState({isValidPassword: this.state.password !== ''})
+  handleInputChange = (value: string, event: SyntheticEvent<HTMLInputElement>) => {
```
didierofrivia: Here you could follow the same approach as I mentioned in the validateAll when submitting the form

app/javascript/src/LoginPage/loginForms/SignupForm.jsx
damianpm: I will go with the simplest approach: single input validation and disabling submit button by default

app/javascript/src/LoginPage/SignupPageWrapper.jsx
```diff
@@ -5,43 +5,28 @@ import {createReactWrapper} from 'utilities/createReactWrapper'
...
-
-const SignupPage = ({user, name, path}: Props) => (
+const SignupPage = ({user, name, path}: SignupProps) => (
   <LoginPage
     brandImgSrc={brandImg}
     brandImgAlt='Red Hat 3scale API Management'
     backgroundImgSrc={PF4DownstreamBG}
     backgroundImgAlt='Red Hat 3scale API Management'
-    loginTitle={`Signup to ${name}`}
+    loginTitle={`Signup to ${String(name)}`}
```
damianpm: `String(name)`is needed because flow complaints about coercion of maybe types

app/javascript/src/LoginPage/utils/formValidation.js
```diff
@@ -0,0 +1,29 @@
...
+    length: {minimum: 1}
+  },
+  email: {
+    presence: true,
+    email: true,
+    length: {minimum: 1}
+  },
+  password: {
+    presence: true,
+    length: {minimum: 6}
```
hallelujah: So this is validating in JS + in Ruby too when the form is submitted right?
It is duplicated there https://github.com/3scale/porta/blob/master/app/models/user.rb#L80
OK for now, but I think this kind of thing will hit us really hard when it gets complicated with logic in both views and in backend

damianpm: We'll handle that in a future task: https://issues.jboss.org/browse/THREESCALE-3352

app/javascript/src/LoginPage/utils/formValidation.js
```diff
@@ -0,0 +1,29 @@
+// @flow
+
+import validate from 'validate.js'
+
+const constraintsTypes = {
```
josemigallas: @hallelujah @damianpm 
To avoid duplication between client and server (regarding validation criteria) this should be generated by Ruby according to what's defined in the model.

damianpm: Yes, but it can be done as a following task: 
https://issues.jboss.org/browse/THREESCALE-3352
Is better to have 2 different tasks, this one is big enough to close it

app/javascript/src/LoginPage/utils/formValidation.js
hallelujah: Good for me 👍

PR #1007, opened 2019-07-16
[bug] Proxy Pro is not working in the UI anymore

hallelujah:
:warning: This is just to see if test passes... it should not

Well the feature has been removed but still this test is strange
Probably some JS rework has made it fail


hallelujah:
Closing, not going to fix it

PR #1010, opened 2019-07-18
[Bugfix] DeleteObjectHierarchyWorker raise URI::GID::MissingModelIdError

Martouta:
I've seen that we have this error while I was working on https://github.com/3scale/porta/pull/981
I separate it because it is unrelated and also more urgent.

### DIFF
--- a/app/workers/delete_object_hierarchy_worker.rb
+++ b/app/workers/delete_object_hierarchy_worker.rb
@@ -53,7 +53,7 @@ class DeleteObjectHierarchyWorker < ActiveJob::Base
   def build_batch
     batch = Sidekiq::Batch.new
     batch.description = batch_description
-    batch_callbacks(batch) { batch.jobs { delete_associations(object) } }
+    batch_callbacks(batch) { batch.jobs { delete_associations } }
     batch
   end
 
@@ -74,7 +74,7 @@ class DeleteObjectHierarchyWorker < ActiveJob::Base
     end
   end
 
-  def delete_associations(object)
+  def delete_associations
     object.class.reflect_on_all_associations.each do |reflection|
       next unless reflection.options[:dependent] == :destroy
       ReflectionDestroyer.new(object, reflection, caller_worker_hierarchy).destroy_later
--- a/app/workers/delete_object_hierarchy_worker.rb
+++ b/app/workers/delete_object_hierarchy_worker.rb
@@ -106,13 +106,13 @@ class DeleteObjectHierarchyWorker < ActiveJob::Base
       main_object.public_send("#{reflection.name.to_s.singularize}_ids").each do |associated_object_id|
         associated_object = reflection.class_name.constantize.new
         associated_object.id = associated_object_id
-        delete_associated_object_later(associated_object)
+        delete_associated_object_later(associated_object) if associated_object.id
       end
     end
 
     def destroy_has_one_association
       associated_object = main_object.public_send(reflection.name)
-      delete_associated_object_later(associated_object) if associated_object
+      delete_associated_object_later(associated_object) if associated_object.try(:id)
     end
 
     def delete_associated_object_later(associated_object)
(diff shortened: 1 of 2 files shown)


app/workers/delete_object_hierarchy_worker.rb
```diff
@@ -74,7 +74,7 @@ def batch_callbacks(batch)
     end
   end
 
-  def delete_associations(object)
+  def delete_associations
```
Martouta: because `object` is [an instance variable](https://github.com/3scale/porta/blob/de3b4578961b7b1c6ccf8cf3f315a84943aec7ef/app/workers/delete_object_hierarchy_worker.rb#L51).

app/workers/delete_object_hierarchy_worker.rb
```diff
@@ -106,13 +106,13 @@ def destroy_has_many_association
       main_object.public_send("#{reflection.name.to_s.singularize}_ids").each do |associated_object_id|
         associated_object = reflection.class_name.constantize.new
         associated_object.id = associated_object_id
-        delete_associated_object_later(associated_object)
+        delete_associated_object_later(associated_object) if associated_object.id
```
Martouta: Imagine a `user` as `main_object` with some `member_permissions`. It has `dependent: :destroy`:
https://github.com/3scale/porta/blob/762be0132ce13f3927e3ed50e39320c92020a2fe/app/models/user/permissions.rb#L7

So doing `main_object.public_send("#{reflection.name.to_s.singularize}_ids")` means `user.member_permissions_ids`, but that's overwritten:
https://github.com/3scale/porta/blob/762be0132ce13f3927e3ed50e39320c92020a2fe/app/models/user/permissions.rb#L48-L52

So `associated_object.id = associated_object_id` does nothing because it is not its real ID (it is not even an integer) and the `associated_object` does not have an ID, so doing `delete_associated_object_later(associated_object)` would raise `URI::GID::MissingModelIdError` because it cannot enqueue an object without an ID.
So this line says "if it doesn't have an ID, don't enqueue, ignore it", and it will be destroyed anyway but without parallelism... when the user is destroyed, ActiveRecord will destroy the member_permission too.

I've added a test for this scenario 😄

app/workers/delete_object_hierarchy_worker.rb
```diff
@@ -106,13 +106,13 @@ def destroy_has_many_association
...
       end
     end
 
     def destroy_has_one_association
       associated_object = main_object.public_send(reflection.name)
-      delete_associated_object_later(associated_object) if associated_object
+      delete_associated_object_later(associated_object) if associated_object.try(:id)
```
Martouta: It is similar to https://github.com/3scale/porta/pull/1010/files#r304992523

I could reproduce it because in one of the tests for [#981 [credit card] unstore if gateway configured](https://github.com/3scale/porta/pull/981) I temporarily added a `dependent: :destroy` in here:
https://github.com/3scale/porta/blob/762be0132ce13f3927e3ed50e39320c92020a2fe/app/models/account/payment_details.rb#L7
And it was raising this same error for this:
https://github.com/3scale/porta/blob/762be0132ce13f3927e3ed50e39320c92020a2fe/app/models/account/payment_details.rb#L28-L46

Meaning that if it doesn't exist, it builds it, so the `associated_object` will be a `PaymentDetail` without an `ID`.


I couldn't add a test for this one because I don't know which what other model this happens, and that was just a temporary change for testing that PR.

test/workers/delete_object_hierarchy_worker_test.rb
```diff
@@ -177,19 +177,20 @@ def setup
...
 
-  class DeleteMemberPermissionTest < DeleteObjectHierarchyWorkerTest
+  class DeleteMemberPermissionThroughUserTest < ActiveSupport::TestCase
+    def setup
+      tenant = FactoryBot.create(:simple_provider)
+      @member = FactoryBot.create(:member, account: tenant)
+      @member_permission = FactoryBot.create(:member_permission, user: member)
+    end
 
-    def test_member_permission
```
Martouta: Correct me if I am wrong but I have the perception that this test wasn't really testing anything at all 😕

app/workers/delete_object_hierarchy_worker.rb
hallelujah: Why not put the if inside the delete_associated_object_later ?

Martouta: Right, done @hallelujah 😎

PR #1011, opened 2019-07-18
[2.7 ER1] Cleanup deleted objects through saving service owner

Martouta:
The idea is to implement this strategy: https://issues.jboss.org/browse/THREESCALE-2085?focusedCommentId=13756894&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13756894

- [X] Save Services in DeletedObject when the service is destroyed.
- [X] The Janitor destroys regularly stale DeletedObject... those are at the moment those objects whose owner has been destroyed, or service fathers, but only when they have been deleted at least 1 week ago and not earlier.
- [X] Destroy all the DeletedObjects whose owner is a Service that does not exist in the services table anymore.


app/workers/janitor_worker.rb
```diff
@@ -7,6 +7,7 @@ def perform
     return unless ThreeScale.config.janitor_worker_enabled
 
     PurgeOldUserSessionsWorker.perform_async
-    PurgeOldEventsWorker.perform_async
+    PurgeStaleObjectsWorker.perform_later(EventStore::Event.name, DeletedObject.name)
+    true
```
duduribeiro: Is this `true` required? 

I think the `assert JanitorWorker.new.perform` works if we return any value different from `nil`

Martouta: the `JanitorWorker.new.perform` removing this true returns `nil` 😞 so i can return any truthy value other than true but i am not sure if it is worth it 😄

duduribeiro: maybe you can test if `PurgeStaleObjectsWorker` was enqueued.. 

or leave the true 😂 not a big deal haha

app/models/deleted_object.rb
```diff
@@ -8,5 +8,14 @@ class DeletedObject < ApplicationRecord
...
+  scope(:deleted_owner, lambda do
+    join_sql = <<-SQL
+      INNER JOIN deleted_objects deleted_owner
+        ON deleted_objects.owner_type = deleted_owner.object_type
+        AND deleted_objects.owner_id = deleted_owner.object_id
+    SQL
+    joins(join_sql)
+  end)
+
+  scope :stale, -> { where.has { ((id.in DeletedObject.deleted_owner.select(:id)) | (object_type == Service.name)) & (created_at <= 1.week.ago) } }
```
macejmic: what's the meaning of `|` in this query?

Martouta: @macejmic It is an 'OR' in [baby_squeel](https://github.com/rzane/baby_squeel)

app/models/service.rb
```diff
@@ -38,6 +38,8 @@ class Service < ApplicationRecord
   before_destroy :stop_destroy_if_last_or_default
   after_destroy :update_account_default_service
 
+  after_commit :achieve_as_deleted, on: :destroy
```
macejmic: `achieve` ? what does it mean? :D

app/models/service.rb
```diff
@@ -518,6 +520,10 @@ def oidc_configuration
 
   private
 
+  def achieve_as_deleted
+    ::DeletedObject.create!(object: self, owner: account)
```
macejmic: what if `create!` would fail? it would rollback the destroy?

Martouta: it is `after_commit :achieve_as_deleted, on: :destroy`, so if the destroy fails, or even if it is done but it is inside a transaction and rollbacks, then it does not reach here 😄

Martouta: ah you mean this `create!` ... then nothing, it wouldn't rollback, i think 🤷‍♀

Martouta: it shouldn't rollback because it is not inside the same transaction but after_commit

app/models/service.rb
Martouta: hahahahah it should be `archieve` as I see from `gmail` 😊

app/models/service.rb
macejmic: ok, you're right.. `after_commit` is being called after the transaction.

app/models/service.rb
Martouta: i will fix it in another PR because I have to fix it for other classes too 😊

hallelujah: then `archive`

PR #1012, opened 2019-07-19
[2709] disable end user switch

macejmic:
THREESCALE-2709

Deactivate "End User Plans" for all SaaS accounts

### DIFF
--- a/app/lib/switches.rb
+++ b/app/lib/switches.rb
@@ -299,7 +299,7 @@ module Switches
     elsif switch == :require_cc_on_signup # visible only for existing providers as of 2016-07-05
       account.provider_can_use?(switch)
     else
-      true
+      switch != :end_users
     end
   end
 end
--- a/app/views/sites/usage_rules/edit.html.slim
+++ b/app/views/sites/usage_rules/edit.html.slim
@@ -48,9 +48,6 @@
           = form.inputs "Service Plan Changing", class: "use-service-plans #{hide_class_unless_service_plans_visible(@settings)}" do
             = render 'shared/plan_change_settings', form: form, permission: :change_service_plan_permission, label: ''
 
-        - if can? :see, :end_users
-          = form.input :end_user_plans_ui_visible
-
     = form.buttons do
       = form.commit_button
 
--- a/lib/tasks/end_users_switch.rake
+++ b/lib/tasks/end_users_switch.rake
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+namespace :end_user_switch do
+  desc 'Disable end user switch for all accounts'
+  task disable_all: :environment do
+    Settings.find_in_batches(batch_size: 5000) do |group|
+      ids = group.collect(&:id)
+      puts "Updating #{ids.length} settings."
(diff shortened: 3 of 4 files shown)


app/lib/switches.rb
```diff
@@ -299,7 +299,7 @@ def visible_ui?(switch)
     elsif switch == :require_cc_on_signup # visible only for existing providers as of 2016-07-05
       account.provider_can_use?(switch)
     else
-      true
+      !(switch == :end_users)
```
Martouta: It can be done also in 1 instruction: 😄 
```ruby
switch != :end_users
```

Martouta:
About the rake task:
1. is it for this PR or for another one?
2. I get the general idea but I want to clarify that `Settings.update_all(end_users_switch: false)` in our DB is scary and I would do it in smaller batches and with a sleep time between them 😄

app/views/sites/usage_rules/edit.html.slim
```diff
@@ -48,9 +48,6 @@
           = form.inputs "Service Plan Changing", class: "use-service-plans #{hide_class_unless_service_plans_visible(@settings)}" do
             = render 'shared/plan_change_settings', form: form, permission: :change_service_plan_permission, label: ''
 
-        - if can? :see, :end_users
-          = form.input :end_user_plans_ui_visible
```
Martouta: I guess we should also filter the param from here:
https://github.com/3scale/porta/blob/43383d8fe110b8ce25954a1748eb12b589966ace/app/controllers/sites/usage_rules_controller.rb#L10
And I dunno if there is an API for this or not 😄

macejmic: this part is only about UI.

app/lib/switches.rb
macejmic: true 🍳

macejmic:
@Martouta yep, an update will be done in batches.. 🍤

lib/tasks/end_users_switch.rake
```diff
@@ -0,0 +1,11 @@
+
+namespace :end_user_switch do
+  desc 'Disable end user switch for all accounts'
+  task disable_all: :environment do
+    Settings.find_in_batches(batch_size: 500) do |group|
+      ids = group.collect(&:id)
+      Settings.where(id: ids).update_all(end_users_switch: 'denied')
```
Martouta: good for me but you should check with ops that `bacth_size` of 500 with `update_all` and later `sleep(2)` is the way to go 😄

lib/tasks/end_users_switch.rake
```diff
@@ -0,0 +1,11 @@
+
+namespace :end_user_switch do
```
Martouta: I've checked `CodeClimate` and it complains in 2 issues:
image

The last one I don't mind if ops agrees (as said in https://github.com/3scale/porta/pull/1012/files#r306419635), but the first one is very easy to fix 😄 just running locally `rubocop lib/tasks/end_users_switch.rake -a` it fixes it alone, otherwise is adding manually that line on the top of the file 😄

PR #1013, opened 2019-07-19
Hide AccountPlan from Account#show view unless it has permission, like in the nav

Martouta:
Closes [THREESCALE-3005](https://issues.jboss.org/browse/THREESCALE-3005)

### Verification steps
Check `http://provider-admin.example.com.lvh.me:3000/buyers/accounts/3` locally
#### Before
image

#### After
image

### DIFF
--- a/app/helpers/plans_helper.rb
+++ b/app/helpers/plans_helper.rb
@@ -1,5 +1,10 @@
 module PlansHelper
 
+  def account_plans_management_visible?
+    settings = current_account.settings
+    can?(:manage, :account_plans) && settings.account_plans.allowed? && settings.account_plans_ui_visible?
+  end
+
   def can_create_plan?(plan)
     case plan.new
     when AccountPlan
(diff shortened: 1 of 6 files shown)


app/views/buyers/accounts/show.html.slim
```diff
@@ -81,7 +81,7 @@ h1 data-hook="account-show"
         = render 'monthly_charging', account: @account
 
   .right_narrow
-    - if can?(:manage, :account_plans)
+    - if can?(:manage, :account_plans) && current_account.settings.account_plans.allowed? && current_account.settings.account_plans_ui_visible?
```
didierofrivia: extract it to helper/decorator method `account_plans_visible?` ?

app/helpers/plans_helper.rb
```diff
@@ -1,5 +1,10 @@
 module PlansHelper
 
+  def account_plans_management_visible?
```
macejmic: nice extraction..

PR #1015, opened 2019-07-22
Improve Service Discovery Errors (AKA improve New API form)

josemigallas:

* Improves the way "error scenarios" are managed in the form, such as when projects is an empty string or when there's any error code.
* Improves *Flow* typings
* Improves the UI flow by adding a loading spinner and disabling inputs when there's an ongoing request
* Simplifies some pieces of logic

**Possible TODOs**
- [ ] Find a better way to show loading state
- [x] Improve the data structured received from the server
- [ ] Reduce amount of requests by checking for service discovery before selecting the radio
- [ ] Get projects/services data only once, not everytime the radio is selected (keep state in parent?)

[THREESCALE-2646: Service Discovery: Improve error messages and visual feedback when ajax request fails](https://issues.jboss.org/browse/THREESCALE-2646)

**Verification steps** 

Having Service Discovery enabled and logged in in Openshift go to `/apiconfig/services/new` and select *Import from Openshift*. Mess around the form and look for any errors.

**Visuals**
Expected:
new-api-good

When empty:
new-api-empty

When error:
new-api-error

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -12,6 +12,7 @@
 [lints]
 
 [options]
+module.system.node.resolve_dirname=node_modules
 module.name_mapper.extension='scss' -> 'empty/object'
 module.system=haste
 module.name_mapper='\(Applications\|Dashboard\|LoginPage\|Navigation\|Onboarding\|Policies\|services\|Stats\|Types\|Users\|utilities\|NewService\)\(.*\)$' -> '<PROJECT_ROOT>/app/javascript/src/\1/\2'
--- a/app/assets/stylesheets/provider/_forms.scss
+++ b/app/assets/stylesheets/provider/_forms.scss
@@ -538,5 +538,5 @@ li.hidden {
 }
 
 .new-service-source-input {
-  padding-left: line-height-times(5 / 10);
+  padding: 0 line-height-times(5 / 10);
 }
--- a/app/assets/stylesheets/provider/_forms.scss
+++ b/app/assets/stylesheets/provider/_forms.scss
@@ -1,3 +1,5 @@
+$fieldset-margin-bottom: line-height-times(1.5);
+
 form.formtastic ol, form.formtastic ul {
   list-style: none;
 }
@@ -310,7 +312,7 @@ form.formtastic {
     position: relative;
     padding: line-height-times(1) 0 0 0;
     border-top: $border-width solid $border-color;
-    margin-bottom: line-height-times(1.5);
+    margin-bottom: $fieldset-margin-bottom;
     -moz-appearance: none;
     -webkit-appearance: none;
     appearance: none;
@@ -533,10 +535,14 @@ li.hidden {
   display: none;
 }
 
-.errorMessage {
-  color: $error-color;
-}
+#new_service_wrapper {
+  .errorMessage {
+    color: $error-color;
+    margin-bottom: line-height-times(1/2);
+    margin-top: -#{$fieldset-margin-bottom};
+  }
 
-.new-service-source-input {
-  padding: 0 line-height-times(5 / 10);
+  .new-service-source-input {
+    padding: 0 line-height-times(5 / 10);
+  }
 }
--- a/app/controllers/provider/admin/service_discovery/cluster_projects_controller.rb
+++ b/app/controllers/provider/admin/service_discovery/cluster_projects_controller.rb
@@ -2,6 +2,7 @@
 
 class Provider::Admin::ServiceDiscovery::ClusterProjectsController < Provider::Admin::ServiceDiscovery::ClusterBaseController
   def index
-    render json: { projects: cluster.projects_with_discoverables.map(&:to_json) }
+    render json: cluster.projects_with_discoverables.map(&:name)
+                                                    .to_json
   end
 end
--- a/app/javascript/src/NewService/components/FormElements/ServiceDiscoveryListItems.jsx
+++ b/app/javascript/src/NewService/components/FormElements/ServiceDiscoveryListItems.jsx
@@ -1,25 +1,42 @@
 // @flow
 
-import React, {useEffect} from 'react'
+import React, {useState, useEffect} from 'react'
 import {Label, Select} from 'NewService/components/FormElements'
+import {fetchData} from 'utilities/utils'
+import {BASE_PATH} from 'NewService'
 
 import type { Option } from 'NewService/types'
 
 type Props = {
-  fetchServices: (namespace: string) => Promise<void>,
-  loading: boolean,
   projects: Option[],
-  services: Option[]
+  onError: (err: string) => void
 }
 
 const ServiceDiscoveryListItems = (props: Props) => {
-  const {fetchServices, loading, projects, services} = props
+  const { projects, onError } = props
+
+  const [services, setServices] = useState([])
+  const [loading, setLoading] = useState(true)
 
   useEffect(() => {
     if (projects && projects.length) {
       fetchServices(projects[0].metadata.name)
     }
   }, [projects])
+
+  const fetchServices = async (namespace: string) => {
+    setLoading(true)
+    setServices([])
+
+    try {
+      const { services } = await fetchData<{services: Option[]}>(`${BASE_PATH}/namespaces/${namespace}/services.json`)
+      setServices(services)
+    } catch (error) {
+      onError(error.message)
+    } finally {
+      setLoading(false)
+    }
+  }
 
   return (
     <React.Fragment>
--- a/app/javascript/src/NewService/components/FormElements/ServiceDiscoveryListItems.jsx
+++ b/app/javascript/src/NewService/components/FormElements/ServiceDiscoveryListItems.jsx
@@ -5,12 +5,13 @@ import {Label, Select} from 'NewService/components/FormElements'
 
 type Props = {
   fetchServices: (namespace: string) => Promise<void>,
-  projects: string[],
-  services: string[]
+  loading: boolean,
+  projects: Option[],
+  services: Option[]
 }
 
 const ServiceDiscoveryListItems = (props: Props) => {
-  const {fetchServices, projects, services} = props
+  const {fetchServices, loading, projects, services} = props
   return (
     <React.Fragment>
       <li id="service_name_input" className="string required">
@@ -19,6 +20,7 @@ const ServiceDiscoveryListItems = (props: Props) => {
           label='Namespace'
         />
         <Select
+          disabled={loading}
           name='service[namespace]'
           id='service_namespace'
           onChange={(e) => fetchServices(e.currentTarget.value)}
@@ -31,6 +33,7 @@ const ServiceDiscoveryListItems = (props: Props) => {
           label='Name'
         />
         <Select
+          disabled={loading}
           name='service[name]'
           id='service_name'
           options={services}
--- a/app/javascript/src/NewService/components/FormElements/ServiceDiscoveryListItems.jsx
+++ b/app/javascript/src/NewService/components/FormElements/ServiceDiscoveryListItems.jsx
@@ -40,7 +40,7 @@ const ServiceDiscoveryListItems = (props: Props) => {
     <React.Fragment>
       <li id="service_name_input" className="string required">
         <Label
-          htmlFor='namespace'
+          htmlFor='service_namespace'
           label='Namespace'
         />
         <Select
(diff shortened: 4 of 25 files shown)


app/controllers/provider/admin/service_discovery/cluster_projects_controller.rb
```diff
@@ -2,6 +2,7 @@
 
 class Provider::Admin::ServiceDiscovery::ClusterProjectsController < Provider::Admin::ServiceDiscovery::ClusterBaseController
   def index
-    render json: { projects: cluster.projects_with_discoverables.map(&:to_json) }
+    render json: cluster.projects_with_discoverables.map(&:metadata)
```
didierofrivia: Are service and project names unique?

app/javascript/src/NewService/components/FormElements/ServiceDiscoveryListItems.jsx
```diff
@@ -1,28 +1,56 @@
...
+import {fetchData} from 'utilities/utils'
+import {BASE_PATH} from 'NewService'
+
+import type { ClusterResource } from 'NewService/types'
 
 type Props = {
-  fetchServices: (namespace: string) => Promise<void>,
-  projects: string[],
-  services: string[]
+  projects: ClusterResource[],
```
didierofrivia: Seems a bit odd to have a type for something that holds just a prop that is a string. Can't we just return the name from Rails and have a type `string[]`

spec/javascripts/NewService/ServiceDiscoveryForm.spec.jsx
```diff
@@ -33,3 +35,40 @@ it('should render `ServiceDiscoveryListItems` child', () => {
...
+
+  it('should fetch projects when first redendered', done => {
+    const projects = [{ name: 'project_00' }]
+    fetch.mockImplementation(url => projects)
+
+    const setState = jest.fn(val => {
+      expect(val).toEqual(projects)
+      done()
+    })
+    const useState = jest.spyOn(React, 'useState')
```
didierofrivia: I don't see a lot of value mocking React.useState just to assert it was called... And specially on init, not when it was expected to be called after a change or with specific values

spec/javascripts/NewService/formElements/FormWrapper.spec.jsx
```diff
@@ -1,38 +1,51 @@
...
 
 it('should render itself', () => {
-  const wrapper = mount(<FormWrapper {...props}/>)
+  const wrapper = shallow(<FormWrapper {...props}/>)
   expect(wrapper.find('#form-id').exists()).toEqual(true)
 })
 
+it('should render an empty input', () => {
+  const wrapper = shallow(<FormWrapper {...props}/>)
+  const button = wrapper.find(`input[name="utf8"][type='hidden']`)
```
didierofrivia: button (?)

spec/javascripts/NewService/formElements/FormWrapper.spec.jsx
```diff
@@ -1,38 +1,51 @@
...
+  expect(button.exists()).toBe(true)
+})
+
 it('should render submit button with proper text', () => {
```
didierofrivia: Maybe also another test to assert submitting the form when clicking the button?

app/controllers/provider/admin/service_discovery/cluster_projects_controller.rb
josemigallas: I believe so

app/javascript/src/NewService/components/FormElements/ServiceDiscoveryListItems.jsx
josemigallas: I understand what you say, I decided to create a type so that it is extensible. Maybe eventually we require other fields like a human name. But it you think that's not likely going to happen I agree it is enough with strings.

spec/javascripts/NewService/ServiceDiscoveryForm.spec.jsx
josemigallas: Rather than mocked it is _spied_, which is the only way [I know] to assert the projects are fetch.
The test doesn't have much value I agree but it is there the increase coverage (for the useEffect).

spec/javascripts/NewService/formElements/FormWrapper.spec.jsx
josemigallas: How would we do that?

spec/javascripts/NewService/ServiceDiscoveryForm.spec.jsx
josemigallas: > I don't see a lot of value mocking React.useState just to assert it was called
have in mind it is `setProjects` what's being asserted, not useState

app/javascript/src/NewService/components/FormElements/ServiceDiscoveryListItems.jsx
didierofrivia: Let's go for just strings to make it obvious and same-same with services. Also, regarding my previous comment in https://github.com/3scale/porta/pull/1015#discussion_r308259340, we could just return the names (strings) array

spec/javascripts/NewService/ServiceDiscoveryForm.spec.jsx
didierofrivia: Yes, spied, sorry. I meant that just asserting `setProjects` was called, like without the projects, was a bit weak given the fact you are spying useState. Since you already have the fetch mocked, maybe asserting that it was called with the mocked projects would be more valuable.

spec/javascripts/NewService/NewServiceForm.spec.jsx
```diff
@@ -1,8 +1,14 @@
...
 import React from 'react'
 import Enzyme, {mount} from 'enzyme'
 import Adapter from 'enzyme-adapter-react-16'
 
-import {NewServiceForm} from 'NewService'
+import {NewServiceForm, ServiceManualForm, ServiceDiscoveryForm} from 'NewService'
+
+import * as utils from 'utilities/utils'
+jest.spyOn(utils, 'CSRFToken')
+  .mockImplementation(() => '')
```
damianpm: Instead of mocking the implementation in every file, maybe we could mock the module

spec/javascripts/NewService/NewServiceForm.spec.jsx
```diff
@@ -12,49 +18,23 @@ const props = {
...
   const wrapper = mount(<NewServiceForm {...props}/>)
-  expect(wrapper.find('#new_service').exists()).toEqual(true)
-  expect(wrapper.find('#service_discovery').exists()).toEqual(false)
-})
 
-// TODO: remove `skip` when this is merged: https://github.com/airbnb/enzyme/pull/2008
-it.skip('should render new Service Discovery form when click on input', () => {
-  const wrapper = mount(<NewServiceForm {...props}/>)
-  wrapper.find('#source_discover').simulate('click')
+  wrapper.find('#source_discover').simulate('change')
```
damianpm: As `simulate()` is being deprecated you could call `onChange` from props()

.flowconfig
```diff
@@ -12,6 +12,7 @@
 [lints]
 
 [options]
+module.system.node.resolve_dirname=node_modules
```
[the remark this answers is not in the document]
damianpm: Nvm, I saw why you include it

spec/javascripts/NewService/NewServiceForm.spec.jsx
josemigallas: Yes you're right! 👍

spec/javascripts/NewService/NewServiceForm.spec.jsx
josemigallas: Yes 100% agree. But in order not to get too much out of scope with this PR I created this [other ticket](https://issues.jboss.org/browse/THREESCALE-3104) for that and other repetitive setups.

app/javascript/src/NewService/components/ServiceDiscoveryForm.jsx
```diff
@@ -1,47 +1,43 @@
...
+  formActionPath: string,
+  setLoadingProjects: boolean => void
+}
 
-const ServiceDiscoveryForm = ({formActionPath}: {formActionPath: string}) => {
-  const [projects, setProjects] = useState([])
-  const [services, setServices] = useState([])
-  const [fetchErrorMessage, setFetchErrorMessage] = useState('')
+const ServiceDiscoveryForm = ({formActionPath, setLoadingProjects}: Props) => {
+  // Don't use named imports so that useState can be mocked in specs
```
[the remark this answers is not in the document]
josemigallas: The reason why is to inject a spied `setProjects` so that the receiving projects can be asserted. Ideally it would simply assert the projects are rendered but since it's an async flow, there's no way to tell Jest to "wait" except by spying `setProjects`.

I read the issue but I can't see any way to workaround this, it seems all it does is to support hooks functionally when using Enzyme, but not direct access to them.

app/javascript/src/NewService/components/FormElements/ServiceDiscoveryListItems.jsx
josemigallas: Ah, I see what you meant now

spec/javascripts/NewService/ServiceDiscoveryForm.spec.jsx
josemigallas: I disagree, we "don't care" about the fetch method here, only that the component is receiving projects at some point. Asserting `fetch` doesn't mean `setProjects` will be called, or that the projects will be added to the state.

In any case, the main reason why `setProjects` has to be asserted is because fetching is inside an async effect and it isn't possible to tell Enzyme to wait for that. The only way is to call `done` when `setProjects` is called.

Does this make sense?

didierofrivia: Well, I never said we should care about the fetch method, just that `setProjects` was receiving the right projects, which you actually are expecting here: https://github.com/3scale/porta/pull/1015/files#diff-9616eeee01eb7fb9cf1dbaf45ae410b9R63 which I missed in my first comment. I just find it a bit dirty, spying the setState, mocking the fetch and setProjects... but I guess it's ok if we can't just mock the fetch and assert after render the list items rendered or the props.

spec/javascripts/NewService/formElements/FormWrapper.spec.jsx
didierofrivia: you render the form within an `act` scope, then find the button and dispatch a click event.

app/assets/stylesheets/provider/_forms.scss
```diff
@@ -530,10 +532,14 @@ li.hidden {
...
 }
 
-.errorMessage {
-  color: $error-color;
-}
+#new_service_wrapper {
+  .errorMessage {
+    color: $error-color;
+    margin-bottom: line-height-times(1/2);
+    margin-top: -#{$fieldset-margin-bottom};
```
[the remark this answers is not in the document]
josemigallas: Yes, it is using the fieldset's margin-bottom. It's been refactored into a variable instead of simply 🍝'ing the value.

spec/javascripts/NewService/ServiceDiscoveryForm.spec.jsx
josemigallas: OK all clear now. It maybe looks a bit hacky but it is the only way to do assertions with `useState` hook (as far as I know). So I will leave it like this if that's OK.

spec/javascripts/NewService/formElements/FormWrapper.spec.jsx
josemigallas: Sure, but what is to be expected after the button is clicked? There's no UI feedback, no callback, just a pure http request.

didierofrivia: Right, sorry, there's no React involved in handling the submission, it's just left to Rails controller. Next time I suggest if you already know there's an obvious case and the person reviewing your code missed that, present it clearly, instead of fishing for a wrong answer. A simple "It's Rails managed, probably I'm missing something?" or "I might be missing something, why would we want to test a Rails form submission?" or "I think it's not needed since it's left to the Rails controller" or simply a "Why?" instead of a "How?" would have just satisfied the inquire and avoid the thread.... including this patronizing comment ;)

josemigallas: OK, my bad 🙈

PR #1016, opened 2019-07-22
Allow choosing between 3scale managed or self managed for on-premises

hallelujah:
Fixes THREESCALE-3001

Relates to https://github.com/3scale/Red_Hat_3scale_Documentation/pull/470

~Trying to fix the issues on a migration if possible (for existing services created on version 2.5)~

- [ ] ~What to set for currently existing Service, should it be self_managed or hosted?~
- [ ] ~What if APIcast is actually self managed but on the same OCP?~
  - [ ] ~On same THREESCALE_WILDCARD_DOMAIN and the route exists?~
  - [ ] ~On another domain, and the route exists?~
- [ ] ~What if APIcast is actually self managed but hosted somewhere else, so the route does not exist~
- [ ] ~What if actually the customers wants it to be `hosted` but sets a route in order to bypass the wildcard-router?~

### DIFF
--- a/test/unit/service_test.rb
+++ b/test/unit/service_test.rb
@@ -650,6 +650,16 @@ class ServiceTest < ActiveSupport::TestCase
       refute_includes gateways, 'service_mesh_istio'
     end
 
+    def test_gateways_with_apicast_custom_url
+      ThreeScale.config.stubs(apicast_custom_url: true)
+      gateways = Service::DeploymentOption.gateways
+
+      assert_includes gateways, 'self_managed'
+      assert_includes gateways, 'hosted'
+      refute_includes gateways, 'plugin_ruby'
+      refute_includes gateways, 'service_mesh_istio'
+    end
+
     def test_plugins
       plugins = Service::DeploymentOption.plugins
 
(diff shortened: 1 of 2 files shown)


hallelujah:
Not going the migration way. Too many unknown and introducing risks for 2.5 on-premises users if we make the migration.

Basically what we @3scale/system consider that setting via the API after the migration the correct deployment_option is better. Zync pod should be scaled to 0.

thomasmaas:
@mikz might have an opinion on this

test/unit/service_test.rb
```diff
@@ -650,6 +650,16 @@ def test_gateways
       refute_includes gateways, 'service_mesh_istio'
     end
 
+    def test_gateways_with_apicast_custom_url
+      ThreeScale.config.stubs(apicast_custom_url: true)
```
guicassolato: This won't really change the test, right?

guicassolato: I understand the idea is to show that even with `apicast_custom_url: true` it now shows the `self_managed` option. Maybe also showing that the test passes with `apicast_custom_url: false`?

hallelujah: It is tested above already

PR #1017, opened 2019-07-23
Raise an error and rollback when the seeds don't seed the DB correctly

Martouta:
Closes [THREESCALE-2951](https://issues.jboss.org/browse/THREESCALE-2951)

Things that I don't know how to test:
1. The password
2. The values of `APICAST_ACCESS_TOKEN`, `MASTER_ACCESS_TOKEN` and `ADMIN_ACCESS_TOKEN`

### DIFF
--- a/app/models/fields_definition.rb
+++ b/app/models/fields_definition.rb
@@ -52,14 +52,13 @@ class FieldsDefinition < ApplicationRecord
 
   alias_attribute :position, :pos
 
-  # this is tested in unit/account_test
-  def self.create_defaults(account)
+  # This smells of :reek:NestedIterators: FieldsDefinition#self.create_defaults! contains iterators nested 2 deep
+  def self.create_defaults!(account)
     targets.each do |target|
       klass = target.constantize
-      klass.required_fields.each do |f|
-        label = DEFAULT_LABELS[f] || f.humanize
-        account.fields_definitions.create!({ :target => target, :name => f,
-                                             :label => label, :required => true })
+      klass.required_fields.each do |field|
+        label = DEFAULT_LABELS[field] || field.humanize
+        account.fields_definitions.create!({target: target, name: field, label: label, required: true})
       end
     end
   end
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -202,17 +202,13 @@ ActiveRecord::Base.transaction do
   end
 
   # Import CMS templates
-  SimpleLayout.new(provider).import! if ENV.fetch('CMS_TEMPLATES', Rails.env.development?).presence
+  SimpleLayout.new(provider).import! if ENV['CMS_TEMPLATES'].present?
 
   ###
   #  Creating Sample Data
   ###
 
-  if Rails.env.development?
-    SignupWorker::SampleDataWorker.new.perform(provider.id)
-  else
-    SignupWorker.enqueue(provider)
-  end
+  SignupWorker.enqueue(provider)
 
 
   puts '='*80
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -213,20 +213,17 @@ ActiveRecord::Base.transaction do
 
   puts <<~INFO
     #{'='*80}
-    \n
     Setup Completed
-    \n
-    Root Domain: #{ThreeScale.config.superdomain}
+
+    Root Domain: #{ThreeScale.config.superdomain}\n
   INFO
 
   if master_login && master_password
     puts <<~INFO
-      \n
       Master Domain: #{master.admin_domain}
       Master User Login: #{master_login}
       Master User Password: #{master_password}
-      Master RW access token: #{master_access_token}
-      \n
+      Master RW access token: #{master_access_token}\n
     INFO
   end
 
@@ -237,7 +234,6 @@ ActiveRecord::Base.transaction do
     Provider User Password: #{user_password}
     APIcast Access Token: #{apicast_access_token}
     #{'Admin Access Token: ' + admin_access_token if admin_access_token}
-    \n
     #{'='*80}
   INFO
 
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,249 +1,242 @@
+# frozen_string_literal: true
+
 require 'dotenv/rails-now'
 
-if Rails.env.test? || System::Application.config.three_scale.core.fake_server
-  require Rails.root.join('test/test_helpers/backend')
-  TestHelpers::Backend::MockCore.mock_core!
-end
 
-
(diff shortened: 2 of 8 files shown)


test/unit/db/seeds_test.rb
```diff
@@ -5,7 +5,178 @@
 class SeedsTest < ActiveSupport::TestCase
   disable_transactional_fixtures!
 
-  test 'the seeds do not fail' do
-    assert Rails.application.load_seed
+  def teardown
```
macejmic: it could be defined on one place:

```
env_variables = {
  MASTER_NAME: 'master name' 
}
```

and then:

```
env_variables.each do |key, value|
  ENV[key] = value
```
and then the same for `teardown`

macejmic: to not left any leftovers by any chance..

Martouta: Right 🚀 💪

Martouta: Done 😎 💃

app/models/fields_definition.rb
```diff
@@ -52,14 +52,13 @@ def self.editable_by(user)
 
   alias_attribute :position, :pos
 
-  # this is tested in unit/account_test
```
Martouta: This comment is not even true anymore 😂 Since long ago it is tested in `test/unit/extra_fields_test.rb`

db/seeds.rb
```diff
@@ -1,249 +1,242 @@
...
+  master_service.service_plans.default!(ServicePlan.first!)
+
+  # Master needs to be contracted with an ApplicationPlan
+  application_plan = ApplicationPlan.create!(issuer: master_service, name: master_plan)
+  master_service.default_application_plan = application_plan
+  range = [0] | Alert::ALERT_LEVELS
+  master_service.notification_settings = { web_buyer: range, email_buyer: range, web_provider: range, email_provider: range}
+  master_service.save!
+
+  application_plan.create_contract_with! master
```
guicassolato: This is the same as `master.buy! application_plan`, right?

db/seeds.rb
```diff
@@ -1,249 +1,242 @@
...
-  provider.update_columns(sample_data: false)
-end
+  ###
+  #  Creating Sample Data
+  ###
 
-# Import CMS templates
-SimpleLayout.new(provider).import! if ENV.fetch('CMS_TEMPLATES', Rails.env.development?).presence
+  if Rails.env.development?
+    SignupWorker::SampleDataWorker.new.perform(provider.id)
```
guicassolato: Nice!

guicassolato: Why not invoking `ImportSimpleLayoutWorker.new.perform(provider.id)` as well though?

db/seeds.rb
```diff
@@ -1,249 +1,242 @@
...
   puts
-end
+  puts "Root Domain: #{ThreeScale.config.superdomain}"
+
+  if master_login && master_password
+    puts
+    puts "Master Domain: #{master.admin_domain}"
+    puts "Master User Login: #{master_login}"
+    puts "Master User Password: #{master_password}"
+    puts "Master RW access token: #{master_access_token}"
```
guicassolato:
```ruby
puts <<~INFO
  Master Domain: #{master.admin_domain}
  Master User Login: #{master_login}
  Master User Password: #{master_password}
  Master RW access token: #{master_access_token}
INFO
```

db/seeds.rb
Martouta: Right!! I will do it 👍 💪

db/seeds.rb
```diff
@@ -1,249 +1,242 @@
...
+                             ])
+  end
 
-###
-#  Creating Provider Account
-###
+  master_user = master.users.create!(username: master_login, password: master_password, password_confirmation: master_password) do |user|
+    user.signup_type = 'minimal'
+    user.role = :admin
+    user.state = 'active'
```
guicassolato: This will not fill `activated_at`. Better keep using `user.activate!`

Martouta: @guicassolato It fills it for sure because it is in the tests 😄

Martouta: ah right, the `activated_at` !!! will do it 👍 💪

app/models/fields_definition.rb
```diff
@@ -52,14 +52,13 @@ def self.editable_by(user)
...
-  def self.create_defaults(account)
+  # This smells of :reek:NestedIterators: FieldsDefinition#self.create_defaults! contains iterators nested 2 deep
+  def self.create_defaults!(account)
     targets.each do |target|
       klass = target.constantize
-      klass.required_fields.each do |f|
-        label = DEFAULT_LABELS[f] || f.humanize
-        account.fields_definitions.create!({ :target => target, :name => f,
-                                             :label => label, :required => true })
+      klass.required_fields.each do |field|
```
duduribeiro: couldn't this logic be moved to `lib/fields/fields.rb`? this way I think it will not smells :reek:NestedIterators:. WDYT?

test/unit/db/seeds_test.rb
```diff
@@ -5,7 +5,187 @@
...
+    assert_equal master_service.id, tenant_app_plan.issuer_id
+
+    assert(apicast = master_account.access_tokens.find_by(name: 'APIcast'))
+    assert_equal 'ro', apicast.permission
+    assert_equal %w[account_management], apicast.scopes
+    assert(master_token = master_account.access_tokens.find_by(name: 'Master Token'))
+    assert_equal 'rw', master_token.permission
+    assert_equal %w[account_management], master_token.scopes
+
+    #### TENANT
```
duduribeiro: WDYT in isolate `test_default_values_for_master` and `test_default_values_for_tenant` ? and maybe create some `assert_default_values` method ? Don't know if it is valid :smile:

Martouta: The problem is it would be very slow to run the seeds more times than necessary 😞 So I dunno what's better... 🤷‍♀

app/models/fields_definition.rb
Martouta: I really don't want to touch this class or method at all for this PR... it is already a risky PR like to make it more risky... I only changed the `!` at the end of the method because it is what it is, and CodeClimate started complaining about things that were already wrong in this code... 🤷‍♀

test/unit/db/seeds_test.rb
duduribeiro: makes totally sense.. You can ignore it 😄

app/models/fields_definition.rb
duduribeiro: 👍

PR #1018, opened 2019-07-23
Zync domain events cascade

mikz:
Create domain changed events for all proxies/providers matching the same domain. 
So for example when it is no longer used it can be recreated by zync for another provider/proxy.

### DIFF
--- a/app/events/domains/provider_domains_changed_event.rb
+++ b/app/events/domains/provider_domains_changed_event.rb
@@ -1,9 +1,13 @@
 class Domains::ProviderDomainsChangedEvent < BaseEventStoreEvent
-  def self.create(provider)
+  def self.create(provider, parent_event = nil)
     new(
+      parent_event_id: parent_event&.event_id,
+      parent_event_type: parent_event&.class&.name,
+
       provider:    provider,
       admin_domains: [ provider.self_domain ],
       developer_domains: [ provider.domain ],
+
       metadata: {
         provider_id: provider.id,
         zync: {
@@ -11,6 +15,18 @@ class Domains::ProviderDomainsChangedEvent < BaseEventStoreEvent
         }
       }
     )
+  end
+
+  def domains
+    admin_domains + developer_domains
+  end
+
+  def parent_event?
+    parent_event_id && parent_event_type
+  end
+
+  def after_commit
+    ProcessDomainEventsWorker.enqueue(self)
   end
 
   def self.valid?(account)
--- a/app/models/proxy.rb
+++ b/app/models/proxy.rb
@@ -69,6 +69,7 @@ class Proxy < ApplicationRecord
             :credentials_location, :error_auth_failed, :error_auth_missing, :authentication_method,
             :error_headers_auth_failed, :error_headers_auth_missing, :error_headers_limits_exceeded, :error_limits_exceeded,
             :error_no_match, :error_headers_no_match, :secret_token, :hostname_rewrite, :sandbox_endpoint,
+            :staging_domain, :production_domain,
             length: { maximum: 255 }
 
   validates :oidc_issuer_type, inclusion: { in: OIDC_ISSUER_TYPES.keys.map(&:to_s), allow_blank: true }, presence: { if: ->(proxy) { proxy.oidc_issuer_endpoint.present? } }
@@ -87,6 +88,7 @@ class Proxy < ApplicationRecord
   validates :endpoint, presence: true, on: :update, if: :require_production_endpoint?
 
   before_save :set_correct_endpoints, if: :set_correct_endpoints?
+  before_save :update_domains
   after_save :publish_events
   before_destroy :publish_events
 
@@ -471,6 +473,15 @@ class Proxy < ApplicationRecord
     URI(sandbox_endpoint || set_sandbox_endpoint).host
   end
 
+  def update_domains
+    domains = {
+      staging_domain: self.class.extract_domain(sandbox_endpoint.presence || default_staging_endpoint),
+      production_domain: self.class.extract_domain(endpoint.presence || default_production_endpoint),
+    }
+    assign_attributes(domains)
+    domains
+  end
+
   def provider
     @provider ||= self.service.account
   end
@@ -635,6 +646,12 @@ class Proxy < ApplicationRecord
 
   def set_port_endpoint
     generate_port(:endpoint)
+  end
+
+  def self.extract_domain(url)
+    URI(url.presence || '').host
+  rescue ArgumentError, URI::InvalidURIError
+    # nothing
   end
 
   def generate_port(proxy_attribute)
--- a/app/workers/process_domain_events_worker.rb
+++ b/app/workers/process_domain_events_worker.rb
@@ -1,0 +1,49 @@
+class ProcessDomainEventsWorker
+  include Sidekiq::Worker
+
+  class_attribute :publisher
+  self.publisher = ->(*args) { Rails.application.config.event_store.publish_event(*args) }
+
+  def perform(event)
+    events = find_providers(event).map { |provider| Domains::ProviderDomainsChangedEvent.create(provider, event) }
+    events += find_proxies(event).map { |proxy| Domains::ProxyDomainsChangedEvent.create(proxy, event) }
+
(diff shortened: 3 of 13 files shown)


app/events/domains/provider_domains_changed_event.rb
```diff
@@ -1,9 +1,13 @@
 class Domains::ProviderDomainsChangedEvent < BaseEventStoreEvent
-  def self.create(provider)
+  def self.create(provider, parent_event = nil)
     new(
+      parent_event_id: parent_event&.event_id,
```
Martouta: I dunno if it is on purpose (I guess not) but when there is no `parent_event`, this event will never be saved either... for this:
https://github.com/3scale/porta/blob/43383d8fe110b8ce25954a1748eb12b589966ace/app/services/notification_delivery_service.rb#L71-L73

mikz: But that is for notifications only. This is not a notification and the test proves the event was created.

Martouta: > But that is for notifications only. This is not a notification.

Right, sorry, I see that you are right that it doesn't affect to this kind of events 😄 Although interestingly it seems to affect to `AccountDeletedEvent` although it is not a notification 🤔 but that last detail is an off-topic 😄 

> the test proves the event was created
The current test is not testing that it is persisted when this data is `nil` 😄 You are doing this (this is an example but all tests are like this):
```ruby
def test_provider_domain_changed_event
    provider = FactoryBot.create(:simple_provider)
    proxy = FactoryBot.create(:simple_proxy)
    proxy.update_attributes!(endpoint: "http://#{provider.admin_domain}")
    event = Domains::ProviderDomainsChangedEvent.create(provider)

    proxy_domains_changed = EventStore::Repository.adapter.where(event_type: 'Domains::ProxyDomainsChangedEvent')

    assert_difference proxy_domains_changed.method(:count) do
      ProcessDomainEventsWorker.new.perform(event)
    end
  end
```
But `event = Domains::ProviderDomainsChangedEvent.create(provider)` is not persisting, it is just being build (you can persist it with `Rails.application.config.event_store.publish_event(event)` and check its persitence for example with `assert EventStore::Event.where(event_id: event.event_id).last`), and `ProcessDomainEventsWorker.new.perform(event)` is testing the persistence but this is not the scenario in which `parent_event` is `nil`, because here parent event is the event you have just build 😄

mikz: You are right 👍 Added those tests into the relevant event tests.

app/models/proxy.rb
```diff
@@ -471,6 +473,15 @@ def sandbox_host
     URI(sandbox_endpoint || set_sandbox_endpoint).host
   end
 
+  def update_domains
+    domains = {
+      staging_domain: self.class.extract_domain(sandbox_endpoint.presence || default_staging_endpoint),
```
guicassolato: Tiny thing. No need to call `presence` again.

db/migrate/20190722114341_add_proxy_domains.rb
```diff
@@ -0,0 +1,20 @@
+class AddProxyDomains < ActiveRecord::Migration
+  def change
+    add_column :proxies, :staging_domain, :string
+    add_column :proxies, :production_domain, :string
+
+    add_index :proxies, [ :staging_domain, :production_domain ]
```
Martouta: The migrations are not done for `postgres` and `oracle` schemas. I tried to do it locally and saw this error (this is a blocking migration):
image

Thankfully it suggests a way:
```ruby
add_index :proxies, [:staging_domain, :production_domain], algorithm: :concurrently
```

mikz: But that works for mysql too? Honestly I have no idea how you handle those.

Martouta: buff, what it suggests is not that easy:
image

image

I guess it should be done in more than 1 migration and the add index should be a migration of its own 🤔

Martouta: > But that works for mysql too?

No idea, we gotta try 🤷‍♀ I am trying one that works for all yet 😂

Martouta: image

image

So I will add a condition for the kind of DB 🤔

app/workers/process_domain_events_worker.rb
```diff
@@ -0,0 +1,49 @@
+class ProcessDomainEventsWorker
+  include Sidekiq::Worker
+
+  class_attribute :publisher
+  self.publisher = ->(*args) { Rails.application.config.event_store.publish_event(*args) }
```
guicassolato: We are starting to to have this in many places and it's already refined in https://github.com/3scale/porta/blob/af47f5f0f34b856571eb0c3a4227ba5f17ed7849/app/events/base_event_store_event.rb#L3. Perhaps we could use it from there

db/migrate/20190722114341_add_proxy_domains.rb
```diff
@@ -0,0 +1,20 @@
...
+
+    reversible do |dir|
+      dir.up do
+        Proxy.reset_column_information
+
+        Proxy.select(:id, :sandbox_endpoint, :endpoint).find_in_batches(batch_size: 1000) do |batch|
```
guicassolato: If we don't need to address the batch for anything then we can simplify with `find_each`.

app/models/proxy.rb
mikz: When again? `sandbox_endpoint` can be an emptu string no?

app/workers/process_domain_events_worker.rb
mikz: Yes, but this is not the time I think.

db/migrate/20190722114341_add_proxy_domains.rb
mikz: Indeed. Use github suggested changes for it? :-D I'm wrapping up other things.

app/workers/process_domain_events_worker.rb
Martouta: I agree that it is not the time 😄 Btw I am doing a PR for refactoring these things in https://github.com/3scale/porta/pull/1023, so I can do it there 👍 but let's wait for after CR1 😄

app/models/proxy.rb
guicassolato: You are right. Sorry. I was thinking about refactoring `extract_domain` and left that part of the thought only inside my head. No need to change anything.

db/migrate/20190722114341_add_proxy_domains.rb
Martouta: > Use github suggested changes for it

I think that way the commits are not signed 🤔

mikz: Merge commits are not signed either then.

guicassolato: I think GH suggested changes doesn't work for multi-line changes.

Martouta: Anyway someone will have to work on this branch locally at some point and commit and push for the migrations 😄

app/workers/process_domain_events_worker.rb
guicassolato: How is that so hard? Define in `BaseEventStoreEvent`

```ruby
def publish
  PUBLISHER.call(self)
end
```

Delete 2 lines of code here (4 and 5) and then

https://github.com/3scale/porta/blob/c62569f715823e343fc729c8a577e1958497b2b7/app/workers/process_domain_events_worker.rb#L11
becomes
```ruby
events.each(&:publish)
```

guicassolato: No need to change it here if we are in such a hurry with the PR but it's a quite simpler change than what we expect to achieve with https://github.com/3scale/porta/pull/1023 and also about not adding more TODOs that we never address.

Martouta:
The failure in postgres cukes is an unrelated random failure.
I'm going to merge and I will create a Jira issue for that bug.

PR #1019, opened 2019-07-23
Threescale 3057 disable pf inline styles injection

damianpm:

PF4/React by default injects css inline, it pollutes our markup and is difficult to find errors happening in our CI

https://issues.jboss.org/browse/THREESCALE-3057

**Verification steps** 
`rm -rf node_modules`
`npm install`


damianpm:
@didierofrivia Maybe you can help me
I've followed the steps described [here](https://github.com/patternfly/patternfly-react-seed/wiki/Disable-Patternfly-React-Auto-Style-Injection)
But I still see a lot of <styles> tags in the header
Am I missing something, maybe in the webpacker config ?

damianpm:
> @didierofrivia Maybe you can help me
> I've followed the steps described [here](https://github.com/patternfly/patternfly-react-seed/wiki/Disable-Patternfly-React-Auto-Style-Injection)
> But I still see a lot of tags in the header
> ~Am I missing something, maybe in the webpacker config ?~

UPDATE:
I talked with PF Team, and there are some styles that can't be moved from now because of an external dependency (tipsy). They are working on that and will be fixed in a future release,
Anyway, this PR reduces the amount of styles injected, I think we can move forward

damianpm:
Error messages from server were missing in Signup Page, I added them in last commit:

errors-list

config/webpack/environment.js
```diff
@@ -25,4 +25,11 @@ environment.loaders.append('style', {
   ]
 })
 
+const nullLoader = {
+  test: /\.css$/,
+  include: stylesheet => stylesheet.indexOf('@patternfly/react-styles/css/') > -1,
+  use: ['null-loader']
+}
+environment.loaders.insert('null', nullLoader, {before: 'style'})
```
josemigallas: Instead of `insert` before 'styles' you could call `append` before appending the styles loader.

Also the styles loader is declared in 1 line, I suggest we do the same with `nullLoader` so that the file looks consistent.

damianpm:
> We should add them in Rails' assets pipeline like so that we have actual control of what we want and what we don't want to include. 

First, we need to agree if we are moving all assets to Rails pipeline or if we keep some assets in the webpacker scope.

Then, if we decide to move everything to Rails we need to decide if we:
1. Copy all PF4 assets to rails assets directory, maybe in `assets/vendor`
2. Import directly from `node_modules`
 2.1 As rails doesn't include `node_modules` in the assets pipeline, we may need to add an additional package like [node-sass-package-importer](https://www.npmjs.com/package/node-sass-package-importer) or [node-sass-import](https://npmjs.com/package/node-sass-package-importer).

WDYT @didierofrivia @josemigallas

didierofrivia:
> I don't think all those stylesheets should be added by default in pf4BaseStyles.js. Those belong to individual components that we don't use in some cases and are messing with the main.css.
> We should add them in Rails' assets pipeline like so that we have actual control of what we want and what we don't want to include. Generally speaking we should be very careful when adding styles to avoid unexpected changes.

We should not involve the assets pipeline, but we should definitely be precise and use only the modules required, not everything.

Webpack should take care of all the node assets, you could also use the webpacker `stylesheet_pack_tag` helper with a customized bundle to selectively include them in the views.

damianpm:
> > I don't think all those stylesheets should be added by default in pf4BaseStyles.js. Those belong to individual components that we don't use in some cases and are messing with the main.css.
> > We should add them in Rails' assets pipeline like so that we have actual control of what we want and what we don't want to include. Generally speaking we should be very careful when adding styles to avoid unexpected changes.
> 
> We should not involve the assets pipeline, but we should definitely be precise and use only the modules required, not everything.
> 
> Webpack should take care of all the node assets, you could also use the webpacker `stylesheet_pack_tag` helper with a customized bundle to selectively include them in the views.

Agree, let's do webpacker deal with node_modules

app/javascript/src/patternflyStyles/pf4Base.css
```diff
@@ -0,0 +1,6 @@
+@import '~@patternfly/patternfly/patternfly-variables.css';
```
damianpm: This file contains the basic PF4 styles, it can be imported in any pf4 component as starting point

app/javascript/src/patternflyStyles/loginPage.css
```diff
@@ -0,0 +1,6 @@
+@import '~@patternfly/patternfly/components/Login/login.css';
```
damianpm: This file contains the imports specific to components in the Login Pages (Login, request password, signup)

app/views/layouts/provider/login.html.slim
```diff
@@ -7,7 +7,7 @@ html lang="en"
 
     = render 'provider/analytics'
     = stylesheet_link_tag 'loginPageFonts'
-    = javascript_pack_tag 'pf4BaseStyles'
+    = javascript_pack_tag 'loginPageStyles'
```
damianpm: @didierofrivia I've found that `using stylesheet_pack_tag to point at a CSS file is discouraged, as "packs" should only be JS files` ([here](https://github.com/rails/webpacker/issues/1575))
I think we can use the js file as entry point, and from there import the css files needed

config/webpack/environment.js
damianpm: @josemigallas done!

PR #1023, opened 2019-07-24
create_and_publish returns the event

Martouta:
Otherwise it is counterintuitive and problematic, and this way it simplifies the code and it is also easier for any newcomer.

Outside the tests, it is used in the following places, all of them will keep working after this because the only expect a truthy value if it worked and a falsy value and if it didn't:

https://github.com/3scale/porta/blob/af47f5f0f34b856571eb0c3a4227ba5f17ed7849/app/models/contract/states.rb#L63-L65

https://github.com/3scale/porta/blob/af47f5f0f34b856571eb0c3a4227ba5f17ed7849/app/models/service_token.rb#L15-L17

https://github.com/3scale/porta/blob/af47f5f0f34b856571eb0c3a4227ba5f17ed7849/app/models/proxy.rb#L259-L263

https://github.com/3scale/porta/blob/af47f5f0f34b856571eb0c3a4227ba5f17ed7849/app/models/cinstance.rb#L269-L271

https://github.com/3scale/porta/blob/af47f5f0f34b856571eb0c3a4227ba5f17ed7849/app/models/account/states.rb#L146-L149

https://github.com/3scale/porta/blob/af47f5f0f34b856571eb0c3a4227ba5f17ed7849/app/models/application_key.rb#L130-L132

https://github.com/3scale/porta/blob/af47f5f0f34b856571eb0c3a4227ba5f17ed7849/app/models/service.rb#L212-L215

https://github.com/3scale/porta/blob/af47f5f0f34b856571eb0c3a4227ba5f17ed7849/app/models/service.rb#L558-L564

https://github.com/3scale/porta/blob/af47f5f0f34b856571eb0c3a4227ba5f17ed7849/app/models/service.rb#L605-L607

https://github.com/3scale/porta/blob/af47f5f0f34b856571eb0c3a4227ba5f17ed7849/app/models/user.rb#L417-L419

https://github.com/3scale/porta/blob/af47f5f0f34b856571eb0c3a4227ba5f17ed7849/app/models/account/domains.rb#L81-L84

https://github.com/3scale/porta/blob/af47f5f0f34b856571eb0c3a4227ba5f17ed7849/app/observers/message_observer.rb#L27-L36

### DIFF
--- a/test/events/domains/proxy_domains_changed_event_test.rb
+++ b/test/events/domains/proxy_domains_changed_event_test.rb
@@ -4,19 +4,10 @@ require 'test_helper'
 
 class Domains::ProxyDomainsChangedEventTest < ActiveSupport::TestCase
   test 'deserialises correctly when the Proxy is deleted' do
-    proxy = FactoryBot.create(:proxy)
-    event = Domains::ProxyDomainsChangedEvent.create(proxy)
-    Rails.application.config.event_store.publish_event(event)
+    proxy = FactoryBot.create(:simple_proxy)
+    event = Domains::ProxyDomainsChangedEvent.create_and_publish!(proxy)
 
     proxy.delete
-
-    assert EventStore::Repository.find_event(event.event_id)
-  end
-
-  test 'can be persisted' do
-    proxy = FactoryBot.create(:simple_proxy)
-    event = Domains::ProxyDomainsChangedEvent.create(proxy)
-    Rails.application.config.event_store.publish_event(event)
 
     assert EventStore::Repository.find_event(event.event_id)
   end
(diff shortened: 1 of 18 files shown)


test/events/domains/proxy_domains_changed_event_test.rb
```diff
@@ -5,19 +5,10 @@
...
-    event = Domains::ProxyDomainsChangedEvent.create(proxy)
-    Rails.application.config.event_store.publish_event(event)
+    event = Domains::ProxyDomainsChangedEvent.create_and_publish!(proxy)
 
     proxy.delete
 
     assert EventStore::Repository.find_event(event.event_id)
   end
-
-  test 'can be persisted' do
```
Martouta: It is redundant 😄 the test above already tests the persistence as well 😄

PR #1031, opened 2019-07-26
Replace FakeWeb gem in favor of WebMock

duduribeiro:

FakeWeb is not active anymore and does not have support for ruby 2.4+.

This will change all the usage of FakeWeb and will replace with WebMock

[THREESCALE-2517](https://issues.jboss.org/browse/THREESCALE-2517)


Gemfile.lock
```diff
@@ -861,7 +854,6 @@ DEPENDENCIES
   equivalent-xml
   factory_bot_rails (~> 4.11.1)
   fakefs (~> 0.18.0)
-  fakeweb!
```
Martouta: Good 😄 but these changes should be applied also to the other Gemfile:
https://github.com/3scale/porta/blob/c567ceab6a86dd06f1dbdb34d006107e628c2010/gemfiles/prod/Gemfile.lock#L869

test/remote/payment_gateways_test.rb
```diff
@@ -1,13 +1,12 @@
 require File.expand_path(File.dirname(__FILE__) + '/../test_helper')
```
Martouta: I know that you didn't touch this but would be nice to change it for:
```ruby
require 'test_helper'
```

test/remote/payment_gateways_test.rb
```diff
@@ -1,13 +1,12 @@
...
-require 'fakeweb'
 
 class PaymentGatewaysTest < ActiveSupport::TestCase
   def setup
     @provider_account = FactoryBot.create(:provider_account)
     @provider_account.billing_strategy = FactoryBot.create(:postpaid_billing)
 
     @buyer_account = FactoryBot.create(:buyer_account, :provider_account => @provider_account)
-    FakeWeb.allow_net_connect = true
+    WebMock.allow_net_connect!
```
Martouta: I know that this is respecting the previous behaviour but we shouldn't make real http requests in the tests 😄

Martouta: I'd even say that we are not using this for this test at all 😕 so I'd try removing it and see if the tests pass 😄

PR #1038, opened 2019-07-29
Fix service discovery cluster endpoint for fetching openshift projects

guicassolato:

It fixes Service Discovery's endpoint to connect to the cluster REST API specifically to fetch Openshift projects info.

The old endpoint `https://kubernetes.default.svc/oapi` is no longer an option in Openshift 4: https://github.com/openshift/api/pull/195

Closes [THREESCALE-3114](https://issues.jboss.org/browse/THREESCALE-3114)

### DIFF
--- a/test/unit/service_discovery/cluster_client_test.rb
+++ b/test/unit/service_discovery/cluster_client_test.rb
@@ -164,23 +164,6 @@ module ServiceDiscovery
       assert_equal %w[project-2], cluster.projects_with_discoverables.map(&:name)
     end
 
-    test 'routes' do
-      cluster.stubs(get_routes: [
-        cluster_route(metadata: { name: 'my-api-staging-route',    uid: '220151f0-1918-4eca-808a-d4d07cea4b16', namespace: 'my-namespace' }, spec: { to: { kind: 'Service', name: 'my-api-staging' } }),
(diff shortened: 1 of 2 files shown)


test/unit/service_discovery/cluster_client_test.rb
```diff
@@ -197,5 +180,38 @@ class ClusterClientTest < ActiveSupport::TestCase
...
+      test '.build_client' do
+        assert_equal client_endpoint, 'https://my-cluster.com:8443/api'
+        assert_equal client_endpoint(api_path: 'k8s'), 'https://my-cluster.com:8443/k8s'
+      end
+
+      test '.build_client_for_projects' do
+        assert_equal client_endpoint_for_projects, 'https://my-cluster.com:8443/apis/project.openshift.io'
+        assert_equal client_endpoint_for_projects(api_path_for_projects: 'projs'), 'https://my-cluster.com:8443/projs'
+        assert_equal client_endpoint_for_projects(api_path: 'k8s'), 'https://my-cluster.com:8443/k8s'
+        assert_equal client_endpoint_for_projects(api_path: 'k8s', api_path_for_projects: 'projs'), 'https://my-cluster.com:8443/projs'
```
josemigallas: Is this correct? Wouldn't it be `https://my-cluster.com:8443/k8s/projs`

guicassolato: It is correct. The idea is to show that `api_path_for_projects` takes precedence and overrides `api_path` if both are provided.

guicassolato: Ok. Removed `api_path` from `.build_client_for_projects` for simplicity.

PR #1039, opened 2019-07-29
Do not allow blank plan in service subscription

duduribeiro:

Currently when you try to subscribe in some service, the plan listing is 
blank and when you try to submit it, it explodes an exception and it 
renders a weird 404 page.

This commit validates if the service plan exists

idea2

[THREESCALE-2944](https://issues.jboss.org/browse/THREESCALE-2944)

**Verification steps** 

- Create user and application
- Create at least 2 services 
- Go to Accounts->Listing and choose created account
- Go to Service Subscriptions tab and try to add one available service subscription
- It should not have a blank option in the listing

### DIFF
--- a/app/controllers/buyers/service_contracts_controller.rb
+++ b/app/controllers/buyers/service_contracts_controller.rb
@@ -139,6 +139,6 @@ private
   end
 
   def service_plan(plan_id = params[:service_contract][:plan_id])
-    @service_plan ||= service.service_plans.find(plan_id)
+    @service_plan ||= service.service_plans.find_by(id: plan_id)
   end
 end
--- a/app/views/buyers/service_contracts/new.html.erb
+++ b/app/views/buyers/service_contracts/new.html.erb
@@ -1,8 +1,8 @@
 <h1>New service subscription</h1>
 
 <%= semantic_form_for @service_contract,
-  :url => admin_buyers_account_service_contracts_path(@account, :service_id => @service.id),
-  :html => { :'data-remote' => true } do |form| %>
+  url: admin_buyers_account_service_contracts_path(@account, service_id: @service.id),
+  html: { data: { remote: true } } do |form| %>
 
   <%= form.inputs do %>
 
@@ -11,11 +11,11 @@
       <p><%= h @service.name %></p>
     </li>
 
-    <%= form.input :plan, :collection => @service_plans, :include_blank => !@service_contract.plan_id? %>
+    <!--  It will prompt only if there is no default service plan selected -->
+    <%= form.input :plan, collection: @service_plans, prompt: !@service_contract.plan_id? %>
   <% end %>
 
   <%= form.buttons do %>
-    <%= form.commit_button 'Create subscription', :button_html => {:'data-disable-with' => 'Creating subscription...' } %>
+    <%= form.commit_button 'Create subscription', button_html: { data: { disable_with: 'Creating subscription...' } } %>
   <%-end%>
 <%-end %>
-
--- a/test/integration/buyers/service_contracts_controller_test.rb
+++ b/test/integration/buyers/service_contracts_controller_test.rb
@@ -15,11 +15,10 @@ class Buyers::ServiceContractsControllerTest < ActionDispatch::IntegrationTest
     @buyer1.buy! @application_plan
     @buyer2.buy! @application_plan
 
-    host! @provider.admin_domain
-    provider_login_with @provider.admins.first.username, 'supersecret'
+    login! @provider
   end
 
-  def test_can_unsubscribe_from_service_contract
+  test 'can unsubscribe from service contract' do
     get admin_buyers_account_service_contracts_path(account_id: @buyer1.id)
 
     assert_response :success
@@ -30,7 +29,7 @@ class Buyers::ServiceContractsControllerTest < ActionDispatch::IntegrationTest
     assert page.xpath("//a[@data-method='delete']").text =~ /Unsubscribe/
   end
 
-  def test_unsubscribe_developers_from_service_with_suspended_application
+  test 'unsubscribe developers from service with suspended application' do
     apps = @buyer1.bought_cinstances.by_service_id(@service_contract.service_id)
     apps.update_all state: 'suspended'
 
@@ -47,7 +46,7 @@ class Buyers::ServiceContractsControllerTest < ActionDispatch::IntegrationTest
     assert_equal 1, @buyer2.bought_cinstances.by_service_id(@service_contract.service_id).count
   end
 
-  def test_unsubscribe_developers_from_service_with_one_not_suspended_application
+  test 'unsubscribe developers from service with one not suspended application' do
     apps = @buyer1.bought_cinstances.by_service_id(@service_contract.service_id)
 
     assert_no_difference(-> {apps.count})  do
@@ -61,7 +60,7 @@ class Buyers::ServiceContractsControllerTest < ActionDispatch::IntegrationTest
     end
   end
 
-  def test_unsubscribe_developers_from_service_with_two_no_suspended_applications
+  test 'unsubscribe developers from service with two not suspended applications' do
     application_plan2 = FactoryBot.create(:application_plan, issuer: @service)
 
     @buyer1.buy! application_plan2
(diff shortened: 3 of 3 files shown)


app/views/buyers/service_contracts/new.html.erb
```diff
@@ -11,11 +11,10 @@
       <p><%= h @service.name %></p>
     </li>
 
-    <%= form.input :plan, :collection => @service_plans, :include_blank => !@service_contract.plan_id? %>
```
thomasmaas: I'm just wondering why we did it like this and if something would break in an edge case if we stop doing it. You don't think so @Martouta ?

Martouta: Right, makes sense 👍 I am checking another PR now but I will check this later 👍

duduribeiro: I thought about having it some edge case. But currently the controller as far as I identified, it doesn't accept a blank service plan.

thomasmaas: so only if no service plan has been marked as default the empty option is rendered. that kind of makes sense to me but would be better to use `prompt: 'Select a plan` instead of include blank

test/functional/buyers/service_contracts_controller_test.rb
```diff
@@ -0,0 +1,27 @@
+require 'test_helper'
+
+class Buyers::ServiceContractsControllerTest < ActionController::TestCase
```
Martouta: These functional tests are deprecated and we intend to change them by integration tests 😄 So these tests should be in:
https://github.com/3scale/porta/blob/11e03b37712d0f73890eae5ddf5aca70bea10626/test/integration/buyers/service_contracts_controller_test.rb#L3

duduribeiro: good to know.. thanks 🍻

duduribeiro:
@hallelujah but it is a validation displaying an error now, after my last change 😄

duduribeiro:
> Instead we should have validation and the controller display an erro

@hallelujah I updated the PR description with the validation

test/test_helpers/form_select_assertions.rb
```diff
@@ -98,6 +98,16 @@ def assert_select_no_button_to(*args)
...
       assert_select_button_to(*args)
     end
+
+    # Matcher to check that a specific element does not exist
+    #
+    # == Example
+    #
+    #  refute_select '#some_id'
+    #
+    def refute_select(selector, args = {})
```
hallelujah: I think this is too generic and does not really strongly assert or refute anything.
If you change the code CSS a bit will still be green so not enough value.
I think tests should be strong enough to protect from mutation.

duduribeiro: @hallelujah any idea? 🤔

duduribeiro: @hallelujah what about:

```
  def refute_select(*args)
    begin
      assert_select(*args)
      return false
    rescue AssertionFailedError
      return true # there is no existing field
    end
  end
end
```

hallelujah: Why refute select?
Can we do assert select instead?
Otherwise I think it is not worth it because anyone can change it from select to equal and it will still pass

duduribeiro: I wanted to test that the select didn't have the `Please select` value

duduribeiro:
@hallelujah I tested now the dropdown texts, wdyt?

test/integration/buyers/service_contracts_controller_test.rb
```diff
@@ -74,10 +74,29 @@ def test_unsubscribe_developers_from_service_with_two_no_suspended_applications
...
-  def test_unauthorized_access_master_on_premises
+  test 'unauthorized access master_on_premises' do
     login! master_account
     ThreeScale.stubs(master_on_premises?: true)
     get admin_buyers_account_service_contracts_path(account_id: @provider.id)
     assert_response :forbidden
   end
+
+  test 'renders a prompt if there is no default service plan' do
+    login_provider @provider
```
Martouta: This shouldn't be necessary because in the setup there is this: https://github.com/3scale/porta/blob/e13572517c900d6f0abe320918d599a62f2d1afb/test/integration/buyers/service_contracts_controller_test.rb#L19
in fact we can change [these 2 lines](https://github.com/3scale/porta/blob/e13572517c900d6f0abe320918d599a62f2d1afb/test/integration/buyers/service_contracts_controller_test.rb#L18-L19) of the setup for just `login! provider`

test/integration/buyers/service_contracts_controller_test.rb
```diff
@@ -74,10 +74,29 @@ def test_unsubscribe_developers_from_service_with_two_no_suspended_applications
...
+      
+    get new_admin_buyers_account_service_contract_path(account_id: @buyer1.id, service_id: @service.id)
+
+    page = Nokogiri::HTML::Document.parse(response.body)
+    assert page.xpath("//select[@id='service_contract_plan_id']/option").map(&:text).include?('Please select')
+  end
+
+  test 'does not render a prompt if there is a default service plan' do
+    @service.update_attributes(default_service_plan: @service_plan)    
+    login_provider @provider
```
Martouta: same as https://github.com/3scale/porta/pull/1039/files#r309387734

test/integration/buyers/service_contracts_controller_test.rb
duduribeiro: 🥇

test/integration/buyers/service_contracts_controller_test.rb
duduribeiro: fixed! thanks so much

app/views/buyers/service_contracts/new.html.erb
```diff
@@ -1,8 +1,8 @@
 <h1>New service subscription</h1>
 
 <%= semantic_form_for @service_contract,
-  :url => admin_buyers_account_service_contracts_path(@account, :service_id => @service.id),
-  :html => { :'data-remote' => true } do |form| %>
+  url: admin_buyers_account_service_contracts_path(@account, service_id: @service.id),
+  html: { 'data-remote': true } do |form| %>
```
hallelujah: FYI you can write `html: {data: {remote: true}}` as well
image

app/views/buyers/service_contracts/new.html.erb
hallelujah: As I understand:

- it is on creation of service contract
- if there is already a plan pre-selected (I guess passed on parameters on the `new` action):
  *  We select this plan

- if there is no plan pre-selected:
  *  the default service plan should be selected as default if it exists
  * If there is no default service plan, then render a blank option


So I think to make it equal in every situation, we just `include_blank: true` and make sure that the above statements are right

app/views/buyers/service_contracts/new.html.erb
```diff
@@ -11,11 +11,11 @@
...
     </li>
 
-    <%= form.input :plan, :collection => @service_plans, :include_blank => !@service_contract.plan_id? %>
+    <!--  It will prompt only if there is no default service plan selected -->
+    <%= form.input :plan, collection: @service_plans, prompt: !@service_contract.plan_id? %>
   <% end %>
 
   <%= form.buttons do %>
-    <%= form.commit_button 'Create subscription', :button_html => {:'data-disable-with' => 'Creating subscription...' } %>
+    <%= form.commit_button 'Create subscription', button_html: { 'data-disable-with': 'Creating subscription...' } %>
```
hallelujah: Same here you can write it in a single data hash (though not mandatory change)

app/views/buyers/service_contracts/new.html.erb
duduribeiro: @hallelujah the problem is that `include_blank: true` will always render the blank option, which is not the desired behave. We wanted to render the blank option only if there is no default service plan. Also, @thomasmaas suggested to change it to `prompt`, to render the message instead of a blank option

app/controllers/buyers/service_contracts_controller.rb
```diff
@@ -139,6 +139,6 @@ def service
   end
 
   def service_plan(plan_id = params[:service_contract][:plan_id])
-    @service_plan ||= service.service_plans.find(plan_id)
+    @service_plan ||= service.service_plans.find_by(id: plan_id)
```
josemigallas: Isn't this the same?

Martouta: @josemigallas `find` raises an error if it doesn't exist and `find_by` doesn't 😄

duduribeiro: @josemigallas no. If not exists `find` explodes with an exception, and `find_by` returns `nil` and it will run the validation.

josemigallas:
It's showing weird scrolls in Chrome
Screen Shot 2019-08-02 at 16 15 16

I think `overflow: hidden` does the trick.

app/controllers/buyers/service_contracts_controller.rb
josemigallas: OK OK don't jump everybody at the same time 😉

duduribeiro: hahaha.. I didn't see @Martouta answer 😂 .. sorry

app/views/buyers/service_contracts/new.html.erb
```diff
@@ -11,11 +11,11 @@
       <p><%= h @service.name %></p>
     </li>
 
-    <%= form.input :plan, :collection => @service_plans, :include_blank => !@service_contract.plan_id? %>
+    <!--  It will prompt only if there is no default service plan selected -->
+    <%= form.input :plan, collection: @service_plans, prompt: !@service_contract.plan_id? %>
```
josemigallas: I have a default service plan selected and the pop up is showing
Screen Shot 2019-08-02 at 16 26 22

duduribeiro: @josemigallas do you have the default plan set? or only have a plan named `default`? 

please check this page and screenshot http://provider-admin.example.com.lvh.me:3000/apiconfig/services/{{ ID }}/application_plans

to remove the `please select`, you need to have a `default plan` marked here

josemigallas: It is checked, yes. The name is just to reassure it hehe

damianpm: I see no "Please select" with default plan
_____
Screenshot 2019-08-02 at 17 02 58
_____

It's true there is a weird scroll bar. But this issue is about disallowing blank plan option, then I'd say it's OK. And we should create a new task to fix the scrolling issue which is happening not only here but in all the modals

PR #1041, opened 2019-07-31
[APIAP] Service with field to act as product

Martouta:
Service with boolean to act as product.

1. Migrate Service act_as_product
2. Service code for acts_as_product? uses the new boolean
3. Rename acts_as_product? -> act_as_product? for Service

### DIFF
--- a/app/lib/backend_api_logic/service_extension.rb
+++ b/app/lib/backend_api_logic/service_extension.rb
@@ -21,6 +21,14 @@ module BackendApiLogic
       config.backend_api
     end
 
+    def act_as_product?
+      act_as_product && account&.provider_can_use?(:api_as_product)
+    end
+
+    def acts_as_traditional_service?
+      !act_as_product?
+    end
+
     private
 
     def create_first_backend_api_config!
--- a/db/migrate/20190731092338_backfill_add_acts_as_product_to_service.rb
+++ b/db/migrate/20190731092338_backfill_add_acts_as_product_to_service.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+class BackfillAddActsAsProductToService < ActiveRecord::Migration
+  disable_ddl_transaction!
+
+  def change
+    Service.find_in_batches(batch_size: 200) do |records|
+      Service.where(id: records.map(&:id)).update_all act_as_product: false
+      sleep(0.5)
+    end
+  end
+end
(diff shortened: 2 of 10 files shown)


app/lib/backend_api_logic/service_extension.rb
```diff
@@ -21,6 +21,14 @@ def find_or_create_first_backend_api!
       config.backend_api
     end
 
+    def acts_as_product?
```
guicassolato: I would rename this to the same name of the column `act_as_product?` so it's not confusing. Plus, when/if we remove the rolling update, it will be as simple deleting this method.

db/migrate/20190731092338_backfill_add_acts_as_product_to_service.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class BackfillAddActsAsProductToService < ActiveRecord::Migration
+  disable_ddl_transaction!
+
+  def change
+    Service.find_in_batches(batch_size: 200) do |records|
+      Service.where(id: records.map(&:id)).update_all act_as_product: false
+      sleep(0.5) if Rails.env.production? || Rails.env.preview?
```
guicassolato: I think we can remove the `if`.

db/migrate/20190731092338_backfill_add_acts_as_product_to_service.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class BackfillAddActsAsProductToService < ActiveRecord::Migration
+  disable_ddl_transaction!
+
+  def change
+    Service.find_in_batches(batch_size: 200) do |records|
+      Service.where(id: records.map(&:id)).update_all act_as_product: false
```
guicassolato: Let's make sure it works in Oracle as well. `update_all` doesn't go through ActiveRecord’s type-casting so the value must be supported as-is in the final SQL command.

Martouta: Ok, I will check 👍

Martouta: It probably should because I ran it locally (with a couple of services in the DB) and it didn't raise any error... but I will check the final value, just in case 😄

Martouta: It works for Oracle 💪 
image

PR #1043, opened 2019-08-01
[ProcessDomainEventsWorker] use perform instead of enqueue

macejmic:


test/workers/process_domain_events_worker_test.rb
```diff
@@ -1,6 +1,13 @@
 require 'test_helper'
 
 class ProcessDomainEventsWorkerTest < ActiveSupport::TestCase
+
+  def test_enqueue
+    proxy = FactoryBot.create(:simple_proxy)
+    event = Domains::ProxyDomainsChangedEvent.create(proxy)
+    ProcessDomainEventsWorker.enqueue(event)
```
Martouta: This isn't testing anything other that it doesn't raise an error 😕

macejmic: and that's the test :)

PR #1044, opened 2019-08-01
Move FilterArguments to app/lib

duduribeiro:

When we change some code in development and we don't restart the
app it explodes with `A copy of ThreeScale::SidekiqLoggingMiddleware
has been removed from the module tree but is still active!`.

This happens because the ThreeScale::SidekiqLoggingMiddleware class
uses an class that stays inside lib folder.

This commit moves it to app/lib.


hallelujah:
@josemigallas I do not understand what you are talking about. Which sidekiq screen?

@duduribeiro The sidekiq middleware is pre-loaded https://github.com/3scale/porta/blob/3f878caa5e76ffe3616f836244396acf4239d2e9/config/initializers/sidekiq.rb#L40

I think leaving this file in the `lib` directory is the best choice for now but we should require it at the top of the file.
Though that does not really solve the current issue until we get rid of https://github.com/3scale/porta/blob/3f878caa5e76ffe3616f836244396acf4239d2e9/config/application.rb#L77

PR #1047, opened 2019-08-02
Migrate Triggers for tables of Backend Api

Martouta:
Migration of triggers for tables of backend api

### DIFF
--- a/db/migrate/20190802133303_create_triggers_backend_apis.rb
+++ b/db/migrate/20190802133303_create_triggers_backend_apis.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+class CreateTriggersBackendApis < ActiveRecord::Migration
+  # At this point it does not matter if it blocks these tables or not because they are not being used yet anyway
+
+  def up
+    self.class.execute_trigger_action(:recreate)
+  end
+
+  def down
+    self.class.execute_trigger_action(:drop)
+  end
+
+  def self.execute_trigger_action(action)
+    backend_api_triggers = System::Database.triggers.select { |trigger| trigger.table =~ /\Abackend_api((s)|(_configs))\z/ }
+    BackendApi.transaction do
+      backend_api_triggers.each do |trigger|
+        methods = [trigger.public_send(action)].flatten # [drop] or [drop, create]
+        methods.each(&ActiveRecord::Base.connection.method(:execute))
+      end
+    end
+  end
+
+end
--- a/db/oracle_schema.rb
+++ b/db/oracle_schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20190722114341) do
+ActiveRecord::Schema.define(version: 20190802133303) do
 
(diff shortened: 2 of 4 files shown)


db/migrate/20190802133303_create_triggers_backend_apis.rb
```diff
@@ -0,0 +1,204 @@
...
+         END IF;
+        END;
+      SQL
+    end
+  end
+
+  def self.create_trigger_backend_apis_mysql
+    safety_assured do
+      execute <<~SQL.strip
+        CREATE TRIGGER backend_apis_tenant_id BEFORE INSERT
```
guicassolato: We should be using https://github.com/3scale/porta/blob/master/lib/system/database/trigger.rb#L5 and https://github.com/3scale/porta/blob/master/lib/system/database/mysql/trigger.rb#L6 for this.

db/migrate/20190802133303_create_triggers_backend_apis.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  def down
+    self.class.execute_trigger_actions do |trigger|
+      ActiveRecord::Base.connection.execute(trigger.drop)
+    end
+  end
+
+  def self.execute_trigger_actions
+    BackendApi.transaction do
+      System::Database.triggers.each do |trigger|
```
guicassolato: It could also be

```ruby
triggers = System::Database.triggers.select { |trigger| trigger.table =~ /\Abackend_api((s)|(_configs))\z/ }
triggers.each(&block)
```

IDK if it helps in terms of readability.

db/migrate/20190802133303_create_triggers_backend_apis.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+class CreateTriggersBackendApis < ActiveRecord::Migration
+  # At this point it does not matter if it blocks these tables or not because they are not being used yet anyway
+
+  def up
+    self.class.execute_trigger_actions do |trigger|
+      connection = ActiveRecord::Base.connection
+      connection.execute(trigger.drop)
+      connection.execute(trigger.create)
```
guicassolato: You can use https://github.com/3scale/porta/blob/00ce6011f14d7b1625b58f917b8afc82be259f36/lib/system/database/trigger.rb#L24-L26

db/migrate/20190802133303_create_triggers_backend_apis.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  def down
+    self.class.execute_trigger_action(:drop)
+  end
+
+  def self.execute_trigger_action(action)
+    backend_api_triggers = System::Database.triggers.select { |trigger| trigger.table =~ /\Abackend_api((s)|(_configs))\z/ }
+    BackendApi.transaction do
+      backend_api_triggers.each { |trigger| trigger.public_send(action) }
```
guicassolato:
```suggestion
      backend_api_triggers.each do |trigger|
        trigger.public_send(action).each(ActiveRecord::Base.connection.method(:execute))
      end
```

PR #1049, opened 2019-08-02
Move some lib folders to app/lib - Part1

duduribeiro:


app/models/account.rb
```diff
@@ -58,8 +54,7 @@ def self.columns
   include CreditCard
   include Gateway
   include States
-  require_dependency 'account/domains'
-  include Domains
+  include Domain
```
Martouta: I am not sure if this change is appropriate or not because it represents the combination of `domain` + `self_domain` for a provider and not only one of them, so I dunno... What do you guys think @3scale/system ?

duduribeiro: agree with you. the only problem is:

If we don't rename it, it clashes with `app/events/domains/*`, because there is a class Domains:: definition there.

hallelujah: Then TenantDomains :)

hallelujah:
I would really want to make it per logical files.

app/models/account.rb
duduribeiro: good idea!

duduribeiro:
@hallelujah will split it into more PR's

app/models/account.rb
Martouta: Imo it is more accurate `ProviderDomains` because this is for `tenants` + `master`, but as I told to Cadu before: I trust you all on this :+1:

duduribeiro:
we decided to split this PR in multiples PR's

PR #1053, opened 2019-08-05
[apiap] Polymorphic owner for Metric

guicassolato:
This is step 1 of migrating metrics from the services to the backend APIs. Next step will be removing `service_id` and using only the new polymorphic owner.

Closes [THREESCALE-3158](https://issues.jboss.org/browse/THREESCALE-3158)

It contains commits of https://github.com/3scale/porta/pull/1059

### DIFF
--- a/db/migrate/20190805135829_backfill_metric_owners.rb
+++ b/db/migrate/20190805135829_backfill_metric_owners.rb
@@ -1,0 +1,10 @@
+class BackfillMetricOwners < ActiveRecord::Migration
+  def up
+    say_with_time 'Updating metric owners...' do
+      Metric.reset_column_information
+      Metric.find_each do |metric|
+        metric.update_columns(owner_id: metric.service_id, owner_type: 'Service') unless metric.owner_type?
+      end
+    end
+  end
+end
--- a/db/oracle_schema.rb
+++ b/db/oracle_schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20190802133303) do
+ActiveRecord::Schema.define(version: 20190805135829) do
 
   create_table "access_tokens", force: :cascade do |t|
     t.integer "owner_id",   precision: 38,                  null: false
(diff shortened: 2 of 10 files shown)


db/migrate/20190805135829_backfill_metric_owners.rb
```diff
@@ -0,0 +1,10 @@
+class BackfillMetricOwners < ActiveRecord::Migration
+  def up
+    say_with_time 'Updating metric owners...' do
+      Metric.reset_column_information
```
hallelujah: Yes as the previous migration will be run at the same time as this one, we need to reset the column information of the model otherwise it will not know about the newly created columns

db/schema.rb
```diff
@@ -1155,8 +1158,11 @@
     t.text     "redirect_url",       limit: 65535
     t.integer  "position",           limit: 4
     t.boolean  "last",                             default: false
+    t.integer  "owner_id",           limit: 8
```
hallelujah: This does not belong to this PR

db/schema.rb
```diff
@@ -1228,6 +1234,7 @@
     t.boolean  "referrer_filters_required",                    default: false
     t.string   "deployment_option",              limit: 255,   default: "hosted"
     t.string   "kubernetes_service_link",        limit: 255
+    t.boolean  "act_as_product",                               default: false
```
hallelujah: This does not belong to this PR

test/unit/metric_test.rb
```diff
@@ -72,6 +72,21 @@ def test_destroyable?
...
+    service = FactoryBot.create(:service, account: provider)
+    service_metric = FactoryBot.build(:metric, service: service)
+    refute service_metric.owner
+    assert service_metric.valid?
+    assert_equal service, service_metric.owner
+
+    backend_api = BackendApi.create(name: 'API', system_name: 'api', account: provider)
+    backend_metric = FactoryBot.build(:metric, owner: backend_api)
+    assert_equal backend_api, backend_metric.owner
+    assert service_metric.valid?
```
guicassolato:
```suggestion
    assert backend_metric.valid?
```

db/oracle_schema.rb
```diff
@@ -817,14 +817,17 @@
...
+    t.datetime "created_at",                             null: false
     t.datetime "updated_at"
-    t.integer  "service_id",    precision: 38
+    t.integer  "service_id",              precision: 38
     t.string   "friendly_name"
-    t.integer  "parent_id",     precision: 38
-    t.integer  "tenant_id",     precision: 38
+    t.integer  "parent_id",               precision: 38
+    t.integer  "tenant_id",               precision: 38
+    t.integer  "owner_id",      limit: 8, precision: 8
```
hallelujah: As @Martouta noticed, this should be only`precision: 38`

test/unit/metric_test.rb
```diff
@@ -72,6 +72,21 @@ def test_destroyable?
     assert metric_method.errors[:system_name].present?
   end
 
+  test 'fill owner' do
+    provider = FactoryBot.create(:simple_provider)
+
+    service = FactoryBot.create(:service, account: provider)
```
Martouta: The only difference between `:service` and `:simple_service` is that `:service` uses a `:provider_account` and `:simple_service` uses a `:simple_provider`, so a possible improvement would be removing the previous line (`provider = FactoryBot.create(:simple_provider)`), here using `service = FactoryBot.create(:simple_service)` and in line 84 doing `backend_api = BackendApi.create(name: 'API', system_name: 'api', account: service.provider)`

But this is completely optional 😄 And in fact I don't think you should repush if this is the only change bcz it is not important at all and we don't have much time 😄

PR #1008, opened 2019-07-18
[apiap] UI base

guicassolato:
### What this PR does / why we need it

It implements the first UI changes preparing for API as a Product.
- [x] Application Plan > edit > Methods & Metrics
- [x] Configuration > "Integration page" => List of Backend APIs + Hiding Mapping Rules
- [x] New vertical navigation sidebar for Backend API sections
- [x] New controllers for Backend APIs
- [x] Backend API > List (missing link to navigate to it)
- [x] Backend API > Overview
- [x] Backend API > Methods & Metrics
- [x] Backend API > Mapping Rules
- [x] Fix existing tests
- [x] Write ~cukes~ tests for the new pages

### Special notes for your reviewer
- ~It contains commits of https://github.com/3scale/porta/pull/1003. Let's get that one merged first.~ [MERGED]
- It depends on https://github.com/3scale/porta/pull/1052 for the mapping rules under the backend APIs to make sense
- It depends on https://github.com/3scale/porta/pull/1053 for the metrics under the backend APIs to make sense

A series of other UI changes are to follow this PR, including:
- Add “Products/Backends” view to the _Dashboard_ ([THREESCALE-2374](https://issues.jboss.org/browse/THREESCALE-2374))
- Add selection of a Backend API to the context menu ([THREESCALE-3086](https://issues.jboss.org/browse/THREESCALE-3086))
- Implement or Hide _Backend Overview > edit_ ([THREESCALE-3088](https://issues.jboss.org/browse/THREESCALE-3088))
- Create proper new/edit pages for the metrics of a Backend API (instead of reusing anything from the service) ([THREESCALE-3089](https://issues.jboss.org/browse/THREESCALE-3089))
- Continue _Backend API > Mapping Rules_ (with pagination in Ruby) ([THREESCALE-3090](https://issues.jboss.org/browse/THREESCALE-3090) and [THREESCALE-3091](https://issues.jboss.org/browse/THREESCALE-3091))
- Fix the style of the header of each Backend API in _Application plan > edit > Methods & Metrics_ ([THREESCALE-3092](https://issues.jboss.org/browse/THREESCALE-3092))
- Make the table of metrics and methods of a Backend API expandable/collapsible in _Application plan > edit > Methods & Metrics_ ([THREESCALE-3093](https://issues.jboss.org/browse/THREESCALE-3093))
- Add option to filter the _Analytics_ by Backend API ([THREESCALE-3094](https://issues.jboss.org/browse/THREESCALE-3094))
[...]

PR #1054, opened 2019-08-06
Create an independent page to manage Mapping Rules

duduribeiro:

Our current page that manages the mapping rules don't have
pagination, causing a timeout if you have a lot of rules.

This commit creates a new independent page to manage Mapping rules
with a paginated list of itens.

[THREESCALE-3002](https://issues.jboss.org/browse/THREESCALE-3002)

independent_mapping_rules

### DIFF
--- a/app/controllers/api/proxy_rules_controller.rb
+++ b/app/controllers/api/proxy_rules_controller.rb
@@ -1,0 +1,70 @@
+class Api::ProxyRulesController < Api::BaseController
+  include ThreeScale::Search::Helpers
+
+  before_action :authorize!, :find_service, :find_proxy
+  before_action :find_proxy_rule, only: %i[edit update destroy]
+
+  activate_menu :serviceadmin, :integration, :mapping_rules
+
+  sublayout 'api/service'
+
+  def index
+    @proxy_rules = @proxy.proxy_rules.order_by(params[:sort], params[:direction])
+                                     .includes(:metric).ordered.paginate(page: params[:page])
+  end
+
+  def new
+    last_position = @proxy.proxy_rules.maximum(:position) || 0
+    next_position = last_position + 1
+    @proxy_rule = @proxy.proxy_rules.build(position: next_position, delta: 1)
+  end
+
+  def create
+    @proxy_rule = @proxy.proxy_rules.build(proxy_rule_params)
+    if @proxy_rule.save
+      redirect_to admin_service_proxy_rules_path(@service), notice: 'Mapping rule was created.'
+    else
+      render :new
+    end
+  end
+
+  def update
+    if @proxy_rule.update_attributes(proxy_rule_params)
+      redirect_to admin_service_proxy_rules_path(@service), notice: 'Mapping rule was updated.'
+    else
+      render :edit
+    end
+  end
+
+  def destroy
+    if @proxy_rule.destroy
+      flash[:notice] = 'The mapping rule was deleted'
+    else
+      flash[:error] = 'The mapping rule cannot be deleted'
+    end
+
+    redirect_to admin_service_proxy_rules_path(@service)
+  end
+
+  private
+
+  def authorize!
+    provider_can_use!(:independent_mapping_rules)
+  end
+
+  def find_proxy
+    @proxy = @service.proxy
+  end
+
+  def find_proxy_rule
+    @proxy_rule = @proxy.proxy_rules.find(params[:id])
+  end
+
+  def find_service
+    @service = current_user.accessible_services.find(params[:service_id])
+  end
+
+  def proxy_rule_params
+    params.require(:proxy_rule).permit(%i[http_method pattern delta metric_id position last redirect_url])
+  end
+end
--- a/app/views/api/proxy_rules/index.html.slim
+++ b/app/views/api/proxy_rules/index.html.slim
@@ -1,0 +1,46 @@
+- content_for :sublayout_title, 'Mapping Rules'
+
+table#mapping_rules.data
+  thead
+    tr
+      th = sortable('proxy_rules.http_method', 'Verb')
+      th = sortable('proxy_rules.pattern', 'Pattern')
+      - if @service.using_proxy_pro?
+        th Redirect
+      th title="increment" +
+      th
+        = sortable('metrics.friendly_name', 'Metric or Method')
+        |  (
+        = link_to 'Define', admin_service_metrics_path(@service), title: 'Define Methods & Metrics for this service'
+        | )
+      th = sortable('proxy_rules.last', 'Last?')
+      th = sortable('proxy_rules.position', 'Position')
+      th.actions
+        = link_to 'Add Mapping Rule', new_admin_service_proxy_rule_path(@service), class: 'action add'
+  tbody
+    - @proxy_rules.each do |rule|
+      tr
+        td.http_method
+          = rule.http_method
+        td.pattern
+          = rule.pattern
+          - if rule.pattern == '/'
+            span.fa.fa-exclamation-triangle.disabled#catch-all-warning title=(t('api.integrations.proxy.proxy_rule_catch_all_warning'))
+        - if @service.using_proxy_pro?
+          td.redirect_url
+            = rule.redirect_url
+        td.delta
+          = rule.delta
+        td.metric
+          = rule.metric_system_name
+        td.last_called_and_position
+          = rule.last
+        td.position
+          = rule.position
+        td.actions
+          = link_to '', edit_admin_service_proxy_rule_path(@service, rule.id), class: 'action edit'
+          = link_to '', admin_service_proxy_rule_path(@service, rule.id), class: 'action delete', data: {confirm: 'Are you sure?'}, method: :delete
+
+  tfoot
+
+= will_paginate(@proxy_rules)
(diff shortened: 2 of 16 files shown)


duduribeiro:
image

I couldn't make this link in the same pattern as the collapsible titles. (even copying the same structure from it)

any idea? 🤔

app/controllers/api/proxy_rules_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+  activate_menu :serviceadmin, :integration, :mapping_rules
+
+  sublayout 'api/service'
+
+  def index
+    @proxy_rules = @proxy.proxy_rules.includes(:metric).ordered.paginate(page: params[:page])
+  end
+
+  def new
+    next_position = @proxy.proxy_rules.maximum(:position) || 0
```
mayorova: The variable name `next_position` is a bit misleading, as it's actually `last_position`, and next position would be last + 1.

app/controllers/api/proxy_rules_controller.rb
```diff
@@ -0,0 +1,64 @@
+class Api::ProxyRulesController < Api::BaseController
```
hallelujah: We should authorize this controller based on the rolling update

app/controllers/api/proxy_rules_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+    end
+
+    redirect_to admin_service_proxy_rules_path(@service)
+  end
+
+  private
+
+  def find_service
+    service_id = params[:service_id]
+    @service   = current_user.accessible_services.find(service_id) if service_id
```
hallelujah:
```suggestion
    @service   = current_user.accessible_services.find_by(id: params[:service_id])
```

app/controllers/api/proxy_rules_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+      flash[:error] = 'The mapping rule cannot be deleted'
+    end
+
+    redirect_to admin_service_proxy_rules_path(@service)
+  end
+
+  private
+
+  def find_service
+    service_id = params[:service_id]
```
hallelujah:
```suggestion
```

josemigallas:
I think the _new mapping rule_ button should go in the table, to the right, like in this other page:
Screen Shot 2019-08-06 at 09 13 14

duduribeiro:
@3scale/system updated the code addressing the comments.

duduribeiro:
@josemigallas updated the styles:

image

image

app/controllers/api/proxy_rules_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+  end
+
+  private
+
+  def authorize_rolling_update
+    provider_can_use!(:independent_mapping_rules)
+  end
+
+  def find_proxy
+    @proxy = @service.proxy
```
Martouta: `@service` can be nil and then this would raise an error for sure

duduribeiro: @Martouta I think we should always rely on service.. wdyt on moving the `find_by` from find_service to `find` ?

with this will throw the exception if service does not exists

app/controllers/api/proxy_rules_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+    @proxy_rule = @proxy.proxy_rules.build(proxy_rule_params)
+    if @proxy_rule.save
+      redirect_to admin_service_proxy_rules_path(@service), notice: 'Mapping rule was created.'
+    else
+      render :new
+    end
+  end
+
+  def edit
+    @proxy_rule = @proxy.proxy_rules.find(params[:id])
```
Martouta: `@proxy_rule = @proxy.proxy_rules.find(params[:id])` this should be extracted to a method and then have a `before_action :find_proxy_rule, only: %i[edit update destroy]`
and then you don't need to repeat this line at the beginning of the 3 methods 😄

duduribeiro: 🥇

app/controllers/api/proxy_rules_controller.rb
Martouta: yes, i think it should be `find` and we rely on its existence 😄 

I also think that you should add an integration test for this controller 👍

app/controllers/api/proxy_rules_controller.rb
duduribeiro: @Martouta I did update the code with these suggestions

app/views/api/proxy_rules/index.html.slim
```diff
@@ -0,0 +1,46 @@
...
+          = rule.pattern
+          - if rule.pattern == '/'
+            span.fa.fa-exclamation-triangle.disabled#catch-all-warning title=(t('api.integrations.proxy.proxy_rule_catch_all_warning'))
+        - if @service.using_proxy_pro?
+          td.redirect_url
+            = rule.redirect_url
+        td.delta
+          = rule.delta
+        td.metric
+          = rule.metric.name
```
Martouta: I suggest this instead:
```suggestion
          = rule.metric_system_name
```

For 3 reasons:

1. the Law of Demeter. Let's just make calls to those classes it directly knows of 😄 Better not to make a chain of call from different places and then if we have to change it, we have to change it everywhere the chain is written.

2. `rule.metric.name` if metric is nil for any reason (although it shouldn't, but weirder things have happened 😄 ), it will raise a `NotMethodError`, but with `metric_system_name` we are already covering that case and it will return just nil:
https://github.com/3scale/porta/blob/8aed1303c2193d09b788696c4159c8cb9032770b/app/models/proxy_rule.rb#L92-L94

3. `system_name` and `name` are the same thing for metrics and there is a message in the `metric` model that I find very confusing but sounds to me like better not to use `name`:

https://github.com/3scale/porta/blob/8aed1303c2193d09b788696c4159c8cb9032770b/app/models/metric.rb#L78-L81

https://github.com/3scale/porta/blob/8aed1303c2193d09b788696c4159c8cb9032770b/app/models/metric.rb#L1

duduribeiro: makes total sense! thanks for it

duduribeiro:
@Martouta added the tests and modified the view

can you re-review?

app/controllers/api/proxy_rules_controller.rb
```diff
@@ -0,0 +1,67 @@
+class Api::ProxyRulesController < Api::BaseController
+  before_action :authorize_rolling_update, :find_service, :find_proxy
```
guicassolato:
```suggestion
  before_action :authorize!, :find_service, :find_proxy
```

app/controllers/api/proxy_rules_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+    else
+      flash[:error] = 'The mapping rule cannot be deleted'
+    end
+
+    redirect_to admin_service_proxy_rules_path(@service)
+  end
+
+  private
+
+  def authorize_rolling_update
```
guicassolato:
```suggestion
  def authorize!
```

app/views/shared/provider/navigation/service/_integration.html.slim
```diff
@@ -6,6 +6,10 @@
          title: 'Methods & Metrics',
          path: admin_service_metrics_path(@service)
 
+= render vertical_nav_item,
+         title: 'Mapping Rules',
```
guicassolato: We should add this new menu item in https://github.com/3scale/porta/blob/73c8e8a616b5367c7b4190624afe8cc93814bfc6/features/old/menu/api_menu.feature#L38-L43

Take a look at https://github.com/3scale/porta/pull/1008/files?file-filters%5B%5D=.feature#diff-311ba69d0277694c725639eaad915d08 in case you need to bypass the rolling update.

app/controllers/api/proxy_rules_controller.rb
```diff
@@ -0,0 +1,67 @@
+class Api::ProxyRulesController < Api::BaseController
+  before_action :authorize_rolling_update, :find_service, :find_proxy
+  before_action :find_proxy_rule, only: %i[edit update destroy]
+
+  activate_menu :serviceadmin, :integration, :mapping_rules
+
+  sublayout 'api/service'
+
+  def index
+    @proxy_rules = @proxy.proxy_rules.includes(:metric).ordered.paginate(page: params[:page])
```
guicassolato: We could make it sortable like other index pages. See https://github.com/3scale/porta/pull/1008/commits/fd28f421bf2739bdd13872aad9fcc84334647cf7

duduribeiro: done!

duduribeiro:
@guicassolato updated with the sortable items.. Can you re-review?

app/views/shared/provider/navigation/service/_integration.html.slim
```diff
@@ -6,6 +6,10 @@
          title: 'Methods & Metrics',
          path: admin_service_metrics_path(@service)
 
+= render vertical_nav_item,
```
guicassolato: As we spoke before, let's not forget to hide from the menu if the rolling update is disabled.

guicassolato:
Screen Shot 2019-08-08 at 10 30 04 AM

I think **+ Metric or Method** could be actually 2 fields:
- Metric or Method to increment
- Increment by

PR #1057, opened 2019-08-06
Upgrade Sidekiq to 5.2.7

duduribeiro:

This commit upgrades our version of Sidekiq. Version 5x includes
a better support for code reloading in development.

Sidekiq changed the default concurrency from 25 to 10 (https://github.com/mperham/sidekiq/issues/3892). I don't know if this will affect something in our app.


Martouta:
> Sidekiq changed the default concurrency from 25 to 10

For sure that affects us. I don't know all the implications that it has, but this is one of them:
https://github.com/3scale/porta/blob/8aed1303c2193d09b788696c4159c8cb9032770b/app/workers/zync_worker.rb#L16-L23

This code is relying on being retried up to 25 times, and it changes the retry time to don't be longer than a week in total because otherwise the event is gone.

duduribeiro:
> > Sidekiq changed the default concurrency from 25 to 10
> 
> For sure that affects us. I don't know all the implications that it has, but this is one of them:
> https://github.com/3scale/porta/blob/8aed1303c2193d09b788696c4159c8cb9032770b/app/workers/zync_worker.rb#L16-L23
> 
> This code is relying on being retried up to 25 times, and it changes the retry time to don't be longer than a week in total because otherwise the event is gone.

@Martouta  in this case we should close this PR ou merge it and rely on the servers to run sidekiq with `-c 25` 😄

duduribeiro:
> > > Sidekiq changed the default concurrency from 25 to 10
> > 
> > 
> > For sure that affects us. I don't know all the implications that it has, but this is one of them:
> > https://github.com/3scale/porta/blob/8aed1303c2193d09b788696c4159c8cb9032770b/app/workers/zync_worker.rb#L16-L23
> > 
> > This code is relying on being retried up to 25 times, and it changes the retry time to don't be longer than a week in total because otherwise the event is gone.
> 
> @Martouta in this case we should close this PR ou merge it and rely on the servers to run sidekiq with `-c 25` 😄

@Martouta about the snippet that you send: it is from retries. I think concurrency should not interfere on it AFAIK

Martouta:
hahahahahahah i read it wrong, sorry 🤦‍♀ 🙈 😊 😂

hallelujah:
Closing in favor of #1703

PR #1058, opened 2019-08-07
Dashboard for APIAP

josemigallas:

Implements the Dashboard for APIAP according to the mockups (available in JIRA).

[THREESCALE-2374: APIAP: Show Products and Backends in Dashboard](https://issues.jboss.org/browse/THREESCALE-2374)

**Notes**:
Current approach: download both products and backends and render them hidden. Then, depending on which input is selected (hidden radio button) show the tab.

The tabs should work without JS enabled but its highlighting won't.

⚠️ There's a bug in Firefox: when going back using the back page button, it might show the wrong tab.


app/views/provider/admin/dashboards/_list_backends.html.slim
```diff
@@ -0,0 +1,21 @@
...
+      tr
+        td
+          = link_to backend.name, %"backend_apis/#{backend.id}"
+        td
+          - if backend.services.any?
+            i class='fa fa-check'
+          - else
+            i class='fa fa-times'
+        td
+          = link_to %"#{backend.method_metrics.length} Methods", %"p/admin/backend_apis/#{backend.id}/metrics"
```
guicassolato:
```suggestion
          = link_to pluralize(backend.method_metrics.size, 'Method'), provider_admin_backend_api_metrics_path(backend)
```

app/views/provider/admin/dashboards/_list_backends.html.slim
```diff
@@ -0,0 +1,21 @@
...
+        td
+          = link_to %"#{backend.mapping_rules.length} Mapping Rules", %"p/admin/backend_apis/#{backend.id}/mapping_rules"
```
guicassolato:
```suggestion
          = link_to pluralize(backend.mapping_rules&.size, 'Mapping Rule'), provider_admin_backend_api_mapping_rules_path(backend)
```

app/views/provider/admin/dashboards/_list_backends.html.slim
```diff
@@ -0,0 +1,21 @@
...
+    tr
+      th Name
+      th In Use
+      th Methods
+      th Mapping Rules
+  tbody
+    - current_account.backend_apis.each do |backend|
+      tr
+        td
+          = link_to backend.name, %"backend_apis/#{backend.id}"
```
guicassolato:
```suggestion
          = link_to backend.name, provider_admin_backend_api_path(backend)
```

app/views/provider/admin/dashboards/_apiap_search_bar.html.slim
```diff
@@ -0,0 +1,10 @@
+div id=id
+
+javascript:
+  document.addEventListener('DOMContentLoaded', function () {
+    var apis = #{json collection}
+    var placeholder = #{json placeholder}
+    var displayApis = () => {}
+
+    window.ApiFilter({ apis, placeholder, displayApis }, #{json id})
```
guicassolato: Filter seems not to be working, nor for backends, neither for products.

app/views/provider/admin/dashboards/_apiap_search_bar.html.slim
```diff
@@ -0,0 +1,10 @@
...
+  })
```
guicassolato: You can add to you vscode configs:

```
"files.insertFinalNewline": true
```

and avoid `TrailingBlankLines` cop

josemigallas: boom! thanks

app/views/provider/admin/dashboards/_apiap_navigation.html.slim
```diff
@@ -0,0 +1,27 @@
+nav.DashboardNavigation.DashboardNavigation-tabs
+  ul.DashboardNavigation-list
+    li.DashboardNavigation-list-item
+      = dashboard_apiap_tab_label 'tab-products', 'Product', @services, icon_name: 'puzzle-piece', current_tab: true
+    li.DashboardNavigation-list-item
+      = dashboard_apiap_tab_label 'tab-backends', 'Backend', current_account.backend_apis, icon_name: 'database'
+
+javascript:
```
hallelujah: Can we put that in a file instead of embedding the javascript?

app/helpers/provider/admin/dashboards_helper.rb
```diff
@@ -35,16 +35,27 @@ def dashboard_navigation_link(link_text, path, options = {})
...
   end
-  
+
   def dashboard_secondary_collection_link(singular_name, collection, path, options = {})
     safe_wrap_with_parenthesis(dashboard_collection_link(singular_name, collection, path, options))
   end
 
+  def dashboard_apiap_tab_label(html_for, singular_name, collection, options = {})
+    label_text = pluralize(number_to_human(collection.size), singular_name, options.fetch(:plural, nil))
+    icon_name = options.fetch(:icon_name, nil)
```
hallelujah:
```suggestion
    icon_name = options[:icon_name]
```

app/views/provider/admin/dashboards/_apiap_navigation.html.slim
josemigallas: Would be better as asset or pack?

app/javascript/packs/dashboard.js
```diff
@@ -2,10 +2,12 @@ import { widget as dashboardWidget } from 'Dashboard/index'
...
 import { render as renderChartWidget } from 'Dashboard/chart'
 import { ApiFilterWrapper } from 'Dashboard/components/ApiFilter'
+import { initialize as tabsWidget } from 'Dashboard/tabs-widget'
 
 document.addEventListener('DOMContentLoaded', () => {
   window.dashboardWidget = dashboardWidget
   window.toggleWidget = toggleWidget
   window.renderChartWidget = renderChartWidget
   window.ApiFilter = ApiFilterWrapper
+  window.tabsWidget = tabsWidget
```
didierofrivia: A good time to stop polluting the window and move the init fn here.

app/views/provider/admin/dashboards/_apiap_search_bar.html.slim
```diff
@@ -0,0 +1,20 @@
+div id=id
+
+javascript:
```
didierofrivia: It's better to move this to the pack, and send the api collection, dom class, etc with data props.

app/javascript/src/Dashboard/tabs-widget.jsx
```diff
@@ -0,0 +1,19 @@
+const CURRENT_TAB_CLASS = 'current-tab'
+
+export function initialize () {
```
didierofrivia: Was this approach simpler than recreating the rendered view in React?

app/javascript/src/Dashboard/tabs-widget.jsx
```diff
@@ -0,0 +1,19 @@
...
+  const tabProducts = document.querySelector('label[for="tab-products"]')
+  const tabBackends = document.querySelector('label[for="tab-backends"]')
+
+  const inputProducts = document.querySelector('input#tab-products')
+  const inputBackends = document.querySelector('input#tab-backends')
+
+  inputProducts.addEventListener('change', () => {
```
didierofrivia: Maybe avoiding repetition with something like..
```
[inputProducts, inputBackends].forEach(
	input => input.addEventListener('change', () => [tabProducts, tabBackends].forEach(
	tab => tab.classList.toggle(CURRENT_TAB_CLASS)
	))
)
```

josemigallas: I wanted to avoid using `toggle` because if somehow the same `change` event is triggered twice (it is possible programmatically), the tabs will look wrong. Maybe I shouldn't be concerned about it?

didierofrivia: I wouldn't be too concerned about it, still this is not something important to change.. I'd prioritize the migration of the js in the view to the pack.

app/javascript/src/Dashboard/tabs-widget.jsx
josemigallas: I think so, the dashboard apis section is rather complex and moving it to React would require extra work. IMO right now this is the quicker and even more efficient way.

app/views/provider/admin/dashboards/_apiap_search_bar.html.slim
josemigallas: I don't like the data-props approach in this case since we can end up having huge amount of data in the HTML, say there are 200 apis, the element would be ridiculously big, WDYT?

didierofrivia: Having the JS in the view is something we should avoid, is not testable and prone to bugs. It's not even transpiled. And no worries about the 200 elements, after my PR you could use the decorator and those will be shallow.

app/javascript/packs/dashboard.js
josemigallas: Moved to its own pack, `api-as-product` that is to be merged into `dashboard` once the rolling update is removed.

spec/javascripts/Dashboard/ApiFilter.spec.jsx
```diff
@@ -1,45 +1,35 @@
...
 
 it('should render itself', () => {
   expect(apiFilter.find('.ApiFilter').exists()).toBe(true)
 })
 
-it('should filter APIs passed in props by name', () => {
-  const displayApis = jest.fn()
-  apiFilter = getWrapper(apis, displayApis)
-
+it.skip('should filter APIs passed in props by name', () => {
```
josemigallas: I had to take this out of the equation, since we no longer pass a displayApis function. Can't be spied since it's neither a class component nor it is an exported function.
I can't think of any other option rather than passing the argument or exporting the method, wdyt @didierofrivia ?

features/old/menu/dashboard.feature
```diff
@@ -19,8 +19,15 @@ Feature: Dashboard
...
     When I go to the provider dashboard
     Then I should see "APIs" in the apis dashboard widget
+    And I should see "1 Product" in the apis dashboard widget
+    And I should see "1 Backend" in the apis dashboard widget
     And I should see the link "0 ActiveDocs" in the apis dashboard widget
-    And I should see the link "New API" in the apis dashboard widget
+    And I should see the link "New Backend" in the apis dashboard widget
+    And I should see the link "API Backend API" in the apis dashboard widget
+    # TODO
+    # And I should see the link "0 methods" in the apis dashboard widget
```
guicassolato: We can implement these already, at least methods and mapping rules.

spec/javascripts/Dashboard/ApiFilter.spec.jsx
damianpm: This test, besides being skipped, is doing nothing, I think we can remove it by now.
You can create a new task to improve this test suite later

damianpm: Approving to not block you, but try to create a task for this

PR #1065, opened 2019-08-09
[WIP] [API Endpoints] Backend API CRUD

Martouta:
Closes [THREESCALE-3201](https://issues.jboss.org/browse/THREESCALE-3201)


app/controllers/admin/api/backend_apis_controller.rb
```diff
@@ -0,0 +1,146 @@
...
+  ##~ op.description = "Creates a Backend API."
+  ##~ op.group = "backend_api"
+  #
+  ##~ op.parameters.add @parameter_access_token
+  ##~ op.parameters.add @parameter_backend_api_name
+  ##~ op.parameters.add @parameter_backend_api_description
+  ##~ op.parameters.add @parameter_backend_api_private_endpoint
+  #
+  def create
+    backend_api = current_account.backend_apis.create(backend_api_params)
```
hallelujah: We should also allow `system_name`

app/controllers/admin/api/backend_apis_controller.rb
```diff
@@ -0,0 +1,146 @@
...
+  private
+
+  attr_reader :backend_api
+
+  def authorize
+    provider_can_use!(:api_as_product)
+    # TODO: I guess it should authorize also that it has permission for the specific action :)
+  end
+
+  def find_backend_api
```
hallelujah: There is no need to find it before action
Rename it to `def backend_api` and memoize the variable

app/controllers/admin/api/backend_apis_controller.rb
```diff
@@ -0,0 +1,146 @@
...
+    @backend_api = current_account.backend_apis.find(params[:id])
+  end
+
+  def backend_api_params
+    # Not sure about :system_name but for Service it is readonly # TODO ?
```
hallelujah: Only on create we can allow setting it.

app/representers/backend_api_representer.rb
```diff
@@ -0,0 +1,16 @@
...
+  wraps_resource
+
+  property :id
+  property :name
+  property :system_name
+  property :description
+  property :private_endpoint
+  property :account_id
+  property :created_at
+  property :updated_at
```
hallelujah: We should also add hypermedia to metrics and proxy_rules

app/representers/backend_apis_representer.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class BackendApisRepresenter < ThreeScale::CollectionRepresenter
+  include Roar::JSON::Collection
```
hallelujah: Right we decided no more XML 👍

spec/acceptance/api/backend_api_spec.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+resource 'BackendApi' do
+  let(:resource) { FactoryBot.create(:backend_api) }
```
hallelujah: Do you need to create it? Does build work?

spec/acceptance/api/backend_api_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    it { subject.should have_properties(expected_properties).from(resource) }
+  end
+
+  json(:collection) do
+    let(:root) { 'backend_apis' }
+    context do
+      let(:collection) { [resource, resource] }
+      it 'contains the backend apis data by its representer' do
+        subject.each do |subject_backend_api|
```
hallelujah: But they are no different so if one pass the other too...

Martouta:
To do this: **This ApiDocs for Backend API maybe shouldn't appear for those who don't have the rolling update enabled**
Michal Macejko gave me the advice to look at this: https://github.com/3scale/porta/blob/master/app/controllers/api_docs/services_controller.rb
He said that the logic is in:
https://github.com/3scale/porta/blob/c9643b6dbbe16a44acf14ef318cfd591f5927e5b/app/controllers/api_docs/services_controller.rb#L122

test/integration/admin/api/backend_apis_controller_test.rb
```diff
@@ -0,0 +1,125 @@
...
+require 'test_helper'
+
+class Admin::API::AccountsControllerTest < ActionDispatch::IntegrationTest
+  def setup
+    @provider = FactoryBot.create(:provider_account)
+    host! @provider.admin_domain
+    @access_token_value = FactoryBot.create(:access_token, owner: @provider.admin_users.first!, scopes: %w[account_management], permission: 'rw').value
+    Logic::RollingUpdates.stubs(enabled?: true)
+    Logic::RollingUpdates::Features::ApiAsProduct.any_instance.stubs(:enabled?).returns(true)
+    @backend_api = FactoryBot.create(:backend_api, account: provider)
```
hallelujah: This is been created for each tests (though create for example does not need it)

test/integration/admin/api/backend_apis_controller_test.rb
```diff
@@ -0,0 +1,125 @@
...
+      assert provider.backend_apis.find_by(id: response_backend_api.dig('backend_api', 'id'))
+    end
+  end
+
+  test 'index can be paginated' do
+    FactoryBot.create_list(:backend_api, 5, account: provider)
+    get admin_api_backend_apis_path(access_token: access_token_value, per_page: 3, page: 2)
+    assert_response :success
+    response_backend_api_ids = JSON.parse(response.body)['backend_apis'].map { |response_backend_api| response_backend_api.dig('backend_api', 'id') }
+    assert_same_elements provider.backend_apis.offset(3).limit(3).pluck(:id), response_backend_api_ids
```
hallelujah: Why same_elements? and not equal?
Aren't they ordered?

spec/acceptance/api/backend_api_spec.rb
Martouta: @hallelujah Then it is going to save it anyway for this:
https://github.com/3scale/porta/blob/c9643b6dbbe16a44acf14ef318cfd591f5927e5b/spec/support/api/context.rb#L71

Martouta: And anyway I need the `id` and `created_at` which are only there for `create` and `build_stubbed` but i cannot do the later for that `save!`

hallelujah: Yes that is why I think there is no need to create it and resave it :)

Martouta: Ok, I will put it as `build` then 😄 as long as we keep in mind that it is actually saving for that 😄

app/representers/backend_api_representer.rb
Martouta: I can't yet because those routes do not exist, but I added the comments and I will reference the comments from those jira issues as well.

spec/acceptance/api/backend_api_spec.rb
Martouta: Fixed 👍

app/controllers/admin/api/backend_apis_controller.rb
```diff
@@ -0,0 +1,148 @@
...
+  ##~ op.httpMethod = "DELETE"
+  ##~ op.summary    = "Backend API Delete"
+  ##~ op.description = "Deletes a backend API."
+  ##~ op.group = "backend_api"
+  #
+  ##~ op.parameters.add @parameter_access_token
+  ##~ op.parameters.add @parameter_backend_api_id_by_id
+  #
+  def destroy
+    backend_api.destroy
```
hallelujah: Good for now though we should probably wait for this PR to be merged to implement it correctly.
Wdyt?
https://github.com/3scale/porta/pull/1144

Martouta: Right 👍 I will wait for that PR and then do it correctly here 😉 👍

PR #1068, opened 2019-08-09
API Service Update/Create with Strong Params

Martouta:
API side of [THREESCALE-2715](https://issues.jboss.org/browse/THREESCALE-2715)
Conclusions in [this comment](https://issues.jboss.org/browse/THREESCALE-2715?focusedCommentId=13770328&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13770328), I've been very conservative to be safe and I did not allow the ones that I am sure of, and I explained why in the Jira issue so you can see the explanation of each one of the ones I am rejecting 👍

### DIFF
--- a/app/controllers/admin/api/services_controller.rb
+++ b/app/controllers/admin/api/services_controller.rb
@@ -43,7 +43,7 @@ class Admin::Api::ServicesController < Admin::Api::ServiceBaseController
   ##~ op.parameters.add :name => "deployment_option", :description => "Deployment option for the gateway: 'hosted' for APIcast hosted, 'self_managed' for APIcast Self-managed option", :dataType => "string", :allowMultiple => false, :required => false, :paramType => "query"
   ##~ op.parameters.add :name => "backend_version", :description => "Authentication mode: '1' for API key, '2' for App Id / App Key, 'oidc' for OpenID Connect", :dataType => "string", :allowMultiple => false, :required => false, :paramType => "query"
   ##~ op.parameters.add @parameter_system_name_by_name
-  ##~ op.parameters.add @parameter_extra_short
+  ##~ op.parameters.add :name => " ", :dataType => "custom", :paramType => "query", :allowMultiple => true, :description => "Extra parameters"
   #
   def create
     service = current_account.create_service(service_params)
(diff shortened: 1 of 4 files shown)


test/integration/admin/api/services_controller_test.rb
```diff
@@ -3,69 +3,126 @@
...
+      assert_persisted_right_params
+      assert_equal 'the-system-name', service.system_name
+    end
+
+    test 'update' do
+      put admin_api_service_path(service, access_token: access_token_value, format: :json), permitted_params.merge(forbidden_params).merge({system_name: 'the-system-name'})
+      assert_response :success
+      service.reload
+      assert_persisted_right_params
+      assert_not_equal 'the-system-name', service.system_name
```
Martouta: Gui suggested instead:
`assert_equal 'old-system-name', service.reload.system_name`
I said:
> i think both then... because if the old value is equal to the new one... bad test 😄

test/integration/admin/api/services_controller_test.rb
```diff
@@ -3,69 +3,214 @@
...
+        created_at: 1.day.from_now,
+        updated_at: 1.day.from_now,
+        kubernetes_service_link: '/api/v1/namespaces/example-project/services/example-api',
+        account_id: provider.id + 1,
+        tenant_id:  provider.id + 1,
+        oneline_description: 'one line description',
+        txt_api: 'text for txt api',
+        txt_features: 'text for txt features',
+        draft_name: 'example of draft name',
+        infobar: 'infobar text',
```
hallelujah: Some are present in a drop or view so I'd suppose it can be modified

```ruby
app/controllers/api/terms_controller.rb
17:      redirect_to edit_admin_service_terms_path(@edited_service)

lib/developer_portal/app/views/developer_portal/settings/terms.html.erb
4:  <%= raw textilize(@service.terms) %>
```

```ruby
lib/developer_portal/lib/liquid/drops/service.rb
198:      def infobar
199:        @service.infobar
```
```ruby
lib/developer_portal/app/views/shared/_footer.html.erb
4:    <% if @service.try(:txt_support).present? %>
```

app/controllers/admin/api/services_controller.rb
```diff
@@ -111,7 +113,12 @@ def can_create
   end
 
   def service_params
-    params.fetch(:service)
+    params.require(:service).permit(:name, :system_name, :description, :support_email, :deployment_option, :backend_version,
```
hallelujah: Those are the same as in the UI controller right?
Probably best to create a `ServiceUpdater` service object and use it in API and UI

test/integration/admin/api/services_controller_test.rb
Martouta: here you are mentioning 3 attributes:
- terms
- txt_support
- infobar

Terms and txt_support are being allowed in the strong params, infobar isn't, but read the comment in the drop:
https://github.com/3scale/porta/blob/ad46a58a31aba8889e419c48da369ae2cfed807e/lib/developer_portal/lib/liquid/drops/service.rb#L196-L200
And read my justification of why I am not allowing it:
https://issues.jboss.org/browse/THREESCALE-2715?focusedCommentId=13769613&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13769613

app/controllers/admin/api/services_controller.rb
Martouta: @hallelujah UI in a following PR 😄 And I don't mind doing that refactoring in that one 😄

Martouta: In particular in https://github.com/3scale/porta/pull/1084 😄 But it isn't really started... I am only testing a weird situation there first 😄

PR #1071, opened 2019-08-12
Remove all Jenkins-related code

Martouta:
Because we don't use Jenkins anymore 😄 

The only references left now are from `git` branches (mostly this one 😂 ) and from `npm` modules (a dependency from `patternfly` so it is not related to us and I didn't touch it)
image


Dockerfile
```diff
@@ -35,6 +35,8 @@ ADD config/examples/*.yml config/
 # Needed for Sphinx ODBC
 ADD config/oracle/odbc*.ini /etc/
 
+ENTRYPOINT ["xvfb-run", "--server-args", "-screen 0 1280x1024x24"]
```
duduribeiro: why do we need this? 😄

Martouta: Oooops, error dealing with conflicts, sorry 🙈 , I will remove it 💪

PR #1072, opened 2019-08-12
Share logic between the 2 pages to manage Mapping Rules

duduribeiro:

Now we have 2 pages to manage mapping rules, 1 for backend apis
and another for the service itself, and some code are being
duplicated between both. This commit tries to reuse some logics
and views for both pages.

### DIFF
--- a/app/controllers/api/proxy_rules_controller.rb
+++ b/app/controllers/api/proxy_rules_controller.rb
@@ -2,36 +2,26 @@
 
 class Api::ProxyRulesController < Api::BaseController
   include ThreeScale::Search::Helpers
+  include ProxyRuleSharedController
 
-  before_action :authorize!, :find_service, :find_proxy
-  before_action :find_proxy_rule, only: %i[edit update destroy]
+  before_action :authorize!
+
+  delegate :proxy_rules, to: :proxy
 
   activate_menu :serviceadmin, :integration, :mapping_rules
 
   sublayout 'api/service'
 
-  def index
-    @proxy_rules = @proxy.proxy_rules.order_by(params[:sort], params[:direction])
-                                     .includes(:metric).ordered.paginate(page: params[:page])
-  end
-
-  def new
-    proxy_rules   = @proxy.proxy_rules
-    last_position = proxy_rules.maximum(:position) || 0
-    next_position = last_position + 1
-    @proxy_rule   = proxy_rules.build(position: next_position, delta: 1)
-  end
-
-  def edit; end
-
   def create
-    @proxy_rule = @proxy.proxy_rules.build(proxy_rule_params)
+    @proxy_rule = proxy_rules.build(proxy_rule_params)
     if @proxy_rule.save
       redirect_to admin_service_proxy_rules_path(@service), notice: 'Mapping rule was created.'
     else
       render :new
     end
   end
+
+  def edit; end
 
   def update
     if @proxy_rule.update_attributes(proxy_rule_params)
@@ -57,19 +47,11 @@ class Api::ProxyRulesController < Api::BaseController
     provider_can_use!(:independent_mapping_rules)
   end
 
-  def find_proxy
-    @proxy = @service.proxy
+  def proxy
+    @proxy ||= service.proxy
   end
 
-  def find_proxy_rule
-    @proxy_rule = @proxy.proxy_rules.find(params[:id])
-  end
-
-  def find_service
-    @service = current_user.accessible_services.find(params[:service_id])
-  end
-
-  def proxy_rule_params
-    params.require(:proxy_rule).permit(%i[http_method pattern delta metric_id position last redirect_url])
+  def service
+    @service ||= current_user.accessible_services.find(params[:service_id])
   end
 end
--- a/app/controllers/provider/admin/backend_apis/mapping_rules_controller.rb
+++ b/app/controllers/provider/admin/backend_apis/mapping_rules_controller.rb
@@ -1,14 +1,39 @@
 # frozen_string_literal: true
 
 class Provider::Admin::BackendApis::MappingRulesController < Provider::Admin::BackendApis::BaseController
+  include ThreeScale::Search::Helpers
+  include ProxyRuleSharedController
+
   activate_menu :backend_api, :mapping_rules
 
-  include ThreeScale::Search::Helpers
+  delegate :proxy_rules, to: :@backend_api
 
-  def index
-    @mapping_rules = @backend_api.mapping_rules
-                                 .order_by(params[:sort], params[:direction])
-                                 .includes(:metric)
-                                 .paginate(pagination_params)
+  def create
+    @proxy_rule = proxy_rules.build(proxy_rule_params)
+    if @proxy_rule.save
+      redirect_to provider_admin_backend_api_mapping_rules_path(@backend_api), notice: 'Mapping rule was created.'
+    else
+      render :new
+    end
+  end
+
+  def edit; end
+
+  def update
+    if @proxy_rule.update_attributes(proxy_rule_params)
+      redirect_to provider_admin_backend_api_mapping_rules_path(@backend_api), notice: 'Mapping rule was updated.'
+    else
+      render :edit
+    end
+  end
+
+  def destroy
+    if @proxy_rule.destroy
+      flash[:notice] = 'The mapping rule was deleted'
+    else
+      flash[:error] = 'The mapping rule cannot be deleted'
+    end
+
+    redirect_to provider_admin_backend_api_mapping_rules_path(@backend_api)
   end
 end
--- a/app/controllers/proxy_rule_shared_controller.rb
+++ b/app/controllers/proxy_rule_shared_controller.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+# This module is used to share behavior between the 2 different Mapping rules controller that we have
+# (1 for BackendApi and another one to the Proxy).
+# The controller that uses this class should implement the `proxy_rules` method that returns a relation containing all
+# the mapping rules for the resource.
+module ProxyRuleSharedController
+  def self.included(base)
+    base.class_eval do
+      before_action :find_proxy_rule, only: %i[edit update destroy]
+    end
+  end
+
+  def index
+    @proxy_rules = proxy_rules.order_by(params[:sort], params[:direction])
+                              .includes(:metric)
+                              .paginate(pagination_params)
+  end
+
+  def new
+    last_position = proxy_rules.maximum(:position) || 0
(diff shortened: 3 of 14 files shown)


app/controllers/api/proxy_rules_controller.rb
```diff
@@ -57,19 +46,25 @@ def authorize!
     provider_can_use!(:independent_mapping_rules)
   end
 
-  def find_proxy
-    @proxy = @service.proxy
+  helper_method :destroy_rule_path
+  def destroy_rule_path(service, rule)
```
guicassolato: I think this does not belong to the controller.

Martouta: I think that this method is not necessary at all 😄 This method is just an alias for `admin_service_proxy_rule_path` so we could use that one directly instead.

app/controllers/api/proxy_rules_controller.rb
```diff
@@ -57,19 +46,25 @@ def authorize!
...
+    admin_service_proxy_rule_path(service, rule)
   end
 
-  def find_proxy_rule
-    @proxy_rule = @proxy.proxy_rules.find(params[:id])
+  helper_method :edit_rule_path
+  def edit_rule_path(service, rule)
```
Martouta: Same as https://github.com/3scale/porta/pull/1072/files#r313297934, but let's keep the conversation only there 😄

app/views/provider/admin/backend_apis/mapping_rules/new.html.slim
```diff
@@ -0,0 +1,5 @@
+= content_for :sublayout_title, 'New Mapping Rule'
```
guicassolato: This is not showing up:

Screen Shot 2019-08-13 at 2 39 04 PM

app/views/provider/admin/backend_apis/mapping_rules/new.html.slim
```diff
@@ -0,0 +1,5 @@
...
+
+= semantic_form_for @proxy_rule, url: provider_admin_backend_api_mapping_rules_path(@backend_api) do |form|
+  = render 'shared/proxy_rules/form', form: form, proxy_rule: @proxy_rule, metrics: @backend_api.metrics,
+                                      using_proxy_pro: true
```
guicassolato:
```suggestion
                                      using_proxy_pro: false
```

app/views/provider/admin/backend_apis/mapping_rules/edit.html.slim
```diff
@@ -0,0 +1,5 @@
+= content_for :sublayout_title, 'Edit Mapping Rule'
```
guicassolato: Same here, the title is not showing up

Screen Shot 2019-08-13 at 2 44 50 PM

app/controllers/api/proxy_rules_controller.rb
duduribeiro: @Martouta @guicassolato  the problem is that I wanted to make a different behave from both controllers. if you look in the `_table` partial, I call this method and each controller calls the specific url_path

app/views/provider/admin/backend_apis/mapping_rules/new.html.slim
duduribeiro: @guicassolato what do you think to add an `h2`?

app/views/provider/admin/backend_apis/mapping_rules/new.html.slim
duduribeiro: 🤦‍♂ my bad! thank you!

app/views/shared/proxy_rules/_form.html.slim
```diff
@@ -1,7 +1,8 @@
 = form.inputs do
   = form.input :http_method, as: :select, collection: ProxyRule::ALLOWED_HTTP_METHODS, include_blank: false, label: 'Verb'
   = form.input :pattern
-  = form.input :metric_id, as: :select, collection: service.metrics, include_blank: false, label: 'Metric or Method to increment'
+  = form.input :redirect_url, as: :string if using_proxy_pro
```
guicassolato: It should work with something like...
```suggestion
  = form.input :redirect_url, as: :string if owner.try(:using_proxy_pro?)
```
and we don't even need to set the local variable.

app/controllers/api/proxy_rules_controller.rb
guicassolato: Can we define a helper method (in a normal helper file) and use `controller_name`? Not pretty but better than polluting the controller with view-related stuff.

app/views/provider/admin/backend_apis/mapping_rules/new.html.slim
guicassolato: 🤷‍♂️

app/controllers/api/proxy_rules_controller.rb
guicassolato: I think you can also do it with https://apidock.com/rails/v4.2.9/ActionDispatch/Routing/PolymorphicRoutes/polymorphic_path

guicassolato:
```ruby
polymorphic_path [:admin, proxy_rule.owner, ProxyRule]
```

BTW, it should be `admin_proxy_proxy_rule_path ` and not `admin_service_proxy_rule_path`

guicassolato: Try

```ruby
owner = proxy_rule.owner
parts = case owner
        when Proxy
          [:admin, owner.service]
        when BackendApi
          [:provider, :admin, owner]
        end
polymorphic_path [*parts, :mapping_rule], id: proxy_rule
```

duduribeiro:
@guicassolato @Martouta I did update the code. can you re-review it? thanks so much

app/controllers/provider/admin/backend_apis/mapping_rules_controller.rb
```diff
@@ -1,14 +1,45 @@
...
+
+  private
 
-  def index
-    @mapping_rules = @backend_api.mapping_rules
-                                 .order_by(params[:sort], params[:direction])
-                                 .includes(:metric)
-                                 .paginate(pagination_params)
+  def proxy_rules
+    @backend_api.mapping_rules
```
guicassolato: `BackendApi#mapping_rules` is an alias, you can also
```ruby
delegate :proxy_rules, to: @backend_api
```
if you want

app/controllers/proxy_rule_shared_controller.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+# This module is used to share behave between the 2 different Mapping rules controller that we have
```
guicassolato:
```suggestion
# This module is used to share behavior between the 2 different Mapping rules controller that we have
```

app/controllers/proxy_rule_shared_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+# (1 for BackendApi and another one to the Service).
```
guicassolato:
```suggestion
# (1 for BackendApi and another one to the Proxy).
```

guicassolato: This one is tricky as a `ProxyRule` either belong to a `BackendApi` or to a `Proxy`, but never to a `Service`. Nevertheless the controller is right below service. Confusing.

app/controllers/provider/admin/backend_apis/mapping_rules_controller.rb
```diff
@@ -1,14 +1,45 @@
 # frozen_string_literal: true
 
 class Provider::Admin::BackendApis::MappingRulesController < Provider::Admin::BackendApis::BaseController
+  include ThreeScale::Search::Helpers
+  include ProxyRuleSharedController
+
+  before_action :find_proxy_rule, only: %i[edit update destroy]
```
guicassolato: This could be injected in `included` by `ProxyRuleSharedController`, no?

duduribeiro: makes total sense.. will try if it will works on included

app/helpers/proxy_rules_helper.rb
```diff
@@ -0,0 +1,12 @@
+module ProxyRulesHelper
+  def proxy_rule_path_for(proxy_rule, edit: false)
+    owner = proxy_rule.owner
+    collection = case owner
+                 when Proxy
+                   [:admin, owner.service, :proxy_rule]
+                 when BackendApi
+                   [:provider, :admin, owner, :mapping_rule]
+                 end
+    edit ? edit_polymorphic_path(collection, id: proxy_rule) : polymorphic_path(collection, id: proxy_rule)
```
guicassolato:
```suggestion
    public_send (edit ? :edit_polymorphic_path : :polymorphic_path), collection, id: proxy_rule
```

duduribeiro:
@guicassolato changed the code based on your review. can you re-review? thanks!

app/controllers/proxy_rule_shared_controller.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+# This module is used to share behavior between the 2 different Mapping rules controller that we have
+# (1 for BackendApi and another one to the Proxy).
+# The controller that uses this class should implement the `proxy_rules` method that returns a relation containing all
+# the mapping rules for the resource.
+module ProxyRuleSharedController
+  def self.included(base)
```
guicassolato: You can also `extend ActiveSupport::Concern` here but all right.

app/helpers/proxy_rules_helper.rb
```diff
@@ -0,0 +1,13 @@
+module ProxyRulesHelper
+  def proxy_rule_path_for(proxy_rule, edit: false)
+    owner = proxy_rule.owner
+    collection = case owner
+                 when Proxy
+                   [:admin, owner.service, :proxy_rule]
+                 when BackendApi
+                   [:provider, :admin, owner, :mapping_rule]
+                 end
+    # edit ? edit_polymorphic_path(collection, id: proxy_rule) : polymorphic_path(collection, id: proxy_rule)
```
guicassolato:
```suggestion
```

app/views/api/proxy_rules/index.html.slim
```diff
@@ -1,46 +1,5 @@
-- content_for :sublayout_title, 'Mapping Rules'
+h2 Mapping Rules
```
guicassolato: I still don't get why it doesn't work with
```
content_for :sublayout_title, 'Mapping Rules'
```

PR #1073, opened 2019-08-13
docker-compose.mk back for now

guicassolato:
We are not ready to remove docker-compose.mk. Putting it back until someone can work on https://github.com/3scale/porta/blob/1ffee4935d9ffb1a2bd025c540f7bca7141531a3/openshift/system/Makefile#L26

### DIFF
--- a/docker-compose.mk
+++ b/docker-compose.mk
@@ -1,0 +1,46 @@
+MKFILE_PATH := $(abspath $(lastword $(MAKEFILE_LIST)))
+PROJECT_PATH := $(patsubst %/,%,$(dir $(MKFILE_PATH)))
+
+DOCKER_COMPOSE_VERSION := 1.21.0
+DOCKER_COMPOSE := $(BIN_PATH)/docker-compose
+DOCKER_COMPOSE_BIN := $(DOCKER_COMPOSE)-$(DOCKER_COMPOSE_VERSION)
+
+ifeq ($(CI),true)
+DOCKER_ENV = CI=true
+else
+DOCKER_ENV = CI=jenkins
+endif
+
+DOCKER_ENV += $(foreach env,$(JENKINS_ENV),$(env)=$(value $(env)))
+DOCKER_ENV += GIT_TIMESTAMP=$(shell git log -1 --pretty=format:%ct)
+DOCKER_ENV += PERCY_PROJECT=3scale/porta PERCY_BRANCH=$(subst origin/,,$(GIT_BRANCH)) PERCY_COMMIT=$(GIT_COMMIT)
+DOCKER_ENV += $(RUBY_ENV)
+DOCKER_ENV += BUNDLE_GEMFILE=$(BUNDLE_GEMFILE)
+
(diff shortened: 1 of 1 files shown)


docker-compose.mk
```diff
@@ -0,0 +1,46 @@
...
+PROJECT_PATH := $(patsubst %/,%,$(dir $(MKFILE_PATH)))
+
+DOCKER_COMPOSE_VERSION := 1.21.0
+DOCKER_COMPOSE := $(BIN_PATH)/docker-compose
+DOCKER_COMPOSE_BIN := $(DOCKER_COMPOSE)-$(DOCKER_COMPOSE_VERSION)
+
+ifeq ($(CI),true)
+DOCKER_ENV = CI=true
+else
+DOCKER_ENV = CI=jenkins
```
Martouta: we don't need jenkins anymore 😄

docker-compose.mk
```diff
@@ -0,0 +1,46 @@
...
+endif
+
+DOCKER_ENV += $(foreach env,$(JENKINS_ENV),$(env)=$(value $(env)))
```
Martouta: same as https://github.com/3scale/porta/pull/1073/files#r313332051

docker-compose.mk
guicassolato: I won't fix `docker-compose.mk`, @Martouta. I'm putting it back so it doesn't break `openshift/system/Makefile`.

Martouta: @guicassolato I'm really not a big fan of introducing unnecessary code to master again after having merged https://github.com/3scale/porta/pull/1071 😞 , so may I fix it in another PR pointing to this one then if you don't want to do it yourself ? 🤔 If you disagree, I won't oppose strongly either 👍

duduribeiro:
@guicassolato what is the fail that happens when this is removed?

docker-compose.mk
guicassolato: @Martouta, https://github.com/3scale/porta/pull/908 was incomplete and introduced a bug.  Don't bother much about `docker-compose.mk`. It's `openshift/system/Makefile` that needs to be fixed.

We can even remove https://github.com/3scale/porta/blob/1ffee4935d9ffb1a2bd025c540f7bca7141531a3/openshift/system/Makefile#L26 and delete `docker-compose.mk` again if we fix the targets that depend on it. They are probably broken already anyway.

Martouta: Ok then as @duduribeiro says, we need to create a Jira issue to completely fix all that for real since this PR is just a patch 😄

guicassolato:
@duduribeiro, `openshift/system/Makefile` depends on `docker-compose.mk`. We either should not have removed the file in https://github.com/3scale/porta/pull/908 or fixed this other Makefile together with the one at the root (better). I just need the https://github.com/3scale/porta/blob/1ffee4935d9ffb1a2bd025c540f7bca7141531a3/openshift/system/Makefile#L28 working before we find time to fix this.

docker-compose.mk
guicassolato: https://issues.jboss.org/browse/THREESCALE-3246

PR #1074, opened 2019-08-13
[major_and_minor_version_path] unknown version 

macejmic:

`major_and_minor_version_path` does not work for "upstream" release name

https://issues.jboss.org/browse/THREESCALE-2250

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -341,10 +341,13 @@ module ApplicationHelper
   def major_and_minor_version_path
     full_version = System::Deploy.info.release || '2.0.0'
     simple_version = /^(?:(\d+)\.)?(?:(\d+))/.match(full_version)
+
     if saas?
-      "red_hat_3scale/#{simple_version[1]}-saas"
+      version = simple_version ? simple_version[1] : full_version
+      "red_hat_3scale/#{version}-saas"
     else
-      "red_hat_3scale_api_management/#{simple_version[1]}.#{simple_version[2]}"
+      version = simple_version ? "#{simple_version[1]}.#{simple_version[2]}" ? full_version
+      "red_hat_3scale_api_management/#{version}"
     end
   end
 
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -339,15 +339,12 @@ module ApplicationHelper
   end
 
   def major_and_minor_version_path
-    full_version = System::Deploy.info.release || '2.0.0'
-    simple_version = /^(?:(\d+)\.)?(?:(\d+))/.match(full_version)
+    info = System::Deploy.info
 
     if saas?
-      version = simple_version ? simple_version[1] : full_version
-      "red_hat_3scale/#{version}-saas"
+      "red_hat_3scale/#{info.major_version}-saas"
     else
-      version = simple_version ? "#{simple_version[1]}.#{simple_version[2]}" ? full_version
-      "red_hat_3scale_api_management/#{version}"
+      "red_hat_3scale_api_management/#{info.minor_version}"
     end
   end
 
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -344,7 +344,7 @@ module ApplicationHelper
     if saas?
       "red_hat_3scale/#{info.major_version}-saas"
     else
-      "red_hat_3scale_api_management/#{info.minor_version}"
+      "red_hat_3scale_api_management/#{info.major_version}.#{info.minor_version}"
     end
   end
 
--- a/lib/system/deploy.rb
+++ b/lib/system/deploy.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module System
   module Deploy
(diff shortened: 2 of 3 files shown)


thomasmaas:
@macejmic how will this work with https://github.com/3scale/porta/blob/THREESCALE-3471/settings_as_usage_rules/config/locales/en.yml#L3

lib/system/deploy.rb
```diff
@@ -15,10 +17,32 @@ class Info
...
+
+      def major_version
+        version.segments[0]
+      end
+
+      private
+
+      class UnknownVersion
+        def segments
+          %w[unknown unknown]
```
guicassolato: This means whenever we pass anything that `Gem::Version.new(release)` cannot parse (e.g. "upstream"), we will show "unknown" instead. Maybe better better showing whatever is in "release" properly escaped in this case.

app/helpers/application_helper.rb
```diff
@@ -339,12 +339,12 @@ def saas?
...
-    full_version = System::Deploy.info.release || '2.0.0'
-    simple_version = /^(?:(\d+)\.)?(?:(\d+))/.match(full_version)
+    info = System::Deploy.info
+
     if saas?
-      "red_hat_3scale/#{simple_version[1]}-saas"
+      "red_hat_3scale/#{info.major_version}-saas"
     else
-      "red_hat_3scale_api_management/#{simple_version[1]}.#{simple_version[2]}"
+      "red_hat_3scale_api_management/#{info.minor_version}"
```
guicassolato: This means showing only the minor version without the major, right? I think in this case we want to show `#{major}.#{minor}`, no?

lib/system/deploy.rb
```diff
@@ -15,10 +17,32 @@ class Info
...
+        end
+      end
+
+      def version
+        @version ||= Gem::Version.new(release)
+      rescue ArgumentError
+        UnknownVersion.new
```
hallelujah:
```suggestion
        UnknownVersion.new(release)
```

lib/system/deploy.rb
```diff
@@ -15,10 +17,32 @@ class Info
...
+        version.segments[1]
+      end
+
+      def major_version
+        version.segments[0]
+      end
+
+      private
+
+      class UnknownVersion
```
hallelujah:
```suggestion
      class UnknownVersion
         attr_reader :segments
        
        def initialize(release)
          @segments = release.to_s.split('.')
        end
        
        def minor_version
           segments[1,-1].joins('.')
        end
```

lib/system/deploy.rb
```diff
@@ -15,10 +17,32 @@ class Info
 
       def initialize(info)
         @revision = info.fetch('revision') { `git rev-parse HEAD 2> /dev/null`.strip }
-        @release = ENV['AMP_RELEASE'].presence
+        @release = info.fetch('release') { '2.0.0' }
```
hallelujah: Why 2.0.0?

lib/system/deploy.rb
```diff
@@ -15,10 +17,32 @@ class Info
...
+      private
+
+      class UnknownVersion
+        def segments
+          %w[unknown unknown]
+        end
+      end
+
+      def version
+        @version ||= Gem::Version.new(release)
```
hallelujah:
```suggestion
        @version ||= VersionParser new(release)
```

lib/system/deploy.rb
```diff
@@ -13,12 +15,36 @@ class Info
       attr_reader :revision, :deployed_at
       attr_reader :release
 
+      delegate :minor_version, :major_version, to: :version
+
       def initialize(info)
         @revision = info.fetch('revision') { `git rev-parse HEAD 2> /dev/null`.strip }
-        @release = ENV['AMP_RELEASE'].presence
+        @release = info.fetch('release') { '2.x' }
```
hallelujah: So 2.0.0 is because of SaaS. It does not make sense to have SaaS related configuration here, so let just put it correctly in the deploy repo

app/helpers/application_helper.rb
```diff
@@ -339,12 +339,12 @@ def saas?
...
-    full_version = System::Deploy.info.release || '2.0.0'
-    simple_version = /^(?:(\d+)\.)?(?:(\d+))/.match(full_version)
+    info = System::Deploy.info
+
     if saas?
-      "red_hat_3scale/#{simple_version[1]}-saas"
+      "red_hat_3scale/#{info.major_version}-saas"
     else
-      "red_hat_3scale_api_management/#{simple_version[1]}.#{simple_version[2]}"
+      "red_hat_3scale_api_management/#{info.major_version}.#{info.minor_version}"
```
guicassolato: This will result in `"red_hat_3scale_api_management/whatever.` if the release is "whatever".

Better:
```suggestion
      version = [info.major_version, info.minor_version].compact.join('.')
      "red_hat_3scale_api_management/#{version}"
```

hallelujah: 👍 Good catch! This is valid but did not change much the initial version

PR #1076, opened 2019-08-13
[zync] Fixes issue when the Zync DB is reset, and Proxy can't be recreated

hallelujah:
Closes THREESCALE-1526

Make sure that the metadata are kept in the successive events on ZyncEvent.create

Steps to reproduce:

1. Setup Porta and launch web server + Sidekiq with ZYNC_ENDPOINT=http://localhost:5000 ZYNC_AUTHENTICATON_TOKEN=foo
2. Setup Zync and launch web server with ZYNC_AUTHENTICATON_TOKEN=foo
3. Drop and recreate zync database: rails db:drop && ./bin/setup
4. Create an application (call it App1)
5. See in the Sidekiq Web UI that Proxy cannot be created and is stuck with UnprocessableEntityError

Expected result:

All the components needed in Zync (Tenant, Service, Proxy, Application) are created

### DIFF
--- a/app/events/zync_event.rb
+++ b/app/events/zync_event.rb
@@ -5,7 +5,7 @@ class ZyncEvent < BaseEventStoreEvent
   # Create Zync Event
 
   def self.create(event, model = event)
-    metadata = event.metadata
+    metadata = event.metadata || {}
     provider_id = metadata.fetch(:provider_id) { model.tenant_id }
 
     attributes = {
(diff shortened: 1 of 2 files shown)


app/events/zync_event.rb
```diff
@@ -19,7 +19,7 @@ def self.create(event, model = event)
     new(
       metadata: {
         provider_id: provider_id,
-      },
+      }.reverse_merge(metadata),
```
hallelujah: Probably only if the event is a ZyncEvent (but not sure)
Anyway this won't do any harm

PR #1082, opened 2019-08-15
Move lib/* to app/lib in order to have eager loading on boot - Part 2

duduribeiro:
Depends on #1049


hallelujah:
Going to close this. I tried to rebase then apply the diff against master it is conflicting

PR #1087, opened 2019-08-19
Rename AuthenticationProvider module to AuthenticationProviders

duduribeiro:

This commit renames the module AuthenticationProvider to 
AuthenticationProviders to avoid clashing it with the already existent 
class with the same name. With this, we can safely drop the 
`require_dependency` for each authentication provider.


duduribeiro:
@hallelujah it is the same problem as the Domain. 

all the provider is in the AuthenticationProviders module namespace, but we already have a class named AuthenticationProviders, and it is clashing. Because of this the require_dependency is required here: https://github.com/3scale/porta/pull/1087/files#diff-ae4c6e150d4b3858cb8bde54e86a01d5L155

PR #1090, opened 2019-08-19
Move lib/cms into api/lib/cms

duduribeiro:

This one moves lib/cms

### DIFF
--- a/app/lib/cms/builtin_pages_support.rb
+++ b/app/lib/cms/builtin_pages_support.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module CMS
   module BuiltinPagesSupport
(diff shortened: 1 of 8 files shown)


app/lib/cms/builtin_pages_support.rb
```diff
@@ -25,8 +27,8 @@ module BuiltinPagesSupport
     def find_builtin_static_page_layout
       system_name = [builtin_template_scope, action_name].compact.join('/')
 
-      page = site_account.builtin_static_pages.find_by_system_name(system_name)
-      layout = page.try!(:layout)
+      page = site_account.builtin_static_pages.find_by(system_name: system_name)
```
Martouta: Please make sure that it does the same thing :smile: :+1:

PR #1093, opened 2019-08-19
Move lib/messenger to app/lib/messenger

duduribeiro:

This one moves lib/messenger


lib/messenger_old.rb
```diff
@@ -1,29 +0,0 @@
-# This class tries to mimic ActiveMailer interface, but sends internal messages
-# insteads of emails.
-#
-# TODO: make this as close to ActiveMailer as possible. Support templates,
-# helpers, and so on.
-class Messenger
```
Martouta: I guess you had your reasons to remove this. I trust you :+1:

duduribeiro: I removed because it looked like no place was using it. 
@hallelujah @guicassolato any idea if this file is being used somewhere?

PR #1095, opened 2019-08-20
[lib] Add a counter to long migration

hallelujah:
So we have at least a bit of information on how our migration is going

### DIFF
--- a/db/migrate/20190716110520_create_the_backend_apis_of_services.rb
+++ b/db/migrate/20190716110520_create_the_backend_apis_of_services.rb
@@ -1,12 +1,19 @@
 # frozen_string_literal: true
 
+require 'progress_counter'
+
 class CreateTheBackendApisOfServices < ActiveRecord::Migration
   def change
     say_with_time 'Migrating proxies api_backend to slugs...' do
       reversible do |dir|
         dir.up do
           Service.transaction do
-            Service.find_each(&:find_or_create_first_backend_api!)
+            progress = ProgressCounter.new(Service.count)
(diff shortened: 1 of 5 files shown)


db/migrate/20190731092338_backfill_add_acts_as_product_to_service.rb
```diff
@@ -1,12 +1,18 @@
...
 
+require 'progress_counter'
+
 class BackfillAddActsAsProductToService < ActiveRecord::Migration
   disable_ddl_transaction!
 
   def change
-    Service.find_in_batches(batch_size: 200) do |records|
-      Service.where(id: records.map(&:id)).update_all act_as_product: false
-      sleep(0.5)
```
hallelujah: Removed this sleep as it should not be needed

db/migrate/20190716110520_create_the_backend_apis_of_services.rb
```diff
@@ -1,12 +1,19 @@
...
 class CreateTheBackendApisOfServices < ActiveRecord::Migration
   def change
     say_with_time 'Migrating proxies api_backend to slugs...' do
       reversible do |dir|
         dir.up do
           Service.transaction do
-            Service.find_each(&:find_or_create_first_backend_api!)
+            progress = ProgressCounter.new(Service.count)
+            Service.find_each(batch_size: 100) do |service|
+              progress.call do
```
guicassolato:
```suggestion
              progress.call(increment: 100) do
```

PR #1077, opened 2019-08-14
Apiap context selector

didierofrivia:
**What this PR does / why we need it**:

Includes backend apis in context selector (navigation)

**Which issue(s) this PR fixes** 
fixes https://issues.jboss.org/browse/THREESCALE-3086

**Verification steps** 

Within the context selector, backend api links should be included

** TO DO:**
* [x] Write new backend api test case
* [x] Fix JS Test
* [x] Fix Lint test
* [x] Fix Cukes
* [x] Fix functional tests

** How it looks like: **

Screenshot 2019-08-16 at 15 48 19
Screenshot 2019-08-16 at 15 48 33
Screenshot 2019-08-16 at 15 52 19

PR #1096, opened 2019-08-20
[apiap] Backend API edit page

guicassolato:

Implements the edit page for backend APIs

Closes [THREESCALE-3276](https://issues.jboss.org/browse/THREESCALE-3276)

### DIFF
--- a/app/controllers/provider/admin/backend_apis_controller.rb
+++ b/app/controllers/provider/admin/backend_apis_controller.rb
@@ -1,18 +1,18 @@
 # frozen_string_literal: true
 
 class Provider::Admin::BackendApisController < Provider::Admin::BaseController
-  before_action :find_backend_api, only: %i[show update]
-  before_action :authorize, only: %i[create update]
+  before_action :find_backend_api, except: %i[index new create]
+  before_action :authorize
 
-  activate_menu :backend_api
+  activate_menu :backend_api, :overview
   layout 'provider'
 
   def index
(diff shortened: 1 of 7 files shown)


app/views/provider/admin/backend_apis/edit.html.slim
```diff
@@ -0,0 +1,8 @@
+h2 Backend API
```
josemigallas: This is different in the services' form:
https://github.com/3scale/porta/blob/f01d12881545e6ebba344fb7514191affb949924/app/views/api/services/edit.html.slim#L1
maybe you could change that one too

josemigallas:
I've noticed the ContextSelector doesn't work anymore. The reason is it takes `@service` as argument but that variable doesn't exist in backend_apis. I think this could be fixed by simply passing `@service || @backend_api` to the component.

guicassolato:
> I've noticed the ContextSelector doesn't work anymore.
The context selector is not part of this PR. Anyway, I tested this branch in combination with https://github.com/3scale/porta/pull/1058 and https://github.com/3scale/porta/pull/1077 in Openshift and the context selector works just fine. What am I missing?

app/controllers/provider/admin/backend_apis_controller.rb
```diff
@@ -25,6 +25,10 @@ def create
     end
   end
 
+  def show; end
```
hallelujah: Do we need to create them?

PR #1097, opened 2019-08-20
Remove require dependency from account

duduribeiro:

    Remove the usage of require_dependency in Account models
    
    - This one removes the needing of require_dependency
    'account/domains'. To this, we needed to rename the
    Account::Domains concern to be Account::ProviderDomains in order
    to not have naming clashes with the other Domains module.
    - Remove the backend_client require_dependency from Account class.

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -192,7 +192,7 @@ class Account < ApplicationRecord
   end
 
   def forum!
-    forum || fail(ActiveRecord::RecordNotFound, "buyer accounts can't have forum")
+    forum || raise(ActiveRecord::RecordNotFound, "buyer accounts can't have forum")
   end
 
   def build_forum(attributes = {})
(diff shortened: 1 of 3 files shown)


app/models/account.rb
```diff
@@ -543,21 +538,17 @@ def on_trial?
...
   end
 
   def finance_support_email
     fse = self[:finance_support_email]
-    if fse.present?
-      fse
-    else
-      support_email
-    end
+    fse.presence || support_email
```
Martouta: in fact it can be done in only 1 line:
```ruby
def finance_support_email
  self[:finance_support_email].presence || support_email
end
```

app/models/account/provider_domains.rb
```diff
@@ -49,13 +51,15 @@ def find_by_domain(domain)
       find_by(domain: domain)
     end
 
+    # rubocop:disable Rails/DynamicFindBy
```
Martouta: I think that it can be done only for the specific line instead of the whole method just as u did in https://github.com/3scale/porta/pull/1097/files#diff-7be675ec904f1d3a73803f8bd354afb1R68

PR #1101, opened 2019-08-20
Move some lonely libs from lib/ to app/lib

duduribeiro:

This one moves some lonely libs from lib/ to app/lib/


config/initializers/cancan_hacks.rb
```diff
@@ -1 +1,54 @@
-require 'cancan_hacks'
+# This makes CanCan work with Inherited resources.
+#
+# Inspiration comes from hexors's comment over here:
+# http://github.com/ryanb/cancan/issues#issue/23
+#
+# TODO: Maybe this is no longer needed, as there is a new version of canca. Check it.
```
Martouta: What happened with this @duduribeiro ?

duduribeiro: Instead of requiring it from lib, I moved the code from `lib/cancan_hacks` to here

Martouta: @duduribeiro I mean this comment, the "Maybe this is not needed anymore because this is for a very old version of cancan"

Martouta: Like... did you investigate if we still need this? because if we don't need it, let's remove it 😄

duduribeiro: ah, gotcha!.. just grabbed the code from there. I think this is fixed by https://github.com/ryanb/cancan/commit/4eee637270e524f3a56341bd595559a96a996551. Will try to update the gem to check if it works!

thanks

Martouta: We spoke in private, conclusion here:
Cadu thinks that is fixed since 1.6.10, and we have 1.7 in our Gemfile, so it seems safe to be removed 😎

config/initializers/cancan_hacks.rb
```diff
@@ -1 +1,54 @@
...
+     def inherits_resources?	
+      included_modules.include?(InheritedResources::BaseHelpers)	
+    end	
+  end	
+end	
+
+ ActionController::Base.send(:include, CanCanHacks)	
+
+ # Module where ability definitions live.	
+module Abilities
```
hallelujah: Uh :)

Why is there an empty module defined here?

duduribeiro: just to be used in the tests test/unit/abilities.. 

it was the way was defined in cancan hacks.. I will drop it

duduribeiro: @hallelujah I dropped the module and updated the tests

duduribeiro:
@hallelujah I opened this ticket to handle the cancan_hacks depreciation: https://issues.jboss.org/browse/THREESCALE-3379

PR #1103, opened 2019-08-21
Backend API metric UI pages

guicassolato:
- New/create metric/method
- Edit/update metric/method
- Destroy metric/method

Closes [THREESCALE-3089](https://issues.jboss.org/browse/THREESCALE-3089)

**TODOs**
- [x] Tests
- [x] Remove the need for service to be loaded https://github.com/3scale/porta/blob/2329f6bc60b4c7b752f047b1dcb3ead2cbbfff6c/app/controllers/provider/admin/backend_apis/metrics_controller.rb#L7
- [x] Append backend API id to the metric system_name

### DIFF
--- a/app/controllers/provider/admin/backend_apis/metrics_controller.rb
+++ b/app/controllers/provider/admin/backend_apis/metrics_controller.rb
@@ -2,11 +2,8 @@
 
 class Provider::Admin::BackendApis::MetricsController < Provider::Admin::BackendApis::BaseController
   before_action :find_metric, except: %i[new create index]
-  before_action :build_metric, only: %i[new create]
 
-  # FIXME: This should be removed. It is still needed because either
-  #        1. the page is still pointing the form to the old metrics controller (new/edit) or
-  #        2. the page uses a helper shared with the old metrics page (app/helpers/services_helper.rb:26) (index)
+  # FIXME: This should be removed. It is still needed because some pages use a helper shared with the old metrics page (app/helpers/services_helper.rb:26) (index)
   before_action :find_service
 
   activate_menu :backend_api, :methods_metrics
@@ -17,19 +14,41 @@ class Provider::Admin::BackendApis::MetricsController < Provider::Admin::Backend
   end
 
   def new
-    render template: '/api/metrics/new'
+    @metric = collection.build
   end
 
   def create
-    # TODO: https://issues.jboss.org/browse/THREESCALE-3089
+    @metric = collection.build(metric_params)
+    if @metric.save
+      onboarding.bubble_update('metric')
+      flash[:notice] = "The #{metric_type} was created"
+      redirect_to provider_admin_backend_api_metrics_path(@backend_api)
+    else
+      flash[:error] = "#{metric_type.capitalize} could not be created"
+      render :new
+    end
   end
 
-  def edit
-    render template: '/api/metrics/edit'
-  end
+  def edit; end
 
   def update
-    # TODO: https://issues.jboss.org/browse/THREESCALE-3089
+    if @metric.update_attributes(metric_params)
+      flash[:notice] = "The #{metric_type} was updated"
+      redirect_to provider_admin_backend_api_metrics_path(@backend_api)
+    else
+      flash[:error] = "#{metric_type.capitalize} could not be updated"
+      render :edit
+    end
+  end
+
+  def destroy
+    if @metric.destroy
+      flash[:notice] = "The #{metric_type} was deleted"
+      redirect_to provider_admin_backend_api_metrics_path(@backend_api)
+    else
+      flash[:error] = "The #{metric_type} could not be deleted"
+      render :edit
+    end
   end
 
   helper_method :bubbles
@@ -49,16 +68,20 @@ class Provider::Admin::BackendApis::MetricsController < Provider::Admin::Backend
     @metric = find_backend_api_metric_by(params[:id])
   end
 
-  def build_metric
-    metrics = if (metric_id = params[:metric_id])
-                find_backend_api_metric_by(metric_id).children
-              else
-                @backend_api.metrics
-              end
-    @metric = metrics.build(params[:metric] || {})
-  end
-
   def find_backend_api_metric_by(id)
     @backend_api.metrics.find(id)
+  end
+
+  def metric_params
+    params.require(:metric).permit(:friendly_name, :system_name, :unit, :description)
+  end
+
+  def metric_type
+    @metric.child? ? 'method' : 'metric'
+  end
+
+  def collection
+    metric_id = params[:metric_id]
+    metric_id ? find_backend_api_metric_by(metric_id).children : @backend_api.metrics
   end
 end
--- a/app/controllers/provider/admin/backend_apis/metrics_controller.rb
+++ b/app/controllers/provider/admin/backend_apis/metrics_controller.rb
@@ -3,9 +3,6 @@
 class Provider::Admin::BackendApis::MetricsController < Provider::Admin::BackendApis::BaseController
   before_action :find_metric, except: %i[new create index]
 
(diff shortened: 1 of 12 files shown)


app/controllers/provider/admin/backend_apis/metrics_controller.rb
```diff
@@ -17,19 +11,41 @@ def index
   end
 
   def new
-    render template: '/api/metrics/new'
+    @metric = collection.build
   end
 
   def create
-    # TODO: https://issues.jboss.org/browse/THREESCALE-3089
+    @metric = collection.build(metric_params.merge(owner: @backend_api))
```
hallelujah: Why do we need owner here?
Is it because of       `find_backend_api_metric_by(metric_id).children`?
Meaning that children (methods) of a metric should have the same owner right?

If we can fix it somehow in `acts_as_tree` it would probably be better

guicassolato: Right. I was lazy. It is because of `find_backend_api_metric_by `. We need to fix it there.

app/controllers/provider/admin/backend_apis/metrics_controller.rb
```diff
@@ -17,19 +11,41 @@ def index
...
+      render :edit
+    end
+  end
+
+  def destroy
+    if @metric.destroy
+      flash[:notice] = "The #{metric_type} was deleted"
+      redirect_to provider_admin_backend_api_metrics_path(@backend_api)
+    else
+      flash[:error] = "#{metric_type.capitalize} could not be deleted"
```
hallelujah:
```suggestion
      flash[:error] = "The #{metric_type} could not be deleted"
```

hallelujah: Same for others capitalize :p

app/controllers/provider/admin/backend_apis/metrics_controller.rb
```diff
@@ -41,24 +57,27 @@ def bubbles
...
+  def metric_params
+    params.require(:metric).permit(:friendly_name, :system_name, :unit, :description)
+  end
+
+  def metric_type
+    @metric.child? ? 'method' : 'metric'
+  end
+
+  def collection
+    if (metric_id = params[:metric_id])
```
hallelujah: Avoid assignment in conditional

app/views/provider/admin/backend_apis/metrics/edit.html.erb
```diff
@@ -0,0 +1,11 @@
+<% if @metric.child? %>
```
hallelujah:
```suggestion
<% content_for :sublayout_title, @metric.child? "Edit Metric" : "Edit Method" %>
```

app/views/provider/admin/backend_apis/metrics/edit.html.erb
```diff
@@ -0,0 +1,11 @@
...
+  <% content_for :sublayout_title, "Edit Method" %>
```
hallelujah:
```suggestion
```

app/views/provider/admin/backend_apis/metrics/edit.html.erb
```diff
@@ -0,0 +1,11 @@
...
+<% else %>
```
hallelujah:
```suggestion
```

app/views/provider/admin/backend_apis/metrics/edit.html.erb
```diff
@@ -0,0 +1,11 @@
...
+  <% content_for :sublayout_title, "Edit Metric" %>
```
hallelujah:
```suggestion
```

app/views/provider/admin/backend_apis/metrics/edit.html.erb
```diff
@@ -0,0 +1,11 @@
...
+<% end %>
```
hallelujah:
```suggestion
```

app/views/provider/admin/backend_apis/metrics/new.html.erb
```diff
@@ -0,0 +1,14 @@
+<% if @metric.child? %>
```
hallelujah: Same as in https://github.com/3scale/porta/pull/1103/files#diff-f8f3bc8ebf3c49aff1e52f17d2f2be6e

app/views/provider/admin/backend_apis/metrics/new.html.erb
```diff
@@ -0,0 +1,11 @@
+<% content_for :sublayout_title, @metric.child? ? "New Method" : "New Metric" %>
+
+<%= semantic_form_for @metric,
+  :url => if @metric.child?
```
josemigallas: Could we define the url in a different line? It would look a bit cleaner in my opinion

guicassolato: Done

josemigallas:
I see this (Sidekick?) error when creating/editing/removing a metric:
Screen Shot 2019-08-30 at 11 11 38

I also cannot create methods, even when having a 'Hits' metric 🤔. The url is `/p/admin/backend_apis/2/metrics/1531/children/new` which looks like the one expected, but I think the request might be missing `metric_id` parameter?
Screen Shot 2019-08-30 at 11 23 59

app/views/provider/admin/backend_apis/metrics/index.html.slim
```diff
@@ -10,99 +10,87 @@ section.Section.u-toggleableBySettingsBox data-state="open"
...
-              td
-                = link_to metric.friendly_name, edit_provider_admin_backend_api_metric_path(@backend_api, metric)
-              td = metric.system_name
-              td = metric.unit
-              td = metric.description
-              - if show_mappings?
-                td
-                  = render 'link_to_mapping_rules', metric: metric
-              td.actions
+  table#metrics.data class='u-sixEqualColumns'
```
josemigallas: Would it make more sense to place Metrics above Methods and not the other way around? Since Metrics are top-level and Methods, children.

guicassolato:
> I see this (Sidekick?) error when creating/editing/removing a metric:
It's not related to this PR.

app/views/provider/admin/backend_apis/metrics/index.html.slim
guicassolato: If we do, it should be done both for products and for backend APIs. I'd say out of scope for this PR.

josemigallas: OK

app/models/metric.rb
```diff
@@ -239,7 +245,12 @@ def associate_to_service_of_parent
...
   def fill_owner
     return true if owner_type?
-    self.owner_id = service_id
-    self.owner_type = 'Service'
+
+    if new_record? && child?
+      self.owner = parent.owner
+    else
+      self.owner_id = service_id
+      self.owner_type = 'Service'
```
Martouta: So if it is an old record OR it is hits, the owner is always a service and not a backend api. Is this still correct ? (I genuinely ask because I really don't know, but it does not sound correct to me)

app/models/metric.rb
```diff
@@ -88,7 +91,11 @@ def name
   #
   # +type+:: Only :hits are supported right now.
   def default?(type)
-    system_name && system_name.to_sym == type.to_sym
+    if type == :hits
+      system_name.to_s =~ self.class.hits_extended_system_name_regex
```
Martouta: I wouldn't ask you to push only for this, but imo we should have an instance method called `hits?` or `extended_name_hits?` or something like that which does exactly this, so it would be more easily readable, and I see this is used somehow often (for example few lines below I see `parent.system_name =~ self.class.hits_extended_system_name_regex` which then would be `parent.THIS_NEW_METHOD?`

app/models/metric.rb
Martouta: Ok, @hallelujah told me that it is correct 👍

PR #1104, opened 2019-08-21
Revert "Rename AuthenticationProvider module to AuthenticationProviders"

hallelujah:
Reverts 3scale/porta#1087

Missing a migration

PR #1105, opened 2019-08-21
Revert "Revert "Rename AuthenticationProvider module to AuthenticationProviders""

hallelujah:
Reverts 3scale/porta#1104

This needs a migration for the column type in the table


hallelujah:
Closing it as we decided that it is much work for tiny benefit
`require_dependency` is meant for those name clashing

> Use this method in code that absolutely needs a certain constant to be defined at that point. A typical use case is to make constant name resolution deterministic for constants with the same relative name in different namespaces whose evaluation would depend on load order otherwise.

PR #1106, opened 2019-08-21
Fixes spec for ApiFilter component

josemigallas:
As part of #1058 the `ApiFilter` was refactored and some specs became outdated.

This PR makes sure ApiFilter has a proper coverage.

### DIFF
--- a/spec/javascripts/Dashboard/ApiFilter.spec.jsx
+++ b/spec/javascripts/Dashboard/ApiFilter.spec.jsx
@@ -23,11 +23,27 @@ it('should render itself', () => {
 })
 
 it('should filter APIs passed in props by name', () => {
-  // TODO
+  // remove / add mean class 'hidden' from DOM Element
+  const remove = jest.fn()
+  const add = jest.fn()
+  jest.spyOn(document, 'getElementById').mockReturnValue({
+    classList: { remove, add }
+  })
+
   const input = apiFilter.find('input')
+
+  // Filter first all apis
   input.simulate('change', { target: { value: 'api' } })
-  input.simulate('change', { target: { value: 'api 1' } })
+
+  expect(remove).toHaveBeenCalledTimes(apis.length)
+  expect(add).toHaveBeenCalledTimes(0)
+
+  remove.mockReset()
+  add.mockReset()
+
+  // Filter only last one
   input.simulate('change', { target: { value: 'api 11' } })
 
-  // expect...
+  expect(remove).toHaveBeenCalledTimes(1)
+  expect(add).toHaveBeenCalledTimes(2)
 })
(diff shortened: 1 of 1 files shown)


spec/javascripts/Dashboard/ApiFilter.spec.jsx
```diff
@@ -22,12 +22,28 @@ it('should render itself', () => {
   expect(apiFilter.find('.ApiFilter').exists()).toBe(true)
 })
 
-it.skip('should filter APIs passed in props by name', () => {
-  // TODO
+it('should filter APIs passed in props by name', () => {
+  // remove / add mean class 'hidden' from DOM Element
+  const remove = jest.fn()
+  const add = jest.fn()
+  jest.spyOn(document, 'getElementById').mockReturnValue({
```
didierofrivia: This smells a bit. I know where it comes from in this case, there's invoked inside the `ApiFilter` component, inside a fn that receives an Api list, then for every item it looks for the element and removes the class according the filter.... I'd change the design of this making it less coupled: the "filter" could be stored in the parent component state and you render the children filtering the Api array, without the need of using classes.

spec/javascripts/Dashboard/ApiFilter.spec.jsx
```diff
@@ -22,12 +22,28 @@ it('should render itself', () => {
...
-  input.simulate('change', { target: { value: 'api 1' } })
+
+  expect(remove).toHaveBeenCalledTimes(apis.length)
+  expect(add).toHaveBeenCalledTimes(0)
+
+  remove.mockReset()
+  add.mockReset()
+
+  // Filter only last one
   input.simulate('change', { target: { value: 'api 11' } })
```
didierofrivia: Also, `simulate` will be deprecated.

josemigallas: Ah, true

spec/javascripts/Dashboard/ApiFilter.spec.jsx
josemigallas: Problem is there is no parent component.. the list is [rendered by rails](https://github.com/3scale/porta/blob/60a39606a3dd26060ab1d1f1ba65aec919f4633b/app/views/provider/admin/dashboards/_section_services.html.slim#L5). This PR's scope is for the existing `ApiFilter`, since tests were [left out in the original PR](https://github.com/3scale/porta/pull/1058#discussion_r316262016). Maybe the question is where these tests should be implemented in gherkin/ruby instead of jest.

I'd leave it as smelly as it is because of time constraints and leave the proper implementation for porta fly.

didierofrivia: Fair enough :)

PR #1108, opened 2019-08-21
List all backend API endpoints inside the APIcast config

josemigallas:

Renders a list of endpoints of all `@service.backend_api_config` instead of a single "Private Base URL".

Without API as Product (still like this for all services that do not act as product):
Screen Shot 2019-08-23 at 1 59 55 PM

With API as Product:
Screen Shot 2019-08-23 at 1 59 25 PM

"Private Base URL" will inflect automatically depending on the number of `backend_api_config` of the service.

[THREESCALE-3282: APIAP: list all backend API endpoints in the proxy config show page](https://issues.jboss.org/browse/THREESCALE-3282)

**Notes for reviewers**:
Go to `/apiconfig/services/<id>/integration`
Either have an APIcast running or create a `service_token` and a `proxy_configs`:
```ruby
# in the rails console
s = Service.find(<id>)
s.service_tokens.create!(value: 'token')
s.proxy.proxy_configs.create(content: { proxy: { hosts: ['foo.com']}}.to_json, environment: 'sandbox').save!
```

### DIFF
--- a/app/assets/stylesheets/provider/_commons.scss
+++ b/app/assets/stylesheets/provider/_commons.scss
@@ -99,6 +99,7 @@ ul, p, ol, dl {
   min-height: line-height-times(1);
   padding: line-height-times(1/2) 0 line-height-times(1/2) 40%;
   border-bottom: $border-width solid $border-color;
+  list-style: none;
 }
 
 .u-dl {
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/provider/_commons.scss
```diff
@@ -99,6 +99,7 @@ ul, p, ol, dl {
   min-height: line-height-times(1);
   padding: line-height-times(1/2) 0 line-height-times(1/2) 40%;
   border-bottom: $border-width solid $border-color;
+  list-style: none;
```
guicassolato: Can this change break other views?

josemigallas: I haven't found any other `dd` with a list inside and this style is usually set to all lists. I think this won't do any harm.

PR #1111, opened 2019-08-22
Fixes API search in non-APIAP dashboards

josemigallas:
🐛
`ApiFilter` is not working when APIAP is disabled, this was introduced in #1058 .

Fixes:
- Create a pack for services searchbar (to be consistent with #1058) that is loaded when APIAP rolling update is off
- Adjust style `.hidden` to affect both service sections and product/backend sections


app/javascript/packs/services_search_bar.js
```diff
@@ -0,0 +1,11 @@
+// TODO: remove this pack when apiap rolling updated is removed
```
Martouta: Add this somewhere in our pending jiras or create new one because otherwise it will be forgotten @macejmic

PR #1112, opened 2019-08-22
Inject Routing policy for Backend API upstreams (THREESCALE-3311)

hallelujah:
So the request is properly mapped to the correct upstream

Closes https://issues.jboss.org/browse/THREESCALE-3311

### DIFF
--- a/app/lib/backend_api_logic/routing_policy.rb
+++ b/app/lib/backend_api_logic/routing_policy.rb
@@ -17,7 +17,7 @@ module BackendApiLogic
       end
 
       def to_a
-        rules = backend_api_configs.each_with_object([]) do |config, rules|
+        rules = backend_api_configs.reorder(path: :desc).each_with_object([]) do |config, rules|
           rule = Rule.new(config).as_json
           rules << rule if rule
         end
--- a/app/lib/backend_api_logic/routing_policy.rb
+++ b/app/lib/backend_api_logic/routing_policy.rb
@@ -57,7 +57,7 @@ module BackendApiLogic
           if path.empty?
             "/.*"
           else
-          "/#{path}/.*|/#{path}/?"
+            "/#{path}/.*|/#{path}/?"
           end
         end
       end
--- a/app/lib/backend_api_logic/routing_policy.rb
+++ b/app/lib/backend_api_logic/routing_policy.rb
@@ -1,0 +1,69 @@
+# frozen_string_literal: true
+
+module BackendApiLogic
+  module RoutingPolicy
+    def policy_chain
+      chain = super
+      return chain unless service.act_as_product?
+      Builder.new(service).to_a.concat(chain)
+    end
+
+    class Builder
+      delegate :proxy, :backend_api_configs, to: :@service
+      delegate :policies_config, to: :proxy
+
+      def initialize(service)
+        @service = service
+      end
+
+      def to_a
+        rules = backend_api_configs.each_with_object([]) do |config, rules|
+          rule = Rule.new(config).as_json
+          rules << rule if rule
+        end
+        return [] if rules.empty?
+        [{
+          name: "routing",
+          version: "builtin",
+          enabled: true,
+          configuration: {
+            rules: rules
+          }
+        }].as_json
+      end
+
+      class Rule
+        delegate :private_endpoint, :path, to: :@config
+
+        def initialize(config)
+          @config = config
+        end
+
+        def as_json
+          return if private_endpoint.blank?
+          {
+            url: private_endpoint,
+            condition: {
+              operations: [
+                match: :path,
+                op: :matches,
+                value: path_to_regex
+              ]
+            }
+          }
+        end
+
+        def path_to_regex
+          if path.empty?
+            "/.*"
+          else
+          "/#{path}/.*|/#{path}/?"
+          end
+        end
+      end
+      private_constant :Rule
+    end
+    private_constant :Builder
+
+  end
+end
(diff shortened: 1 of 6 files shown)


app/models/backend_api_routing_policy_injector.rb
```diff
@@ -0,0 +1,76 @@
+# frozen_string_literal: true
+
+class BackendApiRoutingPolicyInjector
```
guicassolato: Q1: Are we using later this same class to merge all proxy rules of each backend (prepending the backend path) into the proxy config?

if so...
Q2: Can we also use it when showing/counting proxy rules of a service at https://github.com/3scale/porta/blob/9bcb8724c6931d8dd2f0b148c14f193fcc4b4eec/app/presenters/api/integrations_show_presenter.rb#L58?

app/lib/backend_api_logic/routing_policy.rb
```diff
@@ -0,0 +1,88 @@
...
+      end
+
+      class Rule
+        delegate :private_endpoint, :path, to: :@config
+
+        def initialize(config)
+          @config = config
+        end
+
+        def to_h
```
hallelujah: It is expecting to return a `Hash` but actually can return `nil`
I will change the naming

app/lib/backend_api_logic/routing_policy.rb
```diff
@@ -0,0 +1,69 @@
...
+
+      class Rule
+        delegate :private_endpoint, :path, to: :@config
+
+        def initialize(config)
+          @config = config
+        end
+
+        def as_json
+          return if private_endpoint.blank?
```
hallelujah: This is not tested :)
But it makes sense not to route anything if the private_endpoint is not assigned

hallelujah: I will test it :)

test/unit/proxy_test.rb
```diff
@@ -73,6 +74,36 @@ def test_policy_chain
...
+      backend_api1 = FactoryBot.create(:backend_api, account: account, private_endpoint: 'https://private-1.example.com')
+      backend_api2 = FactoryBot.create(:backend_api, account: account, private_endpoint: 'https://private-2.example.com')
+      FactoryBot.create(:backend_api_config, path: '/null', backend_api: null_backend_api, service: service)
+      FactoryBot.create(:backend_api_config, path: '/foo', backend_api: backend_api1, service: service)
+      FactoryBot.create(:backend_api_config, path: '/bar', backend_api: backend_api2, service: service)
+      policy_chain =  [
+        {"name"=>"routing", "version"=>"builtin", "enabled"=>true,
+          "configuration"=>{
+            "operations"=>[
+              {"url"=>"https://echo-api.3scale.net:443", "condition"=>{"operations"=>[{"match"=>"path", "op"=>"matches", "value"=>"/.*"}]}},
```
hallelujah: This one is because of the default backend api created for the service ;)

app/lib/backend_api_logic/routing_policy.rb
```diff
@@ -0,0 +1,69 @@
...
+              ]
+            }
+          }
+        end
+
+        def path_to_regex
+          if path.empty?
+            "/.*"
+          else
+          "/#{path}/.*|/#{path}/?"
```
guicassolato:
```suggestion
            "/#{path}/.*|/#{path}/?"
```

test/unit/proxy_test.rb
```diff
@@ -73,6 +74,36 @@ def test_policy_chain
...
+      service = FactoryBot.create(:simple_service, account: account, act_as_product: true)
+      null_backend_api = FactoryBot.create(:backend_api, account: account, private_endpoint: 'https://foo.baz')
+      null_backend_api.update_columns(private_endpoint: '')
+      backend_api1 = FactoryBot.create(:backend_api, account: account, private_endpoint: 'https://private-1.example.com')
+      backend_api2 = FactoryBot.create(:backend_api, account: account, private_endpoint: 'https://private-2.example.com')
+      FactoryBot.create(:backend_api_config, path: '/null', backend_api: null_backend_api, service: service)
+      FactoryBot.create(:backend_api_config, path: '/foo', backend_api: backend_api1, service: service)
+      FactoryBot.create(:backend_api_config, path: '/bar', backend_api: backend_api2, service: service)
+      policy_chain =  [
+        {"name"=>"routing", "version"=>"builtin", "enabled"=>true,
```
davidor: The "enabled" field is not used in APIcast. I guess it's something that only Porta uses?

test/unit/proxy_test.rb
```diff
@@ -73,6 +74,36 @@ def test_policy_chain
...
+          "configuration"=>{
+            "operations"=>[
```
davidor: Should be called "rules" instead of "operations" according to the schema.
See this example: https://github.com/3scale/APIcast/tree/master/gateway/src/apicast/policy/routing#rule-with-several-operations

test/unit/proxy_test.rb
davidor: Just to be sure, this will always match, so the request will never be proxied to the 2 API backends below. Is that what you need in this example?

The routing policy will route to the first URL that matches.

test/unit/proxy_test.rb
hallelujah: Yes true, but we have it since last year.
It is probably better in a next PR to remove it

test/unit/proxy_test.rb
hallelujah: Correct, This is because of the first backend api that has "/" in the routing. We allow that but it would be better to have ordering of `path` so that we can use sub-routing like {`/foo` and `/foo/bar`} or simply  {`/` and `/foo`}

But that would necessarily implies the need ordering of path. I do not know if we have to fix that now (at least not in this PR)

test/unit/proxy_test.rb
davidor: :+1:

app/models/backend_api_routing_policy_injector.rb
hallelujah: Q1: It is only used in `Proxy#policy_chain`, when saving the configuration to the proxy config.
Q2: I do not understand

hallelujah:
Yes we are adding it anyway.
Might be an issue but It will work because you can add also a path to your backend API which is not the case now.

PR #1115, opened 2019-08-22
[wip] Remove uniqueness from index on metrics(service_id, system_name)

guicassolato:

It removes the uniqueness constraint from the index on `metrics(service_id, system_name)`. The reason is because now a metric may belong to a backend API and not to a service. Without this PR, two different backend APIs each one holding a metric with the same `system_name` would cause a uniqueness constraint violation in the DB due to `NULL` `service_id` in both of them.


hallelujah:
Don't we append the backend_api id to the system_name of a metric?

guicassolato:
Because of https://github.com/3scale/porta/pull/1103, this is no longer needed.

PR #1117, opened 2019-08-22
Updating css-loader to ^2.1.1

damianpm:

Webpack loader `css-loader` must be updated

https://issues.jboss.org/browse/THREESCALE-3303

**Verification steps** 

`npm run update-dependencies`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3860,12 +3860,6 @@
         "lodash.uniq": "^4.5.0"
       }
     },
-    "caniuse-db": {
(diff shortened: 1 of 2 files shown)


damianpm:
> I think you might need to enable modules manually, see [breaking changes here.](https://github.com/webpack-contrib/css-loader/releases#2.0.0)

@josemigallas  CSS Modules are handled by `sass-loader` in https://github.com/3scale/porta/blob/master/config/webpack/environment.js#L21

PR #1119, opened 2019-08-22
Add an option to mark that the Service acts as product

duduribeiro:

This commit adds a new option in the screen to create a service
to inform if it will act as product. If this options is checked, it will
present a select to choose which backend api it will be attached
or if it will create a new one.

This is when you create a Service without checking act as product:
act_as_product_1

This is when you create a Service checking act as product and let it create a new Backend API:
act_as_product_2

This is when you create a Service checking act as product and select an already existent Backend API:
act_as_product_3

[THREESCALE-3294](https://issues.jboss.org/browse/THREESCALE-3294)

**Verification steps** 

1. Go to the screen to create a new Service
2. Create a new service without checking `act as product` checkbox.
3. It should create a new service without attaching any backend api into it.
4. Go to the screen to create a new Service
5. Create a new service checking `act as product` checkbox and selecting `New Backend API` in the Backend API combo box.
6. It should create a new service with a freshly created backend api for this service.
7. Go to the screen to create a new Service
8. Create a new service checking `act as product` checkbox and selecting some already created backend api in the Backend API combo box.
9. It should create a new service with an already existent backend api for this service.

### DIFF
--- a/app/controllers/api/services_controller.rb
+++ b/app/controllers/api/services_controller.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 class Api::ServicesController < Api::BaseController
   include ServiceDiscovery::ControllerMethods
 
@@ -34,8 +36,8 @@ class Api::ServicesController < Api::BaseController
 
   def create
     @service = collection.new # this is done in 2 steps so that the account_id is in place as preffix_key relies on it
-    @service.attributes = params[:service]
-    @service.system_name = params[:service][:system_name]
+    @service.attributes = service_params
+    @service.backend_api_configs = build_backend_api_configs
 
     if can_create? && @service.save
       flash[:notice] =  'Service created.'
@@ -65,8 +67,21 @@ class Api::ServicesController < Api::BaseController
     redirect_to provider_admin_dashboard_path
   end
 
-  protected
+  private
 
+  def build_backend_api_configs
+    return [] if !provider_can_use?(:api_as_product) || params[:service][:backend_api].blank?
+    backend_api = current_account.backend_apis.find(params[:service][:backend_api])
+    [BackendApiConfig.new(backend_api: backend_api)]
+  end
+
+  def service_params
+    allowed_params = %i[system_name name description]
+    allowed_params << :act_as_product if provider_can_use?(:api_as_product)
+    params.require(:service).permit(allowed_params)
+  end
+
+  protected
 
   def service_name_changed?
     @service.previous_changes['name']
--- a/app/javascript/src/NewService/components/FormElements/index.jsx
+++ b/app/javascript/src/NewService/components/FormElements/index.jsx
@@ -5,5 +5,6 @@ export {HiddenServiceDiscoveryInput} from 'NewService/components/FormElements/Hi
 export {Label} from 'NewService/components/FormElements/Label'
 export {Select} from 'NewService/components/FormElements/Select'
 export {ErrorMessage} from 'NewService/components/FormElements/ErrorMessage'
+export {BackendApiSelect} from 'NewService/components/FormElements/BackendApiSelect'
 export {ServiceDiscoveryListItems} from 'NewService/components/FormElements/ServiceDiscoveryListItems'
 export {ServiceManualListItems} from 'NewService/components/FormElements/ServiceManualListItems'
--- a/app/javascript/src/NewService/components/NewServiceForm.jsx
+++ b/app/javascript/src/NewService/components/NewServiceForm.jsx
@@ -4,18 +4,21 @@ import React, {useState} from 'react'
 
 import {ServiceSourceForm, ServiceDiscoveryForm, ServiceManualForm} from 'NewService'
 import {createReactWrapper} from 'utilities/createReactWrapper'
+import type {Api} from 'Types/Api'
 
 type Props = {
   isServiceDiscoveryAccessible: boolean,
   isServiceDiscoveryUsable: boolean,
   serviceDiscoveryAuthenticateUrl: string,
   providerAdminServiceDiscoveryServicesPath: string,
-  adminServicesPath: string
+  adminServicesPath: string,
+  apiap: boolean,
+  backendApis: Api[]
 }
 
 const NewServiceForm = (props: Props) => {
   const {isServiceDiscoveryAccessible, isServiceDiscoveryUsable, serviceDiscoveryAuthenticateUrl,
-    providerAdminServiceDiscoveryServicesPath, adminServicesPath} = props
+    providerAdminServiceDiscoveryServicesPath, adminServicesPath, apiap, backendApis} = props
 
   const [formMode, setFormMode] = useState('manual')
   const [loadingProjects, setLoadingProjects] = useState(false)
@@ -24,8 +27,8 @@ const NewServiceForm = (props: Props) => {
     setFormMode(event.currentTarget.value)
   }
 
-  const formToRender = () => formMode === 'manual' || !isServiceDiscoveryAccessible
-    ? <ServiceManualForm formActionPath={adminServicesPath}/>
+  const formToRender = () => formMode === 'manual'
+    ? <ServiceManualForm formActionPath={adminServicesPath} apiap={apiap} backendApis={backendApis} />
     : <ServiceDiscoveryForm formActionPath={providerAdminServiceDiscoveryServicesPath} setLoadingProjects={setLoadingProjects} />
 
   return (
--- a/app/javascript/src/NewService/components/ServiceManualForm.jsx
+++ b/app/javascript/src/NewService/components/ServiceManualForm.jsx
@@ -1,10 +1,19 @@
 // @flow
 
 import React from 'react'
-import {FormWrapper, ServiceManualListItems} from 'NewService/components/FormElements'
+import {BackendApiSelect, FormWrapper, ServiceManualListItems} from 'NewService/components/FormElements'
 import type {FormProps} from 'NewService/types'
+import type {Api} from 'Types/Api'
 
-const ServiceManualForm = ({formActionPath}: {formActionPath: string}) => {
+type Props = {
+  formActionPath: string,
+  apiap: boolean,
+  backendApis: Api[]
+}
+
+const ServiceManualForm = (props: Props) => {
+  const {formActionPath, apiap, backendApis} = props
+
   const formProps: FormProps = {
     id: 'new_service',
     formActionPath,
@@ -14,6 +23,7 @@ const ServiceManualForm = ({formActionPath}: {formActionPath: string}) => {
   return (
     <FormWrapper {...formProps}>
       <ServiceManualListItems/>
+      {apiap && <BackendApiSelect backendApis={backendApis} />}
     </FormWrapper>
   )
 }
--- a/app/models/backend_api_config.rb
+++ b/app/models/backend_api_config.rb
@@ -8,6 +8,10 @@ class BackendApiConfig < ApplicationRecord
 
   validates :path, length: { in: 0..255, allow_nil: false }, path: true
 
+  after_create do
+    backend_api.metrics.each { |metric| metric.send(:sync_backend) }
+  end
+
   def path=(value)
     super(value.to_s.reverse.chomp("/").reverse.chomp("/"))
   end
--- a/spec/javascripts/NewService/NewServiceForm.spec.jsx
+++ b/spec/javascripts/NewService/NewServiceForm.spec.jsx
@@ -5,6 +5,7 @@ import Enzyme, {mount} from 'enzyme'
 import Adapter from 'enzyme-adapter-react-16'
 
 import {NewServiceForm, ServiceManualForm, ServiceDiscoveryForm} from 'NewService'
+import {BackendApiSelect} from 'NewService/components/FormElements'
 
 import * as utils from 'utilities/utils'
 jest.spyOn(utils, 'CSRFToken')
@@ -17,7 +18,9 @@ const props = {
   isServiceDiscoveryUsable: true,
   serviceDiscoveryAuthenticateUrl: 'authenticate-url',
   providerAdminServiceDiscoveryServicesPath: 'my-path',
-  adminServicesPath: 'my-other-path'
+  adminServicesPath: 'my-other-path',
+  apiap: false,
+  backendApis: []
 }
 
 it('should render itself', () => {
@@ -41,6 +44,11 @@ it('should render the correct form depending on which mode is selected', () => {
   expect(wrapper.find(ServiceDiscoveryForm).exists()).toEqual(false)
 })
 
+it('should not render BackendApiSelect by default', () => {
+  const wrapper = mount(<NewServiceForm {...props}/>)
+  expect(wrapper.find(BackendApiSelect).exists()).toEqual(false)
+})
+
 describe('when Service Discovery is not accessible', () => {
   beforeAll(() => {
     props.isServiceDiscoveryAccessible = false
@@ -58,5 +66,16 @@ describe('when Service Discovery is not accessible', () => {
     expect(wrapper.find('ServiceManualForm').exists()).toEqual(true)
     expect(wrapper.find('form#new_service').exists()).toEqual(true)
     expect(wrapper.find('form#service_source').exists()).toEqual(false)
+  })
+})
+
+describe('when Api as Product is enabled', () => {
+  beforeAll(() => {
+    props.apiap = true
+  })
+
+  it('should render BackendApiSelect', () => {
+    const wrapper = mount(<NewServiceForm {...props}/>)
+    expect(wrapper.find(BackendApiSelect).exists()).toEqual(true)
   })
 })
--- a/spec/javascripts/NewService/formElements/BackendApiSelect.spec.jsx
+++ b/spec/javascripts/NewService/formElements/BackendApiSelect.spec.jsx
@@ -1,0 +1,40 @@
+// @flow
+
+import React from 'react'
+import {act} from 'react-dom/test-utils'
+import Enzyme, {shallow} from 'enzyme'
+import Adapter from 'enzyme-adapter-react-16'
+
+import {BackendApiSelect} from 'NewService/components/FormElements'
+import { BASE_PATH } from 'NewService'
+
+Enzyme.configure({adapter: new Adapter()})
+
+const props = {
+  backendApis: []
+}
+
+it('should render itself', () => {
+  const wrapper = shallow(<BackendApiSelect {...props} />)
+  expect(wrapper.find('#service_act_as_product_input').exists()).toEqual(true)
+})
+
+it('should render Act as Product checkbox', () => {
+  const wrapper = shallow(<BackendApiSelect {...props} />)
+  expect(wrapper.find('input[name="service[act_as_product]"]').exists()).toEqual(true)
+})
+
+it('should not render select element when Act as Product checkbox is unchecked', () => {
+  const wrapper = shallow(<BackendApiSelect {...props} />)
+  expect(wrapper.find('select[name="service[backend_api]"]').exists()).toEqual(false)
+})
+
+it('should render select element when Act as Product checkbox is checked', () => {
+  const wrapper = shallow(<BackendApiSelect {...props} />)
+  expect(wrapper.find('select[name="service[backend_api]"]').exists()).toEqual(false)
+  const eventObj = {currentTarget: {checked: true}}
+  act(() => {
+    wrapper.find('input[name="service[act_as_product]"]').simulate('change', eventObj)
+  })
+  expect(wrapper.find('select[name="service[backend_api]"]').exists()).toEqual(true)
+})
--- a/test/integration/api/services_controller_test.rb
+++ b/test/integration/api/services_controller_test.rb
@@ -5,79 +5,152 @@ require 'test_helper'
 class Api::ServicesControllerTest < ActionDispatch::IntegrationTest
 
   setup do
-    provider = FactoryBot.create(:provider_account)
-    @service = provider.default_service
-    login! provider
+    @provider = FactoryBot.create(:provider_account)
+    @service = @provider.default_service
+    login! @provider
   end
 
   attr_reader :service
 
-  test 'settings renders the right template and contains the right sections' do
-    get settings_admin_service_path(service)
-    assert_response :success
-    page = Nokogiri::HTML::Document.parse(response.body)
+  class SettingsTest < Api::ServicesControllerTest
+    test 'settings renders the right template and contains the right sections' do
+      get settings_admin_service_path(service)
+      assert_response :success
+      page = Nokogiri::HTML::Document.parse(response.body)
 
-    assert_template 'api/services/settings'
-    section_titles = page.xpath("//fieldset[@class='inputs']/legend").text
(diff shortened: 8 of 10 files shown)


app/controllers/api/services_controller.rb
```diff
@@ -65,8 +64,19 @@ def destroy
     redirect_to provider_admin_dashboard_path
   end
 
-  protected
+  private
+
+  def find_backend_api
+    current_account.backend_apis.find(params[:service][:backend_api])
```
Martouta: `current_account.backend_apis.find(params.require(:service).require(:backend_api])` because otherwise if `service` is not in the params, it will raise `NoMethodError`

app/controllers/api/services_controller.rb
```diff
@@ -34,8 +34,7 @@ def settings
 
   def create
     @service = collection.new # this is done in 2 steps so that the account_id is in place as preffix_key relies on it
-    @service.attributes = params[:service]
-    @service.system_name = params[:service][:system_name]
+    @service.attributes = service_params
```
Martouta: only do this is the account is authorized to do it :smile: or if we are not doing the authorizations in depth yet, then at leaaast make sure the rolling update is enabled

app/controllers/api/services_controller.rb
```diff
@@ -65,8 +64,19 @@ def destroy
...
-  protected
+  private
+
+  def find_backend_api
+    current_account.backend_apis.find(params[:service][:backend_api])
+  end
 
+  def service_params
+    params.require(:service).permit(:system_name, :name, :description, :act_as_product).tap do |p|
+      p[:backend_api_configs] = [ BackendApiConfig.new(backend_api: find_backend_api) ] if params[:service][:backend_api].present?
```
Martouta: only do this is the account is authorized to do it :smile: or if we are not doing the authorizations in depth yet, then at leaaast make sure the rolling update is enabled

Martouta: and add a test to make sure this backend api part works and only if the rolling update is enabled

app/javascript/src/NewService/components/FormElements/SelectBackendApi.jsx
```diff
@@ -0,0 +1,38 @@
+// @flow
+
+import React, {useState} from 'react'
+import {Label} from 'NewService/components/FormElements'
+
+const SelectBackendApi = ({backendApis}: {backendApis: Array<{id: number, name: string}>}) => {
+  const [actAsProduct, setActAsProduct] = useState(false)
+
+  const toggleActAsProduct = (event) => {
```
josemigallas:
```suggestion
  const toggleActAsProduct = (event: SyntheticInputEvent<HTMLInputElement>) => {
```

app/javascript/src/NewService/components/FormElements/SelectBackendApi.jsx
```diff
@@ -0,0 +1,38 @@
...
+    setActAsProduct(event.target.checked)
```
josemigallas: In React we should use `currentTarget` as it is safer.
```suggestion
    setActAsProduct(event.currentTarget.checked)
```

app/javascript/src/NewService/components/FormElements/SelectBackendApi.jsx
```diff
@@ -0,0 +1,38 @@
...
+      </li>
+
+      {actAsProduct &&
+        <li id="service_backend_api_input">
+          <Label
+            htmlFor='service_backend_api'
+            label='Backend API'
+          />
+          <select name="service[backend_api]" id="service_backend_api">
+            <option key='new' value=''>New Backend API</option>
```
josemigallas: I think maybe _Create a new Backend API_ would be slightly nicer.

app/javascript/src/NewService/components/NewServiceForm.jsx
```diff
@@ -25,8 +27,8 @@ const NewServiceForm = (props: Props) => {
   }
 
   const formToRender = () => formMode === 'manual' || !isServiceDiscoveryAccessible
-    ? <ServiceManualForm formActionPath={adminServicesPath}/>
-    : <ServiceDiscoveryForm formActionPath={providerAdminServiceDiscoveryServicesPath} setLoadingProjects={setLoadingProjects} />
+    ? <ServiceManualForm formActionPath={adminServicesPath} apiap={apiap} backendApis={backendApis} />
```
josemigallas: No need to propagate `apiap` and `backendApis` down. Since the form doesn't depend on the `formMode`. Simply add the new field after `formToRender()`.

app/javascript/src/NewService/components/FormElements/index.jsx
```diff
@@ -5,5 +5,6 @@ export {HiddenServiceDiscoveryInput} from 'NewService/components/FormElements/Hi
 export {Label} from 'NewService/components/FormElements/Label'
 export {Select} from 'NewService/components/FormElements/Select'
 export {ErrorMessage} from 'NewService/components/FormElements/ErrorMessage'
+export {SelectBackendApi} from 'NewService/components/FormElements/SelectBackendApi'
```
josemigallas: I would call it `BackendApiSelect`, since it's a Select element not a method _to select_ something. Just my opinion.

spec/javascripts/NewService/NewServiceForm.spec.jsx
```diff
@@ -60,3 +67,14 @@ describe('when Service Discovery is not accessible', () => {
     expect(wrapper.find('form#service_source').exists()).toEqual(false)
   })
 })
+
+describe('when Api as Product is enabled', () => {
+  beforeAll(() => {
+    props.apiap = true
+  })
+
+  it('should render the checkbox to mark if it will act as product', () => {
```
josemigallas: _it should render Act as Product checkbox_

spec/javascripts/NewService/NewServiceForm.spec.jsx
```diff
@@ -41,6 +43,11 @@ it('should render the correct form depending on which mode is selected', () => {
   expect(wrapper.find(ServiceDiscoveryForm).exists()).toEqual(false)
 })
 
+it('should not render the ckeckbox to mark if it will act as product', () => {
```
josemigallas: _it should not render Act as Product checkbox by default_

spec/javascripts/NewService/NewServiceForm.spec.jsx
```diff
@@ -41,6 +43,11 @@ it('should render the correct form depending on which mode is selected', () => {
...
+  const wrapper = mount(<NewServiceForm {...props}/>)
+  expect(wrapper.find('input#act_as_product').exists()).toEqual(false)
```
josemigallas: You correct me if I'm wrong but I think the relevant attribute for Rails is `name="service[act_as_product]"`. If so, better assert that one:
```suggestion
  expect(wrapper.find('input[name="service[act_as_product]"]').exists()).toEqual(false)
```

EDIT
In fact, I think the proper way is to assert here de component `SelectBackendApi`, since the responsibility of `NewServiceForm` is to render de component nothing else.
Then, in **SelectBackendApi.spec.jsx** we make all the specific assertions like the input is rendered and the select works, etc.

spec/javascripts/NewService/ServiceDiscoveryForm.spec.jsx
```diff
@@ -72,3 +74,15 @@ describe('fetchProjects', () => {
     expect(fetch).toHaveBeenCalled()
   })
 })
+
+
+describe('when Api as Product is enabled', () => {
```
josemigallas: This is repetition, it's already asserted in `NewServiceForm`.

app/javascript/src/NewService/components/NewServiceForm.jsx
```diff
@@ -10,12 +10,14 @@ type Props = {
   isServiceDiscoveryUsable: boolean,
   serviceDiscoveryAuthenticateUrl: string,
   providerAdminServiceDiscoveryServicesPath: string,
-  adminServicesPath: string
+  adminServicesPath: string,
+  apiap: boolean,
+  backendApis: Array<{id: number, name: string}>
```
josemigallas: I think this prop should be optional:
```suggestion
  backendApis?: Array<{id: number, name: string}>
```

spec/javascripts/NewService/NewServiceForm.spec.jsx
```diff
@@ -17,7 +17,9 @@ const props = {
   isServiceDiscoveryUsable: true,
   serviceDiscoveryAuthenticateUrl: 'authenticate-url',
   providerAdminServiceDiscoveryServicesPath: 'my-path',
-  adminServicesPath: 'my-other-path'
+  adminServicesPath: 'my-other-path',
+  apiap: false,
+  backendApis: []
```
josemigallas: If optional, no need to pass this unless apiap true.

app/javascript/src/NewService/components/FormElements/ServiceDiscoveryListItems.jsx
```diff
@@ -1,17 +1,19 @@
...
 
 import React, {useState, useEffect} from 'react'
-import {Label, Select} from 'NewService/components/FormElements'
+import {Label, Select, SelectBackendApi} from 'NewService/components/FormElements'
 import {fetchData} from 'utilities/utils'
 import {BASE_PATH} from 'NewService'
 
 type Props = {
+  apiap: boolean,
+  backendApis: Array<{id: number, name: string}>,
```
damianpm: Here you can reuse the `BackendApi` type

app/javascript/src/NewService/components/FormElements/ServiceManualListItems.jsx
```diff
@@ -1,36 +1,40 @@
...
-    </li>
-    <li id="service_description_input" className="text optional">
-      <Label
-        htmlFor='service_description'
-        label='Description'
-      />
-      <textarea rows="3" id="service_description" name="service[description]"></textarea>
-    </li>
-  </React.Fragment>
+const ServiceManualListItems = ({apiap, backendApis}: {apiap: boolean, backendApis: Array<{id: number, name: string}>}) => {
```
damianpm: Same here, you can follow the approach suggested in https://github.com/3scale/porta/pull/1119/files#diff-7f4ccafcb147cdb8c05f8f25bdd19e28R6

app/javascript/src/NewService/components/NewServiceForm.jsx
damianpm: True, you are passing `apiap` and `backendApis` to **ServiceManualForm** -> **ServiceManualListItems** then rendering with `{apiap && <SelectBackendApi backendApis={backendApis} />}`

And doing the same for **ServiceDiscoveryForm** -> **ServiceDiscoveryListItems** then `{apiap && <SelectBackendApi backendApis={backendApis} />}`

But it's unnecessary, as you are duplicating the same logic in 2 components. You can simply add **SelectBackendApi** here

app/javascript/src/NewService/components/ServiceDiscoveryForm.jsx
```diff
@@ -9,11 +9,13 @@ import {fetchData} from 'utilities/utils'
 import {PROJECTS_PATH} from 'NewService'
 
 type Props = {
+  apiap: boolean,
```
damianpm: This won't be needed if you do the changes suggested in https://github.com/3scale/porta/pull/1119/files#diff-708bb89f617b27eb6d9826c93af38a90R30

app/javascript/src/NewService/components/ServiceManualForm.jsx
```diff
@@ -4,7 +4,15 @@ import React from 'react'
 import {FormWrapper, ServiceManualListItems} from 'NewService/components/FormElements'
 import type {FormProps} from 'NewService/types'
 
-const ServiceManualForm = ({formActionPath}: {formActionPath: string}) => {
+type Props = {
+  apiap: boolean,
```
damianpm: Same as previous comment

spec/javascripts/NewService/ServiceDiscoveryForm.spec.jsx
damianpm: If you move **SelectBackendApi** outside **ServiceManualForm**  and **ServiceDiscoveryForm** this is not needed

spec/javascripts/NewService/formElements/ServiceDiscoveryListItems.spec.jsx
```diff
@@ -13,7 +13,7 @@ Enzyme.configure({adapter: new Adapter()})
 
 const projects = ['project_00', 'project_01']
 const onError = jest.fn()
-const props = { projects, onError }
+const props = { projects, onError, apiap: false, backendApis: [] }
```
damianpm: Same as prev comment

guicassolato:
When an existing Backend API is selected to start the service with (as opposed to using the option "New Backend API"), we need to make sure the existing metrics of that backend API are synced with apisonator together with the new service.

Right now I know we sync new metrics just added to a backend through all services using the backend, but not existing metrics of a backend when the backend is added to a service.

The same needs to be done when we add a backend API to a _existing_ service. (Another pending issue of APIAP, BTW.)

app/javascript/src/NewService/components/FormElements/SelectBackendApi.jsx
```diff
@@ -0,0 +1,38 @@
+// @flow
+
+import React, {useState} from 'react'
+import {Label} from 'NewService/components/FormElements'
+
+const SelectBackendApi = ({backendApis}: {backendApis: Array<{id: number, name: string}>}) => {
+  const [actAsProduct, setActAsProduct] = useState(false)
```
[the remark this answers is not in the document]
duduribeiro: I can't see error on this too.

app/javascript/src/NewService/components/FormElements/ServiceManualListItems.jsx
```diff
@@ -3,34 +3,37 @@
...
-    <li id="service_description_input" className="text optional">
-      <Label
-        htmlFor='service_description'
-        label='Description'
-      />
-      <textarea rows="3" id="service_description" name="service[description]"></textarea>
-    </li>
-  </React.Fragment>
+const ServiceManualListItems = () => {
+  return (
```
josemigallas: Better remove `return` from here, the following are equivalent:
```
const func = (x) => (
  x + x
)
const func = (x) => {
  return x + x
}

File: app/javascript/src/NewService/components/NewServiceForm.jsx [hunk 22]
```diff
@@ -3,19 +3,23 @@
 import React, {useState} from 'react'
 
 import {ServiceSourceForm, ServiceDiscoveryForm, ServiceManualForm} from 'NewService'
+import {FormWrapper, BackendApiSelect} from 'NewService/components/FormElements'
+import type {BackendApi} from 'NewService/types'
```
josemigallas: This is really just a personal preference but I think it's better to import types after all the JS imports, since they are not really a real Javascript thing.

app/javascript/src/NewService/types/index.js
```diff
@@ -9,3 +9,8 @@ export type FormProps = {
   submitText: string,
   children?: React.Node
 }
+
+export type BackendApi = {
```
josemigallas: Use type `Api` from `Types` instead (perhaps you'll need to turn `link` and `type` into optionals)

test/integration/api/services_controller_test.rb
```diff
@@ -25,15 +28,15 @@ class Api::ServicesControllerTest < ActionDispatch::IntegrationTest
...
   test 'update the settings' do
-    put admin_service_path(service), update_params
+    put admin_service_path(service), params: update_params
     assert_equal 'Service information updated.', flash[:notice]
 
     update_service_params = update_params[:service]
     expected_notification_settings = update_service_params[:notification_settings].transform_values { |notifications| notifications.map(&:to_i) }
     expected_buyer_plan_change_permission = update_service_params[:buyer_plan_change_permission]
     expected_signup_and_use = update_service_params
-                                  .slice(:intentions_required, :buyers_manage_apps, :referrer_filters_required, :custom_keys_enabled, :buyer_can_select_plan)
```
josemigallas: This is probably your IDE's fault, I think it is more readable with the indentation.

test/integration/api/services_controller_test.rb
```diff
@@ -48,36 +51,96 @@ class Api::ServicesControllerTest < ActionDispatch::IntegrationTest
     new_admin_support_email = 'foo.admin.support@example.com'
 
     deprecated_update_params = { service:
-                                 { tech_support_email: new_tech_support_email,
```
josemigallas: Another unnecessary change IMO, more readable with the indentation.

test/integration/api/services_controller_test.rb
```diff
@@ -48,36 +51,96 @@ class Api::ServicesControllerTest < ActionDispatch::IntegrationTest
...
+
+    assert_change of: -> { BackendApi } do
+      post admin_services_path, service: {
+        system_name: 'my_new_product',
+        name: 'My new Product',
+        description: 'This will act as product',
+        act_as_product: true
+      }
+    end
+
```
josemigallas: Double blank line here.
```suggestion
```

test/integration/api/services_controller_test.rb
```diff
@@ -48,36 +51,96 @@ class Api::ServicesControllerTest < ActionDispatch::IntegrationTest
...
+          email_provider: ['', '50', '100', '150'],
+          web_buyer: ['', '50', '100', '150'],
+          email_buyer: ['', '50', '100', '300']
+        }
+      }
     }
   end
 
-end
+  end
```
josemigallas: Wrong indentation I think:
```suggestion
end
```

thomasmaas:
So, i guess i was out of the loop on this one. In my head services and api as product were never intended to exist side by side and thus this feature of marking a service as a product wouldn't exist.

app/models/backend_api_config.rb
```diff
@@ -11,6 +11,10 @@ class BackendApiConfig < ApplicationRecord
   has_one :proxy, through: :service
   has_many :proxy_rules, through: :proxy
 
+  after_create do
```
hallelujah: Is this part of this PR?

app/javascript/src/NewService/types/index.js
damianpm: @josemigallas means from `app/javascript/src/Types/Api.jsx`

app/javascript/src/NewService/components/NewServiceForm.jsx
```diff
@@ -24,8 +27,8 @@ const NewServiceForm = (props: Props) => {
     setFormMode(event.currentTarget.value)
   }
 
-  const formToRender = () => formMode === 'manual' || !isServiceDiscoveryAccessible
```
damianpm: I remember it (`|| !isServiceDiscoveryAccessible`) was needed to not show Service Discovery feature in SAAS.
@josemigallas you made this fix, remember?
Or it's not needed anymore?

spec/javascripts/NewService/formElements/BackendApiSelect.spec.jsx
```diff
@@ -0,0 +1,40 @@
+// @flow
+
+import React from 'react'
+import {act} from 'react-dom/test-utils'
+import Enzyme, {mount, shallow, render} from 'enzyme'
```
damianpm: You are never using `mount` or `render`

spec/javascripts/NewService/formElements/BackendApiSelect.spec.jsx
```diff
@@ -0,0 +1,40 @@
...
+import Adapter from 'enzyme-adapter-react-16'
+
+import {BackendApiSelect} from 'NewService/components/FormElements'
+import * as utils from 'utilities/utils'
```
damianpm: I don't think you need to import `utils`, the scope of this test is only the `select`, not the whole form.
And, if you import it, you should mock the implementation like [here](https://github.com/3scale/porta/blob/0dd700b542328e28d554aa13bb28d6c390637da8/spec/javascripts/NewService/NewServiceForm.spec.jsx#L12)

spec/javascripts/NewService/formElements/BackendApiSelect.spec.jsx
```diff
@@ -0,0 +1,40 @@
...
+  const wrapper = shallow(<BackendApiSelect {...props} />)
+  expect(wrapper.find('#service_act_as_product_input').exists()).toEqual(true)
+})
+
+it('should render Act as Product checkbox', () => {
+  const wrapper = shallow(<BackendApiSelect {...props} />)
+  expect(wrapper.find('input[name="service[act_as_product]"]').exists()).toEqual(true)
+})
+
+it('should not render the field to select the Backend Api by default', () => {
```
damianpm: Maybe "should not render select element when Act as Product checkbox is unchecked" ?

spec/javascripts/NewService/formElements/BackendApiSelect.spec.jsx
```diff
@@ -0,0 +1,40 @@
...
+  const wrapper = shallow(<BackendApiSelect {...props} />)
+  expect(wrapper.find('select[name="service[backend_api]"]').exists()).toEqual(false)
+})
+
+it('should render the field to select the Backend Api when selecting the Act as Product checkbox', () => {
```
damianpm: Maybe "should render select element when Act as Product checkbox is checked" ?

app/javascript/src/NewService/components/NewServiceForm.jsx
josemigallas: I remember that was added after.. but analyzing the current logic, I think it might be redundant. `formMode` is `'manual'` by default and this **cannot change** when `!isServiceDiscoveryAccessible`. This happens because the radio buttons only render when it is `true`.

In other words: `formMode === 'service'` and `isServiceDiscoveryAccessible === false` is not possible, I think.

damianpm: OK

app/controllers/api/services_controller.rb
```diff
@@ -34,8 +36,8 @@ def settings
 
   def create
     @service = collection.new # this is done in 2 steps so that the account_id is in place as preffix_key relies on it
-    @service.attributes = params[:service]
-    @service.system_name = params[:service][:system_name]
```
guicassolato: Why did `system_name` need to be set separately and now we don't?

app/controllers/api/services_controller.rb
```diff
@@ -65,8 +67,21 @@ def destroy
...
+  def build_backend_api_configs
+    return [] if !provider_can_use?(:api_as_product) || params[:service][:backend_api].blank?
+    backend_api = current_account.backend_apis.find(params[:service][:backend_api])
+    [BackendApiConfig.new(backend_api: backend_api)]
+  end
 
+  def service_params
+    allowed_params = %i[system_name name description]
+    allowed_params << :act_as_product if provider_can_use?(:api_as_product)
+    params.require(:service).permit(allowed_params)
```
guicassolato:
```suggestion
    params.require(:service).permit(*allowed_params)
```

guicassolato: Not really needed as Rails flattens the array.

app/models/backend_api_config.rb
guicassolato: This is because you can choose an existing backend API instead of creating a new one.

PR #1120, opened 2019-08-26
Fix links to actions over backend API metrics on app plan page (THREESCALE-3348)

guicassolato:
Closes [THREESCALE-3348](https://issues.jboss.org/browse/THREESCALE-3348)
Screen Shot 2019-08-26 at 1 02 54 PM

### DIFF
--- a/app/controllers/api/metrics_controller.rb
+++ b/app/controllers/api/metrics_controller.rb
@@ -124,7 +124,7 @@ class Api::MetricsController < Api::BaseController
   end
 
   def find_metric
-    @metric = @service.metrics.find(params[:id])
+    @metric = @service.all_metrics.find(params[:id])
   end
 
   def build_metric
--- a/app/lib/backend_api_logic/service_extension.rb
+++ b/app/lib/backend_api_logic/service_extension.rb
@@ -9,6 +9,12 @@ module BackendApiLogic
       has_many :backend_apis, through: :backend_api_configs
 
       alias_method :backend_api, :first_backend_api
+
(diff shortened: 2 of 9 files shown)


app/lib/backend_api_logic/service_extension.rb
```diff
@@ -9,6 +9,12 @@ module ServiceExtension
       has_many :backend_apis, through: :backend_api_configs
 
       alias_method :backend_api, :first_backend_api
+
+      has_many :backend_api_metrics, through: :backend_api_configs
+
+      has_many :all_metrics, ->(object) do
```
guicassolato: Alternative:

```ruby
def all_metrics
  sql = "((#{metrics.to_sql}) UNION ALL (#{backend_api_metrics.to_sql})) AS metrics"
  Metric.from(sql)
end
```

app/controllers/api/metrics_controller.rb
```diff
@@ -124,7 +124,7 @@ def find_service
   end
 
   def find_metric
-    @metric = @service.metrics.find(params[:id])
+    @metric = @service.all_metrics.find(params[:id])
```
hallelujah: We could probably just do Metric.find and verify in Ruby that it belongs to the Service either directly or via BackendApiConfig

app/lib/backend_api_logic/service_extension.rb
guicassolato: This alternative may not be so good though due to https://www.percona.com/blog/2010/10/25/mysql-limitations-part-3-subqueries/

PR #1123, opened 2019-08-26
[modern browser] logout current user if broswer is not modern

macejmic:

https://issues.jboss.org/browse/THREESCALE-2632

**Verification steps** 

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -91,11 +91,11 @@ class ApplicationController < ActionController::Base
 
   def check_browser
     request_format = request.format
-    return true if request_format.xml? || request_format.json?
+    return if request_format.xml? || request_format.json?
 
-    if current_account&.provider_can_use?(:modern_browser_check) && browser_not_modern?
+    if current_user && browser_not_modern?
       logout_keeping_session!
-      flash.now[:error] = 'To protect you, we cannot let you use the browser you are using. Please upgrade your browser and login again.'
+      flash.now[:error] = "The browser you are using doesn't seem to support the X-Frame-Options header. That means we can't protect you against Cross Frame Scripting and thus not guarantee the security of your session. Please upgrade your browser and sign in again."
       redirect_to provider_admin_path
     end
   end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -27,6 +27,7 @@ class ApplicationController < ActionController::Base
 
   before_action :report_google_experiments, if: proc { ThreeScale::Analytics::GoogleExperiments.enabled? }
   before_action :enable_analytics
+  before_action :check_browser
 
   def status
     begin
@@ -87,6 +88,22 @@ class ApplicationController < ActionController::Base
   end
 
   protected
+
+  def check_browser
+    request_format = request.format
+    return true if request_format.xml? || request_format.json?
+
+    if current_account&.provider_can_use?(:modern_browser_check) && browser_not_modern?
+      logout_keeping_session!
+      flash.now[:error] = 'To protect you, we cannot let you use the browser you are using. Please upgrade your browser and login again.'
+      redirect_to provider_admin_path
+    end
+  end
+
+  def browser_not_modern?
+    browser = Browser.new(request.env['HTTP_USER_AGENT'])
+    browser.known? && !browser.modern?
+  end
 
   def cors
     headers['Access-Control-Allow-Origin'.freeze] =
--- a/Gemfile.base
+++ b/Gemfile.base
@@ -133,6 +133,7 @@ gem 'diff-lcs', '~> 1.2'
 gem 'acts-as-taggable-on', '~> 4.0'
 gem 'whenever', '~> 0.9.7', require: false
 gem 'will_paginate', '~> 3.1.0'
+gem 'browser'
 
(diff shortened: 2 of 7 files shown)


app/controllers/application_controller.rb
```diff
@@ -88,6 +89,21 @@ def sublayout
...
+    if current_account && current_account.provider_can_use?(:modern_browser_check) && browser_not_modern?
+      logout_keeping_session!
+      flash.now[:error] = 'To protect you, we cannot let you use the browser you are using. Please upgrade your browser and login in again.'
+      redirect_to provider_admin_path
+    end
+  end
+
+  def browser_not_modern?
+    browser = Browser.new(request.env['HTTP_USER_AGENT'])
+    browser.known? && !browser.modern?
```
macejmic: not sure about this, shall we logout user if the browser is unknown?

hallelujah: Probably not :)

macejmic: agree.

app/controllers/application_controller.rb
```diff
@@ -88,6 +89,22 @@ def sublayout
 
   protected
 
+  def check_browser
+    request_format = request.format
+    return true if request_format.xml? || request_format.json?
+
+    if current_account&.provider_can_use?(:modern_browser_check) && browser_not_modern?
+      logout_keeping_session!
+      flash.now[:error] = 'To protect you, we cannot let you use the browser you are using. Please upgrade your browser and login again.'
```
thomasmaas:
```suggestion
      flash.now[:error] = "The browser you are using doesn't seem to support the X-Frame-Options header. That means we can't protect you against Cross Frame Scripting and thus not guarantee the security of your session. Please upgrade your browser and login again."
```

app/controllers/application_controller.rb
```diff
@@ -88,6 +89,22 @@ def sublayout
 
   protected
 
+  def check_browser
+    request_format = request.format
+    return true if request_format.xml? || request_format.json?
+
+    if current_account&.provider_can_use?(:modern_browser_check) && browser_not_modern?
```
thomasmaas: I think we should keep the number of switches to a minimum as it only leads to edge cases and complexity. Maybe just enable for all? What do you think @hallelujah @guicassolato?

hallelujah: Agree, we can enable for all yes.

app/controllers/application_controller.rb
kevprice83: Where are we configuring the rules for the modern browser compatibility? Right now it looks like we are just using the defaults from the library which are not sane at all. Look at the firefox, opera & IE defaults, they are years old. If we are setting this somewhere else where those versions are relevant then ignore this comment and please point me to that commit. We should list this in our supported configurations.

kevprice83: Maybe we can use [this](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options#Browser_compatibility) as a reference?

macejmic: @kevprice83 defaults from the library mirror X-frame-Options support, taken from: https://caniuse.com/#feat=x-frame-options

kevprice83: Okay makes sense, the only one that maybe needs review is firefox as the default in the browser gem is [17](https://github.com/fnando/browser/blob/v2.6.1/lib/browser/browser.rb#L87) but in https://caniuse.com/#feat=x-frame-options it is 18. Once we have the final list then I'll speak with Andreu to ensure our supported browser list is up-to-date. Thanks @macejmic

app/controllers/application_controller.rb
```diff
@@ -88,6 +89,22 @@ def sublayout
...
+    return true if request_format.xml? || request_format.json?
+
+    if current_user && browser_not_modern?
+      logout_keeping_session!
+      flash.now[:error] = "The browser you are using doesn't seem to support the X-Frame-Options header. That means we can't protect you against Cross Frame Scripting and thus not guarantee the security of your session. Please upgrade your browser and sign in again."
+      redirect_to provider_admin_path
+    end
+  end
+
+  def browser_not_modern?
```
hallelujah: I'd use the positive  `modern_browser?` here

app/controllers/application_controller.rb
```diff
@@ -88,6 +89,22 @@ def sublayout
 
   protected
 
+  def check_browser
+    request_format = request.format
+    return true if request_format.xml? || request_format.json?
```
hallelujah: No need of returning true right?
```ruby
return if modern_browser? || request_format.xml? || request_format.json?
```

app/controllers/application_controller.rb
macejmic: we would need to return true if a browser is unknown.

macejmic: taking into account that, I would prefer `modern_not_modern?`

app/controllers/application_controller.rb
macejmic: true.

macejmic:
@kevprice83 https://github.com/3scale/porta/pull/1123/files#diff-40a9819bb61b5a4203d5bb7c2c5fdde2

PR #1127, opened 2019-08-27
[DO NOT MERGE] Destroy backend apis belonging to the service when it is deleted

duduribeiro:

Currently we have some orphans backend apis in the database caused
by deleting a service. This generates some problems for some
clients when they tries to recreate the service with the same name
and it fails because an already existent backend api with the same
name already exists. This commit changes this behave to delete
all related backend apis when deleting a service when the client
is in the rolling update.

This also adds a rake task to clear orphans backend apis.

[THREESCALE-3317](https://issues.jboss.org/browse/THREESCALE-3317)

**Verification steps** 

1. Create a service.
2. It will create a service and a backend api belonging to it.
3. Destroy the service.
4. It should also destroy the backend api related to it.
5. Try to create a service with the same name.
6. It should create with success.

:warning: Blocked by [ [THREESCALE-3343](https://issues.jboss.org/browse/THREESCALE-3343) | [#1144](https://github.com/3scale/porta/pull/1144) ]

### DIFF
--- a/app/models/service.rb
+++ b/app/models/service.rb
@@ -198,6 +198,7 @@ class Service < ApplicationRecord
 
     before_transition to: [:deleted], do: :deleted_by_state_machine
     after_transition to: [:deleted], do: :notify_deletion
+    after_transition to: [:deleted], do: :destroy_related_backend_apis
   end
 
   def using_proxy_pro?
@@ -515,6 +516,11 @@ class Service < ApplicationRecord
 
   def archive_as_deleted
     ::DeletedObject.create!(object: self, owner: account)
+  end
+
+  def destroy_related_backend_apis
+    return if act_as_product?
+    backend_api&.destroy
   end
 
   def deleted_by_state_machine
--- a/app/workers/destroy_all_deleted_objects_worker.rb
+++ b/app/workers/destroy_all_deleted_objects_worker.rb
@@ -1,7 +1,7 @@
 class DestroyAllDeletedObjectsWorker
   include Sidekiq::Worker
 
-  def perform(class_name)
-    class_name.constantize.deleted.find_each(&DeleteObjectHierarchyWorker.method(:perform_later))
+  def perform(class_name, method = :deleted)
+    class_name.constantize.public_send(method).find_each(&DeleteObjectHierarchyWorker.method(:perform_later))
   end
 end
--- a/test/unit/service_test.rb
+++ b/test/unit/service_test.rb
@@ -361,9 +361,13 @@ class ServiceTest < ActiveSupport::TestCase
     assert_equal 'published', service_plan.state
   end
 
-  class DestroyingServiceTest < ActiveSupport::TestCase
+  class DestroyServiceTest < ActiveSupport::TestCase
 
     disable_transactional_fixtures!
+
+    def setup
+      Account.any_instance.stubs(:provider_can_use?).returns(true)
+    end
 
     test "destroying service destroys it's plans" do
       service          = FactoryBot.create(:service)
@@ -403,21 +407,6 @@ class ServiceTest < ActiveSupport::TestCase
       event = RailsEventStoreActiveRecord::Event.where(event_type: Services::ServiceDeletedEvent).last!
       assert_equal service_id, event.data['service_id']
     end
-  end
-
-  class CreateServiceTest < ActiveSupport::TestCase
-    disable_transactional_fixtures!
-
-    test 'creating service creates a related event' do
-      User.stubs(current: FactoryBot.create(:simple_user))
-      assert_difference(RailsEventStoreActiveRecord::Event.where(event_type: Services::ServiceCreatedEvent).method(:count)) do
-        FactoryBot.create(:simple_service)
-      end
-    end
-  end
-
-  class DestroyServiceTest < ActiveSupport::TestCase
-    disable_transactional_fixtures!
 
     test 'archive as deleted' do
       account = FactoryBot.create(:simple_provider)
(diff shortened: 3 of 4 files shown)


lib/tasks/backend_api.rake
```diff
@@ -0,0 +1,6 @@
+namespace :backend_api do
+  desc 'clear backend_apis that do not belongs to any service'
+  task :destroy_orphans => :environment do
+    BackendApi.where('id NOT IN (SELECT DISTINCT(backend_api_id) FROM backend_api_configs)').destroy_all
```
hallelujah: As @macejmic said in the tchat, you could use `DestroyAllDeletedObjectsWorker` (needs a slight change to make it work)

app/models/service.rb
```diff
@@ -528,6 +529,11 @@ def archive_as_deleted
     ::DeletedObject.create!(object: self, owner: account)
   end
 
+  def destroy_related_backend_apis
+    return if provider_can_use?(:api_as_product)
+    backend_apis.with_single_service.destroy_all
```
hallelujah: No need to look for `with_single_service`

If it is under `api_as_product` then we do not destroy it, otherwise we just destroy the only one backend_api

```suggestion
    backend_api.&.destroy
```

magnusvage:
Hi @duduribeiro ! We intend to use the product feature heavily once it is released. However, this PR gives me some concerns that backend APIs will be linked on a 1-to-1 basis with services/products. We will need for a backend API to be a standalone service as well as included in 0,1 or many products. Deleting a product should never ever touch the backend API.

Maybe I am misunderstanding it all and/or reading it out of context, in which case you can ignore this comment :)

app/models/backend_api.rb
```diff
@@ -29,6 +29,8 @@ class BackendApi < ApplicationRecord
 
   has_system_name(uniqueness_scope: [:account_id])
 
+  scope :orphans, -> { where('id NOT IN (SELECT DISTINCT(backend_api_id) FROM backend_api_configs)') }
```
hallelujah: Do it with LEFT OUTER JOIN and baby_squeel?

app/models/service.rb
```diff
@@ -528,6 +529,11 @@ def archive_as_deleted
     ::DeletedObject.create!(object: self, owner: account)
   end
 
+  def destroy_related_backend_apis
+    return if provider_can_use?(:api_as_product)
```
hallelujah:
```suggestion
    return if act_as_product?
```

duduribeiro:
@guicassolato and @hallelujah I need a re-review on this one please 😄

app/models/backend_api.rb
```diff
@@ -29,6 +29,8 @@ class BackendApi < ApplicationRecord
 
   has_system_name(uniqueness_scope: [:account_id])
 
+  scope :orphans, -> { joining { services.outer }.where('backend_api_id IS NULL') }
```
makikata:
```suggestion
  scope :orphans, -> { joining { services.outer }.where{ BabySqueel[:backend_api_configs].backend_api_id == nil } }
```
I do not know why this does not work

```ruby
joining { services.outer }.where{ backend_api_configs.backend_api_id == nil }
```

It constructs 🤷‍♂ 

```sql
SELECT `backend_apis`.* FROM `backend_apis` LEFT OUTER JOIN `backend_api_configs` ON `backend_api_configs`.`backend_api_id` = `backend_apis`.`id` LEFT OUTER JOIN `services` ON `services`.`id` = `backend_api_configs`.`service_id` WHERE `backend_api_configs_backend_apis`.`backend_api_id` IS NULL
```

See the `WHERE` clause ➡️  `WHERE `backend_api_configs_backend_apis`.`backend_api_id` IS NULL`

lib/tasks/backend_api.rake
```diff
@@ -0,0 +1,8 @@
+namespace :backend_api do
+  desc 'Destroy all orphan backend apis that does not belongs to any service when api_as_product is disabled'
+  task :destroy_orphans => :environment do
+    Account.find_each do |account|
+      account.backend_apis.orphans.destroy_all unless account.provider_can_use?(:api_as_product)
```
makikata: Or the other way around
```suggestion
BackendApi.orphans.each { |b| b.destroy unless account.provider_can_use?(:api_as_product) }
```

It will make less requests

lib/tasks/backend_api.rake
```diff
@@ -0,0 +1,8 @@
+namespace :backend_api do
+  desc 'Destroy all orphan backend apis that does not belongs to any service when api_as_product is disabled'
+  task :destroy_orphans => :environment do
+    Account.find_each do |account|
```
makikata:
```suggestion
```

test/models/backend_api_test.rb
```diff
@@ -23,4 +23,12 @@ def test_default_api_backend
     @backend_api.private_endpoint = 'https://example.org:3/path'
     assert @backend_api.valid?
   end
+
+  test '.orphans should return backend apis that do not belongs to any service' do
+    FactoryBot.create(:backend_api_config)
```
makikata: I do not get this test :)
If I change the factory it will probably fail

Create a Service, then remove the BackendApiConfig linked to it is sufficent right?

app/models/backend_api.rb
guicassolato: Because it's a `LEFT JOIN` and the `BackendApi` no longer exists.

test/models/backend_api_test.rb
duduribeiro: Makes sense.. I constructed the backend_api_config to be more explicit, but I agree with you. it is better to create the service only.

app/models/backend_api.rb
duduribeiro: it worked.. I tested here.

app/models/service.rb
```diff
@@ -528,6 +529,11 @@ def archive_as_deleted
     ::DeletedObject.create!(object: self, owner: account)
   end
 
+  def destroy_related_backend_apis
+    return if act_as_product?
+    backend_api&.destroy
```
Martouta: We should probably wait for https://github.com/3scale/porta/pull/1144 to be merged and then adapt this PR accordingly.
After that I will do a review 😄

Martouta:
I think that we don't need to do this anymore now that we don't automatically create a backend api for each service even when they are not using APIAP.
However, even if I am wrong, what is clear is that this PR is veeeeeeeeery outdated and the task would need to be done from scratch :smile:

PR #1128, opened 2019-08-28
Fixes index view '/backend_apis'

josemigallas:
Removes the empty vertical navigation showing up in `/backend_apis`.
Also fixes Context Selector `backend: undefined`.

**Before**:
Screen Shot 2019-08-28 at 12 19 49

**After**:
Screen Shot 2019-08-28 at 12 19 36


thomasmaas:
This page is not supposed to exist: the tab on the dashboard covers it. What am I missing?

guicassolato:
Now that we have it in the dashboard and in the context switcher menu, we should remove it.

josemigallas:
To be removed in #1130

PR #1130, opened 2019-08-29
Removes 'index' of Backend APIs

josemigallas:

Once Backend APIs are listed in the Dashboard and in the ContextSelector, the view for index `/backend_apis` has to be removed.

This one removes the route `index`, the method in the controller, the view and the unit test.

[THREESCALE-3333: Remove 'index' page for Backend APIs](https://issues.jboss.org/browse/THREESCALE-3333)

**Verification steps** 

* Go to `/backend_apis` -> it should throw a "no route" error


makikata:
I wonder if it is correct to assume this UI
I'd prefer having it in the menu somewhere also. 
We were discussing with @thomasmaas and @guicassolato of mixing it in the Selector.
In my point of view, it is really confusing.

We should probably keep this view and have another nice way to introduce the backend apis. 
They are not part of Account but still neither entirely to API (as they are in another logic).

guicassolato:
> I'd prefer having it in the menu somewhere also.

Me too, but in this case why not having an index of service/products somewhere in the menu as well?

josemigallas:
I don't see why having an index for products and backends would do any harm. However, services has been always listed in the Dashboard and that's where clients are going to look them for. Since products/backends are the new APIs, I don't think it's confusing to keep them in the Dashboard.

Adding a menu for them could also be confusing. It is supposed to be the first thing a user would see.

thomasmaas:
Dashboard is the index for all API's; both Products and Backends.

guicassolato:
Let's move on with this and have less loosen threads from start. In the future, if we decide to have independent index pages for backend APIs and for products, we can always revisit this.

PR #1131, opened 2019-08-29
Implements basic deletion of Backend APIs

josemigallas:

Implements `destroy` method of Backend APIs. It checks whether the backend is used by any product: if so, it can't be destroyed; otherwise it is destroyed and user is redirected to the Dashboard.

[THREESCALE-3342: Implement basic deletion for Backend API](https://issues.jboss.org/browse/THREESCALE-3342)

**Verification steps** 

1. Go to a Backend API and click on 'edit'
2. At the bottom there's the delete button, confirm the alert and
    * if backend is used by any product: it redirects to edit page and shows an error message
    * if backend is **not** used by any product: it redirects to dashboard and backend is deleted

**TODO**:

- [x] Fix tests

### DIFF
--- a/app/controllers/provider/admin/backend_apis_controller.rb
+++ b/app/controllers/provider/admin/backend_apis_controller.rb
@@ -42,7 +42,12 @@ class Provider::Admin::BackendApisController < Provider::Admin::BaseController
   end
 
   def destroy
-    # TODO
+    if @backend_api.services.empty? && @backend_api.destroy
+      redirect_to provider_admin_dashboard_path, notice: 'Backend API deleted'
+    else
+      flash[:error] = 'Backend API could not be deleted'
+      render :edit
+    end
   end
 
   protected
--- a/test/integration/provider/admin/backend_apis_controller_test.rb
+++ b/test/integration/provider/admin/backend_apis_controller_test.rb
@@ -53,4 +53,23 @@ class Provider::Admin::BackendApisControllerTest < ActionDispatch::IntegrationTe
     assert_response :redirect
     assert_equal 'https://new-endpoint.com:443/p', backend_api.reload.private_endpoint
   end
+
+  test 'delete a backend api with products' do
+    backend_api = @provider.backend_apis[0]
+    assert backend_api.backend_api_configs.any?
+
+    delete provider_admin_backend_api_path(backend_api)
+    assert BackendApi.exists? backend_api.id
+    assert_equal 'Backend API could not be deleted', flash[:error]
+  end
+  
+  test 'delete a backend api without any products' do
+    backend_api = @provider.backend_apis[1]
+    assert_not backend_api.backend_api_configs.any?
+    
+    delete provider_admin_backend_api_path(backend_api)
+    assert_redirected_to provider_admin_dashboard_path
+    assert_not BackendApi.exists? backend_api.id
+    assert_equal 'Backend API deleted', flash[:notice]
+  end
 end


app/controllers/provider/admin/backend_apis_controller.rb
```diff
@@ -42,7 +42,13 @@ def update
   end
 
   def destroy
-    # TODO
+    if @backend_api.backend_api_configs.any?
```
makikata: I'd check directly for `backend_api.services`

app/controllers/provider/admin/backend_apis_controller.rb
```diff
@@ -42,7 +42,13 @@ def update
...
+      flash[:error] = "Backend API could not be deleted"
+      render :edit
+    else
+      @backend_api.destroy
```
makikata: For first version :+1:

test/integration/provider/admin/backend_apis_controller_test.rb
```diff
@@ -53,4 +53,21 @@ class Provider::Admin::BackendApisControllerTest < ActionDispatch::IntegrationTe
     assert_response :redirect
     assert_equal 'https://new-endpoint.com:443/p', backend_api.reload.private_endpoint
   end
+
+  test 'delete a backend api with products' do
+    backend_api = @provider.backend_apis[0]
+    assert backend_api.backend_api_configs.any?
+
+    delete provider_admin_backend_api_path(backend_api)
+    assert_nothing_raised(ActiveRecord::RecordNotFound) { backend_api.reload }
```
makikata: `assert_nothing_raised` is noop so no need to do it. Not much value and removed in higher version of testunit ;)
You can also assert the redirection

app/controllers/provider/admin/backend_apis_controller.rb
guicassolato: Is making `BackendApiConfig` as much transparent as possible the idea behind, @makikata?

test/integration/provider/admin/backend_apis_controller_test.rb
josemigallas: _backend_api.reload_ doesn't mean anything from testing perspective. The gist of this test is to not to throw an error when calling `reload`, isn't it? So why not be as explicit as possible for the sake of future readers?

josemigallas: Ideally there should be a `assert record.exist` or something like that, but the exception is the only way I could think of

test/integration/provider/admin/backend_apis_controller_test.rb
```diff
@@ -53,4 +53,21 @@ class Provider::Admin::BackendApisControllerTest < ActionDispatch::IntegrationTe
     assert_response :redirect
     assert_equal 'https://new-endpoint.com:443/p', backend_api.reload.private_endpoint
   end
+
+  test 'delete a backend api with products' do
+    backend_api = @provider.backend_apis[0]
```
duduribeiro: you can use:
@provider.backend_apis.first

app/controllers/provider/admin/backend_apis_controller.rb
```diff
@@ -42,7 +42,13 @@ def update
   end
 
   def destroy
-    # TODO
+    if @backend_api.services.any?
+      flash[:error] = "Backend API could not be deleted"
+      render :edit
+    else
+      @backend_api.destroy
```
guicassolato: What if the `destroy` returns `false`?

guicassolato:
```ruby
def destroy
  if @backend_api.services.empty? && @backend_api.destroy
    return redirect_to provider_admin_dashboard_path, notice: 'Backend API deleted'
  else
    flash[:error] = "Backend API could not be deleted"
    render :edit
  end
end

Comment (josemigallas) [reply app/controllers/provider/admin/backend_apis_controller.rb]: Then it won't be deleted 😃
OK I'll change it

File: test/integration/provider/admin/backend_apis_controller_test.rb (hunk 3 above)
Comment (Martouta) [reply test/integration/provider/admin/backend_apis_controller_test.rb]: you can do `assert backend_api.reload` or  `assert BackendApi.exists?(backend_api.id)`, or:
```ruby
assert_no_difference(BackendApi.method(:count)) do
  delete provider_admin_backend_api_path(backend_api)
end
```

Martouta: Btw, the tests should test the flash message too 😄

app/controllers/provider/admin/backend_apis_controller.rb
```diff
@@ -42,7 +42,12 @@ def update
   end
 
   def destroy
-    # TODO
+    if @backend_api.services.empty? && @backend_api.destroy
+      redirect_to provider_admin_dashboard_path, notice: 'Backend API deleted'
+    else
+      flash[:error] = "Backend API could not be deleted"
```
Martouta: Imo the flash message should be in i18n 😄 but it isn't a huge problem if it is here 👍

test/integration/provider/admin/backend_apis_controller_test.rb
```diff
@@ -53,4 +53,21 @@ class Provider::Admin::BackendApisControllerTest < ActionDispatch::IntegrationTe
...
+    assert backend_api.backend_api_configs.any?
+
+    delete provider_admin_backend_api_path(backend_api)
+    assert_nothing_raised(ActiveRecord::RecordNotFound) { backend_api.reload }
+  end
+    
+  test 'delete a backend api without any products' do
+    backend_api = @provider.backend_apis[1]
```
Martouta: If you do https://github.com/3scale/porta/pull/1131/files#r319517769, then this one should be either `backend_api = @provider.backend_apis.second` or `backend_api = @provider.backend_apis.last` to use the same everywhere in the test 😄

josemigallas: @duduribeiro @Martouta 
I personally don't like that syntax for this particular case. I prefer using them in the logic where it can have a syntactical meaning. Here it's just an index that coincidentally can be referred as first/second/last but that could change anytime.

Martouta: That's fine for me 👍 😄 Specially because it has the assertion later of ww you need anyway (I meant the `assert/assert_not backend_api.backend_api_configs.any?`)

test/integration/provider/admin/backend_apis_controller_test.rb
josemigallas: How can I do that? @Martouta

Martouta: `assert_equal 'the expected message here', flash[:notice]` and `assert_equal 'the expected message here', flash[:error]`

Martouta: Example:
https://github.com/3scale/porta/blob/d57cbd4db56d1c168dfc4be8386f7704f63c9847/test/integration/developer_portal/invitation_signup_test.rb#L69

app/controllers/provider/admin/backend_apis_controller.rb
josemigallas: If I changed it, I would have to change all the messages in the controller. I suggest we do that in a different PR.

PR #1133, opened 2019-08-29
Make independent mapping rules always enabled when APIAP is on

duduribeiro:

We need to make independent_mapping_rules always enabled for
services acting as product.

[THREESCALE-3341](https://issues.jboss.org/browse/THREESCALE-3341)

Depends on https://github.com/3scale/porta/pull/1156

### DIFF
--- a/app/controllers/api/proxy_rules_controller.rb
+++ b/app/controllers/api/proxy_rules_controller.rb
@@ -44,7 +44,7 @@ class Api::ProxyRulesController < Api::BaseController
   private
 
   def authorize!
-    provider_can_use!(:independent_mapping_rules)
+    current_account.independent_mapping_rules_enabled? || raise(CanCan::AccessDenied)
   end
 
   def proxy
--- a/app/helpers/vertical_nav_helper.rb
+++ b/app/helpers/vertical_nav_helper.rb
@@ -240,7 +240,7 @@ module VerticalNavHelper
     items = []
     items << {id: :configuration,   title: 'Configuration',     path: path_to_service(@service)}
     items << {id: :methods_metrics, title: 'Methods & Metrics', path: admin_service_metrics_path(@service)}
-    items << {id: :mapping_rules,   title: 'Mapping Rules',     path: admin_service_proxy_rules_path(@service)} if provider_can_use?(:independent_mapping_rules)
+    items << {id: :mapping_rules,   title: 'Mapping Rules',     path: admin_service_proxy_rules_path(@service)} if current_account.independent_mapping_rules_enabled?
     items << {id: :settings,        title: 'Settings',          path: settings_admin_service_path(@service)}
   end
 
--- a/app/views/api/integrations/apicast/shared/_mapping_rules.html.slim
+++ b/app/views/api/integrations/apicast/shared/_mapping_rules.html.slim
@@ -1,4 +1,4 @@
-- if current_account.provider_can_use?(:independent_mapping_rules)
+- if current_account.independent_mapping_rules_enabled?
   fieldset.inputs.independent_mapping_rules
     legend
       span
--- a/app/views/shared/provider/navigation/service/_integration.html.slim
+++ b/app/views/shared/provider/navigation/service/_integration.html.slim
@@ -7,7 +7,7 @@
          title: 'Methods & Metrics',
          path: admin_service_metrics_path(@service)
 
-- if current_account.provider_can_use?(:independent_mapping_rules)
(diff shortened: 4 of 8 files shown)


app/controllers/api/proxy_rules_controller.rb
```diff
@@ -44,7 +44,7 @@ def destroy
   private
 
   def authorize!
-    provider_can_use!(:independent_mapping_rules)
+    raise ActiveRecord::RecordNotFound unless service.independent_mapping_rules?
```
makikata:
```suggestion
    raise ActiveRecord::RecordNotFound unless service.act_as_product?
```

app/helpers/vertical_nav_helper.rb
```diff
@@ -240,7 +240,7 @@ def service_integration_items
     items = []
     items << {id: :configuration,   title: 'Configuration',     path: path_to_service(@service)}
     items << {id: :methods_metrics, title: 'Methods & Metrics', path: admin_service_metrics_path(@service)} if @service.act_as_traditional_service?
-    items << {id: :mapping_rules,   title: 'Mapping Rules',     path: admin_service_proxy_rules_path(@service)} if provider_can_use?(:independent_mapping_rules)
+    items << {id: :mapping_rules,   title: 'Mapping Rules',     path: admin_service_proxy_rules_path(@service)} if @service.independent_mapping_rules?
```
makikata:
```suggestion
    items << {id: :mapping_rules,   title: 'Mapping Rules',     path: admin_service_proxy_rules_path(@service)} if @service.act_as_product?
```

app/models/service.rb
```diff
@@ -235,6 +235,10 @@ def backend_authentication_value
     end
   end
 
+  def independent_mapping_rules?
+    act_as_product? || provider_can_use?(:independent_mapping_rules)
```
makikata: Well we do not need this rolling update as it is only for 2.5 patch
On master we can just roll it out for API as Product

app/views/api/integrations/apicast/shared/_mapping_rules.html.slim
```diff
@@ -1,4 +1,4 @@
-- if current_account.provider_can_use?(:independent_mapping_rules)
+- if @service.independent_mapping_rules?
```
makikata:
```suggestion
- if @service.act_as_product?
```

app/views/shared/provider/navigation/service/_integration.html.slim
```diff
@@ -7,7 +7,7 @@
          title: 'Methods & Metrics',
          path: admin_service_metrics_path(@service)
 
-- if current_account.provider_can_use?(:independent_mapping_rules)
+- if @service.independent_mapping_rules?
```
makikata:
```suggestion
- if @service.act_as_product?
```

app/controllers/api/proxy_rules_controller.rb
guicassolato: Not sure about `RecordNotFound` here.

Also... instead of https://github.com/3scale/porta/blob/09fa0ef1cff253740ba44a8ee4929d7a1c789c86/app/controllers/api/proxy_rules_controller.rb#L54-L56 maybe we could use https://github.com/3scale/porta/blob/9529edd7d8d5058ea92ea8f2ee990e36bb51ee21/app/controllers/frontend_controller.rb#L111 with a `before_action`. WDYT?

guicassolato: Hmm. This would mean not being able to use `independent_mapping_rules` on its own (and it is already part of `master`) and force us to roll out APIAP to every account. I'd prefer having the option.

duduribeiro: the RecordNotFound is to mimic the same behave that we had with `provider_can_use!`. wdyt?

good idea.. will move it to a before_action so

hallelujah: There are no account using that on SaaS, why should we have it, and just testing it under APIAP?
Less tests and time consuming/context switching

app/controllers/api/proxy_rules_controller.rb
```diff
@@ -44,7 +44,7 @@ def destroy
   private
 
   def authorize!
-    provider_can_use!(:independent_mapping_rules)
+    provider_can_use!(:api_as_product)
```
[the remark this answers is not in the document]
thomasmaas: I would say `provider_can_use!(:api_as_product)` is enough, unless we have customers on saas who are using this?

PR #1134, opened 2019-08-29
Stop using service_prefix column from Account

duduribeiro:

At first we thought that we needed this for the backend_id of a
Service, but if we are not using the service_prefix as it seems
from the SaaS DB, then the backend_id of the service by default
is just its id.

[THREESCALE-2025](https://issues.jboss.org/browse/THREESCALE-2025)

### DIFF
--- a/app/lib/backend/model_extensions/service.rb
+++ b/app/lib/backend/model_extensions/service.rb
@@ -8,7 +8,7 @@ module Backend
       end
 
       def backend_id
-        preffix_key
+        prefix_key
       end
 
       def update_backend_service
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -6,7 +6,7 @@ class Account < ApplicationRecord
   # Hack to remove all the attributes with names matching the regex from @attributes
   # It is enough for rails not persisting them in actual columns.
   def self.columns
-    super.reject {|column| /\Apayment_gateway_(type|options)|deleted_at|first_admin_id\Z/ =~ column.name }
+    super.reject {|column| /\Apayment_gateway_(type|options)|deleted_at|service_preffix|first_admin_id\Z/ =~ column.name }
   end
 
   # need to reset column information to clear column_names and such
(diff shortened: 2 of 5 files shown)


makikata:
The service_prefix is not used by any of our customer in SaaS
Anyway I think it was an old feature that needs to be removed ( I think it is related to end user plan)

It would be good to check with @davidor on Apisonator side if we would need to change something

duduribeiro:
@hallelujah cool. the idea is to destroy the column in a new PR after deploying this one.

duduribeiro:
@hallelujah / @Martouta, I pushed the changes removing the service_preffix column.

test/unit/models_test.rb
```diff
@@ -22,7 +22,7 @@ class ModelsTest < ActiveSupport::TestCase
       'LatestForumPostsPortlet' => %w[content_type], 'TableOfContentsPortlet' => %w[content_type], 'AuthenticationProvider::GitHub' => %w[branding_state account_type],
       'AuthenticationProvider::Keycloak' => %w[account_type], 'AuthenticationProvider::Auth0' => %w[account_type], 'AuthenticationProvider::Custom' => %w[account_type],
       'AuthenticationProvider::ServiceDiscoveryProvider' => %w[account_type], 'AuthenticationProvider::RedhatCustomerPortal' => %w[account_type],
-      'Account' => %w[credit_card_auth_code credit_card_authorize_net_payment_profile_token credit_card_partial_number],
+      'Account' => %w[credit_card_auth_code credit_card_authorize_net_payment_profile_token credit_card_partial_number service_preffix],
```
Martouta: This shouldn't be necessary because this test calls to `Account.columns` and here you are meant to say which of those active columns we don't need to test the string size for now.
https://github.com/3scale/porta/blob/04ae4112ad465694973b0a43d57f08edd78c8510/test/unit/models_test.rb#L35
But in here (https://github.com/3scale/porta/pull/1134/files#diff-1299db718a4681d9e44c3708ebbab542R9) you are removing `service_preffix` from what `Account.columns` returns so we don't need to discard it from here too because it is already discarded

PR #1137, opened 2019-08-30
Fixing broken alignment text in Product overview page

damianpm:

Fixes text alignment issue in Product overview page

https://issues.jboss.org/browse/THREESCALE-3354

Go to **Your Product > Overview**

**Before:**

> before

**After:**

> after


damianpm:
> I think it should be indented 1 to the left, so that it's at te same level but not inside `- if can? :admin, :plans`

@josemi done!

PR #1138, opened 2019-08-30
[Apicast::ProxySource] inject backend API proxy rules

macejmic:

It allows overriding from Product level the Proxy Rules defined in Backend API

https://issues.jboss.org/browse/THREESCALE-3325

### DIFF
--- a/app/decorators/proxy_rule_decorator.rb
+++ b/app/decorators/proxy_rule_decorator.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+class ProxyRuleDecorator < ApplicationDecorator
+
+  self.include_root_in_json = false
+
+  def pattern
+    pattern_value = super
+    backend_api_path ? File.join('/', backend_api_path, pattern_value) : pattern_value
+  end
+
+  private
+
+  def backend_api_path
+    context[:backend_api_path]
+  end
+end
--- a/lib/apicast/proxy_rules_source.rb
+++ b/lib/apicast/proxy_rules_source.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+module Apicast
+  class ProxyRulesSource
+    attr_reader :proxy
+
+    delegate :backend_api_configs, to: :proxy
+
+    def initialize(proxy)
+      @proxy = proxy
+    end
+
+    def to_hash
+      api_proxy_rules = proxy.proxy_rules
+
+      api_proxy_rules += backend_api_configs.flat_map do |config|
+        config.backend_api.proxy_rules.decorate(context: { backend_api_path: config.path })
+      end
+
+      api_proxy_rules.as_json(root: false)
+    end
+  end
+end
--- a/app/models/proxy.rb
+++ b/app/models/proxy.rb
@@ -103,6 +103,7 @@ class Proxy < ApplicationRecord
 
   delegate :account, to: :service, allow_nil: true
   delegate :provider_can_use?, to: :account, allow_nil: true
+  delegate :backend_apis, :backend_api_configs, to: :service
 
   # This smells of :reek:NilCheck
   def authentication_method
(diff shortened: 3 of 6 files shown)


lib/apicast/proxy_rules_source.rb
```diff
@@ -0,0 +1,26 @@
...
+    def initialize(proxy)
+      @proxy = proxy
+    end
+
+    def to_hash
+      api_proxy_rules = []
+
+      backend_api_configs.each do |config|
+        config.backend_api.proxy_rules.each do |proxy_rule|
+          proxy_rule.pattern = "/#{config.path}#{proxy_rule.pattern}"
```
hallelujah: `File.join`?

hallelujah: This is kinda dangerous .... If there is another process that save the proxy_ruyle we might end up saving unwanted changes

How about decorate it? using config as second parameter of the Decorator?

macejmic: true, true.. apicast decorator could be the thing.. thanks!

app/decorators/proxy_rule_decorator.rb
```diff
@@ -0,0 +1,15 @@
...
+  self.include_root_in_json = false
+
+  def pattern
+    pattern_value = super
+    backend_api_path ? File.join("/#{backend_api_path}", pattern_value) : pattern_value
+  end
+
+  private
+
+  def backend_api_path
```
hallelujah: Does it really matter to memoize a value in a Hash?

app/decorators/proxy_rule_decorator.rb
```diff
@@ -0,0 +1,15 @@
+class ProxyRuleDecorator < ApplicationDecorator
+
+  self.include_root_in_json = false
+
+  def pattern
+    pattern_value = super
+    backend_api_path ? File.join("/#{backend_api_path}", pattern_value) : pattern_value
```
hallelujah:
```suggestion
    backend_api_path ? File.join("/", backend_api_path, pattern_value) : pattern_value
```

app/decorators/proxy_rule_decorator.rb
```diff
@@ -0,0 +1,15 @@
+class ProxyRuleDecorator < ApplicationDecorator
```
hallelujah:
```suggestion
# frozen_string_literal: true
class ProxyRuleDecorator < ApplicationDecorator
```

app/decorators/proxy_rule_decorator.rb
macejmic: not that much..

lib/apicast/proxy_rules_source.rb
```diff
@@ -0,0 +1,23 @@
...
+    delegate :backend_api_configs, to: :proxy
+
+    def initialize(proxy)
+      @proxy = proxy
+    end
+
+    def to_hash
+      api_proxy_rules = []
+
+      backend_api_configs.each do |config|
```
hallelujah: We can also use `flat_map` btw

macejmic: 🍤

lib/apicast/proxy_rules_source.rb
```diff
@@ -0,0 +1,21 @@
...
+    def initialize(proxy)
+      @proxy = proxy
+    end
+
+    def to_hash
+      api_proxy_rules = backend_api_configs.flat_map do |config|
+        config.backend_api.proxy_rules.decorate(context: { backend_api_path: config.path })
+      end
+
+      [proxy.proxy_rules, api_proxy_rules].flatten.as_json(root: false)
```
hallelujah: Oh I was thinking of

```ruby
api_proxy_rules = proxy.proxy_rules
api_proxy_rules += backend_api_configs.flat_map do |config|
  # etc ...
end

```

PR #1139, opened 2019-08-30
Remove white spaces when filtering backends in Safari

josemigallas:

* Removes white space from backends table when filtering in Safari
* Fixes name column width so that table doesn't change size when filtering

[THREESCALE-3358: filtering backends should remove white space](https://issues.jboss.org/browse/THREESCALE-3358)

**Verification steps** 

1. Go to dashboard > Backends tab
2. Filter Backends. Make sure there are some with big names.

Verified in:

- [x] Safari
- [x] Chrome
- [x] Firefox
- [x] IE11

### DIFF
--- a/app/javascript/src/Dashboard/components/ApiFilter.jsx
+++ b/app/javascript/src/Dashboard/components/ApiFilter.jsx
@@ -33,7 +33,12 @@ const ApiFilter = ({ apis, domClass, placeholder = 'All APIs' }: Props) => {
 
       const el = document.getElementById(`${domClass}_${id}`)
       if (el) {
-        el.classList.toggle('hidden', !isFilteredApi)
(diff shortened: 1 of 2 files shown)


app/javascript/src/Dashboard/components/ApiFilter.jsx
```diff
@@ -33,7 +33,12 @@ const ApiFilter = ({ apis, domClass, placeholder = 'All APIs' }: Props) => {
 
       const el = document.getElementById(`${domClass}_${id}`)
       if (el) {
-        el.classList.toggle('hidden', !isFilteredApi)
+        // Toggle method second argument not supported in IE11
```
Martouta: Woooow, right! 😱 

image

PR #1143, opened 2019-09-02
Ignore hidden elements in Capybara to avoid false positives

josemigallas:
Some elements are found in even when the are _not visible_, resulting in false positives.

### DIFF
--- a/features/old/accounts/buyers/personal_details.feature
+++ b/features/old/accounts/buyers/personal_details.feature
@@ -39,8 +39,6 @@ Feature: Personal Details
     And I follow "Settings"
     Then I should not see "Personal Details"
 
-
-  @javascript
   Scenario: User Fields
     Given provider "foo.example.com" has the following fields defined for "User":
       | name                 | required | read_only | hidden |
--- a/features/old/proxy.feature
+++ b/features/old/proxy.feature
@@ -66,15 +66,14 @@ Feature: Proxy integration
     Given I'm using a custom API Backend
     Then I should be able to switch back to using the default API Backend
 
-
-  @javascript
+  @javascript @selenium
   Scenario: Got some fancy policy chain
     And I go to the integration show page for service "one"
     And I press "Start using the latest APIcast"
     When I have policies feature enabled
     And I go to the integration page for service "one"
+    And I toggle "Policies"
     Then I should see the Policy Chain
-
 
   Scenario: Got error message when APIcast registry is not setup properly
     And apicast registry is undefined
@@ -96,6 +95,6 @@ Feature: Proxy integration
     And I save the proxy config
     Then the mapping rules should be in the following order:
       | http_method | pattern | delta | metric |
-      | PUT         | /mixers | 1     | hits   |
-      | GET         | /       | 1     | hits   |
       | POST        | /beers  | 2     | hits   |
+      | GET         | /       | 1     | hits   |
+      | PUT         | /mixers | 1     | hits   |
--- a/features/step_definitions/buyers_management/user_steps.rb
+++ b/features/step_definitions/buyers_management/user_steps.rb
@@ -19,13 +19,13 @@ Then /^I should see button to delete buyer (user "[^"]*")$/ do |user|
 end
 
 Then /^I should not see button to delete buyer (user "[^"]*")$/ do |user|
-  user_rm_form = "//form[@action='#{admin_buyers_account_user_path(user.account, user)}'][@method='post']//input[@name='_method'][@value='delete']"
-  assert has_no_xpath?(user_rm_form)
+  user_rm_form = "//form[@action='#{admin_buyers_account_user_path(user.account, user)}'][@method='post']//input[@name='_method'][@value='delete'][@type='hidden']"
+  assert has_no_xpath?(user_rm_form, visible: true)
 end
 
 Then /^I should see button to delete (user "[^"]*")$/ do |user|
-  user_rm_form = "//form[@action='#{provider_admin_account_user_path(user)}'][@method='post']//input[@name='_method'][@value='delete']"
-  assert has_xpath?(user_rm_form)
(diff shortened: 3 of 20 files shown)


josemigallas:
@thomasmaas Yes indeed. I just rescued this branch from the deepest of my local repo and I have to review it (hence the wip)
It may even not be useful anymore.

features/old/applications/providers/create_application.feature
```diff
@@ -84,7 +84,8 @@ Feature: Create application
        And I fill in "Name" with "UltimateWidget"
        And I fill in "Description" with "Awesome ultimate super widget"
        And I select "second_app_plan" from "Application plan"
-      Then I should see "Create a service plan"
+      # FIXME: this is a false positive, the element is invisible
```
Martouta: @josemigallas Create a Jira issue for this plz and reference it here, because otherwise we will forget for sure

features/old/switches/branding.feature
```diff
@@ -13,7 +13,8 @@ Feature: Branding switch
     When I log in as provider "foo.example.com"
      And I go to the dns settings page
      And I follow "Change"
-    Then I should see the invitation to upgrade my plan
+    # FIXME: this is a false positive, the element is invisible
```
Martouta: Same as https://github.com/3scale/porta/pull/1143/files#r321244858
It can be the same issue 👍

features/old/accounts/buyers/personal_details.feature
```diff
@@ -39,8 +39,6 @@ Feature: Personal Details
     And I follow "Settings"
     Then I should not see "Personal Details"
 
-
-  @javascript
```
Martouta: yaaaay! 💃

features/step_definitions/stats/application_steps.rb
```diff
@@ -13,7 +13,7 @@
     And I log in as provider "foo.example.com"
   }
 
-  click_on(text: /\d Apps?/, match: :first)
+  click_on(text: /\d Applications?/i, match: :first)
```
Martouta: Is the question mark a good idea here? 🤔

features/step_definitions/capcha_steps.rb
```diff
@@ -7,7 +7,7 @@
 end
 
 Then /^I should see the captcha$/ do
-  page.should have_selector(RECAPTCHA_SCRIPT)
+  page.should have_selector(RECAPTCHA_SCRIPT, visible: false)
```
Martouta: I asked Josemi about this, because I find very weird the `I should see the captcha` in the title and then in this line `visible: false`, and he told me that it is because it is a `<script>…</script>` so it is not visible in the browser.

Keeping in mind that he didn't add this, just refactored, ok 👍 Just adding the comment here because the reason is not obvious 😄

features/step_definitions/cms/email_template_steps.rb
```diff
@@ -27,6 +27,5 @@
   t = CMS::EmailTemplate.dup.extend(CMS::EmailTemplate::ProviderAssociationExtension)
 
   content = t.find_default_by_name(name).published
-
-  page.find :xpath, XPath.descendant(:textarea)[XPath.text.is(content)]
+  page.find :xpath, XPath.descendant(:textarea)[XPath.text.is(content)], visible: :hidden
```
Martouta: Similarly to https://github.com/3scale/porta/pull/1143/files#r321299524, Josemi explained that it is because the editor to do the templates is a JS library that uses a hidden textarea (like in background) that is used to send the form.

features/old/applications/providers/create_application.feature
```diff
@@ -84,7 +84,8 @@ Feature: Create application
...
+      # Then I should see "Create a service plan"
```
hallelujah: Or there is some bug elsewhere to prevent it to be visible

features/old/switches/branding.feature
```diff
@@ -13,7 +13,8 @@ Feature: Branding switch
...
+    # Then I should see the invitation to upgrade my plan
```
hallelujah: Same, there is probably a bug that prevents it to be visible

features/old/accounts/buyers/personal_details.feature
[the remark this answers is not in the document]
josemigallas: It's asserting the `*` is present (to identify required fields) and it's hidden by JS. Since now capybara ignores hidden elements, it fails. I'll fix it.

features/step_definitions/navigation_steps.rb
```diff
@@ -10,7 +10,7 @@
 end
 
 When 'I navigate to the accounts page' do
-  click_link(text: /\A\d Accounts?\z/)
+  click_link href: admin_buyers_accounts_path
```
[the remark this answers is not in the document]
josemigallas: Yes but it's supposed to find a link and the critical property IMO is the path not the title. The title can change anytime, it's just a string, the path isn't. I think this is more robust but maybe I'm mistaken.

features/old/proxy.feature
```diff
@@ -66,16 +66,15 @@ Feature: Proxy integration
     Given I'm using a custom API Backend
     Then I should be able to switch back to using the default API Backend
 
-
-  @javascript
+  @javascript @selenium
```
hallelujah: No need of `@selenium` anymore, any `@javascript` is run under a real browser

features/step_definitions/buyers_management/user_steps.rb
```diff
@@ -19,13 +19,13 @@
 end
 
 Then /^I should not see button to delete buyer (user "[^"]*")$/ do |user|
```
hallelujah: I ever wonder the use of those I should NOT see steps

They are old but we should probably strie for getting rid of them (later)

features/old/proxy.feature
josemigallas: Sounds good!

PR #1144, opened 2019-09-02
Implement Mark as Deleted for Backend Api

duduribeiro:
This commit changed the behave of the destroy action from the
Backend API controller to make it delete the object in background.

[THREESCALE-3343](https://issues.jboss.org/browse/THREESCALE-3343)


hallelujah:
So I am thinking of not really use any other column but only do it when the service is not a product

hallelujah:
I do not think this column is needed at all,
If the only service tied to it is an old service

app/controllers/provider/admin/backend_apis_controller.rb
```diff
@@ -42,8 +42,8 @@ def update
   end
 
   def destroy
-    if @backend_api.services.empty? && @backend_api.destroy
-      redirect_to provider_admin_dashboard_path, notice: 'Backend API deleted'
+    if @backend_api.services.empty? && DeleteObjectHierarchyWorker.perform_later(@backend_api)
```
Martouta: The problem of doing it this way is from the moment it is requested to destroy to when it is actually destroyed, there can be a long time (maybe 5 minutes or more) and many things can happen in between:
1. It can be requested to destroy more than once, then that means that we will enqueue it more than once and if there are executed at the same time, we will make many of the things to destroy repeatedly at the same time... not very efficient...
2. It can be seen in the show view and in the index view, and it can be updated in the same time... maybe even with side effects. and it may even have traffic in the meantime...
3. If it fails and for some reason either it does not retry (for example we clean our redis for that deleting queue, which we have already done in the past in case of emergency) or it retries so many times that it won't retry anymore... we will have it kinda silently failed and the backend_api will keep being used normally... and in the backend_api we have nothing changed for us to even identify which backend_api aimed to be deleted.

So IMO we should do it as we are doing for a `service`, having a `state` that becomes `deleted` in the controller:
https://github.com/3scale/porta/blob/dd5c49fd0bc9903d3f6f88b1f0e0d98b4d2962ac/app/controllers/api/services_controller.rb#L64-L68
creating an event after the state has successfully changed:
https://github.com/3scale/porta/blob/dd5c49fd0bc9903d3f6f88b1f0e0d98b4d2962ac/app/models/service.rb#L200
https://github.com/3scale/porta/blob/dd5c49fd0bc9903d3f6f88b1f0e0d98b4d2962ac/app/models/service.rb#L614-L617

the event should contain only the info we need:
https://github.com/3scale/porta/blob/dd5c49fd0bc9903d3f6f88b1f0e0d98b4d2962ac/app/events/services/service_scheduled_for_deletion_event.rb#L1-L13

and then through a subscriber of the event, actually doing the deletion in background:
https://github.com/3scale/porta/blob/dd5c49fd0bc9903d3f6f88b1f0e0d98b4d2962ac/app/lib/event_store/repository.rb#L165

https://github.com/3scale/porta/blob/dd5c49fd0bc9903d3f6f88b1f0e0d98b4d2962ac/app/subscribers/service_deletion_subscriber.rb#L1-L12

Martouta: And I also find it simpler if we always do it the same way than if for every model we follow a different approach

hallelujah:
So decision is to use the mark as deleted because of the points that @Martouta issued.
Those are really valid, and we should just follow what we have in Service

test/integration/provider/admin/backend_apis_controller_test.rb
```diff
@@ -11,11 +13,13 @@ class Provider::Admin::BackendApisControllerTest < ActionDispatch::IntegrationTe
...
   attr_reader :provider
 
-  test '#index' do
+  test '#index does not list deleted backend apis' do
+    FactoryBot.create(:backend_api, account: @provider, state: BackendApi::DELETED_STATE)
     get provider_admin_backend_apis_path
     assert_response :success
-    assert_select '#backend_apis tr', count: @provider.backend_apis.count+1
-    @provider.backend_apis.each do |backend_api|
+    expected_backend_apis = 3 # 1 default + 2 created in setup
```
Martouta: I think it is a safer approach to do: `expected_backend_apis = BackendApi.count - 1`
being the `-1` the one you just created.
This way we don't rely on having 1 or 0 on the default which is something that can change as we have seen and shouldn't be related to this test.

db/migrate/20190904144157_add_state_to_backend_apis.rb
```diff
@@ -0,0 +1,8 @@
+class AddStateToBackendApis < ActiveRecord::Migration
+  disable_ddl_transaction! if System::Database.postgres?
+
+  def change
+    add_column :backend_apis, :state, :string
```
hallelujah: We could add a default here I think. On SaaS:

```
irb(main):001:0> BackendApi.count
=> 37183
```

It will be blocking but reasonably, and in fact it might not block at all, there are very few requests that modifies the private_endpoint

db/migrate/20190904144307_backfill_state_for_backend_apis.rb
```diff
@@ -0,0 +1,11 @@
+class BackfillStateForBackendApis < ActiveRecord::Migration
+  disable_ddl_transaction!
+
+  def change
```
hallelujah: Please extract it to a rake task so we can do it after migration.

app/models/backend_api.rb
```diff
@@ -16,6 +17,7 @@ class BackendApi < ApplicationRecord
   delegate :default_api_backend, to: :class
 
   validates :name,        length: { maximum: 511 }, presence: true
```
hallelujah: So I was wondering

Why do we have 511 characters here????

Martouta: hahahahaha good question :joy: i will check :smile:

Martouta: https://github.com/3scale/porta/blob/7131eb1fb93b6f0be507cfa575e091726a66c1e4/db/schema.rb#L235
https://github.com/3scale/porta/commit/c2204dcdf8cf3e5a9484e635df3bba7544e8b529#diff-e2de550cd7105b0cc383c49f401ca88dR6

db/migrate/20190904144157_add_state_to_backend_apis.rb
```diff
@@ -0,0 +1,8 @@
+class AddStateToBackendApis < ActiveRecord::Migration
+  disable_ddl_transaction! if System::Database.postgres?
+
+  def change
+    safety_assured { add_column :backend_apis, :state, :string, default: :published, null: false }
```
Martouta: This will fill up the DB :wink:

PR #1156, opened 2019-09-04
Remove column act_as_product

hallelujah:
There is no more need to have it since we will just rely on `provider_can_use?(:api_as_product)`

Proxy Rules, Methods & Metrics are present in Product/Service level as well as in BackendApi level
This will cover some use cases that we did not address when specifying
This will also ease the development, tests and documentation as nothing changes for customer point of view for their old services
The migration will be transparent.


app/lib/backend_api_logic/service_extension.rb
```diff
@@ -22,7 +22,7 @@ def find_or_create_first_backend_api!
     end
 
     def act_as_product?
-      act_as_product && account&.provider_can_use?(:api_as_product)
+      provider_can_use?(:api_as_product)
```
Martouta: `account&.provider_can_use?(:api_as_product)`
Because this module belong to the service, not to the account.

features/step_definitions/service_steps.rb
```diff
@@ -62,5 +62,5 @@
 end
 
 Given /^the service acts as product$/ do
```
Martouta: As a first step ok, but for later on, this name has nothing to do with what it actually does anymore 😂

app/lib/backend_api_logic/service_extension.rb
Martouta: Spoke with Hery, I showed him that the test `test/unit/proxy_test.rb:74` fails without the `account&.` but it passes with it:
image
image

He told me that service also implements it:
https://github.com/3scale/porta/blob/c4896133d3e1d53f96cee0bdae80dbb00d80a482/app/lib/logic/rolling_updates.rb#L349-L355

So either the tests is wrong or it should be fixed in the module.
He will fix it 👍 💪

Martouta: Done, good job: 👍 💪 🥇 🕺 
image

app/lib/backend_api_logic/routing_policy.rb
```diff
@@ -4,10 +4,14 @@ module BackendApiLogic
   module RoutingPolicy
     def policy_chain
       chain = super
-      return chain unless service.act_as_product?
+      return chain if using_one_regular_backend?
       Builder.new(service).to_a.concat(chain)
     end
 
+    def using_one_regular_backend?
+      backend_api_configs.where { path.not_in(['', '/']) }.count < 2
```
Martouta: I dunno if this will work in Oracle or not but in Oracle the `''` in path is saved as `null`, and I dunno if this will work if it is going to compare `NULL` with this `''`. For me the safest approach is `backend_api_configs.where { path.not_in([nil, '', '/']) }.count < 2` but not sure if necessary and also i dunno if this has a test somewhere 🤔

app/lib/backend_api_logic/routing_policy.rb
```diff
@@ -19,7 +23,7 @@ def initialize(service)
       def to_a
         rules = backend_api_configs.reordering { sift(:path_desc) }.each_with_object([]) do |config, rules|
```
Martouta:
```suggestion
        rules = backend_api_configs.reordering { sift(:path_desc) }.each_with_object([]) do |config, collection|
```

hallelujah: Bad rebase, thank you
There were some conflicts

features/step_definitions/provider_steps.rb
```diff
@@ -294,3 +294,7 @@ def fill_and_submit_form_to_create_tenant(username:)
     assert_selector('.inline-errors', :text => error_message[:message])
   end
 end
+
+Given /^the account has Service acting as Product$/ do
```
guicassolato:
```suggestion
Given /^the account has (\w+) rolling update (enabled|disabled)$/ do |feature, enabled|
```

features/step_definitions/provider_steps.rb
```diff
@@ -294,3 +294,7 @@ def fill_and_submit_form_to_create_tenant(username:)
...
+  @provider.stubs(:provider_can_use?).with(:api_as_product).returns(true)
```
guicassolato:
```suggestion
  @provider.stubs(:provider_can_use?).with(feature).returns(enabled == 'enabled')
```

features/old/menu/api_menu.feature
```diff
@@ -54,7 +54,7 @@ Feature: API menu
     | Settings                  |
 
   Scenario: Integration sub menu structure for API as Product
-    Given the service acts as product
+    Given the account has Service acting as Product
```
guicassolato: I know this is intended to be readable for dummies, but who are kidding? It's just not anyway. IMHO the code would be easier to understand if it clearly stated what it does and in this case it has to do with a rolling update being enabled.

```suggestion
    Given the account has api_as_product rolling update enabled
```

If you don't agree, it's all right too, but then can we please describe it

```
Given the account uses API as product
```

or

```
Given the account has all services acting as products
```

?

features/old/menu/api_menu.feature
```diff
@@ -64,7 +64,7 @@ Feature: API menu
     | Settings                  |
 
   Scenario: Integration sub menu structure for API as Product without independent mapping rules
-    Given the service acts as product
+    Given the account has Service acting as Product
```
guicassolato:
```suggestion
    Given the account has api_as_product rolling update enabled
```

lib/tasks/apiap.rake
```diff
@@ -1,87 +0,0 @@
-# frozen_string_literal: true
```
guicassolato: Shoot! This was added to stage by mistake. Sorry and thanks for removing it.

app/models/backend_api_config.rb
```diff
@@ -14,9 +14,7 @@ class BackendApiConfig < ApplicationRecord
     backend_api.metrics.each { |metric| metric.send(:sync_backend) }
   end
 
-  sifter :path_desc do
-    System::Database.oracle? ? 'path DESC NULLS LAST' : {path: :desc}
-  end
+  scope :with_subpath, -> { common_query = where.not(path: '/'); System::Database.oracle? ? common_query.where('path is NOT NULL') : common_query.where.not(path: '') }
```
Martouta: This is meant to be temporary so we can merge and unblock many PRs depending on this one.

Martouta:
Merging as it works now and it is blocking many PRs.
Improvements in another PR 👍

Martouta:
Improvements in https://github.com/3scale/porta/pull/1185 😄

PR #1157, opened 2019-09-04
Add Fedora installation instructions into INSTALL.md

duduribeiro:
This commit reorganize some topics from the doc and also includes
the instruction on how to setup the application using Fedora (29).

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -20,13 +20,7 @@ The project relies on a [`Makefile`](https://www.gnu.org/software/make/manual/ht
 make help
 ```
 
-##### Running Tests
-Download and build all the images and run the test suite in the container:
-```bash
-make test
-```
-
-##### Entering a Running Container
+### Entering a Running Container
 Download and build all the images and start a shell session inside the container:
 ```bash
 make bash
@@ -34,7 +28,7 @@ make bash
 
 All the source and dependencies for this project will be in place, making it possible to run porta and the tests from inside the container. See [Run Porta](#run-porta)
 
-##### Running the application
+### Running the application
 
 It's also possible to run the application by using only Docker. Firstly, setup the database by runnning `dev-setup` from your terminal:
 
@@ -66,7 +60,7 @@ make dev-stop
 
 #### Ruby version
 
-The project supports **Ruby 2.3.x**.
+The project supports **Ruby 2.4.x**.
 
 Verify you have a proper version by running on your terminal:
 ```bash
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -194,7 +188,121 @@ Run [NPM](https://www.npmjs.com/) to install all the required Node modules:
...
+
+#### Database (Postgres / MySQL / Oracle)
+
+Postgres, MySQL or Orcacle has to be running for the application to work. The easiest way to do it is in a [Docker](https://www.docker.com/) container by simply running:
+
+```shell
+docker run -d -p 5433:5432 -e POSTGRES_USER=postgres -e POSTGRES_DB=3scale_system_development --name postgres10 circleci/postgres:10.5-alpine
+```
+
+Alternatively you can run Redis directly on your machine by following [this article](https://www.if-not-true-then-false.com/2012/install-postgresql-on-fedora-centos-red-hat-rhel/).
```
hallelujah: Are you sure it is about Redis and not Postgres? Maybe a typo?
Hmmm 2012 article from an unofficial blog, I'd like to avoid that in a README ;)
If you can find something more official from Postgres github account or website that would be better

INSTALL.md
```diff
@@ -194,7 +188,121 @@ Run [NPM](https://www.npmjs.com/) to install all the required Node modules:
...
+
+```shell
+bundle config --local build.mysql2 "--with-ldflags=-L/usr/local/opt/openssl/lib --with-cppflags=-I/usr/local/opt/openssl/include"
+```
+
+and run `bundle install` again.
+
+#### NPM
+
+Run [NPM](https://www.npmjs.com/) to install all the required Node modules:
```
hallelujah: We support NPM version 6.4 (I think) @damianpm @josemigallas to confirm
Though it should work with 6.4+

INSTALL.md
```diff
@@ -194,7 +188,121 @@ Run [NPM](https://www.npmjs.com/) to install all the required Node modules:
...
+ruby -v
+```
+
+> You can use [rbenv](https://github.com/rbenv/rbenv) or [rvm](https://rvm.io/) to install ruby.
+
+#### Node version
+
+The project supports **Version: 8.X.X**.
+
+You might want to use [nvm](https://github.com/creationix/nvm/) to install and work with specific Node versions:
```
hallelujah: OK so this is not enforcing the use of NVM, probably say then to install version 8 of nodejs and ensure that environment variables like `PATH` are setup correctly

INSTALL.md
damianpm: Yes, installing node 8 it comes with 6.9 if I'm not wrong

INSTALL.md
```diff
@@ -194,7 +188,96 @@ Run [NPM](https://www.npmjs.com/) to install all the required Node modules:
...
+
+#### NPM
+
+Run [NPM](https://www.npmjs.com/) to install all the required Node modules:
+
+```bash
+npm install
+```
+
+## Setup Database
```
srijitamukherjee: For the Provider  account, no random password gets generated. 

I executed the following command:
USER_PASSWORD=p MASTER_PASSWORD=p bin/rake db:drop db:setup
>>>> Everything worked fine on the terminal but on the login portal, password 'p' isn't accepted as a valid password. Both the USER_PASSWORD and MASTER_PASSWORD needs to be updated with a password of at least 6 characters, else login fails.
I think it's worth mentioning here.

INSTALL.md
```diff
@@ -194,7 +188,96 @@ Run [NPM](https://www.npmjs.com/) to install all the required Node modules:
...
 ```
 
-#### Setup Database
+## Manual setup on Fedora (29)
+
+### Prerequisites
+
+#### Ruby version
+
+The project supports **Ruby 2.3.x**.
```
Martouta: I think that from now on we are only supposed to support at least `2.4.X`, but I am not sure. @hallelujah knows better.

duduribeiro:
> Not sure, the comment I mentioned is worth going into the file or not.
> Everything else looks good.

@srijitamukherjee I think it makes sense!

INSTALL.md
duduribeiro: 👍

duduribeiro: @Martouta  @hallelujah,  Is it a good idea to have different ruby versions  on CI (>= 2.3) so we can exactly now which versions do we support?

INSTALL.md
Martouta: This is already fixed 😄 https://github.com/3scale/porta/pull/1163

INSTALL.md
duduribeiro: changed it to 2.4

INSTALL.md
hallelujah: This is fixed now in master by https://github.com/3scale/porta/pull/1163

INSTALL.md
hallelujah: Only 2.4 :)
2.3 is EOL and we removed support of it
2.5 is blocked by rails upgrade and many gems update

PR #1158, opened 2019-09-05
[Contract] smart counter cache

macejmic:

One of the most problematic query during tenant deletion is updating `contracts_count`.
```
*************************** 21. row ***************************
     Id: 30653
   User: systemdb
   Host: 10.128.226.31:15651
     db: preview01
Command: Query
   Time: 10
  State: updating
   Info: UPDATE `plans` SET `contracts_count` = COALESCE(`contracts_count`, 0) - 1 WHERE `plans`.`type` IN ('ApplicationPlan') AND `plans`.`id` = 957972
```

Since the whole account is being deleted, there's no need to update this attribute.

https://issues.jboss.org/browse/THREESCALE-2988

**Verification steps** 

### DIFF
--- a/app/models/contract.rb
+++ b/app/models/contract.rb
@@ -22,8 +22,10 @@ class Contract < ApplicationRecord
   after_destroy :destroy_customized_plan
   after_commit :notify_plan_changed
 
-  belongs_to :plan, counter_cache: true
-  validate   :correct_plan_subclass?
+  belongs_to :plan
+  counter_culture :plan, column_name: proc { |object| object.counter_culture_enabled? ? :contracts_count : nil }, column_names: { true => :contracts_count }
+
+  validate :correct_plan_subclass?
   # this breaks nested saving of records, when validating there is no user_account yet, its new record
   # validates_presence_of :user_account
 
@@ -107,6 +109,10 @@ class Contract < ApplicationRecord
   end
 
   delegate :paid?, :to => :plan
+
+  def counter_culture_enabled?
+    !provider_account&.scheduled_for_deletion? && !issuer&.deleted?
+  end
 
   def messenger
     (self.class.name.to_s << "Messenger").constantize
--- a/app/models/plan.rb
+++ b/app/models/plan.rb
@@ -184,7 +184,7 @@ class Plan < ApplicationRecord
   end
 
   def reset_contracts_counter
-    self.class.reset_counters(id, :contracts)
+    Contract.counter_culture_fix_counts(where: { id: id })
   end
 
   def can_be_destroyed?
--- a/Gemfile.base
+++ b/Gemfile.base
@@ -134,6 +134,7 @@ gem 'acts-as-taggable-on', '~> 4.0'
 gem 'whenever', '~> 0.9.7', require: false
 gem 'will_paginate', '~> 3.1.0'
 gem 'browser'
+gem 'counter_culture'
 
 gem 'rack-x_served_by', '~> 0.1.1'
 gem 'deadlock_retry'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -176,6 +176,9 @@ GEM
       activerecord (>= 3.0.0)
     addressable (2.6.0)
       public_suffix (>= 2.0.2, < 4.0)
+    after_commit_action (1.1.0)
+      activerecord (>= 3.0.0)
+      activesupport (>= 3.0.0)
     after_commit_queue (1.1.0)
       rails (>= 3.0)
     aggregate_root (0.3.6)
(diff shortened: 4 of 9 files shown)


hallelujah:
@macejmic What issue does it solve? Could you please describe it?

Martouta:
@hallelujah The 1st query of the 2 mentioned in the Jira issue.
I quote here from the description:

> {noformat}
> *************************** 21. row ***************************
>      Id: 30653
>    User: systemdb
>    Host: 10.128.226.31:15651
>      db: preview01
> Command: Query
>    Time: 10
>   State: updating
>    Info: UPDATE `plans` SET `contracts_count` = COALESCE(`contracts_count`, 0) - 1 WHERE `plans`.`type` IN ('ApplicationPlan') AND `plans`.`id` = 957972
> {noformat}
> 
> 
> {noformat}
> *************************** 11. row ***************************
>      Id: 30643
>    User: systemdb
>    Host: 10.128.226.31:15640
>      db: preview01
> Command: Query
>    Time: 10
>   State: updating
>    Info: DELETE FROM `cms_templates` WHERE `cms_templates`.`provider_id` = 80 AND (`cms_templates`.`type` != 'CMS::LegalTerm')
> {noformat}
> 
> I have checked the reason of the 1st one only, and it is that we have a 'contracts_count' for application_plans. I will explain with an example. Imagine that we have these objects and associations:
> 
> {noformat}
> Tenant ID 2
>   -> Service ID 2
>          -> ApplicationPlan ID 2
>                  -> Cinstance/Application ID 2 (For Developer ID 3)
>                  -> Cinstance/Application ID 3 (For Developer ID 4)
>                  -> Cinstance/Application ID 4 (For Developer ID 5)
>          -> ApplicationPlan ID 3
> {noformat}
> 
> So for ApplicationPlan with ID 2 the field 'contracts_count' is 3, and it is there for the assumption that this field is read much more than updated so it is more efficient this way.
> 
> Then in the deletion in background what it currently happens is (imagine that we make the call to delete service ID 2 in background):
> 
> {noformat}
> Destroy Cinstance/Application ID 2.
> Update ApplicationPlan ID 2 to have 'contracts_count' as '2' now.
> Destroy Cinstance/Application ID 3.
> Update ApplicationPlan ID 2 to have 'contracts_count' as '1' now.
> Destroy Cinstance/Application ID 4.
> Update ApplicationPlan ID 2 to have 'contracts_count' as '0' now.
> Destroy ApplicationPlan ID 2.
> Destroy ApplicationPlan ID 3.
> Destroy Service ID 2.
> {noformat}
> 
> So this problematic query shouldn't be executed at all for this scenario because the plan is about to be destroyed as well anyway.
> 
> However, this execution in background must work no matter what is the object to be destroyed in background, so imagine that we make a call to destroy in background the Developer ID 3 only, and nothing else will be destroyed at all (the service will stay :) and the application plans will stay too :) we don't want to delete them in this example), then what happens right now is:
> 
> {noformat}
> Destroy Cinstance/Application ID 2.
> Update ApplicationPlan ID 2 to have 'contracts_count' as '2' now.
> Destroy Developer ID 3.
> {noformat}
> 
> And in this case the update makes sense because the plan will stay (and because it is updated only once!).

Martouta:
FYI, this PR does not fix all [THREESCALE-2988](https://issues.jboss.org/browse/THREESCALE-2988), but the first half of it 😄 👍

app/models/contract.rb
```diff
@@ -108,6 +110,10 @@ def self.by_plan_type(type)
 
   delegate :paid?, :to => :plan
 
+  def counter_culture_enabled?
+    !provider_account.scheduled_for_deletion?
```
Martouta:
```suggestion
    !provider_account.scheduled_for_deletion? && !issuer.deleted?
```

Because imo it is the same if we are deleting the whole service but not the provider, then the plan is gonna be destroyed anyway so no need to updated the counter

macejmic: true true.

app/models/contract.rb
```diff
@@ -108,6 +110,10 @@ def self.by_plan_type(type)
 
   delegate :paid?, :to => :plan
 
+  def counter_culture_enabled?
+    !provider_account.scheduled_for_deletion? || issuer.deleted?
```
Martouta:
```suggestion
    !provider_account.scheduled_for_deletion? && !issuer.deleted?
```

app/models/plan.rb
```diff
@@ -184,7 +184,7 @@ def reorder!(account, ids)
   end
 
   def reset_contracts_counter
-    self.class.reset_counters(id, :contracts)
+    Contract.counter_culture_fix_counts(where: { id: id })
```
hallelujah: Shouldn't it be 
```suggestion
    Contract.counter_culture_fix_counts(where: { id: id }, only: :contracts)
```

Martouta:
As Gui reminded us, it would be nice to test that in preview with big accounts/services 😄

app/models/plan.rb
macejmic: there's only one counter cache specified, but true, it won't hurt.

hallelujah: So leave it :)

PR #1159, opened 2019-09-05
Updating css-loader to 2.x

damianpm:

Updating `css-loader` to 2.x

Note: a previous [PR](https://github.com/3scale/porta/pull/1151) was reverted because master was failing. Trying again now that master issues are fixed

https://issues.jboss.org/browse/THREESCALE-3303

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3857,12 +3857,6 @@
         "lodash.uniq": "^4.5.0"
       }
     },
-    "caniuse-db": {
-      "version": "1.0.30000990",
-      "resolved": "https://registry.npmjs.org/caniuse-db/-/caniuse-db-1.0.30000990.tgz",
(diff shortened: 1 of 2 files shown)


Martouta:
@damianpm @josemigallas I don't understand why this is ok but https://github.com/3scale/porta/pull/1117 had to be reverted in https://github.com/3scale/porta/pull/1151
I just trust you though 🤷‍♀

damianpm:
> @damianpm @josemigallas I don't understand why this is ok but #1117 had to be reverted in #1151
> I just trust you though 🤷‍♀

@Martouta Because master was failing wit an unknown npm error, so I reverted several PR's. I think the issue was flow, fixed in https://github.com/3scale/porta/pull/1153.
I hope this time we have no issues :)

PR #1160, opened 2019-09-05
Remove 'act_as_product' in favor of rolling update check

josemigallas:
Based onto #1156

### DIFF
--- a/app/javascript/src/NewService/components/FormElements/BackendApiSelect.jsx
+++ b/app/javascript/src/NewService/components/FormElements/BackendApiSelect.jsx
@@ -1,6 +1,6 @@
 // @flow
 
-import React, {useState} from 'react'
+import React from 'react'
 import {Label} from 'NewService/components/FormElements'
 import type {Api} from 'Types/Api'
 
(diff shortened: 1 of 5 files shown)


app/javascript/src/NewService/components/FormElements/BackendApiSelect.jsx
```diff
@@ -1,43 +1,24 @@
...
-}
+const BackendApiSelect = ({backendApis}: Props) => (
+  <li id="service_backend_api_input">
+    <Label
+      htmlFor='service_backend_api'
+      label='Backend API'
+    />
+    <select name="service[backend_api]" id="service_backend_api">
+      <option key='new' value=''>Create a new Backend API</option>
+      {backendApis.map(({id, name}) => <option key={id} value={id}>{name}</option>)}
```
damianpm: Maybe we can check if `backendApis` before:
```suggestion
      {backendApis && backendApis.map(({id, name}) => <option key={id} value={id}>{name}</option>)}
```

spec/javascripts/NewService/formElements/BackendApiSelect.spec.jsx
```diff
@@ -1,40 +1,26 @@
...
-it('should render Act as Product checkbox', () => {
-  const wrapper = shallow(<BackendApiSelect {...props} />)
-  expect(wrapper.find('input[name="service[act_as_product]"]').exists()).toEqual(true)
-})
-
-it('should not render select element when Act as Product checkbox is unchecked', () => {
-  const wrapper = shallow(<BackendApiSelect {...props} />)
-  expect(wrapper.find('select[name="service[backend_api]"]').exists()).toEqual(false)
-})
+it('should render select element for backend apis with an empty option', () => {
```
damianpm: 1st option is showing text "Create a new Backend API" is not empty

spec/javascripts/NewService/formElements/BackendApiSelect.spec.jsx
```diff
@@ -1,40 +1,26 @@
...
 
-it('should render select element when Act as Product checkbox is checked', () => {
-  const wrapper = shallow(<BackendApiSelect {...props} />)
-  expect(wrapper.find('select[name="service[backend_api]"]').exists()).toEqual(false)
-  const eventObj = {currentTarget: {checked: true}}
-  act(() => {
-    wrapper.find('input[name="service[act_as_product]"]').simulate('change', eventObj)
-  })
-  expect(wrapper.find('select[name="service[backend_api]"]').exists()).toEqual(true)
+  expect(select.children()).toHaveLength(backendApis.length + 1)
```
damianpm: Should be good to also test the content not only the length

app/lib/logic/rolling_updates.rb
```diff
@@ -348,9 +348,9 @@ def rolling_update(name)
 
     module Service
       private
-
+        # This smells :reek:NilCheck
```
duduribeiro: just align the # with the method name

app/models/service.rb
```diff
@@ -48,7 +48,7 @@ class Service < ApplicationRecord
   end
 
   def self.columns
-    super.reject { |column| %w[tech_support_email admin_support_email].include?(column.name) }
+    super.reject { |column| %w[act_as_product tech_support_email admin_support_email].include?(column.name) }
```
duduribeiro: 💯

spec/javascripts/NewService/formElements/BackendApiSelect.spec.jsx
josemigallas: _empty option_ refers to one that hasn't any value, which is the case.

app/javascript/src/NewService/components/FormElements/BackendApiSelect.jsx
josemigallas: Since `backendApis` is not optional it won't be ever undefined, as long as we comply with Flow.

spec/javascripts/NewService/formElements/BackendApiSelect.spec.jsx
damianpm: Then maybe "without value"

PR #1164, opened 2019-09-06
Fix Create Application wrong test

josemigallas:

The test passes when it should not. In order to show "create new service plan" there has to be no service plans.

https://issues.jboss.org/browse/THREESCALE-3390

### DIFF
--- a/features/old/applications/providers/create_application.feature
+++ b/features/old/applications/providers/create_application.feature
@@ -78,6 +78,7 @@ Feature: Create application
    Scenario: The service of the selected application plans hasn´t service plan
      Given a service "second" of provider "foo.example.com"
        And a published application plan "second_app_plan" of service "second" of provider "foo.example.com"
+       And the service "second" does not have service plan
        And buyer "bob" is not subscribed to the default service of provider "foo.example.com"
        And I go to the provider side create application page for "bob"
       Then I should see "New Application"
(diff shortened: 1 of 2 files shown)


features/old/applications/providers/create_application.feature
```diff
@@ -78,6 +78,7 @@ Feature: Create application
    Scenario: The service of the selected application plans hasn´t service plan
      Given a service "second" of provider "foo.example.com"
        And a published application plan "second_app_plan" of service "second" of provider "foo.example.com"
+       And the service "second" does not have service plan
```
thomasmaas: I don't understand. Normally there would be a (default) service plan, how would it be possible to create an app against a service that doesn't have a (default) service plan?

hallelujah: This test is to show some link to create a service plan if it happens it is deleted

image

Anyway clicking on this link leads to a 404
We can either remove the test and also the feature or fix it

hallelujah: image

A default service plan is created everytime a Service is created. But in case it is deleted, there was a feature that shows a link to create a service plan. Though this link leads to nowhere

So we either delete this entire test and the feature, or fix the broken link or the broken behaviour (in case the service plan is deleted)

PR #1165, opened 2019-09-06
Fixes "create new service plan" link in New Application form

josemigallas:

In Create Application form, when there isn't any Service Plan associated with a Service, a link is shown "create a new service plan". Currently that link is broken, now this PR sets the `href` of such element with the appropriate URL to the service _Service Plans_ page.

**Verification steps** 

1. Go to New Application
2. Make sure you have a service with no service plans, select a plan from that service
3. Click on "create a new service plan" below the _Service Plan" dropdown and verify it goes to the proper page

### DIFF
--- a/app/helpers/buyers/applications_helper.rb
+++ b/app/helpers/buyers/applications_helper.rb
@@ -6,6 +6,7 @@ module Buyers::ApplicationsHelper
       data-services_contracted='#{ services_contracted(buyer) }'
       data-service_plan_contracted_for_service='#{ service_plan_contracted_for_service(buyer) }'
       data-relation_service_and_service_plans='#{ relation_service_and_service_plans(provider) }'
+      data-create_service_plan_path='#{ create_service_plan_path }'
       data-relation_plans_services= '#{ relation_plans_services(provider) }' >".html_safe
 
   end
@@ -39,6 +40,10 @@ module Buyers::ApplicationsHelper
     provider.application_plans.includes(:service).each_with_object({}) do |application_plan, hash|
       hash[application_plan.id] = application_plan.service.id
     end.to_json
+  end
+
+  def create_service_plan_path
+    admin_service_service_plans_path ':service_id'
   end
 
   def last_traffic(cinstance)
--- a/app/javascript/src/Applications/create_application.js
+++ b/app/javascript/src/Applications/create_application.js
@@ -47,6 +47,14 @@ export class CreateApplication {
     return this.relationPlansServices[this.selectedPlan]
   }
 
+  get createServicePlanPath (): string {
+    return this.metadata.data('create_service_plan_path')
+  }
+
+  getCreateServicePlanPathForService (serviceId: number): string {
+    return this.createServicePlanPath.replace(':service_id', String(serviceId))
+  }
+
   checkSelectedPlan () {
     const service = this.serviceOfSelectedPlan
     const servicePlans = this.getServicePlansForService(service)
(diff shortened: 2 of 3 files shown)


app/helpers/buyers/applications_helper.rb
```diff
@@ -41,6 +42,12 @@ def relation_plans_services(provider)
     end.to_json
   end
 
+  def relation_service_and_create_service_plan_path(provider)
```
hallelujah: We do not need this helper
Just use the view

app/javascript/src/Applications/create_application.js
```diff
@@ -47,6 +47,14 @@ export class CreateApplication {
     return this.relationPlansServices[this.selectedPlan]
   }
 
+  get relationServiceAndCreateServicePlanPath (): {[number]: string} {
+    return this.metadata.data('relation_service_and_create_service_plan_path')
+  }
+
+  getCreateServicePlanPathForService (serviceId: number): string {
+    return this.relationServiceAndCreateServicePlanPath[serviceId]
```
hallelujah: Something like this
in the view:

```ruby
data: {admin_service_service_plans_path: admin_service_service_plans_path(':service_id') }
```

And in the JS:
```javascript
return this.metadata.data('admin_service_service_plans_path').replace(':service_id', number)
```

app/helpers/buyers/applications_helper.rb
josemigallas: The reason why using this is to avoid hardcoding the path `admin_service_service_plans_path`, because Ruby cannot be used from within the JS file.

app/javascript/src/Applications/create_application.js
josemigallas: OK

app/javascript/src/Applications/create_application.js
```diff
@@ -47,6 +47,14 @@ export class CreateApplication {
     return this.relationPlansServices[this.selectedPlan]
   }
 
+  get createServicePlanPath (): string {
+    return this.metadata.data('create_service_plan_path')
+  }
+
+  getCreateServicePlanPathForService (serviceId: number): string {
+    return this.createServicePlanPath.replace(':service_id', String(serviceId))
```
Martouta: Why not directly this?
```javascript
getCreateServicePlanPathForService (serviceId: number): string {
  return this.metadata.data('create_service_plan_path').replace(':service_id', String(serviceId))
}
```
Then you wouldn't need the previous method.

I ask because in this same file, a bit above, you have this:
https://github.com/3scale/porta/blob/ec754ae33e58eae72c969180e9f58b64c6289f3f/app/javascript/src/Applications/create_application.js#L38-L40

So just for the sake of following always the same pattern inside the file 🤔 Besides that url with `:service_id` actually does not point to any valid place.
But this is 100% up to the JS team to decide 😄

app/helpers/buyers/applications_helper.rb
```diff
@@ -6,6 +6,7 @@ def metadata_new_app(buyer, provider)
       data-services_contracted='#{ services_contracted(buyer) }'
       data-service_plan_contracted_for_service='#{ service_plan_contracted_for_service(buyer) }'
       data-relation_service_and_service_plans='#{ relation_service_and_service_plans(provider) }'
+      data-create_service_plan_path='#{ create_service_plan_path }'
```
Martouta: I haven't tried but this might work just with a proc here instead of creating a method that is not useful for anything else. So it would be:
`data-create_service_plan_path='#{ proc { admin_service_service_plans_path ':service_id' } }'`

PR #1173, opened 2019-09-10
[Bugfix][Oracle] path order in routing_policy working for null/empty values

Martouta:
Test it locally with oracle DB for the test: `test/unit/proxy_test.rb:103`
For Oracle the empty string is saved as `NULL`.

### DIFF
--- a/app/lib/backend_api_logic/routing_policy.rb
+++ b/app/lib/backend_api_logic/routing_policy.rb
@@ -17,7 +17,7 @@ module BackendApiLogic
       end
 
       def to_a
-        rules = backend_api_configs.reorder(path: :desc).each_with_object([]) do |config, rules|
+        rules = backend_api_configs.reordering { sift(:path_desc) }.each_with_object([]) do |config, rules|
           rule = Rule.new(config).as_json
           rules << rule if rule
         end
@@ -54,7 +54,7 @@ module BackendApiLogic
         end
 
         def path_to_regex
-          if path.empty?
+          if path.to_s.empty?
             "/.*"
           else
             "/#{path}/.*|/#{path}/?"
--- a/app/models/backend_api_config.rb
+++ b/app/models/backend_api_config.rb
@@ -14,6 +14,10 @@ class BackendApiConfig < ApplicationRecord
     backend_api.metrics.each { |metric| metric.send(:sync_backend) }
   end
 
+  sifter :path_desc do
+    System::Database.oracle? ? 'path DESC NULLS LAST' : {path: :desc}
+  end
+
   delegate :private_endpoint, to: :backend_api
 
   def path=(value)


app/lib/backend_api_logic/routing_policy.rb
```diff
@@ -17,7 +17,7 @@ def initialize(service)
       end
 
       def to_a
-        rules = backend_api_configs.reorder(path: :desc).each_with_object([]) do |config, rules|
+        rules = backend_api_configs.reorder('CASE WHEN path IS NULL THEN 1 ELSE 0 END, path DESC').each_with_object([]) do |config, rules|
```
guicassolato: We can use `COALESCE` instead (with sift) and make it Oracle only just in case.

Martouta: Spoke with Gui, and we discovered that baby squeel does not offer reorder construct yet: https://github.com/rzane/baby_squeel/issues/73

Martouta: Coalesce is not ordering properly for me:
image

Martouta: `backend_api_configs.reorder('CASE WHEN path IS NULL THEN 1 ELSE 0 END, path DESC')` works for the 3 DBs, however it might be a better idea for speed in the other DBs (not sure, so share your opinions) doing the specific query of Oracle only on Oracle so the others do not execute stuff they do not need, like this:
```ruby
reorder = System::Database.oracle? ? 'CASE WHEN path IS NULL THEN 1 ELSE 0 END, path DESC' : {path: :desc}
backend_api_configs.reorder(reorder)
```

Martouta: in fact then i could make it oracle specific like:
```ruby
reorder = System::Database.oracle? ? 'path DESC NULLS LAST' : {path: :desc}
rules = backend_api_configs.reorder(reorder).each_with_object([]) do |config, rules|
# ...
```

Martouta: So these are the 2 options, and both work in all DBs:
```ruby
rules = backend_api_configs.reorder('CASE WHEN path IS NULL THEN 1 ELSE 0 END, path DESC').each_with_object([]) do |config, rules|
# ...
```

```ruby
reorder = System::Database.oracle? ? 'path DESC NULLS LAST' : {path: :desc}
rules = backend_api_configs.reorder(reorder).each_with_object([]) do |config, rules|
# ...
```

Martouta: Right, good 👍 So I will push the last once 👍 It is also definitely more understandable at first glance 😄

Martouta: Just discovered that `reordering` exists now in baby squeel 😄because i tried it locally after seeing https://www.bountysource.com/issues/53307207-reordering-without-explicit-direction-doesn-t-work-with-last-in-rails-4
So I will do the sifter 😄

PR #1176, opened 2019-09-10
Sync backend API 'hits' metrics with apisonator as children of service 'hits'

guicassolato:
Closes [THREESCALE-3399](https://issues.jboss.org/browse/THREESCALE-3399)

### DIFF
--- a/app/lib/backend/model_extensions/metric.rb
+++ b/app/lib/backend/model_extensions/metric.rb
@@ -23,19 +23,31 @@ module Backend
 
       def sync_backend
         execute_per_service do |service|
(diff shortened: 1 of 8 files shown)


app/lib/backend/model_extensions/metric.rb
```diff
@@ -23,20 +23,32 @@ def cache_service_association
 
       def sync_backend
         execute_per_service do |service|
-          ::BackendMetricWorker.sync(service.backend_id, id, system_name)
+          ::BackendMetricWorker.perform_async(*args_to_backend_metric_worker(service))
```
hallelujah: Is this because it may take too long or because we want it to be asynchronous?

PR #1180, opened 2019-09-12
Remove warnings, errors and improve Jest specs

josemigallas:
* Remove errors and warnings that were showing up in the log
* Improve some tests
* Remove some _todo_ and _skip_

### DIFF
--- a/app/javascript/src/Policies/components/PoliciesForm.jsx
+++ b/app/javascript/src/Policies/components/PoliciesForm.jsx
@@ -10,6 +10,7 @@ import type { ChainPolicy } from 'Policies/types/Policies'
 import type { UpdatePolicyConfigAction } from 'Policies/actions/PolicyConfig'
 
 class PolicyForm extends Form {
+  // TODO: this method will be renamed to UNSAFE_componentWillReceiveProps in 1.9 and eventually removed.
   componentWillReceiveProps (nextProps: Props) {
     if (nextProps.schema !== this.state.schema) {
       this.setState({...this.getStateFromProps(nextProps), ...{errors: []}})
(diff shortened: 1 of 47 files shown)


spec/javascripts/Stats/inlinechart/inlineChart.spec.jsx
```diff
@@ -89,26 +91,13 @@ describe('<InlineChart/>', () => {
   })
 
   it('calls updateState method with total of 1', () => {
-    const data = {
```
damianpm: I think this tests is useless, after removing `data`, you are calling it with the default data object where `total: 10`, then it's just asserting `toHaveBeenCalled` but not with `data:1`

spec/javascripts/Stats/inlinechart/inlineChart.spec.jsx
```diff
@@ -89,26 +91,13 @@ describe('<InlineChart/>', () => {
...
-    }
-
     wrapper.instance().generateC3Chart = jest.fn()
     wrapper.update()
     wrapper.instance().updateState(data)
     expect(wrapper.instance().generateC3Chart).toHaveBeenCalled()
   })
 
   it('calls updateState method with total > 1', () => {
-    const data = {
```
damianpm: Same here, it's not asserting `total > 1`

spec/javascripts/Stats/inlinechart/inlineChart.spec.jsx
josemigallas: You're right, I didn't realize the "total" was different.. the test is wrong then, it should fail. I will fix it.

app/javascript/src/Policies/components/PoliciesForm.jsx
```diff
@@ -10,6 +10,7 @@ import type { ChainPolicy } from 'Policies/types/Policies'
 import type { UpdatePolicyConfigAction } from 'Policies/actions/PolicyConfig'
 
 class PolicyForm extends Form {
+  // TODO: this method will be renamed to UNSAFE_componentWillReceiveProps in 1.9 and eventually removed.
```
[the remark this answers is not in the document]
josemigallas: Done

PR #1188, opened 2019-09-13
Fix `Folded services have no overview data` scenario test

duduribeiro:
This commit fixes this scenario in the cucumber test. Apparently,
it never worked right.

You can test by checking the commit that introduced it and running
```
git checkout 832920cf
features/old/services/multiservice.feature:61
```


hallelujah:
Probably because it is not loaded yet?
Can you make it fail (without your commits) and make a screenshot of the page?

Martouta:
Fixed this morning in https://github.com/3scale/porta/pull/1191

PR #1190, opened 2019-09-13
Improve Tabs in Dashboard

josemigallas:

- [x] Remember what tab is selected
- [x] Replace current tabs with PF4 ones
- [x] Fix hover style

after

~⚠️**PF4 Styles**⚠️ There's still an issue with the hover style. For some reason I can't get the [::after style](https://github.com/3scale/porta/pull/1190/files#diff-b04944eeb7c91095409e39e62cf6cf8cR54) work, despite the fact that it works for `.pf-m-current`.~

[THREESCALE-3357: Improve product and backend tabs on dashboard](https://issues.jboss.org/browse/THREESCALE-3357)

**Verification steps** 

* Feels and looks like [PF4 tabs](https://pf4.patternfly.org/components/Tabs/examples/)
* After refreshing the page, same tab is selected (it doesn't show Products by default)

### DIFF
--- a/app/assets/stylesheets/provider/_theme.scss
+++ b/app/assets/stylesheets/provider/_theme.scss
@@ -107,3 +107,4 @@
 @import 'provider/print';
 
 @import 'provider/vertical_nav';
+@import 'provider/patternfly_tabs_fake';
--- a/app/assets/stylesheets/provider/admin/_dashboard.scss
+++ b/app/assets/stylesheets/provider/admin/_dashboard.scss
@@ -429,14 +429,17 @@ a:hover .u-notice {
   }
 
   &--content {
-    & > input,
-    #products,
-    #backends {
+    & > input {
       display: none;
     }
 
-    & > input:checked ~ #products,
-    & > input:checked ~ #backends {
+    & > #products,
+    & > #backends {
+      display: none;
+    }
+
+    & > #products.active,
+    & > #backends.active {
       display: block;
     }
 
--- a/app/assets/stylesheets/provider/admin/_dashboard.scss
+++ b/app/assets/stylesheets/provider/admin/_dashboard.scss
@@ -31,10 +31,6 @@
         }
       }
     }
-
-    .pf-c-tabs {
-      padding-bottom: 1rem;
-    }
   }
 
   &-title {
--- a/app/assets/stylesheets/provider/admin/_dashboard.scss
+++ b/app/assets/stylesheets/provider/admin/_dashboard.scss
@@ -31,6 +31,10 @@
         }
       }
     }
+
+    .pf-c-tabs {
+      padding-bottom: 1rem;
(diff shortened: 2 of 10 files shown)


app/assets/stylesheets/provider/admin/_dashboard.scss
```diff
@@ -31,6 +31,10 @@
         }
       }
     }
+
+    // .pf-c-tabs {
+    //   padding-bottom: 1rem;
+    // }
```
thomasmaas: maybe remove this?

thomasmaas:
@damianpm I know it's not PF compliant but could we make the vertical distance between APIs and the horizontal line the same as Audience? And then subsequently lower the words in the tabs and make them all caps the same as the menu items in audience?

app/assets/stylesheets/provider/admin/_dashboard.scss
damianpm: Didn't see that, removed @thomasmaas

damianpm:
> @damianpm I know it's not PF compliant but could we make the vertical distance between APIs and the horizontal line the same as Audience? And then subsequently lower the words in the tabs and make them all caps the same as the menu items in audience? And while at it, also add some horizontal padding inside the tabs so the content has some space to breath.
> 
> Screenshot 2019-09-19 10 58 39

Updated @thomasmaas 

after

app/assets/stylesheets/provider/_theme.scss
```diff
@@ -107,3 +107,4 @@
 @import 'provider/print';
 
 @import 'provider/vertical_nav';
+@import 'provider/patternfly_tabs_fake';
```
didierofrivia: Why "fake" ?

app/javascript/src/Dashboard/tabs-widget.jsx
```diff
@@ -1,17 +1,36 @@
-// @flow
+const TAB_PRODUCTS = 'tab-products'
+const TAB_BACKENDS = 'tab-backends'
+
+function updateActiveTab (activeTab, inactiveTab) {
```
didierofrivia: I believe this widget exists because the dashboard tabs are rendered by rails and we are just adding the interactive functionality... it's a bit pitty tho..
Also, the code is very imperative... maybe we could enhance it managing a list and toggling the active class and defining the `active` one at initialization or by rails. 
I know PF needs all those classes and the tree rendered properly, but that doesn't mean we just can't make it simpler and the `addEventListener` part can be simplified for sure since its duplicated code.
Give it a thought and assess whether its possible to do it on time or not.

damianpm: @didierofrivia AFAIK it needs to be merged ASAP.
I'd rather create a new task to improve it in the near future.

PR #1191, opened 2019-09-16
Fixes test by making assertion case insensitive

josemigallas:
Some title in system are upper case and Capybara is case-sensitive by default. Adding a regex with the flag `/i` fixes this issue.

Error:
```
failed Folded services have no overview data

    Scenario: Folded services have no overview data

And service "Fancy API" is folded

Message:

    Unable to find visible css "section" with text "Fancy API" within #<Capybara::Node::Element tag="section" path="/html/body/div[1]/div/main/div/div[1]/section[2]"> (Capybara::ElementNotFound)
```

EDIT
And also it's required to specify `visible: false` so that the element is found and asserted.
Using `visible: :hidden` could make the test throw an error, which would be also valid but not so elegant.

### DIFF
--- a/features/step_definitions/dashboard_steps.rb
+++ b/features/step_definitions/dashboard_steps.rb
@@ -7,14 +7,14 @@ def service_for_name (name)
   page.find("#service_#{service_id}")
 end
 
-def hits_for_name (name)
+def hits_for_name (name, opts = {})
   service_id = service_id_for_name(name)
-  page.find_by_id("dashboard-widget-service_id-#{service_id}service_hits")
+  page.find_by_id("dashboard-widget-service_id-#{service_id}service_hits", opts)
 end
 
-def top_traffic_for_name (name)
+def top_traffic_for_name (name, opts = {})
   service_id = service_id_for_name(name)
-  page.find_by_id("dashboard-widget-service_id-#{service_id}service_top_traffic")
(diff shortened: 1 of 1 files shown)


thomasmaas:
@josemigallas There is no way to make capybara case insensitive at the config level so all tests become case insensitive?

josemigallas:
@thomasmaas 
From the documentation:
>Note: All searches in Capybara are case sensitive. This is because Capybara heavily uses XPath, which doesn't support case insensitivity.

And a quick search didn't reveal any setting to turn this off. My bet is no, it is not possible. In any case, we should merge it to fix master and create a ticket to perhaps investigate this further.

Martouta:
> we could override _find_ methods

To override ActiveRecord methods usually means asking for problems 😄

josemigallas:
But it isn't, it's a method from [Capybara](https://www.rubydoc.info/github/jnicklas/capybara/Capybara/Node/Finders#find-instance_method).

Martouta:
Ah, cool, less bad then, but still better the idea of making the `find_i`. Less problematic 😄

Martouta:
also, not your fault but if we want to find something, we should not find by id and then assert the error that it is not there because we can also change the id and make the test pass although not for the right reason XD in that case we should find by whatever is visible when it renders...
but as i said, it is not your fault and it is not quick

PR #1194, opened 2019-09-16
Fix missing configs in Custom Policies

josemigallas:
NPM silently installed `react-jsonschema-form@1.8.0` which has a change in their `Form` component that breaks our implementation.

Since we rely on `1.3.0`, let's install that _exact_ version.

fixes https://issues.jboss.org/browse/THREESCALE-3475

- [ ] check if this was shipped in 2.6 (i guess not but would be good to be sure)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -940,15 +940,6 @@
         "regenerator-runtime": "^0.13.2"
       }
     },
-    "@babel/runtime-corejs2": {
-      "version": "7.5.5",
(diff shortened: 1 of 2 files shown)


Martouta:
This is what we have in 2.6 (I think):
https://github.com/3scale/porta/blob/d41e5bdbd6fccd493de26e0c06eb5c5b9052a0cd/package-lock.json#L17107-L17109

So it is important to check if it works there.

thomasmaas:
@josemigallas please make sure to check if this bug was shipped in 2.6

PR #1195, opened 2019-09-16
Make registry_policies available for Edit and Update actions

duduribeiro:
Currently when something happens when trying to update something
in the integration page, it fails and the registry_policies is
not called because it is only set in the edit action. This commit
changes it and makes it available as a helper method and the view
only calls the method instead of relying on a variable saved in
the action.

### DIFF
--- a/app/controllers/api/integrations_controller.rb
+++ b/app/controllers/api/integrations_controller.rb
@@ -13,21 +13,14 @@ class Api::IntegrationsController < Api::BaseController
   rescue_from ActiveRecord::StaleObjectError, with: :edit_stale
 
   def edit
-    begin
-      @registry_policies = Policies::PoliciesListService.call!(current_account, proxy: @proxy)
-    rescue StandardError => error
-      @error = error
-    end
-    @latest_lua = current_account.proxy_logs.first
-    @deploying =  ThreeScale::TimedValue.get(deploying_hosted_proxy_key)
-    @ever_deployed_hosted = current_account.hosted_proxy_deployed_at.present?
+    fetch_registry_policies
   end
 
   def settings; end
 
   def update
     if @service.using_proxy_pro? && !@proxy.apicast_configuration_driven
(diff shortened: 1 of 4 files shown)


thomasmaas:
There must have been a recent commit that caused this. Did we pinpoint it?

Martouta:
The next thing to find out imo is if this also happens in 2.6

duduribeiro:
> There must have been a recent commit that caused this. Did we pinpoint it?

This never worked. I tested in 2.6. It also never re-displayed the policies after some update fails

josemigallas:
This should point to `master` @duduribeiro

app/controllers/api/integrations_controller.rb
```diff
@@ -13,21 +13,14 @@ class Api::IntegrationsController < Api::BaseController
...
-    @ever_deployed_hosted = current_account.hosted_proxy_deployed_at.present?
+    set_registry_policies
   end
 
   def settings; end
 
   def update
     if @service.using_proxy_pro? && !@proxy.apicast_configuration_driven
-      proxy_pro_update
+      return redirect_to(edit_path) if proxy_pro_update
```
hallelujah: I think we can remove this line, proxy_pro was removed. Some dust to clean :)

app/controllers/api/integrations_controller.rb
```diff
@@ -133,6 +125,12 @@ def toggle_apicast_version
 
   protected
 
+  def set_registry_policies
```
Martouta:
```suggestion
  def fetch_registry_policies
```

app/controllers/api/integrations_controller.rb
Martouta: probably remove the whole method of `proxy_pro_update`

duduribeiro:
Fixed by: https://github.com/3scale/porta/pull/1228

PR #1184, opened 2019-09-13
[API Endpoints] Metrics under Backend API

Martouta:
Part of [THREESCALE-3209](https://issues.jboss.org/browse/THREESCALE-3209)

Done like we do for metrics of services.
Methods in another PR in https://github.com/3scale/porta/pull/1201 because otherwise this one would be over 1000 lines 😄 

- [X] [Changes from previous API endpoints](https://github.com/3scale/porta/pull/1184#issuecomment-532258751)
- [X] New endpoints for metrics under backend api

### New endpoints
Remember that you need the rolling update called `api_as_product` enabled to be able to test this locally in `{{your-url}}/p/admin/api_docs`

#### Index
image

image

- Request: `curl -v  -X GET "http://provider-admin.example.com.lvh.me:3000/admin/api/backend_apis/2/metrics.json?access_token=2793672eac7366b9e9502de4188d0f40f1b6a3e5144fdc1c3617ad7d3646eae7"`

- Response:
```json
{
 "metrics": [
  {
   "metric": {
    "id": 7,
    "system_name": "hits.2",
    "friendly_name": "Hits",
    "description": "Number of API hits",
    "unit": "hit",
    "created_at": "2019-09-17T21:17:25Z",
    "updated_at": "2019-09-17T21:17:25Z",
    "links": [
     {
      "rel": "backend_api",
      "href": "http://provider-admin.example.com/admin/api/backend_apis/2"
     },
     {
      "rel": "self",
      "href": "http://provider-admin.example.com/admin/api/backend_apis/2/metrics/7"
     }
    ]
   }
  }
 ]
}
```

#### Show
image

image

- Request: `curl -v  -X GET "http://provider-admin.example.com.lvh.me:3000/admin/api/backend_apis/2/metrics/7.json?access_token=2793672eac7366b9e9502de4188d0f40f1b6a3e5144fdc1c3617ad7d3646eae7"`

- Response:
```json
{
 "metric": {
  "id": 7,
  "system_name": "hits.2",
  "friendly_name": "Hits",
  "description": "Number of API hits",
  "unit": "hit",
  "created_at": "2019-09-17T21:17:25Z",
  "updated_at": "2019-09-17T21:17:25Z",
  "links": [
   {
    "rel": "backend_api",
    "href": "http://provider-admin.example.com/admin/api/backend_apis/2"
   },
   {
    "rel": "self",
    "href": "http://provider-admin.example.com/admin/api/backend_apis/2/metrics/7"
   }
  ]
 }
}
```

[...]

PR #1197, opened 2019-09-17
Refactor Test ApiDocs only show APIAP with RU enabled (to be dynamic)

Martouta:
This way it tests all APIAP new API endpoints dynamically 😄 
It was originally part of https://github.com/3scale/porta/pull/1184 but that PR is too big now and everything that can be extracted to other PRs, better to do so 😄

### DIFF
--- a/test/integration/api_docs/services_controller_test.rb
+++ b/test/integration/api_docs/services_controller_test.rb
@@ -76,18 +76,19 @@ class ApiDocs::ServicesControllerTest < ActionDispatch::IntegrationTest
 
     test 'show backend_api endpoints only under rolling update enabled' do
       Logic::RollingUpdates.stubs(enabled?: true)
-      select_endpoint_id = Proc.new { |api| api['path'] == '/admin/api/backend_apis/{id}.json' }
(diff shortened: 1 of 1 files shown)


test/integration/api_docs/services_controller_test.rb
```diff
@@ -74,21 +74,22 @@ def test_show_service_management
...
+      assert_equal actual_backed_api_routes.length, JSON.parse(response.body)['apis'].each_with_object(Set.new, &select_endpoints).length
 
       Logic::RollingUpdates::Features::ApiAsProduct.any_instance.stubs(enabled?: false)
-      get '/api_docs/services/account_management_api.json'
-      assert_empty JSON.parse(response.body)['apis'].select(&select_endpoint_id)
-      assert_empty JSON.parse(response.body)['apis'].select(&select_endpoint_collection)
-    end
+			get '/api_docs/services/account_management_api.json'
+			assert_empty JSON.parse(response.body)['apis'].each_with_object(Set.new, &select_endpoints)
+		end
```
guicassolato:
```suggestion
  end
  
  private
  
  def backed_api_routes
    Rails.application.routes.named_routes.select { |name, route| name.to_s.match(/\Aadmin_api_backend_api.*\z/) }
  end
```

PR #1200, opened 2019-09-17
Include Rake task to destroy orphans backend apis

duduribeiro:
Although the deletion process of Services is async, even after
several hours or even days a given Service cannot be created again
using the same system_name from a previously deleted one. This
happens because the backend api is still created. Some customers
reported this bug.

This commit includes a Rake task to allow us to run this in cli
to destroy orphans backend apis

### DIFF
--- a/app/models/backend_api.rb
+++ b/app/models/backend_api.rb
@@ -30,6 +30,8 @@ class BackendApi < ApplicationRecord
 
   has_system_name(uniqueness_scope: [:account_id])
 
+  scope :orphans, -> { joining { services.outer }.where { BabySqueel[:backend_api_configs].backend_api_id == nil } }
+
   scope :oldest_first, -> { order(created_at: :asc) }
 
   def self.default_api_backend
--- a/test/models/backend_api_test.rb
+++ b/test/models/backend_api_test.rb
@@ -30,6 +30,18 @@ class BackendApiTest < ActiveSupport::TestCase
     assert @backend_api.valid?
   end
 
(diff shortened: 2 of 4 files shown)


Martouta:
Patch for [THREESCALE-3490](https://issues.jboss.org/browse/THREESCALE-3490) 👍 💪 🥇

app/models/backend_api.rb
```diff
@@ -30,6 +30,8 @@ class BackendApi < ApplicationRecord
 
   has_system_name(uniqueness_scope: [:account_id])
 
+  scope :orphans, -> { joining { services.outer }.where{ BabySqueel[:backend_api_configs].backend_api_id == nil } }
```
Martouta:
```suggestion
  scope :orphans, -> { joining { services.outer }.where { BabySqueel[:backend_api_configs].backend_api_id == nil } }
```
For this:
image

test/models/backend_api_test.rb
```diff
@@ -30,6 +30,16 @@ def test_default_api_backend
...
   end
 
+  test '.orphans should return backend apis that do not belongs to any service' do
+    FactoryBot.create(:service)
+    service_to_delete = FactoryBot.create(:service)
+    orphan_backend_api = service_to_delete.backend_api_configs.first.backend_api
+
+    service_to_delete.destroy!
+
+    assert_equal [orphan_backend_api], BackendApi.orphans
```
hallelujah: Well there is only one backend api in the DB so this test does not really compare anything

PR #1203, opened 2019-09-18
[plan] manual counter cache

macejmic:

> Replace this comment with references to related issues. This will ensure the issue(s) are closed automatically when the PR gets merged.
> e.g. `fixes #<issue number>(, fixes #<issue_number>, ...)`

**Verification steps** 

### DIFF
--- a/app/models/cinstance.rb
+++ b/app/models/cinstance.rb
@@ -5,7 +5,7 @@ class Cinstance < Contract
 
   delegate :backend_version, to: :service, allow_nil: true
 
-  belongs_to :plan, class_name: 'ApplicationPlan', foreign_key: :plan_id, inverse_of: :cinstances, counter_cache: :contracts_count
+  belongs_to :plan, class_name: 'ApplicationPlan', foreign_key: :plan_id, inverse_of: :cinstances
   alias application_plan plan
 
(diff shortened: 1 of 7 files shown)


app/models/cinstance.rb
```diff
@@ -5,7 +5,7 @@ class Cinstance < Contract
 
   delegate :backend_version, to: :service, allow_nil: true
 
-  belongs_to :plan, class_name: 'ApplicationPlan', foreign_key: :plan_id, inverse_of: :cinstances, counter_cache: :contracts_count
```
hallelujah: So I have a simpler solution, a bit hacky and might break future Rails upgrade so just add good tests with it:

See
https://github.com/rails/rails/blob/4-2-stable/activerecord/lib/active_record/associations/builder/belongs_to.rb#L32

And 

https://github.com/rails/rails/blob/v4.2.11.1/activerecord/lib/active_record/associations/builder/belongs_to.rb#L58

```ruby
belongs_to :plan, class_name: 'ApplicationPlan', foreign_key: :plan_id, inverse_of: :cinstances, counter_cache: :contracts_count

def belongs_to_counter_cache_after_update(reflection)
   return true if your_condition
   super
end
```

PR #1204, opened 2019-09-18
[WIP] Factory metric without having to pass service_id: nil

Martouta:
Otherwise when we create a metric for a backend api, it creates a service anyway, and we could simply pass `service: nil` but it is not very intuitive then.

### DIFF
--- a/app/lib/api_docs/provider_data.rb
+++ b/app/lib/api_docs/provider_data.rb
@@ -62,7 +62,7 @@ module ApiDocs
     end
 
     def metrics
-      @metrics ||= @account.metrics.includes(:service).top_level
+      @metrics ||= @account.top_level_metrics.includes(:owner)
     end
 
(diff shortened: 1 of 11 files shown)


test/factories/simple.rb
```diff
@@ -128,4 +128,9 @@
     api_backend { 'http://api.example.net:80' }
     secret_token { '123' }
   end
+
+  factory(:simple_metric, class: Metric) do
```
hallelujah: There is a presence validation on owner

Martouta:
Your idea of changing the metric factory to use owner instead of service is what i first tried but it is a loooot of work and we don't have time for it right now. I know it seems simple and that's why I tried, but it isn't, it breaks a lot...
What i can do for now however is close this PR, create a jira issue for doing it that way later on and right now just stick to creating metrics like `FactoryBot.create(:metric, owner: ..., service_id: nil)` 🤷‍♀ 
WDYT?

Martouta:
I will take the other approach 😄

PR #1205, opened 2019-09-18
[Bugfix] Cuke Feature cms/advanced/groups_and_sections

Martouta:
And in CircleCI it fails particularly [in Oracle](https://circleci.com/gh/3scale/porta/100521#tests/containers/20) 🤷‍♀ 

#### Verification steps
Just run locally `DATABASE_URL="oracle-enhanced://rails:railspass@127.0.0.1:1521/systempdb" bundle exec cucumber features/old/cms/advanced/groups_and_sections.feature:25` 😄 
But anyway you can see the results in CircleCI in this same PR 👍

### DIFF
--- a/features/step_definitions/web_steps.rb
+++ b/features/step_definitions/web_steps.rb
@@ -31,7 +31,7 @@ end
 
 When /^(?:|I )follow( invisible)? "([^"]*)"(?: within "([^"]*)")?$/ do |invisible, link, selector|
   with_scope(selector) do
-    click_link(link, visible: !invisible)
+    click_link(link, exact: true, visible: !invisible)
   end
 end
 
--- a/features/old/accounts/accounts.feature
+++ b/features/old/accounts/accounts.feature
@@ -40,14 +40,14 @@ Feature: Account management
     And user "bob" activates himself
     And current domain is the admin domain of provider "foo.example.com"
     When I log in as provider "bob"
-    And I follow "Account"
+    And I go to the provider account page
     # FIXME: as the Edit button now resides elsewhere, this does not assert anything
     Then I should not see link "Edit" within "#account_details"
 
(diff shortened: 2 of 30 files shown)


hallelujah:
Are there AJAX call in this feature? I'd prefer not waiting for the requests for every steps where "I go to"

features/step_definitions/web_steps.rb
```diff
@@ -21,6 +21,7 @@ def with_scope(locator)
 
 When /^(?:|I )go to (.+)$/ do |page_name|
   visit path_to(page_name)
+  wait_for_requests
```
hallelujah: Are there AJAX call in this feature? I'd prefer not waiting for the requests for every steps where "I go to"

Martouta:
@hallelujah Yes, that's why I said that I am not sure if it is a good idea, but I didn't find anything better to do... what do you propose? 🤔 
The test itself contains `@javascript` now but removing this tag happens exactly the same.
image

features/step_definitions/web_steps.rb
```diff
@@ -31,7 +31,7 @@ def with_scope(locator)
 
 When /^(?:|I )follow( invisible)? "([^"]*)"(?: within "([^"]*)")?$/ do |invisible, link, selector|
   with_scope(selector) do
-    click_link(link, visible: !invisible)
+    click_link(link, exact: true, visible: !invisible)
```
hallelujah: Oh yes this will make the page load a bit to find the right element before clicking
On slow ajax request the click_link is fatal, it catches "ApiDocs" instead of "Docs" because it is a regex match ...

hallelujah:
Approved! because it is solving the issue, in fact most of those cukes need rework but not the purpose of this PR

PR #1206, opened 2019-09-18
Fix Random Bug in Test: Order Services by ID before fetching the 2nd [Test Export to CSV]

Martouta:
The order in postgres changes depending on the execution, so if we fetch the `2nd` and the `2nd` changes for each execution, then sometimes it is going to raise an error because it does not fetch the one we want. So this PR just orders by ID before fetching, so it will always fetch the same one 😄 

Fixes [this random bug](https://circleci.com/gh/3scale/porta/101044?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link).

#### Verification steps
Run several times locally `DATABASE_URL="postgresql://postgres:@localhost:5433/systempdb" bundle exec m test/unit/csv/applications_exporter_test.rb:54`

### DIFF
--- a/test/unit/csv/applications_exporter_test.rb
+++ b/test/unit/csv/applications_exporter_test.rb
@@ -52,7 +52,7 @@ class Csv::ApplicationsExporterTest < ActiveSupport::TestCase
 
   test 'to_csv' do
     Timecop.freeze(Time.utc(2011,1,1)) do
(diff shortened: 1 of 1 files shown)


hallelujah:
Yes postgres does not order automatically by ID like MySQL

PR #1208, opened 2019-09-18
replace_path property in the routing policies injected for backend APIs

guicassolato:
Closes [THREESCALE-3489](https://issues.jboss.org/browse/THREESCALE-3489)

### DIFF
--- a/app/lib/backend_api_logic/routing_policy.rb
+++ b/app/lib/backend_api_logic/routing_policy.rb
@@ -54,7 +54,7 @@ module BackendApiLogic
                 value: path_to_regex
               ]
             }
-          }
+          }.merge(replace_path)
         end
 
         def path_to_regex
@@ -63,6 +63,13 @@ module BackendApiLogic
           else
             "/#{path}/.*|/#{path}/?"
           end
+        end
+
+        def replace_path
+          return {} if path.blank?
+
+          private_base_path = StringUtils::StripSlash.strip_slash(URI.parse(private_endpoint).path)
+          { replace_path: "{{original_request.path | replace: '/#{path}', '/#{private_base_path}'}}" }
         end
       end
       private_constant :Rule
--- a/test/unit/backend_api_logic/routing_policy_test.rb
+++ b/test/unit/backend_api_logic/routing_policy_test.rb
@@ -1,0 +1,65 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+module BackendApiLogic
+  class RoutingPolicyTest < ActiveSupport::TestCase
+    setup do
+      @service = FactoryBot.create(:simple_service)
+      other_backend_api = FactoryBot.create(:backend_api, system_name: 'foo')
+      @backend_apis = [service.first_backend_api, other_backend_api]
+      @service.backend_api_configs.create(backend_api: other_backend_api, path: 'foo')
+    end
+
+    attr_reader :service, :backend_apis
+    delegate :proxy, to: :service
+
+    test '#with_subpaths?' do
+      assert proxy.with_subpaths?
+      other_proxy = FactoryBot.create(:proxy)
+      refute other_proxy.with_subpaths?
+    end
+
+    test '#policy_chain' do
+      injected_rules = [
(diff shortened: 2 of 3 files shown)


app/lib/backend_api_logic/routing_policy.rb
```diff
@@ -54,7 +54,7 @@ def as_json
                 value: path_to_regex
               ]
             }
-          }
+          }.merge(replace_path || {})
```
hallelujah:
```suggestion
          }.merge(replace_path)
```

app/lib/backend_api_logic/routing_policy.rb
```diff
@@ -64,6 +64,13 @@ def path_to_regex
             "/#{path}/.*|/#{path}/?"
           end
         end
+
+        def replace_path
+          return if path.to_s.empty?
```
hallelujah:
```suggestion
          return {} if path.to_s.empty?
```

app/lib/backend_api_logic/routing_policy.rb
```diff
@@ -64,6 +64,13 @@ def path_to_regex
             "/#{path}/.*|/#{path}/?"
           end
         end
+
+        def replace_path
+          return {} if path.to_s.empty?
```
hallelujah: Is it a case where we need to cast it with `to_s`?

```suggestion
          return {} if path.blank?
```

test/unit/backend_api_logic/routing_policy_test.rb
```diff
@@ -0,0 +1,64 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+module BackendApiLogic
+  class RoutingPolicyTest < ActiveSupport::TestCase
+    setup do
+      @service = FactoryBot.create(:service)
```
Martouta:
```suggestion
      @service = FactoryBot.create(:simple_service)
```

The only difference between those 2 factories is that the account it creates is simple or with a bunch of stuff, and in this test you don't seem to want to use the account at all so it seems fair enough to make it as efficient as possible then :smile:

test/unit/backend_api_logic/routing_policy_test.rb
```diff
@@ -0,0 +1,64 @@
...
+
+    test '#with_subpaths?' do
+      assert proxy.with_subpaths?
+      other_proxy = FactoryBot.create(:proxy)
+      refute other_proxy.with_subpaths?
+    end
+
+    test '#policy_chain' do
+      injected_rules = [
+        { url: service.backend_apis.last.private_endpoint,  condition: { operations: [match: :path, op: :matches, value: '/foo/.*|/foo/?'] }, replace_path: '{{original_url.path | replace: "/foo", "/"}}' },
```
Martouta: Check this PR: https://github.com/3scale/porta/pull/1206
In this case, for `service.backend_apis.last` and the next line `service.backend_apis.first`, the order will be random for Postgres and sometimes we will get random CircleCI failures for that :disappointed: So you can fix it by something like:
```ruby
injected_rules = [
  { url: service.backend_apis.order(:id).last.private_endpoint,  condition: { operations: [match: :path, op: :matches, value: '/foo/.*|/foo/?'] }, replace_path: '{{original_url.path | replace: "/foo", "/"}}' },
  { url: service.backend_apis.order(:id).first.private_endpoint, condition: { operations: [match: :path, op: :matches, value: '/.*'] } }
]
```

guicassolato: Taken care by @hallelujah already: https://github.com/3scale/porta/blob/83a66a93c9bf04a3ae383327e111c324bd1b8096/app/models/backend_api_config.rb#L6

app/lib/backend_api_logic/routing_policy.rb
```diff
@@ -64,6 +64,13 @@ def path_to_regex
             "/#{path}/.*|/#{path}/?"
           end
         end
+
+        def replace_path
+          return {} if path.blank?
+
+          private_base_path = URI.parse(private_endpoint).path.to_s.sub(/^\//, '').sub(/\/$/, '')
```
hallelujah: You might want to use https://github.com/3scale/porta/blob/83a66a93c9bf04a3ae383327e111c324bd1b8096/app/lib/string_utils/strip_slash.rb#L8

guicassolato: Hmmm. It should not be `gsub`.

PR #1168, opened 2019-09-06
Do not create backend_api automatically when create a Service

duduribeiro:
Closes [THREESCALE-3368](https://issues.jboss.org/browse/THREESCALE-3368
)

This PR removes the auto creation of a backend api when a
product is created. We don't need to have a backend api every time
that a product is created.

The idea is that product and backend are two different things that
can be independently created and linked together at will.
All services are/will be products and there will then be no need
for a checkbox.

When using RU APIAP, the private endpoint is not created.
It creates a Service with default api_backend when RU APIAP is disabled

PR #1211, opened 2019-09-19
Do not create automatically create backend when creating a product

hallelujah:
Do not create backend_api automatically when create a Service

It is extracted from https://github.com/3scale/porta/pull/1168

- Keeping only changes in the backend, not in the UI
- Normally we do not have to touch the UI

### DIFF
--- a/app/controllers/admin/api/services_controller.rb
+++ b/app/controllers/admin/api/services_controller.rb
@@ -46,7 +46,9 @@ class Admin::Api::ServicesController < Admin::Api::ServiceBaseController
   ##~ op.parameters.add :name => " ", :dataType => "custom", :paramType => "query", :allowMultiple => true, :description => "Extra parameters"
   #
   def create
-    service = current_account.create_service(service_params)
+    service = current_account.services.build
+    create_service = ServiceCreator.new(service: service)
+    create_service.call!(service_params)
     service.reload if service.persisted? # It has been touched
     respond_with(service)
   end
--- a/app/controllers/admin/api/services_controller.rb
+++ b/app/controllers/admin/api/services_controller.rb
@@ -48,7 +48,7 @@ class Admin::Api::ServicesController < Admin::Api::ServiceBaseController
   def create
     service = current_account.services.build
     create_service = ServiceCreator.new(service: service)
-    create_service.call!(service_params)
+    create_service.call(service_params)
     service.reload if service.persisted? # It has been touched
     respond_with(service)
   end
--- a/app/controllers/api/services_controller.rb
+++ b/app/controllers/api/services_controller.rb
@@ -37,10 +37,9 @@ class Api::ServicesController < Api::BaseController
 
   def create
     @service = collection.new # this is done in 2 steps so that the account_id is in place as preffix_key relies on it
-    @service.attributes = service_params
-    @service.backend_api_configs = build_backend_api_configs
+    creator = ServiceCreator.new(service: @service)
 
-    if can_create? && @service.save
+    if can_create? && creator.call(service_params)
       flash[:notice] =  'Service created.'
       onboarding.bubble_update('api')
       redirect_to admin_service_path(@service)
@@ -70,15 +69,14 @@ class Api::ServicesController < Api::BaseController
 
   private
 
-  def build_backend_api_configs
-    return [] if !provider_can_use?(:api_as_product) || params[:service][:backend_api].blank?
-    backend_api = current_account.backend_apis.find(params[:service][:backend_api])
-    [BackendApiConfig.new(backend_api: backend_api)]
-  end
-
   def service_params
-    allowed_params = %i[system_name name description]
-    params.require(:service).permit(allowed_params)
+    permitted_params = [:name, :system_name, :description, :support_email, :deployment_option, :backend_version,
+                        :intentions_required, :buyers_manage_apps, :referrer_filters_required,
+                        :buyer_can_select_plan, :buyer_plan_change_permission, :buyers_manage_keys,
+                        :buyer_key_regenerate_enabled, :mandatory_app_key, :custom_keys_enabled, :state_event,
+                        :txt_support, :terms,
+                        {notification_settings: [web_provider: [], email_provider: [], web_buyer: [], email_buyer: []]}]
+    params.require(:service).permit(permitted_params)
   end
 
   protected
--- a/app/lib/backend_api_logic/proxy_extension.rb
+++ b/app/lib/backend_api_logic/proxy_extension.rb
@@ -20,11 +20,7 @@ module BackendApiLogic
     protected
 
     def validate_backend_api?
-      if account.provider_can_use?(:api_as_product)
-        backend_api.private_endpoint_changed?
-      else
-        true
-      end
+      !account.provider_can_use?(:api_as_product) || backend_api.private_endpoint_changed?
     end
 
     def save_backend_api
--- a/app/lib/backend_api_logic/proxy_extension.rb
+++ b/app/lib/backend_api_logic/proxy_extension.rb
@@ -5,24 +5,30 @@ module BackendApiLogic
     extend ActiveSupport::Concern
 
     included do
-      delegate :backend_api, to: :service, allow_nil: true
-      delegate :private_endpoint, :private_endpoint=, to: :backend_api, allow_nil: true, prefix: true
+      delegate :api_backend, :api_backend=, :backend_api, :backend_api_proxy, to: :service
+      delegate :private_endpoint, :private_endpoint=, to: :backend_api, prefix: true
       delegate :default_api_backend, to: 'BackendApi'
 
-      alias_method :api_backend, :backend_api_private_endpoint
-      alias_method :api_backend=, :backend_api_private_endpoint=
-
-      validates :backend_api, nested_association: {report: {private_endpoint: :api_backend}}, associated: true
       before_save :save_backend_api
 
       has_many :backend_api_configs, through: :service
       accepts_nested_attributes_for :backend_api_configs
+
+      validates :backend_api, nested_association: {report: {private_endpoint: :api_backend}}, associated: true, if: :validate_backend_api?
     end
 
     protected
 
+    def validate_backend_api?
+      if account.provider_can_use?(:api_as_product)
+        backend_api.private_endpoint_changed?
+      else
+        true
+      end
+    end
+
     def save_backend_api
-      backend_api.save
+      backend_api_private_endpoint && backend_api_proxy.update!(private_endpoint: backend_api_private_endpoint)
     end
   end
 end
--- a/app/lib/backend_api_logic/service_extension.rb
+++ b/app/lib/backend_api_logic/service_extension.rb
@@ -8,7 +8,8 @@ module BackendApiLogic
       has_many :backend_api_configs, inverse_of: :service, dependent: :destroy
       has_many :backend_apis, through: :backend_api_configs
 
-      alias_method :backend_api, :first_backend_api
+      delegate :backend_api, to: :backend_api_proxy
+      delegate :api_backend, :api_backend=, to: :backend_api
 
       has_many :backend_api_metrics, through: :backend_api_configs
 
@@ -17,6 +18,9 @@ module BackendApiLogic
       end, class_name: 'Metric'
     end
 
+    # TODO: This is used by db/migrate/20190716110520_create_the_backend_apis_of_services.rb
+    # We should Remove this after 2.7 is released
+    # https://issues.jboss.org/browse/THREESCALE-3517
     def first_backend_api
       @backend_api ||= find_or_create_first_backend_api!
     end
@@ -26,8 +30,6 @@ module BackendApiLogic
       config = backend_api_configs.first || create_first_backend_api_config!
       config.backend_api
     end
-
-    private
 
     def create_first_backend_api_config!
       backend_api = account.backend_apis.build(
@@ -39,6 +41,45 @@ module BackendApiLogic
       constructor = new_record? ? :build : :create!
       attrs = {backend_api: backend_api, path: ''}
       backend_api_configs.public_send(constructor, attrs)
+    end
+
+    class BackendApiProxy
+      attr_reader :service
+      delegate :account, :backend_apis, :backend_api_configs, to: :service
+      delegate :name, :system_name, to: :service, prefix: true
+
+      def initialize(service)
+        @service = service
+      end
+
+      def backend_api_config
+        @backend_api_config ||= backend_api_configs.first ||
+                                backend_api_configs.build(path: '', backend_api: backend_api)
+      end
+
+      def backend_api
+        @backend_api ||= backend_api_configs.first&.backend_api || account.backend_apis.build(system_name: service_system_name, name: "#{service_name} Backend API", description: "Backend API of #{service_name}")
+      end
+
+      def update!(attrs = {})
+        BackendApi.transaction do
+          backend_api_config.path = attrs[:path] if attrs.key?(:path)
+          backend_api_config.backend_api = backend_api
+          backend_api.private_endpoint = attrs[:private_endpoint] if attrs.key?(:private_endpoint)
+          backend_api.save!
+          backend_api_config.save!
+        end
+      end
+
+      def update(attrs = {})
+        update!(attrs)
+      rescue ActiveRecord::RecordInvalid
+        false
+      end
+    end
+
+    def backend_api_proxy
+      @backend_api_proxy ||= BackendApiProxy.new(self)
     end
   end
 end
--- a/app/models/account/provider_methods.rb
+++ b/app/models/account/provider_methods.rb
@@ -334,7 +334,7 @@ module Account::ProviderMethods
       plan.create_contract_with!(self)
     end
 
-    services.create! name: "API"
+    ServiceCreator.new(service: services.build(name: 'API')).call
   end
 
   def create_default_fields_definitions
--- a/app/models/backend_api.rb
+++ b/app/models/backend_api.rb
@@ -59,6 +59,7 @@ class BackendApi < ApplicationRecord
   private
 
   def set_private_endpoint
+    return if account.provider_can_use?(:api_as_product)
     self.private_endpoint ||= default_api_backend
   end
 
--- a/app/models/proxy.rb
+++ b/app/models/proxy.rb
@@ -102,8 +102,8 @@ class Proxy < ApplicationRecord
   alias_attribute :production_endpoint, :endpoint
   alias_attribute :staging_endpoint, :sandbox_endpoint
 
-  delegate :account, to: :service, allow_nil: true
-  delegate :provider_can_use?, to: :account, allow_nil: true
+  delegate :account, to: :service
+  delegate :provider_can_use?, to: :account
   delegate :backend_apis, :backend_api_configs, to: :service
 
   # This smells of :reek:NilCheck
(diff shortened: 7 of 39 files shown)


app/lib/backend_api_logic/service_extension.rb
```diff
@@ -9,7 +9,7 @@ module ServiceExtension
       has_many :backend_apis, through: :backend_api_configs
 
       delegate :backend_api, to: :backend_api_proxy
-      delegate :api_backend, to: :backend_api
+      delegate :api_backend, to: :backend_api, allow_nil: true
```
hallelujah: Why can it be nil? It must not, that is the whole point of the backend_api_proxy

app/lib/backend_api_logic/service_extension.rb
```diff
@@ -8,7 +8,9 @@ module ServiceExtension
       has_many :backend_api_configs, inverse_of: :service, dependent: :destroy
       has_many :backend_apis, through: :backend_api_configs
 
-      alias_method :backend_api, :first_backend_api
+      delegate :backend_api, to: :backend_api_proxy
+      delegate :api_backend, to: :backend_api, allow_nil: true
```
hallelujah: I do not get it, it can't be nil that is the whole point of the backend_api_proxy

app/lib/backend_api_logic/service_extension.rb
```diff
@@ -17,28 +19,37 @@ module ServiceExtension
...
+        @service = service
+      end
+
+      def backend_api_config
+        @backend_api_config ||= backend_api.backend_api_configs.first ||
+                                backend_api_configs.build(path: '', backend_api: backend_api)
+      end
+
+      def backend_api
+        return unless account
```
hallelujah: Please avoid returning `nil` object. This is just breaking NullObject pattern
why is account nil?

app/services/api_classification_service.rb
```diff
@@ -11,6 +11,7 @@ def initialize(test_hosts: TEST_API_HOSTS)
   # @return [Symbol] either `:real` or `:test`
 
   def test(url)
+    return Category.new(uri: nil, test_api: false) unless url
```
hallelujah: Is there a test for this change?

app/services/proxy_test_service.rb
```diff
@@ -129,7 +129,9 @@ def test_request(uri)
   end
 
   def ignore_failures?
-    api_classifier.test(@proxy.api_backend).test_api?
+    api_backend = @proxy.api_backend
+    return false unless api_backend
```
hallelujah: I think this does not belong to this PR. lets extract it

features/provider/integration/edit.feature
```diff
@@ -9,11 +9,13 @@ Feature: Edit Integration
      And all the rolling updates features are off
      And apicast registry is stubbed
 
-  Scenario: Edit a tested integration has a link to the analytics usage
-    Given the service has been successfully tested
-    When I go to the service integration page
-     And I follow "the analytics section"
-    Then I should be on the provider stats usage page
+  # Since we are skiping the test request, this is not rendered anymore.
```
hallelujah: yes but not in this PR

test/integration/api/services_controller_test.rb
```diff
@@ -82,42 +82,39 @@ def update_params
...
       Account.any_instance.stubs(:provider_can_use?).returns(true)
       @provider.settings.allow_multiple_services!
     end
 
-    test 'should create a new Backend API if none was selected' do
+    test 'should not create the default Backend API if API as Product is enabled' do
       Account.any_instance.stubs(:provider_can_use?).with(:api_as_product).returns(true).at_least_once
 
-      assert_change of: -> { BackendApi } do
+      assert_change of: -> { BackendApi.count }, by: 0 do
```
hallelujah: `assert_no_change`

app/services/api_classification_service.rb
duduribeiro: we can not have URL now, when private_endpoint is empty..

test/integration/api/services_controller_test.rb
hallelujah: I think you should test for when the RU is disabled as well!

duduribeiro: there is one already

```
    test 'should create the default Backend API if API as Product is disabled' do
      Account.any_instance.stubs(:provider_can_use?).with(:api_as_product).returns(false).at_least_once
```

app/services/proxy_test_service.rb
```diff
@@ -129,7 +129,7 @@ def test_request(uri)
   end
 
   def ignore_failures?
-    api_classifier.test(@proxy.api_backend).test_api??
```
Martouta: :joy:

duduribeiro: two `??` to be sure!!

app/controllers/api/services_controller.rb
```diff
@@ -69,17 +68,23 @@ def destroy
...
     allowed_params = %i[system_name name description]
     params.require(:service).permit(allowed_params)
   end
 
+  private
+
+  def service_params
+    permitted_params = [:name, :system_name, :description, :support_email, :deployment_option, :backend_version,
+      :intentions_required, :buyers_manage_apps, :referrer_filters_required,
+      :buyer_can_select_plan, :buyer_plan_change_permission, :buyers_manage_keys,
```
Martouta: You can automatically fix this bad identation by just doing `rubocop app/controllers/api/services_controller.rb --only Layout/AlignArray -a`

app/models/backend_api.rb
```diff
@@ -25,6 +25,7 @@ class BackendApi < ApplicationRecord
 
   alias_attribute :api_backend, :private_endpoint
 
+  # before_validation :set_private_endpoint, :set_port_private_endpoint
```
Martouta: Why do we need this line twice but one of them commented?

duduribeiro: mistake 😄 .. will remove it

app/models/proxy.rb
```diff
@@ -139,9 +139,9 @@ def find_policy_config_by(name:, version:)
 
   def policy_chain
     # TODO: We need to remove this rolling update as it should be available for everyone using APIcast V2
-    return unless provider_can_use?(:policies)
+    return [] unless provider_can_use?(:policies)
     raw_config = policies_config
-    return if raw_config.blank?
+    return [] if raw_config.blank?
```
Martouta: personally i don't mind this way :+1: but i would find it simpler as:
```ruby
def policy_chain
  # TODO: We need to remove this rolling update as it should be available for everyone using APIcast V2
  return [] unless provider_can_use?(:policies)
  (policies_config.presence || []).each_with_object([]) do |config, chain|
    chain << config.slice('name', 'version', 'configuration') if config['enabled']
  end
end
```

db/seeds.rb
```diff
@@ -69,7 +69,9 @@
   master_user.activate!
 
   # Creating the master service
-  master_service = Service.create!(account: master, name: master_service)
+  # master_service = Service.create!(account: master, name: master_service)
```
Martouta: :eyes:

app/controllers/api/services_controller.rb
```diff
@@ -69,17 +68,23 @@ def destroy
 
   private
 
-  def build_backend_api_configs
-    return [] if !provider_can_use?(:api_as_product) || params[:service][:backend_api].blank?
-    backend_api = current_account.backend_apis.find(params[:service][:backend_api])
-    [BackendApiConfig.new(backend_api: backend_api)]
-  end
-
   def service_params
```
Martouta: You need to remove this method or it is repeated! :scream:

app/lib/backend_api_logic/service_extension.rb
```diff
@@ -17,28 +19,36 @@ module ServiceExtension
...
+          backend_api_config.path = attrs[:path] if attrs.key?(:path)
+          backend_api_config.backend_api = backend_api
+          backend_api.private_endpoint = attrs[:private_endpoint] if attrs.key?(:private_endpoint)
+          backend_api.save!
+          backend_api_config.save!
+        end
+      end
     end
 
-    def find_or_create_first_backend_api!
```
hallelujah: This is used in a migration!

test/unit/signup/account_manager_test.rb
```diff
@@ -126,6 +126,20 @@ class ProviderAccountManagerTest < Signup::AccountManagerTest
       assert_equal 'API', account.first_service!.name
     end
 
+    test 'creating a provider will create default Backend API if api_as_product is disabled' do
```
Martouta: really? :scream: shouldn't it be when it is enabled, it creates it, and when it is disabled, it does not?

test/unit/backend_api_logic/service_extension_test.rb
```diff
@@ -3,50 +3,14 @@
...
-  end
-
-  test 'first_backend_api default values' do
-    service = FactoryBot.build(:simple_service, system_name: nil)
-    assert(backend_api = service.first_backend_api)
-    assert_equal "#{service.name} Backend API", backend_api.name
-    assert_equal  "Backend API of #{service.name}", backend_api.description
-  end
-
-  test 'first_backend_api when the service its own system_name' do
```
hallelujah: I guess those tests can just be left with the RU disabled?

app/lib/backend_api_logic/proxy_extension.rb
```diff
@@ -5,24 +5,30 @@ module ProxyExtension
     extend ActiveSupport::Concern
 
     included do
-      delegate :backend_api, to: :service, allow_nil: true
-      delegate :private_endpoint, :private_endpoint=, to: :backend_api, allow_nil: true, prefix: true
-      delegate :default_api_backend, to: 'BackendApi'
+      delegate :api_backend,       to: :service
+      delegate :api_backend=,      to: :service
```
Martouta: I think these 2 lines can be done in 1 single delegate in either 1 line or 2 lines

app/lib/backend_api_logic/proxy_extension.rb
```diff
@@ -5,24 +5,30 @@ module ProxyExtension
...
       has_many :backend_api_configs, through: :service
       accepts_nested_attributes_for :backend_api_configs
+
+      validates :backend_api, nested_association: {report: {private_endpoint: :api_backend}}, associated: true, if: :validate_backend_api?
     end
 
     protected
 
+    def validate_backend_api?
+      backend_api.private_endpoint && backend_api.changed?
```
Martouta: Shhouldn't we check that what changed is the private_endpoint in particular?

app/lib/backend_api_logic/service_extension.rb
```diff
@@ -8,7 +8,9 @@ module ServiceExtension
       has_many :backend_api_configs, inverse_of: :service, dependent: :destroy
       has_many :backend_apis, through: :backend_api_configs
 
-      alias_method :backend_api, :first_backend_api
+      delegate :backend_api, to: :backend_api_proxy
+      delegate :api_backend, to: :backend_api
+      delegate :api_backend=, to: :backend_api
```
Martouta: These 2 lines can be:
```suggestion
      delegate :api_backend, :api_backend=, to: :backend_api
```

app/lib/backend_api_logic/service_extension.rb
```diff
@@ -17,6 +19,8 @@ module ServiceExtension
       end, class_name: 'Metric'
     end
 
+    # This is used by db/migrate/20190716110520_create_the_backend_apis_of_services.rb
+    # We should Remove this after 2.7 is released
```
Martouta: Good, but put a `TODO``in the comment, and better if u create a jira issue, because this looks like the kind of thing that stays there for ages...

test/unit/signup/account_manager_test.rb
hallelujah: No that is correct
When it is on API as Product, we do not create a default backend API per service

Martouta: @hallelujah And when API as Product is disabled, we create a backend API for the service... what for? if api as product is disabled, we don't need a backend api at all, no?

hallelujah: We need it because we moved the private endpoint to BackendApi
We create a default one with echo-api

app/lib/backend_api_logic/proxy_extension.rb
hallelujah: It was like that before but I do not know why @duduribeiro changed it ...

app/models/account/provider_methods.rb
```diff
@@ -334,7 +334,8 @@ def create_first_service
       plan.create_contract_with!(self)
     end
 
-    services.create! name: "API"
+    creator = ServiceCreator.new(service: services.build(name: 'API'))
+    creator.call!
```
Martouta: If it is considered simple enough in one line in [ProviderSignup](https://github.com/3scale/porta/pull/1211/files#diff-ba591bb65af24eb68a38faa20a776e17R67), then probably also here :thinking:

app/lib/backend_api_logic/proxy_extension.rb
hallelujah: In fact, why should proxy know that it is private_endpoint that has changed?

I'd even say it should be something like:

```suggestion
      backend_api.validate_for_save?
```

```ruby
    def validate_for_save?
      if account.provider_can_use?(:api_as_product)
        # some condition
      else
        # some condition
     end    
    end
```

Will take a look tomorrow

Martouta: good point :+1: :muscle:

Martouta:
Please review feom master. The currently failing cucumber is fixed in master

app/lib/backend_api_logic/service_extension.rb
```diff
@@ -40,5 +42,37 @@ def create_first_backend_api_config!
...
+    class BackendApiProxy
+      attr_reader :service
+      delegate :account, :backend_apis, :backend_api_configs, to: :service
+
+      def initialize(service)
+        @service = service
+      end
+
+      def backend_api_config
+        @backend_api_config ||= backend_api.backend_api_configs.first ||
```
guicassolato: This should do it, no?

```suggestion
        @backend_api_config ||= backend_api_configs.first ||
```

If `Service` doesn't have any `BackendApiConfig`, then `#backend_api` will do `account.backend_apis.build`. Yet `backend_api.backend_api_configs.first` is `nil`, so the fallback invokes `#backend_api` again and builds a first `BackendApiConfig` using the `BackendApi` just built.

If `Service` already has a `BackendApiConfig`, then `#backend_api` will return the `BackendApi` of the first `BackendApiConfig` of the `Service` (let's call it `service_a`). However, the **first** `BackendApiConfig` of this returned `BackendApi` may not be the one with `service_a`, but, for example, with another `Service` called `service_b`; ergo `service_a.backend_api_proxy.backend_api_config.service == service_b`. 🤔

hallelujah: Good catch! thank you!

app/lib/backend_api_logic/proxy_extension.rb
```diff
@@ -5,24 +5,30 @@ module ProxyExtension
...
       has_many :backend_api_configs, through: :service
       accepts_nested_attributes_for :backend_api_configs
+
+      validates :backend_api, nested_association: {report: {private_endpoint: :api_backend}}, associated: true, if: :validate_backend_api?
     end
 
     protected
 
+    def validate_backend_api?
+      if account.provider_can_use?(:api_as_product)
```
Martouta: Same as:
```ruby
def validate_backend_api?
  !account.provider_can_use?(:api_as_product) || backend_api.private_endpoint_changed?
end
```

app/representers/proxy_representer.rb
```diff
@@ -4,7 +4,7 @@ class ProxyRepresenter < ThreeScale::Representer
   property :service_id
   property :endpoint
   property :deployed_at
-  property :api_backend
+  property :api_backend, render_nil: true
```
Martouta: I don't understand why now we render it even when it is `nil`. Although I don't find it a problem either. But if you consider it somehow important, then it should have a test (and I couldn't find it :thinking: )

hallelujah: Well because before the api_backend was never nil, it was always set.
But with Roar, if a value is nil it is not rendered by default which makes the acceptance spec fails

app/services/service_creator.rb
```diff
@@ -0,0 +1,54 @@
...
+  def save_default_backend_api(params)
+    if provider_can_use?(:api_as_product)
+      return true if %i[path private_endpoint].none? { |key| params.key?(key) }
+      backend_api_proxy.update!(params)
+    else
+      backend_api_proxy.update!(private_endpoint: BackendApi.default_api_backend)
+    end
+  end
+
+  def save_assigned_backend_api(attrs)
```
guicassolato:
```suggestion
  def save_assigned_backend_api!(attrs)
```

...as it ends with `config.save!`.


Despite Code Climate complaining, I don't really about having a quiet version without the bang.

app/services/service_creator.rb
```diff
@@ -0,0 +1,54 @@
...
+  def save!(params)
+    @service.save!
+    if @backend_api
+      save_assigned_backend_api(params)
+    else
+      save_default_backend_api(params)
+    end
+  end
+
+  def save_default_backend_api(params)
```
guicassolato:
```suggestion
  def save_default_backend_api!(params)
```

...as it ends with `update!`.

Despite Code Climate complaining, I don't really about having a quiet version without the bang.

guicassolato: Hmm. I see now they are `private`. Maybe not needed then.

app/controllers/admin/api/services_controller.rb
```diff
@@ -48,7 +48,7 @@ def index
   def create
     service = current_account.services.build
     create_service = ServiceCreator.new(service: service)
-    create_service.call!(service_params)
+    create_service.call(service_params)
     service.reload if service.persisted? # It has been touched
```
Martouta: I think that this line is not needed anymore... It was there for the way that it was creating the service back then. I remember that there was a test failing if I originally removed that line back then, so u should try now and see what happens :smile:

hallelujah: Yes I remember this one :)

duduribeiro: @hallelujah I pushed the change. should I remove?

Martouta: Confirmed: we still need the line :joy:

test/integration/provider/admin/backend_apis_controller_test.rb
```diff
@@ -55,18 +55,19 @@ class Provider::Admin::BackendApisControllerTest < ActionDispatch::IntegrationTe
...
 
     delete provider_admin_backend_api_path(backend_api)
     assert BackendApi.exists? backend_api.id
     assert_equal 'Backend API could not be deleted', flash[:error]
   end
-  
+
   test 'delete a backend api without any products' do
-    backend_api = @provider.backend_apis[1]
+    backend_api = @provider.backend_apis.second
```
guicassolato: Depending on the order from the db here.

test/unit/backend_api_logic/routing_policy_test.rb
```diff
@@ -6,6 +6,7 @@ module BackendApiLogic
   class RoutingPolicyTest < ActiveSupport::TestCase
     setup do
       @service = FactoryBot.create(:simple_service)
+      first_backend_api = FactoryBot.create(:backend_api_config, path: '', service: @service).backend_api
       other_backend_api = FactoryBot.create(:backend_api, system_name: 'foo')
       @backend_apis = [service.first_backend_api, other_backend_api]
```
guicassolato:
```suggestion
      @backend_apis = [first_backend_api, other_backend_api]
```

PR #1202, opened 2019-09-17
Integration Settings as Application Usage Rules

didierofrivia:
**What this PR does / why we need it**:

The current Integration > Settings needs to be migrated to Applications > Usage Rules when rolling update `api_as_product` is enabled

**Which issue(s) this PR fixes** 

fixes https://issues.jboss.org/browse/THREESCALE-3471

**Verification steps** 

* With Rolling update on:
Screen Shot 2019-09-19 at 09 13 25

* With Rolling update off:
Screen Shot 2019-09-19 at 09 14 50

**Special notes for your reviewer**:
Mind the rolling update feature needs to be either `false` or an array with the provider_id (s)

PR #1215, opened 2019-09-19
[Usage Rules] Separated view and route

didierofrivia:
This aims to enhance the previous https://github.com/3scale/porta/pull/1202, making it a bit easier to tackle https://issues.jboss.org/browse/THREESCALE-3451

Improves  https://issues.jboss.org/browse/THREESCALE-3471

This is an enhancement of the previous https://github.com/3scale/porta/pull/1202

* The views are duplicated, but it's needed in order to get independence for method
settings
* In the future, we just adjust the cukes to Usage Rules instead of old
Settings, and we can safely remove the old settings
* Clean, no conditionals

### DIFF
--- a/app/controllers/api/services_controller.rb
+++ b/app/controllers/api/services_controller.rb
@@ -12,7 +12,7 @@ class Api::ServicesController < Api::BaseController
   load_and_authorize_resource :service, through: :current_user,
     through_association: :accessible_services, except: [:create]
 
-  with_options only: %i[edit update settings] do |actions|
+  with_options only: %i[edit update settings usage_rules] do |actions|
     actions.sublayout 'api/service'
   end
 
@@ -30,8 +30,13 @@ class Api::ServicesController < Api::BaseController
   end
 
   def settings
-    activate_menu_args = current_account.provider_can_use?(:api_as_product) ? %i[serviceadmin applications usage_rules] : %i[serviceadmin integration settings]
-    activate_menu activate_menu_args
+    activate_menu :serviceadmin, :integration, :settings
+    @alert_limits = Alert::ALERT_LEVELS
+  end
+
+  def usage_rules
+    raise ActiveRecord::RecordNotFound unless provider_can_use?(:api_as_product)
+    activate_menu :serviceadmin, :applications, :usage_rules
     @alert_limits = Alert::ALERT_LEVELS
   end
 
--- a/app/helpers/vertical_nav_helper.rb
+++ b/app/helpers/vertical_nav_helper.rb
@@ -222,8 +222,8 @@ module VerticalNavHelper
     items << {id: :listing,           title: 'Listing',           path: admin_service_applications_path(@service)}
     items << {id: :application_plans, title: 'Application Plans', path: admin_service_application_plans_path(@service)} if can?(:manage, :plans)
     if current_account.provider_can_use?(:api_as_product)
-      items << {                          title: 'Settings'}
-      items << {id: :usage_rules,       title: 'Usage Rules',       path: settings_admin_service_path(@service)}
+      items << {title: 'Settings'}
+      items << {id: :usage_rules, title: 'Usage Rules', path: usage_rules_admin_service_path(@service)}
     end
     items
   end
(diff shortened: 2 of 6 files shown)


app/views/api/services/usage_rules.html.slim
```diff
@@ -0,0 +1,22 @@
+- # TODO: When APIAP is default, adjust the old settings cukes to usage rules
+- content_for :sublayout_title, 'Usage Rules'
```
hallelujah: Hmm it would be safer to use it with `:flush` otherwise it will append to any previous content 
```suggestion
- content_for :sublayout_title, 'Usage Rules', flush: true
```

app/views/api/services/usage_rules.html.slim
```diff
@@ -0,0 +1,22 @@
...
+
+- if (can? :manage, :service_plans) && !current_account.settings.service_plans_ui_visible?
+	= form_tag polymorphic_path([:masterize, :admin, @service, ServicePlan]), :method => :post, :class => 'service autosubmit formtastic', :remote => true do
+		fieldset.inputs
+			legend Default Service Plan
+			ol
+				li.string.optional
```
hallelujah: Are those tabs?
The indentation is really weird, it should be 2 spaces

didierofrivia: That's true, I've copy/pasta from the other view... I'll take a look and see if the IDE did some nasty thing..

Martouta: hahahah exactly, i was really doubting if it was just me seeing it weird for using a different laptop and setup now :joy:

app/views/api/services/forms/_usage_rules.html.slim
```diff
@@ -0,0 +1,55 @@
...
+	= form.input :buyers_manage_apps
+
+	- if @service.backend_version.app_keys_allowed?
+		= form.input :buyers_manage_keys
+
+	= form.input :referrer_filters_required, as: :boolean
+
+	= form.input :custom_keys_enabled
+
+	- if @service.backend_version == "2"
```
Martouta: not sure if it should be `- if @service.backend_version.app_keys_allowed?` :thinking:

app/views/api/services/usage_rules.html.slim
didierofrivia: and yeah, it was the IDE settings. fixed

app/views/api/services/forms/_usage_rules.html.slim
```diff
@@ -0,0 +1,55 @@
...
+  = form.input :buyers_manage_apps
+
+  - if @service.backend_version.app_keys_allowed?
+    = form.input :buyers_manage_keys
+
+  = form.input :referrer_filters_required, as: :boolean
+
+  = form.input :custom_keys_enabled
+
+  - if @service.backend_version == "2"
```
Martouta: not sure if it should be `- if @service.backend_version.app_keys_allowed?` :thinking:

didierofrivia: I didn't want to change much, just move the file around :)

app/helpers/vertical_nav_helper.rb
```diff
@@ -223,7 +223,7 @@ def service_applications
     items << {id: :application_plans, title: 'Application Plans', path: admin_service_application_plans_path(@service)} if can?(:manage, :plans)
     if current_account.provider_can_use?(:api_as_product)
       items << {                          title: 'Settings'}
-      items << {id: :usage_rules,       title: 'Usage Rules',       path: settings_admin_service_path(@service)}
+      items << {id: :usage_rules,       title: 'Usage Rules',       path: usage_rules_admin_service_path(@service)}
```
duduribeiro: I think you can remove this spaces before `path`

app/controllers/api/services_controller.rb
```diff
@@ -30,8 +30,12 @@ def edit
   end
 
   def settings
-    activate_menu_args = current_account.provider_can_use?(:api_as_product) ? %i[serviceadmin applications usage_rules] : %i[serviceadmin integration settings]
-    activate_menu activate_menu_args
+    activate_menu :serviceadmin, :integration, :settings
+    @alert_limits = Alert::ALERT_LEVELS
+  end
+
+  def usage_rules
```
hallelujah: It is under rolling update right?

didierofrivia: The link is under rolling updates, here: https://github.com/3scale/porta/blob/THREESCALE-3471/enhance_for_3451/app/helpers/vertical_nav_helper.rb#L224-L227

PR #1222, opened 2019-09-21
THREESCALE-2542 Ability to configure JWT Claim with ClientID

hallelujah:
Closes THREESCALE-2542

### What it does

Ability to configure through API and UI the JWT Claim With ClientID
By default in the UI they are pre-filled to `plain` and `azp` if OIDC is chosen as authentication methid, so customers are not confused.

In the API they are not set as the default in APIcast no values means `plain` and `azp` anyway

### Notes

This  PR introduces a new storage for the Proxy model: GatewayConfiguration
Instead of using columns, configurations are stored into a JSON field through `ActiveRecord::Base::store` method

It will be more flexible in the future, no need of migration (proxies table is quite core in the project)

Also we could migrate probably the old columns to that one but no need for now.

### DIFF
--- a/app/lib/gateway_settings/proxy_extension.rb
+++ b/app/lib/gateway_settings/proxy_extension.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+module GatewaySettings
+  module ProxyExtension
+    extend ActiveSupport::Concern
+    included do
+      has_one :gateway_configuration, dependent: :delete, inverse_of: :proxy, autosave: true
+      # TODO: In the future, stop using columns and put all configuration to this class
+      delegate(*GatewayConfiguration.accessors, to: :gateway_configuration)
+    end
+
+    def gateway_configuration
+      super || build_gateway_configuration
+    end
+  end
+end
--- a/app/controllers/admin/api/services/proxies_controller.rb
+++ b/app/controllers/admin/api/services/proxies_controller.rb
@@ -6,7 +6,7 @@ class Admin::Api::Services::ProxiesController < Admin::Api::Services::BaseContro
   represents :json, entity: ::ProxyRepresenter::JSON
   represents :xml, entity: ::ProxyRepresenter::XML
 
-  wrap_parameters ::Proxy, include: Proxy.attribute_names + %w[api_backend]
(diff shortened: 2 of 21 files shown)


thomasmaas:
Don’t forget default values as described in the issue

app/lib/gateway_settings/proxy_extension.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module GatewaySettings
+  module ProxyExtension
+    extend ActiveSupport::Concern
+    included do
+      has_one :gateway_configuration, dependent: :delete, inverse_of: :proxy, autosave: true
```
Martouta: why `dependent: :delete` instead of `dependent: :destroy` ? :thinking:

hallelujah: Because there is nothing else we can do, it is just a storage, no other dependencies nor callbacks

Basically I want to extract all setttings (only data) from the Proxy model
Proxy should have mandatory data, the settings are just configurations

hallelujah:
About the defaults I am not sure if we need them. It was APIcast request but here are my thoughts:

- APIcast already sets it to `azp` if nothing is passed https://github.com/3scale/APIcast/pull/1034/files#diff-fe57f478c9d93c2245426baff7e5171bR19
- now it works without the JWT claim in APIcast
- there is no need of the JWT claim client_id, it is only for some customers, most won't use it
- there is no need to store something optional in the database.
- it might be confusing, some might think that it is mandatory (thus the default), I think they should be `nil`

Looking at https://github.com/3scale/APIcast/pull/1034/files#diff-fe57f478c9d93c2245426baff7e5171bR228-R237 it only works if both of them are defined

So this PR just makes sure that either both fields are defined or none.

hallelujah:
So defaults are added @thomasmaas but only in the view.
I would need you to polish the view, or it is good enough

thomasmaas:
- explain more and be precise with labels
- update structure and order for all 3 auth settings so they always have their specific fields first

Screenshot 2019-09-23 12 51 51
Screenshot 2019-09-23 12 51 12
Screenshot 2019-09-23 12 49 39

thomasmaas:
In my mind we should just save the new values to db when the user decides to submit the form, why keep forever faking it?

PR #1223, opened 2019-09-21
Fix Cuke Master Step ensuring that master exists or creates it

Martouta:
Fixes the random failure [in CircleCI](https://circleci.com/gh/3scale/porta/105017#tests/containers/18) with title `failed The 404 page` and in `features/percy/provider.feature:5:in Given the master account allows signups`

### DIFF
--- a/features/step_definitions/master_steps.rb
+++ b/features/step_definitions/master_steps.rb
@@ -1,4 +1,5 @@
 Given(/^the master account allows signups$/) do
+  master_account # Ensure that the master account exists, otherwise it creates it
   step 'provider "master" has multiple applications disabled'
   step 'provider "master" has default service and account plan'
   step 'a default application plan "Base" of provider "master"'


hallelujah:
Approved but I wonder
If you rin this cuke alone with an empty DB it should always fail no?
How can this be random? It should be another cuke or step that deletes the master account aomehow or makes it invalid

Martouta:
Executed in isolation works locally in the current master. I think because of this:
https://github.com/3scale/porta/blob/434182f7f4889f04fe4efbdd26e052be22034f5b/features/support/initial_data.rb#L13
So in this CircleCI failure, maybe another feature removed master :woman_shrugging:

hallelujah:
We should look at which feature removed master or makes it invalid

Martouta:
Makes sense :+1: I will make a Jira issue to check this after 2.7 ER1

PR #1224, opened 2019-09-22
Destroy Backend APIs for destroyed objects when RU api_as_product disabled

Martouta:
Closes [THREESCALE-3317](https://issues.jboss.org/browse/THREESCALE-3317)

### DIFF
--- a/app/models/backend_api.rb
+++ b/app/models/backend_api.rb
@@ -33,6 +33,7 @@ class BackendApi < ApplicationRecord
   has_system_name(uniqueness_scope: [:account_id])
 
   scope :orphans, -> { joining { services.outer }.where { BabySqueel[:backend_api_configs].backend_api_id == nil } }
+  scope :of_product, ->(service) { joining { services }.where { BabySqueel[:backend_api_configs].service_id == service.id } }
 
   scope :oldest_first, -> { order(created_at: :asc) }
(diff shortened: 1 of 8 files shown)


app/workers/delete_service_hierarchy_worker.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class DeleteServiceHierarchyWorker < DeleteObjectHierarchyWorker
+  def perform(*)
+    purge_backend_apis unless service.account.provider_can_use?(:api_as_product)
```
hallelujah: Add a TODO here to remove this line when api_as_product is shipped
Otherwise we might just remove the conditional thus removing all shared BackendApis

app/models/backend_api.rb
```diff
@@ -33,6 +33,7 @@ class BackendApi < ApplicationRecord
   has_system_name(uniqueness_scope: [:account_id])
 
   scope :orphans, -> { joining { services.outer }.where { BabySqueel[:backend_api_configs].backend_api_id == nil } }
+  scope :of_product, ->(service) { joining { services }.where { BabySqueel[:backend_api_configs].service_id == service.id } }
```
hallelujah: It would be nice to test this scope :)

Martouta: I will try the `service.backend_apis` :+1:

PR #1214, opened 2019-09-19
[API Endpoints] Backend API Configs

Martouta:
Closes [THREESCALE-3207 - [API Endpoints] Backend API Configs CRUD](https://issues.jboss.org/browse/THREESCALE-3207
)
#### Create
image

image

- Request: `curl -v  -X POST "http://provider-admin.example.com.lvh.me:3000/admin/api/services/2/backend_apis.json" -d 'access_token=04bddce348f2445c48640ef0918e04b05988c662ae945f0c3aa91ede88b1b7db&backend_api_id=2&path=%2Ffoo'`

- Response:
```json
{
 "path": "foo",
 "links": [
  {
   "rel": "service",
   "href": "http://provider-admin.example.com/admin/api/services/2"
  },
  {
   "rel": "backend_api",
   "href": "http://provider-admin.example.com/admin/api/backend_apis/2"
  }
 ]
}
```

#### Index
image

image

- Request: `curl -v  -X GET "http://provider-admin.example.com.lvh.me:3000/admin/api/services/2/backend_apis.json?access_token=04bddce348f2445c48640ef0918e04b05988c662ae945f0c3aa91ede88b1b7db"`

- Response:
```json
[
 {
  "path": "foo",
  "links": [
   {
    "rel": "service",
    "href": "http://provider-admin.example.com/admin/api/services/2"
   },
   {
    "rel": "backend_api",
    "href": "http://provider-admin.example.com/admin/api/backend_apis/2"
   }
  ]
 },
 {
  "path": "foo/bar",
  "links": [
   {
    "rel": "service",
    "href": "http://provider-admin.example.com/admin/api/services/2"
   },
   {
    "rel": "backend_api",
    "href": "http://provider-admin.example.com/admin/api/backend_apis/3"
   }
  ]
 }
]
```

#### Update
image

image

- Request: `curl -v  -X PUT "http://provider-admin.example.com.lvh.me:3000/admin/api/services/2/backend_apis/2.json" -d 'access_token=04bddce348f2445c48640ef0918e04b05988c662ae945f0c3aa91ede88b1b7db&path=%2Ffoo%2Fupdated'`

- Response:
```json
{
 "path": "foo/updated",
 "links": [
  {
   "rel": "service",
   "href": "http://provider-admin.example.com/admin/api/services/2"
  },
  {
   "rel": "backend_api",
   "href": "http://provider-admin.example.com/admin/api/backend_apis/2"
  }
 ]
}
```

#### Delete
image

image

[...]

PR #1225, opened 2019-09-22
[PR JUST TO UNBLOCK QE!][DO NOT REVIEW OR MERGE!] API to relate Services with BackendAPI

Martouta:
Just for QE to be able to relate BackendAPIs with Products/Services and undo it.
The real PR is in https://github.com/3scale/porta/pull/1214 but this one is just for QE and not meant to be ever merged.

### DIFF
--- a/app/controllers/admin/api/backend_apis/services_controller.rb
+++ b/app/controllers/admin/api/backend_apis/services_controller.rb
@@ -1,0 +1,54 @@
+# frozen_string_literal: true
+
+class Admin::Api::BackendApis::ServicesController < Admin::Api::BackendApis::BaseController
+  wrap_parameters Service, include: Service.attribute_names | %w[state_event]
+  representer Service
+
+  # TODO: ActiveDocs
+  # TODO: Refactor
+  # TODO: should it all be called product instead of service?
+  # TODO: paginate
+
+  def index
+    services = backend_api.services.accessible
+    respond_with(services)
+  end
+
+  def create
+    service = current_account.services.build
+    create_service = ServiceCreator.new(service: service, backend_api: backend_api)
+    create_service.call(service_params)
+    service.reload if service.persisted? # It has been touched # TODO: Add test for this
+    respond_with(service)
+  end
+
+  def show
+    respond_with(service)
+  end
+
+  def update
+    service.update(service_params)
+    respond_with(service)
+  end
+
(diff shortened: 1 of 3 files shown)


app/controllers/admin/api/backend_apis/services_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+  def index
+    services = backend_api.services.accessible
+    respond_with(services)
+  end
+
+  def create
+    service = current_account.services.build
+    create_service = ServiceCreator.new(service: service, backend_api: backend_api)
+    create_service.call(service_params)
+    service.reload if service.persisted? # It has been touched # TODO: Add test for this
```
hallelujah: I think this is not needed

app/controllers/admin/api/backend_apis/services_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+class Admin::Api::BackendApis::ServicesController < Admin::Api::BackendApis::BaseController
+  wrap_parameters Service, include: Service.attribute_names | %w[state_event]
+  representer Service
+
+  # TODO: ActiveDocs
+  # TODO: Refactor
+  # TODO: should it all be called product instead of service?
+  # TODO: paginate
+
+  def index
```
hallelujah: This could be useful in https://github.com/3scale/porta/blob/9b79b8b4177e0c9cd0fba99f6084bf9adabc4927/app/controllers/admin/api/backend_apis_controller.rb

hallelujah:
OK I got the gist of it 👍 
Some of them might be useful

app/controllers/admin/api/backend_apis/services_controller.rb
Martouta: I think the same but i am investigating so far: https://github.com/3scale/porta/pull/1214/commits/1cbef5e33fe93f0f32f3389b525a3e1db355025f
the difference is the `updated_at`, so i think the key is to change the way we create to save at the end only, but we'll see...

Martouta: Anyway, let's keep the conversation in the other PR plz, this one is only for QE to be able do work in the meantime

Martouta:
Mission accomplished :sunglasses:

PR #1226, opened 2019-09-23
Add Backend API config to product

guicassolato:
Closes [THREESCALE-3321](https://issues.jboss.org/browse/THREESCALE-3321)

ezgif-5-b3a5367b70b1

### DIFF
--- a/app/models/backend_api_config.rb
+++ b/app/models/backend_api_config.rb
@@ -11,6 +11,8 @@ class BackendApiConfig < ApplicationRecord
 
   has_many :backend_api_metrics, through: :backend_api, source: :metrics
 
+  validates :service_id, :backend_api_id, presence: true
+  validates :path, uniqueness: { scope: :service_id, case_sensitive: false }
   validates :path, length: { in: 0..255, allow_nil: false }, path: true
 
   scope :with_subpath, (lambda do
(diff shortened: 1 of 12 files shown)


config/abilities/provider_admin.rb
```diff
@@ -78,6 +78,7 @@
 
     if account.provider_can_use?(:api_as_product)
       can :manage, BackendApi
+      can :manage, BackendApiConfig
```
Martouta: Ok so I definitely have to reuse this then :+1: _taking notes..._ :smile:

app/models/backend_api_config.rb
```diff
@@ -11,6 +11,8 @@ class BackendApiConfig < ApplicationRecord
 
   has_many :backend_api_metrics, through: :backend_api, source: :metrics
 
+  validates :backend_api_id, presence: true
```
Martouta:
```suggestion
  validates :backend_api_id, :service_id, presence: true
```

test/unit/abilities/provider_admin_test.rb
```diff
@@ -82,6 +82,14 @@ def test_services
       assert_can ability, :show, service_3
     end
 
+    test 'backend apis' do
+      assert_can ability, :manage, BackendApi
+    end
+
+    test 'backend api configs' do
+      assert_can ability, :manage, BackendApiConfig
```
Martouta: would be nice to also have a test for when the RU is disabled :smile:

PR #1228, opened 2019-09-23
[apiap] Extract Policies to its own page

hallelujah:
Fixes THREESCALE-3450


app/controllers/api/policies_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  protected
+
+  def proxy_params
+    params.require(:proxy).permit(:policies_config)
+  end
+
+  def check_permission!
+    provider_can_use! :api_as_product
+    authorize! :edit, service
```
hallelujah: We have the service permission (which does not really work well sadly), Just thinking about the future

guicassolato:
I fixed https://github.com/3scale/porta/blob/8b63356fc1e8c105de40445a06b19f9fcc7c7f06/features/old/menu/api_menu.feature#L59-L67 in https://github.com/3scale/porta/pull/1230. You'll probably gonna need it here.

app/javascript/src/Policies/styles/policies.scss
```diff
@@ -15,12 +15,6 @@ $error-color: #c00;
   line-height: 1.5;
   padding-bottom: 2rem;
 
-  .PolicyChain, .PolicyRegistryList, .PolicyConfiguration {
```
guicassolato: I think the partial will continue to be rendered in the integration form for non-APIAP accounts. Won't this change affect the way those users see the Policies section in the page?

guicassolato: Even for APIAP-viewers some minor fixes need to be done:

Screen Shot 2019-09-25 at 1 17 30 PM

thomasmaas:
@hallelujah did you think about that policies page should not appear in nav when user has selected Istio?

PR #1231, opened 2019-09-23
[WIP] Validate that for a BackendApiConfig, its Service and BackendAPI belong to the same tenant

Martouta:
Without this validation, we can have big problems :stuck_out_tongue_closed_eyes:

### DIFF
--- a/app/models/backend_api_config.rb
+++ b/app/models/backend_api_config.rb
@@ -14,6 +14,7 @@ class BackendApiConfig < ApplicationRecord
   validates :service_id, :backend_api_id, presence: true
   validates :path, uniqueness: { scope: :service_id, case_sensitive: false }
   validates :path, length: { in: 0..255, allow_nil: false }, path: true
+  validate :validate_service_and_backend_api_belong_to_the_same_tenant
 
   scope :with_subpath, (lambda do
     common_query = where.not(path: '/')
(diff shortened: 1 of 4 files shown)


hallelujah:
We have the tenant_id triggers already

So it should not be an issue, and if it is then all our models need this check

Martouta:
> We have the tenant_id triggers already
> So it should not be an issue

In the trigger we are putting the same tenant_id as the backend_api:
https://github.com/3scale/porta/blob/d76fdf8e0bab92a31cfe3ae2021fa78b42380d49/lib/system/database/definitions/mysql.rb#L452-L456

That does not mean that the backend_api and the service belong to the same tenant

> then all our models need this check

Give me an example then of a model in this situation :stuck_out_tongue_closed_eyes: And if we need it, we do anyway :woman_shrugging: In this case the backend_api_config belongs both to a backend_api and to a service, and somehow in the controllers we are going to allow the id of one of them, and that means that if the customers are allowed to set an id of an existing object that does not belong to them, we are in trouble...
For the API in particular I am searching the object inside the scope of the current_account or it returns 404, and I dunno if @guicassolato is checking that in the controller of the UI (imo he should), but still I am not sure if it is safer to also have this check in the model :thinking:

Martouta:
I added [this comment](https://github.com/3scale/porta/pull/1230#discussion_r327504663) for this problem in the review of the UI.

hallelujah:
In the other case we should just load the BackendApi with `current_account.backend_appi.find(params[:backend_api_id])` and done

Martouta:
Right :+1: Conclusion:
image

PR #1235, opened 2019-09-24
Show exclamation mark for each product with out of date config

thomasmaas:
- [x] visual implementation dashboard (@thomasmaas)

 _hiding other icons to make this one stand out, other icons should be removed anyway as icons are no longer supported in vertical nav so doesn't make sense._

Screenshot 2019-09-24 14 29 16

- [x] visual implementation vertical nav (@damianpm) (not supported in PF4 but we will do it anyway)

Screenshot 2019-09-24 13 46 24

- [x] working method (@hallelujah)
- [ ] tests

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -367,6 +367,6 @@ module ApplicationHelper
   end
 
   def has_out_of_date_configuration?(service)
-    true
+    service.pending_affecting_changes?
   end
 end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -365,4 +365,8 @@ module ApplicationHelper
   def unlinked_paid_account?
     !current_account.field_value('red_hat_account_verified_by').presence && current_account.paid?
   end
+
+  def has_out_of_date_configuration?(service)
+    true
+  end
 end
--- a/app/javascript/src/Navigation/components/PF4NavProxy.jsx
+++ b/app/javascript/src/Navigation/components/PF4NavProxy.jsx
@@ -8,7 +8,12 @@ const Nav = ({id, children}) => (
   </nav>
 )
 
-const NavExpandable = ({id, title, isExpanded, isActive, children}) => {
+const OutdatedIcon = ({title}) => {
+  const isVisibleIcon = title === 'Integration'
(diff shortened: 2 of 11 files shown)


Martouta:
For this kind of tasks, I appreciate some screenshots in the description :stuck_out_tongue_winking_eye:

app/helpers/application_helper.rb
```diff
@@ -365,4 +365,8 @@ def call_to_action?
   def unlinked_paid_account?
     !current_account.field_value('red_hat_account_verified_by').presence && current_account.paid?
   end
+
+  def has_out_of_date_configuration?(service)
```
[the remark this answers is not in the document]
hallelujah: That works in a helper :)

hallelujah:
@thomasmaas yes do not worry no need to be in the RU apiap
It is not particular to APIAP (example changing authentication method)

app/views/shared/provider/navigation/_vertical_nav.html.slim
```diff
@@ -3,8 +3,9 @@
 - sections = vertical_nav_sections || []
 - active_section = active_submenu
 - active_item = active_sidebar
+- out_of_date_config = has_out_of_date_configuration?('foo')
```
damianpm: What argument `has_out_of_date_configuration?` expects?

thomasmaas: service

app/javascript/src/Navigation/components/PF4NavProxy.jsx
```diff
@@ -8,7 +8,12 @@ const Nav = ({id, children}) => (
   </nav>
 )
 
-const NavExpandable = ({id, title, isExpanded, isActive, children}) => {
+const OutdatedIcon = ({title}) => {
+  const isVisibleIcon = title === 'Integration'
+  return isVisibleIcon ? <i className='fa fa-exclamation-triangle' title='The Integration is out of date, promote!'></i> : null
```
guicassolato: I'm thinking about using "promote". It will be both staging and production configs that will be out of date, right? Yet, it appears that we currently only use the "promote" for when updating prod.

thomasmaas: yes, there will be a new button: https://issues.jboss.org/browse/THREESCALE-3539

PR #1229, opened 2019-09-23
Integration Settings part 1

didierofrivia:
Screenshot 2019-09-24 at 09 10 56

**What this PR does / why we need it**:

This is the first iteration of the unified integration settings

**Which issue(s) this PR fixes** 

Partially fixes https://issues.jboss.org/browse/THREESCALE-3451

**Verification steps** 

TODO

**Special notes for your reviewer**:

This includes:
* Deployment options
* Authentication just the auth methods selection, missing its settings

PR #1238, opened 2019-09-24
Integration Settings part 2

didierofrivia:
This needs to be merge after https://github.com/3scale/porta/pull/1229

This is the second iteration of the unified integration settings

Partially fixes https://issues.jboss.org/browse/THREESCALE-3451

**Verification steps** 

TODO

### DIFF
--- a/app/views/api/services/forms/_authentication_settings.html.slim
+++ b/app/views/api/services/forms/_authentication_settings.html.slim
@@ -1,41 +1,40 @@
 = f.inputs class: 'auth-settings-wrapper' do
   - # oidc
-  = f.inputs id: 'oidc', disabled: true, class: 'auth-settings hidden'
+  - is_oidc = @service.backend_version === 'oidc'
+  = f.inputs id: 'service_proxy_authentication_method_oidc_settings', disabled: !is_oidc, class: "auth-settings #{'hidden' unless is_oidc}"
     = render 'api/services/forms/oidc', f: f, proxy: @service.proxy
 
-  - # oauth
-  = f.inputs id: 'oauth', class: 'auth-settings hidden'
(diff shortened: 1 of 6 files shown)


app/views/api/services/forms/_authentication_settings.html.slim
```diff
@@ -0,0 +1,40 @@
+= f.inputs class: 'auth-settings-wrapper' do
+  - # oidc
```
Martouta: I assume that it is not your fault but all these 'ruby comments' should be slim comments, so for example instead of `- # oidc`, it should be `/ oidc` and then it is not rendered either.

app/views/api/services/forms/_authentication_settings.html.slim
```diff
@@ -0,0 +1,40 @@
...
+  - is_oidc = @service.backend_version === 'oidc'
+  = f.inputs id: 'service_proxy_authentication_method_oidc_settings', disabled: !is_oidc, class: "auth-settings #{'hidden' unless is_oidc}"
+    = render 'api/services/forms/oidc', f: f, proxy: @service.proxy
+
+  - # auth user key (1)
+  - is_auth_user = @service.backend_version === '1'
```
Martouta: It should be a method to be called from `@service` and definitely not with the same `is_...`.
But I guess this is also copy-pasted :smile:

app/views/api/services/forms/_authentication_settings.html.slim
```diff
@@ -0,0 +1,40 @@
...
+
+  = f.input :hostname_rewrite, label: 'Host Header'
+  = f.input :secret_token, label: 'Secret Token'
+
+  = f.input :credentials_location, as: :radio,
+          collection: Proxy.credentials_collection
+
+= f.inputs 'Gateway Response' do
+  = f.inputs 'Authentication Failed Error' do
+    = f.input :error_status_auth_failed, label: :status
```
Martouta: There is a lot of repetition in these lines and we should automate it a bit more at some point when we add one new kind or remove one or edit it somehow.
But again, just to keep in mind later on if we do :+1:

PR #1239, opened 2019-09-25
Implement search by pattern for Mapping Rules

duduribeiro:
This commit implements a search by pattern for mapping rules. It
includes a search field in the mapping rules page for both Backend
APIs and Services. It also adds an index for ProxyRule class using
Sphinx.

### DIFF
--- a/app/controllers/proxy_rule_shared_controller.rb
+++ b/app/controllers/proxy_rule_shared_controller.rb
@@ -12,9 +12,11 @@ module ProxyRuleSharedController
   end
 
   def index
-    @proxy_rules = proxy_rules.order_by(params[:sort], params[:direction])
-                              .includes(:metric)
-                              .paginate(pagination_params)
+    @search = ThreeScale::Search.new(params[:search])
+    query   = ProxyRuleQuery.new(owner_type: params[:owner_type], owner_id: owner_id,
+                                 direction: params[:direction], sort: params[:sort], per_page: per_page)
+    @proxy_rules = query.search_for(@search['query'], proxy_rules)
+                        .paginate(pagination_params)
   end
 
   def new
--- a/app/models/proxy_rule.rb
+++ b/app/models/proxy_rule.rb
@@ -3,7 +3,6 @@
 require 'addressable/template'
 
 class ProxyRule < ApplicationRecord
-
   acts_as_list scope: :proxy, add_new_at: :bottom
   scope :ordered, -> { order(position: :asc) }
 
@@ -16,6 +15,8 @@ class ProxyRule < ApplicationRecord
   validates :delta, numericality: { :only_integer => true, :greater_than => 0 }
 
   before_validation :fill_owner
+
+  after_commit { IndexProxyRuleWorker.perform_later(id) }
 
   include ThreeScale::Search::Scopes
 
--- a/app/queries/proxy_rule_query.rb
+++ b/app/queries/proxy_rule_query.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+class ProxyRuleQuery
+  def initialize(owner_type:, owner_id:, sort: nil, direction: nil, per_page: ThreeScale::Search::Helpers::MAX_PER_PAGE)
+    @direction  = direction
+    @owner_type = owner_type
+    @owner_id   = owner_id
+    @per_page   = per_page.to_i
+    @sort       = sort
+  end
+
+  def search_for(query, scope = ProxyRule.all)
+    scope = scope.order_by(@sort, @direction).includes(:metric)
+    return scope if query.blank?
+    options = {
+      ids_only: true, star: true, per_page: @per_page, ignore_scopes: true,
+      with: { owner_type: @owner_type, owner_id: @owner_id }
+    }
+    ids = ProxyRule.search(ThinkingSphinx::Query.escape(query), options)
+    scope.where(id: ids, owner_type: @owner_type, owner_id: @owner_id)
+  end
+end
--- a/app/views/shared/proxy_rules/_table.html.slim
+++ b/app/views/shared/proxy_rules/_table.html.slim
@@ -11,6 +11,13 @@ table#mapping_rules.data
       th = sortable('proxy_rules.position', 'Position')
       th.actions
         = link_to 'Add Mapping Rule', new_rule_path, class: 'action add'
+    tr.search
+      = search_form do |s|
+        th
+        th colspan="3"
+          = s.text_field :query, size: 20, class: :query, placeholder: 'Search for Pattern'
+        th colspan="4"
+          = s.submit 'Search'
   tbody
     - proxy_rules.each do |rule|
       tr
--- a/app/workers/application_job.rb
+++ b/app/workers/application_job.rb
@@ -1,0 +1,4 @@
+# frozen_string_literal: true
+
+class ApplicationJob < ActiveJob::Base
+end
--- a/app/workers/index_proxy_rule_worker.rb
+++ b/app/workers/index_proxy_rule_worker.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+class IndexProxyRuleWorker < ApplicationJob
+  def perform(proxy_rule_id)
+    proxy_rule = ProxyRule.find(proxy_rule_id)
+    ThinkingSphinx::RealTime::Callbacks::RealTimeCallbacks.new(
+      :proxy_rule
+    ).after_save(proxy_rule)
+  end
+end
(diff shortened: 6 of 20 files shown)


thomasmaas:
Screen capture?

duduribeiro:
search

@thomasmaas ☝️

duduribeiro:
search_2

test/unit/proxy_rule_test.rb
```diff
@@ -156,4 +156,18 @@ class ProxyRuleTest < ActiveSupport::TestCase
     assert_equal backend_api, backend_proxy_rule.owner
     assert backend_proxy_rule.valid?
   end
+
+  test '.by_query uses sphinx if query given' do
```
Martouta: Run it inside `ThinkingSphinx::Test.run do ... end`

test/unit/proxy_rule_test.rb
```diff
@@ -156,4 +156,18 @@ class ProxyRuleTest < ActiveSupport::TestCase
...
+    backend_api = FactoryBot.build_stubbed(:backend_api)
+    proxy_rule = FactoryBot.create(:proxy_rule, owner: backend_api, pattern: "/path/test")
+    ProxyRule.expects(:search).once.returns([proxy_rule])
+
+    assert_equal [proxy_rule], ProxyRule.by_query('test', proxy_rule.owner_type, proxy_rule.owner_id)
+  end
+
+  test '.by_query does not use sphinx if no query given' do
```
Martouta: Same as https://github.com/3scale/porta/pull/1239/files#r328479988

app/models/proxy_rule.rb
```diff
@@ -25,6 +28,16 @@ class ProxyRule < ApplicationRecord
 
   ALLOWED_HTTP_METHODS = %w[GET POST DELETE PUT PATCH HEAD OPTIONS].freeze
 
+  scope :by_query, ->(query, owner_type, owner_id) do
```
Martouta: RuboCop is complaining about this right now, but just rebase from master and it won't complain about this anymore :wink:

app/models/proxy_rule.rb
```diff
@@ -25,6 +28,16 @@ class ProxyRule < ApplicationRecord
...
+    return if query.blank?
```
Martouta:
```suggestion
    return none if query.blank?
```
Not sure if it will work, but then this way we return the same time and we can chain scopes... so it won't be null anymore. Btw, lambda checks that the params are passed so 'query' should never be nil :thinking: so imo we shouldn't even have this checking the type at all

app/models/proxy_rule.rb
```diff
@@ -25,6 +28,16 @@ class ProxyRule < ApplicationRecord
...
+    options = {
+                ids_only: true, star: true, per_page: 1_000_000, ignore_scopes: true,
```
Martouta: image
image

app/workers/index_proxy_rule_worker.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class IndexProxyRuleWorker < ActiveJob::Base
```
Martouta:
```suggestion
class IndexProxyRuleWorker < ActiveJob::Base # rubocop:disable Rails/ApplicationJob
```

app/views/shared/proxy_rules/_table.html.slim
```diff
@@ -11,6 +11,12 @@ table#mapping_rules.data
       th = sortable('proxy_rules.position', 'Position')
       th.actions
         = link_to 'Add Mapping Rule', new_rule_path, class: 'action add'
+    tr.search
+      = search_form do |s|
+        th colspan="6"
```
thomasmaas:
```suggestion
        th colspan="3"
```

No need for the search box to be this wide i think

thomasmaas: Also, search input should be in the second column as the first column is the verb

app/models/proxy_rule.rb
duduribeiro: My first shoot was returning a `none` relation. But it doesn't work because it ignores prepended and postponed relations in the query. eg:

with `return none ..` 
```
[1] pry(main)> ProxyRule.by_query('', 'Proxy', 22).where(position: 2).to_sql
=> ""
```
it returned the empty relation and ignored the where position.

with `return nil`
```
[1] pry(main)> ProxyRule.by_query('', 'Proxy', 22).where(position: 2).to_sql
=> "SELECT `proxy_rules`.* FROM `proxy_rules` WHERE `proxy_rules`.`position` = 2"
```
it skipped the where(id: ..) and made the `where(position: )` one.

app/workers/application_job.rb
```diff
@@ -0,0 +1,4 @@
+# frozen_string_literal: true
+
+class ApplicationJob < ActiveJob::Base
```
Martouta: :joy: :joy: :joy: :joy: :joy: 
Cadu 1 - RuboCop 0

features/support/env.rb
```diff
@@ -36,6 +36,10 @@ def initialize(a,b,c)
 
 transactional = non_transactional.map {|t| "~#{t}" }
 
+Before do
+  IndexProxyRuleWorker.stubs(:perform_later)
```
Martouta: Why? :thinking:

app/models/proxy_rule.rb
hallelujah: This could be a query see `app/queries`

features/support/env.rb
hallelujah: @Martouta 
It is run after_commit, so all tests with `disable_transactional_fixtures` and involving Proxy (basically all, because you create an account + service + proxy) will trigger indexation of Proxy.

We wanted to use ActiveJob::TestHelper `perform_enqueued_jobs` but it does not have the `:only` options in Rails 4.2, we need to update to rails 5.x

test/monkey_patches/active_job_test_helper.rb
```diff
@@ -0,0 +1,109 @@
+require 'active_job/test_helper'
+require 'active_job/queue_adapters/test_adapter'
+
+# TODO: Remove this Monkey patch after we move to Rails 5.
+# This is used to ensure that we run only a specific job in a Test.
+# It was extracted from
+# https://github.com/rails/rails/blob/fc5dd0b85189811062c85520fd70de8389b55aeb/activejob/lib/active_job/test_helper.rb#L368
```
hallelujah: This is acceptable! Really :)

Martouta:
Since you are going to push again anyway, personally I'd appreciate if the update/install of Gems could be a separate commit before the actual changes in the code :smile:

features/support/env.rb
duduribeiro: yeah, cucumber enables Sidekiq.inline! for all features, without this, all cucumber tests fails

app/controllers/proxy_rule_shared_controller.rb
```diff
@@ -12,7 +12,10 @@ def self.included(base)
   end
 
   def index
-    @proxy_rules = proxy_rules.order_by(params[:sort], params[:direction])
+    @search = ThreeScale::Search.new(params[:search])
+    query   = ProxyRuleQuery.new(owner_type: params[:owner_type], owner_id: owner_id)
+    @proxy_rules = query.search_for(@search['query'], proxy_rules)
+                              .order_by(params[:sort], params[:direction])
                               .includes(:metric)
```
hallelujah: Can those order and includes be in the query object?

duduribeiro: yeah, we can do it

app/queries/proxy_rule_query.rb
```diff
@@ -0,0 +1,21 @@
...
+    @owner_type = owner_type
+    @owner_id   = owner_id
+    @sort       = sort
+  end
+
+  def search_for(query, scope = ProxyRule.all)
+    scope = scope.order_by(@sort, @direction).includes(:metric)
+    return scope if query.blank?
+    options = {
+      ids_only: true, star: true, per_page: 1_000_000, ignore_scopes: true,
```
hallelujah: 1 million per page? Are you sure?

duduribeiro: yeah.. don't make sense haha.. will add the same maximum as we paginate.

app/queries/proxy_rule_query.rb
```diff
@@ -0,0 +1,22 @@
...
+    @owner_id   = owner_id
+    @per_page   = per_page.to_i
+    @sort       = sort
+  end
+
+  def search_for(query, scope = ProxyRule.all)
+    scope = scope.order_by(@sort, @direction).includes(:metric)
+    return scope if query.blank?
+    options = {
+      ids_only: true, star: true, per_page: 20, ignore_scopes: true,
```
hallelujah: Ouch I dis not check correctly
The per page does not use the variable in. Initialize

PR #1241, opened 2019-09-25
Ensure uniqueness of service_id + backend_api_id

Martouta:
We need this for now for the API. Because we are hiding the 'backend api config' from the customer, so we are not returning its ID, so without this PR, the only way to identify a backend api config in `/admin/api/services/:service_id/backend_apis/` to update it or destroy it is by the field path, which is clearly a bad idea because it would be part of the path of the request and because it can be empty.
Which this PR, it is just the ID of the backend_api :smile:

### DIFF
--- a/app/models/backend_api.rb
+++ b/app/models/backend_api.rb
@@ -34,25 +34,6 @@ class BackendApi < ApplicationRecord
 
   scope :orphans, -> { where.has { id.not_in(BackendApiConfig.selecting { :backend_api_id }) } }
 
-  scope :not_used_by, ->(service_id) {
-    # TODO: Baby Squeel
-    # It used to be:
-    # where.has do
-    #   not_exists BackendApiConfig.by_service(service_id).by_backend_api(BabySqueel[:backend_apis].id).select(:id)
-    # end
-    # And that works for MySQL and Postgres but not Oracle
-    sql_query = <<~SQL
-      (
-        NOT EXISTS (
-          SELECT id
-          FROM backend_api_configs
-          WHERE service_id = ? AND backend_api_configs.backend_api_id = backend_apis.id
-        )
-      )
-    SQL
-    where(sql_query, service_id)
-  }
-
   scope :oldest_first, -> { order(created_at: :asc) }
   scope :accessible, -> { where.not(state: DELETED_STATE) }
 
@@ -65,6 +46,12 @@ class BackendApi < ApplicationRecord
     end
 
     after_transition to: [DELETED_STATE], do: :schedule_deletion
+  end
+
+  def self.not_used_by(service_id)
+    where.has do
+      not_exists BackendApiConfig.by_service(service_id).by_backend_api(BabySqueel[:backend_apis].id).select(:id)
+    end
   end
 
   def self.default_api_backend
--- a/app/models/backend_api.rb
+++ b/app/models/backend_api.rb
@@ -36,11 +36,7 @@ class BackendApi < ApplicationRecord
 
   scope :not_used_by, -> do |service_id|
     where.has do
-      not_exists(
-        BackendApiConfig
-        .by_service(service_id)
-        .by_backend_api(BabySqueel[:backend_apis].id)
-      )
+      not_exists BackendApiConfig.by_service(service_id).by_backend_api(BabySqueel[:backend_apis].id).select(:id)
     end
   end
 
--- a/app/models/backend_api.rb
+++ b/app/models/backend_api.rb
@@ -34,7 +34,7 @@ class BackendApi < ApplicationRecord
 
   scope :orphans, -> { where.has { id.not_in(BackendApiConfig.selecting { :backend_api_id }) } }
 
-  scope :without_service, -> do |service_id|
+  scope :not_used_by, -> do |service_id|
     where.has do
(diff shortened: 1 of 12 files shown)


thomasmaas:
Ok, we still need the select of backends from @gui's pull request only to contain backends that haven't been added yet.

test/integration/provider/admin/backend_apis_controller_test.rb
```diff
@@ -58,7 +58,7 @@ class Provider::Admin::BackendApisControllerTest < ActionDispatch::IntegrationTe
 
   test 'delete a backend api with products' do
     backend_api = @provider.backend_apis.order(:id).first
-    FactoryBot.create(:backend_api_config, service: @provider.first_service, backend_api: backend_api)
+    FactoryBot.create(:backend_api_config, service: FactoryBot.create(:simple_service, account: @provider), backend_api: backend_api)
```
guicassolato: If you had to change to different the service here, it's because a `BackendApiConfig` already existed with `backend_api`. I think if you remove this line the test will pass just the same.

app/models/backend_api.rb
```diff
@@ -34,6 +34,16 @@ class BackendApi < ApplicationRecord
 
   scope :orphans, -> { where.has { id.not_in(BackendApiConfig.selecting { :backend_api_id }) } }
 
+  scope :without_service, (lambda do |param_service_id|
```
guicassolato:
```suggestion
  scope :without_service, ->(service_id) do
```

app/models/backend_api.rb
```diff
@@ -34,6 +34,16 @@ class BackendApi < ApplicationRecord
...
+    where.has do
+      not_exists(
+        BackendApiConfig
+        .by_service(param_service_id)
```
guicassolato:
```suggestion
        .by_service(service_id)
```

app/models/backend_api.rb
```diff
@@ -34,6 +34,16 @@ class BackendApi < ApplicationRecord
...
+        .by_backend_api(BabySqueel[:backend_apis].id)
+      )
+    end
+  end)
```
guicassolato:
```suggestion
  end
```

app/models/backend_api.rb
Martouta: Then RuboCop would complain

app/models/backend_api.rb
Martouta: Then Rubocop would complain

app/models/backend_api.rb
Martouta: with the `->` i mean, because it is multiline

guicassolato: Then I disagree with Rubocop. The way I suggested is more readable IMO.

guicassolato: It complains about multiline when we use `{..}` instead of `begin..end` I think.

app/models/backend_api_config.rb
```diff
@@ -18,9 +18,13 @@ class BackendApiConfig < ApplicationRecord
   has_many :backend_api_metrics, through: :backend_api, source: :metrics
 
   validates :service_id, :backend_api_id, presence: true
+  validates :backend_api_id, uniqueness: { scope: :service_id }
   validates :path, uniqueness: { scope: :service_id, case_sensitive: false }
   validates :path, length: { in: 0..255, allow_nil: false }, path: true
 
+  scope :by_service,     ->(param_service_id)     { where.has { service_id     == param_service_id     } }
```
guicassolato:
```suggestion
  scope :by_service,     ->(service_id)     { where.has { self.service_id     == service_id     } }
```

app/models/backend_api_config.rb
```diff
@@ -18,9 +18,13 @@ class BackendApiConfig < ApplicationRecord
...
+  scope :by_backend_api, ->(param_backend_api_id) { where.has { backend_api_id == param_backend_api_id } }
```
guicassolato:
```suggestion
  scope :by_backend_api, ->(backend_api_id) { where.has { self.backend_api_id == backend_api_id } }
```

app/models/backend_api.rb
Martouta: It still complains :disappointed:

app/models/backend_api.rb
```diff
@@ -34,6 +34,16 @@ class BackendApi < ApplicationRecord
 
   scope :orphans, -> { where.has { id.not_in(BackendApiConfig.selecting { :backend_api_id }) } }
 
+  scope :without_service, (lambda do |service_id|
```
guicassolato:
```suggestion
  scope :not_used_by, (lambda do |service_id|
```

Martouta: `not _used_by` lacks a what :smile: like: `not_used_by_product` :thinking: 
WDYT @guicassolato ?

Martouta:
I will wait for tests of all DBs to know if it works or not. After that, I will push at least these 2 changes:
- [Rename](https://github.com/3scale/porta/pull/1241#discussion_r328269969) `without_service` to `not_used_by`
- Update the lambda after https://github.com/3scale/porta/pull/1245 is merged

PR #1242, opened 2019-09-25
Integration Settings part 3

didierofrivia:
Screenshot 2019-09-25 at 17 37 09

This needs to be merge after #1229

This is the third iteration of the unified integration settings, hides the old configuration settings

Which issue(s) this PR fixes

Partially fixes https://issues.jboss.org/browse/THREESCALE-3451

Verification steps

TODO

Special notes for your reviewer:

### DIFF
--- a/app/controllers/api/integrations_controller.rb
+++ b/app/controllers/api/integrations_controller.rb
@@ -87,6 +87,7 @@ class Api::IntegrationsController < Api::BaseController
   end
 
   def show
+    @apiap = apiap?
     respond_to do |format|
       format.html do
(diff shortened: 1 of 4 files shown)


app/controllers/api/integrations_controller.rb
```diff
@@ -87,6 +87,7 @@ def promote_to_production
   end
 
   def show
+    @apiap = apiap?
```
Martouta: This is to be able to use it from the views... so imo it should be a helper instead :thinking: 
but i know it is a temporary thing bcz soon everything will be apiap and we will remove this, so ok :+1:

PR #1243, opened 2019-09-25
More agressive APIcast upgrade message plus removal of downgrading in ui

thomasmaas:
closes https://issues.jboss.org/browse/THREESCALE-3266

Screenshot 2019-09-25 19 10 41

Next steps would be to have custom policies fully operational in SaaS (@hallelujah), agree on decommissioning date (@3scale/product) and communicate it.

### DIFF
--- a/features/step_definitions/apicast_steps.rb
+++ b/features/step_definitions/apicast_steps.rb
@@ -16,3 +16,8 @@ Given(/^apicast registry is undefined$/) do
   ThreeScale.config.sandbox_proxy.stubs(:self_managed_apicast_registry_url).returns(nil)
   JSONClient.expects(:get).with(nil).raises(SocketError)
 end
+
+Given(/^I toggle the apicast version$/) do
+  proxy = @provider.default_service.proxy
+  proxy.toggle!(:apicast_configuration_driven) if !proxy.oidc? || !proxy.apicast_configuration_driven # rubocop:disable Rails/SkipsModelValidations
+end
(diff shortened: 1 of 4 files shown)


features/step_definitions/apicast_steps.rb
```diff
@@ -16,3 +16,8 @@
   ThreeScale.config.sandbox_proxy.stubs(:self_managed_apicast_registry_url).returns(nil)
   JSONClient.expects(:get).with(nil).raises(SocketError)
 end
+
+Given(/^I toggle the apicast version$/) do
+  proxy = @provider.default_service.proxy
+  proxy.toggle!(:apicast_configuration_driven) if !proxy.oidc? || !proxy.apicast_configuration_driven # rubocop:disable Rails/SkipsModelValidations
```
Martouta: This currently does exactly the same as the controller:
https://github.com/3scale/porta/blob/4d295c0bfacf59a45a087895355e98e5676c9ad8/app/controllers/api/integrations_controller.rb#L115-L134

This is because before this branch, the cukes where doing it through the UI, but now it is removed from the UI and the cukes must test it for accounts that already have these values from before this PR, so now it is a 'given'.

Martouta: it should work because it does the same as the controller
it is very bad because it does not set the same value for each cuke, but more like "depending on what u have at this moment of the code, change it to another value"
it is asking for problems because it is not "given this value..."
maybe the given ends up with a value that is not what we want...
i will fix it in another commit

PR #1244, opened 2019-09-25
Proxy config affecting change events

guicassolato:

It defines a new event to be issued whenever an object that is part of a proxy config changes (or gets created, or destroyed). This event will asynchronously touch a tracking record in a separate table, also introduced, that holds a 1-1 relationship with proxy.

This will make possible for the UI to highlight for the user a potential outdated proxy config, thus being relevant to https://github.com/3scale/porta/pull/1235.

Closes [THREESCALE-3555](https://issues.jboss.org/browse/THREESCALE-3555)

**TODOs**
- [x] Create table to store the change history on proxy config affecting objects
- [x] Generate the schema for Oracle and PostgreSQL
- [x] Define a new type of event to be issued whenever a proxy config affecting change occurs (plus subscriber and worker)
- [x] Issue the event on every model of proxy config affecting object (in a `after_commit`, for each proxy potentially affected)
  - `ProxyRule` (`Metric` deletion will kick `ProxyRule` deletion, so already covered)
  - ~`Policy`~
  - `Proxy` (Maybe only a few attributes – `policies_config`)
  - `GatewayConfiguration`

Not sure if we should issue one event per proxy (passing proxy and affecting object) or just one event per change, passing only the affecting object and let the creation of the event to handle the logics of triggering one background job per proxy. In the latter case, the event would have to be aware of the different types of affecting objects (so it can get to the proxies).

To sum up (very roughly exemplifying):

*Option A*
```
# where the event is triggered (many places in the code) 
Event.new(object)

# what the event does
for each proxy using object
  Notifier.new(proxy, object)
```

*Option B*
```
# where the event is triggered (many places in the code) 
for each proxy using object
  Event.new(proxy, object)

# what the event does
Notifier.new(proxy, object)
```

_Update:_ going with Option B.

----

I won't tag this as APIAP-only because I think this can be used for non-APIAP accounts as well.

### DIFF
--- a/app/lib/proxy_config_affecting_changes.rb
+++ b/app/lib/proxy_config_affecting_changes.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+module ProxyConfigAffectingChanges
+  module ProxyRuleExtension
+    extend ActiveSupport::Concern
+
+    included do
+      include ProxyConfigAffectingChanges
+
+      after_commit :issue_proxy_affecting_change_events
+
+      def issue_proxy_affecting_change_events
+        case owner
+        when Proxy
+          issue_proxy_affecting_change_event(owner)
+        when BackendApi
+          owner.proxies.each(&method(:issue_proxy_affecting_change_event))
+        end
+      end
+    end
+  end
+
+  module ProxyExtension
+    extend ActiveSupport::Concern
+
+    PROXY_CONFIG_AFFECTING_ATTRIBUTES = %w[policies_config].freeze # TODO: add more attributes here
+
+    included do
+      include ProxyConfigAffectingChanges
+
+      after_commit :issue_proxy_affecting_change_events, on: :update
+
+      def issue_proxy_affecting_change_events
+        changes_attributes = previous_changes.keys
+        return if changes_attributes.include?('created_at') || (changes_attributes & PROXY_CONFIG_AFFECTING_ATTRIBUTES).empty?
+        issue_proxy_affecting_change_event(self)
+      end
+    end
+  end
+
+  def issue_proxy_affecting_change_event(proxy)
+    ProxyConfigs::AffectingObjectChangedEvent.create_and_publish!(proxy, self)
+  end
+end
--- a/app/models/proxy.rb
+++ b/app/models/proxy.rb
@@ -6,6 +6,7 @@ class Proxy < ApplicationRecord
   include BackendApiLogic::ProxyExtension
   prepend BackendApiLogic::RoutingPolicy
   include GatewaySettings::ProxyExtension
+  include ProxyConfigAffectingChanges::ProxyExtension
 
   DEFAULT_POLICY = { 'name' => 'apicast', 'humanName' => 'APIcast policy', 'description' => 'Main functionality of APIcast.',
                      'configuration' => {}, 'version' => 'builtin', 'enabled' => true, 'removable' => false, 'id' => 'apicast-policy'  }.freeze
--- a/app/models/proxy.rb
+++ b/app/models/proxy.rb
@@ -20,6 +20,9 @@ class Proxy < ApplicationRecord
 
   validates :error_status_no_match, :error_status_auth_missing, :error_status_auth_failed, :error_status_limits_exceeded, presence: true
 
+  has_one :proxy_config_affecting_change, dependent: :delete
+  private :proxy_config_affecting_change
+
   uri_pattern = URI::DEFAULT_PARSER.pattern
 
   URI_OPTIONAL_PORT = /\Ahttps?:\/\/[a-zA-Z0-9._-]*(:\d+)?\Z/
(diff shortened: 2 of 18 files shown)


app/models/proxy.rb
```diff
@@ -511,6 +514,23 @@ def service_mesh_integration?
     Service::DeploymentOption.service_mesh.include?(deployment_option)
   end
 
+  def find_or_create_proxy_config_affecting_change
+    proxy_config_affecting_change || create_proxy_config_affecting_change
+  end
+  alias affecting_change_history find_or_create_proxy_config_affecting_change
+  private :find_or_create_proxy_config_affecting_change
+
+  def pending_affecting_changes??(environment: :sandbox)
```
Martouta:
```suggestion
  def pending_affecting_changes?(environment: :sandbox)
```

guicassolato: oops

app/lib/proxy_config_affecting_changes.rb
```diff
@@ -0,0 +1,45 @@
...
+    PROXY_CONFIG_AFFECTING_ATTRIBUTES = %w[policies_config].freeze # TODO: add more attributes here
+
+    included do
+      include ProxyConfigAffectingChanges
+
+      after_commit :issue_proxy_affecting_change_events, on: :update
+
+      def issue_proxy_affecting_change_events
+        changes_attributes = previous_changes.keys
+        return if changes_attributes.include?('created_at')
```
hallelujah:
```suggestion
        return if created_at_changed? || (changes_attributes & PROXY_CONFIG_AFFECTING_ATTRIBUTES).empty?
```

app/models/proxy.rb
```diff
@@ -511,6 +515,23 @@ def service_mesh_integration?
...
+  private :find_or_create_proxy_config_affecting_change
+
+  def pending_affecting_changes?(environment: :sandbox)
+    return unless apicast_configuration_driven?
+    config = latest_config(environment: environment)
+    return false unless config
+    config.created_at < affecting_change_history.updated_at
+  end
+
+  def latest_config(environment: :sandbox)
```
hallelujah: I think we won't need it for production, as production is always a promotion from sandbox
Diff between production and sandbox can already be done by checking the ProxyConfig

app/models/proxy.rb
```diff
@@ -515,6 +519,19 @@ def service_mesh_integration?
     Service::DeploymentOption.service_mesh.include?(deployment_option)
   end
 
+  def find_or_create_proxy_config_affecting_change
+    proxy_config_affecting_change || create_proxy_config_affecting_change
+  end
+  alias affecting_change_history find_or_create_proxy_config_affecting_change
+  private :find_or_create_proxy_config_affecting_change
+
+  def pending_affecting_changes?
```
hallelujah: Using this method (probably in the view) creates records right?
If yes, can we avoid that somehow?

app/lib/proxy_config_affecting_changes.rb
guicassolato: No, it's not `changed_attributes`, it's `changes_attributes`, a variable defined in the line before.

app/models/proxy.rb
guicassolato: Well, if we have a proxy config, then we can also have an affecting change history I think. It also saves us from having to migrate data.

app/lib/proxy_config_affecting_changes.rb
```diff
@@ -0,0 +1,44 @@
...
+    PROXY_CONFIG_AFFECTING_ATTRIBUTES = %w[policies_config].freeze # TODO: add more attributes here
+
+    included do
+      include ProxyConfigAffectingChanges
+
+      after_commit :issue_proxy_affecting_change_events, on: :update
+
+      def issue_proxy_affecting_change_events
+        changes_attributes = previous_changes.keys
+        return if changes_attributes.include?('created_at') || (changes_attributes & PROXY_CONFIG_AFFECTING_ATTRIBUTES).empty?
```
hallelujah: So you can use `previously_changed?` https://github.com/3scale/porta/blob/d21f8d5e7a6dc740db6652e0604dcf0b639edb52/lib/previously_changed.rb

app/lib/proxy_config_affecting_changes.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    PROXY_CONFIG_AFFECTING_ATTRIBUTES = %w[policies_config].freeze # TODO: add more attributes here
+
+    included do
+      include ProxyConfigAffectingChanges
+
+      after_commit :issue_proxy_affecting_change_events, on: :update
+
+      def issue_proxy_affecting_change_events
+        changes_attributes = previous_changes.keys
```
hallelujah: So you can use `previously_changed?` https://github.com/3scale/porta/blob/d21f8d5e7a6dc740db6652e0604dcf0b639edb52/lib/previously_changed.rb

Which will be introduced in rails 5.0

app/models/proxy.rb
hallelujah: fair enough

PR #1245, opened 2019-09-25
set rubocop lambda style to literal

guicassolato:
So

```ruby
scope :blah, (lambda do |param|
  # ...
  param...
  #...
  # ...
end)
```

can be written 

```ruby
scope :blah, ->(param) do
  # ...
  param...
  #...
  # ...
end
```

See https://github.com/rubocop-hq/rubocop/issues/1520

### DIFF
--- a/app/models/deleted_object.rb
+++ b/app/models/deleted_object.rb
@@ -8,7 +8,7 @@ class DeletedObject < ApplicationRecord
     scope scoped_class.to_s.underscore.pluralize.to_sym, -> { where(object_type: scoped_class) }
   end
 
-  scope(:deleted_owner, lambda do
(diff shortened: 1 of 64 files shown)


app/models/deleted_object.rb
```diff
@@ -8,7 +8,7 @@ class DeletedObject < ApplicationRecord
     scope scoped_class.to_s.underscore.pluralize.to_sym, -> { where(object_type: scoped_class) }
   end
 
-  scope(:deleted_owner, lambda do
+  scope(:deleted_owner, -> do
```
Martouta: Perhaps now we can do: `scope :deleted_owner, -> do`
And then in like 18: `end`

PR #1249, opened 2019-09-26
[Rolling Updates] Plugin deployment option are deprecated

hallelujah:
We want to deprecate Plugins
This is the first step to do it, on-prem they should not be shown anymore

We still need to think of what to do with customers that use it.
My suggestion is not to have a migration path, either they use APIcast V2 or nothing

Partially closes THREESCALE-2395

### DIFF
--- a/app/helpers/services_helper.rb
+++ b/app/helpers/services_helper.rb
@@ -36,19 +36,6 @@ module ServicesHelper
   end
 
   def deployment_options
-    options = Service.deployment_options(current_account)
-    {}.merge(options['Gateway'])
-        .merge(service_mesh_active? ? options['Service Mesh'] : {})
-        .merge(plugins_active? ? options['Plugin'] : {})
-  end
-
-  private
-
-  def service_mesh_active?
-    current_account.provider_can_use?(:service_mesh_integration)
-  end
-
-  def plugins_active?
-    false
+    Service.deployment_options(current_account).values.reduce(&:merge)
   end
 end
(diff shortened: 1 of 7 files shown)


app/concerns/api_as_product.rb
```diff
@@ -0,0 +1,15 @@
...
+  module Controllers
+    extend ActiveSupport::Concern
+
+    included do
+      helper_method :apiap?
+    end
+
+    protected
+
+    def apiap?
```
guicassolato: 😍

thomasmaas:
Migration path is not needed, they can keep using unsupported plugins. They just won't see it reflected in the UI. And we no longer need to store it as an option in the db. but question remains, do we change that field in their db or do we just remove logic in view layer based on that setting?

hallelujah:
This will remove the view on-prem
I do not want to change anything now.

thomasmaas:
@hallelujah I'm seeing this:

Screenshot 2019-10-02 21 09 24

we were already not showing plugins on this page but this pr makes them show again…

hallelujah:
Well lets close it then ...
I would have preferred a way to deprecate smoothly the plugins when an agreement is reached than just removing from the UI but OK.

I think this PR should have been merged 1 week ago before extracting the settings to other pages.

hallelujah:
Reopening as we want to deprecate nicely the plugins

hallelujah:
Closing. Lets remove the code definitively without transition instead

PR #1255, opened 2019-09-27
[settings] Auth settings behaviour updated

didierofrivia:
istio-auth

This is the final part of the integration settings

Fixes https://issues.jboss.org/browse/THREESCALE-3559

Notes for reviewers:
* This needs to be migrated to React component
* Needs a cuke

### DIFF
--- a/app/javascript/src/Settings/authentication-widget.jsx
+++ b/app/javascript/src/Settings/authentication-widget.jsx
@@ -3,6 +3,10 @@ import 'Settings/styles/authentication.scss'
 const AUTH_WRAPPER_ID = 'auth-wrapper'
 const AUTH_METHOD_CLASS = 'auth-method'
 const AUTH_SETS_CLASS = 'auth-settings'
+const INTEGRATION_WRAPPER_ID = 'integration-wrapper'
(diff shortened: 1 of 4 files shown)


app/javascript/src/Settings/authentication-widget.jsx
```diff
@@ -3,16 +3,23 @@ import 'Settings/styles/authentication.scss'
 const AUTH_WRAPPER_ID = 'auth-wrapper'
 const AUTH_METHOD_CLASS = 'auth-method'
 const AUTH_SETS_CLASS = 'auth-settings'
+const INTEGRATION_WRAPPER_ID = 'integration-wrapper'
+const INTEGRATION_METHODS_CLASS = 'integration-method'
+const ISTIO_ID = 'service_deployment_option_service_mesh_istio'
```
hallelujah: Lets use service mesh instead of Istio, be as generic as possible

thomasmaas:
@hallelujah there were things missing in this pull request. @didierofrivia knows. It’s js. We should fix before deploying.

hallelujah:
@thomasmaas that's fine, I told him to make small PR even not functionnaly complete so they can be reviewed and merged quickly

PR #1256, opened 2019-09-27
Policies update button

josemigallas:

* Hides Form's submit button when a policy is being edited (so that there is only 1 button at the same time)
* Renders Form's submit button disabled by default and enables it after any change (so that it makes more obvious that the changes are yet to be saved)

[THREESCALE-3546: Policies: remove double update button](https://issues.jboss.org/browse/THREESCALE-3546)

policies


app/javascript/src/Policies/reducers/PolicyChain.jsx
```diff
@@ -22,6 +22,12 @@ function addPolicy (state: ChainState, action: AddPolicyToChainAction): ChainSta
 }
 
 function updateChain (_state: ChainState, action: UpdatePolicyChainAction): ChainState {
+  // HACK: enable the submit button after any change is made
```
damianpm: Hmmm, if this is a reducer then it shouldn't modify the dom.
If this hack is urgent, then please create a task to improve it

josemigallas: I think it's going to be removed whenever we refactor the policies anyways. The reason to do it this way it is not to add much more complexity for a thing that.. shouldn't be done, at all.. for starters the submit button should be part of the React component.

damianpm: Yes, but is not clear what `whenever we refactor` means, it may be next week or next year. If we have a task to refactor policies, just add a comment, if not create a new one

app/views/api/policies/edit.html.slim
```diff
@@ -3,4 +3,4 @@
 = semantic_form_for @proxy, url: admin_service_policies_path(@service) do |f|
   = render 'api/integrations/apicast/shared/policies'
   = f.buttons do
-    = f.commit_button t('formtastic.actions.policies.update'), button_html: {class: 'important-button', name: 'update', id: 'policies-button-sav', title: 'Update Policies'}
+    = f.commit_button t('formtastic.actions.policies.update'), button_html: {class: 'disabled-button', name: 'update', id: 'policies-button-sav', title: 'Update Policies', disabled: true}
```
guicassolato: How about making the value of this button "Update Policy Chain" so it's double clear "Update Policy" is one thing, "Update Policy Chain" is another?

app/javascript/src/Policies/reducers/PolicyChain.jsx
didierofrivia: This is way to hacky. Ideally we could check the pristine state with Redux, but if we can't...what about saving the original json (the proxy config generated) and compare it when rendering the form? That'd be only a string comparison, not so expensive

josemigallas: I can try that

PR #1257, opened 2019-09-27
Extract apiap data migrations to rake tasks

guicassolato:

Rails migrations should not really contain changing-state DML code (INSERT, UPDATE, DELETE), as they are not sufficiently tested and tend not to follow changes in the code.

Closes [THREESCALE-3556](https://issues.jboss.org/browse/THREESCALE-3556)
 
**Verification steps** 

Try upgrade 3scale from 2.6 to a newer version. You should see no migration exceptions.

**Missing**
- [ ] Add tests

### DIFF
--- a/lib/tasks/proxy.rake
+++ b/lib/tasks/proxy.rake
@@ -1,3 +1,8 @@
+# frozen_string_literal: true
+
+require 'benchmark'
+require 'progress_counter'
+
 namespace :proxy do
   desc "create proxies for every service"
   task :create_proxies => :environment do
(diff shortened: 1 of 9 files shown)


hallelujah:
Thank you 🚀

lib/tasks/proxy.rake
```diff
@@ -32,4 +37,17 @@ namespace :proxy do
   task :set_correct_endpoint_hosted => :environment do
     Service.includes(:proxy).where(proxies: {endpoint: nil}, deployment_option: 'hosted').find_each(&:deployment_option_changed)
   end
+
+  desc 'Update proxy rules owners'
+  task update_proxy_rule_owners: :environment do
```
hallelujah: So I think for now it is acceptable to just document it.
Let see later if we can have a mechanism to make some post-migration automatic and resilient over uprades

PR #1258, opened 2019-09-27
3265 endpoint objects status

macejmic:

API endpoint: object deletion status for objects that are deleted asynchronously

https://issues.jboss.org/browse/THREESCALE-3265

### DIFF
--- a/app/controllers/admin/api/objects_controller.rb
+++ b/app/controllers/admin/api/objects_controller.rb
@@ -57,7 +57,7 @@ class Admin::Api::ObjectsController < Admin::Api::BaseController
   ##~ op = e.operations.add
   ##~ op.httpMethod = "GET"
   ##~ op.summary = "Object deletion status for objects that are deleted asynchronously"
-  ##~ op.description = "Returns an object status. (200/404). Useful for those objects that deleted asynchronously in order to know if the deletion has been completed(404) or not(200)"
+  ##~ op.description = "Returns an object status (200/404). Useful for those objects that are deleted asynchronously in order to know if the deletion has been completed(404) or not(200)."
   ##~ op.group = "objects"
   #
   ##~ op.parameters.add @parameter_access_token
--- a/app/controllers/admin/api/objects_controller.rb
+++ b/app/controllers/admin/api/objects_controller.rb
@@ -1,0 +1,95 @@
+# frozen_string_literal: true
+
+class Admin::Api::ObjectsController < Admin::Api::BaseController
+
+  class AllowedStatusObject
+    attr_reader :controller, :status_object
+
+    def initialize(controller, status_object)
+      @controller = controller
+      @status_object = status_object
+    end
+
+    def authorize
+      raise NotImplementedError
+    end
+
+    private
+
+    def raise_access_denied
+      raise CanCan::AccessDenied
+    end
+  end
+
+  class BuyerAccountObject < AllowedStatusObject
+    def authorize
+      controller.authorize! :read, status_object
+    end
+  end
+
+  class ServiceObject < AllowedStatusObject
+    def authorize
+      raise_access_denied unless controller.accessible_services.exists?(service_id)
+    end
+
+    def service_id
+      controller.status_object_id
+    end
+  end
+
+  class ProxyObject < ServiceObject
+    delegate :service_id, to: :status_object
+  end
+
+  class BackendApiObject < AllowedStatusObject
+    def authorize
+      controller.provider_can_use!(:api_as_product)
+    end
+  end
+
+  ALLOWED_OBJECTS = %w[service buyer_account proxy backend_api].freeze
+
+  before_action :authorize_status_object_type
+
+  ##~ e = sapi.apis.add
+  ##~ e.path = "/admin/api/objects/status.json"
+  #
+  ##~ op = e.operations.add
+  ##~ op.httpMethod = "GET"
+  ##~ op.summary = "Object deletion status for objects that are deleted asynchronously"
+  ##~ op.description = "Returns an object status. (200/404). Useful for those objects that deleted asynchronously in order to know if the deletion has been completed(404) or not(200)"
+  ##~ op.group = "objects"
+  #
+  ##~ op.parameters.add @parameter_access_token
+  ##~ op.parameters.add :dataType => "string", :required => true, :paramType => "query", :name => "object_type", :description => "Object type has to be service, buyer_account, proxy or backend_api."
+  ##~ op.parameters.add :dataType => "string", :required => true, :paramType => "query", :name => "object_id", :description => "Object ID."
+  def status
+    status_object = current_account.public_send(association_name).find(status_object_id)
+
+    authorize_status_object(status_object)
+
+    head :ok
+  end
+
+  def status_object_type
+    @status_object_type ||= params.fetch(:object_type).to_s
+  end
+
+  def status_object_id
+    @status_object_id ||= params.fetch(:object_id)
+  end
+
+  private
+
+  def association_name
+    status_object_type.pluralize
+  end
+
+  def authorize_status_object_type
+    raise(CanCan::AccessDenied) if ALLOWED_OBJECTS.exclude?(status_object_type)
+  end
+
+  def authorize_status_object(status_object)
+    "#{self.class}::#{status_object_type.classify}Object".constantize.new(self, status_object).authorize
+  end
+end
--- a/app/lib/logic/rolling_updates.rb
+++ b/app/lib/logic/rolling_updates.rb
@@ -360,11 +360,11 @@ module Logic
         klass.helper_method :provider_can_use?
       end
 
-      protected
-
       def provider_can_use!(feature)
         raise ActiveRecord::RecordNotFound unless provider_can_use?(feature)
       end
+
+      protected
 
       def provider_can_use?(fresh_feature)
         return false if Logic::RollingUpdates.skipped?
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -487,6 +487,8 @@ without fake Core server your after commit callbacks will crash and you might ge
       # called from heroku deploy hook after heroku proxy deploy script
       post 'heroku-proxy/deployed', to: 'heroku_proxy#deployed', as: :heroku_proxy_deployed
 
+      get 'objects/status' => 'objects#status', as: :objects_status, controller: :objects
+
       namespace :personal, defaults: { format: :json } do
         resources :access_tokens, except: %i[new edit]
       end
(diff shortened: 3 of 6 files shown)


doc/active_docs/Account Management API.json
```diff
@@ -5644,6 +5644,41 @@
         }
       ]
     },
+    {
+      "path": "/admin/api/objects/status.json",
+      "operations": [
+        {
+          "httpMethod": "GET",
+          "summary": "Object status",
```
thomasmaas:
```suggestion
          "summary": "Object deletion status for objects that are deleted asynchronously",
```

doc/active_docs/Account Management API.json
```diff
@@ -5644,6 +5644,41 @@
...
+          "description": "Returns an object status. (200/404)",
```
thomasmaas:
```suggestion
          "description": "Returns an object status. (200/404). Useful for those objects that deleted asynchronously in order to know if the deletion has been completed(404) or not(200)",
```

doc/active_docs/Account Management API.json
```diff
@@ -5644,6 +5644,41 @@
...
+              "required": true,
+              "paramType": "query",
+              "threescale_name": "access_token"
+            },
+            {
+              "dataType": "string",
+              "required": true,
+              "paramType": "query",
+              "name": "object_type",
+              "description": "Object type could be service, account, proxy or backend_api."
```
thomasmaas:
```suggestion
              "description": "Object type has to be service, account, proxy or backend_api."
```

config/routes.rb
```diff
@@ -203,6 +203,8 @@ def call(env)
     get 'admin', to: 'admin#show'
 
     namespace :admin do
+      get 'objects/status' => 'objects#status', as: :objects_status
```
Martouta:
```suggestion
      get 'objects/status' => 'objects#status', as: :objects_status, controller: :objects
```

doc/active_docs/Account Management API.json
Martouta: The ActiveDocs are autogenerated, so this shouldn't be changed here but in the documentation in the controller :smile:

macejmic:
@Martouta 

1. I don't think we need to check the permissions as we're not returning any attributes about the reading object. So there are not violation that we would broke return just an object delation info. 

2. I really that this endpoint is super helpful for saas providers too. As the background delation was mainly designed because of them. (and also for a provider delation of course). Making this for master only would means this feature would be useful only for on-prem customers. (as we don't need it)

Scoping services etc. by the current account would be good. But if user is trying to get status of account, that's not the case.

thomasmaas:
- Agreed that this endpoint should exist at provider level
- Not agreed that just because we don't return any specifics we shouldn't take roles into acount: the fact something is there or not is information so should be treated as such

Martouta:
@macejmic 

1. At least let's just check that the account checking this has the same tenant_id as the object being checked :smile: 

2. That makes sense, so imo it should be both for master and for tenant (the endpoint), but for tenant is harder for what I said of the endpoint not available anymore. In case that it helps, we fixed this for Zync doing this:
https://github.com/3scale/porta/blob/b6ff26ffe59feeee0aedc9a4cfefd90ca49d02ba/lib/routing_constraints.rb#L28-L29

macejmic:
1. yep, agree that we should check at least the `tenant_id`

2. this use case is when `current_account.master?` and then you don't have to fix anything

ok, let me change the logic and i'll get back to you.. thanks!

app/controllers/admin/api/objects_controller.rb
```diff
@@ -0,0 +1,95 @@
...
+
+  before_action :authorize_status_object_type
+
+  ##~ e = sapi.apis.add
+  ##~ e.path = "/admin/api/objects/status.json"
+  #
+  ##~ op = e.operations.add
+  ##~ op.httpMethod = "GET"
+  ##~ op.summary = "Object deletion status for objects that are deleted asynchronously"
+  ##~ op.description = "Returns an object status. (200/404). Useful for those objects that deleted asynchronously in order to know if the deletion has been completed(404) or not(200)"
```
thomasmaas:
```suggestion
  ##~ op.description = "Returns an object status (200/404). Useful for those objects that are deleted asynchronously in order to know if the deletion has been completed(404) or not(200)."
```

macejmic:
@thomasmaas done.

Martouta:
@macejmic rebase plz :smile:

macejmic:
@Martouta done.

test/integration/admin/api/objects_controller_test.rb
```diff
@@ -0,0 +1,80 @@
...
+  def setup
+    @provider = FactoryBot.create(:provider_account)
+
+    host! @provider.admin_domain
+
+    login_provider @provider
+  end
+
+  def test_status_404
+    get admin_api_objects_status_path(object_type: 'service', object_id: 1)
```
Martouta: Is this a service that does not exist or that belongs to another tenant? I think that we should test both. And I also think that it is risky to test it with a hardcoded ID number because it may vary in the tests (specially in Oracle)

test/integration/admin/api/objects_controller_test.rb
```diff
@@ -0,0 +1,80 @@
...
+    assert_response 404
+  end
+
+  def test_status_200
+    service = FactoryBot.create(:service, account: @provider)
```
Martouta: Just for efficiency:
```suggestion
    service = @provider.default_service
```

Because the factory `provider_account` already creates a very complete service so we don't need to create another one for this scenario :smile:

config/routes.rb
```diff
@@ -487,6 +487,8 @@ def call(env)
       # called from heroku deploy hook after heroku proxy deploy script
       post 'heroku-proxy/deployed', to: 'heroku_proxy#deployed', as: :heroku_proxy_deployed
 
+      get 'objects/status' => 'objects#status', as: :objects_status, controller: :objects
```
Martouta: We should accept only JSON for new endpoints :smile:

test/integration/admin/api/objects_controller_test.rb
```diff
@@ -0,0 +1,80 @@
...
+    CONTROLLER.any_instance.expects(:accessible_services).returns(Service.none)
+    get admin_api_objects_status_path(object_type: 'proxy', object_id: service.proxy.id)
+    assert_response 403
+
+    CONTROLLER.any_instance.expects(:accessible_services).returns(Service.all)
+    get admin_api_objects_status_path(object_type: 'proxy', object_id: service.proxy.id)
+    assert_response 200
+  end
+
+  def test_status_backend_api
```
Martouta: We lack the scenario of a service with state 'deleted' but not yet completely deleted.

test/integration/admin/api/objects_controller_test.rb
```diff
@@ -0,0 +1,80 @@
...
+    CONTROLLER.any_instance.expects(:accessible_services).returns(Service.all)
+    get admin_api_objects_status_path(object_type: 'service', object_id: service.id)
+    assert_response 200
+
+    service.destroy!
+    get admin_api_objects_status_path(object_type: 'service', object_id: service.id)
+    assert_response 404
+  end
+
+  def test_status_buyer_account
```
Martouta: In our current implementation in porta, buyer users can only be either destroyed for real or not, but we don't destroy them in background yet. This is does not mean that this endpoint for buyers is bad, it is actually good, but what I mean is that we lack the endpoint for tenants :thinking: and it is the hardest one because when we mark a tenant as scheduled for deletion, we disable its domain, so we should be able to enable the domain only for this endpoint or something like that :thinking:

app/controllers/admin/api/objects_controller.rb
```diff
@@ -0,0 +1,95 @@
...
+    end
+  end
+
+  class ProxyObject < ServiceObject
+    delegate :service_id, to: :status_object
+  end
+
+  class BackendApiObject < AllowedStatusObject
+    def authorize
+      controller.provider_can_use!(:api_as_product)
```
Martouta: We have a cancan ability for this which is what we are using in the BackendApisController (both in UI and API), so maybe we should do:
```suggestion
      controller.authorize! :read, status_object
```

test/integration/admin/api/objects_controller_test.rb
macejmic: hardcoded ID, true true.

app/lib/logic/rolling_updates.rb
```diff
@@ -359,9 +359,7 @@ module Controller
         helper_method :provider_can_use?
         private :provider_can_use?
       end
-
-      private
```
Martouta: If you do https://github.com/3scale/porta/pull/1258/files#r330067462 , you won't need to remove this :smile:

test/integration/admin/api/objects_controller_test.rb
macejmic: sure. but we should assert: `assert_response 200`

Martouta: ah right, sorry :+1: :see_no_evil:

app/controllers/admin/api/objects_controller.rb
```diff
@@ -0,0 +1,95 @@
+# frozen_string_literal: true
+
+class Admin::Api::ObjectsController < Admin::Api::BaseController
+
+  class AllowedStatusObject
```
guicassolato: This is an "authorizer", right? How about, for example:

```suggestion
  class Authorizer
```

...and of course also name the children classes `<ObjectType>Authorizer`.

It could be more literal too of course, like `StatusQueryAuthorizer`, if you prefer.

app/controllers/admin/api/objects_controller.rb
```diff
@@ -0,0 +1,95 @@
...
+    attr_reader :controller, :status_object
```
guicassolato:
```suggestion
    attr_reader :controller, :object
```

app/controllers/admin/api/objects_controller.rb
```diff
@@ -0,0 +1,95 @@
...
+
+    private
+
+    def raise_access_denied
+      raise CanCan::AccessDenied
+    end
+  end
+
+  class BuyerAccountObject < AllowedStatusObject
+    def authorize
```
guicassolato: As this will raise an error...

```suggestion
    def authorize!
```

app/controllers/admin/api/objects_controller.rb
macejmic: true true.

Martouta:
Closing in favor of https://github.com/3scale/porta/pull/1287
I won't remove the branch until we tag. We can reopen if we need to.

PR #1259, opened 2019-09-27
Cannot destroy a BackendApi used by Services

Martouta:
Closes [THREESCALE-3571](https://issues.jboss.org/browse/THREESCALE-3571)

- [X] Cannot destroy a BackendApi with BackendApiConfigs (except for destroy by association) is validated from the model and automatically applied to all our controllers and deletion workers. This includes the same condition for changing the state to 'deleted' and to destroy for real.
- [X] BackendApiConfig.accessible does not have to check anymore the state of its backend apis because all of its backend apis will be accessible now.
- [X] BackendApiRepresenter returns its usage.

image

image

- [X] When tries to destroy via the controller of the UI (skipping the UI itself), it shows an accurate and understandable error message.
image

### DIFF
--- a/app/controllers/provider/admin/backend_apis_controller.rb
+++ b/app/controllers/provider/admin/backend_apis_controller.rb
@@ -37,7 +37,7 @@ class Provider::Admin::BackendApisController < Provider::Admin::BaseController
   end
 
   def destroy
-    if @backend_api.services.empty? && @backend_api.mark_as_deleted
+    if @backend_api.mark_as_deleted
       redirect_to provider_admin_dashboard_path, notice: 'Backend will be deleted shortly.'
     else
       flash[:error] = 'Backend could not be deleted'
--- a/app/controllers/provider/admin/backend_apis_controller.rb
+++ b/app/controllers/provider/admin/backend_apis_controller.rb
@@ -40,7 +40,7 @@ class Provider::Admin::BackendApisController < Provider::Admin::BaseController
     if @backend_api.mark_as_deleted
       redirect_to provider_admin_dashboard_path, notice: 'Backend will be deleted shortly.'
     else
-      flash[:error] = 'Backend could not be deleted'
+      flash[:error] = @backend_api.errors.full_messages.to_sentence
       render :edit
     end
   end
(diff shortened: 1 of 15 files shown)


app/controllers/provider/admin/backend_apis_controller.rb
```diff
@@ -37,7 +37,7 @@ def update
   end
 
   def destroy
-    if @backend_api.services.empty? && @backend_api.mark_as_deleted
+    if @backend_api.mark_as_deleted
       redirect_to provider_admin_dashboard_path, notice: 'Backend will be deleted shortly.'
     else
       flash[:error] = 'Backend could not be deleted'
```
thomasmaas:
```suggestion
      flash[:error] = 'Backend could not be deleted because it's used in a Product'
```

Martouta: Imo that's not a good idea. Then we should show the flash containing he errors in the model, because that is not the only possible reason why it couldn't be deleted.

Martouta: I mean like: `flash[:error] = @backend_api.errors.full_messages.to_sentence`

config/locales/en.yml
```diff
@@ -1247,7 +1249,7 @@ en:
           Enables you to block any direct developer requests to your API backend; each 3scale API gateway call to your API backend contains a request header called <code>X-3scale-proxy-secret-token</code>.
           The value of this header can be set by you here. It's up to you ensure your backend only allows calls with this secret header.
         jwt_claim_with_client_id_type: |
-          Process the ClientID Token Claim value as a string or as a liquid template. When set to 'Liquid' you can define more complex rules. e.g. If 'some_claim' is an array you can select the first value this like {{ some_claim | first }}.
```
Martouta: Automatically fixed by atom :smile:

app/views/provider/admin/backend_apis/forms/_delete.html.slim
```diff
@@ -28,4 +28,3 @@
       - msg = t('api.backend_apis.edit.delete_confirmation', name: h(backend_api.name))
       - delete_options = { data: { confirm: msg }, method: :delete, label: "I understand the consequences, proceed to delete '#{h(backend_api.name)}' backend" }
       = delete_link_for(provider_admin_backend_api_path(backend_api), delete_options)
-
```
Martouta: Automatically fixed by atom :smile:

PR #1260, opened 2019-09-27
Promote to Staging button

guicassolato:
1. We will be able to hide the button once https://github.com/3scale/porta/pull/1244 gets merged. (Ideally also https://github.com/3scale/porta/pull/1235.)
2. We also need to solve a previous step when it's the first staging config to be generated. Right now we are pointing the user to the integration edit page where the "Save & test" is broken, apart from other fields that should not appear for APIAP-accounts.
3. We can easily improve the flash message shown after the config is promoted in https://github.com/3scale/porta/blob/95cf0c394bb8ee73131184080361e2be03d8f8a3/app/controllers/api/integrations_controller.rb#L33

### DIFF
--- a/app/views/api/integrations/apicast/configuration_driven/_apicast.html.slim
+++ b/app/views/api/integrations/apicast/configuration_driven/_apicast.html.slim
@@ -34,6 +34,13 @@ section.Configuration class=('Configuration--is-promoted' unless deployment_opti
         dt.u-dl-term Secret Token
         dd.u-dl-definition = @proxy.secret_token
 
+    div
+      = semantic_form_for @proxy, url: admin_service_integration_path(@service) do |f|
+        = f.buttons class: "buttons buttons-inline" do
+          = f.hidden_field :lock_version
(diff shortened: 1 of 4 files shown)


app/views/api/integrations/apicast/configuration_driven/_apicast.html.slim
```diff
@@ -34,6 +34,13 @@ section.Configuration class=('Configuration--is-promoted' unless deployment_opti
         dt.u-dl-term Secret Token
         dd.u-dl-definition = @proxy.secret_token
 
+    div
+      = semantic_form_for @proxy, url: admin_service_integration_path(@service) do |f|
+        = f.buttons class: "buttons buttons-inline" do
+          = f.hidden_field :lock_version
+          = f.button "Promote v. #{@show_presenter.last_sandbox_config.version+1} to Staging",
```
hallelujah: How does this work with Service Mesh?

Normally we show only one button for it, there is no deploy to Staging

PR #1263, opened 2019-09-30
Calling has_out_of_date_configuration? from vertical nav

damianpm:
In https://github.com/3scale/porta/pull/1235 the exclamation mark was introduced in the dashboard, but it was missing in the vertical navigation

**With JS on:**

> js-on

**With JS off:**

> js-off

### DIFF
--- a/app/helpers/vertical_nav_helper.rb
+++ b/app/helpers/vertical_nav_helper.rb
@@ -199,7 +199,7 @@ module VerticalNavHelper
 
     if can? :manage, :plans
       sections << {id: :ActiveDocs,  title: 'ActiveDocs',  path: admin_service_api_docs_path(@service)}
-      sections << {id: :integration, title: 'Integration', items: service_integration_items}
+      sections << {id: :integration, title: 'Integration', items: service_integration_items, outOfDateConfig: has_out_of_date_configuration?(@service)}
     end
 
     sections
--- a/app/views/layouts/provider.html.slim
+++ b/app/views/layouts/provider.html.slim
@@ -16,7 +16,7 @@ html[lang="en"]
   body
     #wrapper.pf-c-page
       = render partial: '/shared/provider/header'
-      = render partial: 'shared/provider/navigation/vertical_nav', :locals => {service: @service} unless active_menu == :dashboard
+      = render partial: 'shared/provider/navigation/vertical_nav' unless active_menu == :dashboard
       #border.pf-c-page__main class="self_clear" role="main"
         = render 'shared/provider/flash', flash: flash
         main.Main#content class=('Main--callToAction' if call_to_action?)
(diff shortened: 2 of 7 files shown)


app/views/layouts/provider.html.slim
```diff
@@ -16,7 +16,7 @@ html[lang="en"]
   body
     #wrapper.pf-c-page
       = render partial: '/shared/provider/header'
-      = render partial: 'shared/provider/navigation/vertical_nav' unless active_menu == :dashboard
+      = render partial: 'shared/provider/navigation/vertical_nav', :locals => {service: @service} unless active_menu == :dashboard
```
guicassolato: In some views we may not have the `@service`.

guicassolato: Let's fix in:
https://github.com/3scale/porta/blob/faeb8446346471aae12f4aa5502e36a4605d8c56/app/helpers/application_helper.rb#L369-L371

Make it:
```ruby
def has_out_of_date_configuration?(service)
  return unless service
  service.pending_affecting_changes?
end
```

app/views/shared/provider/navigation/_vertical_nav.html.slim
```diff
@@ -3,11 +3,12 @@
 - sections = vertical_nav_sections || []
 - active_section = active_submenu
 - active_item = active_sidebar
+- out_of_date_config = service ? has_out_of_date_configuration?(service) : false
```
josemigallas: Since this is only for `Integration` I think it would be better to add it as a field in `Section` and set it in `vertical_nav_helper` directly.

app/views/shared/provider/navigation/_vertical_nav_item.html.slim
```diff
@@ -12,7 +13,7 @@
   li.pf-c-nav__item class='pf-m-expandable pf-m-expanded' class=('pf-m-current' if is_section_active)
     a.pf-c-nav__link href='#!'
       = title
-      i.fa.fa-exclamation-triangle
+      i class=('fa fa-exclamation-triangle' if integration_out_of_date)
```
josemigallas: This is adding an `<i>` tag always. It should be:
```suggestion
      - if integration_out_of_date
        i.fa.fa-exclamation-triangle
```

app/helpers/vertical_nav_helper.rb
```diff
@@ -199,7 +199,7 @@ def service_nav_sections
 
     if can? :manage, :plans
       sections << {id: :ActiveDocs,  title: 'ActiveDocs',  path: admin_service_api_docs_path(@service)}
-      sections << {id: :integration, title: 'Integration', items: service_integration_items}
+      sections << {id: :integration, title: 'Integration', items: service_integration_items, outOfDateConfig: has_out_of_date_configuration?(@service)}
```
josemigallas:
```suggestion
      sections << {id: :integration, title: 'Integration', items: service_integration_items, out_of_data_config: has_out_of_date_configuration?(@service)}
```

damianpm: No, is not automagic

PR #1266, opened 2019-09-30
ProxyConfigAffectingChange touches the proxy

guicassolato:
Closes [THREESCALE-3570](https://issues.jboss.org/browse/THREESCALE-3570)


Martouta:
Makes sense so far :+1: And then it will have to be removed from other models :smile: because it is not needed anymore from there :smile:

PR #1267, opened 2019-09-30
Scope ProxyRule's acts_as_list by owner

guicassolato:
Closes [THREESCALE-3569](https://issues.jboss.org/browse/THREESCALE-3569)

### DIFF
--- a/app/models/proxy_rule.rb
+++ b/app/models/proxy_rule.rb
@@ -6,7 +6,7 @@ class ProxyRule < ApplicationRecord
   acts_as_list scope: %i[owner_id owner_type], add_new_at: :bottom
   scope :ordered, -> { order(position: :asc) }
 
-  belongs_to :proxy, touch: true
+  belongs_to :proxy
   belongs_to :owner, polymorphic: true # FIXME: we should touch the owner here, but it will raise ActiveRecord::StaleObjectError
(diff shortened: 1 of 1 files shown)


app/models/proxy_rule.rb
```diff
@@ -3,7 +3,7 @@
 require 'addressable/template'
 
 class ProxyRule < ApplicationRecord
-  acts_as_list scope: :proxy, add_new_at: :bottom
+  acts_as_list scope: %i[owner_id owner_type], add_new_at: :bottom
   scope :ordered, -> { order(position: :asc) }
 
   belongs_to :proxy, touch: true
```
Martouta: I wonder if it should be: :thinking: 
```suggestion
  belongs_to :owner, touch: true
```

guicassolato: I don't think it's supported by thinking_sphinx. We have it somewhere else too: https://github.com/3scale/porta/blob/ab2b46785b741acc7714112aeea8eb6a2954cd61/app/models/plan.rb#L18

Martouta:
The problem with the tests is that with the `touch: true` it updates the proxy and then when it tries to destroy the proxy, it cannot do it because the instance in memory is not the same as the one updated (with the updated_at)

PR #1268, opened 2019-09-30
APIAP only works with APIcast V2

hallelujah:
Hide and unauthorize policies or backends when service can't use them


thomasmaas:
This is only relevant for SaaS so no hurries

PR #1272, opened 2019-09-30
BackendApiConfig as an affecting object of Proxy config

guicassolato:
Currently `BackendApiConfig` object are not signalling a proxy config affecting change. They should. So we show the exclamation mark to the user whenever a backend is added to a product, edited ot removed.

### DIFF
--- a/app/lib/proxy_config_affecting_changes.rb
+++ b/app/lib/proxy_config_affecting_changes.rb
@@ -26,6 +26,9 @@ module ProxyConfigAffectingChanges
     PROXY_CONFIG_AFFECTING_ATTRIBUTES = %w[policies_config].freeze # TODO: add more attributes here
 
     included do
+      has_one :proxy_config_affecting_change, dependent: :delete
+      private :proxy_config_affecting_change
+
       include ProxyConfigAffectingChanges
 
       after_commit :issue_proxy_affecting_change_events, on: :update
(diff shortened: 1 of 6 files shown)


app/lib/proxy_config_affecting_changes.rb
```diff
@@ -35,6 +38,33 @@ def issue_proxy_affecting_change_events
...
+  module BackendApiConfigExtension
+    extend ActiveSupport::Concern
+
+    included do
+      include ProxyConfigAffectingChanges
+
+      after_commit :issue_proxy_affecting_change_events
+
+      def issue_proxy_affecting_change_events
+        backend_api.proxies.each(&method(:issue_proxy_affecting_change_event))
```
Martouta:
```suggestion
        backend_api.proxies.find_each(&method(:issue_proxy_affecting_change_event))
```
So if there are many, I feel a bit safer this way than loading them all to memory at once :thinking:

app/lib/proxy_config_affecting_changes.rb
```diff
@@ -10,11 +10,13 @@ module ProxyRuleExtension
       after_commit :issue_proxy_affecting_change_events
 
       def issue_proxy_affecting_change_events
+        return unless owner
```
Martouta: There is no harm with this :+1: but in which case this is possible? :thinking:

guicassolato: If it has been deleted but the ProxyRule was still pending to execute this callback.

PR #1273, opened 2019-09-30
[LOW PRIORITY][Fix] Escaped HTML entities in edit API page

Martouta:
Closes [THREESCALE-2547](https://issues.jboss.org/browse/THREESCALE-2547)

image

Not sure if it is a good idea but I am escaping JSON instead of HTML to obtain this. For this:
https://github.com/rails/rails/blob/a273da7619ac6a2b2f98532a5610238c68ad219b/activesupport/lib/active_support/core_ext/string/output_safety.rb#L10-L11

### DIFF
--- a/app/helpers/api/services_helper.rb
+++ b/app/helpers/api/services_helper.rb
@@ -46,7 +46,7 @@ module Api::ServicesHelper
   end
 
   def delete_service_link(service, options = {})
-    msg = t('api.services.forms.definition_settings.delete_confirmation', name: h(service.name))
+    msg = t('api.services.forms.definition_settings.delete_confirmation', name: j(service.name))
     delete_link_for(admin_service_path(service), {data: { confirm: msg }, method: :delete}.merge(options) )
   end
 
--- a/app/helpers/cinstances_helper.rb
+++ b/app/helpers/cinstances_helper.rb
@@ -53,7 +53,7 @@ Are you sure?}
   end
 
   def delete_application_link(application)
-    msg = t('api.applications.edit.delete_confirmation', name: h(application.name))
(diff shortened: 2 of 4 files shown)


app/helpers/api/services_helper.rb
```diff
@@ -46,7 +46,7 @@ def friendly_service_setting service, setting
   end
 
   def delete_service_link(service, options = {})
-    msg = t('api.services.forms.definition_settings.delete_confirmation', name: h(service.name))
+    msg = t('api.services.forms.definition_settings.delete_confirmation', name: j(service.name))
     delete_link_for(admin_service_path(service), {data: { confirm: msg }, method: :delete}.merge(options) )
```
hallelujah: Leave above as is and  escaping the whole message `msg` here?

Martouta: Right, better! :+1:

Martouta: Waaait, maybe there's no need :thinking: because the rest of the message is what we write ourselves in the locales, so we may wanna add some HTML there on purpose and it is a trustable source :thinking:

hallelujah: Yes but we need to escape the name anyway

Martouta: @hallelujah , sorry but I am not understanding you right now :thinking:

This commit right now escapes the name only which is what is introduced from the user from outside, then you suggested to escape the whole message (`msg`), which contains the message written by ourselves in the locales + the name (in this commit, the name is escaped but not the rest of the message). So I said that maybe we shouldn't escape the whole message because the rest of the message is what we wrote ourselves in the locales so it is trustable and we also might include some HTML there intentionally.

PR #1274, opened 2019-09-30
always show apicast config for apiap

thomasmaas:
Closes [THREESCALE-3575](https://issues.jboss.org/browse/THREESCALE-3575)

People on APIAP should always get the same config show  page.

This will lead to issues with people currently using plugins. To be continued. Hoping the tests will fail.


guicassolato:
I think we can close this. It will be covered by https://github.com/3scale/porta/pull/1281.

PR #1275, opened 2019-10-01
Fix bug of promoting to staging when there is no sandbox config

hallelujah:


guicassolato:
I think we can close this. It's covered by https://github.com/3scale/porta/pull/1281.

Martouta:
Closing with Hery's permission and moving the issue to Gui's PR :smile:

PR #1276, opened 2019-10-01
Fixes update buttons in new Policies page

josemigallas:

Fixes broken logic regarding update buttons in Policies:
* Move button into a fieldset
* Changes query from `input#id` to simply the id.

**Verification steps** 

update

### DIFF
--- a/app/javascript/src/Policies/components/PoliciesWidget.jsx
+++ b/app/javascript/src/Policies/components/PoliciesWidget.jsx
@@ -53,7 +53,7 @@ const PolicyList = ({ registry, chain, originalChain, policyConfig, ui, boundAct
     closePolicyRegistry: boundActionCreators.closePolicyRegistry
   }
 
-  const buttonsFieldset = document.querySelector('fieldset.buttons')
+  const buttonsFieldset = document.querySelector('[id^="edit_proxy_"] > fieldset.buttons')
   if (buttonsFieldset) {
     // classList.toggle second argument is not supported in IE11
     if (ui.chain) {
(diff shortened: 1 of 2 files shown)


app/javascript/src/Policies/components/PoliciesWidget.jsx
```diff
@@ -53,7 +53,7 @@ const PolicyList = ({ registry, chain, originalChain, policyConfig, ui, boundAct
     closePolicyRegistry: boundActionCreators.closePolicyRegistry
   }
 
-  const buttonsFieldset = document.querySelector('fieldset.buttons')
+  const buttonsFieldset = document.querySelector('#edit_proxy_2 > fieldset.buttons')
```
Martouta: Why #edit_proxy_2 ? What is this '2'? :thinking:

josemigallas: It's the ID of the form, I guess that comes from `semantic_form_for()`.

josemigallas: In APIAP, the `fieldset.buttons` is a direct child of `#edit_proxy_2`. In non-APIAP it is not. This is the way I found to affect only APIAP.

thomasmaas: But wouldn't `#edit_proxy_2` only be valid of a proxy with id=2 and couldn't it be any number?

josemigallas: OK I didn't know that

thomasmaas: I'm just guessing…

thomasmaas: But classes would be saver than ids

josemigallas: Done

thomasmaas: When policies are in integration edit page they sit inside a container with id `#integration-tabs` (i know, also an id). Maybe you can use that?

PR #1277, opened 2019-10-01
[APIAP] Update Proxy public/staging urls from Integration -> Settings

Martouta:
Closes [THREESCALE-3576](https://issues.jboss.org/browse/THREESCALE-3576)

### Verification Steps
1. In `config/rolling_updates.yml` you will need `api_as_product: true`
2. In `config/settings.yml` you will need `apicast_custom_url: http://example.com.lvh.me` for this:
https://github.com/3scale/porta/blob/17d259fa26abe0bbbfc710c869cc3bf7c3a26470/app/helpers/api/integrations_helper.rb#L71-L75
3. Run the server with `UNICORN_WORKERS=8 bundle exec rails server -b 0.0.0.0`
4. Through the browser, go to `http://provider-admin.example.com.lvh.me:3000/apiconfig/services/:service_id/settings`
image
5. Change that data and update :dancer:

### DIFF
--- a/app/controllers/api/services_controller.rb
+++ b/app/controllers/api/services_controller.rb
@@ -43,7 +43,7 @@ class Api::ServicesController < Api::BaseController
     @service = collection.new # this is done in 2 steps so that the account_id is in place as preffix_key relies on it
     creator = ServiceCreator.new(service: @service)
 
-    if can_create? && creator.call(service_params)
+    if can_create? && creator.call(create_params)
       flash[:notice] =  'Service created.'
       onboarding.bubble_update('api')
       redirect_to admin_service_path(@service)
@@ -55,7 +55,7 @@ class Api::ServicesController < Api::BaseController
   end
 
   def update
-    if @service.update_attributes(service_params)
+    if integration_settings_result.update(service_attributes: service_params, proxy_attributes: proxy_params)
       flash[:notice] =  'Service information updated.'
       onboarding.bubble_update('api') if service_name_changed?
       onboarding.bubble_update('deployment') if integration_method_changed? && !integration_method_self_managed?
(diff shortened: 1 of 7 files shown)


app/controllers/api/services_controller.rb
```diff
@@ -75,24 +75,43 @@ def destroy
 
   private
 
+  attr_reader :service
+
+  def integration_settings_result
+    ApiIntegration::SettingsResult.new(service: service, proxy: service.proxy)
+  end
+
+  def create_params
```
Martouta: I know that this is very bad but I encountered conflicts with this during this same morning :scream: And I don't have time to fix it properly for today.

hallelujah: Ah you mean repeating it :)

Yes you can DRY with `user_attribute_names` later

thomasmaas:
@Martouta it allows you to change the url and then it keeps that url when you switch to 3scale managed apicast. That's not the correct behaviour.

Martouta:
Right, @thomasmaas , good catch! :+1: Fixed now :+1: :muscle:

app/controllers/api/services_controller.rb
Martouta: In fact it should be done properly in [THREESCALE-3363](https://issues.jboss.org/browse/THREESCALE-3363), but that is after 2.7

app/views/api/integrations/apicast/shared/_proxy_endpoints.html.slim
```diff
@@ -1,4 +1,5 @@
-= f.inputs "API gateway" do
+- is_service_mesh = proxy.service_mesh_integration?
+= f.inputs "API gateway", class: "apicast-only-settings #{'hidden' if is_service_mesh}", disabled: is_service_mesh do
```
Martouta: I've followed @didierofrivia 's work:
https://github.com/3scale/porta/blob/4f3b989575b030b601c29012fa1ae30d43c8e7b2/app/views/api/services/forms/_authentication_settings.html.slim#L20
https://github.com/3scale/porta/blob/4f3b989575b030b601c29012fa1ae30d43c8e7b2/app/views/api/services/forms/_integration_settings_apiap.html.slim#L1
image
https://github.com/3scale/porta/blob/4f3b989575b030b601c29012fa1ae30d43c8e7b2/app/javascript/packs/settings.js#L4
https://github.com/3scale/porta/blob/4f3b989575b030b601c29012fa1ae30d43c8e7b2/app/javascript/src/Settings/authentication-widget.jsx#L3-L38

app/lib/api_integration/settings_result.rb
```diff
@@ -0,0 +1,49 @@
...
+    errors
+  end
+
+  private
+
+  # ActiveModel::Errors needs this method to read the errors correctly
+  def read_attribute_for_validation(attr)
+    public_send(attr)
+  end
+end
```
duduribeiro: why to put this in `lib` ? 🤔 

I think it is better to put it in `app/models` or `app/services`.. what do you think?

PR #1278, opened 2019-10-01
Improve error message when the apicast registry url is not present + 'Start by'

Martouta:
Before this PR:
image
image

After this PR:
image
image


Martouta:
I'm not sure @thomasmaas but I think so because it is a config that the sysadmins of on-prem can change (and it can happen accidentally). Anyway, there is no harm in a clear error message :smile:

PR #1281, opened 2019-10-01
Proxy config workflow

guicassolato:

It implements the new workflow for generating/promoting proxy configs on APIAP, mostly changing the UI.

_Non-APIAP_

non-apiap

_APIAP_

apiap

Closes
- [THREESCALE-3581](https://issues.jboss.org/browse/THREESCALE-3581)
- [THREESCALE-3575](https://issues.jboss.org/browse/THREESCALE-3575)
- [THREESCALE-3539](https://issues.jboss.org/browse/THREESCALE-3539)

**TODOs**
- [x] Fix promote buttons for APIAP vs non-APIAP
- [x] Fix promote buttons for service mesh integration (See https://github.com/3scale/porta/pull/1274)
- [x] Enable more attributes of `Proxy` to trigger `AffectingObjectChangesEvent`
- [x] Fix "Integrate this API" -> "Integrate this Product" in the Dashboard
- [x] Add "Create a backend" link in "Add Backend to Product" page (create BackendApiConfig)
- [ ] Write 🥒 for the new scenarios
- [ ] ~Make it impossible to generate proxy config if product has no backend api config~ (separate PR)
- [ ] ~Throw 404 when apiap-user tries to access `/apiconfig/services/2/integration/edit`~ (separate PR)

### DIFF
--- a/app/helpers/api/integrations_helper.rb
+++ b/app/helpers/api/integrations_helper.rb
@@ -92,4 +92,30 @@ module Api::IntegrationsHelper
     title = deployment_option_is_service_mesh?(service) ? 'Service Mesh' : 'APIcast'
     t(:edit_deployment_configuration, scope: :api_integrations_controller, deployment: title )
   end
+
+  def promote_to_staging_button_options(proxy)
+    return disabled_promote_button_options if proxy.any_sandbox_configs? && !proxy.pending_affecting_changes?
+
+    label = deployment_option_is_service_mesh?(proxy.service) ? 'Update Configuration' : "Promote v. #{proxy.next_sandbox_config_version} to Staging"
+    promote_button_options(label)
+  end
+
+  def promote_to_production_button_options(proxy)
+    return disabled_promote_button_options if proxy.environments_have_same_config?
+
+    label = "Promote v. #{proxy.next_production_config_version} to Production"
+    promote_button_options(label)
+  end
+
+  PROMOTE_BUTTON_COMMON_OPTIONS = { button_html: { class: 'PromoteButton', data: { disable_with: 'promoting…' } } }.freeze
+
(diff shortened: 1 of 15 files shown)


app/helpers/api/integrations_helper.rb
```diff
@@ -92,4 +92,30 @@ def edit_deployment_option_title(service)
     title = deployment_option_is_service_mesh?(service) ? 'Service Mesh' : 'APIcast'
     t(:edit_deployment_configuration, scope: :api_integrations_controller, deployment: title )
   end
+
+  def promote_to_staging_button_options(proxy)
+    classes = ['PromoteButton']
+    disabled = proxy.any_sandbox_configs? && !proxy.pending_affecting_changes?
+    if disabled
```
josemigallas: Suggestion: extract each button (disabled, important) to its own method and
```
if (proxy.any_sandbox_configs? && ...)
  disabled_button
...

File: app/presenters/api/integrations_show_presenter.rb [hunk 2]
```diff
@@ -42,6 +53,18 @@ def environments_have_same_config?
     end
   end
 
+  def apiap?
+    provider_can_use? :api_as_product
+  end
+
+  def apicast_config_ready?
+    if apiap?
```
josemigallas: Question: Is this check redundant? Wouldn't `.backend_apis.any?` return `nil` always if false?

guicassolato: Not necessarily. `backend_apis.any?` may be truthy and yet `apiap? => false`.

Martouta: In fact we currently create a backend api for each service in saas and we have apiap false there :smile:

PR #1284, opened 2019-10-02
[settings] Disabling auth options when service mesh and not oidc

didierofrivia:

Fixes a bug when Service Mesh integration was saved and auth options were displayed.

Fixes https://issues.jboss.org/browse/THREESCALE-3591

**Verification steps** 
APIAP and Istio should be enabled in rolling updates.

Visit Integration -> Settings
When picking Istio integration and auth key, hit `update service` then authentication settings should not be available. 

Testing will be done when replaced by react component.

### DIFF
--- a/app/views/api/services/forms/_authentication_settings.html.slim
+++ b/app/views/api/services/forms/_authentication_settings.html.slim
@@ -1,18 +1,18 @@
 - # oidc
-- is_oidc = @service.backend_version === 'oidc'
-- is_service_mesh = @service.proxy.deployment_option === 'service_mesh_istio'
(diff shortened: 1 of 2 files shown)


app/views/api/services/forms/_authentication_settings.html.slim
```diff
@@ -1,18 +1,18 @@
 - # oidc
 - is_oidc = @service.backend_version === 'oidc'
 - is_service_mesh = @service.proxy.deployment_option === 'service_mesh_istio'
```
hallelujah:
```suggestion
- is_service_mesh = @service.proxy.service_mesh_integration?
```

PR #1287, opened 2019-10-02
[api#ednpoint] objects status endpoint

macejmic:

API endpoint: object deletion status for objects that are deleted asynchronously

https://issues.jboss.org/browse/THREESCALE-3265

### DIFF
--- a/app/controllers/admin/api/objects_controller.rb
+++ b/app/controllers/admin/api/objects_controller.rb
@@ -1,0 +1,60 @@
+# frozen_string_literal: true
+
+class Admin::Api::ObjectsController < Admin::Api::BaseController
+  clear_respond_to
+  respond_to :json
+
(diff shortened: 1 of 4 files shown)


app/controllers/admin/api/objects_controller.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+class Admin::Api::ObjectsController < Admin::Api::BaseController
+  clear_respond_to
+  respond_to :json
+
+  ALLOWED_OBJECTS = %w[service buyer_account proxy backend_api].freeze
```
hallelujah: Does BuyerAccount exist?

macejmic: right.. there should be `account`

PR #1290, opened 2019-10-02
CircleCI with Ruby 2.4

Martouta:
Fixes [THREESCALE-3589 - NoMethodError: undefined method `match?' for "\"4\"":String](https://issues.jboss.org/browse/THREESCALE-3589)


hallelujah:
We can just change also system-builder latest to be ruby-2.4
But LGTM

PR #1280, opened 2019-10-01
[WIP][Bugfix] Create a Backend API for each new tenant

Martouta:
Closes [THREESCALE-3580](https://issues.jboss.org/browse/THREESCALE-3580)

PR #1292, opened 2019-10-02
Create first backend api of every tenant

guicassolato:
Closes [THREESCALE-3580](https://issues.jboss.org/browse/THREESCALE-3580)

Alternative to https://github.com/3scale/porta/pull/1280.

### DIFF
--- a/test/unit/db/seeds_test.rb
+++ b/test/unit/db/seeds_test.rb
@@ -114,6 +114,7 @@ class SeedsTest < ActiveSupport::TestCase
     assert tenant_account.approved?
     assert tenant_account.state_changed_at.present?
     assert_equal ApplicationPlan.find_by(name: 'enterprise').id, tenant_account.bought_cinstance&.plan.id
+    assert_equal 'API', tenant_account.default_service&.name
 
     assert_equal 2, tenant_account.users.count
     tenant_user = tenant_account.users.but_impersonation_admin.first!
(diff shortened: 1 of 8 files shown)


test/unit/db/seeds_test.rb
```diff
@@ -114,6 +114,7 @@ def teardown
     assert tenant_account.approved?
     assert tenant_account.state_changed_at.present?
     assert_equal ApplicationPlan.find_by(name: 'enterprise').id, tenant_account.bought_cinstance&.plan.id
+    assert_equal 'API', tenant_account.default_service&.name
```
hallelujah: This is weird ...
Why do we check for nil in a test?

Martouta: just to don't raise a weird error if it is not there. It is a clearer message imo. Anyway, small improvements tomorrow instead of today plz :joy:

test/unit/signup/account_manager_test.rb
```diff
@@ -133,11 +133,12 @@ class ProviderAccountManagerTest < Signup::AccountManagerTest
       assert_equal BackendApi.default_api_backend, account.default_service.api_backend
     end
 
-    test 'creating a provider will not create default Backend if api_as_product is enabled' do
+    test 'creating a provider will create default Backend if api_as_product is enabled' do
       Account.any_instance.stubs(:provider_can_use?).returns(false)
       Account.any_instance.stubs(:provider_can_use?).with(:api_as_product).returns(true)
       account = signup_account_manager.create(signup_params).account
-      assert_nil account.default_service.api_backend
+      assert account.backend_apis.first
```
hallelujah: I would also test that there is only one even if I create another service

test/unit/db/seeds_test.rb
hallelujah: But why in a test????
So the test will fail anyway, `'API' == nil` or error raising makes the test fail... but does not have the same meaning.

When reading it, one can get confused that it could be absent and thus "fix" it with `assert_nil`

Martouta: Then we could do:
```ruby
assert (service = tenant_account.default_service)
assert_equal 'API', service.name
```

PR #1293, opened 2019-10-03
Fix Oracle Tests for the UI Controller of BackendApiConfigs

Martouta:
Fixes [these tests](https://circleci.com/gh/3scale/porta/113726#tests/containers/5).

### DIFF
--- a/test/integration/api/backend_api_configs_controller_test.rb
+++ b/test/integration/api/backend_api_configs_controller_test.rb
@@ -17,10 +17,10 @@ class Api::BackendApiConfigsControllerTest < ActionDispatch::IntegrationTest
   attr_reader :service, :backend_api
 
   test '#index' do
-    service.backend_api_configs.create(backend_api: backend_api)
+    service.backend_api_configs.create!(backend_api: backend_api, path: 'whatever')
     get admin_service_backend_api_configs_path(service)
     assert_response :success
-    assert_select 'table#backend_api_configs tbody tr', count: 2
+    assert_select 'table#backend_api_configs tbody tr', count: service.backend_apis.count
   end
 
   test '#new' do
@@ -33,7 +33,7 @@ class Api::BackendApiConfigsControllerTest < ActionDispatch::IntegrationTest
 
     backend_apis_not_in_use = backend_apis.take(2)
     backend_api_in_use = backend_apis.last
-    service.backend_api_configs.create(backend_api: backend_api_in_use)
+    service.backend_api_configs.create!(backend_api: backend_api_in_use, path: 'whatever')
 
     get new_admin_service_backend_api_config_path(service)
     backend_apis_not_in_use.each { |backend_api| assert_select 'select#backend_api_config_backend_api_id option[value=?]', backend_api.id }


test/integration/api/backend_api_configs_controller_test.rb
```diff
@@ -17,10 +17,10 @@ class Api::BackendApiConfigsControllerTest < ActionDispatch::IntegrationTest
   attr_reader :service, :backend_api
 
   test '#index' do
-    service.backend_api_configs.create(backend_api: backend_api)
+    service.backend_api_configs.create!(backend_api: backend_api, path: 'whatever')
     get admin_service_backend_api_configs_path(service)
     assert_response :success
-    assert_select 'table#backend_api_configs tbody tr', count: 2
+    assert_select 'table#backend_api_configs tbody tr', count: service.backend_apis.count
```
guicassolato: We should expect a fixed number of backend api configs listed here.

test/integration/api/backend_api_configs_controller_test.rb
```diff
@@ -17,10 +17,10 @@ class Api::BackendApiConfigsControllerTest < ActionDispatch::IntegrationTest
   attr_reader :service, :backend_api
 
   test '#index' do
-    service.backend_api_configs.create(backend_api: backend_api)
+    service.backend_api_configs.create!(backend_api: backend_api, path: 'whatever')
```
guicassolato: Why does this only affect Oracle?

Martouta: @guicassolato no idea :woman_shrugging: It could be that it is not database related but just randomly regardless of the DB.

test/integration/api/backend_api_configs_controller_test.rb
Martouta: @guicassolato I don't understand why :thinking: The goal is that it should show as many as backend apis we have for that service, no? independently of how many there are :thinking: 

But in that case i would suggest adding as the first line of the test: `service.backend_apis.destroy_all`

test/integration/api/backend_api_configs_controller_test.rb
Martouta: @guicassolato 
I am running tests in all DBs in other PRs and apparently it happens for Oracle tests always and only Oracle. But that is not necessarily related to the DB itself as I discovered for https://github.com/3scale/porta/pull/1205

hallelujah: Because empty String  is `nil` in oracle?
Can you investigate this clue?

Martouta: no @hallelujah , it is not related to that. The error message is `ActiveRecord::RecordNotUnique: OCIError: ORA-00001: unique constraint (RAILS.INDEX_BACKEND_API_CONFIGS_ON_PATH_AND_SERVICE_ID) violated` and it happens because in https://github.com/3scale/porta/pull/1292 we associate a backend api with empty path for every first service of a tenant, so now it tries to create a backend api with empty path again (empty path I mean '/') for the same service and it says that it can't.

test/integration/api/backend_api_configs_controller_test.rb
hallelujah: :+1: I think the `service.backend_apis.count` is more robust to any changes

Martouta: I agree with @hallelujah :+1:

guicassolato: I'm sorry, but this is incorrect. If the page shows more than 2 backend api configs, then we have a bug. (Specially if it's only for one DB.) Changing this is only hiding that bug.

Martouta: For anybody reading this: Gui, Hery and I are talking about it outside here right now :smile:

PR #1294, opened 2019-10-03
Improve the tests of creating a backend api for each new tenant

Martouta:
Improves tests of https://github.com/3scale/porta/pull/1292 according to the reviews.

### DIFF
--- a/test/unit/db/seeds_test.rb
+++ b/test/unit/db/seeds_test.rb
@@ -197,7 +197,7 @@ class SeedsTest < ActiveSupport::TestCase
     tenant_user = tenant_account.users.but_impersonation_admin.first!
     assert_equal ENV_VARIABLES['USER_LOGIN'], tenant_user.username
     assert_equal ENV_VARIABLES['USER_EMAIL'], tenant_user.email
-    assert(tenant_bought_cinstance = enant_account.bought_cinstance)
+    assert(tenant_bought_cinstance = tenant_account.bought_cinstance)
     assert_equal ApplicationPlan.find_by(name: ENV_VARIABLES['PROVIDER_PLAN']).id, tenant_bought_cinstance.plan_id
   end
 
--- a/test/unit/db/seeds_test.rb
+++ b/test/unit/db/seeds_test.rb
@@ -90,14 +90,14 @@ class SeedsTest < ActiveSupport::TestCase
       assert master_service.metrics.find_by(system_name: system_name, unit: 'hit', friendly_name: description)
     end
 
-    assert (tenant_app_plan = master_service.application_plans.default)
+    tenant_app_plan = master_service.application_plans.default
     assert_equal 'enterprise', tenant_app_plan.name
     assert_equal master_service.id, tenant_app_plan.issuer_id
 
-    assert (apicast = master_account.access_tokens.find_by(name: 'APIcast'))
+    apicast = master_account.access_tokens.find_by(name: 'APIcast')
     assert_equal 'ro', apicast.permission
     assert_equal %w[account_management], apicast.scopes
-    assert (master_token = master_account.access_tokens.find_by(name: 'Master Token'))
+    master_token = master_account.access_tokens.find_by(name: 'Master Token')
     assert_equal 'rw', master_token.permission
     assert_equal %w[account_management], master_token.scopes
 
@@ -208,9 +208,9 @@ class SeedsTest < ActiveSupport::TestCase
   private
 
   def assert_expected_backend_api
-    assert (service = Account.tenants.first!.default_service)
+    service = Account.tenants.first!.default_service
     assert_equal 1, service.backend_apis.count
-    assert (backend_api = service.backend_apis.accessible.first)
+    backend_api = service.backend_apis.accessible.first
     assert_equal BackendApi.default_api_backend, backend_api.private_endpoint
     assert_equal service.system_name, backend_api.system_name
     assert_equal "#{service.name} Backend", backend_api.name
--- a/test/unit/db/seeds_test.rb
+++ b/test/unit/db/seeds_test.rb
@@ -80,7 +80,8 @@ class SeedsTest < ActiveSupport::TestCase
     assert_equal master_service.id, master_app_plan.issuer_id
     assert_equal 'enterprise', master_app_plan.name
 
-    assert_equal ApplicationPlan.find_by(name: 'Master Plan').id, master_account.bought_cinstance&.plan.id
+    assert(master_bought_cinstance = master_account.bought_cinstance)
+    assert_equal ApplicationPlan.find_by(name: 'Master Plan').id, master_bought_cinstance.plan_id
 
     master_account_plan = master_account.default_account_plan
     assert_equal master_account.id, master_account_plan.issuer_id
@@ -113,8 +114,10 @@ class SeedsTest < ActiveSupport::TestCase
     assert tenant_account.sample_data.presence
     assert tenant_account.approved?
     assert tenant_account.state_changed_at.present?
-    assert_equal ApplicationPlan.find_by(name: 'enterprise').id, tenant_account.bought_cinstance&.plan.id
-    assert_equal 'API', tenant_account.default_service&.name
+    assert(tenant_bought_cinstance = tenant_account.bought_cinstance)
+    assert_equal ApplicationPlan.find_by(name: 'enterprise').id, tenant_bought_cinstance.plan_id
+    assert(tenant_service = tenant_account.default_service)
+    assert_equal 'API', tenant_service.name
 
     assert_equal 2, tenant_account.users.count
     tenant_user = tenant_account.users.but_impersonation_admin.first!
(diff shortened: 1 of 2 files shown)


test/unit/db/seeds_test.rb
```diff
@@ -80,7 +80,8 @@ def teardown
     assert_equal master_service.id, master_app_plan.issuer_id
     assert_equal 'enterprise', master_app_plan.name
 
-    assert_equal ApplicationPlan.find_by(name: 'Master Plan').id, master_account.bought_cinstance&.plan.id
+    assert(master_bought_cinstance = master_account.bought_cinstance)
```
guicassolato: Apparently we don't need `master_bought_cinstance` again in the same test. Why setting a variable now?

Martouta: @guicassolato it is just a possible way to solve what @hallelujah said in https://github.com/3scale/porta/pull/1292#discussion_r330738420 of the nil check in tests. But I am opened to other proposals :smile: :+1:

test/unit/signup/account_manager_test.rb
```diff
@@ -126,23 +126,29 @@ class ProviderAccountManagerTest < Signup::AccountManagerTest
       assert_equal 'API', account.first_service!.name
     end
 
-    test 'creating a provider will create default Backend if api_as_product is disabled' do
+    test 'creating a provider will create default Backend regardless of api_as_product' do
```
guicassolato: This is not "regardless APIAP". This is actually "without APIAP".

test/unit/signup/account_manager_test.rb
```diff
@@ -126,23 +126,29 @@ class ProviderAccountManagerTest < Signup::AccountManagerTest
...
       Account.any_instance.stubs(:provider_can_use?).returns(false)
       Account.any_instance.stubs(:provider_can_use?).with(:api_as_product).returns(false)
       account = signup_account_manager.create(signup_params).account
-      assert_equal BackendApi.default_api_backend, account.default_service.api_backend
+      assert_expected_backend_api(account)
```
guicassolato: The reason why it was kept the assertion of `account.default_service.api_backend` here is to abstract implementation details. For non-APIAP scenario. the seeds do not care about existing backend_apis.

```suggestion
      assert_equal BackendApi.default_api_backend, account.default_service.api_backend
```

test/unit/signup/account_manager_test.rb
```diff
@@ -126,23 +126,29 @@ class ProviderAccountManagerTest < Signup::AccountManagerTest
...
     end
 
     test 'creating a provider will create default Backend if api_as_product is enabled' do
       Account.any_instance.stubs(:provider_can_use?).returns(false)
       Account.any_instance.stubs(:provider_can_use?).with(:api_as_product).returns(true)
       account = signup_account_manager.create(signup_params).account
-      assert account.backend_apis.first
-      assert_equal BackendApi.default_api_backend, account.default_service.api_backend
+      assert_expected_backend_api(account)
```
guicassolato:
```suggestion
      assert (backend_api = account.backend_apis.first)
      assert_equal BackendApi.default_api_backend, backend_api.private_endpoint
```

test/unit/signup/account_manager_test.rb
Martouta: Right sorry, bcz I tried refactoring in one and then I came back to the previous version and forgot the change this :+1:

test/unit/signup/account_manager_test.rb
Martouta: @guicassolato I did not remove that assertion at all:
https://github.com/3scale/porta/pull/1294/files#diff-e76abd3a417934b609ad05f2a208fa52R149
The goal of putting it in a method was just to don't repeat all the same assertions for both tests.

test/unit/signup/account_manager_test.rb
```diff
@@ -126,23 +126,29 @@ class ProviderAccountManagerTest < Signup::AccountManagerTest
...
 
     private
 
+    def assert_expected_backend_api(account)
+      assert_equal 1, account.backend_apis.count
+      assert(service = account.default_service)
+      assert_equal 1, service.backend_apis.count
+      assert_equal BackendApi.default_api_backend, service.api_backend
+    end
+
```
guicassolato: This is not needed. The end result of APIAP and non-APIAP scenarios may be the same, but how we validate that in the test should be different.

test/unit/db/seeds_test.rb
```diff
@@ -209,9 +214,9 @@ def teardown
   private
 
   def assert_expected_backend_api
-    assert (service = Account.tenants.first!.default_service)
+    assert(service = Account.tenants.first!.default_service)
```
guicassolato:
```suggestion
    assert (service = Account.tenants.first!.default_service)
```

test/unit/db/seeds_test.rb
```diff
@@ -209,9 +214,9 @@ def teardown
...
     assert_equal 1, service.backend_apis.count
-    assert (backend_api = service.backend_apis.accessible.first)
+    assert(backend_api = service.backend_apis.accessible.first)
```
guicassolato:
```suggestion
    assert (backend_api = service.backend_apis.accessible.first)
```

test/unit/signup/account_manager_test.rb
Martouta: Why it should be different? we want the exact same result, no? so why should we write the same twice? :confused:

test/unit/signup/account_manager_test.rb
Martouta: This is exactly what we had and I am trying to tackle this:
https://github.com/3scale/porta/pull/1292#discussion_r330740443

test/unit/signup/account_manager_test.rb
guicassolato: It's about abstracting implementation details. For APIAP we need to validate that a backend api was created. For non-APIAP what's important is if the service has a given private endpoint set. We don't care about which type of object is caring that info in the latter.

If you will, keep the two test cases as they were and describe them
- "First service has a private endpoint" (This means it is functional for non-APIAP accounts)
- "First backend API is created" (This means https://github.com/3scale/porta/pull/1292 fixed the bug it was meant to)

Martouta: Ok, I understand you now :+1:

test/unit/db/seeds_test.rb
Martouta: Ok but then I will do the same for the entire file. I don't mine `assert(` or `assert (` but it should be the same for the whole test :+1:

test/unit/db/seeds_test.rb
Martouta: same as https://github.com/3scale/porta/pull/1294#discussion_r330980991

test/unit/signup/account_manager_test.rb
Martouta: https://github.com/3scale/porta/pull/1294#discussion_r330980774

test/unit/signup/account_manager_test.rb
Martouta: https://github.com/3scale/porta/pull/1294#discussion_r330980774

test/unit/signup/account_manager_test.rb
guicassolato: It is not the same. I changed second assertion.

@hallelujah's suggestion is valid. I agree. To tackle that, let's add:
```
assert_equal 1, account.backend_apis.count
FactoryBot.create(:simple_service, account: account)
assert_equal 1, account.backend_apis.count
```

test/unit/db/seeds_test.rb
guicassolato: I'm all right with fixing the nil check. I'm just saying we don't need to set the variable.

test/unit/db/seeds_test.rb
guicassolato: Actually, we should avoid setting variables while invoking methods with parameters. (Kinda weird thing to say for Ruby but all right 😁)

Martouta: I disagree with that last part but that'ss a whole different topic :smile:

test/unit/signup/account_manager_test.rb
Martouta: I disagree with that last way to test it because creating a new service by the factory is not testing the ProviderAccountManager what is what this test should be for

test/unit/db/seeds_test.rb
```diff
@@ -90,14 +90,14 @@ def teardown
       assert master_service.metrics.find_by(system_name: system_name, unit: 'hit', friendly_name: description)
     end
 
-    assert(tenant_app_plan = master_service.application_plans.default)
+    assert (tenant_app_plan = master_service.application_plans.default)
```
duduribeiro: I prefer without the space between assert and `(` 😂 .. wdyt?

Martouta: https://github.com/3scale/porta/pull/1294#discussion_r330977927
OMG :joy:

duduribeiro: we should have a style guide hahaha

Martouta:
Proposal from the team:
```
tenant_app_plan = master_service.application_plans.default
assert(tenant_app_plan) 
```

PR #1298, opened 2019-10-04
Passing username from params to login form

damianpm:

When username is present in URL params it must be used to pre-fill username input

Screenshot 2019-10-04 at 11 11 23

### DIFF
--- a/app/javascript/src/LoginPage/loginForms/Login3scaleForm.jsx
+++ b/app/javascript/src/LoginPage/loginForms/Login3scaleForm.jsx
@@ -17,7 +17,10 @@ import {
 } from '@patternfly/react-core'
 
 type Props = {
-  providerSessionsPath: string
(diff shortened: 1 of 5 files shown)


app/javascript/src/LoginPage/loginForms/Login3scaleForm.jsx
```diff
@@ -17,11 +17,14 @@ import {
...
-  providerSessionsPath: string
+  providerSessionsPath: string,
+  session: {
+    username: ?string
+  }
 }
 
 type State = {
-  username: string,
+  username: ?string,
```
josemigallas: I think this is not necessary. Its value is not going to be `undefined`.

damianpm: True, good catch

PR #1300, opened 2019-10-04
When deleting in parallel, try once to unstore credit card before destroying the buyer

Martouta:
Closes [THREESCALE-3637](https://issues.jboss.org/browse/THREESCALE-3637)

### Tasks
- [X] Delete provider buyers before payment setting
- [X] Do not work the hierarchy for buyers twice
- [X] Delete PaymentDetail before its Account

### Verification Steps
I tested also in preview to be more realistic and the way to test it there is:
1. Get a some tenants scheduled for deletion. Some of them big ones. For example by doing: `tenants = Account.tenants.scheduled_for_deletion.first(10)`
2. Do the deletion in Background: `tenants.each { |tenant| DeleteAccountHierarchyWorker.perform_later(tenant) }`
3. Keep track of what's going on in the UI by the URL `/sidekiq/queues` , and check which ones are destroyed by doing in the terminal: `tenants.map { |tenant| Account.exists?(tenant.id) }`. That until all of them are completely destroyed.
4. Download the Sidekiq logs and search there for the order of how everything has been destroyed. For example, this is the format of the data you are looking for:
```json
"arguments": [
  {"_aj_globalid": "gid://system/Settings/3"},
  ["Hierarchy-Account-1", "Hierarchy-PaymentGatewaySetting-4", "Hierarchy-Account-2", "Hierarchy-Settings-2"],
  "destroy"
]
```

### DIFF
--- a/app/workers/delete_account_hierarchy_worker.rb
+++ b/app/workers/delete_account_hierarchy_worker.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 class DeleteAccountHierarchyWorker < DeleteObjectHierarchyWorker
-
   def perform(*)
     return unless account.should_be_deleted?
     super
@@ -21,4 +20,11 @@ class DeleteAccountHierarchyWorker < DeleteObjectHierarchyWorker
     end
   end
 
+  def destroyable_association?(reflection)
+    if called_from_provider_hierarchy? && account.gateway_setting.persisted?
+      super && reflection.class_name != Account.name
+    else
+      super
+    end
+  end
 end
--- a/test/workers/delete_payment_setting_hierarchy_worker_test.rb
+++ b/test/workers/delete_payment_setting_hierarchy_worker_test.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class DeletePaymentSettingHierarchyWorkerTest < ActiveSupport::TestCase
+  include ActiveJob::TestHelper
+  
+  def setup
+    tenant = FactoryBot.create(:simple_provider)
+    tenant.schedule_for_deletion!
+    @payment_gateway_setting = FactoryBot.create(:payment_gateway_setting, account: tenant, tenant_id: tenant.id)
+    @buyers = FactoryBot.create_list(:simple_buyer, 2, provider_account: tenant)
+  end
+
+  attr_reader :payment_gateway_setting, :buyers
+
+  test 'it performs for the buyers is it is destroyed by the provider association' do
+    perform_enqueued_jobs do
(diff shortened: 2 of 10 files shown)


app/workers/delete_account_hierarchy_worker.rb
```diff
@@ -21,4 +20,12 @@ def batch_description
     end
   end
 
+  # :reek:NilCheck
```
macejmic: wouldn't be better to make `destroyable_association?` a protected method and rewrite that logic? it could be nice one liner..

macejmic: and no additional `loop` would be necessary.

Martouta: ah right :+1: good one :+1:

guicassolato:
One of the tasks in the description says:

> Delete PaymentDetail after its Account

I don't see that in the code. Actually, the other way around. So I think you meant _before_, not _after_. Specially because deleting the payment detail _after_ the account would have no meaning, given the propagation policy of the foreign key:

https://github.com/3scale/porta/blob/12c1b2d39cb540fcd614db499c896b0d18db9ccc/db/schema.rb#L1544

Martouta:
@guicassolato right, sorry, I meant this:
- https://github.com/3scale/porta/pull/1300/files#diff-79739d7eb5a7ffdc6620e2af64fcfe20R7
- https://github.com/3scale/porta/pull/1300/files#diff-9a6f99081e89cd7feb3ad45390033075R84

I will fix the description accordingly :+1:

test/workers/delete_payment_setting_hierarchy_worker_test.rb
```diff
@@ -0,0 +1,29 @@
...
+    tenant = FactoryBot.create(:simple_provider)
+    tenant.schedule_for_deletion!
+    @payment_gateway_setting = FactoryBot.create(:payment_gateway_setting, account: tenant, tenant_id: tenant.id)
+    @buyers = FactoryBot.create_list(:simple_buyer, 2, provider_account: tenant)
+  end
+
+  attr_reader :payment_gateway_setting, :buyers
+
+  test 'it performs for the buyers is it is destroyed by the provider association' do
+    Sidekiq::Testing.inline! do
```
duduribeiro: since this is a activejob worker, what do you think about using:
```
perform_enqueued_jobs(only: DeletePaymentSettingHierarchyWorker)
```
?

Martouta: @duduribeiro i can do that but without the `only` because i want it to always execute the whole hierarchy chain, no matter which worker uses for the whole chain.

PR #1302, opened 2019-10-04
Do not allow access to edit or update IntegrationsController for apiap

Martouta:
[THREESCALE-3625](https://issues.jboss.org/browse/THREESCALE-3625)
APIAP-users should not have access to `/apiconfig/services/:service_id/integration/edit` and before this PR they do by typing the URL on the browser. This PR blocks access to the `edit` action of this controller for APIAP users.

### DIFF
--- a/app/controllers/api/integrations_controller.rb
+++ b/app/controllers/api/integrations_controller.rb
@@ -4,6 +4,7 @@ class Api::IntegrationsController < Api::BaseController
   before_action :find_service
   before_action :find_proxy
   before_action :authorize
+  before_action :hide_for_apiap, only: :edit
 
   activate_menu :serviceadmin, :integration, :configuration
   sublayout 'api/service'
@@ -203,6 +204,10 @@ class Api::IntegrationsController < Api::BaseController
 
   def message_bus?(proxy)
     proxy.oidc? && ZyncWorker.config.message_bus
+  end
+
+  def hide_for_apiap
+    raise ActiveRecord::RecordNotFound if apiap?
   end
 
   def authorize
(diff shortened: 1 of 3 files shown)


app/controllers/api/integrations_controller.rb
```diff
@@ -205,6 +206,10 @@ def message_bus?(proxy)
     proxy.oidc? && ZyncWorker.config.message_bus
   end
 
+  def authorize_not_apiap
```
duduribeiro: wdyt about `authorize_unless_apiap` ? 🤔 

or

```
unauthorize_apiap
```

Martouta: u r right :+1: Better :+1:

app/controllers/api/integrations_controller.rb
```diff
@@ -4,6 +4,7 @@ class Api::IntegrationsController < Api::BaseController
   before_action :find_service
   before_action :find_proxy
   before_action :authorize
+  before_action :unauthorize_apiap, only: %i[edit update]
```
macejmic: `unauthorize` sounds pretty weird to me.. what about simple `raise_if_apiap` ?

Martouta: @duduribeiro @macejmic WDYT about `raise_not_found_if_apipap`? because simply raise... it is like 'raise what?'

guicassolato:
```suggestion
  before_action :unauthorize_apiap, only: :edit
```

guicassolato: I think we still need the update.

guicassolato: This will happen with the "Promote" buttons if we unauthorize the `update` action:

Screen Shot 2019-10-08 at 10 29 38 AM

PR #1304, opened 2019-10-07
Update Wizard with steps for Product/Backend

josemigallas:

It updates the wizard with the new features of APIAP. It should change only for providers with `api_as_product` enabled.

TODO:

- [x] Make it work
- [x] Add some tests
- [x] Fix those tests! 🥒🔥

First screen hasn't changed (only the icon):
Screen Shot 2019-10-07 at 17 28 44

Then a screen to create a Product and a Backend:
Screen Shot 2019-10-07 at 17 28 50

Screen Shot 2019-10-07 at 17 29 35

And next a screen to connect Backend with Product (icon is a bit lame, it would be interesting to have some cubes connecting to the pile of cubes):
Screen Shot 2019-10-09 at 16 17 28

Continues with testing the endpoint and that's it.

[THREESCALE-3372: Update the wizard to explain API as a product](https://issues.jboss.org/browse/THREESCALE-3372)

**Verification steps** 

Wizard: `/p/admin/onboarding/wizard/intro`


guicassolato:
It looks all right. Is Backend following Product something decided and closed to opinions or can I still propose first Backend and then Product?

features/provider/onboarding/wizard.feature
```diff
@@ -14,3 +14,18 @@ Feature: Onboarding Wizard
...
      And goes to what's next
     Then goes to API page
+
+  @emails
+  Scenario: Provider goes through the APIAP wizard
+    Given a provider signs up and activates his account
+    When user starts the onboarding wizard
+     And goes to Add a Product page
+     And adds the echo Product
+     And goes to Add a Backend page
```
josemigallas: @thomasmaas
Do you know why this step isn't working? Is it because it's `And` and not `Then`?

thomasmaas: No, And and Then have no consequences for the test.

thomasmaas:
@josemigallas 

- are you taking into account that apiap steps should be behind rolling update?
- I keep thinking the order should be: 
Screenshot 2019-10-10 12 02 54
Screenshot 2019-10-10 12 03 00
Screenshot 2019-10-10 12 03 13

josemigallas:
@thomasmaas 
- Rolling update: yes
- Order: definitely a valid option but let's discuss with the wider group.

josemigallas:
@thomasmaas Some suggestions for your texts:

**API**
- _Add your API_ should be only singular. Mainly because the wizard adds only 1 API and also to keep it simple. The possibility of more APIs is already mentioned when adding the Product, which makes more sense.
- Metrics and Methods: it could be more intuitive by adding _allows you to keep track of requests by defining Metrics and Methods_, wdyt?
- _please use the Echo-api_ -> _you can use the Echo-api_? Less aggressive?

**Product**
- I'm not sure of the _design project_ terminology. It sounds too strong and IMO it's better to simply "create a project". Not only a Project isn't really _designed_ in the context of 3scale but also the wizard is only adding a name.

  Maybe in the documentation, where the Product concept is fully described, it would make sense to use the word _design_.

- _with which your costumers interact_ -> _which your costumers interact with_

- _For the purpose of this exercise_ sounds too academical, no? Maybe _First let's keep it simple, ..._

thomasmaas:
@josemigallas i made some changes based on your feedback. I want to keep the more generic titles (add API(s) & design product) as i think it helps in understanding the potentials.

app/views/provider/admin/onboarding/wizard/backend_api/new.html.slim
```diff
@@ -0,0 +1,28 @@
...
+      ' In 3scale each of your APIs is represented by what we call a
+      strong Backend
+      | .
+    p.text.text--up
+      ' A Backend contains the
+      strong> Base URL
+      ' of your API and helps you to keep track of requests through
+      strong> Metrics, Methods and Mapping Rules.
+    p.text.text--up
+      ' Let's add your first Backend (If you don’t have a publicly accessible API yet, please
```
guicassolato: "publicly" makes sense for hosted apicast in Saas, but possibly not so much for some other scenarios.

```suggestion
      ' Let's add your first Backend (If you don’t have an accessible API yet, please
```

app/controllers/provider/admin/onboarding/wizard/product_controller.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+class  Provider::Admin::Onboarding::Wizard::ProductController < Provider::Admin::Onboarding::Wizard::BaseController
+  respond_to :html
+
+  def new
+    @service = service
```
damianpm: This is repeated, you may extract this to a `before_action` call

PR #1306, opened 2019-10-07
Index on proxy_configs(proxy_id, environment, version)

guicassolato:

Defines a database index on `proxy_configs(proxy_id, environment, version)`, so the SQL query corresponding to the following scope can perform better: https://github.com/3scale/porta/blob/ab53f034d27d61db26710706256a4dd30b3bd6fd/app/models/proxy_config.rb#L40-L47 

Closes [THREESCALE-3054](https://issues.jboss.org/browse/THREESCALE-3054)


guicassolato:
Benchmarked in preview env. Migration took 1.237s to create the new index.

PR #1307, opened 2019-10-07
adds liquid check for admin role

kevprice83:
Fixes [THREESCALE-1142](https://issues.jboss.org/browse/THREESCALE-1142)

404 Not Found error is returned for member users when they try to navigate to some of the presented menu items. These menu items shouldn't be rendered if the user is not an admin user and therefore restricted access is enforced whilst hiding those menus.

### DIFF
--- a/lib/developer_portal/app/views/developer_portal/_users_menu.html.liquid
+++ b/lib/developer_portal/app/views/developer_portal/_users_menu.html.liquid
@@ -5,10 +5,14 @@
     </li>
 
     {% if provider.account_management_enabled? %}
-      {% include 'menu_item' with urls.users %}
+      {% if current_user.admin? %}
+        {% include 'menu_item' with urls.users %}
+      {% endif %}
 
       {% if provider.multiple_users_allowed? %}
-        {% include 'menu_item' with urls.invitations %}
+        {% if current_user.admin? %}
+          {% include 'menu_item' with urls.invitations %}
+        {% endif %}
       {% endif %}
     {% endif %}
 


kevprice83:
@Martouta is there somewhere we are testing the developer portal templates? The only tests I see for [developer portal](https://github.com/3scale/porta/tree/master/test/support/functional/developer_portal) doesn't seem to have much...

thomasmaas:
@kevprice83 shouldn't all items in `users_menu` only be accessible for admin users?

kevprice83:
@thomasmaas according to the solution [here](https://access.redhat.com/solutions/3528941) a member should not have access to the 2 tabs protected in this PR. Do we have any documentation that says there are further restrictions? I couldn't find anything...

thomasmaas:
I don't have access to that content but i can imagine only admins should have access to things like payment details…

kevprice83:
@thomasmaas I mean that a member can access the invoice tabs (among others) without getting a 404, so if they should be restricted then I guess there is some restriction that is not being enforced in the code. This PR is hiding the tabs that return a 404 for a member user.

kevprice83:
@thomasmaas yeah it could be done in one line, if you prefer that let me know and I'll update it, for existing accounts the KCS I linked above shows the same solution that customers can apply themselves.

thomasmaas:
To really make this a valid pull request a regression test should be added. Maybe @Martouta can give you some hints on how to do that.

kevprice83:
@thomasmaas @Martouta I've been looking through all our tests and I cannot find any that would be like _integration_ tests on the developer portal, I see some [javascript tests](https://github.com/3scale/porta/tree/master/spec/javascripts) & [unit tests for liquid tags](https://github.com/3scale/porta/tree/master/test/unit/liquid). Maybe I'm looking in the wrong place?

thomasmaas:
@kevprice83 did you check https://github.com/3scale/porta/tree/master/test/integration/developer_portal?

PR #1308, opened 2019-10-07
Avoid Empty paths for BackendApiConfig

duduribeiro:

MySQL and Postgres saves `''` as empty string, but Oracle saves it as null.
Now for the first service of each tenant, we link it to a new backend api, with empty string path in the backend api config. The problem is that we are checking the uniqueness of service_id and path for backend_api_config but for Oracle, it searches by empty string although it is saved as null, so it does not find it by Rails and it tries to save it but raises an error.
We also have this problem if we try to use `where(path: '')` or a `find_by`, that it tries to search by `''` but it is saved as null so it does not find it although it is there. 

This commit changes this behave to make the application to save `'/'` instead of the empty string. 

[THREESCALE-3631](https://issues.jboss.org/browse/THREESCALE-3631)

### DIFF
--- a/app/lib/backend_api_logic/routing_policy.rb
+++ b/app/lib/backend_api_logic/routing_policy.rb
@@ -43,6 +43,10 @@ module BackendApiLogic
           @config = config
         end
 
+        def config_path
+          @config_path ||= ConfigPath.new(path)
+        end
+
         def as_json
           return if private_endpoint.blank?
           {
@@ -51,24 +55,16 @@ module BackendApiLogic
               operations: [
                 match: :path,
                 op: :matches,
-                value: path_to_regex
+                value: config_path.to_regex
               ]
             }
           }.merge(replace_path)
         end
 
-        def path_to_regex
-          if path.to_s.empty?
-            "/.*"
-          else
-            "/#{path}/.*|/#{path}/?"
-          end
-        end
-
         def replace_path
-          return {} if path.blank?
+          return {} if config_path.blank?
 
-          { replace_path: "{{original_request.path | remove_first: '/#{path}'}}" }
+          { replace_path: "{{original_request.path | remove_first: '#{config_path.path}'}}" }
         end
       end
       private_constant :Rule
--- a/app/models/backend_api_config.rb
+++ b/app/models/backend_api_config.rb
@@ -14,22 +14,17 @@ class BackendApiConfig < ApplicationRecord
   self.default_sort_column = 'backend_apis.name'
   self.default_sort_direction = :asc
 
-  attribute :path, ActiveRecord::Type::StringNotNil.new
-
   has_many :backend_api_metrics, through: :backend_api, source: :metrics
 
   validates :service_id, :backend_api_id, presence: true
   validates :backend_api_id, uniqueness: { scope: :service_id }
   validates :path, uniqueness: { scope: :service_id, case_sensitive: false }
-  validates :path, length: { in: 0..255, allow_nil: false }, path: true
+  validates :path, length: { in: 1..255, allow_nil: false }, path: true
 
   scope :by_service,     ->(service_id)     { where.has { self.service_id     == service_id     } }
   scope :by_backend_api, ->(backend_api_id) { where.has { self.backend_api_id == backend_api_id } }
 
-  scope :with_subpath, -> do
-    common_query = where.not(path: '/')
-    System::Database.oracle? ? common_query.where('path is NOT NULL') : common_query.where.not(path: '')
-  end
+  scope :with_subpath, -> { where.not(path: ConfigPath::EMPTY_PATH) }
 
   scope :accessible, -> do
     joining { service }.where.has { (service.state != ::Service::DELETE_STATE) }
@@ -38,6 +33,6 @@ class BackendApiConfig < ApplicationRecord
   delegate :private_endpoint, to: :backend_api
 
   def path=(value)
-    super(StringUtils::StripSlash.strip_slash(value))
+    super(ConfigPath.new(value).path)
   end
 end
--- a/app/models/config_path.rb
+++ b/app/models/config_path.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+class ConfigPath
+  EMPTY_PATH = '/'
+
+  attr_reader :path
+
+  def initialize(value)
+    self.path = value
+  end
+
+  def to_regex
+    return '/.*' if empty?
+
+    "#{path}/.*|#{path}/?"
+  end
+
+  def empty?
+    path == EMPTY_PATH
+  end
+
+  private
+
+  def path=(value)
+    @path = EMPTY_PATH + StringUtils::StripSlash.strip_slash(value)
+  end
+end
--- a/lib/tasks/data_migration/fix_backend_config_path.rb
+++ b/lib/tasks/data_migration/fix_backend_config_path.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+namespace :data_migration do
+  desc 'Backfill Paths for Backend Api Configs'
+  task :fix_backend_config_path do
+    query = System::Database.oracle? ? BackendApiConfig.where('path is NULL') : BackendApiConfig.where(path: '')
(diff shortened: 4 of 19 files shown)


db/migrate/20191007161920_backfill_paths_for_backend_api_configs.rb
```diff
@@ -0,0 +1,12 @@
+class BackfillPathsForBackendApiConfigs < ActiveRecord::Migration
+  disable_ddl_transaction!
+
+  def change
+    query = System::Database.oracle? ? BackendApiConfig.where('path is NULL') : BackendApiConfig.where(path: '')
```
guicassolato: We should avoid DML in Rails migrations as they are not tested and tend to break the upgrade path of onprem customers.

Martouta:
This line: https://github.com/3scale/porta/blob/fedd08b63974d66bece271d608704a8a053674b4/app/models/backend_api_config.rb#L24
Should become:
```ruby
validates :path, length: { in: 1..255, allow_nil: false }, path: true
```

app/models/backend_api_config.rb
```diff
@@ -27,8 +27,7 @@ class BackendApiConfig < ApplicationRecord
   scope :by_backend_api, ->(backend_api_id) { where.has { self.backend_api_id == backend_api_id } }
 
   scope :with_subpath, -> do
-    common_query = where.not(path: '/')
-    System::Database.oracle? ? common_query.where('path is NOT NULL') : common_query.where.not(path: '')
+    where.not(path: ConfigPath::EMPTY_PATH)
```
Martouta: This is great, but we can also simplify its test accordingly. So we should just remove the next 2 lines:
https://github.com/3scale/porta/blob/fedd08b63974d66bece271d608704a8a053674b4/test/models/backend_api_config_test.rb#L30
https://github.com/3scale/porta/blob/fedd08b63974d66bece271d608704a8a053674b4/test/models/backend_api_config_test.rb#L37

Martouta: Btw this scope now can be done in 1 line:
```ruby
scope :with_subpath, -> { where.not(path: ConfigPath::EMPTY_PATH) }
```

app/lib/backend_api_logic/routing_policy.rb
```diff
@@ -51,24 +51,16 @@ def as_json
...
-          if path.to_s.empty?
-            "/.*"
-          else
-            "/#{path}/.*|/#{path}/?"
-          end
-        end
-
         def replace_path
-          return {} if path.blank?
+          return {} if config_path.blank?
```
Martouta: `blank?` internally calls the `empty?` of config_path (if it is a ConfigPath and not nil).

https://github.com/rails/rails/blob/e9d6b85f3e834ceea2aeabe4cbaa96a7c73eb896/activesupport/lib/active_support/core_ext/object/blank.rb#L4-L18

Just saying this because I found like black magic how this was working. And I see that it is well tested in:
https://github.com/3scale/porta/blob/fedd08b63974d66bece271d608704a8a053674b4/test/unit/backend_api_logic/routing_policy_test.rb#L46

Good :+1:

app/lib/backend_api_logic/routing_policy.rb
```diff
@@ -37,7 +37,7 @@ def to_a
       end
 
       class Rule
-        delegate :private_endpoint, :path, to: :@config
+        delegate :private_endpoint, :config_path, to: :@config
```
Martouta: Imo 'config_path' shouldn't be a method belonging to BackendApiConfig but belonging to this specific class, like this:
image

And then the test:
image

'config_path' is not data itself but manipulation of it, as business logic, and needed only for this class, so i that's why i would put it here instead of in the model.

test/unit/tasks/data_migration/fix_backend_config_path_test.rb
```diff
@@ -0,0 +1,20 @@
+require 'test_helper'
+
+class Tasks::FixBackendConfigPathTest < ActiveSupport::TestCase
+  fixtures :backend_api_configs
```
Martouta: I don't know how this works so I better ask... is this testing the case of oracle as well having it internally saved as `null` o not? because we should definitely test that too.

test/models/backend_api_config_test.rb
```diff
@@ -5,18 +5,18 @@ def setup
     @config = BackendApiConfig.new
```
Martouta: Do not forget to do and test also this: https://github.com/3scale/porta/pull/1308#issuecomment-540135411

Martouta: btw also this might not be necessary anymore to convert:
https://github.com/3scale/porta/blob/734c29a867ae443da0c54c17a04d310a338b51b6/app/models/backend_api_config.rb#L17

test/models/backend_api_config_test.rb
```diff
@@ -5,18 +5,18 @@ def setup
...
+    assert_equal '/hello', @config.path
 
     @config.path = 'hello/my/name/is'
-    assert_equal 'hello/my/name/is', @config.path
-
-    @config.path = 'hello/my/name/is/'
-    assert_equal 'hello/my/name/is', @config.path
+    assert_equal '/hello/my/name/is', @config.path
 
-    @config.path = '//hello/my/name/is/john//'
```
hallelujah: Why do you change this test?

Martouta:
You will need to change also this line:
https://github.com/3scale/porta/blob/3b298e604d410724aea773913164ccd3881a8c86/lib/tasks/services.rake#L34
It should become:
`service.backend_api_configs.create(backend_api: backend_api, path: '/')`

test/unit/tasks/data_migration/fix_backend_config_path_test.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  setup do
+    @config_without_path = backend_api_configs(:empty_path)
+    @config_with_path    = backend_api_configs(:with_path)
+  end
+
+  test 'migrates all configs with empty path to have a backslash on it' do
+    assert_empty @config_without_path.path
```
Martouta:
```suggestion
    assert @config_without_path.path.blank?
```
To fix [the oracle test](https://circleci.com/gh/3scale/porta/117572#tests/containers/1).

app/models/config_path.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  private
+
+  def path=(value)
+    @path = value.dup
+    @path = remove_duplicated_slashes(@path)
+    @path.insert(0, EMPTY_PATH) unless @path.start_with?(EMPTY_PATH)
+  end
+
+  def remove_duplicated_slashes(path)
```
hallelujah: Use https://github.com/3scale/porta/blob/3b298e604d410724aea773913164ccd3881a8c86/app/lib/string_utils/strip_slash.rb#L8-L10

lib/tasks/data_migration/fix_backend_config_path.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+namespace :data_migration do
+  desc 'Backfill Paths for Backend Api Configs'
+  task :fix_backend_config_path do
+    query = System::Database.oracle? ? BackendApiConfig.where('path is NULL') : BackendApiConfig.where(path: '')
+    progress = ProgressCounter.new(query.count)
+    BackendApiConfig.transaction do
+      query.find_in_batches(batch_size: 200) do |records|
```
hallelujah: Can you do `select(:id)` so we do not load all the fields?

app/models/config_path.rb
```diff
@@ -0,0 +1,33 @@
...
+    path == EMPTY_PATH
+  end
+
+  private
+
+  def path=(value)
+    @path = value.dup
+    @path = remove_duplicated_slashes(@path)
+    @path.insert(0, EMPTY_PATH) unless @path.start_with?(EMPTY_PATH)
+  end
```
hallelujah: Can be simplified by:

```suggestion
  def path=(value)
    super(EMPTY_PATH + StringUtils::strip_slashes(value))
  end
```

PR #1310, opened 2019-10-08
[CircleCI] Do not run nightly 2.4 tests twice

Martouta:
Until 1 before I opened this PR, right before merging https://github.com/3scale/porta/pull/1290, we had the following builds related to DBs running every night:
1. mysql-container (for Ruby 2.3)
2. postgres-container (for Ruby 2.3)
3. oracle-db-container (for Ruby 2.3)
4. mysql-ruby24 (for Ruby 2.4)
5. postgres-ruby24 (for Ruby 2.4)
6. oracle-ruby24 (for Ruby 2.4)
But now the 3 first ones already run with 2.4, so we should not run the 3 last ones anymore because they are exactly the same :smile: 

#### Validation
image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -73,7 +73,7 @@ attach-to-workspace: &attach-to-workspace
     at: .
 
 system-builder-ruby24: &system-builder-ruby24
-  image: quay.io/3scale/system-builder:latest
+  image: quay.io/3scale/system-builder:ruby24
   environment:
     BUNDLE_FROZEN: true
     BUNDLE_PATH: 'vendor/bundle'
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -73,7 +73,7 @@ attach-to-workspace: &attach-to-workspace
     at: .
 
 system-builder-ruby24: &system-builder-ruby24
-  image: quay.io/3scale/system-builder:ruby24
+  image: quay.io/3scale/system-builder:latest
   environment:
     BUNDLE_FROZEN: true
     BUNDLE_PATH: 'vendor/bundle'
@@ -1080,10 +1080,11 @@ workflows:
             - assets_precompile
           <<: *only-master-filter
 
-######## Nightly workflows
+
+  ######## Nightly workflows
 
 
-  mysql_build_ruby24:
+  mysql_nightly_build_ruby24:
     jobs:
       - notify_start:
           <<: *only-master-filter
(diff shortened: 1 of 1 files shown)


guicassolato:
We should also get rid of the executors.

.circleci/config.yml
```diff
@@ -72,7 +72,7 @@ attach-to-workspace: &attach-to-workspace
   attach_workspace:
     at: .
 
-system-builder-ruby24: &system-builder-ruby24
+system-builder-latest: &system-builder-latest
```
gsaslis: i think i prefer the explicit `ruby24` tag than the more generic `latest` 🤔

duduribeiro: Agree. `Latest` gives the impression that we will run into the latest version of ruby

.circleci/config.yml
```diff
@@ -72,7 +72,7 @@ attach-to-workspace: &attach-to-workspace
...
   image: quay.io/3scale/system-builder:ruby24
```
hallelujah:
```suggestion
  image: quay.io/3scale/system-builder:latest
```

.circleci/config.yml
```diff
@@ -1135,207 +1079,3 @@ workflows:
           requires:
             - assets_precompile
           <<: *only-master-filter
-
-
-######## Nightly workflows
```
hallelujah: Why do we remove the nightly's?

It is good to keep them, at least for postgres and oracle because the other workflows are manual approval only

Martouta: This is what I mentioned yesterday that I don't know how to do... As I said, the goal is to run the tests once for each DB + JS, but I don't know how to do that and I suspected with this change they wouldn't run at all.
I don't want them to run twice, that is why I tried by removing those lines, but I felt it was wrong.
I need help though :stuck_out_tongue:

Martouta: This :smile: 
image

hallelujah: Just keep them, the executor should be the same as for the Pull Requests workflow

Validate the config before pushing with: `circleci config validate`

Martouta: Awesome! Thanks! :muscle: :dancer:

.circleci/config.yml
```diff
@@ -73,7 +73,7 @@ attach-to-workspace: &attach-to-workspace
     at: .
 
 system-builder-ruby24: &system-builder-ruby24
-  image: quay.io/3scale/system-builder:ruby24
+  image: quay.io/3scale/system-builder:latest
```
hallelujah: Preferably use ruby24 but then tag it in the system-builder. I will do

PR #1313, opened 2019-10-08
Fixes TypeError: json is null in CMS

josemigallas:

Catches when json is `null` and shows an error message.

Also minor lint fixes.

https://app.bugsnag.com/3scale-networks-sl/system/errors/5d9ba1d604b3fe0019229e5d?event_id=5d9ba1d60050a3b0d87f0000&i=sk&m=nw

**Verification steps** 
1. Override AJAX call so it returns `null`
2. There should be no error in the console
3. An error message should be shown in the sidebar

### DIFF
--- a/app/views/layouts/_cms.html.erb
+++ b/app/views/layouts/_cms.html.erb
@@ -25,6 +25,7 @@
   <div id="cms-sidebar-layouts" class="cms-sidebar-listing"></div>
   <div id="cms-sidebar-partials" class="cms-sidebar-listing"></div>
   <div id="cms-sidebar-portlets" class="cms-sidebar-listing"></div>
+  <p id="cms-sidebar-error-empty">The folder tree could not load. Please retry.</p>
 </div>
 
 <div id="tab-content">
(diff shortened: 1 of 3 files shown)


app/views/layouts/_cms.html.erb
```diff
@@ -25,6 +25,7 @@
   <div id="cms-sidebar-layouts" class="cms-sidebar-listing"></div>
   <div id="cms-sidebar-partials" class="cms-sidebar-listing"></div>
   <div id="cms-sidebar-portlets" class="cms-sidebar-listing"></div>
+  <p id="cms-sidebar-error-empty">The folder tree could not load. Please retry.</p>
```
thomasmaas:
```suggestion
  <p id="cms-sidebar-error-empty">Something went wrong, please reload the page.</p>
```

thomasmaas:
We're moving the error from the console to the ui, how is this helpful for the user? A refresh really fixes the problem?

josemigallas:
@thomasmaas That is what I wanted to know earlier, I'd like to know where this error may come from. I think it's just an AJAX failure, maybe connection broke or something. We're moving a false error IMO, from being visible to something controlled. If AJAX returns null it's up to the script to show a message instead of a blank screen.

About the refresh, it should, but can't be sure 🤷‍♂.

PR #1315, opened 2019-10-09
[settings] New integration settings component

didierofrivia:

The new integration settings page needed a good refactor, it was crafted in a hacky fashion for 2.7 ER1

fixes https://issues.jboss.org/browse/THREESCALE-3563

**Verification Steps**
Apply this patch to make it work:
```
diff --git a/app/javascript/packs/settings.js b/app/javascript/packs/settings.js
index e9bb1a69..0bfeb4b6 100644
--- a/app/javascript/packs/settings.js
+++ b/app/javascript/packs/settings.js
@@ -1,5 +1,5 @@
-import { initialize as authenticationWidget } from 'Settings/authentication-widget'
+import { initSettings as settingsWidgets } from 'Settings/index'

-  authenticationWidget()
+  settingsWidgets({elementId: 'settings'})
 })
```

 
**TODO**
* [x] Tests
* [x] Fix Styles
* [x] Create issue for "Rails Helper" https://issues.jboss.org/browse/THREESCALE-3822
* [x] Create issue for "Remove old implementation" https://issues.jboss.org/browse/THREESCALE-3823

### DIFF
--- a/app/javascript/src/Settings/components/AuthenticationSettingsFieldset.jsx
+++ b/app/javascript/src/Settings/components/AuthenticationSettingsFieldset.jsx
@@ -1,0 +1,47 @@
+// @flow
+
+import * as React from 'react'
+import { FormFieldset, FormLegend } from 'Form'
+import { FormCollection, TextInputGroup } from 'Settings/components/Common'
+import { OidcFieldset } from 'Settings/components/OidcFieldset'
+import type { FieldGroupProps, TypeItemProps } from 'Settings/types'
+
+const OIDC_AUTH_METHOD = 'oidc'
+const API_KEY_METHOD = '1'
+const APP_ID_KEY_METHOD = '2'
+
+type Props = {
+  isServiceMesh: boolean,
+  authenticationMethod: string,
+  apiKeySettings: FieldGroupProps,
+  appIdKeyPairSettings: FieldGroupProps[],
+  oidcSettings: {
+    basicSettings: TypeItemProps,
+    flowSettings: FieldGroupProps[],
+    jwtSettings: TypeItemProps
+  }
+}
+
+const AuthenticationSettingsFieldset = ({
+  isServiceMesh,
+  authenticationMethod,
+  apiKeySettings,
+  appIdKeyPairSettings,
+  oidcSettings
+}: Props) => {
+  const isOidc = authenticationMethod === OIDC_AUTH_METHOD
+  const isApiKey = authenticationMethod === API_KEY_METHOD
+  const isAppIdKey = authenticationMethod === APP_ID_KEY_METHOD
+  return (
+    (!isServiceMesh || isOidc) && <FormFieldset id='fieldset-AuthenticationSettings'>
+      <FormLegend>Authentication Settings</FormLegend>
+      { isApiKey && <FormCollection collection={[apiKeySettings]} ItemComponent={TextInputGroup} legend='API KEY (USER_KEY) BASICS' /> }
+      { isAppIdKey && <FormCollection collection={appIdKeyPairSettings} ItemComponent={TextInputGroup} legend='APP_ID AND APP_KEY PAIR BASICS' /> }
+      { isOidc && <OidcFieldset {...oidcSettings} isServiceMesh={isServiceMesh} /> }
+    </FormFieldset>
+  )
+}
+
+export {
+  AuthenticationSettingsFieldset
+}
--- a/app/javascript/src/Settings/components/Common/FormCollection.jsx
+++ b/app/javascript/src/Settings/components/Common/FormCollection.jsx
@@ -1,0 +1,24 @@
+// @flow
+
+import * as React from 'react'
+import { FormFieldset, FormLegend } from 'Form'
+import type { FieldGroupProps } from 'Settings/types'
+
+type Props = {
+  collection: FieldGroupProps[],
+  ItemComponent: (FieldGroupProps) => React.Element<empty>,
+  legend: string
+}
+
+const FormCollection = ({collection, ItemComponent, legend}: Props) => {
+  return (
+    <FormFieldset id={`fieldset-${legend.replace(/\s+/g, '')}`}>
+      <FormLegend>{legend}</FormLegend>
+      { collection.map(itemProps => <ItemComponent {...itemProps} key={itemProps.name} />) }
+    </FormFieldset>
+  )
+}
+
+export {
+  FormCollection
+}
--- a/app/javascript/src/Settings/components/Common/RadioFieldset.jsx
+++ b/app/javascript/src/Settings/components/Common/RadioFieldset.jsx
@@ -1,0 +1,50 @@
+// @flow
+
+import * as React from 'react'
+import { useState } from 'react'
+import { FormFieldset, FormLegend } from 'Form'
+import { Radio } from '@patternfly/react-core'
+import type { FieldGroupProps, FieldCatalogProps } from 'Settings/types'
+
+type CheckEvent = SyntheticEvent<HTMLButtonElement>
+
+type Props = FieldGroupProps & FieldCatalogProps
+
+const useSelectedOnChange = (value, onChange) => (
+  typeof onChange === 'function'
+    ? [value, onChange]
+    : useState(value).map(x => typeof x === 'function' ? (_c, e: CheckEvent) => x(e.currentTarget.value) : x)
+)
+
+const RadioFieldset = ({
+  children,
+  legend,
+  name,
+  value,
+  catalog,
+  onChange,
+  ...props
+}: Props) => {
+  const [selectedOnChange, setSelectedOnChange] = useSelectedOnChange(value, onChange)
+  return (
+    <FormFieldset id={`fieldset-${name}`} {...props} >
+      <FormLegend>{legend}</FormLegend>
+      {Object.keys(catalog).map(key => (
+        <Radio
+          key={key}
+          value={key}
+          isChecked={selectedOnChange === key}
+          name={`service[${name}]`}
+          onChange={setSelectedOnChange}
+          label={catalog[key]}
+          id={`service_method_${name}_${key}`}
+        />
+      ))}
+      {children}
+    </FormFieldset>
+  )
+}
+
+export {
+  RadioFieldset
+}
--- a/app/javascript/src/Settings/components/Common/SelectGroup.jsx
+++ b/app/javascript/src/Settings/components/Common/SelectGroup.jsx
@@ -1,0 +1,44 @@
+// @flow
+
+import * as React from 'react'
+import { useState } from 'react'
+import { FormGroup, Select, SelectOption } from '@patternfly/react-core'
+import type { FieldGroupProps, FieldCatalogProps } from 'Settings/types'
+
+type Props = FieldGroupProps & FieldCatalogProps
+
+const SelectGroup = ({
+  label,
+  name,
+  hint,
+  value,
+  catalog
+}: Props) => {
+  const [ selectedValue, setSelectedValue ] = useState(value)
+  const [ isExpanded, setIsExpanded ] = useState(false)
+  const onSelect = (_e, selection) => {
+    setIsExpanded(false)
+    setSelectedValue(selection.key)
+  }
+
+  return (
+    <FormGroup label={label} fieldId={`service_proxy_attributes_${name}_select`} helperText={hint}>
+      <input name={`service[proxy_attributes][${name}]`} type='hidden' value={selectedValue} />
+      <Select
+        selections={catalog[selectedValue]}
+        onToggle={setIsExpanded}
+        onSelect={onSelect}
+        isExpanded={isExpanded}
+        id={`service_proxy_attributes_${name}_select`}
+      >
+        {Object.keys(catalog).map(key => (
+          <SelectOption key={catalog[key]} value={{ key, toString: () => catalog[key] }} />
+        ))}
+      </Select>
+    </FormGroup>
+  )
+}
+
+export {
+  SelectGroup
+}
--- a/app/javascript/src/Settings/components/Common/TextInputGroup.jsx
+++ b/app/javascript/src/Settings/components/Common/TextInputGroup.jsx
@@ -1,0 +1,38 @@
+// @flow
+
+import * as React from 'react'
+import { useState } from 'react'
+import { FormGroup, TextInput } from '@patternfly/react-core'
+import type { FieldGroupProps } from 'Settings/types'
+
+const TextInputGroup = ({
+  defaultValue,
+  placeholder,
+  label,
+  name,
+  hint,
+  value,
+  isDefaultValue = false,
+  readOnly = false,
+  inputType = 'text'
+}: FieldGroupProps) => {
+  const [ inputValue, setInputValue ] = useState(value)
+  const onChange = (value, _e) => setInputValue(value)
+  return (
+    <FormGroup label={label} fieldId={`service_proxy_attributes_${name}_input`} helperText={hint}>
+      <TextInput
+        id={`service_proxy_attributes_${name}_input`}
+        name={`service[proxy_attributes][${name}]`}
+        placeholder={placeholder}
+        value={isDefaultValue ? defaultValue : inputValue}
+        type={inputType}
+        onChange={onChange}
+        isReadOnly={readOnly}
+      />
+    </FormGroup>
+  )
+}
+
+export {
+  TextInputGroup
+}
--- a/app/javascript/src/Settings/components/Common/TypeItemCombo.jsx
+++ b/app/javascript/src/Settings/components/Common/TypeItemCombo.jsx
@@ -1,0 +1,26 @@
+// @flow
+
+import * as React from 'react'
+import { FormFieldset, FormLegend } from 'Form'
+import { TextInputGroup, SelectGroup } from 'Settings/components/Common'
+import type { FieldGroupProps, FieldCatalogProps } from 'Settings/types'
+
+type Props = {
+  type: FieldCatalogProps & FieldGroupProps,
+  item: FieldGroupProps,
+  legend: string
+}
+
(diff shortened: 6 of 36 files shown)


app/javascript/src/Settings/components/AuthenticationSettingsFieldset.jsx
```diff
@@ -0,0 +1,47 @@
...
+}: Props) => {
+  const isOidc = authenticationMethod === OIDC_AUTH_METHOD
+  const isApiKey = authenticationMethod === API_KEY_METHOD
+  const isAppIdKey = authenticationMethod === APP_ID_KEY_METHOD
+  return (
+    <FormFieldset id='fieldset-AuthenticationSettings'>
+      { (!isServiceMesh || isOidc) && <FormLegend>Authentication Settings</FormLegend> }
+      { !isServiceMesh && isApiKey && <FormCollection collection={[apiKeySettings]} ItemComponent={TextInputGroup} legend='API KEY (USER_KEY) BASICS' /> }
+      { !isServiceMesh && isAppIdKey && <FormCollection collection={appIdKeyPairSettings} ItemComponent={TextInputGroup} legend='APP_ID AND APP_KEY PAIR BASICS' /> }
+      { isOidc && <OidcFieldset {...oidcSettings} isServiceMesh={isServiceMesh} /> }
```
josemigallas: According with lines 37:40, is it possible to render an empty FormFieldSet? Namely when everything is false.

app/javascript/src/Settings/components/Common/FormCollection.jsx
```diff
@@ -0,0 +1,24 @@
...
+  collection: FormSettingsProps[],
+  ItemComponent: (FormSettingsProps) => React.Element<empty>,
+  legend: string
+}
+
+const FormCollection = ({collection, ItemComponent, legend}: Props) => {
+  return (
+    <FormFieldset id={`fieldset-${legend.replace(/\s+/g, '')}`}>
+      <FormLegend>{legend}</FormLegend>
+      { collection.map((itemProps, i) => <ItemComponent {...itemProps} key={i} />) }
```
josemigallas: We should avoid using index as keys:
```suggestion
      { collection.map(itemProps => <ItemComponent {...itemProps} key={itemProps.name} />) }
```

app/javascript/src/Settings/components/Common/RadioFieldset.jsx
```diff
@@ -0,0 +1,45 @@
...
+const RadioFieldset = (
+  {
+    children,
+    legend,
+    name,
+    value,
+    catalog,
+    onChange,
+    ...props
+  }: FormSettingsProps & FormCatalogProps) => {
```
josemigallas: I'd rather have a `type Props = FormSettingsProps & FormCatalogProps`.

app/javascript/src/Settings/components/Common/RadioFieldset.jsx
```diff
@@ -0,0 +1,45 @@
...
+
+import * as React from 'react'
+import { useState } from 'react'
+import { FormFieldset, FormLegend } from 'Form'
+import { Radio } from '@patternfly/react-core'
+import type { FormSettingsProps, FormCatalogProps } from 'Settings/types'
+
+type CheckEvent = SyntheticEvent<HTMLButtonElement>
+
+const RadioFieldset = (
```
josemigallas:
```suggestion
const RadioFieldset = ({
  children
  ...
```

app/javascript/src/Settings/components/Common/RadioFieldset.jsx
```diff
@@ -0,0 +1,45 @@
...
+  {
+    children,
+    legend,
+    name,
+    value,
+    catalog,
+    onChange,
+    ...props
+  }: FormSettingsProps & FormCatalogProps) => {
+  const [selected, onChangeFn] = typeof onChange === 'function'
```
josemigallas: We could extract this into a custom hook

app/javascript/src/Settings/components/Common/SelectGroup.jsx
```diff
@@ -0,0 +1,49 @@
+// @flow
+
+import * as React from 'react'
+import { useState } from 'react'
+import { FormGroup, Select, SelectOption } from '@patternfly/react-core'
+import type { FormSettingsProps, FormCatalogProps } from 'Settings/types'
+
+type Props = FormSettingsProps & FormCatalogProps
+
+const findInObject = (obj: { [key: string]: string }) => (key: string): string =>
```
josemigallas: I think this is too much, it is equivalent as doing `catalog[key]` and delegate the default value to the select.

app/javascript/src/Settings/components/Common/SelectGroup.jsx
```diff
@@ -0,0 +1,49 @@
...
+    <FormGroup label={label} fieldId={`service_proxy_attributes_${name}_select`} helperText={hint}>
+      <input name={`service[proxy_attributes][${name}]`} type='hidden' value={selectedValue} />
+      <Select
+        selections={catalog[selectedValue]}
+        onToggle={setIsExpanded}
+        onSelect={onSelect}
+        isExpanded={isExpanded}
+        id={`service_proxy_attributes_${name}_select`}
+      >
+        {Object.keys(catalog).map((key, i) => (
```
josemigallas: We should not use index as keys (specially when we have a proper key).

spec/javascripts/Settings/components/Common/__snapshots__/FormCollection.spec.jsx.snap
```diff
@@ -0,0 +1,31 @@
...
+exports[`should render correctly 1`] = `
+<FormFieldset
+  id="fieldset-SomeGoodBands"
+>
+  <FormLegend>
+    Some Good Bands
+  </FormLegend>
+  <GoodBand
+    key="0"
+    name="The Rolling Stones"
```
josemigallas:
```suggestion
    name="Justin Bieber"
```

app/javascript/src/Settings/components/Common/SelectGroup.jsx
```diff
@@ -0,0 +1,49 @@
...
+import { useState } from 'react'
+import { FormGroup, Select, SelectOption } from '@patternfly/react-core'
+import type { FormSettingsProps, FormCatalogProps } from 'Settings/types'
+
+type Props = FormSettingsProps & FormCatalogProps
+
+const findInObject = (obj: { [key: string]: string }) => (key: string): string =>
+  Object.keys(obj).find(k => obj[k] === key) || obj[Object.keys(obj)[0]]
+
+const SelectGroup = (
```
josemigallas:
```suggestion
const SelectGroup = ({
  label,
  ...
```

app/javascript/src/Settings/components/Common/TextInputGroup.jsx
```diff
@@ -0,0 +1,39 @@
...
+    placeholder,
+    label,
+    name,
+    hint,
+    value,
+    isDefaultValue = false,
+    readOnly = false,
+    inputType = 'text'
+  }: FormSettingsProps) => {
+  const [ inputValue, setValue ] = useState(value)
```
josemigallas:
```suggestion
  const [value, setValue] = useState(value)
```
or
```suggestion
  const [inputValue, setInputValue] = useState(value)
```

app/javascript/src/Settings/components/Common/TextInputGroup.jsx
```diff
@@ -0,0 +1,39 @@
...
+  const onChange = (value) => setValue(value)
```
josemigallas: I think this is equivalent:
```suggestion
  const onChange = setValue
```

didierofrivia: It would be, but it's coded that way to avoid React warning about hooks are not prepared for callbacks. This happens because of PF `OnChange` function signature, being the second argument the `event`. I'll make it more explicit

app/javascript/src/Settings/components/Common/TextInputGroup.jsx
josemigallas: @didierofrivia I've seem a couple of these in the PR, make this comment count for all of them :P your call to follow naming convention or not.

app/javascript/src/Settings/components/Common/SelectGroup.jsx
```diff
@@ -0,0 +1,45 @@
...
+      <input name={`service[proxy_attributes][${name}]`} type='hidden' value={selectedValue} />
+      <Select
+        selections={catalog[selectedValue]}
+        onToggle={setIsExpanded}
+        onSelect={onSelect}
+        isExpanded={isExpanded}
+        id={`service_proxy_attributes_${name}_select`}
+      >
+        {Object.keys(catalog).map(key => (
+          <SelectOption key={catalog[key]} value={{key, toString: toString(catalog[key])}} />
```
josemigallas: I reckon you're fond of function composition but I think sometimes it makes the code more convoluted than it should. IMO this one should be declared inline:
```suggestion
          <SelectOption key={catalog[key]} value={{ key, toString: catalog[key] }} />
```

app/javascript/src/Settings/components/Common/TextInputGroup.jsx
didierofrivia: You are more than right! :)

app/javascript/src/Settings/components/Common/SelectGroup.jsx
josemigallas: OK if the array option is not suitable, then I propose using underscore's `each` and `map` that can iterate over objects. They can be installed independently (rather than the whole library) and it would improve reading IMO

josemigallas: Or we create our own util method 🤷‍♂

didierofrivia: I don't think it's needed anymore, I got rid of that fn and made it easier to understand, I wasn't aware of passing an object with`toString` fn as a `value` in the past. wdyt?

app/javascript/src/Settings/types/index.jsx
```diff
@@ -0,0 +1,48 @@
...
+      flowSettings: FieldGroupProps[],
+      jwtSettings: TypeItemProps
+    }
+  },
+  credentialsLocation: FieldGroupProps & FieldCatalogProps,
+  security: FieldGroupProps[],
+  gatewayResponse: {
+    legend: string,
+    collection: FieldGroupProps[]
+  }[]
```
josemigallas: I'd create a new type to be clearer: `gatewayResponse: Thing[]`

app/javascript/src/Settings/index.jsx
```diff
@@ -0,0 +1,22 @@
+// @flow
+
+import * as React from 'react'
+import { render } from 'react-dom'
+import { Form } from 'Settings/components/Form'
+import { SETTINGS_DEFAULT } from 'Settings/defaults'
+import type { SettingsProps } from 'Settings/types'
+
+const initSettings = ({settings = SETTINGS_DEFAULT, elementId}: {settings: SettingsProps, elementId: string}) => {
```
josemigallas: Wrap this in `createReactWrapper` instead, it's the same behavior.

app/javascript/src/Settings/components/OidcFieldset.jsx
```diff
@@ -0,0 +1,53 @@
...
+  <TypeItemCombo {...props} legend='OIDC BASICS' inputType='url' />
+)
+
+const JsonWebToken = (props: TypeItemProps) => (
+  <TypeItemCombo {...props} legend='JSON Web Token (JWT) Claim with ClientID' inputType='text' />
+)
+
+const FlowItem = ({name, label, checked}: FieldGroupProps) => {
+  const [ isChecked, setIsChecked ] = useState(checked)
+  const onChange = (check) => setIsChecked(check)
```
josemigallas:
```suggestion
  const onChange = setIsChecked
```

didierofrivia: This is the same case as in https://github.com/3scale/porta/pull/1315#discussion_r340999672, look: https://www.patternfly.org/v4/documentation/react/components/checkbox/#checkbox-props. I'll make it explicit adding the second (not used) argument

app/javascript/src/Settings/components/Common/TypeItemCombo.jsx
```diff
@@ -0,0 +1,26 @@
...
+import { TextInputGroup, SelectGroup } from 'Settings/components/Common'
+import type { FieldGroupProps, FieldCatalogProps } from 'Settings/types'
+
+type Props = {
+  type: FieldCatalogProps & FieldGroupProps,
+  item: FieldGroupProps,
+  legend: string
+}
+
+const TypeItemCombo = ({ type, item, legend }: Props) => {
```
josemigallas: I'm not sure what this component represent, "type item combo"?

app/javascript/src/Settings/components/Common/TextInputGroup.jsx
```diff
@@ -0,0 +1,38 @@
...
+  label,
+  name,
+  hint,
+  value,
+  isDefaultValue = false,
+  readOnly = false,
+  inputType = 'text'
+}: FieldGroupProps) => {
+  const [ inputValue, setInputValue ] = useState(value)
+  const onChange = (value, _e) => setInputValue(value)
```
josemigallas:
```suggestion
  const onChange = setInputValue
```

didierofrivia: As I commented in https://github.com/3scale/porta/pull/1315#discussion_r340999672, I can't use point free notation here because of React annoying warning.

app/javascript/src/Settings/components/Common/SelectGroup.jsx
```diff
@@ -0,0 +1,44 @@
...
+    <FormGroup label={label} fieldId={`service_proxy_attributes_${name}_select`} helperText={hint}>
+      <input name={`service[proxy_attributes][${name}]`} type='hidden' value={selectedValue} />
+      <Select
+        selections={catalog[selectedValue]}
+        onToggle={setIsExpanded}
+        onSelect={onSelect}
+        isExpanded={isExpanded}
+        id={`service_proxy_attributes_${name}_select`}
+      >
+        {Object.keys(catalog).map(key => (
```
josemigallas: With underscore `map` it would look much better (IMO):
```suggestion
        {map(catalog, (value, key) => <SelectOption key={key} value={{ key, toString: () => value }} />
```

didierofrivia: I'd rather install Ramda, still is not worth it, it's used just in 2 places. I'd rather do it when we find ourselves using more than one fn.

app/javascript/src/Settings/components/Common/RadioFieldset.jsx
```diff
@@ -0,0 +1,50 @@
...
+import type { FieldGroupProps, FieldCatalogProps } from 'Settings/types'
+
+type CheckEvent = SyntheticEvent<HTMLButtonElement>
+
+type Props = FieldGroupProps & FieldCatalogProps
+
+const useSelectedOnChange = (value, onChange) => (
+  typeof onChange === 'function'
+    ? [value, onChange]
+    : useState(value).map(x => typeof x === 'function' ? (_c, e: CheckEvent) => x(e.currentTarget.value) : x)
```
josemigallas: I don't think it is needed to do these kind of _setState_ transformations since `_c === e.currentTarget.value`. This PF's way to avoid us digging into events.

app/javascript/src/Settings/components/OidcFieldset.jsx
josemigallas: Sorry, I'm repeating myself, forgot about https://github.com/3scale/porta/pull/1315#discussion_r340999672

app/javascript/src/Settings/components/Common/SelectGroup.jsx
josemigallas: I don't know Ramda but the advantage of underscore is we can install just 1 method, which is almos the same as adding our own helper function.

I still rather have this extracted into a function, even if it's used in 2 places. Also not super important either :P

spec/javascripts/Settings/components/Common/__snapshots__/FormCollection.spec.jsx.snap
guicassolato: Can we please keep it The Rolling Stones?

app/javascript/src/Settings/components/Common/SelectGroup.jsx
didierofrivia: I mean, it's not like we are applying hardcore hacky functions in each place... we are just mapping some Object keys, using built-in prototype, not quite a big deal :)
It was applicable to  https://github.com/3scale/porta/pull/1315#discussion_r341023581, but since it was simplified, I don't think it's worth to install a new dependency or create an util fn just for this. And I know that we could just import the fn, and that we will most probably end up installing either underscore or Ramda... but I don't believe that line justify making it part of this PR.

app/javascript/src/Settings/defaults.jsx
```diff
@@ -0,0 +1,263 @@
+const INTEGRATION_METHOD_DEFAULTS = {
+  value: 'hosted',
+  name: 'deployment_option',
+  catalog: {
+    hosted: 'APIcast a 3scale managed',
```
damianpm: Typo ? `'APIcast 3scale managed'`

app/javascript/src/Settings/components/Common/SelectGroup.jsx
damianpm: You can import only `map` but anyway you need to add a dependency, I wouldn't do it, at least not here.

PR #1325, opened 2019-10-14
Inject `owner_id` into the routing policy rules

guicassolato:
`owner_id` is the id of the backend api of the backend api config behind the policy rule.

This will allow the gateway to be match only proxy rules that belong to the same backend as the request will end up been proxied to.

Proxy rules already include enough information for the match. Whenever a proxy rule belongs to a Backend API (as opposed to belonging to the Product), the gateway will find:

```json
"proxy_rules": [
  { ...
    "owner_id" : "<id-of-the-backend>",
    "owner_type" : "BackendApi"
  }
]
```

With this PR, the routing policy injected for the backend apis will also include:

```json
"rules": [
  { ...
    "owner_id" : "<id-of-the-backend>"
  }
]
```

----

Related to [THREESCALE-3623](https://issues.jboss.org/browse/THREESCALE-3623)
Related to https://github.com/3scale/APIcast/pull/1125

### DIFF
--- a/app/lib/backend_api_logic/routing_policy.rb
+++ b/app/lib/backend_api_logic/routing_policy.rb
@@ -37,7 +37,7 @@ module BackendApiLogic
       end
 
       class Rule
-        delegate :private_endpoint, :path, to: :@config
+        delegate :private_endpoint, :path, :backend_api_id, to: :@config
 
         def initialize(config)
           @config = config
@@ -51,6 +51,8 @@ module BackendApiLogic
           return if private_endpoint.blank?
           {
             url: private_endpoint,
+            owner_id: backend_api_id,
+            owner_type: BackendApi.name,
             condition: {
               operations: [
                 match: :path,
--- a/test/unit/backend_api_logic/routing_policy_test.rb
+++ b/test/unit/backend_api_logic/routing_policy_test.rb
@@ -22,9 +22,11 @@ module BackendApiLogic
     end
 
     test '#policy_chain' do
+      backend_api1 = backend_apis.first
+      backend_api2 = backend_apis.last
       injected_rules = [
-        { url: backend_apis.last.private_endpoint,  condition: { operations: [match: :path, op: :matches, value: '/foo/.*|/foo/?'] }, replace_path: "{{original_request.path | remove_first: '/foo'}}" },
-        { url: backend_apis.first.private_endpoint, condition: { operations: [match: :path, op: :matches, value: '/.*'] } }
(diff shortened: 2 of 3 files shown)


app/lib/backend_api_logic/routing_policy.rb
```diff
@@ -47,6 +47,7 @@ def as_json
           return if private_endpoint.blank?
           {
             url: private_endpoint,
+            owner_id: backend_api_id,
```
hallelujah: Well I really prefer not using owner_id
We are supposed to be compayible with any other gateway.

The `owner_type` and `owner_id` are internal to Porta
Unless you want them to be part of a specification of Gateway configuration

guicassolato: I presented the options to apicast team. They say adding `owner_id` to the policy rules should be enough.

I don't want to make it such big of a deal. For now, moving on with `owner_id` and if they change their mind and prefer something else, not hard to change anyway.

hallelujah: Yes I understand it is not a big deal but it is just that as specification it is not really clean and once it is out it is hard to go back. Sure it is just a word but it can make a difference.
Up to you

guicassolato: Let's see how it goes: https://github.com/3scale/APIcast/pull/1125#discussion_r334571141

Martouta: I agree with your comment there @guicassolato , imo it should be either `backend_api_id` or `owner_id` + `owner_type`, but only `owner_id` is a bit asking for problems :smile:

hallelujah: @guicassolato It would be nice to add `owner_type` and done with this (aka can be merged ;) )

guicassolato:
It's led by @3scale/apicast. They would need to change first and we would follow. An eventual migration would require `owner_id` to be supported simultaneously to whatever replaces it for a while, then removed for good when no existing config depended upon it any longer.

app/lib/backend_api_logic/routing_policy.rb
guicassolato: It's apicast that needs to add `owner_type`, not us, and they did.

guicassolato: Oh. I get it now. For semantics you mean. Only to make it clear who the "owner" is. Ok I will add it.

guicassolato: Done.

PR #1326, opened 2019-10-14
[DO NOT MERGE] Prevents deployment when updating Proxy in APIAP

josemigallas:

Adds a `apiap?` condition when updating so that it isn't deployed. This way there is no proxy_config created.

**JIRA**:

[THREESCALE-3626: Proxy configs for products that do not have any backend (API)](https://issues.jboss.org/browse/THREESCALE-3626)

### DIFF
--- a/app/controllers/admin/api/services/proxies_controller.rb
+++ b/app/controllers/admin/api/services/proxies_controller.rb
@@ -63,7 +63,7 @@ class Admin::Api::Services::ProxiesController < Admin::Api::Services::BaseContro
   ##~ op.parameters.add name: "jwt_claim_with_client_id_type", description: "JWT Claim With ClientId Type. Either `plain` or `liquid`", dataType: "string", paramType: "query", required: false
   #
   def update
-    if proxy.update_attributes(proxy_params)
+    if proxy.update_attributes(proxy_params) && !apiap?
       if proxy.service_mesh_integration?
         proxy.deploy!
       elsif proxy.apicast_configuration_driven
--- a/app/controllers/admin/api/services/proxies_controller.rb
+++ b/app/controllers/admin/api/services/proxies_controller.rb
@@ -33,7 +33,7 @@ class Admin::Api::Services::ProxiesController < Admin::Api::Services::BaseContro
   ##~ op            = e.operations.add
   ##~ op.httpMethod = "PATCH"
   ##~ op.summary    = "Proxy Update"
-  ##~ op.description = "Changes the Proxy settings. This will create a new APIcast configuration version for the Staging environment with the updated settings."
+  ##~ op.description = "Changes the Proxy settings. This will create a new APIcast configuration version for the Staging environment with the updated settings. For servcice mesh, it will also create one for the Production environment"
   ##~ op.group = "proxy"
   #
   ##~ op.parameters.add @parameter_access_token
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -296,4 +296,8 @@ class ApplicationController < ActionController::Base
       res
     end
   end
+
+  def apiap?
+    current_account.provider_can_use?(:api_as_product)
+  end
 end
--- a/app/controllers/frontend_controller.rb
+++ b/app/controllers/frontend_controller.rb
@@ -29,10 +29,6 @@ class FrontendController < ApplicationController
 
   private
 
-  def apiap?
-    current_account.provider_can_use?(:api_as_product)
-  end
-
   helper_method :apiap?
 
   def do_nothing_if_head
--- a/test/integration/user-management-api/services/proxies_test.rb
+++ b/test/integration/user-management-api/services/proxies_test.rb
@@ -49,15 +49,13 @@ class Admin::Api::Services::ProxiesTest < ActionDispatch::IntegrationTest
     params = provider_key_params.merge(proxy: { endpoint: 'https://alaska.wild' })
 
     Proxy.update_all(apicast_configuration_driven: false)
-    Proxy.any_instance.expects(:deploy_v2).never
-    assert_difference @service.proxy.proxy_configs.production.method(:count), 0 do
+    assert_no_difference @service.proxy.proxy_configs.sandbox.method(:count) do
       put(admin_api_service_proxy_path(params))
     end
     assert_response :success
 
     Proxy.update_all(apicast_configuration_driven: true)
-    Proxy.any_instance.expects(:deploy_v2).once
-    assert_difference @service.proxy.proxy_configs.production.method(:count), 1 do
+    assert_difference @service.proxy.proxy_configs.sandbox.method(:count), 1 do
       put(admin_api_service_proxy_path(params))
     end
     assert_response :success
@@ -70,15 +68,13 @@ class Admin::Api::Services::ProxiesTest < ActionDispatch::IntegrationTest
     params = provider_key_params.merge(proxy: { endpoint: 'https://alaska.wild' })
 
     Proxy.update_all(apicast_configuration_driven: false)
-    Proxy.any_instance.expects(:deploy_v2).never
-    assert_difference @service.proxy.proxy_configs.production.method(:count), 0 do
+    assert_no_difference @service.proxy.proxy_configs.sandbox.method(:count) do
       put(admin_api_service_proxy_path(params))
     end
     assert_response :success
 
     Proxy.update_all(apicast_configuration_driven: true)
-    Proxy.any_instance.expects(:deploy_v2).once
-    assert_difference @service.proxy.proxy_configs.production.method(:count), 0 do
+    assert_no_difference @service.proxy.proxy_configs.sandbox.method(:count) do
       put(admin_api_service_proxy_path(params))
     end
     assert_response :success
@@ -87,16 +83,18 @@ class Admin::Api::Services::ProxiesTest < ActionDispatch::IntegrationTest
   def test_update_for_service_mesh
     rolling_updates_on
     rolling_update(:api_as_product, enabled: false)
-    
+
     params = provider_key_params.merge(proxy: { credentials_location: 'headers' })
     @service.update_column :deployment_option, 'service_mesh_istio'
-    
-    assert_difference @service.proxy.proxy_configs.production.method(:count), 1 do
-      put(admin_api_service_proxy_path(params))
+
+    assert_difference @service.proxy.proxy_configs.sandbox.method(:count), 1 do
+      assert_difference @service.proxy.proxy_configs.production.method(:count), 1 do
+        put(admin_api_service_proxy_path(params))
+      end
     end
     assert_response :success
   end
-  
+
   def test_update_for_service_mesh_with_apiap
     rolling_updates_on
     rolling_update(:api_as_product, enabled: true)
@@ -104,8 +102,10 @@ class Admin::Api::Services::ProxiesTest < ActionDispatch::IntegrationTest
     params = provider_key_params.merge(proxy: { credentials_location: 'headers' })
     @service.update_column :deployment_option, 'service_mesh_istio'
 
-    assert_difference @service.proxy.proxy_configs.production.method(:count), 0 do
-      put(admin_api_service_proxy_path(params))
+    assert_no_difference @service.proxy.proxy_configs.sandbox.method(:count) do
+      assert_no_difference @service.proxy.proxy_configs.production.method(:count) do
+        put(admin_api_service_proxy_path(params))
+      end
     end
     assert_response :success
   end
--- a/test/integration/user-management-api/services/proxies_test.rb
+++ b/test/integration/user-management-api/services/proxies_test.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 require 'test_helper'
 
 class Admin::Api::Services::ProxiesTest < ActionDispatch::IntegrationTest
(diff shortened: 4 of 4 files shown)


app/controllers/admin/api/services/proxies_controller.rb
```diff
@@ -63,13 +63,7 @@ def show
...
   #
   def update
-    if proxy.update_attributes(proxy_params)
-      if proxy.service_mesh_integration?
-        proxy.deploy!
-      elsif proxy.apicast_configuration_driven
-        proxy.deploy_v2
-      end
-    end
+    proxy.deploy! if proxy.update_attributes(proxy_params) && !apiap?
```
guicassolato: Why changing the logic here? Calling just `proxy.deploy!` is not the same as we had before. Now it will also check `Proxy#deployable?` and potentially call `Proxy#deploy_v1`.

app/controllers/frontend_controller.rb
```diff
@@ -67,21 +63,21 @@ def session_is_secure?
   helper_method :sudo
 
   def sudo
-    return_path = request.xhr? ? request.fullpath : request.headers.fetch('Referer') { request.fullpath }
+    return_path = request.xhr? ? request.fullpath : request.headers.fetch("Referer") { request.fullpath }
```
guicassolato: Single quotes and double quotes do not mean the same in Ruby. We should keep the single quotes in this case.

```suggestion
    return_path = request.xhr? ? request.fullpath : request.headers.fetch('Referer') { request.fullpath }
```

app/controllers/frontend_controller.rb
```diff
@@ -67,21 +63,21 @@ def session_is_secure?
...
     Sudo.new(return_path: return_path, user_session: user_session, xhr: request.xhr?)
   end
 
   def ask_for_secure_session!
     respond_to do |format|
-      format.html { render 'provider/admin/sudo/show' }
+      format.html { render "provider/admin/sudo/show" }
```
guicassolato:
```suggestion
      format.html { render 'provider/admin/sudo/show' }
```

app/controllers/frontend_controller.rb
```diff
@@ -108,25 +104,25 @@ def analytics
   #
   def pick_buyer_or_provider_layout
     if site_account.master?
-      current_account ? 'provider' : 'provider/login'
+      current_account ? "provider" : "provider/login"
```
guicassolato:
```suggestion
      current_account ? 'provider' : 'provider/login'
```

app/controllers/frontend_controller.rb
```diff
@@ -108,25 +104,25 @@ def analytics
...
+                 site_account.accessible_services
+               end
 
     @service = id.present? ? services.find(id) : services.default
   end
 
   def ensure_provider
     unless current_account.provider?
-      render_error 'Not authorized', :status => :not_authorized
+      render_error "Not authorized", :status => :not_authorized
```
guicassolato:
```suggestion
      render_error 'Not authorized', :status => :not_authorized
```

test/integration/user-management-api/services/proxies_test.rb
```diff
@@ -65,6 +65,28 @@ def test_update_for_service_mesh
     assert_response :success
   end
 
+  def test_update_should_deploy
+    Account.any_instance.stubs(:provider_can_use?).with(:api_as_product).returns(false)
+    params = provider_key_params.merge(proxy: { endpoint: 'https://alaska.wild' })
+    
+    Proxy.update_all(apicast_configuration_driven: false)
```
guicassolato: Actually we want to test `apicast_configuration_driven == true` here

```suggestion
```

guicassolato: ...and also when it's `service_mesh_integration?`

test/integration/user-management-api/services/proxies_test.rb
```diff
@@ -65,6 +65,28 @@ def test_update_for_service_mesh
...
+    put(admin_api_service_proxy_path(params))
+    assert_response :success
+  end
+  
+  def test_update_for_apiap_should_not_deploy
+    Account.any_instance.stubs(:provider_can_use?).with(:api_as_product).returns(true)
+    params = provider_key_params.merge(proxy: { endpoint: 'https://alaska.wild' })
+
+    Proxy.update_all(apicast_configuration_driven: false)
+    Proxy.any_instance.expects(:deploy!).never
```
guicassolato: Better testing that no new proxy config was created. Asserting that a method has never been called is too error-prone.

app/controllers/frontend_controller.rb
Martouta: Since we are refactoring this anyway:
```ruby
def ensure_provider
  return true if current_account.provider?
  render_error 'Not authorized', status: :not_authorized
  false
end
```

app/controllers/frontend_controller.rb
josemigallas: Spending hours to config a good Ruby formatter in VSCode for this.. darn it! I'll fix all the quotes, not a voluntary change anyway.

test/integration/user-management-api/services/proxies_test.rb
josemigallas: OK, makes sense.

app/controllers/admin/api/services/proxies_controller.rb
josemigallas: There is definitely some duplicated and odd behavior here, although I refactored it wrong.

For example when service mesh integration, the same condition is run 3 times:
```ruby
if proxy.service_mesh_integration?         # true (proxies_controller.rb)
  if ([...] || service_mesh_integration?)  # again, true (proxy.rb)
    if service_mesh_integration?           # 3rd time, true (proxy.rb)
      deploy_service_mesh_integration
```
In case of apicast, it bypasses `deployable?` when run from the controller, but it doesn't when run from the model. That's inconsistent, isn't it? Could you clarify this to me?

Finally for `deploy_v1` you're right.

josemigallas: I'm leaving everything like it is here and proposing a simplification in #1327

test/integration/user-management-api/services/proxies_test.rb
```diff
@@ -41,25 +43,64 @@ def test_crud_provider_key
...
+    params = provider_key_params.merge(proxy: { endpoint: 'https://alaska.wild' })
+
+    Proxy.update_all(apicast_configuration_driven: false)
+    assert_difference @service.proxy.proxy_configs.production.method(:count), 0 do
+      put(admin_api_service_proxy_path(params))
+    end
+    assert_response :success
+
+    Proxy.update_all(apicast_configuration_driven: true)
+    assert_difference @service.proxy.proxy_configs.production.method(:count), 0 do
```
Martouta: Right now it is `sandbox` and not `production`

josemigallas: Only here?

Martouta: i think in the whole test @josemigallas . What I don't know if the environment is a param that can be sent and then the sandbox is just the default :thinking:

josemigallas: Changing everything to `sandbox` and also removing all of them make the test pass, I guess that's the answer?

josemigallas: I'm actually wrong, removing it from everywhere makes some fail. I'll change it to sandbox.

Martouta: i don't know @josemigallas , sorry :pray: I can check more in depth after the afternoon sync :smile:

Martouta: I see that you already resolved it :muscle: for anybody else who reads this, the key is in the description of the endpoint:
https://github.com/3scale/porta/blob/22b1411b8eda4b4d1e54818d89db209f39c9dbf6/app/controllers/admin/api/services/proxies_controller.rb#L36

app/controllers/admin/api/services/proxies_controller.rb
```diff
@@ -63,7 +63,7 @@ def show
   ##~ op.parameters.add name: "jwt_claim_with_client_id_type", description: "JWT Claim With ClientId Type. Either `plain` or `liquid`", dataType: "string", paramType: "query", required: false
   #
   def update
-    if proxy.update_attributes(proxy_params)
+    if proxy.update_attributes(proxy_params) && !apiap?
```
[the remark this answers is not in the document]
josemigallas: @guicassolato will correct me if I'm wrong:
The idea is simply that customer using APIAP will have to deploy manually, that's all. I guess your concern is that this is a quick-workaround rather than a solution. To me, it makes sense to merge this solution temporarily because I understand this is bug in production... and in the meantime we'd work on a more robust one.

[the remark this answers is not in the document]
Martouta: @hallelujah the goal of the issue is not to automatically deploy when the proxy is updated, yes, but this is just making for the API side what we are already doing for the UI side :smile:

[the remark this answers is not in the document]
hallelujah: Ah one other strange thing that we get rid off :) 👍 
Though we need to put it in the release note that this is a breaking change.

Also we need to think about what to do next when we clean up this mess and everybody will be under apiap

test/integration/user-management-api/services/proxies_test.rb
```diff
@@ -41,25 +43,64 @@ def test_crud_provider_key
...
+  
+  def test_update_for_service_mesh_with_apiap
+    rolling_updates_on
+    rolling_update(:api_as_product, enabled: true)
+
     params = provider_key_params.merge(proxy: { credentials_location: 'headers' })
     @service.update_column :deployment_option, 'service_mesh_istio'
 
-    assert_difference @service.proxy.proxy_configs.production.method(:count), +1 do
+    assert_difference @service.proxy.proxy_configs.sandbox.method(:count), 0 do
```
hallelujah: `assert_no_change`

test/integration/user-management-api/services/proxies_test.rb
```diff
@@ -41,25 +43,64 @@ def test_crud_provider_key
...
   end
 
   def test_update_for_service_mesh
     rolling_updates_on
+    rolling_update(:api_as_product, enabled: false)
+    
+    params = provider_key_params.merge(proxy: { credentials_location: 'headers' })
+    @service.update_column :deployment_option, 'service_mesh_istio'
+    
+    assert_difference @service.proxy.proxy_configs.sandbox.method(:count), 1 do
```
hallelujah: For service mesh we expect the production to change too :)

test/integration/user-management-api/services/proxies_test.rb
Martouta: or `assert_no_difference` :smile:

test/integration/user-management-api/services/proxies_test.rb
Martouta: @hallelujah Then imo the ActiveDocs description is not entirely related with what it should do:
https://github.com/3scale/porta/blob/22b1411b8eda4b4d1e54818d89db209f39c9dbf6/app/controllers/admin/api/services/proxies_controller.rb#L36

josemigallas: What should it say?

app/controllers/application_controller.rb
```diff
@@ -296,4 +296,8 @@ def safe_return_to(url)
       res
     end
   end
+
+  def apiap?
+    current_account.provider_can_use?(:api_as_product)
```
duduribeiro: this method is already defined in `app/controllers/frontend_controller.rb`, should we redefine here again?

josemigallas: No, it is not anymore 😉
https://github.com/3scale/porta/pull/1326/files#diff-1862ca4b36d9904f13fb697bb4daab9eL32-L35

test/integration/user-management-api/services/proxies_test.rb
hallelujah: For servcice mesh, it will also create a new APIcast configuration for production

app/controllers/application_controller.rb
duduribeiro: 👍

app/controllers/admin/api/services/proxies_controller.rb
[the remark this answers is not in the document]
hallelujah: Still that means that there is no way to create a proxy config as this controller does not have any `create` method

https://github.com/3scale/porta/blob/ecd417bbcbbf96c83d96e99f61902d6f143f3c8b/app/controllers/admin/api/services/proxy/configs_controller.rb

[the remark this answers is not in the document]
hallelujah: So lets not merge it for 2.7 because this is a breaking change without any deprecation notice.

Customers should not carry our mistakes, so steps to achieve it:

- Deprecate the behaviour (creation of proxy onfig on proxy update) in 2.7
- Add another entry point `#create` in proxy configs or a deploy method in `proxies` controller
- remove the behaviour in 2.8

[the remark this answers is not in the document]
josemigallas: We could also fix the bug instead of changing the behavior. I understand this is a quick solution but it would also be possible to keep the logic as it is *but* preventing the creation of a proxy config when there are no backends.

@guicassolato @hallelujah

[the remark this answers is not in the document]
josemigallas: >Still that means that there is no way to create a proxy config as this controller does not have any create method

I'm not sure I follow, why would we need a create method? Proxy configs are created automatically when deploying and deploying is still possible, right? The only change is that users will have to do it manually.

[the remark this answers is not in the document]
hallelujah: This is all explained in this issue https://issues.jboss.org/browse/THREESCALE-3758 @josemigallas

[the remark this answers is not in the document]
guicassolato: @hallelujah, sorry for taking so long to get back to this. I think this PR is fine and we should move on with it.

This line of code that triggered the discussion means that, by using the **Proxy Update** endpoint, APIAP accounts will be able to save the proxy but this won't deploy the new config to staging anymore. In order to "promote" the new config to staging we already have another endpoint, which is the **Proxy Config Promote**. (Yes, it should be called **Proxy Config Create**!) It accepts an "environment" parameter, which, by passing "sandbox", promotes the latest proxy configs to staging, no?

For non-APIAP accounts, the PR keeps the same behavior.

hallelujah:
@guicassolato this is fine, we have already the deploy endpoint

https://github.com/3scale/porta/blob/166536429a54f568ab808afe1d4f3673c988307c/app/controllers/admin/api/services/proxies_controller.rb#L89-L96

PR #1327, opened 2019-10-15
[WIP] Refactor / Simplify deployment logic in Proxy Config

josemigallas:
https://issues.jboss.org/browse/THREESCALE-3751

There is some repetition in `Proxy#deploy!` and `ProxiesController#update` that could be simplified. And the proxy configs deployment logic should be streamlined, if possible.

This PR only reduces duplication, hoping that logic doesn't change.

However, I still don't understand the difference between `deploy_v1`, `deploy_v2` and `deploy_production_v2`.

Also I run the formatter in the affected files.

### DIFF
--- a/app/controllers/admin/api/services/proxies_controller.rb
+++ b/app/controllers/admin/api/services/proxies_controller.rb
@@ -63,13 +63,7 @@ class Admin::Api::Services::ProxiesController < Admin::Api::Services::BaseContro
   ##~ op.parameters.add name: "jwt_claim_with_client_id_type", description: "JWT Claim With ClientId Type. Either `plain` or `liquid`", dataType: "string", paramType: "query", required: false
   #
   def update
-    if proxy.update_attributes(proxy_params)
-      if proxy.service_mesh_integration?
-        proxy.deploy!
-      elsif proxy.apicast_configuration_driven
-        proxy.deploy_v2
-      end
-    end
+    proxy.deploy_mesh_or_apicast if proxy.update_attributes(proxy_params)
 
     respond_with(proxy)
   end
--- a/app/models/proxy.rb
+++ b/app/models/proxy.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 require 'ipaddr'
 require 'resolv'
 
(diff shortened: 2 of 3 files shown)


app/models/proxy.rb
```diff
@@ -347,16 +348,24 @@ def deploy!
     deploy
   end
 
+  def deploy_mesh_or_apicast
```
hallelujah: I'd do another complete object for this logic, and use it here.
That way we can isolate the "deploy" logic with the update_attributes one

josemigallas: By object do you mean a new class? Where would it be located, rails-ly speaking?

app/models/proxy.rb
```diff
@@ -601,13 +605,7 @@ def create_default_proxy_rule
...
     return true unless deployable?
 
-    if service_mesh_integration?
-      deploy_service_mesh_integration
-    elsif apicast_configuration_driven
-      deploy_v2
-    else
-      deploy_v1
-    end
+    deploy_mesh_or_apicast || deploy_v1
```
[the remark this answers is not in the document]
josemigallas: That's correct. I believe that is the current behavior:
`deploy_mesh_or_apicast` is equivalent to the old 604-608 lines, that means `deploy_v1` is called when `service_mesh_integration? == false` and `apicast_configuration_driven == false`.
Of course that means the method must return a truthy value when either of them are true but AFAIK that condition is met (or at least tests confirm this).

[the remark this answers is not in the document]
josemigallas: OK I see what you mean now

app/models/proxy.rb
```diff
@@ -408,21 +303,21 @@ def send_api_test_request!
...
-                        status: result.code
+                      success: success,
+                      uri: result.uri.to_s,
+                      status: result.code
                     )
 
     success
   end
 
   def enabled
```
Martouta: I dunno if this suggestion is easy or not, but since it is expecting a boolean, the method should be called `enabled?`

hallelujah: Better live that for later

app/controllers/admin/api/services/proxies_controller.rb
```diff
@@ -63,13 +63,7 @@ def show
...
   #
   def update
-    if proxy.update_attributes(proxy_params)
-      if proxy.service_mesh_integration?
-        proxy.deploy!
-      elsif proxy.apicast_configuration_driven
-        proxy.deploy_v2
-      end
-    end
+    proxy.deploy_mesh_or_apicast if proxy.update_attributes(proxy_params)
```
hallelujah: There is no method `deploy_mesh_or_apicast`

Martouta: Right :joy: Good catch :+1:

josemigallas:
Closing in favor of #1470

PR #1331, opened 2019-10-15
[DO NOT MERGE - JUST A POC] Add API CRUD services/:service_id/backend_apis

Martouta:
The current master version :smile: 

The biggest problem of this is that we are calling backeng api to what it isn't.


guicassolato:
We're going with the https://github.com/3scale/porta/pull/1330 alternative.

PR #1332, opened 2019-10-15
Simplify ModelsTest. It needs to be dynamic and agnostic to the models we have.

Martouta:
Just simpler :smile: No need to be specific, it should do it the regular way anyway

### DIFF
--- a/app/lib/backend_api_logic/proxy_extension.rb
+++ b/app/lib/backend_api_logic/proxy_extension.rb
@@ -20,6 +20,7 @@ module BackendApiLogic
     protected
 
     def validate_backend_api?
+      return unless account # TODO: remove this line when we remove the Rolling Update :)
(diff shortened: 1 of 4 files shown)


app/lib/backend_api_logic/proxy_extension.rb
```diff
@@ -20,6 +20,7 @@ module ProxyExtension
     protected
 
     def validate_backend_api?
+      return unless account # TODO: remove this line when we remove the Rolling Update :)
```
hallelujah: Modifying the models and logic only for tests to pass is calling for huge problems.
Tests are not agnostics, they precisely need to know what are the objects they are testing.

You can make a test more generic but if you can't then ... too bad you can't

Martouta:
Ok, when we remove the rolling update, we won't need to have the nil check, so i will repush at that moment :+1:

PR #1340, opened 2019-10-17
Load Policies widget from a pack (instead of from the template)

josemigallas:

Move all initialization logic from the template (slim) into a pack. Append all necessary data in the HTML as datasets.

**Verification steps**:

**Note**:
This is part 2 of the refactorization of Policies:

Previous step #1338.

A follow-up PR will be created once this one is merged.


app/javascript/packs/policies.js
```diff
@@ -1,5 +1,19 @@
-import initPolicies from 'Policies/index'
+// @flow
+
+import { initPolicies as PoliciesWidget } from 'Policies'
 
 document.addEventListener('DOMContentLoaded', () => {
-  window.initPolicies = initPolicies
+  const policiesContainer = document.getElementById('policies')
+
+  if (!policiesContainer) {
```
didierofrivia: Better something like...
```js
if (!policiesContainer) throw TypeError('Policies Widget needs a valid DOM Element to render')
```

josemigallas: Changed it. Anyways, in a future commit I'm wrapping it inside a `ReactWrapper` so it's gonna throw this error.

PR #1342, opened 2019-10-17
Fix broken integration page links when apiap is on

damianpm:

In API overview page, links to Integration page `/apiconfig/services/<id>/integration/edit` should not shown when APIAP is on.

https://issues.jboss.org/browse/THREESCALE-3705

**APIAP ON:**
apiap-on

**APIAP OFF:**
apiap-off

### DIFF
--- a/app/views/api/services/show.html.slim
+++ b/app/views/api/services/show.html.slim
@@ -122,16 +122,18 @@ section class="service-widget"
           - if service.deployment_option.present?
             - deployment_option_phrased = t(service.proxy.deployment_option, scope: "deployment_options.phrased")
           - unless service.has_traffic?
-            li.item.integration
+            li class=("item integration #{'is-hidden' if apiap?}")
               - case service.proxy.deployment_option.presence
               - when 'hosted'
-                p On the #{link_to( 'Integration page', edit_admin_service_integration_path(service, :anchor => 'staging') )}, add your <em>API base URL</em> to the <em>Private Base URL</em> field in the staging box and hit <strong>Update & Test</strong>. Once the staging box is green, hit <strong>Deploy</strong> in the production box and you have completed a basic integration.
-                = important_icon_link "Configure #{deployment_option_phrased}",
-                  'hdd-o', edit_admin_service_integration_path(service, :anchor => 'staging'), :class => "outline-button next"
+                - unless apiap?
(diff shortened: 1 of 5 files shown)


app/views/api/services/show.html.slim
```diff
@@ -122,16 +122,18 @@ section class="service-widget"
           - if service.deployment_option.present?
             - deployment_option_phrased = t(service.proxy.deployment_option, scope: "deployment_options.phrased")
           - unless service.has_traffic?
-            li.item.integration
+            li class=("item integration #{'is-hidden' if apiap?}")
```
thomasmaas: we should remove it from the html if apiap, not hide it with css

damianpm: The paragraph is not rendered when apiap, check here
https://github.com/3scale/porta/pull/1342/files#diff-2eaca81d1f3c950ba0e8e12823649c5dR128

This class is to not show an empty `li`

thomasmaas: you're right, i was too quick, ignore!

app/views/provider/admin/dashboards/widgets/_service_top_traffic.html.slim
```diff
@@ -23,10 +23,11 @@ article.DashboardWidget id = widget.id
     - else
       p
         ' In order to show Top Applications you need to have at least one application sending traffic to the #{friendly_service_name(service)}.
-      p
-        ' Consider
-        => link_to 'making some test calls', edit_admin_service_integration_path(widget_service_id, :anchor => 'staging')
-        | from the Integration page to get a feel for what you'd see here.
+      - unless apiap?
```
thomasmaas: so instead of hiding everything the link should be changed

damianpm: Pointing to `apiconfig/services/2/integration` ?

thomasmaas: `admin_service_integration_path(widget_service_id)`

thomasmaas: (we should also tell them to promote config to staging if it has not been done but lets leave that for a later iteration)

PR #1343, opened 2019-10-17
Give Master Admin permission to see Backend API

josemigallas:

Adds permissions for master tenant to see Backend APIs.

[THREESCALE-3725: In master tenant, master backend service overview page shows access denied](https://issues.jboss.org/browse/THREESCALE-3725)

**Verification steps** 

1. Open the master tenant
2. Go to "Master Service Backend API" (from the context selector)

Some permissions might be still missing.

### DIFF
--- a/config/abilities/master_admin.rb
+++ b/config/abilities/master_admin.rb
@@ -48,6 +48,11 @@ Ability.define do |user|
     can :manage, :authentication_providers
     can :manage, :web_hooks
 
+    if user.account.provider_can_use?(:api_as_product)
+      can :manage, BackendApi
+      can :manage, BackendApiConfig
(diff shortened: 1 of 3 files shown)


config/abilities/master_admin.rb
```diff
@@ -48,6 +48,11 @@
     can :manage, :authentication_providers
     can :manage, :web_hooks
 
+    if user.account.provider_can_use?(:api_as_product)
```
thomasmaas: is this valid given master is not provider?

josemigallas: As confirmed by @Martouta, provider means both master and tenants so yes it is.

PR #1344, opened 2019-10-17
Fix mapping rules link in Methods and Metrics page

duduribeiro:

Currently our Methods and Metrics page have a link to mapping rules
link, but it is linking only to the mapping rules inside the integration
pages and it is not considering the independent mapping rules.

This commit fixes this behave by considering the independent mapping
rules to create the link.

[THREESCALE-3719](https://issues.jboss.org/browse/THREESCALE-3719)

**Verification steps** 

1. Enable APIAP
2. Enter in `/apiconfig/services/SERVICE_ID/metrics` page
3. The link should redirect to the independent mapping rules page: `/apiconfig/services/SERVICE_ID/proxy_rules`
4. Disable APIAP
5. Enter in `/apiconfig/services/SERVICE_ID/metrics` page
6. The link should redirect to the integration page: `apiconfig/services/SERVICE_ID/integration/edit`

### DIFF
--- a/app/helpers/api/mapping_rules_helper.rb
+++ b/app/helpers/api/mapping_rules_helper.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+module Api::MappingRulesHelper
+  def link_to_mapping_rules(service:)
+    path, data = if current_account.independent_mapping_rules_enabled?
+                   [admin_service_proxy_rules_path(service), {}]
+                 else
+                   [edit_admin_service_integration_path(service, anchor: 'mapping-rules'), { behavior: 'open-mapping-rules' }]
+                 end
+
+    link_to 'Mapping Rules', path, data: data
+  end
+end
--- a/app/views/api/metrics/index.html.slim
+++ b/app/views/api/metrics/index.html.slim
@@ -13,7 +13,7 @@ section.Section.u-toggleableBySettingsBox data-state="open"
     ' .
     - if show_mappings?
       ' A method needs to be mapped to one or more URL patterns in the
-      => link_to 'Mapping Rules', edit_admin_service_integration_path(@service, anchor: 'mapping-rules'), data: { behavior: 'open-mapping-rules' }
+      => link_to_mapping_rules(service: @service)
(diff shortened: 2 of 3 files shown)


app/views/api/metrics/index.html.slim
```diff
@@ -13,7 +13,7 @@ section.Section.u-toggleableBySettingsBox data-state="open"
     ' .
     - if show_mappings?
       ' A method needs to be mapped to one or more URL patterns in the
-      => link_to 'Mapping Rules', edit_admin_service_integration_path(@service, anchor: 'mapping-rules'), data: { behavior: 'open-mapping-rules' }
+      => link_to_mapping_rules(service: @service)
```
thomasmaas: Isn't this link also used when independent_mapping_rules are disabled?

thomasmaas: never mind, you added the anchor to the helper. problem is it doesn't need the anchor when independent_mapping_rules are enabled

app/helpers/api/mapping_rules_helper.rb
```diff
@@ -0,0 +1,13 @@
...
+
+module Api::MappingRulesHelper
+  def link_to_mapping_rules(service:)
+    path, data = if current_account.independent_mapping_rules_enabled?
+                   [admin_service_proxy_rules_path(service), {}]
+                 else
+                   [edit_admin_service_integration_path(service, anchor: 'mapping-rules'), { behavior: 'open-mapping-rules' }]
+                 end
+
+    link_to 'Mapping Rules', path, anchor: 'mapping-rules', data: data
```
thomasmaas: anchor isn't valid for independent_mapping_rules right?

josemigallas: Is it possible to include `anchor` inside `data`?

duduribeiro: this anchor shouldn't be here. it is already present in the else condition

duduribeiro:
closed because it was fixed by https://github.com/3scale/porta/pull/1342

PR #1351, opened 2019-10-21
Settings: Fix staging and production URL grayed out

didierofrivia:
With `ENABLE_APICAST_CUSTOM_URL=1`
with-custom-url

With `ENABLE_APICAST_CUSTOM_URL=0`
without-custom-url

Default Hosted APIcast url behaviour:
updated_default_behaviour

Fixes a bug with proxy staging and production URLs and integration method selected.

Fixes https://issues.jboss.org/browse/THREESCALE-3753

** Notes **
Missing cuke for feature. https://issues.jboss.org/browse/THREESCALE-3762

### DIFF
--- a/app/javascript/src/Settings/authentication-widget.jsx
+++ b/app/javascript/src/Settings/authentication-widget.jsx
@@ -31,6 +31,8 @@ const toggleDisabled = toggleAttrInSetting('disabled')
 
 const toggleReadOnly = toggleAttrInSetting('readonly')
 
+const resetValue = (isReadOnly) => isReadOnly ? toggleAttrInSetting('value')(false) : settings => settings
+
 export function initialize () {
   const authWrapper = document.getElementById(AUTH_WRAPPER_ID)
   const authSettingsWrapper = document.getElementById(AUTH_SETS_WRP_ID)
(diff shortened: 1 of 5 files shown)


thomasmaas:
Checking locally:

1. URL for self managed doesn't get saved
2. when switching to self managed would be good to not show the URL of 3scale managed but just the the placeholder, which is what we do now in SaaS production:

app/javascript/src/Settings/authentication-widget.jsx
```diff
@@ -11,11 +11,31 @@ const INTEGRATION_METHODS_CLASS = 'integration-method'
...
 const OIDC_ID = 'service_proxy_authentication_method_oidc'
+const PROXY_ENDPOINT_CLASS = 'proxy-endpoint'
+const PROXY_ENDPOINTS_ID = 'proxy-endpoints'
+const SELF_MANAGED = 'service_deployment_option_self_managed'
 
-const toggleActive = (setting, active) => {
+const toggle = (...toggleFns) => active => setting => toggleFns.map(toggle => toggle(active)).reduce((s, t) => t(s), setting)
+
+const toggleAttrInSetting = (attr) => (active) => (setting) => {
+  active ? setting.setAttribute(attr, attr) : setting.removeAttribute(attr)
```
hallelujah: Shouldn't be it
```suggestion
  active ? setting.setAttribute(attr, true) : setting.removeAttribute(attr)
```

didierofrivia: It doesn't work with `readonly` sadly

hallelujah: W3C :trollface:

PR #1352, opened 2019-10-21
[do not merge] [wip] Hotfix: Do not disable promote to staging

hallelujah:
We are unsure yet if the proxy config affecting change is fully complete

Fixes: https://issues.jboss.org/browse/THREESCALE-3760

Proposal is to quick fix it (because of ER2)

Then probably use in a later PR something like:

https://github.com/3scale/porta/blob/master/app/services/apicast_v2_deployment_service.rb#L13-L15

### DIFF
--- a/app/helpers/api/integrations_helper.rb
+++ b/app/helpers/api/integrations_helper.rb
@@ -94,7 +94,8 @@ module Api::IntegrationsHelper
   end
 
   def promote_to_staging_button_options(proxy)
-    return disabled_promote_button_options if proxy.any_sandbox_configs? && !proxy.pending_affecting_changes?
+    # FIXME: Hot fix to https://issues.jboss.org/browse/THREESCALE-3760
+    # return disabled_promote_button_options if proxy.any_sandbox_configs? && !proxy.pending_affecting_changes?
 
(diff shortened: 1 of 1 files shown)


app/helpers/api/integrations_helper.rb
```diff
@@ -101,7 +101,8 @@ def promote_to_staging_button_options(proxy)
   end
 
   def promote_to_production_button_options(proxy)
-    return disabled_promote_button_options if proxy.environments_have_same_config?
+    # FIXME: Hot fix to https://issues.jboss.org/browse/THREESCALE-3760
```
thomasmaas: issue is about promote to staging button and this changes behavior promote to production button

guicassolato:
Let's wait a little bit before merging this. In the end we may need it for 2.7 GA, but for now https://github.com/3scale/porta/pull/1354 already fixed the specific issue reported in https://issues.jboss.org/browse/THREESCALE-3760.

hallelujah:
Closing, it seems we do not need it. We are going to [improve the proxy affecting changes.](https://github.com/3scale/porta/pull/1658)

PR #1354, opened 2019-10-21
Issue proxy affecting change on backend api private endpoint update

guicassolato:
This is an alternative to https://github.com/3scale/porta/pull/1352. Instead of always enabling the "Promote to staging" button, it makes `BackendApi` to issue a `ProxyConfigs::AffectingObjectChangedEvent` whenever `private_endpoint` is updated. The former is a broader "fix" that prevents from any other (yet unkown) flaws of the proxy affecting changes mechanism. This is a more specific approach addressing the issue reported in https://issues.jboss.org/browse/THREESCALE-3760.

### DIFF
--- a/app/lib/proxy_config_affecting_changes.rb
+++ b/app/lib/proxy_config_affecting_changes.rb
@@ -78,6 +78,21 @@ module ProxyConfigAffectingChanges
     end
   end
 
+  module BackendApiExtension
+    extend ActiveSupport::Concern
+
+    included do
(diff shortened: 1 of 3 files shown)


hallelujah:
:+1:

It is not really an alternative as we do not know how many other changes are not taking into account.

thomasmaas:
@hallelujah I would agree if this was for CR1 but given it's for ER2 I think this is the approach we should take.

PR #1357, opened 2019-10-21
Add fix to Wizard

josemigallas:
Follows up on #1304 , the last commit did not get pushed.. so here it is.

It adds the "backend api path" to the test request.

Screen Shot 2019-10-21 at 17 36 08

### DIFF
--- a/app/forms/onboarding/request_form.rb
+++ b/app/forms/onboarding/request_form.rb
@@ -1,7 +1,9 @@
+# frozen_string_literal: true
+
 class Onboarding::RequestForm < Reform::Form
   include ThreeScale::Reform
 
-  SLASH = '/'.freeze
+  SLASH = '/'
 
   model :proxy
 
@@ -29,7 +31,10 @@ class Onboarding::RequestForm < Reform::Form
   end
 
   def proxy_base_url
-    URI model.sandbox_endpoint.to_s
+    path = model.backend_api_configs.first&.path || ''
+    base = model.sandbox_endpoint
+
+    URI.join(base, path).to_s
   end
 
   def uri
@@ -48,7 +53,7 @@ class Onboarding::RequestForm < Reform::Form
 
     status = proxy_test_service.perform
   ensure
-    model.update_column(:api_test_success, status.try!(:success?))
+    model.update_column(:api_test_success, status&.success?)
   end
 
   protected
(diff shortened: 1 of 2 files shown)


app/controllers/provider/admin/onboarding/wizard/product_controller.rb
```diff
@@ -25,6 +24,10 @@ def update
 
   private
 
+  def new_service
```
thomasmaas: i find this naming very confusing given it's an existing service?

app/services/proxy_test_service.rb
```diff
@@ -96,8 +97,11 @@ def api_test_path
       uri = URI(override)
     end
 
-    uri.merge!(proxy.api_test_path.to_s)
+    api_config = proxy.backend_api_configs && proxy.backend_api_configs.first
+    path = api_config ? api_config.path + '/' : ''
 
+    uri.merge!(path + proxy.api_test_path)
```
damianpm: I think `uri.merge!(path + proxy.api_test_path.to_s)` will fix the failing test

app/services/proxy_test_service.rb
```diff
@@ -96,7 +97,10 @@ def api_test_path
       uri = URI(override)
     end
 
-    uri.merge!(proxy.api_test_path.to_s)
+    api_config = proxy.backend_api_configs&.first
+    path = api_config ? api_config.path + '/' : ''
+
+    uri.merge!(path + proxy.api_test_path)
```
damianpm: I think `uri.merge!(path + proxy.api_test_path.to_s)` will fix the failing test

app/forms/onboarding/request_form.rb
```diff
@@ -23,13 +25,17 @@ def api_name
   end
 
   def api_base_url
+    binding.pry
```
hallelujah: 🚫 😱

josemigallas: 💩 !!!

app/forms/onboarding/request_form.rb
```diff
@@ -29,7 +31,10 @@ def api_base_url
   end
 
   def proxy_base_url
-    URI model.sandbox_endpoint.to_s
+    path = model.backend_api_configs.first&.path || ''
```
Martouta: not sure if it should be like this:
```suggestion
    path = model.backend_api_configs.first&.path || SLASH
```

but @duduribeiro must say :smile:

duduribeiro: It can be both.. if you send the empty slash, it will save the `/`. But if we want to have this clear in the UI, I think `path = model.backend_api_configs.first&.path || SLASH` is better!

josemigallas: According to my experiments, the `''` value is equivalent to doing nothing

PR #1358, opened 2019-10-22
Removing box-shadow in nested forms in integration page

damianpm:

Some nested forms in configuration page are showing box-shadow.

**Before:**
before

**After:**
after

**NO APIAP**

**Integration page:**
no-apiap-integration-page

**Configuration page:**
no-apiap-configuration-page


damianpm:
> Sure it also works with no apiap?

@thomasmaas Yes, I added some screenshots

PR #1360, opened 2019-10-22
Update jest to 24.9.0 to fix a security issue of handlebars dependency

damianpm:

Updating jest to `^24.9.0` to fix a vulnerability issue with a dependency (handlebars)

**Verification steps** 

`npm run update-dependencies`
`npm run jest`

All JS tests must run and pass as ususal

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8084,9 +8084,9 @@
       "dev": true
     },
     "handlebars": {
-      "version": "4.2.0",
-      "resolved": "https://registry.npmjs.org/handlebars/-/handlebars-4.2.0.tgz",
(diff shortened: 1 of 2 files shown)


josemigallas:
Would it be a good idea to start using fixed versions, to avoid these kind of problems?

damianpm:
> Would it be a good idea to start using fixed versions, to avoid these kind of problems?

Not sure, if we fix all versions we won't get updated warnings

josemigallas:
I guess we would get them whenever Dependabot opens a new PR, no?

damianpm:
I think dependabot is only watching Ruby

PR #1363, opened 2019-10-22
Ignore empty endpoint when generating curl command

duduribeiro:
[THREESCALE-3783](https://issues.jboss.org/browse/THREESCALE-3783)

We have a problem in integrations page when the user access it with a
proxy that has a blank endpoint. It is raising an error when trying to
construct the URI for it.

This commit changes it to avoid constructing the command for blank
endpoints.

### DIFF
--- a/app/helpers/api/integrations_helper.rb
+++ b/app/helpers/api/integrations_helper.rb
@@ -8,7 +8,7 @@ module Api::IntegrationsHelper
     attr_reader :proxy
 
     def command
-      return unless base_endpoint
+      return if base_endpoint.blank?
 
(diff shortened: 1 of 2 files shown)


app/helpers/api/integrations_helper.rb
```diff
@@ -8,7 +8,7 @@ def initialize(proxy)
     attr_reader :proxy
 
     def command
-      return unless base_endpoint
+      return unless base_endpoint.present?
```
Martouta:
```suggestion
      return if base_endpoint.blank?
```

PR #1365, opened 2019-10-23
[invoice] word break normal

macejmic:

Invoice in admin is not readable.

https://issues.jboss.org/browse/THREESCALE-3791


didierofrivia:
Do we know what introduced the breaking change? cc @damianpm @josemigallas @thomasmaas

thomasmaas:
I think it was me in https://github.com/3scale/porta/pull/1317

PR #1367, opened 2019-10-23
Mitigates ActiveRecord::StaleObjectError on proxy update

guicassolato:
Concurrent updates of a proxy record, particularly when one of the threads is trying to "touch" the object, can cause ActiveRecord to raise an `ActiveRecord::StaleObjectError`. When initiated in the scope of an HTTP request, the update may end up rejected with a `409 Conflict` response due to this race condition.

Example of recently introduced background operations now touching the proxy object at a much higher frequency than before:

https://github.com/3scale/porta/blob/42779a0ae2f7bf3060550b2e34aaeb2e0aeb2497/app/workers/proxy_config_affecting_change_worker.rb#L7
https://github.com/3scale/porta/blob/42779a0ae2f7bf3060550b2e34aaeb2e0aeb2497/app/models/proxy_config_affecting_change.rb#L4

Closes [THREESCALE-3772](https://issues.jboss.org/browse/THREESCALE-3772)

**Special notes to the reviewer**
An alternative approach would be enforcing pessimist locking on the proxy update, though Rails 4 is known to have issues with ActiveRecord locking and touch operations: https://github.com/rails/rails/commit/9721b45b40c045a132b7177f0e86b0f5d567a2df.

### DIFF
--- a/app/models/proxy.rb
+++ b/app/models/proxy.rb
@@ -194,7 +194,6 @@ class Proxy < ApplicationRecord
 
     def default_production_endpoint_apiap; end
 
-
     protected
 
     delegate :provider, to: :service
@@ -533,6 +532,11 @@ class Proxy < ApplicationRecord
 
   def service_mesh_integration?
     Service::DeploymentOption.service_mesh.include?(deployment_option)
+  end
+
+  def update_attributes(*)
+    reload # helps to prevent ActiveRecord::StaleObjectError. Using pessimistic locking wouldcould be an alternative.
+    super
   end
 
   protected
--- a/test/unit/proxy_test.rb
+++ b/test/unit/proxy_test.rb
@@ -683,4 +683,26 @@ class ProxyTest < ActiveSupport::TestCase
       proxy.update_attributes(updated_at: Time.utc(2019, 9, 26, 12, 20))
     end
   end
+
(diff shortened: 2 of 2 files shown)


app/models/proxy.rb
```diff
@@ -535,6 +534,11 @@ def service_mesh_integration?
     Service::DeploymentOption.service_mesh.include?(deployment_option)
   end
 
+  def update_attributes(*)
+    reload # helps to prevent ActiveRecord::StaleObjectError. Using pessimistic locking wouldcould be an alternative.
```
guicassolato: Not happy with this 😕

hallelujah:
Well pessimistic lock will be bad (it will block all unicorn workers) because proxy affecting changes are created in the view (during a web request): creating a lock on proxy

See my comment: https://github.com/3scale/porta/pull/1244/files#r329042359

Another approach would be removing the `touch: true` and find another way to commit the updated at after X changes, like a queue

guicassolato:
@hallelujah, the proxy affecting changes may be generated in the view, but they async processed. We issue `ProxyConfigs::AffectingObjectChangedEvent`'s processed by `ProxyConfigAffectingChangeWorker`. The actual touch happens in the background.

Martouta:
Wow! :clap: :1st_place_medal:

PR #1368, opened 2019-10-23
Redirect to show when promotion to Staging fails

josemigallas:

Render `:show` instead of `:edit` in APIAP.
Redirect to `:show` instead whenever the controller goes to `:edit`.
Normally when navigating to `/edit` an error 404 is thrown, that is not changed.

[THREESCALE-3788: Failing promotion to staging brings user to config edit page](https://issues.jboss.org/browse/THREESCALE-3788)

**Verification steps** 

1. Go to `apiconfig/services/<service_id>/integration`
2. Try to promote in a way that it fails 😅
3. It should not redirect to the config edit page

### DIFF
--- a/app/controllers/api/integrations_controller.rb
+++ b/app/controllers/api/integrations_controller.rb
@@ -23,6 +23,8 @@ class Api::IntegrationsController < Api::BaseController
   def settings; end
 
   def update
+    @show_presenter = Api::IntegrationsShowPresenter.new(@proxy)
+
     if @service.using_proxy_pro? && !@proxy.apicast_configuration_driven
       proxy_pro_update
     elsif @proxy.save_and_deploy(proxy_params)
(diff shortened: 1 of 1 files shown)


app/controllers/api/integrations_controller.rb
```diff
@@ -109,6 +110,10 @@ def show
     end
   end
 
+  def instantiate_presenter
```
hallelujah: I do not see the benefit of having it as a method

josemigallas: In order to have it also in the `edit` method: https://github.com/3scale/porta/pull/1368/files#diff-03ca76903039d35f4565cfda67324befR26
, when rendering :show.

app/controllers/api/integrations_controller.rb
```diff
@@ -6,6 +6,8 @@ class Api::IntegrationsController < Api::BaseController
   before_action :authorize
   before_action :hide_for_apiap, only: :edit
   before_action :find_registry_policies, only: %i[edit update]
+  before_action :instantiate_presenter, only: :update
+  before_action :redirect_to_show, if: :apiap?, only: :edit
```
guicassolato: This won't even get to execute while we keep `hide_for_apiap`.
`redirect_to_show` need to be defined too.

PR #1370, opened 2019-10-24
Disabling submit button when policies endpoint is wrong

damianpm:

In Policies page (`apiconfig/services/<id>/policies/edit`), when fetch to policies endpoint fails for any reason, an error message is shown: "Failed to fetch APIcast policies from ...", but the submit button is enabled and the form can be submitted. This PR disables submit button when fetch APIcast policies fails.

Try with APIAP on. When APIAP off, the button is not displayed at all in the integration page.

https://issues.jboss.org/browse/THREESCALE-3793

**Before:**
before

**After:**
after


josemigallas:
`true if @error` would have been nicer 😛

PR #1371, opened 2019-10-24
Shorten labels auth options for apiap

thomasmaas:
before:
Screenshot 2019-10-24 10 29 23

after:
Screenshot 2019-10-23 09 44 29

### DIFF
--- a/app/javascript/src/Settings/styles/authentication.scss
+++ b/app/javascript/src/Settings/styles/authentication.scss
@@ -2,7 +2,8 @@
   display: none !important;
 }
 
-#service_deployment_option_input {
+#service_deployment_option_input,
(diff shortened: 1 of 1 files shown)


josemigallas:
Why hiding it when it can be removed?

Also, maybe those the spans text could be set in titles, to not to lose the information.

app/javascript/src/Settings/styles/authentication.scss
```diff
@@ -2,7 +2,8 @@
   display: none !important;
 }
 
-#service_deployment_option_input {
+#service_deployment_option_input,
```
damianpm: Why not using same style of `inline-hints`?
authentication

PR #1372, opened 2019-10-25
[BackgroundDeletion] module

macejmic:

Association that should be deleted on the background should be defined manually, instead of taking them by the `depenend` option from the association.

https://issues.jboss.org/browse/THREESCALE-3702

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -58,6 +58,27 @@ class Account < ApplicationRecord
   include States
   include ProviderDomains
 
+  self.background_deletion = [
+    :users,
+    :mail_dispatch_rules,
+    :api_docs_services,
+    :services,
+    :contracts,
+    :account_plans,
+    [:settings, { action: :destroy, has_many: false }],
+    [:payment_detail, { action: :destroy, has_many: false }],
+    [:buyer_accounts, { action: :destroy, class_name: 'Account' }],
+    [:payment_gateway_setting, { action: :destroy, has_many: false }],
+    [:profile, { action: :delete, has_many: false }],
+    [:templates, { action: :delete }],
+    [:sections, { action: :delete, class_name: 'CMS::Section' }],
+    [:provided_sections, { action: :delete, class_name: 'CMS::Section' }],
+    [:redirects, { action: :delete }],
+    [:files, { action: :delete }],
+    [:builtin_pages, { action: :delete }],
+    [:provided_groups, { action: :delete }]
+  ]
+
   #TODO: this needs testing?
   scope :providers, -> { where(provider: true) }
 
--- a/app/models/concerns/background_deletion.rb
+++ b/app/models/concerns/background_deletion.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+module BackgroundDeletion
+  extend ActiveSupport::Concern
+
+  included do
+    class_attribute :background_deletion, default: [], instance_writer: false
+  end
+
+  class Reflection
+
+    DEFAULT_DESTROY_METHOD = 'destroy'
+    DEFAULT_HAS_MANY_OPTION = true
+
+    attr_reader :name, :options
+
+    def initialize(association_config)
+      config = Array(association_config)
+
+      @name = config[0]
+      @options = config[1].presence || {}
+    end
+
+    def many?
+      options.fetch(:has_many) { DEFAULT_HAS_MANY_OPTION }
+    end
+
+    def class_name
+      options[:class_name].presence || name.to_s.singularize.classify
+    end
+
+    def background_destroy_method
+      options[:action].to_s.presence || DEFAULT_DESTROY_METHOD
+    end
+  end
+end
--- a/app/workers/delete_object_hierarchy_worker.rb
+++ b/app/workers/delete_object_hierarchy_worker.rb
@@ -11,7 +11,7 @@ class DeleteObjectHierarchyWorker < ApplicationJob
   queue_as :deletion
 
   before_perform do |job|
-    @object, workers_hierarchy, @destroy_method = job.arguments
+    @object, workers_hierarchy, @background_destroy_method = job.arguments
     id = "Hierarchy-#{object.class.name}-#{object.id}"
     @caller_worker_hierarchy = Array(workers_hierarchy) + [id]
     info "Starting #{job.class}#perform with the hierarchy of workers: #{caller_worker_hierarchy}"
@@ -21,7 +21,7 @@ class DeleteObjectHierarchyWorker < ApplicationJob
     info "Finished #{job.class}#perform with the hierarchy of workers: #{caller_worker_hierarchy}"
   end
 
-  def perform(_object, _caller_worker_hierarchy = [], _destroy_method = 'destroy')
+  def perform(_object, _caller_worker_hierarchy = [], _background_destroy_method = 'destroy')
     build_batch
   end
 
@@ -37,7 +37,8 @@ class DeleteObjectHierarchyWorker < ApplicationJob
     workers_hierarchy = options['caller_worker_hierarchy']
     info "Starting DeleteObjectHierarchyWorker##{method_name} with the hierarchy of workers: #{workers_hierarchy}"
     object = GlobalID::Locator.locate(options['object_global_id'])
-    DeletePlainObjectWorker.perform_later(object, workers_hierarchy, destroy_method)
+    background_destroy_method = @background_destroy_method.presence || 'destroy'
+    DeletePlainObjectWorker.perform_later(object, workers_hierarchy, background_destroy_method)
     info "Finished DeleteObjectHierarchyWorker##{method_name} with the hierarchy of workers: #{workers_hierarchy}"
   rescue ActiveRecord::RecordNotFound => exception
     info "DeleteObjectHierarchyWorker##{method_name} raised #{exception.class} with message #{exception.message}"
@@ -52,7 +53,7 @@ class DeleteObjectHierarchyWorker < ApplicationJob
   def build_batch
     batch = Sidekiq::Batch.new
     batch.description = batch_description
-    batch_callbacks(batch) { batch.jobs { delete_associations } }
+    batch_callbacks(batch) { batch.jobs { destroy_and_delete_associations } }
     batch
   end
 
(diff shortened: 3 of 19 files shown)


app/models/concerns/background_deletation.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module BackgroundDeletation
```
hallelujah: Deletion ;)

macejmic: right :)

app/models/concerns/background_deletion.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module BackgroundDeletion
+  extend ActiveSupport::Concern
+
+  delegate :background_destroy_associations, to: 'self.class'
```
hallelujah: class_attribute?

app/workers/delete_object_hierarchy_worker.rb
```diff
@@ -95,7 +95,11 @@ def destroyable_associations
   end
 
   def destroyable_association?(reflection)
-    %i[destroy delete_all].include?(reflection.options[:dependent])
+    if object.respond_to?(:background_destroy_associations)
+      object.background_destroy_associations.include?(reflection.name)
+    else
+      raise "The old way! BackgroundDeletation should be used for #{object.class.name}"
```
hallelujah: Can you use a tagged error and correct the "deletation" to "deletion"?

app/workers/delete_object_hierarchy_worker.rb
```diff
@@ -107,18 +111,21 @@ class ReflectionDestroyer
...
     def destroy_later
       reflection.macro == :has_many ? destroy_has_many_association : destroy_has_one_association
     end
 
-    attr_reader :main_object, :reflection, :caller_worker_hierarchy, :destroy_method
+    attr_reader :main_object, :reflection, :caller_worker_hierarchy
 
     private
 
+    def destroy_method
```
hallelujah:
```suggestion
    def background_destroy_method
```

app/workers/delete_object_hierarchy_worker.rb
macejmic: this won't be here.. it's just for CI.

app/models/concerns/background_deletion.rb
macejmic: could be.

app/workers/delete_object_hierarchy_worker.rb
macejmic: 🍤

app/models/concerns/background_deletion.rb
macejmic: worked, nice! :)

app/workers/delete_object_hierarchy_worker.rb
macejmic: but when I thinking about it, maybe it would make it.. so will change it.

app/models/concerns/background_deletion.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+module BackgroundDeletion
+  extend ActiveSupport::Concern
+
+  included do
+    class_attribute :background_destroy_associations, default: []
+    class_attribute :background_delete_associations, default: []
```
hallelujah: `instance_writer: false`

app/workers/delete_object_hierarchy_worker.rb
```diff
@@ -94,8 +97,28 @@ def destroyable_associations
...
-    %i[destroy delete_all].include?(reflection.options[:dependent])
+    should_be_destroyed_or_deleted?(reflection, :background_destroy_associations)
+  end
+
+  def deletable_association?(reflection)
+    should_be_destroyed_or_deleted?(reflection, :background_delete_associations)
+  end
+
+  def should_be_destroyed_or_deleted?(reflection, list_name)
+    list_defined = object.respond_to?(list_name)
```
hallelujah: you can use `object.method(list_name)` and rescue from `NameError`

That way you can just do `method.call.include?`

By the way raising an error when it is dependent but not in the list is too hard failure. we should continue the excecution of deleting no?

app/workers/delete_object_hierarchy_worker.rb
```diff
@@ -94,8 +97,28 @@ def destroyable_associations
...
+
+    if !list_defined && %i[destroy delete_all].include?(reflection.options[:dependent])
```
hallelujah: About my comment https://github.com/3scale/porta/pull/1372/files#r340507324
I think we should not raise an error if it is not in the list

Martouta:
We spoke about the way to go, and as @macejmic summarized, this is the way:

1. simplify the configuration, two arrays might not be the best idea
2. get rid of reflect_on_all_associations if possible (we have that list defined)
3. get rid of reflections, the last purpose what be to know macro (has_many) etc.

app/workers/delete_object_hierarchy_worker.rb
```diff
@@ -21,7 +21,7 @@ class DeleteObjectHierarchyWorker < ApplicationJob
     info "Finished #{job.class}#perform with the hierarchy of workers: #{caller_worker_hierarchy}"
   end
 
-  def perform(_object, _caller_worker_hierarchy = [], _destroy_method = 'destroy')
+  def perform(_object, _caller_worker_hierarchy = [], _background_destroy_method = 'destroy')
```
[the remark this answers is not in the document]
Martouta: Personally I prefer in this case to keep it like `(_object, _caller_worker_hierarchy = [], _background_destroy_method = 'destroy')`  instead of `perform(*)` because:
1. It has default values by doing it this way.
2. All of them are actually used, but saved in the `before_perform`, so this way we enforce to send exactly 3 params while the other way we accept anything and it is prone to errors. Besides, putting them names makes more clear what we expect and what each param means.

app/workers/delete_object_hierarchy_worker.rb
```diff
@@ -73,49 +74,71 @@ def batch_callbacks(batch)
...
+    class FakeReflection
+      attr_reader :name, :options
+
+      def initialize(name, options)
+        @name = name
+        @options = options
+      end
+
+      def many?
+        return options[:has_many] if options.keys.include?(:has_many)
```
hallelujah:
```suggestion
return options.fetch(:has_many){ plural? }
```

app/workers/delete_object_hierarchy_worker.rb
```diff
@@ -73,49 +74,71 @@ def batch_callbacks(batch)
...
+      end
+
+      def background_destroy_method
+        options[:action].to_s
+      end
+
+      private
+
+      def plural?
+        string_name = name.to_s
```
hallelujah: Well if name is `:users` then `:users.to_s.pluralize == "users"` that works fine

BUT invariable word will not:

`"information".pluralize == "information"`

app/models/account.rb
```diff
@@ -58,6 +58,27 @@ def self.columns
   include States
   include ProviderDomains
 
+  self.background_deletion = {
+    users: { action: :destroy },
```
Martouta: I dunno if we should make that the default action is 'destroy' and we don't need to repeat `action: :destroy` everywhere 😄

app/workers/delete_object_hierarchy_worker.rb
```diff
@@ -73,49 +74,71 @@ def batch_callbacks(batch)
...
-  end
-
   def callback_options
     { 'object_global_id' => object.to_global_id, 'caller_worker_hierarchy' => caller_worker_hierarchy }
   end
 
   class ReflectionDestroyer
 
-    def initialize(main_object, reflection, caller_worker_hierarchy)
+    class FakeReflection
```
Martouta: 'fake' 😂 😂 😂 I think it is better for example 'SimpleReflection', among many other options

app/workers/delete_object_hierarchy_worker.rb
```diff
@@ -73,49 +74,71 @@ def batch_callbacks(batch)
...
+      end
+
+      def many?
+        return options[:has_many] if options.keys.include?(:has_many)
+
+        plural?
+      end
+
+      def class_name
+        options[:class_name].presence || name.to_s.singularize.classify
```
Martouta: I hope that it isn't a problem, but just in case... `options[:class_name]` is a String and `name.to_s.singularize.classify` is a class 🤔 perhaps it is safer to do:
`options[:class_name].presence || name.to_s.singularize.classify.name`

Martouta: In fact I think that if it is a class and not a string, it will not do the conditional well in here:
https://github.com/3scale/porta/pull/1372/files#diff-ff63dfaf8bfa4a850361d41a988711a1R165

app/workers/delete_object_hierarchy_worker.rb
```diff
@@ -73,49 +74,71 @@ def batch_callbacks(batch)
...
+      end
+
+      def background_destroy_method
+        options[:action].to_s
```
Martouta: If you do as proposed in https://github.com/3scale/porta/pull/1372/files#r348058625, then this would be:
`options[:action].to_s || 'destroy'`

app/models/account.rb
Martouta: In that case, it would be:
```ruby
  self.background_deletion = [
    :users,
    :mail_dispatch_rules,
    # many other here
    buyer_accounts: { class_name: 'Account' },
    # many other here
    profile: { action: :delete },
    # many other here
  ].freeze
```

app/workers/delete_object_hierarchy_worker.rb
```diff
@@ -73,49 +74,71 @@ def batch_callbacks(batch)
     end
   end
 
-  def delete_associations
-    destroyable_associations.each do |association|
-      ReflectionDestroyer.new(object, association, caller_worker_hierarchy).destroy_later
+  def destroy_and_delete_associations
+    Array(object.background_deletion).each do |association, options|
```
Martouta: If you do it like https://github.com/3scale/porta/pull/1372/files#r348064637, then this would be just `object.background_deletion.each do |association, options|`

app/workers/delete_object_hierarchy_worker.rb
Martouta: I agree with Hery that this is playing with fire, for example for 'Settings' 🤔 
I'd go for enforce the option 'has_many/has_one' or just 'one/many' 😄

app/models/account.rb
macejmic: let me think about it..

app/workers/delete_object_hierarchy_worker.rb
macejmic: it's not bulletproof but it does work for 95% of cases.. for that 5 % is going to fail (TDD) and then you can specify `has_many` option.

app/workers/delete_object_hierarchy_worker.rb
macejmic: hehe :D could be..

app/workers/delete_object_hierarchy_worker.rb
macejmic: `classsify` does not return class, but rather a class name: https://apidock.com/rails/ActiveSupport/Inflector/classify

app/workers/delete_object_hierarchy_worker.rb
macejmic: it's either configure everything manually or https://github.com/3scale/porta/pull/1372#discussion_r348069440

macejmic: i have an idea how to prevent `information. pluralize ` problem..

guicassolato:
I was thinking about something a bit different to register in the models, something like

```
class Parent < ApplicationRecord
  has_many :children, dependent: { background: :destroy }
end
```

...and of course extend https://apidock.com/rails/v4.2.9/ActiveRecord/Associations/CollectionAssociation/delete_all.

In fact, I think that could even make a nice a gem, although not strictly necessary.

WDYT?

macejmic:
@guicassolato extracting it to a gem would be awesome.. much work right now tho, but definitely on my todo list.

PR #1373, opened 2019-10-25
Fix Policies when having twice of the same type

josemigallas:

When having 2 policies of the same type, clicking between one and the other wasn't updating the UI.

This happens because React is not detecting a change in `PolicyConfig` props and therefore the view isn't updated. The fix is to render/destroy the detailed view when it's supposed to be displayed/hidden by css.

[THREESCALE-1924: When adding a 2nd URL rewriting policy the UI doesn't work as expected](https://issues.jboss.org/browse/THREESCALE-1924)

**Verification steps** 

1. Go to Integration > Policies
2. Add 2 policies of the same kind, i.e. _Upstream_
3. Open the first one, edit its config, click on _Update Policy_
4. Open the second one, its config should be empty.

Alternative:
Having 2 policies of the same kind with different configs:
1. Open 1, close it.
2. Open the other, configs should change.

### DIFF
--- a/app/javascript/src/Policies/components/PoliciesForm.jsx
+++ b/app/javascript/src/Policies/components/PoliciesForm.jsx
@@ -20,7 +20,6 @@ class PolicyForm extends Form {
 
 type Props = {
   schema?: string,
-  visible: boolean,
   policy: ChainPolicy,
   submitForm: (ChainPolicy) => ThunkAction,
   removePolicy: (ChainPolicy) => ThunkAction,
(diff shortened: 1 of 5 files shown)


app/javascript/src/Policies/components/PoliciesForm.jsx
```diff
@@ -20,7 +20,6 @@ class PolicyForm extends Form {
 
 type Props = {
   schema?: string,
-  visible: boolean,
```
hallelujah: Why the visible property is removed?

PR #1375, opened 2019-10-25
3scale release migration test

eguzki:

A test to validate 3scale release upgrade procedure is needed. High level, the test does the following:
* Installs previous 3scale release
* Run migration / upgrade procedure
* Check migration has been successful

Implemented as a CircleCi test to validate 3scale release upgrade procedure. The test goes as follows:

* installs _previous_ 3scale release. Previous release is set as job parameter to be easily changed when a new version is released. Currently, previous release is *2.6*
* builds current master release of porta image
* Run the `system-app` prehook command (defined in templates/operator) from recently built porta master image.
* Test will pass if prehook command returns successfully (exit code is 0). 

Currently the test is quite simple. IMO, it could be improved in, at least, two ways:
* Installed 3scale _previous_ release contains just initial data. It could be interesting adding some kind of data (like proxy info, backend info, product info, whatever...) that could lead to edge cases to test upgrade procedure.
* Test will pass if prehook command returns successfully (exit code is 0). This could lead to false positives. For instance, the migration process misses upgrading some table. The upgrade procedure will still succeed, but the migration process is clearly failing to be complete.

Those enhancements, though, could be added later in other PR's. 

https://issues.jboss.org/browse/THREESCALE-3808

Data model and actual data migration process can be broken and nobody notices it. 

**Verification steps** 

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -295,6 +295,42 @@ commands: # reusable commands with parameters
       - *store-log-artifacts
       - *upload-coverage
 
+  install-openshift:
+    steps:
+      - run:
+          name: Install OpenShift Client Tools
+          working_directory: /tmp
+          command: |
+            curl --fail -L  https://github.com/openshift/origin/releases/download/v3.11.0/openshift-origin-client-tools-v3.11.0-0cbc58b-linux-64bit.tar.gz | tar -xzf-
+            sudo mv /tmp/openshift-origin-client-tools-*-linux-64bit/oc /usr/local/bin/
+            sudo mv /tmp/openshift-origin-client-tools-*-linux-64bit/kubectl /usr/local/bin/
+      - run:
+          name: Configure Docker
+          command: |
+            echo '{"insecure-registries": ["172.30.0.0/16"]}' | sudo tee --append /etc/docker/daemon.json
+            sudo systemctl restart docker
+      - run:
+          name: Start and Configure OpenShift Cluster
+          working_directory: /tmp/openshift
+          command: |
+            DOCKER_HOST_IP=$(docker run --net=host codenvy/che-ip)
+            oc cluster up --public-hostname=${DOCKER_HOST_IP} --enable=persistent-volumes \
+              --enable=registry --enable=router
+            oc login https://${DOCKER_HOST_IP}:8443 -u system:admin --insecure-skip-tls-verify=true > /dev/null
+            oc adm policy add-cluster-role-to-user cluster-admin developer > /dev/null
+            oc adm policy add-scc-to-group hostmount-anyuid system:serviceaccounts
+            oc login https://${DOCKER_HOST_IP}:8443 -u developer --insecure-skip-tls-verify=true > /dev/null
+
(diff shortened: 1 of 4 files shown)


hallelujah:
@eguzki Should we do it in the CI of porta?
I think it is better fit on product side, with QE tests.

.circleci/config.yml
```diff
@@ -905,11 +941,67 @@ jobs:
...
+            oc create -f https://raw.githubusercontent.com/3scale/3scale-operator/<< parameters.threescale_release_version >>-stable/pkg/3scale/amp/auto-generated-templates/amp/amp-eval.yml
+            oc new-app 3scale-api-management-eval --param WILDCARD_DOMAIN=lvh.me
+            oc wait --for=condition=available --timeout=-1s $(oc get dc --output=name)
+          no_output_timeout: 30m
+      - checkout
+      - run:
+          name: Build system master image
+          command: |
+            oc create -f test/migration/build-system-master.json
+            oc start-build amp-system-master --wait --follow
```
guicassolato: Ideally, we'd use nightly builds image, instead of building a new one.

.circleci/config.yml
```diff
@@ -295,6 +295,42 @@ commands: # reusable commands with parameters
       - *store-log-artifacts
       - *upload-coverage
 
+  install-openshift:
+    steps:
+      - run:
+          name: Install OpenShift Client Tools
+          working_directory: /tmp
+          command: |
+            curl --fail -L  https://github.com/openshift/origin/releases/download/v3.11.0/openshift-origin-client-tools-v3.11.0-0cbc58b-linux-64bit.tar.gz | tar -xzf-
```
guicassolato: Why 3.11? Better OpenShift 4.X, no?

.circleci/config.yml
eguzki: Then you could not check if the current PR is breaking upgrade procedure (by enabling manually the migration test)

.circleci/config.yml
eguzki: Haven't tried that yet. Not even for local dev. 

You need to use the codeready containers thing.
https://developers.redhat.com/blog/2019/09/05/red-hat-openshift-4-on-your-laptop-introducing-red-hat-codeready-containers/

You need lots of resources.

* 4 virtual CPUs (vCPUs)
* 8 GB of memory
* 35 GB of storage space

Martouta:
@eguzki @hallelujah @guicassolato I dunno what happened with this PR but:
1. Either finish it or close it or leave a comment mentioning why it is on hold.
2. Do not merge it before https://github.com/3scale/porta/pull/1665. And in fact, you will have to rebase after that PR and fix conflicts.

eguzki:
QE should implement this test. They agreed to implement. 

Second option is implementing in porta. This option allowes you (as porta developer) to test it if you suspect your PR could break.

Third option is to run nightly in 3scale operator CI pipeline. Nightly we will run tests and report to you if the test fail.

Pick your choice

hallelujah:

PR #1376, opened 2019-10-25
Reorganization of APIcast configuration view for APIAP

guicassolato:
It reorganizes how APIcast Configuration info is displayed for APIAP accounts, making it more clear that Backends are composing parts of Products, they also bring their own mapping rules to the configuration.

It also now shows the backends in the same order as they will be injected into the routing policy rules.

Screen Shot 2019-10-25 at 5 19 26 PM

Related to [THREESCALE-3781](https://issues.jboss.org/browse/THREESCALE-3781) (although it doesn't really show the full list of mapping rules, but it already communicates better to the user)

### DIFF
--- a/app/helpers/api/integrations_helper.rb
+++ b/app/helpers/api/integrations_helper.rb
@@ -221,10 +221,17 @@ module Api::IntegrationsHelper
 
   def backend_routing_rule(backend_api_config)
     path = StringUtils::StripSlash.strip_slash(backend_api_config.path.presence)
-    code = content_tag :code do
-      "/#{path} => "
+    content_tag :code do
+      "/#{path} => #{backend_api_config.backend_api.private_endpoint}"
     end
-    endpoint = backend_api_config.backend_api.private_endpoint
-    code + endpoint
+  end
+
+  def proxy_rules_preview(proxy_rules, path: nil)
+    last_rule = proxy_rules.last
+    return 'None' unless last_rule
+    code = content_tag(:code) { "#{path}#{last_rule.pattern} => #{last_rule.metric.name}" }
+    rules_size = proxy_rules.size
+    more = rules_size > 1 ? " and #{rules_size - 1} more." : ''
+    code + more
   end
 end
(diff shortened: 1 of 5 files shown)


app/helpers/api/integrations_helper.rb
```diff
@@ -221,10 +221,17 @@ def disabled_promote_button_options
...
-    endpoint = backend_api_config.backend_api.private_endpoint
-    code + endpoint
+  end
+
+  def proxy_rules_preview(proxy_rules, path: nil)
+    last_rule = proxy_rules.last
+    return 'None' unless last_rule
+    code = content_tag(:code) { "#{path}#{last_rule.pattern} => #{last_rule.metric.name}" }
+    rules_size = proxy_rules.size
+    more = rules_size > 1 ? " and #{rules_size - 1} more." : ''
```
hallelujah: Just asking for future adjustment, this `and X more` can be a link @thomasmaas ?

app/views/api/integrations/apicast/configuration_driven/_apicast.html.slim
```diff
@@ -6,36 +6,40 @@ section.Configuration class=('Configuration--is-promoted' unless deployment_opti
...
       - unless deployment_option_is_service_mesh?(@service)
         dt.u-dl-term Credential Location
         dd.u-dl-definition = @proxy.credentials_location
         dt.u-dl-term Secret Token
         dd.u-dl-definition = @proxy.secret_token
 
-
       - if apiap?
+        - unless deployment_option_is_service_mesh?(@service)
+          - for backend_api_config in @service.backend_api_configs.sorted_for_proxy_config.includes(:backend_api)
```
hallelujah: Can it be a `collection.each` instead of a `for ... in`?

app/helpers/api/integrations_helper.rb
guicassolato: +1 to this idea.

guicassolato: https://github.com/3scale/porta/pull/1379

PR #1382, opened 2019-10-28
Use create_or_find_by instead of find_or_create_by with retry

duduribeiro:
[THREESCALE-3785](https://issues.jboss.org/browse/THREESCALE-3785)
We have a problem in MailDispatchRule that can be from a Race Condition.
There are some RecordNotUnique errors exploding when the
DeleteObjectHierarchy Worker runs. This commit changes this to first try
to insert in the database, if it fails, we execute the find by.

### DIFF
--- a/app/models/mail_dispatch_rule.rb
+++ b/app/models/mail_dispatch_rule.rb
@@ -6,18 +6,10 @@ class MailDispatchRule < ApplicationRecord
 
   scope :enabled, -> { where(dispatch: true) }
 
-  def self.fetch_with_retry!(options, &block)
-    retries = 0
-
-    begin
-      MailDispatchRule.find_or_create_by!(options, &block)
-    rescue ActiveRecord::RecordNotUnique
-      if retries > 10
-        raise
-      else
-        retries += 1
-        retry
-      end
-    end
+  # This is copied from Rails 6 source. We should remove it soon as we move to Rails 6.
+  def self.create_or_find_by!(options, &block)
+    transaction(requires_new: true) { create!(options, &block) }
+  rescue ActiveRecord::RecordNotUnique
+    find_by!(options, &block)
   end
 end
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -400,7 +400,7 @@ class Account < ApplicationRecord
 
   # @param [SystemOperation] operation
   def fetch_dispatch_rule(operation)
-    MailDispatchRule.fetch_with_retry!(system_operation: operation, account: self) do |m|
+    MailDispatchRule.create_or_find_by!(system_operation: operation, account: self) do |m|
       m.dispatch = false if %w[weekly_reports daily_reports new_forum_post].include?(operation.ref)
       m.emails = emails.first
     end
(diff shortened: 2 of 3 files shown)


app/models/mail_dispatch_rule.rb
```diff
@@ -6,18 +6,10 @@ class MailDispatchRule < ApplicationRecord
...
-    rescue ActiveRecord::RecordNotUnique
-      if retries > 10
-        raise
-      else
-        retries += 1
-        retry
-      end
-    end
+  # This is copied from Rails 6 source. We should remove it soon as we move to Rails 6.
+  def self.create_or_find_by!(options, &block)
```
guicassolato:
```suggestion
  def self.create_or_find_by!(attributes, &block)
```

app/models/mail_dispatch_rule.rb
```diff
@@ -6,18 +6,10 @@ class MailDispatchRule < ApplicationRecord
...
+    transaction(requires_new: true) { create!(options, &block) }
```
guicassolato:
```suggestion
    transaction(requires_new: true) { create!(attributes, &block) }
```

app/models/mail_dispatch_rule.rb
```diff
@@ -6,18 +6,10 @@ class MailDispatchRule < ApplicationRecord
...
+  rescue ActiveRecord::RecordNotUnique
+    find_by!(options, &block)
```
guicassolato:
```suggestion
    find_by!(attributes, &block)
```

Martouta:
When you do and push the changes requested by Gui, please run the tests of all DBs

hallelujah:
This might not fix the underlying issue:

There might be some concurrency deadlock issue: https://stackoverflow.com/questions/37972925/mysql-locking-in-duplicate-key-error

Meaning that there might be different threads doing the same thing at the same time. The `deadlock_retry` gem retries X times the request already, but this is trying it up to 10 times. Meaning it might try it 10 multiplied by X times ... creating more deadlock.

To my point of view, before removing an account or after creating an account, we should ensure that the mail dispatch rules are created so it does not have to create them in a Sidekiq thread

Martouta:
Check [this comment](https://issues.jboss.org/browse/THREESCALE-3785?focusedCommentId=13806588&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13806588) by Andrew:

> Just a reminder on this one - let us know if this results in a risky change.
> If so, we can keep off 2.7 CR1, into master and for saas and 2.8.
> Just let Cath know if in doubt.

So perhaps we should do this for 2.8 instead of 2.7 CR1 :thinking: WDYT? @hallelujah @duduribeiro @secretsurf

hallelujah:
@Martouta definitely out of 2.7 CR1 scope.

hallelujah:
@duduribeiro closing?

Martouta:
@hallelujah @duduribeiro the error keeps happening from the tests i have done deleting tenants in preview :smile: but it did not happen in production in saas because we are not deleting tenants there yet :joy: 
Not saying that I am against closing this PR at all, just adding that extra info that seems that you guys don't know :smile:

Martouta:
Closing because it does not fix the issue.

PR #1386, opened 2019-10-30
Segment REST + Bugfix of number of requests

Martouta:
### What this PR means to solve
1. We used to use the `GraphQL` API to delete users in Segment because that was the only API available for this matter back then. But now that API is deprecated and not working anymore, and this PR moves to the `REST` API. Closes [THREESCALE-3395 - Segment GraphQL deprecation on 2019, September 10th](https://issues.jboss.org/browse/THREESCALE-3395)
2. We used to make a request for each user destroyed to Segment right away, but we used to have problems of `rate limit`, so this PR also aims to fix that. Closes [THREESCALE-2840 - Segment::ClientError in SegmentDeleteUserWorker@default](https://issues.jboss.org/browse/THREESCALE-2840)

### Tasks
- [X] The 1st commit only removes all the previous implementation of removing users from Segment. It only does not remove the code and test to fetch [the config file](https://github.com/3scale/deploy/pull/610).
- [X] When a user is deleted, it is saved in DeletedObject.
- [X] The job config runs a worker daily that deletes from segment those users that have been deleted previously.
- [X] Remove from "DeletedObject" those users that have already been deleted in Segment. This way those users won't be tried to be removed the next day again. And because we don't need them there anymore at that point.
- [X] Remove all the previously deleted users from Segment. For the time when the GraphQL has not worked and we have been destroying users in SaaS anyway, so now we will need to use this new implementation to remove also those. The plan is to have a csv with all the users we have in Segment, and run a rake task that for each one of them, if it is not in our DB as a user, we save it as a deleted object. This way the job cron will remove them from Segment as well later on.

### Verification steps
I've also tested it in preview, but just to know if it works, it can be done locally as explained here.
1. Check a few user ids that we actually want to remove for real. Do it by doing this in the rails console of Preview: `Account.deleted_since.first(3).map { |account| account.users.pluck(:id) }.flatten`. For the next steps, let's imagine that those user ids are these: `[-1, -2, -3]`. I wrote them in negative on purpose just in case that someone does not see that it is just an example and actually executes, so we won't be removing real data of real non-deleted users by accident.
2. Locally, in the development environment, in the rails console, create the users with those ids and destroy them to create the DeletedObject by the code.
3. Copy the config of Segment from `config/examples/features.yml` to `config/features.yml`. Use the correct configuration credentials from [the deploy repository](https://github.com/3scale/deploy/pull/610). Do not forget to enable the config.
4. Run the low queue of Sidekiq. `bundle exec sidekiq -q low -L log/sidekiq.log`
5. In the rails console opened after the config is changed, run the worker. `SegmentDeleteUsersWorker.perform_now`
6. See the [regulations created](https://app.segment.com/3scale/settings/deletion-requests) as expected. You can also do locally from your terminal: `curl "https://platform.segmentapis.com/v1beta/workspaces/{workspace_name}/regulations" --header "Authorization: Bearer YOUR_TOKEN"`


app/models/deleted_object.rb
```diff
@@ -17,5 +17,14 @@ class DeletedObject < ApplicationRecord
     joins(join_sql)
   end
 
-  scope :stale, -> { where.has { ((id.in DeletedObject.deleted_owner.select(:id)) | (object_type == Service.name)) & (created_at <= 1.week.ago) } }
+  scope :stale, -> do
+    where.has do
+      (
+        (id.in DeletedObject.deleted_owner.select(:id)) \
+          | (object_type == Service.name) \
+          | (object_type == Account.name)
```
Martouta: wrong, i gotta remove this. TODO

app/models/user.rb
```diff
@@ -121,7 +121,7 @@ def moderatable
   # after_validation :reset_lost_password_token
 
   after_save :nullify_authentication_id, if: :any_sso_authorizations?
-  after_commit :notify_deletion, on: :destroy
+  after_destroy :archive_as_deleted
```
Martouta: I have to check and explain this. TODO

Martouta: Because the account must still be present :smile: and this should be part of the transaction.

test/unit/deleted_object_test.rb
```diff
@@ -36,25 +40,31 @@ class DeletedObjectTest < ActiveSupport::TestCase
...
     service = FactoryBot.create(:simple_service)
     metric = FactoryBot.create(:metric, service: service)
     deleted_child_but_owner_persisted_recent = DeletedObject.create(object: metric, owner: service, created_at: (1.week - 1.day).ago).id
 
     service = FactoryBot.create(:simple_service)
+    account = service.account
     metric = FactoryBot.create(:metric, service: service)
     deleted_child_and_owner_service_deleted_recent = DeletedObject.create(object: metric, owner: service, created_at: (1.week - 1.day).ago).id
-    deleted_object_service_but_owner_persisted_recent = DeletedObject.create(object: service, owner: service.account, created_at: (1.week - 1.day).ago).id
+    deleted_child_and_owner_account_deleted_recent = DeletedObject.create(object: account.admin_user, owner: account, created_at: (1.week - 1.day).ago).id
```
guicassolato: Maybe we could use `Timecop` here to make the whole test easier to read.

app/services/segment/delete_users_service.rb
```diff
@@ -0,0 +1,43 @@
...
+      end
+    end
+
+    def config
+      Features::SegmentDeletionConfig.config
+    end
+
+    def request_body(user_ids)
+      <<~JSON
+        {
```
guicassolato:
```ruby
{
  regulation_type: 'Suppress_With_Delete',
  attributes: {
    name: 'userId',
    values: user_ids.map(&:to_s)
  }
}.to_json
```

app/workers/segment_delete_users_worker.rb
```diff
@@ -0,0 +1,21 @@
...
+  # :reek:FeatureEnvy can be ignored here for the 'response'
+  def perform
+    return unless Features::SegmentDeletionConfig.enabled?
+    config = Features::SegmentDeletionConfig.config
+    DeletedObject.users.select(:id, :object_id).order(:id).find_in_batches(batch_size: config.request_size) do |records|
+      Segment::DeleteUsersService.call(records.map(&:object_id))
+      records.each(&DeleteObjectHierarchyWorker.method(:perform_later))
+      sleep(config.wait_time)
+    end
+  rescue Segment::ClientError, Segment::UnexpectedResponseError => error
```
guicassolato: Should we abort the entire job because one batch had one failed request? 🤔

Martouta: If it is ClientError probably so because it means that we are doing something wrong, either wrong url, or RateLimitExceeded or something like that, and it will be the same for all of them 🤔

lib/tasks/segment/segment.rake
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+require_dependency 'csv'
+
+namespace :segment do
+  desc 'Save as deleted objects the users from the imported segment csv'
+  task :save_deleted_users, [:file_name] => [:environment] do |_task, args|
+    file_path =  Rails.root.join('config', args[:file_name])
```
guicassolato: Why in `config`?

Martouta: Right, it is not configuration, it is actually the user data... but where then? 🤔

test/unit/services/segment/delete_users_service_test.rb
```diff
@@ -0,0 +1,73 @@
...
+  def delete_request(status:)
+    uri = "#{config.root_uri}/workspaces/#{config.workspace}/#{config.api}"
+    stub_request(:post, uri).with(
+        body: stubbed_request_body,
+        headers: {'Authorization'=>"Bearer #{config.token}", 'Content-Type'=>'application/json; charset=utf-8'}).
+      to_return(status: status, body: 'body response')
+  end
+
+  def stubbed_request_body
+    <<~JSON
```
Martouta: Same as https://github.com/3scale/porta/pull/1386/files#r347377481

app/models/deleted_object.rb
```diff
@@ -17,5 +17,10 @@ class DeletedObject < ApplicationRecord
     joins(join_sql)
   end
 
-  scope :stale, -> { where.has { ((id.in DeletedObject.deleted_owner.select(:id)) | (object_type == Service.name)) & (created_at <= 1.week.ago) } }
+  scope :stale, -> do
+    where.has do
+      ((id.in DeletedObject.deleted_owner.select(:id)) | (object_type == Service.name)) \
```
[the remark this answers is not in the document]
Martouta: > is this initial parenthesis required?
> does this work?
> `id.in(DeletedObject.deleted_owner.select(:id)) | object_type == Service.name`

I will try.

> also, you really meant to use the binary operations or the boolean ones can handle it?
> ||  - &&

this is [baby_squeel](https://github.com/rzane/baby_squeel) 😄 example from the documentation:
image
```
Author.where.has { (name =~ 'Ray%') & (id < 5) | (name.lower =~ 'zane%') & (id > 100) }
# SELECT "authors".* FROM "authors"
# WHERE ("authors"."name" LIKE 'Ray%' AND "authors"."id" < 5 OR LOWER("authors"."name") LIKE 'zane%' AND "authors"."id" > 100)
```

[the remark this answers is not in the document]
duduribeiro: 👍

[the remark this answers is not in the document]
Martouta: It doesn't work without the parentheses with the example you gave.
you can see that by changing the code as you said, then doing in the terminal locally `bundle exec rails console` and inside it `DeletedObject.stale` to see the SQL that it generates in MySQL. But in this case, it raises an error 😄

lib/tasks/segment/segment.rake
Martouta: I've just made it as "responsibility of the user to pass the whole path, be it in config or anywhere else"

app/workers/segment_delete_users_worker.rb
Martouta: I have done it the following way:
- If it fails for ServerError, it will retry, and it will continue where it left it for the users that were left to work with, because it has removed the records that has already worked on, and ONLY those, because if it fails, it does not reach the next line.
- If it fails for ClientError, it is our fault, and we don't want it to keep retrying for other users because it will keep failing.

WDYT of what it should do @guicassolato @duduribeiro @hallelujah @macejmic ?

macejmic: yeah, that makes sense for me..

app/models/user.rb
macejmic: couldn't be audit used for this?

Martouta: I am not sure if that is a good idea because it has a very different purpose :thinking: It could be an option but should be very well thought.

Martouta: I feel that is very scary because:
1. We might remove the audit before time and not be able to use it for this at some point.
2. Making queries in the audit is not really ideal.
3. We use the audits as a log mostly, not to "use to do X thing, and then remove once used", which would change a lot.

lib/tasks/segment/segment.rake
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+require_dependency 'csv'
+
+namespace :segment do
+  desc 'Save as deleted objects the users from the imported segment csv'
+  task :save_deleted_users, [:file_name] => [:environment] do |_task, args|
+    file_path =  Rails.root.join(args[:file_name])
```
guicassolato: I think it didn't even need to be in the app directory anyway, but an absolute path instead.

Martouta: @guicassolato This isn't in the 'app' directory, this is anywhere inside the Rails root. It can be app, lib, config, etc.
However, if you mean that we should accept outside the rails root as well, ok :+1: It could be a feasible option :+1:

PR #1389, opened 2019-10-31
Fix broken bubbles

damianpm:

Some popup bubbles are broken, sometimes showing a space between the icon (?) and the text box, and sometimes are displayed at the right (outside the browser window).

**Before:**
before

**After:**
after

**Brach deployed to preview**: https://multitenant-admin.preview01.3scale.net/p/admin/dashboard

**Note:** Bubbles needs a whole refactor, this is just a quick-fix
New task created for 2.8 release https://issues.jboss.org/browse/THREESCALE-3854

### DIFF
--- a/app/assets/javascripts/help.js.erb
+++ b/app/assets/javascripts/help.js.erb
@@ -20,19 +20,30 @@ if(typeof Help == 'undefined'){
         var layerTopPos  = 10,
             layerLeftPos = 10;
 
+        var textRight = $image.offset().left + $helperLayer.width();
 
         var offset = (function(i, l){
-            var top  = i.offset().top - (i.offset().top - i.position().top) + i.height() - 1,
-                left = i.offset().left - (i.offset().left - i.position().left);
+            var top  = i.offset().top - (i.offset().top - i.position().top) + i.height() - 1;
+            var left = i.offset().left - (i.offset().left - i.position().left);
 
+            if (textRight > $(window).width()) {
+             left = left - l.width();
+            }
             return {top:top, left:left};
         })($image, $helperLayer);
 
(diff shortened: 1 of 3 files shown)


damianpm:
> Bit scary, sure you checked all places where we show help icons?

All I could find were OK, but we could deploy to preview and check with more data.
Can do it next Tuesday, no need to merge right now

didierofrivia:
Check the cuke failing, it might be that now a help bubble blocks the delete button. Also, we should create a jira for replacing this old implementation favoring https://www.patternfly.org/v4/documentation/react/components/popover/

features/old/buyers/users.feature
```diff
@@ -57,7 +57,7 @@ Feature: Buyer users management
   Scenario: Delete buyer user
     When I go to the buyer user page for "bob"
     And I follow "Edit"
-    Then I follow "Delete" and I confirm dialog box
+    Then I follow invisible "Delete" and I confirm dialog box
```
didierofrivia: hmmm I don't think this is the way to go. Usually, a feature indicates what the user should see and interact with. If we are tampering a cuke step with clicking an invisible button/link, the feature is just a fallacy.

damianpm: The problem is I can't reproduce it because the test pass locally, I was trying this to see if this was the problem, but its still failing.

damianpm:
> Check the cuke failing, it might be that now a help bubble blocks the delete button. Also, we should create a jira for replacing this old implementation favoring https://www.patternfly.org/v4/documentation/react/components/popover/

Created: https://issues.jboss.org/browse/THREESCALE-3854

PR #1391, opened 2019-10-31
Display a deprecated message when Service uses plugin deployment

duduribeiro:
[THREESCALE-2395](https://issues.jboss.org/browse/THREESCALE-2395)

Our plugin deployment option is deprecated. To avoid needing some data migration in database, this commit sets the deployment option as hosted if the current option is set to plugins and also displays a message to the user to update their configuration.

Screenshot 2019-11-04 21 12 35

### DIFF
--- a/app/models/service.rb
+++ b/app/models/service.rb
@@ -521,6 +521,18 @@ class Service < ApplicationRecord
     proxy.apicast_configuration_driven && !proxy.service_mesh_integration?
   end
 
+  # TODO: Remove this when no one use plugins
+  def plugin_deployment?
+    deployment = self[:deployment_option]
+    DeploymentOption.plugins.include?(deployment)
+  end
+
+  # TODO: Remove this when no one use plugins
+  def deployment_option
+    return 'hosted' if plugin_deployment?
+    super
+  end
+
   private
 
   def archive_as_deleted
--- a/app/models/service.rb
+++ b/app/models/service.rb
@@ -523,14 +523,7 @@ class Service < ApplicationRecord
 
   # TODO: Remove this when no one use plugins
   def plugin_deployment?
-    deployment = self[:deployment_option]
-    DeploymentOption.plugins.include?(deployment)
-  end
-
-  # TODO: Remove this when no one use plugins
-  def deployment_option
-    return 'hosted' if plugin_deployment?
-    super
+    DeploymentOption.plugins.include?(deployment_option)
   end
 
   private
--- a/app/views/layouts/api/_service.html.slim
+++ b/app/views/layouts/api/_service.html.slim
@@ -2,6 +2,7 @@
   h1
     = yield :sublayout_title
 
+/ TODO: Remove this when no one use plugins
 = render 'shared/service/deprecated_plugin_warning'
 
 = yield
--- a/app/views/layouts/api/_service.html.slim
+++ b/app/views/layouts/api/_service.html.slim
@@ -2,4 +2,6 @@
   h1
     = yield :sublayout_title
 
+= render 'shared/service/deprecated_plugin_warning'
+
 = yield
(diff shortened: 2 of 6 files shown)


app/models/service.rb
```diff
@@ -521,6 +521,18 @@ def can_use_backends?
...
 
+  # TODO: Remove this when no one use plugins
+  def plugin_deployment?
+    deployment = self[:deployment_option]
+    DeploymentOption.plugins.include?(deployment)
+  end
+
+  # TODO: Remove this when no one use plugins
+  def deployment_option
+    return 'hosted' if plugin_deployment?
```
hallelujah: I think it should be self_managed
People using plugin do not really have any apicast in front. It would enforce them to have one hosted
Also zync will create a route automatically in OpenShift if it is hosted

app/views/shared/service/_deprecated_plugin_warning.html.slim
```diff
@@ -0,0 +1,7 @@
+- if @service&.plugin_deployment?
```
hallelujah: I do not get why the nil check

app/models/service.rb
thomasmaas: I’d say the logical setting is what you get out of the box when you install 3scale. That would be 3scale managed. Other question is if it would work.

app/views/layouts/api/_service.html.slim
```diff
@@ -2,4 +2,6 @@
   h1
     = yield :sublayout_title
 
+= render 'shared/service/deprecated_plugin_warning'
```
josemigallas: Maybe add a comment here as well? `/ TODO: Remove this when no one use plugins`

app/models/service.rb
```diff
@@ -521,6 +521,18 @@ def can_use_backends?
     proxy.apicast_configuration_driven && !proxy.service_mesh_integration?
   end
 
+  # TODO: Remove this when no one use plugins
+  def plugin_deployment?
+    deployment = self[:deployment_option]
+    DeploymentOption.plugins.include?(deployment)
```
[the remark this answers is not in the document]
duduribeiro: 😂 .. I hate to omit the parenthesis

app/models/service.rb
thomasmaas: Yes, everything is shown as if 3scale managed APIcast was being used. And when the customer clicks ok, setting will actually be updated in db. I agree this is not the final solution but it is kind of a nice in between with heads up to customer. Next step is remove all of it.

hallelujah: We can remove the code later but a data migration (DML) should be done IMO
Anyway plugins are not shown anymore right?

hallelujah: They are still shown (I just verified), deprecation notice was already issued long time ago. This is adding more work and has really unknown behaviours

Also this undesirable behaviour is worse than leaving it there: Zync will just create routes if it is hosted unlike before while it was still plugin.

If we just display the message and not touch the `deployment_option` method I would merge the PR.

thomasmaas: Why is it wrong that zync creates the routes? If they change to apicast because they want to that would also happen, no?

app/views/shared/service/_deprecated_plugin_warning.html.slim
hallelujah: This smells then :)

app/models/service.rb
hallelujah: It is a change of behaviour without notice zync will create tons of routes ...
Not to mention that some users reported that already https://github.com/3scale/zync/issues/236

If they want to use the hosted gateway yes, but they might also want to use the self managed gateway or the service mesh.

Agree that it makes sense to have default to hosted but I'd say the timing is pretty bad to take this decision ...

Martouta:
Quoted from Hery:

> OK for me we would need to:
> 
> 1. Merge you PR with only the notice https://github.com/3scale/porta/pull/1391
> 2. Merge my PR removing the plugins in the api and UI https://github.com/3scale/porta/pull/1249
> 3. Fix https://github.com/3scale/porta/blob/master/app/models/proxy.rb#L121 by migrating all deployment_option from service to Proxy
> 4. Remove the column from Service with the code first then with a real migration
> 5. make a DML task to put all plugins to hosted or self_managed TBD
> 6. Remove plugin code

PR #1394, opened 2019-11-04
Prevent trailing slash in patterns of proxy rules of backend api configs with a path

guicassolato:
This will make proxy rules whose pattern is the "catch-all" pattern `/` not to end with the slash if the pattern has been prepended by a backend api path.

**Example 1**
Backend path: `/mybackend`
Proxy rule pattern: `/`
Actual proxy rule pattern in the config: `/mybackend` (it used to be `/mybackend/` before this PR)

**Example 2**
Backend path: `/mybackend`
Proxy rule pattern: `/hello`
Actual proxy rule pattern in the config: `/mybackend/hello` (same as before)

**Example 3**
Backend path: `/`
Proxy rule pattern: `/`
Actual proxy rule pattern in the config: `/` (same as before)

**Example 4**
Backend path: `/`
Proxy rule pattern: `/hello`
Actual proxy rule pattern in the config: `/hello` (same as before)

----

Closes [THREESCALE-3833](https://issues.jboss.org/browse/THREESCALE-3833)

### DIFF
--- a/app/decorators/proxy_rule_decorator.rb
+++ b/app/decorators/proxy_rule_decorator.rb
@@ -5,7 +5,10 @@ class ProxyRuleDecorator < ApplicationDecorator
 
   def pattern
     pattern_value = object.pattern
-    backend_api_path ? File.join('/', backend_api_path, pattern_value) : pattern_value
+    return pattern_value unless backend_api_path
+    parts = ['/', backend_api_path]
+    parts << pattern_value unless pattern_value == '/'
+    File.join(*parts)
   end
 
   def metric_system_name
--- a/test/decorators/proxy_rule_decorator_test.rb
+++ b/test/decorators/proxy_rule_decorator_test.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class ProxyRuleDecoratorTest < Draper::TestCase
+  setup do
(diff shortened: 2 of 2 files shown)


app/decorators/proxy_rule_decorator.rb
```diff
@@ -5,7 +5,10 @@ class ProxyRuleDecorator < ApplicationDecorator
 
   def pattern
     pattern_value = object.pattern
-    backend_api_path ? File.join('/', backend_api_path, pattern_value) : pattern_value
+    return pattern_value unless backend_api_path
+    parts = ['/', backend_api_path]
+    parts << pattern_value unless pattern_value == '/'
+    File.join(*parts)
```
didierofrivia: Question: is it not better to manipulate the path with the rails `UrlHelper` ?

porueesq:
@guicassolato @thomasmaas You can create a JIRA and we'll add the note.

app/decorators/proxy_rule_decorator.rb
Martouta: I agree with @didierofrivia , this way it seems that it is somehow related to a File, but it isn't :thinking: 
another alternative would be `parts.join('/')`

Martouta: Spoken with Gui. I will push this fix after lunch in another PR :+1:

PR #1395, opened 2019-11-04
Remove custom policies from code base

josemigallas:

Removes all traces of custom policies since they are neither used nor planned to. Making this a separate commit so that it can be reversed if needed at some point.

**Verification steps**:
* Everything should work as usual

**Note**:
This is part of the refactorization of Policies:


app/controllers/provider/admin/registry/policies_controller.rb
```diff
@@ -1,7 +1,7 @@
 #frozen_string_literal: true
 
 class Provider::Admin::Registry::PoliciesController < Provider::Admin::BaseController
-  activate_menu :account, :integrate, :custom_policies
+  activate_menu :account, :integrate
```
hallelujah: Don't we remove the whole controller then?

josemigallas: Good point, and you're probably right. That controller isn't used anywhere else, is it?

didierofrivia: Remove it, it's not used. We are using `PoliciesService` in`integrations_controller` to fetch APIcast registry.

didierofrivia:
You might want to remove this presenter as well: https://github.com/3scale/porta/blob/master/app/presenters/policy_schema_presenter.rb

app/controllers/provider/admin/registry/policies_controller.rb
josemigallas: OK. How about [app/controllers/admin/api/registry/policies_controller.rb](https://github.com/3scale/porta/blob/master/app/controllers/admin/api/registry/policies_controller.rb)?

didierofrivia: If the API is not documented / published... remove it as well.

guicassolato: I think app/controllers/admin/api/registry/policies_controller.rb _is_ documented and published to all Saas accounts. Even if no one is using it, removing it can be trickier. I'd recommend checking with @3scale/product, as they may want to issue a note about it.

PR #1398, opened 2019-11-05
Service create error messages

didierofrivia:
Screenshot 2019-11-05 at 14 47 54

**BEFORE**
Screenshot 2019-11-05 at 15 23 36

**AFTER**
Screenshot 2019-11-07 at 14 29 21

It was showing misleading error messages when creating a new Service/Product.

fixes https://issues.jboss.org/browse/THREESCALE-3394

**Verification steps** 

1. Browse to the Dashboard and click on "NEW API / NEW PRODUCT".
2. Use the following sample fields:
```
Name: Invalid API
System name: I am using spaces
Description: Invalid API
```
3. Click in "Add API / Create Producte".

### DIFF
--- a/app/controllers/api/services_controller.rb
+++ b/app/controllers/api/services_controller.rb
@@ -22,7 +22,7 @@ class Api::ServicesController < Api::BaseController
 
   def new
     activate_menu :dashboard
-    @service = collection.build params[:service]
+    @service = ServicePresenter.new(collection.build(params[:service]))
   end
 
   def edit
@@ -40,7 +40,7 @@ class Api::ServicesController < Api::BaseController
   end
 
   def create
-    @service = collection.new # this is done in 2 steps so that the account_id is in place as preffix_key relies on it
+    @service = ServicePresenter.new(collection.new)
     creator = ServiceCreator.new(service: @service)
 
     if can_create? && creator.call(create_params)
--- a/app/controllers/api/services_controller.rb
+++ b/app/controllers/api/services_controller.rb
@@ -48,7 +48,7 @@ class Api::ServicesController < Api::BaseController
       onboarding.bubble_update('api')
       redirect_to admin_service_path(@service)
     else
-      flash.now[:error] = flash_message(@service.errors.keys.first, {scope: %i[flash services create errors]})
+      flash.now[:error] = @service.errors.full_messages.to_sentence.presence || I18n.t!('flash.services.create.errors.default', {resource_type: product_or_service_type})
       activate_menu :dashboard
       render :new
     end
@@ -153,12 +153,5 @@ class Api::ServicesController < Api::BaseController
 
   def authorize_admin_plans
     authorize! :admin, :plans
-  end
-
-  def flash_message(key, opts = {})
-    options = opts.reverse_merge(resource_type: product_or_service_type)
-    I18n.t!(key, options)
-  rescue I18n::MissingTranslationData
-    t(:default, options)
   end
 end
--- a/app/controllers/api/services_controller.rb
+++ b/app/controllers/api/services_controller.rb
@@ -48,7 +48,7 @@ class Api::ServicesController < Api::BaseController
       onboarding.bubble_update('api')
       redirect_to admin_service_path(@service)
     else
-      flash.now[:error] = t('flash.services.create.error', resource_type: product_or_service_type)
+      flash.now[:error] = flash_message(@service.errors.keys.first, {scope: %i[flash services create errors]})
       activate_menu :dashboard
       render :new
     end
@@ -153,5 +153,12 @@ class Api::ServicesController < Api::BaseController
 
   def authorize_admin_plans
     authorize! :admin, :plans
+  end
+
+  def flash_message(key, opts = {})
+    options = opts.reverse_merge(resource_type: product_or_service_type)
+    I18n.t!(key, options)
+  rescue I18n::MissingTranslationData
+    t(:default, options)
   end
 end
--- a/app/javascript/src/NewService/components/FormElements/ServiceManualListItems.jsx
+++ b/app/javascript/src/NewService/components/FormElements/ServiceManualListItems.jsx
@@ -1,36 +1,51 @@
 // @flow
 
-import React from 'react'
+import React, {useState} from 'react'
 import {Label} from 'NewService/components/FormElements'
+import type {ServiceFormTemplate} from 'NewService/types'
 
-const ServiceManualListItems = () =>
-  <React.Fragment>
-    <li id="service_name_input" className="string required">
-      <Label
-        htmlFor='service_name'
-        label='Name'
-        required
-      />
-      <input maxLength="255" id="service_name" type="text" name="service[name]" autoFocus="autoFocus"/>
-    </li>
-    <li id="service_system_name_input" className="string required">
-      <Label
-        htmlFor='service_system_name'
-        label='System name'
-        required
-      />
-      <input maxLength="255" id="service_system_name" type="text" name="service[system_name]"/>
-      <p className="inline-hints">
-        Only ASCII letters, numbers, dashes and underscores are allowed.
-      </p>
-    </li>
-    <li id="service_description_input" className="text optional">
-      <Label
-        htmlFor='service_description'
-        label='Description'
-      />
-      <textarea rows="3" id="service_description" name="service[description]"></textarea>
-    </li>
-  </React.Fragment>
+type Props = ServiceFormTemplate
 
+const ServiceManualListItems = ({service, errors}: Props) => {
+  const [name, setName] = useState(service.name)
+  const [systemName, setSystemName] = useState(service.system_name)
+  const [description, setDescription] = useState(service.description)
+  const onChange = (fn) => (e) => fn(e.currentTarget.value)
+
+  return (
+    <React.Fragment>
+      { /* $FlowFixMe boolean coercion in string literals */ }
+      <li id="service_name_input" className={`string required ${errors && errors.name ? 'error' : ''}`}>
+        <Label
+          htmlFor='service_name'
+          label='Name'
+          required
+        />
+        <input onChange={onChange(setName)} value={name} maxLength="255" id="service_name" type="text" name="service[name]" autoFocus="autoFocus"/>
+        { !!errors.name && <p className="inline-errors">{errors.name}</p> }
+      </li>
+      { /* $FlowFixMe boolean coercion in string literals */ }
+      <li id="service_system_name_input" className={`string required ${errors && errors.system_name ? 'error' : ''}`}>
+        <Label
+          htmlFor='service_system_name'
+          label='System name'
+          required
+        />
+        <input onChange={onChange(setSystemName)} value={systemName} maxLength="255" id="service_system_name" type="text" name="service[system_name]"/>
+        <p className={`${errors.system_name ? 'inline-errors' : 'inline-hints'}`}>
+          Only ASCII letters, numbers, dashes and underscores are allowed.
+        </p>
+      </li>
+      { /* $FlowFixMe boolean coercion in string literals */ }
+      <li id="service_description_input" className={`text optional ${errors && errors.description ? 'error' : ''}`}>
+        <Label
+          htmlFor='service_description'
+          label='Description'
+        />
+        <textarea onChange={onChange(setDescription)} value={description} rows="3" id="service_description" name="service[description]"></textarea>
+        { !!errors.description && <p className="inline-errors">{errors.description}</p> }
+      </li>
+    </React.Fragment>
+  )
+}
 export {ServiceManualListItems}
--- a/app/presenters/service_presenter.rb
+++ b/app/presenters/service_presenter.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+class ServicePresenter < SimpleDelegator
+  def as_json(*)
+    {service: stringify_nil_values(filtered_values)}.merge(error_messages)
+  end
+
+  private
+
+  def filtered_values
+    __getobj__.as_json(root: false, only: %i[name system_name description])
+  end
+
+  def error_messages
+    {errors: __getobj__.errors.messages.presence || {} }
+  end
+
+  def stringify_nil_values(hash)
+    hash.transform_values(&:to_s)
+  end
+end
--- a/app/views/api/services/new.html.slim
+++ b/app/views/api/services/new.html.slim
@@ -1,6 +1,7 @@
 = javascript_pack_tag 'new_service'
 
 #new_service_wrapper data-new-service-data={ isServiceDiscoveryAccessible: service_discovery_accessible?,
+                                             template: @service.as_json,
                                              isServiceDiscoveryUsable: service_discovery_usable?,
                                              serviceDiscoveryAuthenticateUrl: service_discovery_presenter.authorize_url,
                                              providerAdminServiceDiscoveryServicesPath: provider_admin_service_discovery_services_path,
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -886,6 +886,11 @@ en:
         service_plan:
           has_contracts: 'This service plan cannot be deleted! At least one contract depends on it.'
 
+        service:
+          attributes:
+            system_name:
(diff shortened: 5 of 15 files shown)


features/step_definitions/flash_steps.rb
```diff
@@ -1,3 +1,7 @@
 Then /^I should see the flash message "([^"]*)"$/ do |flash_text|
   find(:xpath, "//div[@id='flash-messages']").text.should have_content(flash_text)
 end
+
+Then /^I should see a flash message "([^"]*)"$/ do |flash_text|
+  find(:xpath, "//div[@id='flashWrapper']").text.should have_content(flash_text)
```
didierofrivia: Maybe there's a way to find in multiple `:xpath` ? then I could just use the original step? cc: @hallelujah @guicassolato @Martouta ?

thomasmaas:
@didierofrivia  Why aren't we using in form feedback like we do in other places?
Screenshot 2019-11-05 09 43 44

josemigallas:
> @didierofrivia Why aren't we using in form feedback like we do in other places?
> Screenshot 2019-11-05 09 43 44

I was about to suggest the same. This form is not Formtastic but React, therefore inline errors have to be implemented by hand

didierofrivia:
@thomasmaas @josemigallas In all honesty, I didn't even check the view or anything, just the controller implementation and made the smallest change keeping its behaviour. I could go with form feedback

Martouta:
I like more what Thomas and Josemi suggested.
Besides, this way for any possible error we have to add the validation and the error message and it gets more complicated.
What they suggest is rendering the new/edit from the object that contains the errors instead of Object.new again :smile: 
And regarless of that, we should show the message of the plans only if it is the case, independently of the errors in the model.

app/presenters/service_presenter.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+class ServicePresenter < SimpleDelegator
+  def as_json(*)
+    {service: stringify_nil_values(filtered_values(__getobj__))}.merge(error_messages(__getobj__))
```
didierofrivia: Is there a better way to "pipe" functions in Ruby? cc: @guicassolato @hallelujah @Martouta

hallelujah: Why do you use `__getobj__` here? You could use it in the private method ...

app/presenters/service_presenter.rb
```diff
@@ -0,0 +1,21 @@
...
+  def filtered_values(hash)
+    hash.as_json(root: false, only: %i[name system_name description])
+  end
+
+  def error_messages(hash)
+    {errors: hash.errors.any? ? hash.errors.messages : {}}.as_json
+  end
+
+  def stringify_nil_values(hash)
+    hash.map {|key, value| [key, value.presence ? value : '']}.to_h
```
hallelujah: all those value are string right? So you can just use 
```suggestion
    hash.transform_values(&:to_s)
```

app/presenters/service_presenter.rb
```diff
@@ -0,0 +1,21 @@
...
+
+class ServicePresenter < SimpleDelegator
+  def as_json(*)
+    {service: stringify_nil_values(filtered_values(__getobj__))}.merge(error_messages(__getobj__))
+  end
+
+  private
+
+  def filtered_values(hash)
+    hash.as_json(root: false, only: %i[name system_name description])
```
hallelujah:
```suggestion
    __getobj__.as_json(root: false, only: %i[name system_name description])
```

app/presenters/service_presenter.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+class ServicePresenter < SimpleDelegator
+  def as_json(*)
+    {service: stringify_nil_values(filtered_values(__getobj__))}.merge(error_messages(__getobj__))
+  end
+
+  private
+
+  def filtered_values(hash)
```
hallelujah:
```suggestion
  def filtered_values
```

app/presenters/service_presenter.rb
```diff
@@ -0,0 +1,21 @@
...
+    hash.as_json(root: false, only: %i[name system_name description])
+  end
+
+  def error_messages(hash)
+    {errors: hash.errors.any? ? hash.errors.messages : {}}.as_json
```
hallelujah:
```suggestion
    {errors: hash.errors.any? ? hash.errors.messages : {}}.as_json
```
```suggestion
    {errors: __getobj__.errors.messages.presence || {} }
```

app/presenters/service_presenter.rb
didierofrivia: @hery those might or might not be `nil` values. The objective is to avoid feeding JS null values

hallelujah: This will enforce them to be all `String` values `nil.to_s => ''`

hallelujah: Do not ping `@hery` because it is not me @didierofrivia

app/javascript/src/NewService/components/FormElements/ServiceManualListItems.jsx
```diff
@@ -1,36 +1,52 @@
...
-      <Label
-        htmlFor='service_description'
-        label='Description'
-      />
-      <textarea rows="3" id="service_description" name="service[description]"></textarea>
-    </li>
-  </React.Fragment>
+type Props = NewServiceTemplateProps
 
+const ServiceManualListItems = (props: Props) => {
```
josemigallas:
```suggestion
const ServiceManualListItems = ({ service, errors }: Props) => {
```

app/javascript/src/NewService/components/FormElements/ServiceManualListItems.jsx
```diff
@@ -1,36 +1,52 @@
...
+  const {service, errors} = props
+  const [name, setName] = useState(service.name)
+  const [systemName, setSystemName] = useState(service.system_name)
+  const [description, setDescription] = useState(service.description)
+  const onChange = (fn) => (e) => fn(e.currentTarget.value)
+
+  return (
+    <React.Fragment>
+      { /* $FlowFixMe boolean coercion in string literals */ }
+      <li id="service_name_input" className={`string required ${!!errors.name && 'error'}`}>
```
josemigallas: I don't think you need `!!`

app/views/api/services/new.html.slim
```diff
@@ -1,6 +1,7 @@
 = javascript_pack_tag 'new_service'
 
 #new_service_wrapper data-new-service-data={ isServiceDiscoveryAccessible: service_discovery_accessible?,
+                                             template: @service.as_json,
```
josemigallas: I'm not convince of the naming here. 'Template' doesn't sound right and it gets more confusing in React.
The form is for a _service_, as it is defined in the controller. Perhaps _new controller_ even, but template sounds out of place IMHO.

Turns out this object contains errors and it really is a `NewServiceTemplateProps`. So is it a template, a service or a props? I think it would make more sense for the frontend to pass both `service` and `errors` to the Form component as two different props. This way it'd more explicit and we'd be handling well-kown entities (a service and an array of error messages).

We would avoid also defining prop types such as `NewServiceTemplateProps` that belong at a component level. I think `ServiceFormTemplate` would work better as a general type (reusable not only for "new service" forms but also for "edit service" ones) and then `type NewServiceFormProps = { template: ServiceFormTemplate }`.

Hope it makes sense!

features/step_definitions/flash_steps.rb
josemigallas: This is what you're looking for:
```
find(:xpath, "//div[@id='flash-messages'] | //div[@id='flashWrapper']").text.should have_content(flash_text)
```

config/locales/en.yml
```diff
@@ -1137,7 +1137,10 @@ en:
     services:
       create:
         notice: "%{resource_type} created."
-        error: "Couldn't create %{resource_type}. Check your Plan limits" # TODO: this is not always true... there are other reasons of failure
+        errors:
+          name: "Name invalid. It Can't be blank."
```
josemigallas: ?
```suggestion
          name: "Name can't be blank."
```

features/old/services/multiservice.feature
```diff
@@ -27,6 +27,18 @@ Feature: Multiservice feature
      And I press "Create Product"
     Then I should see "Less fancy API"
 
+  @javascript
+  Scenario: Create new product: Fail scenario
```
josemigallas: We're 1 cuke short:
- Create new product: Fails with empty name
- Create new product: Fails with wrong system name

app/views/api/services/new.html.slim
didierofrivia: You are right about the name of the types, I'll change accordingly.

app/javascript/src/NewService/components/FormElements/ServiceManualListItems.jsx
didierofrivia: You don't, but since you are coercing then for printing the class name, it's cool to work with booleans instead of strings.

features/old/services/multiservice.feature
didierofrivia: Not really, we are asserting the flash message being shown, not the kind of error per se.

josemigallas: Ok, then the test should not refer to any particular string. Meaning it should be more generic like:
```
I enter a wrong value
I click on submit
I see a flash error
```
My point is that we're asserting the errors together (I'm also considering the snapshot here) and that is wrong: there should be one test case for each field. If it's not in the specs (because of use of snapshots) then this is the right place, I think.

didierofrivia: I have to disagree, we are asserting the *correct* flash message was shown here, as flash messages work (first encountered, first shown) and what is most important, what the user will see.  This is a feature, not an integration or unit test... you address (almost) real user experience.
The flash messaging system is already tested, the I18n error messages as well, etc...
Then, in JS, we are testing the invalid fields and their messages with the template.

app/javascript/src/NewService/components/FormElements/ServiceManualListItems.jsx
josemigallas: Fair enough.
There's two problems that I just realized, tho:
* `errors.name` is a maybe type, potentially throwing an error. This is obfuscated by the FlowFixMe comment.
* When `name` is falsy, the string will be `string required false`. It has to be `${!!errors.name ? 'error' : ''}`

features/old/services/multiservice.feature
josemigallas: OK

config/locales/en.yml
```diff
@@ -1137,7 +1137,10 @@ en:
     services:
       create:
         notice: "%{resource_type} created."
-        error: "Couldn't create %{resource_type}. Check your Plan limits" # TODO: this is not always true... there are other reasons of failure
+        errors:
+          name: "Name can't be blank."
```
Martouta: The problem of this approach is that we have to rewrite all possible errors of all validations in here, so when we add/edit/remove validations from service, we also have to change it here 😞

didierofrivia: It's not the only approach, that's for free. Either we leave it as it is, or roll back to the default ugly one or remove the entire flash message. There's also feedback per field.

Martouta: > Either we leave it as it is, or roll back to the default ugly one or remove the entire flash message

I don't think so :smile: I will write the alternative I have in mind this afternoon :wink:

didierofrivia: Bare in mind there's already implemented the per field error feedback, the flash message is complementary

Martouta: @didierofrivia https://github.com/3scale/porta/pull/1437 is pointing to this PR to make clearer what we mean 😄

damianpm:
@didierofrivia when trying to run the server I'm getting this error:

`/Users/damian/.rvm/gems/ruby-2.4.6/gems/railties-4.2.11.1/lib/rails/application.rb:234:in `config_for': Could not load configuration. No such file - /Users/damian/Documents/3scale/porta/config/google_experiments.yml (RuntimeError)`

What I'm missing?

app/controllers/api/services_controller.rb
```diff
@@ -40,15 +40,15 @@ def usage_rules
...
+    @service = ServicePresenter.new(collection.new)
     creator = ServiceCreator.new(service: @service)
 
     if can_create? && creator.call(create_params)
       flash[:notice] = t('flash.services.create.notice', resource_type: product_or_service_type)
       onboarding.bubble_update('api')
       redirect_to admin_service_path(@service)
     else
-      flash.now[:error] = t('flash.services.create.error', resource_type: product_or_service_type)
+      flash.now[:error] = flash_message(@service.errors.keys.first, {scope: %i[flash services create errors]})
```
Martouta: :scream: What happened @didierofrivia ?

Martouta: This is the old version

Martouta: image

image

app/javascript/src/NewService/components/FormElements/ServiceManualListItems.jsx
```diff
@@ -1,36 +1,51 @@
...
+        </p>
+      </li>
+      { /* $FlowFixMe boolean coercion in string literals */ }
+      <li id="service_description_input" className={`text optional ${!!errors.description && 'error'}`}>
+        <Label
+          htmlFor='service_description'
+          label='Description'
+        />
+        <textarea onChange={onChange(setDescription)} value={description} rows="3" id="service_description" name="service[description]"></textarea>
+        { !!errors.description && <p className="inline-errors">{errors.description}</p> }
```
damianpm: What are the possible errors for "Description" ?
I've just created a product with empty description.
I think the description is optional, in that case better remove this `p`

josemigallas: Maybe when it's too long? Would the server return an error or ..?

app/controllers/api/services_controller.rb
didierofrivia: oops!

app/javascript/src/NewService/components/FormElements/ServiceManualListItems.jsx
didierofrivia: `Only ASCII letters, numbers, dashes and underscores are allowed`

didierofrivia: ok, it validates length, maximum 65535... a lot, but still potentially could fail

app/javascript/src/NewService/components/FormElements/ServiceManualListItems.jsx
```diff
@@ -1,36 +1,51 @@
...
+const ServiceManualListItems = ({service, errors}: Props) => {
+  const [name, setName] = useState(service.name)
+  const [systemName, setSystemName] = useState(service.system_name)
+  const [description, setDescription] = useState(service.description)
+  const onChange = (fn) => (e) => fn(e.currentTarget.value)
+
+  return (
+    <React.Fragment>
+      { /* $FlowFixMe boolean coercion in string literals */ }
+      <li id="service_name_input" className={`string required ${!!errors.name && 'error'}`}>
```
josemigallas: There's two problems that I just realized, tho:
* `errors.name` is a maybe type, potentially throwing an error. This is obfuscated by the FlowFixMe comment.
* When `name` is falsy, the string will be `string required false`. It has to be `${!!errors.name ? 'error' : ''}`

josemigallas:
```suggestion
      <li id="service_name_input" className={`string required ${errors && errors.name ? 'error' : ''}`}>
```

didierofrivia: * `errors.name` could be `undefined`, but not `null`, so won't be throwing any exceptions.
* And yes! nice catch, it would be the wrong string ("false")

josemigallas: You're right, I was thinking about `errors` being undefined, but that can't happen.

PR #1404, opened 2019-11-07
Cleaning PF4 stylesheets

damianpm:

PF4 stylesheets have been included in an inconsistent manner: duplicated styles, styles copied from node_modules and included as ours, hardcoded values, overrides, etc.

This PR include all PF4 base styles with webpacker importing them from node_modules and delete copied and duplicated styles.

To be created ?

**Verification steps** 
You may need to run
```
npm run update-dependencies
bundle exec rake webpack:compile && bundle exec rake assets:precompile
```

Pay special attention to the **header**, until now our header is not responsive, but **PF4 is responsive**.
At some point, our "page" should match PF4 official demo: https://pf4.patternfly.org/documentation/core/demos/page/default-nav

With this PR now we are taking advantage of PF4 responsive system.
But, the header may still need a future refactor (e.g. Context selector should be on the right side, plus some different styles when small screen, etc.)

~**IE11 is not supported by PF4**, so this won't work in IE11. But as we are deprecating IE11 at the end of this year, we should be good to continue.~
**UPDATE**: Adding **partial IE11 support**, including a specific stylesheet only for IE11 using browser detection.
In IE11 the app will not change, header will remain fixed, not responsive.

**Firefox/Chrome/Opera:**
new

**IE11:**
ie11-3

Branch [deployed to preview](https://multitenant-admin.preview01.3scale.net/p/admin/dashboard)


didierofrivia:
LGTM. The thing is about the timing of this, not sure if merge it right now as it is... or under a rolling update... or merge it in a month and a half. What do you reckon @thomasmaas ?

didierofrivia:
Or merge it to a different branch, say... `ui-ie11less`

thomasmaas:
Top right icons has extra margin at the right compared to current header:
Screenshot 2019-11-12 15 27 45

PR #1406, opened 2019-11-08
Fix Products' search bar in Dashboard

josemigallas:

Css rule `display: none;` was being overridden by `display: table`.

* Moved `hidden` definition to a higher priority place so that it overrides `display: table`
* Sets min-height so that the card doesn't collapse

products

[THREESCALE-3871: Products search bar in Dashboard not working](https://issues.jboss.org/browse/THREESCALE-3871)

**Verification steps** 

1. Open the dashboard, Products tab
2. Use the search bar: products should hide and show accordingly with the search query

Todo:
- [ ] Add tests

### DIFF
--- a/features/provider/dashboard/tabs.feature
+++ b/features/provider/dashboard/tabs.feature
@@ -1,0 +1,36 @@
+Feature: Dashboard search bar
+  In order to navigate easily to products and backends
+  As a provider
+  I want to be able to filter them by name
+
+  Background:
+    Given a provider is logged in
+    And a service "My Fancy Product"
+    And a service "My Regular Product"
+    And a backend api "My Fancy Backend API"
+    And a backend api "My Regular Backend API"
+    And I go to the provider dashboard
+
+  @javascript
+  Scenario: Filtering products
+    When I select the products tab
+    And I should see "My Fancy Product" in the apis dashboard products tabs section
+    And I should see "My Regular Product" in the apis dashboard products tabs section
+    Then I search for "My Fan" using the products search bar
+    And I should see "My Fancy Product" in the apis dashboard products tabs section
+    And I should not see "My Regular Product" in the apis dashboard products tabs section
+    Then I search for "My Foo Foo" using the products search bar
+    And I should not see "My Fancy Product" in the apis dashboard products tabs section
+    And I should not see "My Regular Product" in the apis dashboard products tabs section
+
+  @javascript
+  Scenario: Filtering backends
+    When I select the backends tab
+    And I should see "My Fancy Backend API" in the apis dashboard backends tabs section
+    And I should see "My Regular Backend API" in the apis dashboard backends tabs section
+    Then I search for "My Fan" using the backends search bar
+    And I should see "My Fancy Backend API" in the apis dashboard backends tabs section
+    And I should not see "My Regular Backend API" in the apis dashboard backends tabs section
+    Then I search for "My Foo Foo" using the backends search bar
+    And I should not see "My Fancy Backend API" in the apis dashboard backends tabs section
+    And I should not see "My Regular Backend API" in the apis dashboard backends tabs section
--- a/features/step_definitions/dashboard_steps.rb
+++ b/features/step_definitions/dashboard_steps.rb
@@ -1,18 +1,20 @@
-def service_id_for_name (name)
-  page.find_by_id('apis').find('section', :text => %r{#{name}}i)[:id][/\d+/]
+# frozen_string_literal: true
+
+def service_id_for_name(name)
+  page.find_by_id('apis').find('section', text: /#{name}/i)[:id][/\d+/]
 end
 
-def service_for_name (name)
+def service_for_name(name)
   service_id = service_id_for_name(name)
   page.find("#service_#{service_id}")
 end
 
-def hits_for_name (name, opts = {})
+def hits_for_name(name, opts = {})
   service_id = service_id_for_name(name)
   page.find_by_id("dashboard-widget-service_id-#{service_id}service_hits", opts)
 end
 
-def top_traffic_for_name (name, opts = {})
+def top_traffic_for_name(name, opts = {})
   service_id = service_id_for_name(name)
   page.find_by_id("dashboard-widget-service_id-#{service_id}service_top_traffic", opts)
 end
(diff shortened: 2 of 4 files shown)


features/provider/dashboard/tabs.feature
```diff
@@ -0,0 +1,25 @@
...
+  Scenario: Filtering products
+    When the products tab is selected
+    Then I should see "API" in the apis dashboard products tabs section
+    And I search for "AP" using the products search bar
+    Then I should see "API" in the apis dashboard products tabs section
+    And I search for "foo" using the products search bar
+    Then I should not see "API" in the apis dashboard products tabs section
+
+  @javascript
+  Scenario: Filtering backends
```
Martouta: not a big deal for CR2 but the 2 scenarios shouldn't have the same name

josemigallas: What names would you give them?

Martouta: my bad :woman_facepalming: for checking too quickly :woman_facepalming:

features/step_definitions/dashboard_steps.rb
```diff
@@ -41,8 +43,18 @@ def top_traffic_for_name (name, opts = {})
...
 
 When(/^I (fold|unfold) service "([^"]*)"$/) do |action, service_name|
-  step %{service "#{service_name}" is #{action == 'fold' ? 'unfolded' : 'folded'}}
+  step %(service "#{service_name}" is #{action == 'fold' ? 'unfolded' : 'folded'})
 
   service = service_for_name(service_name)
   service.find('.DashboardSection-toggle').click
 end
+
+When(/^the (products|backends) tab is selected$/) do |tab|
```
guicassolato: I think that `When... is selected` implies no action of the user involved, so I'd expect it to be assertion instead of a click. I understand the purpose of this step is not to be an assertion, so maybe better `I select on the (product|backends) tab`?

features/step_definitions/service_steps.rb
```diff
@@ -2,6 +2,10 @@
   provider.services.create! :name => name, :mandatory_app_key => false
 end
 
+Given /^a backend api "([^"]*)" of (provider "[^"]*")$/ do |name, provider|
+  provider.backend_apis.create!(:name => name, :private_endpoint => 'https://foo')
+end
```
Martouta: I propose this to avoid being so redundant with the url of the provider that does not even matter:
```
Given /^a backend api "([^"]*)")$/ do |name|
  @provider.backend_apis.create!(:name => name, :private_endpoint => 'https://foo')
end

Given /^(?:a )?default service of has name "([^"]*)"$/ do |name|
  @provider.first_service!.update_attribute(:name, name)
end
```

features/provider/dashboard/tabs.feature
```diff
@@ -0,0 +1,36 @@
...
+  In order to navigate easily to products and backends
+  As a provider
+  I want to be able to filter them by name
+
+  Background:
+    Given a provider is logged in
+    And a default service of provider "foo.example.com" has name "My Fancy Product"
+    And a service "My Regular Product" of provider "foo.example.com"
+    And a backend api "My Fancy Backend API" of provider "foo.example.com"
+    And a backend api "My Regular Backend API" of provider "foo.example.com"
```
Martouta: If you follow [this](https://github.com/3scale/porta/pull/1406/files#r344196825), then this part would be:
```
And default service has name "My Fancy Product"
And a service "My Regular Product"
And a backend api "My Fancy Backend API"
And a backend api "My Regular Backend API"
```

features/step_definitions/service_steps.rb
josemigallas: I can't be sure that would be backwards compatible. For this particular cucumber and for the step I added it would be, however keep in mind this is used by a massive number of old features.. it's kinda risky

Martouta: not risky because it is adding new stuff without removing anything :smile:

josemigallas: Right 🤦‍♂

features/step_definitions/dashboard_steps.rb
```diff
@@ -1,18 +1,20 @@
-def service_id_for_name (name)
-  page.find_by_id('apis').find('section', :text => %r{#{name}}i)[:id][/\d+/]
+# frozen_string_literal: true
+
+def service_id_for_name(name)
+  page.find_by_id('apis').find('section', :text => /#{name}/i)[:id][/\d+/]
 end
 
-def service_for_name (name)
+def service_for_name(name)
```
duduribeiro: 💯

features/step_definitions/service_steps.rb
```diff
@@ -2,6 +2,14 @@
   provider.services.create! :name => name, :mandatory_app_key => false
 end
 
+Given /^a backend api "([^"]*)")$/ do |name|
+  @provider.backend_apis.create!(:name => name, :private_endpoint => 'https://foo')
```
duduribeiro: just 💅 

but I prefer the `name: name, private_endpoint: 'https://foo'` syntax.. wdyt?

features/provider/dashboard/tabs.feature
```diff
@@ -0,0 +1,36 @@
...
+    And a service "My Fancy Product"
+    And a service "My Regular Product"
+    And a backend api "My Fancy Backend API"
+    And a backend api "My Regular Backend API"
+    And I go to the provider dashboard
+
+  @javascript
+  Scenario: Filtering products
+    When I select the products tab
+    And I should see "My Fancy Product" in the apis dashboard products tabs section
```
guicassolato: Perhaps we can improve this later with a more concise table of what should be seen, but good for now.

features/step_definitions/service_steps.rb
```diff
@@ -2,6 +2,14 @@
...
+end
+
+Given /^a service "([^"]*)")$/ do |name|
+  @provider.services.create! :name => name, :mandatory_app_key => false
```
duduribeiro: same here

features/step_definitions/dashboard_steps.rb
```diff
@@ -1,18 +1,20 @@
-def service_id_for_name (name)
-  page.find_by_id('apis').find('section', :text => %r{#{name}}i)[:id][/\d+/]
+# frozen_string_literal: true
+
+def service_id_for_name(name)
+  page.find_by_id('apis').find('section', :text => /#{name}/i)[:id][/\d+/]
```
duduribeiro: same here

features/step_definitions/service_steps.rb
Martouta: my bad, sorry, because and copied&pasted from another one :blush:

PR #1410, opened 2019-11-12
Simplify policy lists header and its styles

josemigallas:

Policy lists are the chain and the registry. Form also shares the same header.
* Extracts a `HeaderButton` component
* Simplifie styles, removing some classes and combining repeated things.

**Verification steps**:
* At Policy Chain button should be 'add':
Screen Shot 2019-11-18 at 12 54 12
* At Edit Policy or Select a Policy shuold be "cancel":
Screen Shot 2019-11-18 at 12 54 18

**Note**:
This is part 7 of the refactorization of Policies:

Previous step #1408.

### DIFF
--- a/app/javascript/src/Policies/components/HeaderButton.jsx
+++ b/app/javascript/src/Policies/components/HeaderButton.jsx
@@ -1,0 +1,37 @@
+// @flow
+
+import * as React from 'react'
+
+import type { ThunkAction } from 'Policies/types'
+
+type Props = {
+  type: 'add' | 'cancel',
+  onClick: () => ThunkAction,
+  children?: React.Node
+}
+
+const getIcon = type => {
+  switch (type) {
+    case 'add':
+      return 'fa fa-plus-circle'
+    case 'cancel':
+      return 'fa fa-times-circle'
+  }
+}
+
+const getClass = type => {
+  switch (type) {
+    case 'add':
+      return 'PolicyChain-addPolicy'
+    case 'cancel':
+      return 'PolicyChain-addPolicy--cancel'
+  }
+}
+
+const HeaderButton = ({ type, onClick, children }: Props) => (
+  <div className={getClass(type)} onClick={onClick}>
+    <i className={getIcon(type)} />{children}
+  </div>
+)
+
+export { HeaderButton }
--- a/app/javascript/src/Policies/components/HeaderButton.jsx
+++ b/app/javascript/src/Policies/components/HeaderButton.jsx
@@ -10,27 +10,19 @@ type Props = {
   children?: React.Node
 }
 
-const getIcon = type => {
-  switch (type) {
-    case 'add':
-      return 'fa fa-plus-circle'
-    case 'cancel':
-      return 'fa fa-times-circle'
-  }
+const classNames = {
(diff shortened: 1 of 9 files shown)


spec/javascripts/Policies/components/HeaderButton.spec.jsx
```diff
@@ -0,0 +1,29 @@
+// @flow
+
+import React from 'react'
+import { mount } from 'enzyme'
+
+import { HeaderButton } from 'Policies/components/HeaderButton'
+
+it('should render itself', () => {
+  const wrapper = mount(<HeaderButton type='add' onClick={jest.fn()} />)
+  expect(wrapper.find(HeaderButton).exists()).toBe(true)
```
damianpm: I think we were starting to use jest snapshots

spec/javascripts/Policies/components/HeaderButton.spec.jsx
```diff
@@ -0,0 +1,29 @@
...
+
+it('should handle clicks', () => {
+  const spy = jest.fn()
+  const wrapper = mount(<HeaderButton type='add' onClick={spy} />)
+
+  wrapper.props().onClick()
+  expect(spy).toHaveBeenCalled()
+})
+
+it('should have different stules for each type', () => {
```
damianpm: Typo: "styles"
Or maybe better to rephrase like: "should render different icons"

spec/javascripts/Policies/components/HeaderButton.spec.jsx
didierofrivia: Only when checking html structure, not just rendering

app/javascript/src/Policies/components/HeaderButton.jsx
```diff
@@ -0,0 +1,37 @@
...
+import type { ThunkAction } from 'Policies/types'
+
+type Props = {
+  type: 'add' | 'cancel',
+  onClick: () => ThunkAction,
+  children?: React.Node
+}
+
+const getIcon = type => {
+  switch (type) {
```
didierofrivia: is there anything in lodash like Ramda's cond? https://ramdajs.com/docs/#cond ?

josemigallas: Yes there is, the same function in fact.
https://www.npmjs.com/package/lodash.cond

josemigallas: It doesn't look very good IMO:
```javascript
const getIcon = type => [
  [ type === 'add', 'fa fa-plus-circle' ],
  [ type === 'cancel', 'fa fa-times-circle' ]
]
```
I changed for a dictionary, I think it's looks simpler now.

spec/javascripts/Policies/components/HeaderButton.spec.jsx
josemigallas: But you're right @damianpm, I'll add some snapshots

app/javascript/src/Policies/components/HeaderButton.jsx
didierofrivia: yeah, much better! 🏅

PR #1415, opened 2019-11-13
User#kill_user_session uses delete_all instead of destroy_all for the memory

Martouta:
As the title says :smile:


hallelujah:
Wdyt about using delete also here https://github.com/3scale/porta/blob/882567eefeae654f3bc3219bb3dbc82d587f0cf8/app/workers/user_session_sweeper_worker.rb#L8

It will be just faster than instantiating an object then deleting it

PR #1429, opened 2019-11-19
Move 'lib/signup' to 'app/lib/signup'

Martouta:
Extracted from https://github.com/3scale/porta/pull/1396
All this part has tests in `test/unit/signup` :smile: 
You can also run `bundle exec rails console` and try to use any of them. Like `Signup::SignupParams.new`


test/unit/signup/account_manager_test.rb
```diff
@@ -291,7 +291,7 @@ class PlansWithDefaultsTest < Signup::AccountManagerTest
 
       should 'yield errors to proc' do
         errors = []
-        @plans.error_proc = proc { |error| errors << error }
```
Martouta: because this was the only reason to have this:
https://github.com/3scale/porta/blob/f5bdc1a124993080add3d00099b4c7e27e438055/lib/signup/account_manager.rb#L84

as you can see, it is not called from any other class other than itself:
image

app/lib/signup/account_manager.rb
```diff
@@ -96,39 +96,21 @@ def selected=(plans)
     end
 
     delegate :[], to: :plans
-    delegate :to_a, to: :values
```
Martouta: Because there is already another `to_a` defined, and this one wouldn't even work because that `values` is not defined :joy:

guicassolato:
Better changing all class definitions to

```ruby
module Signup
  class X
    ...
  end
end
```

instead of

```ruby
class Signup::X
  ...
end
```

Martouta:
@guicassolato Done :+1: the 5th commit is exactly that :smile: :dancer:

app/lib/signup/account_manager.rb
```diff
@@ -0,0 +1,82 @@
...
+      end
+    end
+
+    private
+
+    delegate :transaction, to: :manager_account
+
+    class_attribute :account_builder, default: proc {}
+
+    # This method smells of :reek:TooManyStatements and :reek:DuplicateMethodCall
```
duduribeiro: This still smells this?

Martouta: Yup :smile: 
`:reek:DuplicateMethodCall` for 2 times `raise ActiveRecord::Rollback`, and `:reek:TooManyStatements` because it expects a method to have less lines.

PR #1430, opened 2019-11-19
Move PermalinkFu into 'app' folder

Martouta:
PermalinkFu generates a [permalink](https://en.wikipedia.org/wiki/Permalink) for Forum and for Topic.
This in the `master` branch looks like this in the models:
https://github.com/3scale/porta/blob/90b79d87d93dfb552e9ddae3c7fc3d187a090e2a/app/models/forum.rb#L8
https://github.com/3scale/porta/blob/90b79d87d93dfb552e9ddae3c7fc3d187a090e2a/app/models/topic.rb#L67

### What each commit does
1. `mv lib/permalink_fu.rb app/lib/permalink_fu.rb`
2. `mv test/unit/permalink_fu_test.rb test/unit/lib/permalink_fu_test.rb`
3. All the actual changes inside the files :smile:
4. `mv test/unit/lib/permalink_fu_test.rb test/unit/permalink_fu_test.rb`
5. Ensure Permalink is always auto-generated and not written from the outside.

#### Changes in the file(s)
- `PermalinkFu` becomes a `concern` included in the 2 models that actually used instead of `ActiveRecord::Base.send :include, PermalinkFu` in an initializer. It additionally allows to use a less verbose include :smile: 
- Rename `has_permalink` -> `permalink`. DLS-friendly :smile:
- Removes the class attribute and option `permalink_field`. Because in our case it is always just `:permalink`.
- The class attribute `permalink_attributes` becomes `permalink_attribute` in singular. Because that is all we need. And that also allows simplifying the code very much.
- Remove the `Validator` class because reached this point is not actually validating anything but just adding an error message if permalink is empty (not generated).
- Remove `translation_to` and `translation_from` considering that they are not used since the 27th of September of 2013 when we stopped using the `iconv` gem.
- Remove the line `require 'digest/sha1'` because it is already required without being so specific and in this case, it is not used anyway.
- The errors are not in `base` anymore but in the `permalink_field`. So before the error was something like `:base, name must contain latin characters` and now it became `:name, must contain latin characters`. It is more correct and it does not break any current integration because there aren't API endpoints for forums or topics.
- Test tests everything we need. It was lacking some tests before. Besides, it tests for the classes and scenarios that we actually need instead of stubbing the `where` and returning what we would like it to return because that is too error-prone.
- It simplifies the code in general.
- It also fixes specifically this part:
https://github.com/3scale/porta/blob/90b79d87d93dfb552e9ddae3c7fc3d187a090e2a/lib/permalink_fu.rb#L84-L90
Now what I've done to make it safe and as efficient as I could is instead of doing a query to the DB for every possible number, do it with the regex directly.
Some information: There are actually some forums in the database with the same permalink. I guess that happened before creating this validation of having to be unique.


app/lib/permalink_fu.rb
```diff
@@ -0,0 +1,56 @@
...
+  private
+
+  def create_unique_permalink
+    base_permalink = permalink.presence || build_permalink_from_attributes
+    self.permalink = generate_unique_permalink(base_permalink)
+  end
+
+  def generate_unique_permalink(base_permalink)
+    # TODO: CONCEPTS THAT I MUST CHECK:
+    # 1: If the permalink must be unique, why doing count of the exact string? it must be always 0 or 1 then...
```
Martouta: And why there is not a unique validation neither in the model nor in the schema?

app/lib/permalink_fu.rb
```diff
@@ -0,0 +1,45 @@
...
+      validate :permalink_must_contain_latin_characters
+    end
+  end
+
+  private
+
+  def create_unique_permalink
+    base_permalink = permalink.presence || build_permalink_from_attribute
+    count = where_match_permalink_with_conditions(base_permalink).count
+    self.permalink = count.positive? ? "#{base_permalink}-#{count + 1}" : base_permalink
```
Martouta: I should ensure that this one is unique :smile:

guicassolato: This is not thread-safe.

app/lib/permalink_fu.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+module PermalinkFu
+  extend ActiveSupport::Concern
+
+  included do
+    def permalink_must_contain_latin_characters
+      return if permalink.present?
```
guicassolato: If `permalink == '危険'`, then permalink contains latin characters? Weird.

app/models/application_record.rb
```diff
@@ -8,6 +8,17 @@ def self.user_attribute_names
     attribute_names
   end
 
+  sifter(:regexp) do |column, matcher|
```
guicassolato: Nice 😍

test/unit/lib/permalink_fu_test.rb
```diff
@@ -0,0 +1,86 @@
+# frozen_string_literal: true
```
guicassolato: `test/unit/lib` is a Pandora's box. It's like having a `others` dir. I know it wasn't this PR the one to open the box, but please let's close it before it's too late. We have `app/lib` for convenience while moving things from `lib`, but even that one needs to be better organised IMO.

app/lib/permalink_fu.rb
Martouta: I know but what can I do about it? I kept the same behaviour we had and I only made it more efficient and clear.
I cannot add a uniqueness validation because as I said in the description of the PR:

> There are actually some forums in the database with the same permalink. I guess that happened before creating this validation of having to be unique.

Martouta: And `forum` does not have a `created_at` to add that validation only for new models 😞

guicassolato: I guess it’s not worth the effort of trying to fix. As you said, it’s been like this since forever and AFAIR forum is not something to last very long.

app/lib/permalink_fu.rb
Martouta: We shouldn't allow to do that. Permalink must be only autogenerated.
I will ensure that :+1:

PR #1432, opened 2019-11-19
[WIP] Move SidekiqLockWorker into the 'app' folder

Martouta:
Extracted from https://github.com/3scale/porta/pull/1396

The purpose of this file is, as it says in a comment inside it:

```
# We don't actually need this in production. However, in the tests sidekiq-lock's `#lock` returns always `nil` due to
  # Sidekiq::Middleware not being active in this environment and so the thread instance variable that holds the lock is
  # never set. With effect, the tests fail with `undefined method acquire! for nil:NilClass` even though the
  # implementation is correct.
  #
  # Sidekiq::Middleware could be activated or sidekiq-lock test helpers used instead, but since we have so many
  # different types of tests, it would be either tedious to make it work to all of them or we could end up not testing
  # the actual implementation.
```

You can test that it works by running `bundle exec m test/workers/backend_metric_worker_test.rb`.  With the file, it passes, without the file, it raises errors 💃 

However, since this is only for the tests, it may be better to have it as a test helper or something like that :thinking: What do you think @3scale/system ?


guicassolato:
It's not that we don't _need_ that file in production, because we do.

The `ThreeScale::SidekiqLockWorker` module is a hack based upon https://github.com/emq/sidekiq-lock. The usual way to use this gem is by including `Sidekiq::Lock::Worker` in the workers and, in the tests, requiring `sidekiq/lock/testing/inline`, so `#set_sidekiq_lock` and `#clear_sidekiq_lock` helpers are available.

We didn't want to use the helpers methods in the tests because it's a different implementation that what runs in production. In theory, we should trust the abstraction, I know, but the gem was never mature enough.

At the other extreme of the helper methods, we could have activated `Sidekiq::Lock::Middleware` in the test, but that seemed too much back then. So we went with the hack "for the time being".

Today, I'd rather drop this lock-specific gem and maybe try something with https://github.com/sensortower/sidekiq-throttled. I guess we can use the following `sidekiq_throttle` option to set the lock at the same level of what we have now:

```ruby
concurrency: { limit: 1, key_suffix: ->(*args) { BackendMetricWorker.lock_workers(*args) } }
```

Furthermore, https://github.com/sensortower/sidekiq-throttled would also give us some throttling possibilities in other processes where rate limiting background jobs make sense, e.g. transactions with payment gateways.

Martouta:
Ahaaaaaa, nice, thanks @guicassolato :muscle: I will try :+1:

test/workers/backend_metric_worker_test.rb
```diff
@@ -1,77 +0,0 @@
-# frozen_string_literal: true
-
-require 'test_helper'
-
-class BackendMetricWorkerTest < ActiveSupport::TestCase
```
Martouta: I won't remove this for real, it is just temporarily to see what happens in CircleCI and I am rewriting it locally :wink:

Martouta:
Closing in favor of https://github.com/3scale/porta/pull/1498 because it has become a very different approach.

PR #1434, opened 2019-11-19
Enable Policies update button after rearranging chain

josemigallas:

When polices change their place within the chain, the update button is enabled.

[THREESCALE-3941: Policies cannot be updated after reordering them](https://issues.jboss.org/browse/THREESCALE-3941)

**Verification steps** 

1. In Product-> Integration-> Policies-> add multiple policies and update
2. Reload policies page and try to reorder policies
3. Try to click on "Update Policy Chain"

### DIFF
--- a/app/javascript/src/Policies/util.jsx
+++ b/app/javascript/src/Policies/util.jsx
@@ -52,12 +52,14 @@ function parsePolicy (key: string, policy: RawPolicy): RegistryPolicy {
 }
 
 function isPolicyChainChanged (chain: ChainPolicy[], originalChain: ChainPolicy[]) {
-  if (originalChain.length !== chain.length) {
+  const chainLength = chain.length
+  if (originalChain.length !== chainLength) {
     return true
   }
 
-  for (const policy of chain) {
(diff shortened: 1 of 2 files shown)


app/javascript/src/Policies/util.jsx
```diff
@@ -56,8 +56,9 @@ function isPolicyChainChanged (chain: ChainPolicy[], originalChain: ChainPolicy[
     return true
   }
 
-  for (const policy of chain) {
-    const originalPolicy = originalChain.find(p => p.uuid === policy.uuid)
+  for (let i = 0; i < chain.length; i++) {
```
Martouta: These options are more efficient:

```suggestion
  for (let i = 0, length = chain.length; i < length; i++) {
```

```javascript
  for (let i = chain.length - 1; i >= 0 ; i--) {
```

josemigallas: Good! However the length can be declared even earlier. I'll change it.

damianpm: +1 for length caching

PR #1437, opened 2019-11-19
Automatically fetch model error message when fails to create service

Martouta:
This is my proposal for https://github.com/3scale/porta/pull/1398/files#r347518158
This way the result is exactly the same, but we don't have to manually add/edit/remove every validation from the model also in the locale, because many are already done by ActiveRecord and the less repetition, the less changes to mess up from us while maintaining the code 😄

### DIFF
--- a/app/controllers/api/services_controller.rb
+++ b/app/controllers/api/services_controller.rb
@@ -48,7 +48,7 @@ class Api::ServicesController < Api::BaseController
       onboarding.bubble_update('api')
       redirect_to admin_service_path(@service)
     else
-      flash.now[:error] = flash_message(@service.errors.keys.first, {scope: %i[flash services create errors]})
+      flash.now[:error] = @service.errors.full_messages.to_sentence.presence || I18n.t!('flash.services.create.errors.default', {resource_type: product_or_service_type})
       activate_menu :dashboard
       render :new
     end
@@ -153,12 +153,5 @@ class Api::ServicesController < Api::BaseController
 
   def authorize_admin_plans
     authorize! :admin, :plans
-  end
-
-  def flash_message(key, opts = {})
-    options = opts.reverse_merge(resource_type: product_or_service_type)
-    I18n.t!(key, options)
-  rescue I18n::MissingTranslationData
-    t(:default, options)
   end
 end
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -886,6 +886,11 @@ en:
         service_plan:
           has_contracts: 'This service plan cannot be deleted! At least one contract depends on it.'
 
+        service:
+          attributes:
+            system_name:
+              invalid: invalid. Only ASCII letters, numbers, dashes and underscores are allowed.
+
         cinstance:
           attributes:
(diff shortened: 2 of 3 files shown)


test/integration/api/services_controller_test.rb
```diff
@@ -174,7 +174,7 @@ def update_params
     end
   end
 
-  class BackendApiCreation < Api::ServicesControllerTest
+  class BackendApiCreationTest < Api::ServicesControllerTest
```
Martouta: Unrelated to this PR, but all Test classes should be named ending by `Test` to avoid duplication with the logic classes

app/controllers/api/services_controller.rb
```diff
@@ -48,7 +48,7 @@ def create
       onboarding.bubble_update('api')
       redirect_to admin_service_path(@service)
     else
-      flash.now[:error] = flash_message(@service.errors.keys.first, {scope: %i[flash services create errors]})
+      flash.now[:error] = @service.errors.full_messages.to_sentence.presence || I18n.t!('flash.services.create.errors.default', {resource_type: product_or_service_type})
```
Martouta: @didierofrivia `@service.errors.full_messages.to_sentence.presence` fetches them all and makes it a sentence, but alternatively you can also pick only the first one by `@service.errors.full_messages.first.presence`

config/locales/en.yml
```diff
@@ -886,6 +886,11 @@ en:
         service_plan:
           has_contracts: 'This service plan cannot be deleted! At least one contract depends on it.'
 
+        service:
+          attributes:
+            system_name:
+              invalid: invalid. Only ASCII letters, numbers, dashes and underscores are allowed.
```
Martouta: This is because the default message is just 'invalid'. But this way we are allowed to override any message we want for any specific error and not stick to whatever message for the entire attribute.

test/integration/api/services_controller_test.rb
```diff
@@ -210,4 +210,19 @@ def setup
...
+      post admin_services_path, service: { name: '' }
+      assert_equal 'Couldn\'t create Product. Check your Plan limits', flash[:error]
+
+      @provider.settings.allow_multiple_services!
+
+      post admin_services_path, service: { name: '' }
+      assert_equal 'Name can\'t be blank', flash[:error]
+
+      post admin_services_path, service: { name: 'example-service', system_name: '###' }
+      assert_equal 'System name invalid. Only ASCII letters, numbers, dashes and underscores are allowed.', flash[:error]
```
Martouta: @didierofrivia I also respected your cucumber for you to see that it passes: https://github.com/3scale/porta/pull/1398/files#diff-9186c909b21b5e5da01c4f9ab0d72b89R40

But I admit that I felt tempted to remove it because our cukes 🥒 usually become unmaintainable nightmares 😄

app/controllers/api/services_controller.rb
didierofrivia: Ok, cool! Looks good to me

PR #1419, opened 2019-11-15
Redefine Policies types

josemigallas:
**What this PR does / why we need it**:

Big refactor of types, removing duplications and using names as much clear as possible:
* Policies return by rails is an object of `RawRegistryPolicy`. It is parsed and added into the registry component as an array of `IPolicy`
* policies_config is passed as props as an array of `PolicyConfig`. Then they are parsed and added to the chain, copied from the registry -> `ChainPolicy`
* Some variable/prop names also changed for more consistency and clarity

**Verification steps**:
* `npm run update-dependencies`
* Navigate to `apiconfig/services/<id>/policies/edit`
* Policies widget should work as usual

**Note**:
This is part 8 of the refactorization of Policies:
[THREESCALE-2221: Policies: Normalise the use of policies schema and its types](https://issues.jboss.org/browse/THREESCALE-2221)

Previous step #1410.

PR #1438, opened 2019-11-19
Policies/parse registry in rails

josemigallas:

The registry is being returned in a weird format. The frontend is expecting an array but it is insteaf a dictionary. This PR moves the "parsing" logic to the controller (perhaps a decorator would be better) in order to simplify the frontend logic.

**Verification steps**:

**Note**:
This is part 9 of the refactorization of Policies:

Previous step #1419.

### DIFF
--- a/app/javascript/src/Policies/types/Policies.jsx
+++ b/app/javascript/src/Policies/types/Policies.jsx
@@ -11,22 +11,14 @@ export type PolicyConfig = {
   enabled: boolean,
 }
 
-// Represents each of the items contained in the registry object
-// returned by rails (@registry_policies)
-export type RawRegistryPolicy = {
+// Represents policies of the Registry
+export type RegistryPolicy = {
   $schema: string,
   configuration: Configuration,
   description: [string],
   name: string,
   summary: string,
-  version: string
-}
-
-// Represents the registry object returned by the server (@registry_policies)
-export type RawRegistry = { [string]: RawRegistryPolicy[] }
-
-// Represents policies of the Registry
-export type RegistryPolicy = RawRegistryPolicy & {
+  version: string,
   data?: Configuration,
   humanName: string
 }
--- a/app/presenters/policies_list_presenter.rb
+++ b/app/presenters/policies_list_presenter.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+class PoliciesListPresenter
+  def initialize(policies)
+    @policies = policies
+  end
+
+  def registry
+    @policies.map do |key, value|
+      policy = value.first
+      policy.merge :name => key, :humanName => policy['name'], :data => {}
+    end
+  end
+end
--- a/test/integration/api/policies_controller_test.rb
+++ b/test/integration/api/policies_controller_test.rb
@@ -7,7 +7,7 @@ class Api::PoliciesControllerTest < ActionDispatch::IntegrationTest
     @provider = FactoryBot.create(:provider_account)
     @service = @provider.default_service
     login! @provider
-    Policies::PoliciesListService.expects(:call!)
+    Policies::PoliciesListService.expects(:call!).returns({})
   end
 
 
--- a/app/controllers/api/integrations_controller.rb
+++ b/app/controllers/api/integrations_controller.rb
@@ -137,10 +137,7 @@ class Api::IntegrationsController < Api::BaseController
 
   def find_registry_policies
     policies_list = Policies::PoliciesListService.call!(current_account, proxy: @proxy)
-    @registry_policies = policies_list.map do |key, value|
-      policy = value.first
-      policy.merge :name => key, :humanName => policy['name'], :data => {}
-    end
+    @registry_policies = PoliciesListPresenter.new(policies_list).registry
   rescue StandardError => error
     @error = error
   end
--- a/app/controllers/api/integrations_controller.rb
+++ b/app/controllers/api/integrations_controller.rb
@@ -136,7 +136,11 @@ class Api::IntegrationsController < Api::BaseController
   protected
 
   def find_registry_policies
-    @registry_policies ||= Policies::PoliciesListService.call!(current_account, proxy: @proxy)
+    policies_list = Policies::PoliciesListService.call!(current_account, proxy: @proxy)
+    @registry_policies = policies_list.map do |key, value|
+      policy = value.first
+      policy.merge :name => key, :humanName => policy['name'], :data => {}
+    end
   rescue StandardError => error
     @error = error
   end
--- a/app/controllers/api/policies_controller.rb
+++ b/app/controllers/api/policies_controller.rb
@@ -42,7 +42,11 @@ class Api::PoliciesController < Api::BaseController
   end
 
   def find_resources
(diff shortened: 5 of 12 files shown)


app/presenters/policies_list_presenter.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class PoliciesListPresenter
+  def initialize(policies)
+    @policies = policies
+  end
+
+  def get_registry
```
Martouta: We never name in ruby a method like `get_...`, we just put what it returns or does directly.

app/presenters/policies_list_presenter.rb
```diff
@@ -0,0 +1,15 @@
...
+    return nil unless @policies
```
Martouta: It should never be in real execution time, the only time it can be nil is from the tests for this line:
https://github.com/3scale/porta/blob/6336b1d300bdf28252b500ff294c550fc6514e60/test/integration/api/policies_controller_test.rb#L10
But because it is being mocked without returning anything, which cannot happen in real life because if it does not return what we need it is because there has been an error raised and then it does not even reach here anyway.
I think that we should adapt the test to the real code instead of type checking unnecessarily

app/presenters/policies_list_presenter.rb
```diff
@@ -0,0 +1,15 @@
...
+    @policies.map do |key, value|
+      policy = value.first
+      policy.merge :name => key, :humanName => policy['name'], :data => {}
```
Martouta: I see that the purpose of this object is only this, but can't this be done in the JS execution directly so we don't have to loop it twice?

app/presenters/policies_list_presenter.rb
josemigallas: Busted! 🤣

app/presenters/policies_list_presenter.rb
josemigallas: Fair enough

app/presenters/policies_list_presenter.rb
josemigallas: Of course it can but it shouldn't... the less computation in the browser the better. What do you mean by _loop it twice_? Where is the first loop?

app/javascript/src/Policies/types/Policies.jsx
```diff
@@ -11,22 +11,14 @@ export type PolicyConfig = {
   enabled: boolean,
 }
 
-// Represents each of the items contained in the registry object
-// returned by rails (@registry_policies)
-export type RawRegistryPolicy = {
+// Represents policies of the Registry
+export type RegistryPolicy = {
```
guicassolato:
```suggestion
export type PolicyRegistry = {
```

app/presenters/policies_list_presenter.rb
Martouta: > the less computation in the browser the better.

Good point :+1: Although the only computation for each element (the fact of being in a loop is another story that I will comment later) is the `:humanName => policy['name']`, but I trust your criteria on this.

> What do you mean by loop it twice? Where is the first loop?

I meant looping once here for this and later in the JS side in checking each element one by one to render them :smile:

josemigallas: Oh but the point is precisely that: with this presenter, there will be no loop in JS. This is to prepare the data to be consumed directly by JS, which reduce the complexity in the React part significantly.

Martouta: Awesome :pray: :muscle: :1st_place_medal: 

In that case, I suggest keeping this Presenter but with the 2 improvements I mentioned above:

```ruby
# frozen_string_literal: true

class PoliciesListPresenter
  def initialize(policies)
    @policies = policies
  end

  def registry
    @registry ||= policies.map do |key, value|
      policy = value.first
      policy.merge { name: key, humanName: policy['name'], data: {} }
    end
  end

  private

   attr_reader :policies
end
```

test/integration/api/policies_controller_test.rb
```diff
@@ -7,7 +7,7 @@ def setup
     @provider = FactoryBot.create(:provider_account)
     @service = @provider.default_service
     login! @provider
-    Policies::PoliciesListService.expects(:call!)
+    Policies::PoliciesListService.expects(:call!).returns([])
```
guicassolato: This smells. It mock an invalid behaviour, as `Policies::PoliciesListService` won't ever return an `Array`.

Martouta: @guicassolato the thing is that it either returns an array or raises an error, but it never returns `nil` as it was doing before this change... and adapting the real execution code to something that won't happen in real execution outside the tests just for this mock is even worse imo

guicassolato: I think it never return an array.

josemigallas: It should be an object, my bad!

guicassolato: It should be a `String`.

P.S.: In Ruby, it's always an object, even when it's an `Array` ;)

Martouta: Right, it is a json/hash, not an Array, my bad :pray: So it should be its actual response :wink: but definitely not `nil` as it can never happen in real life.

test/integration/api/policies_controller_test.rb
```diff
@@ -7,7 +7,7 @@ def setup
     @provider = FactoryBot.create(:provider_account)
     @service = @provider.default_service
     login! @provider
-    Policies::PoliciesListService.expects(:call!)
+    Policies::PoliciesListService.expects(:call!).returns({})
```
guicassolato:
```suggestion
    Policies::PoliciesListService.expects(:call!).returns({}.to_json)
```

test/integration/api/policies_controller_test.rb
guicassolato: Sorry, a `Hash`, not a `String`

josemigallas: > It should be a `String`.
> 
> P.S.: In Ruby, it's always an object, even when it's an `Array` ;)

Yes I meant a hash (javascript speaker here!)

app/presenters/policies_list_presenter.rb
```diff
@@ -0,0 +1,14 @@
...
+
+class PoliciesListPresenter
+  def initialize(policies)
+    @policies = policies
+  end
+
+  def registry
+    @policies.map do |key, value|
+      policy = value.first
+      policy.merge :name => key, :humanName => policy['name'], :data => {}
```
guicassolato: Let's pretend this lower camel case here didn't happen :)

test/integration/api/policies_controller_test.rb
```diff
@@ -7,7 +7,7 @@ def setup
     @provider = FactoryBot.create(:provider_account)
     @service = @provider.default_service
     login! @provider
-    Policies::PoliciesListService.expects(:call!)
+    Policies::PoliciesListService.expects(:call!).returns({}.to_json)
```
guicassolato: No! I was wrong and deleted the comment. Just `{}` is fine.

```suggestion
    Policies::PoliciesListService.expects(:call!).returns({})
```

PR #1442, opened 2019-11-19
Move lib/liquid to app/lib

duduribeiro:
Extracted from #1396

lib/liquid into app/lib


app/lib/liquid/template_support.rb
```diff
@@ -74,29 +76,24 @@ def render_template(context, options)
...
-      original = {}
-
       @_assigned_drops ||= {}
       @_template_assigns ||= {}
 
       report_and_supress_exceptions do
-        diff = original.keys - @_assigned_drops.keys
         overriden = @_template_assigns.keys & @_assigned_drops.keys
 
-        if diff.present?
```
hallelujah: why removing that?

duduribeiro: this code is never executed. diff is always empty

PR #1443, opened 2019-11-19
Move lib/payment into app/lib

duduribeiro:
Extracted from #1396

lib/payment into app/lib

### DIFF
--- a/app/lib/payment/braintree_blue_errors_handler.rb
+++ b/app/lib/payment/braintree_blue_errors_handler.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module Payment
   class BraintreeBlueErrorsHandler < AbstractErrorsHandler
     class BraintreeResultError < StandardError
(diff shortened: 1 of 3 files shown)


duduribeiro:
How to test: All Payments should be still working

app/lib/payment/braintree_blue_errors_handler.rb
```diff
@@ -23,17 +25,17 @@ def messages
...
       messages.compact!
       messages
     end
 
     # This will notify that some errors are not sent by Braintree
-    def notify_no_errors_given!
-      return if @result.errors.any?
-      error = BraintreeResultError.new @result
+    def notify_no_errors_given
+      return if result.errors.any?
```
Martouta: Is there an attr_reader for result defined? :scream:

Martouta: Yup, you are right, there is :smile: 
image
image

PR #1445, opened 2019-11-20
[oracle] Fix creation of unique index on metrics(owner_type, owner_id, system_name)

guicassolato:
The index uses columns recently added combined with existing system_name. This makes Oracle to activate the uniqueness constraint and fail if the owner_id and owner_type haven't yet been backfilled.

Closes [THREESCALE-3944](https://issues.jboss.org/browse/THREESCALE-3944)

Alternative to https://github.com/3scale/porta/pull/1444.

### DIFF
--- a/db/migrate/20190805135839_add_index_to_metrics_owner_and_system_name.rb
+++ b/db/migrate/20190805135839_add_index_to_metrics_owner_and_system_name.rb
@@ -2,7 +2,7 @@ class AddIndexToMetricsOwnerAndSystemName < ActiveRecord::Migration
   disable_ddl_transaction!
 
    def change
-    return if ActiveRecord::SchemaMigration.find_by(version: 20190805135730) # this migration was repositioned back
+    return if ActiveRecord::Base.connection.index_exists?(:metrics, [:owner_type, :owner_id, :system_name])
 
     index_options = { unique: true }
     index_options[:algorithm] = :concurrently if System::Database.postgres?
--- a/db/migrate/20190805135839_add_index_to_metrics_owner_and_system_name.rb
+++ b/db/migrate/20190805135839_add_index_to_metrics_owner_and_system_name.rb
@@ -1,0 +1,11 @@
+class AddIndexToMetricsOwnerAndSystemName < ActiveRecord::Migration
+  disable_ddl_transaction!
+
+   def change
+    return if ActiveRecord::SchemaMigration.find_by(version: 20190805135730) # this migration was repositioned back
+
+    index_options = { unique: true }
+    index_options[:algorithm] = :concurrently if System::Database.postgres?
+    add_index :metrics, [:owner_type, :owner_id, :system_name], index_options
+  end
+end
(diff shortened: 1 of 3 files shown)


db/migrate/20190805135839_add_index_to_metrics_owner_and_system_name.rb
```diff
@@ -0,0 +1,11 @@
+class AddIndexToMetricsOwnerAndSystemName < ActiveRecord::Migration
+  disable_ddl_transaction!
+
+   def change
+    return if ActiveRecord::SchemaMigration.find_by(version: 20190805135730) # this migration was repositioned back
```
hallelujah: That is a nice trick playing with the timestamp of migration, to make it work for already migrated environments

Martouta: I will write what I think here, but I am not entirely sure so feel free to ignore this comment if you consider it appropriate after reading it. Because I know that this is urgent and I will have to leave early today.

image

image


image

I don't understand... where the hell is the migration that u r trying to reference? it is my first `find` and it is not anywhere for me :confused: 

I guess that you actually meant to reference `20190805135839_add_index_to_metrics_owner_and_system_name.rb` (`20190805135839`) or `db/migrate/20190716110000_add_owner_to_metrics.rb` (`20190805135839`).

Anyway, I find this dangerous for 2 reasons, the first one is kinda small and subjective but the second one actually worries me:

1. By typing a number, it is too easy to mess up, and even if you don't mess up, it is not straightforward to understand what this number stands for (which migration in this case), especially if someone later removes the file of the migration that you are referencing to (yes, it can be searched anyway, but it is not the simplest and fastest thing ever).

2. We've seen that at some point later in time, a migration, even an old one, changes, and you have an example of this in this same PR changing `db/migrate/20190716110000_add_owner_to_metrics.rb`:
image

Although in this case, it isn't a big deal because the only people who have executed are ourselves in SaaS, and in this case is under control, but just giving you an example of this because this sometimes happens.

Another example of indirectly changing old migrations is the other possible approach you took for this PR, by changing this:
image

What I mean with this is that if a migration is executed before these changes, it does one thing, if it is executed later, it does something else, and here you are skipping X migration no matter if it was executed before the change or after the change, and keeping in mind all possible changes in time sounds like a huge nightmare to me.

By that, I mean that for me the safest approach (and also more understandable for what I mentioned in first concern) is instead of checking if X migration has been executed, I would look in the DB directly if the index exists, if it does, then return, if it doesn't, create it this way.

Just my 2 cents, but as I said at the beginning of the comment, feel free to ignore this if it is urgent because I won't be here later today :+1:

Martouta: Change of plans, I will probably be here when you read it :+1:

guicassolato: @Martouta, in this universe, the migration 20190805135730 no longer exists. However in the past of Saas but also to some of the countless quantum dimensions out there, the migration is very alive, either written in the books of history or in a future still to come. Either way, time traveling is dangerous, my friend, as so as playing with Rails migrations in a Git dominated world.

Martouta: Troll mode on, I see 😝

Martouta: image

It can be found but for the reasons I exposed above, I feel it is better to do "return if the index already exists". It is clearer and safer :smile:

PR #1452, opened 2019-11-25
Skip ModelsTest for all ActiveRecord classes declared inside the Tests

Martouta:
Fixes this random bug:
image

In [this case](https://app.circleci.com/jobs/github/3scale/porta/131905), it is because it was trying to test this class:
https://github.com/3scale/porta/blob/00f267bb5cf8dbdbe05e36a2071e8345c31e5516/test/unit/provider_constraints_test.rb#L5-L11

But the purpose of this test is not for classes inside the tests, which is only problematic, so I am skipping them all here :smile:

### DIFF
--- a/test/unit/models_test.rb
+++ b/test/unit/models_test.rb
@@ -18,18 +18,18 @@ class ModelsTest < ActiveSupport::TestCase
       'Profile' => %w[logo_file_name logo_content_type state], 'UserSession' => %w[key user_agent], 'CMS::Partial' => %w[content_type],
       'CMS::EmailTemplate' => %w[content_type], 'CMS::Builtin' => %w[title], 'CMS::Builtin::StaticPage' => %w[title], 'CMS::Builtin::Page' => %w[title content_type],
       'CMS::Builtin::Partial' => %w[title content_type], 'CMS::Portlet' => %w[content_type], 'CMS::Builtin::LegalTerm' => %w[title content_type],
-      'CMS::PortletTest::CustomPortlet' => :all, 'CMS::Portlet::Base' => %w[content_type], 'ExternalRssFeedPortlet' => %w[content_type],
+      'CMS::Portlet::Base' => %w[content_type], 'ExternalRssFeedPortlet' => %w[content_type],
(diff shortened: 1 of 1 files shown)


test/unit/models_test.rb
```diff
@@ -18,18 +18,18 @@ class ModelsTest < ActiveSupport::TestCase
...
-      'Account' => %w[credit_card_auth_code credit_card_authorize_net_payment_profile_token credit_card_partial_number],
-      'DeadlockTest::Model' => :all, 'ThreeScale::SearchTest::Model' => :all
+      'Account' => %w[credit_card_auth_code credit_card_authorize_net_payment_profile_token credit_card_partial_number]
     }
 
     Rails.application.eager_load!
     models = ActiveRecord::Base.descendants - [BackendApi, Service, Proxy]
 
     validate_columns_for = ->(model, options = {}) do
+      next if model.name.match(/^.+Test::.+$/)
```
hallelujah: raise ?

Martouta: why raise? I don't want to raise anything, I just do not want to test those classes... :confused:

Martouta: because the purpose of this test is to ensure that all strings have validations of length so when the customer tries to use it, it does not crash in the DB and we return a proper error.
But that does not apply to the classes inside Test

hallelujah: OK but thing is that maybe some models are not defined with `Test` as suffix

Sadly there is no `source_location` for class to see where it is defined
Best is to look for ActiveRecord or ApplicationRecord in test / spec features and see if we define some models there and suffix them with Test

It is not the case here so my guess is that random failure is still not solved:

```shell
hery@Tiburon ~/Devel/3scale/porta (master●)$ rg '< ActiveRecord::Base' test                                                                                                                                                             [2.4]
test/unit/deadlock_test.rb
6:  class Model < ActiveRecord::Base

test/unit/fields/fields_test.rb
4:  class Model < ActiveRecord::Base

test/unit/three_scale/search_test.rb
5:  class Model < ActiveRecord::Base
```

Martouta: https://github.com/3scale/porta/pull/1461

PR #1455, opened 2019-11-26
[WIP]Updating snapshots when running jest

damianpm:

Recently we started using jest snapshots to test React components, but snapshots are not being updated automatically and jest is failing in ci.

Adding `--updateSnapshot` to our `jest` npm script.

**Verification steps** 

`npm run jest`

Expected result: All jest test must be green, and snapshots updated when needed

### DIFF
--- a/spec/javascripts/Settings/components/Common/__snapshots__/SelectGroup.spec.jsx.snap
+++ b/spec/javascripts/Settings/components/Common/__snapshots__/SelectGroup.spec.jsx.snap
@@ -145,6 +145,7 @@ exports[`should render correctly 1`] = `
             ariaLabelledBy=""
             className=""
             createText="Create"
+            customContent={null}
             direction="down"
             id="service_proxy_attributes_jukebox_select"
             isCreatable={false}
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -12,7 +12,7 @@
     "flow:install": "flow-typed install",
     "lint": "flow && eslint --ext .js,.jsx app/javascript spec/javascripts",
     "test": "karma start --single-run",
-    "jest": "jest",
+    "jest": "jest --updateSnapshot",
```
josemigallas: ⚠️I'm not sure about this, I think it will make the snapshot check useless. The snapshots will be updated every time by CircleCI, therefore they will _always_ be correct. I think the idea of the snapshots is that _we_ manually take care of them. But I can be wrong,  @didierofrivia ?

damianpm: The problem is right now we are supposed to update them manually and we are forgetting, tests are failing when already merged in master.
If we add it here, they will fail in our PR's, when circle ci runs `npm run jest`, not in master.

josemigallas: I think it's the other way around, if CircleCI updates the snapshot it will never fail. It runs `jest` as a part of the javascript pipeline

didierofrivia: We should ***never*** auto update the templates. We should first identify why it didn't failed in the PR merged and then act accordingly.

damianpm:
Fixed in https://github.com/3scale/porta/pull/1454

PR #1458, opened 2019-11-26
Suspend account if payment failures are to high

duduribeiro:
[THREESCALE-3947](https://issues.jboss.org/browse/THREESCALE-3947)

We have a problem that an attacker can try to find a valid CVV for a specific credit card making too many requests in the credit card validation endpoint. This commit prevents this type of attack from suspending the account if the failure count is greater than the threshold (currently 10). If a successful request is made, it will reset the failure count.


app/models/payment_gateway_setting.rb
```diff
@@ -37,4 +39,25 @@ def active_gateway_type
       errors.add(:gateway_type, :invalid)
     end
   end
+
+  def increment_failure_count
+    lock!
+    incremented_failure_count = failure_count + 1
+    self.gateway_settings = symbolized_settings.merge({failure_count:  incremented_failure_count})
```
hallelujah: I was thinking of using ActiveRecord::Store as a refactoring

`store :gateway_settings, accessors: some_stuffs`

app/controllers/provider/admin/account/payment_gateways/braintree_blue_controller.rb
```diff
@@ -48,6 +48,7 @@ def hosted_success
       end
 
     else
+      current_account.mark_as_failed_and_suspend_when_too_much_payment_failure
```
hallelujah: This can be inside a Service object and invoked after_action right?

app/models/account/gateway.rb
```diff
@@ -88,4 +88,13 @@ def log_payment_gateway_change
                       "changed payment gateway to #{payment_gateway_type}")
   end
 
+  def mark_as_failed_and_suspend_when_too_much_payment_failure
```
hallelujah: I think we should extract this to its own Service object

app/models/payment_gateway_setting.rb
```diff
@@ -37,4 +39,25 @@ def active_gateway_type
       errors.add(:gateway_type, :invalid)
     end
   end
+
+  def increment_failure_count
+    lock!
```
hallelujah: This is a pessimistic lock we should try to avoid that unless we are sure that under spam conditions it does not end up with deadlocks or a hold query. This will for sure make the service down as the unicorn will all timeouts

app/models/payment_gateway_setting.rb
```diff
@@ -37,4 +39,18 @@ def active_gateway_type
       errors.add(:gateway_type, :invalid)
     end
   end
+
+  def increment_failure_count
+    incremented_failure_count = failure_count + 1
+    self.gateway_settings = symbolized_settings.merge({failure_count:  incremented_failure_count})
```
guicassolato: This doesn't look thread-safe.

guicassolato: Maybe not an issue, but at least it can explain in case we once realize `FAILURE_THRESHOLD` is not being ensured.

app/services/payment/spam_protection_service.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
+module Payment
+  class SpamProtectionService
+    MAX_REQUESTS_IN_A_PERIOD = 10
```
guicassolato: I think we need to make this more scalable as 10 may end up being a short limit for some enterprise accounts.

app/services/payment/spam_protection_service.rb
```diff
@@ -0,0 +1,56 @@
...
+      total_requests > MAX_REQUESTS_IN_A_PERIOD
+    end
+
+    private
+
+    attr_accessor :account, :gateway_setting, :payment_status, :user_session
+
+    def increment_total_requests
+      redis.incr(key)
+      redis.expire(key, 1.hour)
```
guicassolato: https://github.com/3scale/porta/pull/1458/files#r353282385.

app/services/payment/spam_protection_service.rb
duduribeiro: we are rewriting it to have 10 times in a timespan of 1 hour

app/models/payment_gateway_setting.rb
duduribeiro: we are moving it to redis

app/services/payment/spam_protection_service.rb
```diff
@@ -0,0 +1,30 @@
...
+      limiter = ActionLimiter.new(@account)
+      limiter.perform! 'invalid_payment'
+    rescue ActionLimiter::ActionLimitsExceededError
+      account.force_suspend!
+      user_session&.revoke!
+    end
+
+    private
+
+    attr_accessor :account, :gateway_setting, :payment_status, :user_session
```
hallelujah: You do not need writer for those

app/services/action_limiter.rb
```diff
@@ -0,0 +1,46 @@
...
+  # Pass a block to perform
+  # @yield
+  #   limiter = ActionLimiter.new(User.current_guest)
+  #   limiter.perform! 'signup' do
+  #     UserSignup.create(user_params)
+  #   end
+  def perform!(action, &block)
+    @limiter.add(action)
+    raise ActionLimitsExceededError if @limiter.exceeded?(action, threshold: @threshold, interval: @timespan)
+    block&.call
```
hallelujah: I would not use `block&.call` but instead `yield if block_given?`
Why the change?

app/services/payment/spam_protection_service.rb
```diff
@@ -0,0 +1,30 @@
...
+      @payment_status  = payment_status
+      @gateway_setting = account.gateway_setting
+      @user_session    = user_session
+    end
+
+    def call
+      return if payment_successful?
+
+      limiter = ActionLimiter.new(@account)
+      limiter.perform! 'invalid_payment'
```
hallelujah: This does not rescue Redis Error ... use `perform` instead

app/services/action_limiter.rb
duduribeiro: I changed because I don't need to pass any block in the client using this method. I will change it to yield if block given

app/services/payment/spam_protection_service.rb
```diff
@@ -0,0 +1,30 @@
...
+    rescue ActionLimiter::ActionLimitsExceededError
+      account.force_suspend!
+      user_session&.revoke!
+    end
+
+    private
+
+    attr_reader :account, :gateway_setting, :payment_status, :user_session
+
+    def payment_successful?
```
guicassolato:
```suggestion
    alias_method :payment_successful?, :payment_status
```

Martouta: Do we even need this method? in line 13 it could be just `return if payment_status` :thinking:

app/services/payment/spam_protection_service.rb
```diff
@@ -0,0 +1,30 @@
...
+module Payment
+  class SpamProtectionService
+    def initialize(account, payment_status, user_session = nil)
+      @account         = account
+      @payment_status  = payment_status
+      @gateway_setting = account.gateway_setting
+      @user_session    = user_session
+    end
+
+    def call
```
Martouta: imo this should return a truthy value if it works and a falsy one if it doesn't. But it isn't a big deal in this case because it is only used in a callback without caring about the response

app/services/payment/spam_protection_service.rb
Martouta: Or even in the intializer `@payment_successful = payment_status` and we make it more understandable

app/services/payment/spam_protection_service.rb
```diff
@@ -0,0 +1,30 @@
...
+      return if payment_successful?
+
+      limiter = ActionLimiter.new(@account)
+      limiter.perform 'invalid_payment'
+    rescue ActionLimiter::ActionLimitsExceededError
+      account.force_suspend!
```
guicassolato: Should we log something here so afterwards we know the reason why the account got suspended?

app/services/payment/spam_protection_service.rb
duduribeiro: Agree with @Martouta.

app/services/payment/spam_protection_service.rb
Martouta: I agree with @guicassolato . Easier to debug later.

app/services/payment/spam_protection_service.rb
guicassolato: I think it's about readability, but `return if payment_successful` (without the question mark) is acceptable enough.

app/services/payment/spam_protection_service.rb
duduribeiro: @guicassolato what about sending an email? And if yes, I send the email for whom? 🤔

guicassolato: No need for an email IMHO.

The intention here is to prevent brute force attacks and the new code already takes care of doing whatever one would do in case an email was sent. Besides, it's not like the email message would add anything, as the account data is most likely irrelevant in this case. Finally, sending an email would increase risk of DDoS attack.

duduribeiro: I ment an email to us/admins to inform that someone is suspended. But I think I'll add it on the logs

app/services/action_limiter.rb
```diff
@@ -0,0 +1,46 @@
...
+class ActionLimiter
+  class ActionLimitsExceededError < StandardError
+    def initialize(*)
+      super("Rate limit exceeded")
+    end
+  end
+
+  TIME_SPAN = 1.hour
+  BUCKET_INTERVAL = 10.minutes
+  THRESHOLD = 10
```
Martouta: Should this be from a config file instead? @duduribeiro @guicassolato @hallelujah

app/services/payment/spam_protection_service.rb
duduribeiro: I didn't add any return because it wasn't needed in the usage of it.

app/services/action_limiter.rb
hallelujah: Yes we can and also we need sane constants defaults in case the config does not exists ;)

app/services/payment/spam_protection_service.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module Payment
+  class SpamProtectionService
```
hallelujah: I think we should rename this to `MultipleFailureChecker`
It is not really spam protection but multiple failure check

We have our strategy based on:

- ban if failed multiple times (timespan 3 days for example)
- ban if attempt to request multiple times (timespan 1 hour for example)

app/models/account/states.rb
```diff
@@ -82,6 +82,10 @@ module Account::States
         transition all => :suspended, if: :tenant?
       end
 
+      event :force_suspend do
```
hallelujah: I'd avoid creating another event

app/services/action_limiter.rb
Martouta: imo if the config does not exist, the feature is disabled :smile:

app/services/payment/spam_protection_service.rb
Martouta: Right, I agree there is no need for now but I mentioned it because it does not add any complexity and can be handy.
Like:
```ruby
def call
  return true if payment_successful

  limiter = ActionLimiter.new(@account)
  limiter.perform 'invalid_payment'
rescue ActionLimiter::ActionLimitsExceededError
  account.force_suspend!
  user_session&.revoke!
  Rails.logger.info "spam_protection_service: account_suspended id: #{account.id}"
  nil
end
```

But as you said, there is no need so feel free to don't do it :wink:

app/services/payment/spam_protection_service.rb
```diff
@@ -0,0 +1,27 @@
...
+    def initialize(account, payment_status, user_session = nil)
```
Martouta: Not mandatory but we could skip the nil check that we have below by using here the _Null Object pattern_:
```suggestion
    def initialize(account, payment_status, user_session = UserSession.new)
```

app/models/account/states.rb
hallelujah:
```ruby
        transition all => :suspended, if: :can_be_suspended?

def can_be_suspended?
  tenant? || marked_for_suspsension
end

attr_accessor :marked_for_suspsension
```

app/services/payment/spam_protection_service.rb
```diff
@@ -0,0 +1,27 @@
...
+      @account            = account
+      @payment_successful = payment_status
+      @gateway_setting    = account.gateway_setting
+      @user_session       = user_session
+    end
+
+    def call
+      return if payment_successful
+
+      limiter = ActionLimiter.new(@account)
```
Martouta:
```suggestion
      limiter = ActionLimiter.new(account)
```

Since you created the reader method anyway :smile:

app/services/payment/spam_protection_service.rb
guicassolato:
```ruby
return true if payment_successful
```

is confusing IMO. I'd expect the service to return true if it did something.

app/services/payment/spam_protection_service.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module Payment
+  class SpamProtectionService
+    def initialize(account, payment_status, user_session = nil)
+      @account            = account
+      @payment_successful = payment_status
+      @gateway_setting    = account.gateway_setting
```
Martouta: We are not using this at all, no?

Kill it with fire

app/services/payment/spam_protection_service.rb
Martouta: @guicassolato that makes sense too. then:

```ruby
def call
  return if payment_successful

  limiter = ActionLimiter.new(@account)
  limiter.perform 'invalid_payment'
rescue ActionLimiter::ActionLimitsExceededError
  account.force_suspend!
  user_session&.revoke!
  Rails.logger.info "spam_protection_service: account_suspended id: #{account.id}"
  nil
end
```

As: `nil` means that it didn't do anything, truthy means that it did (for the limiter.perform)

PR #1459, opened 2019-11-27
Fix BackendApiConfig issuing ProxyConfigs::AffectingObjectChangedEvent on proxy update

guicassolato:
Closes [THREESCALE-3989](https://issues.jboss.org/browse/THREESCALE-3989)

### DIFF
--- a/app/lib/proxy_config_affecting_changes.rb
+++ b/app/lib/proxy_config_affecting_changes.rb
@@ -71,7 +71,7 @@ module ProxyConfigAffectingChanges
       after_commit :issue_proxy_affecting_change_events
 
       def issue_proxy_affecting_change_events
-        return unless service
+        return unless service && previous_changes.present?
 
         issue_proxy_affecting_change_event(service.proxy)
       end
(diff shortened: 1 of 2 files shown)


app/lib/proxy_config_affecting_changes.rb
```diff
@@ -71,7 +71,7 @@ module BackendApiConfigExtension
       after_commit :issue_proxy_affecting_change_events
 
       def issue_proxy_affecting_change_events
-        return unless service
+        return unless service && previous_changes.present?
```
Martouta: I think it is [easier to read](http://www.railstips.org/blog/archives/2008/12/01/unless-the-abused-ruby-conditional/) the following way:
```suggestion
        return if !service || previous_changes.blank?
```

guicassolato: Interesting. I find it easier as is now.

Martouta: Oki :+1: Either way for sure not a big deal :+1: approved :green_heart:

PR #1461, opened 2019-11-27
[WIP][Refactor] ModelsTest skips test classes by checking where they are defined

Martouta:
After [this comment,](https://github.com/3scale/porta/pull/1452#discussion_r351341171) we spoke and as it can be inherited from ActiveRecord::Base or from Account or from any class, we decided to check where the class is defined and Hery discovered this way to do it :smile:


Martouta:
It cannot find the classes that do not have any method defined. Mission abort :stuck_out_tongue:

PR #1462, opened 2019-11-27
Remove Airbrake from codebase

josemigallas:

Remove usage of Airbrake and references in the code.

[THREESCALE-3905: Remove Airbrake from code](https://issues.jboss.org/browse/THREESCALE-3905)


app/models/finance/billing_strategy.rb
```diff
@@ -349,18 +345,12 @@ def needs_credit_card?
...
   def bill_and_charge_each(options = {})
     buyer_ids = options[:buyer_ids]
     @failed_buyers = []
 
-    if provider.nil?
-      airbrake(:error_message => "WARNING: tried to use billing strategy #{self.id} which has no account",
-               :error_class => 'InvalidData')
-      return
-    end
+    return if provider.nil?
```
Martouta:
```suggestion
    return unless provider
```

Imo it should become the first line of the method now :smile:

app/controllers/frontend_controller.rb
```diff
@@ -164,7 +164,7 @@ def ensure_master_domain
   end
 
   def error_request_data
-    defined?(Airbrake) ? airbrake_request_data : {}
+    {}
   end
```
Martouta: Imo we should remove this method completely now.
image
image

We should get rid of that `options` param and simplify the method `notify_about_wrong_domain` to:

```ruby
def notify_about_wrong_domain(url, expected)
   # if there is no referer there is not a link leading there and we don't care about it
   return unless request.referer
   Rails.logger.info("ThreeScale::Warnings::WrongDomain: URL '#{url}' is #{expected} only.")
end
```

In fact, we may be able to completely remove that method as well and then turn `render_wrong_domain_error` into:

```ruby
def render_wrong_domain_error(expected)
  Rails.logger.info("ThreeScale::Warnings::WrongDomain: URL '#{request.url}' is #{expected} only.") if request.referer
  render_error "The path '#{request.path}' is not accessible on domain #{request.host}", status: :not_found
end
```

app/models/finance/billing_strategy.rb
```diff
@@ -104,10 +104,6 @@ def self.daily(options = {})
         message = "BillingStrategy #{id}(#{name}) failed utterly"
 
         Rails.logger.error(message)
-        airbrake(:error_message => message,
-                 :error_class => 'BillingError',
-                 :exception => e)
-
```
Martouta: I wonder if we should do a `System::ErrorReporting.report_error` here instead. I guess not because the next line raises an error, and it is not rescued anywhere, then it will notify bugsnag (and will do a Rails logger error as well, I think).
But I don't know if it is rescued somewhere or not. Maybe @guicassolato knows.

config/examples/settings.yml
```diff
@@ -11,7 +11,6 @@ base: &default
   asset_host: <%= ENV.fetch('RAILS_ASSET_HOST', nil) %>
   daily_weekly_reports_pref: <%= ENV.fetch('ENABLE_DAILY_WEEKLY_REPORTS_PREF', '1') == '1' %>
   readonly_custom_domains_settings: <%= ENV.fetch('READONLY_CUSTOM_DOMAINS_SETTINGS', '1') == '1' %>
-  airbrake_api_key: <%= ENV['AIRBRAKE_API_KEY'] %>
```
Martouta: I think u will have to remove it also from the deploy repo :smile:

app/lib/system/error_reporting.rb
```diff
@@ -16,7 +16,6 @@ def report_error(exception, rack_env: nil, logger: Rails.logger, **parameters)
 
       logger.error('Exception') { {exception: {class: exception.class, message: (exception.try(:message) || exception.to_s), backtrace: (exception.try(:backtrace) || [])[0..3]}, parameters: parameters} }
 
-      ::Airbrake.notify_or_ignore(exception, options) if defined?(Airbrake)
```
Martouta: This implies that we don't need to declare the `options` of the first line of the method anymore :smile:

app/controllers/frontend_controller.rb
josemigallas: Is this useless?
```ruby
      options[:error_message] = "URL '#{url}' is #{expected} only."
      options[:error_class] = 'ThreeScale::Warnings::WrongDomain'
```

Martouta: yes @josemigallas , because the only places where this method is called, later they only call `render_wrong_domain_error` which does not use that at all :joy: 

This was useful only for Airbrake really long ago, and they forgot to remove those options [when they removed the Airbrake line](https://github.com/3scale/system/commit/c6361dfe85a4a70e37fa50de9103969762a0c4b7#diff-618ac6be5874259ac2ff1b1e897fb693L32).

config/examples/settings.yml
josemigallas: https://github.com/3scale/deploy/pull/621

app/models/finance/billing_strategy.rb
josemigallas: In other places the error was reported manually.. so it would make sense to me to do it here too.

hallelujah: No need if it raises the error unless you want to add cuatom data to the error

app/lib/system/error_reporting.rb
```diff
@@ -11,12 +11,9 @@ module ErrorReporting
       options
     end
 
-    def report_error(exception, rack_env: nil, logger: Rails.logger, **parameters)
-      options = OPTIONS.call(parameters, rack_env)
-
+    def report_error(exception, logger: Rails.logger, **parameters)
```
guicassolato: Oops, this change will break https://github.com/3scale/porta/blob/c1fbcb787ff4b4d8f257a06a7a22705d0ec08563/app/lib/three_scale/message_bus_config.rb#L28

Martouta:
Plz do not forget to fix this :smile: 
image

app/lib/three_scale/hash_hacks.rb
```diff
@@ -17,11 +17,6 @@ def transform_values
       end
     end
 
-    def sweep
```
Martouta: What is [this](https://github.com/3scale/system/commit/c217bac039a8eb926275a755e47e2ac0c0c4fa4b) for and how do we know that it is safe to remove it?

app/lib/system/error_reporting.rb
Martouta: Right, so we just need to remove that parameter from that line too :smile: 
image

josemigallas: What about [OPTIONS](https://github.com/3scale/porta/blob/a9bd838dbcd11d1594c2dbfad6330745547067ab/app/lib/system/error_reporting.rb#L7-L11)? Where is it used?

app/lib/three_scale/hash_hacks.rb
josemigallas: I looked up usages and found none.. but it doesn't belong to the PR anyways so I'll revert it

app/models/finance/billing_strategy.rb
```diff
@@ -374,10 +364,6 @@ def bill_and_charge_each(options = {})
 
         msg = "Failed to bill or charge #{name}(#{buyer_id}) of provider(#{provider_id}): #{exception.message}\n"
         error(msg, buyer)
-        airbrake(:error_message => msg,
```
guicassolato: Although the method was called `airbrake`, it's been actually reporting to Bugsnag:

https://github.com/3scale/porta/tree/master/app/models/finance/billing_strategy.rb#L416

https://github.com/3scale/porta/tree/master/app/lib/system/error_reporting.rb#L20

Removing this call will change that. Is this really what we want?

app/lib/system/error_reporting.rb
Martouta: @josemigallas it is not used at all anymore. u can remove it :smile:

app/models/finance/billing_strategy.rb
josemigallas: By the way, according to the second condition [`Rails.env.preview?`](https://github.com/3scale/porta/blob/c1fbcb787ff4b4d8f257a06a7a22705d0ec08563/app/models/finance/billing_strategy.rb#L415), errors will be sent to Bugsnag when in preview, but that seems to contradict [`error_reporting_stages`](https://github.com/3scale/porta/blob/e2baf42638754496838ca5b579f62f72bd4bda24/config/initializers/bugsnag.rb#L5-L6).

guicassolato: I agree. Let's `delegate :report_error, to: 'System::ErrorReporting'` and replace those `airbrake` calls to a simple `report_error`.

app/models/finance/billing_strategy.rb
```diff
@@ -410,17 +410,7 @@ def add_cost(contract, name, description, cost, plan = contract.plan)
...
-      if Rails.env.production? || Rails.env.preview?
-        System::ErrorReporting.report_error(*args)
-      end
-    end
-  end
-
-  # so that we have #airbrake as instance AND class method
-  extend ErrorHandling
-  include ErrorHandling
+  delegate :report_error, to System::ErrorReporting
```
Martouta:
```suggestion
  delegate :report_error, to 'System::ErrorReporting'
```
sorry :pray:

PR #1457, opened 2019-11-26
Refactor redux actions in Policies widget

josemigallas:
**What this PR does / why we need it**:

Refactorization and cleanup of the redux actions and action creators.
* Rename actions with more (hopefully) clear names.
* Adds descriptions to each action
* Use actions/index as a proper index

**Verification steps**:
* `npm run update-dependencies`
* Navigate to `apiconfig/services/<id>/policies/edit`
* Policies widget should work as usual

**Note**:
This is part 10 of the refactorization of Policies:
[THREESCALE-2221: Policies: Normalise the use of policies schema and its types](https://issues.jboss.org/browse/THREESCALE-2221)

Previous step #1438.

PR #1463, opened 2019-11-28
Cleanup and refactor Policies

josemigallas:

General cleanup with small refactorizations here and there.

**Verification steps**:

**Note**:
This is part 11 of the refactorization of Policies:

Previous step #1457 .


app/javascript/src/Policies/components/PolicyChain.jsx
```diff
@@ -28,26 +28,24 @@ const SortableItem = SortableElement(({value, editPolicy, index}) => {
...
-          key={`item-${index}`}
-          index={index}
-          value={policy}
-          editPolicy={editPolicy}
-        />
-      ))}
-    </ul>
-  )
-})
+const SortableList = SortableContainer(({ items, editPolicy }) => (
```
didierofrivia: These kind of changes are always quite annoying... nothing really changes here, just returning implicitly... would be nice if GH can avoid marking as all lines has changed...

josemigallas: Yes it would..

PR #1464, opened 2019-11-28
BackendClient::Application::Utilization with the metric's extended system_name

guicassolato:
This PR fixes a bug in the `BackendClient::Application::Utilization` module that, despite receiving usage data from apisonator properly separated by each metric's true/extended system_name, was ignoring the extension and grouping the records by `system_name` without the suffix. This was causing a 500 error while at `/apiconfig/services/<service_id>/applications/<application_id>`.

Closes [THREESCALE-4013](https://issues.jboss.org/browse/THREESCALE-4013)

### DIFF
--- a/app/lib/backend_api_logic/metric_extension.rb
+++ b/app/lib/backend_api_logic/metric_extension.rb
@@ -10,9 +10,8 @@ module BackendApiLogic
       validate :unique_extended_system_name, if: :backend_api_metric?
       before_save :extend_system_name, if: :backend_api_metric?
 
-      def system_name
-        attributes['system_name'].to_s.gsub(/#{Regexp.escape(SYSTEM_NAME_SUFFIX_SEPARATOR)}\d+\z/, '')
-      end
+      alias_method :system_name, :system_name_without_suffix
+      public :system_name
     end
 
     def backend_api_metric?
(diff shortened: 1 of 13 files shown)


app/lib/backend_api_logic/metric_extension.rb
```diff
@@ -37,14 +36,20 @@ def hits_extended_system_name_regex
       end
     end
 
+    def extended_system_name
+      parts = [system_name_without_suffix]
+      parts << owner_id if backend_api_metric?
+      parts.compact.join SYSTEM_NAME_SUFFIX_SEPARATOR
+    end
```
Martouta: 1. it saves in "parts" the `system_name_without_suffix` which is always a string of the first part of the system_name, even if system_name is empty, then it is `''`
2. It adds to "parts" the `owner_id` if `owner_type == 'BackendApi'`. In this case owner_id cannot be `nil` unless the BackendApi is built but not persisted.
3. it returns "parts" as a string separated by "-", having removed first the `nil` values of part. But not the empty strings, so it means that if the owner_id is `nil`, it returns only the first part of "parts" (the result of `system_name_without_suffix`), but if it is the other way around, that `system_name_without_suffix` is an empty string, and owner_id has a value, it will return something like '-3' :joy: I just hope that this cannot happen.

app/lib/backend_api_logic/metric_extension.rb
```diff
@@ -10,9 +10,8 @@ module MetricExtension
       validate :unique_extended_system_name, if: :backend_api_metric?
       before_save :extend_system_name, if: :backend_api_metric?
 
-      def system_name
-        attributes['system_name'].to_s.gsub(/#{Regexp.escape(SYSTEM_NAME_SUFFIX_SEPARATOR)}\d+\z/, '')
-      end
+      alias_method :system_name, :system_name_without_suffix
+      public :system_name
```
Martouta: Do we need this line? :scream: why?

guicassolato: Because `#system_name_without_suffix` is protected.

app/lib/backend_api_logic/metric_extension.rb
guicassolato: This is fine.

PR #1465, opened 2019-11-29
Merge all routing policies into one

guicassolato:
This is one of the options we have to fix [THREESCALE-4016](https://issues.jboss.org/browse/THREESCALE-4016) It corresponds to the first solution proposed in the dev notes.

### DIFF
--- a/app/lib/backend_api_logic/routing_policy.rb
+++ b/app/lib/backend_api_logic/routing_policy.rb
@@ -5,7 +5,15 @@ module BackendApiLogic
     def policy_chain
       chain = super
       return chain unless with_subpaths?
-      Builder.new(service).to_a.concat(chain)
+
+      other_routing_policies, other_policies = chain.partition { |policy| policy['name'] == 'routing' }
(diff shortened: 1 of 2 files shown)


app/lib/backend_api_logic/routing_policy.rb
```diff
@@ -5,7 +5,15 @@ module RoutingPolicy
...
       chain = super
       return chain unless with_subpaths?
-      Builder.new(service).to_a.concat(chain)
+
+      other_routing_policies, other_policies = chain.partition { |policy| policy['name'] == 'routing' }
+      other_routing_rules = other_routing_policies.flat_map { |policy| policy['configuration']['rules'] }
+
+      routing_policy = Builder.new(service).to_h
+
+      return other_policies if routing_policy['configuration']['rules'].concat(other_routing_rules).empty?
```
hallelujah: `rules` can be `nil` in some policies. But agree that for `routing` it must be an Array


```
[{"configuration":{},"name":"apicast","version":"builtin","enabled":true},{"configuration":{"rules":[]},"name":"routing","version":"builtin","enabled":true}]
```

PR #1466, opened 2019-12-03
Add Captcha to password reset form in Dev Portal

josemigallas:
* Adds recaptcha to the reset password form in Dev portal (provided **Spam Protection** is enabled)
* refactor and extract `spam_protection` logic from `sign_up` and `password_reset` into a new form class `SpamProtected`.

Screen Shot 2019-12-03 at 11 12 10

[THREESCALE-1135: Add Captcha element in the password reset form](https://issues.jboss.org/browse/THREESCALE-1135)

**Verification steps** 
1. Go to admin/account/password/new in the Developer Portal
2. There should be a captcha
3. Form can't be submitted without completing the captcha

### DIFF
--- a/lib/developer_portal/app/controllers/developer_portal/admin/account/passwords_controller.rb
+++ b/lib/developer_portal/app/controllers/developer_portal/admin/account/passwords_controller.rb
@@ -10,9 +10,7 @@ class DeveloperPortal::Admin::Account::PasswordsController < ::DeveloperPortal::
   before_action :find_user, :only => [:show, :update]
 
   def create
-    unless spam_check(@buyer)
-      return redirect_to new_admin_account_password_url
-    end
+    return redirect_to new_admin_account_password_url unless spam_check(buyer)
 
     if user = @provider.buyer_users.find_by_email(params[:email])
       user.generate_lost_password_token!
@@ -24,10 +22,7 @@ class DeveloperPortal::Admin::Account::PasswordsController < ::DeveloperPortal::
     end
   end
 
-  def new
-    @buyer = @provider.buyers.build
-    assign_drops(account: Liquid::Drops::Account.new(@buyer))
-  end
+  def new; end
 
   def show
     assign_drops password_reset_token: @token
@@ -46,6 +41,10 @@ class DeveloperPortal::Admin::Account::PasswordsController < ::DeveloperPortal::
   end
 
   private
+
+  def buyer
+    @buyer ||= @provider.buyers.build
+  end
 
   def password_params
     params.require(:user).permit(:password, :password_confirmation)
--- a/lib/developer_portal/app/controllers/developer_portal/admin/account/passwords_controller.rb
+++ b/lib/developer_portal/app/controllers/developer_portal/admin/account/passwords_controller.rb
@@ -1,4 +1,8 @@
+# frozen_string_literal: true
+
 class DeveloperPortal::Admin::Account::PasswordsController < ::DeveloperPortal::BaseController
+  include ThreeScale::SpamProtection::Integration::Controller
+
   liquify prefix: 'password'
 
   skip_before_action :login_required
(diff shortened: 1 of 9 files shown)


lib/developer_portal/app/controllers/developer_portal/admin/account/passwords_controller.rb
```diff
@@ -16,7 +22,9 @@ def create
     end
   end
 
-  def new; end
+  def new
+    assign_drops(account: Liquid::Drops::Account.new(buyer))
```
hallelujah: Do you need to expose this to the liquid?

You can just use the registers (current_account or site_account) if only used in ruby files and not liquid

https://github.com/3scale/porta/blob/3968319a9fa477de9997ad7860bbc8ab5268d465/lib/liquid/template_support.rb#L132-L133

josemigallas: Does that mean I can use `current_account` directly without calling `assign_drop`?

lib/developer_portal/app/views/developer_portal/password/new.html.liquid
```diff
@@ -6,7 +6,7 @@
         Forgot password
       </div>
       <div class="panel-body">
-        {% form 'password_reset', { class: 'form-horizontal' } %}
+        {% form 'password_reset', account, class: 'form-horizontal' %}
```
josemigallas: @hallelujah
Could I use `current_account` here directly?

lib/developer_portal/app/controllers/developer_portal/admin/account/passwords_controller.rb
hallelujah: yes just use context.registers[:current_account] in the ruby code

lib/developer_portal/app/views/developer_portal/password/new.html.liquid
hallelujah: You do not need to use that here, but in the ruby file  `form`

hallelujah: Especially if you change it here, then you will need a migration for all the current customers (because this liquid file is stored in the database)
Which is something really to avoid as most as possible

lib/developer_portal/lib/liquid/forms/spam_protected.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+module Liquid
+  module Forms
+    class SpamProtected < Forms::Create
+      def initialize(context, _object_name, html_attributes = {})
+        # FIXME: object_name should be removed and the template password/new fixed. We're
+        # accidentally passing a bracket '{'
+        super(context, 'site_account', html_attributes)
```
hallelujah: I do not understand this one, why `'site_account'`?  Is it a drop or the actual provider account?

test/unit/liquid/forms_test.rb
```diff
@@ -27,7 +27,9 @@ class Liquid::FormsTest < ActiveSupport::TestCase
   end
 
   test 'password_reset form' do
-    form = get('password_reset', "  ")
+    form = get('password_reset', 'site_account')
+    account = FactoryBot.create(:account)
+    form.context['site_account'] = Liquid::Drops::Message.new(account)
```
hallelujah: Why is this a `Liquid::Drops::Message` instead of `Account` or maybe `Liquid::Drops::Account`

lib/developer_portal/lib/liquid/forms/spam_protected.rb
guicassolato: I think that's because of how `Liquid::Forms::Base#object` works.

https://github.com/3scale/porta/blob/5a719dc90fc3b7a277cb64755c1b4868a1279a54/lib/developer_portal/lib/liquid/forms/base.rb#L29

hallelujah: OK so the form has `object` which how it builds the class names/attributes values
And it gets it from a drop which gets back the real `model`

Thanks for enlightening me, then the test needs to be fixed IMO

test/unit/liquid/forms_test.rb
hallelujah: After reading this https://github.com/3scale/porta/pull/1466#discussion_r373538298
It should be a `Liquid::Drops::Account`

PR #1469, opened 2019-12-04
Rename proxy deployment service to v1

josemigallas:

This is a simple refactorization.

`ProviderProxyDeploymentService` handles only APIcast v1 deployments and this is going to be removed soon. In order to make a new service for v2, this one is to be renamed to `ProxyDeploymentServiceV1`. This way it will be easier to remove and also we won't mix responsibilities.

Works towards [THREESCALE-3751: Tech-Debt: Create a Proxy Deployment Service](https://issues.jboss.org/browse/THREESCALE-3751), but doesn't fix it.

**Verification steps** 

Everything deployment-wise should work the same.

### DIFF
--- a/app/models/proxy.rb
+++ b/app/models/proxy.rb
@@ -326,7 +326,7 @@ class Proxy < ApplicationRecord
   end
 
   def async_deploy(user)
-    ProviderProxyDeploymentService.async_deploy(user, self)
+    ProviderProxyDeploymentV1Service.async_deploy(user, self)
   end
 
   def hosts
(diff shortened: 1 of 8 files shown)


app/models/proxy.rb
```diff
@@ -326,7 +326,7 @@ def save_and_async_deploy(attrs, user)
   end
 
   def async_deploy(user)
-    ProviderProxyDeploymentService.async_deploy(user, self)
+    ProviderProxyDeploymentServiceV1.async_deploy(user, self)
```
guicassolato: Sorry. I know I first suggested this name, but can we change it to `ProviderProxyDeploymentV1Service` instead? The way it is now it makes it look like it's the service that is v1, when in fact this is a reference to apicast version.

josemigallas: Oki doki

app/services/provider_proxy_deployment_v1_service.rb
```diff
@@ -1,6 +1,6 @@
 require_dependency 'apicast/sandbox'
 
-class ProviderProxyDeploymentService
+class ProxyDeploymentV1Service
```
Martouta: then you should do `mv app/services/provider_proxy_deployment_v1_service.rb app/services/proxy_deployment_v1_service.rb`

PR #1470, opened 2019-12-04
New Deployment Service for Proxy

josemigallas:

Goes after #1469

[THREESCALE-3751: Tech-Debt: Create a Proxy Deployment Service](https://issues.jboss.org/browse/THREESCALE-3751)

**Verification steps** 

Everything should work the same.

### DIFF
--- a/app/controllers/admin/api/services/proxies_controller.rb
+++ b/app/controllers/admin/api/services/proxies_controller.rb
@@ -63,13 +63,7 @@ class Admin::Api::Services::ProxiesController < Admin::Api::Services::BaseContro
   ##~ op.parameters.add name: "jwt_claim_with_client_id_type", description: "JWT Claim With ClientId Type. Either `plain` or `liquid`", dataType: "string", paramType: "query", required: false
   #
   def update
-    if proxy.update_attributes(proxy_params)
-      if proxy.service_mesh_integration?
-        proxy.deploy!
-      elsif proxy.apicast_configuration_driven
-        proxy.deploy_v2
-      end
-    end
+    ProxyDeploymentService.call(proxy) if proxy.update_attributes(proxy_params)
 
     respond_with(proxy)
   end
@@ -87,11 +81,8 @@ class Admin::Api::Services::ProxiesController < Admin::Api::Services::BaseContro
   ##~ op.parameters.add @parameter_access_token
   ##~ op.parameters.add @parameter_service_id_by_id_name
   def deploy
-    if proxy.service_mesh_integration?
-      proxy.deploy!
-    elsif proxy.apicast_configuration_driven
-      proxy.deploy_v2
-    end
+    ProxyDeploymentService.call(proxy)
+
     respond_with(proxy)
   end
 
--- a/app/controllers/admin/api/services/proxy/policies_controller.rb
+++ b/app/controllers/admin/api/services/proxy/policies_controller.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 class Admin::Api::Services::Proxy::PoliciesController < Admin::Api::Services::BaseController
-
   wrap_parameters ::Proxy
   representer Proxy::PoliciesConfig
 
@@ -41,7 +40,7 @@ class Admin::Api::Services::Proxy::PoliciesController < Admin::Api::Services::Ba
   #
   def update
     if proxy.update_attributes(proxy_params)
-      proxy.deploy_v2 if proxy.apicast_configuration_driven
+      ApicastV2DeploymentService.new(@proxy).call(environment: :sandbox) if proxy.apicast_configuration_driven
     end
 
     policies_config = Proxy::PoliciesConfig.new(proxy.policies_config)
--- a/app/controllers/admin/api/services/proxy/policies_controller.rb
+++ b/app/controllers/admin/api/services/proxy/policies_controller.rb
@@ -39,8 +39,8 @@ class Admin::Api::Services::Proxy::PoliciesController < Admin::Api::Services::Ba
   ##~ op.parameters.add name: "policies_config", description: "Proxy policies chain", dataType: "string", paramType: "query", required: true
   #
   def update
-    if proxy.update_attributes(proxy_params)
-      ApicastV2DeploymentService.new(@proxy).call(environment: :sandbox) if proxy.apicast_configuration_driven
+    if proxy.update_attributes(proxy_params) && proxy.apicast_configuration_driven
+      ApicastV2DeploymentService.new(@proxy).call(environment: :sandbox)
     end
 
     policies_config = Proxy::PoliciesConfig.new(proxy.policies_config)
--- a/app/controllers/api/integrations_controller.rb
+++ b/app/controllers/api/integrations_controller.rb
@@ -58,7 +58,7 @@ class Api::IntegrationsController < Api::BaseController
   end
 
   def update_production
-    @proxy.deploy_production
+    ProxyDeploymentService.call(@proxy, environment: :production)
     ThreeScale::TimedValue.set(deploying_hosted_proxy_key, true, 5*60 )
     ThreeScale::Analytics.track(current_user, 'Hosted Proxy deployed')
     flash[:notice] = flash_message(:update_production_success)
@@ -79,7 +79,7 @@ class Api::IntegrationsController < Api::BaseController
   end
 
   def promote_to_production
-    if @proxy.deploy_production
+    if ProxyDeploymentService.call(@proxy, environment: :production)
       flash[:notice] = flash_message(:promote_to_production_success)
     else
       flash[:error] = flash_message(:promote_to_production_error)
--- a/app/forms/onboarding/request_form.rb
+++ b/app/forms/onboarding/request_form.rb
@@ -47,7 +47,7 @@ class Onboarding::RequestForm < Reform::Form
   end
 
   def test_api!
-    return unless model.deploy!
+    return unless ProxyDeploymentService.call(model, v1_compatible: true)
 
     proxy_test_service = ProxyTestService.new(model)
 
--- a/app/models/proxy.rb
+++ b/app/models/proxy.rb
@@ -330,7 +330,7 @@ class Proxy < ApplicationRecord
   end
 
   def async_deploy(user)
-    ProxyDeploymentV1Service.async_deploy(user, self)
+    ApicastV1DeploymentService.async_deploy(user, self)
   end
 
   def hosts
@@ -426,7 +426,7 @@ class Proxy < ApplicationRecord
 
   def sandbox_deployed?
     proxy_log = provider.proxy_logs.latest_first.first or return sandbox_config_saved?
-    proxy_log.created_at > self.created_at && proxy_log.status == ProxyDeploymentV1Service::SUCCESS_MESSAGE
+    proxy_log.created_at > self.created_at && proxy_log.status == ApicastV1DeploymentService::SUCCESS_MESSAGE
   end
 
   def sandbox_config_saved?
--- a/app/models/proxy.rb
+++ b/app/models/proxy.rb
@@ -359,11 +359,11 @@ class Proxy < ApplicationRecord
 
     analytics.track('Sandbox Proxy updated', analytics_attributes.merge(success: saved))
 
-    saved && deploy
-  end
+    return false unless saved
 
-  def deploy!
-    deploy
+    success = ProxyDeploymentService.call(self, v1_compatible: true)
+    analytics.track('Sandbox Proxy Deploy', success: success)
+    success
   end
 
   def authentication_params_for_proxy(opts = {})
@@ -444,28 +444,8 @@ class Proxy < ApplicationRecord
       (self.api_backend ? URI(self.api_backend).host : 'none')
   end
 
-  def deploy_production
-    if apicast_configuration_driven
-      deploy_production_v2
-    elsif ready_to_deploy?
-      provider.deploy_production_apicast
-    end
-  end
-
   def ready_to_deploy?
     api_test_success
-  end
-
-  def deploy_v2
-    deployment = ApicastV2DeploymentService.new(self)
-
-    deployment.call(environment: 'sandbox'.freeze)
-  end
-
-  def deploy_production_v2
-    newest_sandbox_config = proxy_configs.sandbox.newest_first.first
-
-    newest_sandbox_config.clone_to(environment: :production) if newest_sandbox_config
   end
 
   def set_correct_endpoints
@@ -631,33 +611,6 @@ class Proxy < ApplicationRecord
     if (hits = service.metrics.first)
       proxy_rules.create(http_method: 'GET', pattern: '/', delta: 1, metric: hits)
     end
-  end
-
-  def deploy
-    return true unless deployable?
-
-    if service_mesh_integration?
-      deploy_service_mesh_integration
-    elsif apicast_configuration_driven
-      deploy_v2
-    else
-      deploy_v1
-    end
-  end
-
-  def deploy_service_mesh_integration
-    return unless provider_can_use?(:service_mesh_integration)
-    deploy_v2 && deploy_production_v2
-  end
-
-  def deploy_v1
-    deployment = ProxyDeploymentV1Service.new(provider)
-
-    success = deployment.deploy(self)
-
-    analytics.track('Sandbox Proxy Deploy', success: success)
-
-    success
   end
 
   def set_api_test_path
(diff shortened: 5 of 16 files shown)


app/controllers/admin/api/services/proxies_controller.rb
```diff
@@ -63,13 +63,7 @@ def show
...
   #
   def update
-    if proxy.update_attributes(proxy_params)
-      if proxy.service_mesh_integration?
-        proxy.deploy!
-      elsif proxy.apicast_configuration_driven
-        proxy.deploy_v2
-      end
-    end
+    ProxyDeploymentService.new(proxy).deploy if proxy.update_attributes(proxy_params)
```
hallelujah: Generally we use (by convention) `call` instead of `deploy` but that works too, no biggie :)

app/controllers/admin/api/services/proxy/policies_controller.rb
```diff
@@ -41,7 +40,7 @@ def show
   #
   def update
     if proxy.update_attributes(proxy_params)
-      proxy.deploy_v2 if proxy.apicast_configuration_driven
+      ProxyDeploymentService.new(proxy).deploy_v2 if proxy.apicast_configuration_driven
```
hallelujah: I would not have 2 methods for a Service, only one `#call`. It should handle all the logic inside. It should act as a blackbox. This way you do not leak any logic outside.

app/controllers/api/integrations_controller.rb
```diff
@@ -58,7 +58,7 @@ def update
   end
 
   def update_production
-    @proxy.deploy_production
+    ProxyDeploymentService.new(@proxy).deploy_production
```
hallelujah: Same as https://github.com/3scale/porta/pull/1470/files#r354185065

app/controllers/api/integrations_controller.rb
```diff
@@ -79,7 +79,7 @@ def update_onpremises_production
   end
 
   def promote_to_production
-    if @proxy.deploy_production
+    if ProxyDeploymentService.new(@proxy).deploy_production
```
hallelujah: Same as https://github.com/3scale/porta/pull/1470/files#r354185065

app/controllers/admin/api/services/proxy/policies_controller.rb
josemigallas: How about `deploy` and `deploy_production`? They don't depend on any variables (I think), would we need 2 services? `ProxyDeployment` and `ProxyDeploymentProduction`.. Or would it be better to pass an argument? `ProxyDeploymentService.call(@proxy, :production)`?

hallelujah: Well maybe just do 

```ruby
class ProxyDeploymentService
  def initialize(proxy, stage: :staging)
       @stage = stage
       @proxy = proxy
  end

  def call
    case @stage
    when :staging, :sandbox
    when :production 
    end
  end
  end
end
```

app/services/proxy_deployment_service.rb
```diff
@@ -0,0 +1,58 @@
+# frozen_string_literal: true
+
+class ProxyDeploymentService
+  delegate :apicast_configuration_driven, to: :@proxy
```
guicassolato:
```suggestion
  delegate :apicast_configuration_driven,
           :deployable?,
           :service_mesh_integration?,
           :provider,
           :proxy_configs, to: :@proxy
```

app/controllers/admin/api/services/proxy/policies_controller.rb
```diff
@@ -40,9 +39,7 @@ def show
   ##~ op.parameters.add name: "policies_config", description: "Proxy policies chain", dataType: "string", paramType: "query", required: true
   #
   def update
-    if proxy.update_attributes(proxy_params)
-      proxy.deploy_v2 if proxy.apicast_configuration_driven
-    end
+    proxy.deploy if proxy.update_attributes(proxy_params)
```
guicassolato: I see behaviour changing a bit here. What happens now if proxy is not configuration driven?

guicassolato: I know policies don't exist in this case, but does the endpoint prevent it?

app/services/proxy_deployment_service.rb
```diff
@@ -0,0 +1,62 @@
...
+class ProxyDeploymentService
+  delegate :apicast_configuration_driven,
+           :deployable?,
+           :service_mesh_integration?,
+           :provider,
+           :proxy_configs, to: :@proxy
+  delegate :deployable?, to: :@proxy
+  delegate :service_mesh_integration?, to: :@proxy
+  delegate :provider, to: :@proxy
+  delegate :proxy_configs, to: :@proxy
```
guicassolato: No need for this anymore.

app/controllers/admin/api/services/proxy/policies_controller.rb
josemigallas: Although it looks like a change, it's been always evaluated inside `#deploy`. Here, this line: https://github.com/3scale/porta/pull/1470/files#diff-7cad44f925331c18a0c6ddeaeb8126e6R25-R31
I might have overlooked something but I'm positive it's the same behavior, only a couple of duplicated lines less.

app/services/proxy_deployment_service.rb
josemigallas: Damn it Github!

app/controllers/admin/api/services/proxy/policies_controller.rb
guicassolato: Consider

```ruby
proxy.update_attributes(proxy_params) == true
proxy.apicast_configuration_driven == false
proxy.service_mesh_integration? == false
```

What happened before and what happens now? Was `proxy.deploy_v1` invoked before? Will it be invoked now?

guicassolato:
```suggestion
    proxy.deploy if proxy.update_attributes(proxy_params) && proxy.apicast_configuration_driven
```

app/models/proxy.rb
```diff
@@ -355,11 +355,20 @@ def save_and_deploy(attrs = {})
 
     analytics.track('Sandbox Proxy updated', analytics_attributes.merge(success: saved))
 
-    saved && deploy
+    if saved
```
guicassolato:
```suggestion
    return false unless saved
    
    success = deploy
    analytics.track('Sandbox Proxy Deploy', success: success)
    success
```

app/models/proxy.rb
```diff
@@ -355,11 +355,20 @@ def save_and_deploy(attrs = {})
...
+      success = deploy
+      analytics.track('Sandbox Proxy Deploy', success: success)
+    end
+
+    saved && success
```
guicassolato: Remove this if https://github.com/3scale/porta/pull/1470/files#r354454921 of course.

app/services/proxy_deployment_service.rb
```diff
@@ -0,0 +1,58 @@
...
+
+  def self.call(*args)
+    new(*args).call
+  end
+
+  def initialize(proxy)
+    @proxy = proxy
+  end
+
+  def call
```
guicassolato:
```suggestion
  def call(environment: :sandbox)
    return true unless deployable?
    return deploy_production if environment == :production
```

Not entirely sure, but I'd start with an option and later maybe worth splitting into one separate classes, one per environment, as they seem to share virtually no logic. At least this first service abstracts already from start all the complexity. The refactor in to more specific classes can be done in a next iteration.

@hallelujah , WDYT?

app/controllers/admin/api/services/proxy/policies_controller.rb
josemigallas: OK gotcha, I mixed it up with the other controller

josemigallas: How could I include this in the tests?

app/services/proxy_deployment_service.rb
josemigallas: I added the parameter in the constructor but I agree there should be a second class

hallelujah: I think they share some logic. Having 2 classes might not be necessary but does not do any harm either  👍

app/services/proxy_deployment_service.rb
```diff
@@ -0,0 +1,70 @@
...
+    case @stage
+    when :staging, :sandbox
+      deploy
+    when :production
+      deploy_production
+    else
+      raise "nope" # ??
```
hallelujah: Yes :) Use a tagged error and rescue it where needed:

- controller
- sidekiq 😱 Yes because we do async deploy https://github.com/3scale/porta/blob/3968319a9fa477de9997ad7860bbc8ab5268d465/app/lib/logic/rolling_updates.rb#L160-L164

app/services/proxy_deployment_service.rb
```diff
@@ -0,0 +1,70 @@
...
+    end
+  end
+
+  def deploy_v2
```
hallelujah: Good for now but I would just remove this method later ;) And replace it with the service everywhere

josemigallas: This is only because of this line https://github.com/3scale/porta/pull/1470/files#diff-c1ade6e6897661208cbfd3d0c6ebcf8dR43 :/

app/controllers/admin/api/services/proxy/policies_controller.rb
hallelujah: In fact if we could have `Proxy#deployable?` and use it in the service.
This way you remove complexity and coupling.

You just use the service instead of `proxy.deploy` and can test it inside the service only

```ruby
class ProxyDeploymentService
   def initialize(proxy)
     @proxy = proxy
     # etc ...
   end
end

```

josemigallas: OK that's one missing test but the problem with this line of code is that it calls `deploy_v2` only, not `deploy_v1` or `deploy_production_v2`. And `service.call` runs at least one of them, which is way we need another method.

josemigallas:
@3scale/system-ruby feel free to review this again, I did some big changes. It turned out that some components were calling a `deploy_v1` when they shouldn't so I had to add another flag to the service.

app/services/proxy_deployment_service.rb
```diff
@@ -0,0 +1,71 @@
...
+      deploy_production
+    else
+      raise UnknownStageError, "Unknown environment: #{@stage}"
+    end
+  end
+
+  private
+
+  def deploy
+    return true unless deployable?
```
guicassolato: The following makes more sense to me:

```suggestion
    return unless deployable?
```

Not being deployable does not mean that deployed it is, so returning `true` is confusing.

app/services/proxy_deployment_service.rb
```diff
@@ -0,0 +1,71 @@
...
+    if apicast_configuration_driven
+      deploy_production_v2
+    elsif @proxy.ready_to_deploy?
+      provider.deploy_production_apicast
+    end
+  end
+
+  # Deprecated
+  def deploy_v1
+    ProxyDeploymentV1Service.new(provider).deploy(@proxy)
```
guicassolato: It's now clear IMO that the name of this service should have been `ApicastV1DeploymentService`, instead of `ProxyDeploymentV1Service `.

app/services/proxy_deployment_service.rb
josemigallas: Shouldn't we do that in another PR? One that affects master

guicassolato: I think this is part of this PR as it introduces `ProxyDeploymentService`, but maybe that's is just me.

josemigallas: But it's just a refactorization, it does not change any logic. By changing the return value.. it would change the behavior of current master.

josemigallas: https://github.com/3scale/porta/pull/1470/files#diff-3ab35f0eaf168ec265ee8e446edc8ca8L633

app/services/proxy_deployment_service.rb
```diff
@@ -0,0 +1,71 @@
...
+  end
+
+  private
+
+  def deploy
+    return true unless deployable?
+
+    if service_mesh_integration?
+      deploy_v2 && deploy_production_v2
+    elsif apicast_configuration_driven
```
hallelujah:
```suggestion
    elsif apicast_configuration_driven?
```

app/services/proxy_deployment_service.rb
```diff
@@ -0,0 +1,71 @@
...
+      deploy_v2
+    elsif @v1_compatible
+      deploy_v1
+    end
+  end
+
+  # TODO: in order to include this method into 'call', do I need another service?
+  def deploy_production
+    if apicast_configuration_driven
```
hallelujah:
```suggestion
    if apicast_configuration_driven?
```

app/forms/onboarding/request_form.rb
```diff
@@ -47,7 +47,7 @@ def proxy_auth_params
   end
 
   def test_api!
-    return unless model.deploy!
+    return unless ProxyDeploymentService.call(model, v1_compatible: true)
```
hallelujah: I will leave it for now, but please put a comment that once APIcast V1 is gone we should remove it

app/services/proxy_deployment_service.rb
josemigallas: Is your suggestion to call [that method](https://github.com/3scale/porta/blob/3c20040cb487edc35e7a1b87e05fa89ee61d18f5/app/helpers/api/integrations_helper.rb#L158) instead or to rename this one?

hallelujah: I think boolean methods are more explicit
Just for readability and convention

test/unit/services/proxy_deployment_service_test.rb
```diff
@@ -0,0 +1,94 @@
...
+
+    assert service.call
+  end
+
+  test 'deploy production not apicast configuration driven but api test success' do
+    @proxy.expects(:apicast_configuration_driven).returns(false)
+    @proxy.expects(:api_test_success).returns(true)
+
+    service = ProxyDeploymentService.new(@proxy, environment: :production)
+    service.expects(:deploy_v2).never
```
hallelujah: Expecting never is not really testing. Better replace by some expectation: state of an object or creation of proxy config with some attributes testing

test/unit/services/proxy_deployment_service_test.rb
```diff
@@ -0,0 +1,94 @@
...
+
+    assert service.call
+  end
+
+  test 'deploy neither service mesh or apicast configuration driven' do
+    @proxy.expects(:service_mesh_integration?).returns(false)
+    @proxy.expects(:apicast_configuration_driven).returns(false)
+
+    service = ProxyDeploymentService.new(@proxy)
+    service.expects(:deploy_v2).never
```
hallelujah: Same as https://github.com/3scale/porta/pull/1470/files#r378210027

test/unit/services/proxy_deployment_service_test.rb
```diff
@@ -0,0 +1,94 @@
...
+    refute service.call
+  end
+
+  # Deprecated
+  test 'deploy neither service mesh or apicast configuration driven but v1 compatible' do
+    @proxy.expects(:service_mesh_integration?).returns(false)
+    @proxy.expects(:apicast_configuration_driven).returns(false)
+
+    service = ProxyDeploymentService.new(@proxy, v1_compatible: true)
+    service.expects(:deploy_v2).never
```
hallelujah: Same as https://github.com/3scale/porta/pull/1470/files#r378210027

app/controllers/admin/api/services/proxy/policies_controller.rb
```diff
@@ -41,7 +40,7 @@ def show
   #
   def update
     if proxy.update_attributes(proxy_params)
-      proxy.deploy_v2 if proxy.apicast_configuration_driven
+      ApicastV2DeploymentService.new(@proxy).call(environment: :sandbox) if proxy.apicast_configuration_driven
```
Martouta: I propose 1 single if statement:


```ruby
if proxy.update_attributes(proxy_params) && proxy.apicast_configuration_driven
end
```

app/services/proxy_deployment_service.rb
josemigallas: I agree. Do you mean I create an alias? The method `apicast_configuration_driven` is already defined and I don't think it's a good idea to refactor it.

hallelujah: Yes an alias is OK for me

by the way it is also in an helper https://github.com/3scale/porta/blob/b13c634f92c4b5c121528cfcd434a175bda8be4a/app/helpers/api/integrations_helper.rb#L158-L161

josemigallas: I saw it but couldn't find a way to use it cleanly, I'd rather use the alias if that's OK.

PR #1476, opened 2019-12-12
Move menu system

duduribeiro:
Extracted from #1396

lib/menu_system into app/lib


app/lib/menu_system.rb
```diff
@@ -81,11 +82,11 @@ def active_sidebar
   end
 
   def active_menu?(level, title)
-    return unless title.present?
+    return if title.blank?
     # opposite of #humanize: "Developer Portal => :developer_portal
     # and also "EndUser Plans" => :end_user_plans
 
-    active_menus[level].try!(:to_s).try!(:underscore) == title.to_s.underscore.parameterize.underscore
+    active_menus[level]&.to_s&.underscore == title.to_s.underscore.parameterize.underscore
```
hallelujah: I wonder why we need to use `&.` here 🤔 

`to_s` should be available everywhere no?

And we are never going to compare with `nil` because there is a return if `title.blank?` above

duduribeiro: 👍

Martouta: good catch

PR #1477, opened 2019-12-12
Remove lib/color.rb

duduribeiro:
Extracted from #1396

 In order to remove the lib folder from autoload, this commit removes the
 lib/color file, since it was not being used in any place. Only one test
 was using some of its constant.

### DIFF
--- a/features/support/hooks.rb
+++ b/features/support/hooks.rb
@@ -1,4 +1,5 @@
-require 'color'
+BOLD  = "\033[1m"
+CLEAR = "\033[0m"
 
 Before '@ignore-backend' do
   stub_backend_get_keys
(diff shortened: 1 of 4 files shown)


features/support/hooks.rb
```diff
@@ -1,4 +1,5 @@
-require 'color'
+BOLD  = "\033[1m"
```
hallelujah: We should just use the colorize gem ....

duduribeiro: installing one gem just for this case? isn't this too much? 🤔

Martouta: i don't know that gem, but if u end up installing it, do it only for the test environment

hallelujah: it is already installed `bundle show colorize`

features/support/hooks.rb
```diff
@@ -188,13 +188,13 @@
...
-print_banner = -> (title, step) do
+print_banner = ->(title, step) do
   step_name = (step.try(:actual_keyword) || step.keyword) + step.name
-  Rails.logger.info <<-NEXT
+  Rails.logger.info <<~NEXT
 
-| #{title}: #{Color::BOLD}#{step_name}#{Color::CLEAR} |
-| #{'=' * (step_name.length + title.length + 2)} |
-#{step.multiline_arg}
+    | #{title}: #{step_name.colorize(:bold)} |
```
hallelujah:
```suggestion
    | #{title}: #{step_name.bold} |
```

duduribeiro: @hallelujah updated! 

thanks

PR #1478, opened 2019-12-12
[lib] Move lib/error_handling.rb to app/controllers/concerns

hallelujah:
Though I do not like using ActiveSupport::Concern, I will just move it there
It makes sense "for now" to have it there until we have another round of code improvement


app/controllers/concerns/error_handling.rb
```diff
@@ -25,19 +27,22 @@ def handle_access_denied(exception)
       handle_error(exception, :forbidden)
     end
 
-    def handle_error(exception, status)
+    def log_error(exception, status)
       logger.error "Handling Exception: #{exception} with status #{status}"
-      logger.debug exception.backtrace.join("\n") if exception.try!(:backtrace)
-
+      logger.debug exception.backtrace.join("\n") if exception&.backtrace
       @_exception_handled = exception
```
guicassolato: Why is this instance var no longer set by `#handle_error`?

PR #1479, opened 2019-12-12
Move lib/numeric_hacks to app/lib

duduribeiro:
lib/numeric_hacks into app/lib


app/lib/numeric_hacks.rb
```diff
@@ -19,12 +21,14 @@ def percentage_change_from(other)
...
+    if zero? && is_other_zero
       0.0
-    elsif other.zero?
+    elsif is_other_zero
       100.0
     else
-      (self.to_f / other) * 100.0
+      (to_f / other) * 100.0
     end
   end
```
Martouta:
```ruby
def percentage_ratio_of(other)
  if zero?
    0.0
  elsif other.zero?
    100.0
  else
    (to_f / other) * 100.0
  end
end
```

Because `(0/X) * 100.0` is always 0 anyway, so the comparison of the 2 values being 0 to return 0 does not make sense to me and we only need to compare this one.

PR #1480, opened 2019-12-13
Move lib/filters to app/lib

duduribeiro:
lib/filters into app/lib


app/lib/filters/provider_required.rb
```diff
@@ -10,7 +12,7 @@ module ControllerMethods
 
     def provider_required
       if current_account.nil? || !current_account.provider?
```
Martouta:
```suggestion
      if !current_account || current_account.buyer?
```

Martouta: Spoken in private :sunglasses: :+1:

app/lib/filters/provider_required.rb
```diff
@@ -8,13 +10,11 @@ def provider_required(options = {})
...
     protected
 
-    def provider_required
-      if current_account.nil? || !current_account.provider?
-        render_error 'Access denied', :status => :forbidden
-        false
-      else
-        true
-      end
+    def provider_require
```
Martouta:
```suggestion
    def provider_required
```

:see_no_evil:

duduribeiro: 🤦‍♂

Martouta: :joy:

PR #1484, opened 2019-12-13
[lib] Move lib/gitlab to app/lib

duduribeiro:
Extracted from #1396

lib/gitlab into app/lib


Martouta:
Originally extracted from [here](https://gitlab.com/gitlab-org/gitlab-foss/blob/4f6620a5ae00383a015379f95408ed7f1be1bdbb/lib/gitlab/testing/request_blocker_middleware.rb)

PR #1486, opened 2019-12-13
[lib] Move lib/layoutless_ajax_rendering to app/lib

duduribeiro:
lib/layoutless_ajax_rendering into app/lib


app/lib/layoutless_ajax_rendering.rb
```diff
@@ -10,16 +12,14 @@ module LayoutlessAjaxRendering
...
   # this method is called by render to get options
   def _normalize_args(*)
     options = super
 
-    if layoutless_rendering && !options.has_key?(:layout)
-      if request.xhr? || request.headers[PJAX]
-        options[:layout] = false
-      end
+    if layoutless_rendering && !options.key?(:layout)
+      options[:layout] = false if request.xhr? || request.headers[PJAX]
```
Martouta: For sure this can be simplified to at least this:
`options[:layout] ||= false if layoutless_rendering && (request.xhr? || request.headers[PJAX])`

But also, I think that we might be able to get rid of PJAX :thinking: not sure yet

Martouta: Ok, we are still using JPAX. I have created [THREESCALE-4138](https://issues.redhat.com/browse/THREESCALE-4138)

Martouta: So I suggest for now:
```ruby
def _normalize_args(*)
  options[:layout] ||= false if layoutless_rendering && (request.xhr? || request.headers[PJAX])
end
```

Martouta: or even better if this works:
```ruby
```ruby
def _normalize_args(*)
  super.tap do |options|
    options[:layout] ||= false if layoutless_rendering && (request.xhr? || request.headers[PJAX])
  end
end
```
```

PR #1487, opened 2019-12-16
Ignoring Developer portal assets in codeclimate

damianpm:

Codeclimate should ignore developer portal assets, but some directories were missing from exclusion in codeclimate.yml
(e.g. https://github.com/3scale/porta/pull/1468 was failing because of https://codeclimate.com/github/3scale/porta/pull/1468)

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -84,6 +84,7 @@ exclude_paths:
   - 'lib/action_dispatch/rails5_ssl.rb'
   - "lib/developer_portal/app/views/developer_portal/javascripts/*"
   - "lib/developer_portal/app/views/developer_portal/css/*"
+  - "lib/developer_portal/app/assets/stylesheets/**/*"
   - "assets/bundles/*.js"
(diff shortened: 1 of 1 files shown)


.codeclimate.yml
```diff
@@ -84,6 +84,7 @@ exclude_paths:
   - 'lib/action_dispatch/rails5_ssl.rb'
   - "lib/developer_portal/app/views/developer_portal/javascripts/*"
   - "lib/developer_portal/app/views/developer_portal/css/*"
+  - "lib/developer_portal/app/assets/stylesheets/**/*"
```
hallelujah: We should not have assets here

damianpm: Well we have this: https://github.com/3scale/porta/blob/master/lib/developer_portal/app/assets/stylesheets/plans_widget.scss
I just renamed from css to scss, but it was on that directory
Should I move it to `lib/developer_portal/app/views/developer_portal/css/` ?

PR #1489, opened 2019-12-17
[Oracle][Docker] Setup Docker Instructions + Use AppArmor default docker profile to run Oracle

Martouta:
This is needed in order to work in Ubuntu (and some other Linux distributions).
It works for Mac because although it does not have 'AppArmor', it just ignores the option and works anyway.
In Fedora, @srijitamukherjee has tested it and we have the same problem. So I have created [THREESCALE-4171](https://issues.redhat.com/browse/THREESCALE-4171) for that.

### DIFF
--- a/SETUP_ORACLE.md
+++ b/SETUP_ORACLE.md
@@ -1,0 +1,39 @@
+# Setup Oracle
+
+## Install Oracle Instant Client
+
+1. Go to [the official Oracle Instant Client Downloads site](https://www.oracle.com/database/technologies/instant-client/downloads.html) and download the following **.rpm** packages for your operative system:
+
+  - Basic
+  - SQL Plus
+  - SDK
+
+2. Create a folder in `/opt/oracle` if it does not exist yet (with `mkdir -p /opt/oracle`) and save these packages there.
+
+3. Install the dependency `libaio1` and the package `alien`.
+In Ubuntu that is done with `sudo apt-get install libaio1 alien`.
+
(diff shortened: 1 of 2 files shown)


SETUP_ORACLE.md
```diff
@@ -0,0 +1,39 @@
...
+  - Basic
+  - SQL Plus
+  - SDK
+
+2. Create a folder in `/opt/oracle` if it does not exist yet (with `mkdir -p /opt/oracle`) and save these packages there.
+
+3. Install the dependency `libaio1` and the package `alien`.
+In Ubuntu that is done with `sudo apt-get install libaio1 alien`.
+
+4. Go to `/opt/oracle` through the terminal (with `cd /opt/oracle`) and use alient to convert all those **.rpm** packages to **.deb** (with `sudo alien --to-deb --scripts *.rpm`).
```
hallelujah:
```suggestion
4. Go to `/opt/oracle` through the terminal (with `cd /opt/oracle`) and use **alien** to convert all those **.rpm** packages to **.deb** (with `sudo alien --to-deb --scripts *.rpm`).
```

hallelujah: You can also download the zip files. Unzip them in a directory and point `ORACLE_HOME` environment to this directory
That's what we do on the Dockerfiles

Martouta: Cool, thanks! :muscle:

SETUP_ORACLE.md
```diff
@@ -0,0 +1,39 @@
+# Setup Oracle
```
duduribeiro: 💯 🥇

PR #1498, opened 2019-12-19
Replace our sidekiq log for sidekiq throttled

Martouta:
Replaces our implementation of "Sidekiq Lock" for the Gem "Sidekiq Throttled" which we could also use for more workers.

Idea [suggested](https://github.com/3scale/porta/pull/1432#issuecomment-555435662) by Gui:
image

Continuation of https://github.com/3scale/porta/pull/1432

I have tested it also in preview to try to do it for many instances at once and having a backend. Although, I didn't know how to test it properly, so what I did is:
1. `RAILS_ENV=preview bundle exec rake backend:storage:rewrite` which uses internally `perform_now` in my PR. it shows a decent percentage output but that only means that it does not raise an error XD however this is not what I really meant to check in preview because it is not asynchronous.
2. Deleting a few providers, hoping that some callbacks will be triggered for its associations, like for its metrics the `sync_backend`  and that apparently worked according to the logs that I downloaded from preview:
image


app/lib/backend/model_extensions/metric.rb
```diff
@@ -28,11 +28,11 @@ def sync_backend!
       end
 
       def sync_backend_for_service(service)
-        ::BackendMetricWorker.perform_async(service.backend_id, id, extended_system_name)
+        ::BackendMetricWorker.perform_later(service.backend_id, id)
```
duduribeiro: 💯

app/workers/backend_metric_worker.rb
```diff
@@ -1,44 +1,19 @@
...
+  sidekiq_throttle({
+                     concurrency: {
+                       limit: 1,
+                       key_suffix: ->(service_id, metric_id, *) { "service:#{service_id}/metric:#{metric_id}" },
+                       ttl: 15.minutes.to_i
+                     }
+                   })
 
-  def save_or_delete_metric(service_backend_id, metric_id)
+  def perform(service_backend_id, metric_id)
```
duduribeiro: 🎖

app/workers/backend_metric_worker.rb
```diff
@@ -1,44 +1,19 @@
...
-
-  def retry_job(backend_id, metric_id, *args)
-    raise LockError if last_attempt?
-    self.class.perform_async(backend_id, metric_id, *args)
-  end
+  sidekiq_throttle({
+                     concurrency: {
+                       limit: 1,
+                       key_suffix: ->(service_id, metric_id, *) { "service:#{service_id}/metric:#{metric_id}" },
+                       ttl: 15.minutes.to_i
```
hallelujah: The lock was 1 hour before now it is 15 minutes.

https://github.com/3scale/porta/pull/1498/files#diff-81adfba00f06f2460ffd621178a978b5L13

Martouta: No special reason. I didn't know the ideal number. I tried it and seemed to work properly and that number sounds reasonable to me. But I definitely don't mind any number in particular.

app/workers/backend_metric_worker.rb
```diff
@@ -1,44 +1,19 @@
...
+  queue_as :backend_sync
 
-  sidekiq_options queue: :backend_sync,
-                  retry: 3,
-                  lock: {
-                    timeout: 1.hour.in_milliseconds,
-                    name: proc { |*args| BackendMetricWorker.lock_workers(*args) }
-                  }
-  sidekiq_retry_in do |_count|
-    5.minutes.to_i
```
guicassolato: We could have kept this in case we wanted:

```ruby
  rescue_from(StandardError) do # <=== perhaps with something better than `StandardError` here
    retry_job wait: 5.minutes
  end 
```

Martouta: @guicassolato I am on PTO and I will come back the 13th. Feel free to push a PR for it if you want or we can talk about it when I come back 😁

PR #1500, opened 2019-12-19
Move lib/string_hacks into app/lib

duduribeiro:
Extracted from #1396

lib/string_hacks into app/lib


hallelujah:
No this one is also required in initializer so it should not be moved

PR #1501, opened 2019-12-19
Move lib/normalize_path_attribute into app/lib

duduribeiro:
Extracted from #1396

lib/normalize_path_attribute into app/lib


app/lib/normalize_path_attribute.rb
```diff
@@ -43,16 +45,16 @@ def verify_path_format(*attrs)
...
-      _value.gsub!(/[^\/\w-]+/, '-') # only letters, digits, - and _ are allowed
-      _value.gsub!(/[-]+/, '-')      # no doble dashes
-      _value.gsub!(/\A[-]+/, '')     # no dashes at the beginning
-      _value.gsub!(/[-]+\Z/, '')     # no dashes at the end
-      _value.gsub!(/\/-/, '/')       # no dash after slashes
-      _value.gsub!(/-\//, '/')       # no dash before slashes
-     end
-    _value
+  def normalize_path(old_path)
+    path = old_path
```
Martouta:
```suggestion
    path = old_path.dup
```

Because otherwise, they are 2 pointers to the same spot in memory:
image
image

And seeing its usages, this is mean to be used as a real non-bang method, giving a copy of the string nornalized and not editing the original one.

https://github.com/3scale/porta/blob/5ad9e2a5e96759b79756658999e8fe977578c7c9/lib/normalize_path_attribute.rb#L34

https://github.com/3scale/porta/blob/5ad9e2a5e96759b79756658999e8fe977578c7c9/app/forms/onboarding/request_form.rb#L15-L17

https://github.com/3scale/porta/blob/5ad9e2a5e96759b79756658999e8fe977578c7c9/app/forms/onboarding/request_form.rb#L61-L65

PR #1508, opened 2019-12-20
Move lib/system_name into app/lib

duduribeiro:
Extracted from #1396

lib/system_name into app/lib


app/lib/system_name.rb
```diff
@@ -26,37 +28,37 @@ def has_system_name(opts = {})
...
-        end
-
-        self.system_name = internal_system_name
+  def generate_system_name!
+    if system_human_name.present?
+      slug = system_human_name.parameterize.underscore
+      self.system_name = slug.presence || internal_system_name
+    else
+      # TODO: remove this branch when Feature#name validation is added
+      # and this ticket resolved https://github.com/3scale/system/issues/908
```
Martouta: This probably can already be cleaned but I won't blame u if u decide that it is out of scope :smile: But then I would appreciate creating a Jira issue about this and writing the link here in a comment :stuck_out_tongue:

duduribeiro: Added as comment: https://issues.redhat.com/browse/THREESCALE-4179

PR #1509, opened 2019-12-20
Move unrequired files from lib/three_scale into app/lib

duduribeiro:
Extracted from #1396

unrequired files from lib/three_scale into app/lib


config/initializers/sidekiq.rb
```diff
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require 'three_scale/sidekiq_logging_middleware'
```
hallelujah: I think because it is required we should not move into app/lib

PR #1513, opened 2019-12-20
Remove lib from autoload

duduribeiro:
This commit removes the lib directory from the autoload path. It also
forces the inclusion of the database files that are still present in the
folder.

### DIFF
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -4,7 +4,6 @@ System::Application.configure do
   # Disable css/jquery animations in tests, makes percy much happier
   config.middleware.use Rack::NoAnimations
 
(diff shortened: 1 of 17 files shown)


config/environments/test.rb
```diff
@@ -4,7 +4,8 @@
   # Disable css/jquery animations in tests, makes percy much happier
   config.middleware.use Rack::NoAnimations
 
-  config.eager_load = false # true if you use a tool to preload your test environment
+  config.eager_load = true
+  config.enable_dependency_loading = false
```
hallelujah: OK so lets enable that everywhere (config/application.rb) but in development

PR #1514, opened 2019-12-23
Move lib/progress_counter into app/lib

duduribeiro:
lib/progress_counter into app/lib


duduribeiro:
@hallelujah yeah, it is required only for some migrations.

PR #1515, opened 2020-01-03
Validation of labels of domain names according to RFC 1035

guicassolato:
According to RFC 1035, domain host names must be 255 characters-long or less, and its labels must be 63 characters-long or less. Currently we do not validate domain names against this constraint which is causing OpenShift to reject non-compliant routes created by Zync.

One not-so-obvious case where this issue can cause us trouble is when creating a new product. The `system_name` of the product, chosen by the user, will end up being used to generate the proxy endpoints (depending on the sandbox proxy config). A sufficiently long `system_name` may produce a domain name containing a lable that is longer than 63 characters. The route will then fail to be created or updated by Zync, requiring a manual intervention by the user to fix it. The creation of the product succeeds in this case though; yet, the endpoints cannot be used.

This PR fixes this behavior by enforcing every domain name validated with https://github.com/3scale/porta/blob/bdc91b894eac16fbd81afd4f05198eb5cb8beee9/app/validators/uri_validator.rb to comply with the limits of characters specified by the RFC. Moreover, it makes proxy endpoints to be validated using this validator instead of using the old simple regex https://github.com/3scale/porta/blob/bdc91b894eac16fbd81afd4f05198eb5cb8beee9/app/models/proxy.rb#L26.

The PR also makes sure that validation errors on the proxy endpoints due to this new requirement are propagated to the service, specifically marking the `system_name` attribute. This is done with the support of https://github.com/3scale/porta/blob/bdc91b894eac16fbd81afd4f05198eb5cb8beee9/app/services/service_creator.rb.

Closes [THREESCALE-2932](https://issues.redhat.com/browse/THREESCALE-2932).

### DIFF
--- a/app/validators/uri_validator.rb
+++ b/app/validators/uri_validator.rb
@@ -8,7 +8,7 @@ class UriValidator < ActiveModel::EachValidator
   def validate_each(record, attribute, value)
     valid = begin
               uri = URI.parse(value)
-              valid_scheme?(uri.scheme) && uri.host.present? && !forbidden_part?(record, uri)
+              valid_scheme?(uri.scheme) && valid_host?(uri.host) && !forbidden_part?(record, uri)
             rescue URI::InvalidURIError
               false
             end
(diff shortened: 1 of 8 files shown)


app/validators/uri_validator.rb
```diff
@@ -26,6 +26,10 @@ def valid_scheme?(scheme)
     end
   end
 
+  def valid_host?(host)
+    host.present? && (host.size <= 255) && !host.split('.').any? { |label| label.size > 63 }
```
hallelujah: I would separate it in another method
```suggestion
    host.present? && valid_parts?
```

```ruby
def valid_domain_parts
   host.split('.').map(&:size).any?(&63.method(:'>'))
end

PR #1516, opened 2020-01-03
Sparse BackendMetricWorker retries by 5 minutes from each other

guicassolato:
`BackendMetricWorker` was [recently refactored](https://github.com/3scale/porta/pull/1498) as an `ApplicationJob`, removing the 5-minute wait between job retries. See https://github.com/3scale/porta/pull/1498/files#r362426938.

This PR reinstantes the wait between job retries.

### DIFF
--- a/app/workers/backend_metric_worker.rb
+++ b/app/workers/backend_metric_worker.rb
@@ -13,6 +13,10 @@ class BackendMetricWorker < ApplicationJob
                      }
                    })
 
+  rescue_from StandardError do |_exception|
+    retry_job wait: 5.minutes
+  end
+
   def perform(service_backend_id, metric_id)
     metric = Metric.find_by(id: metric_id)
     service = Service.find_by(id: service_backend_id)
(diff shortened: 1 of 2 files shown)


guicassolato:
`BackendMetricWorker::LockError` was one of the reasons the job could fail, but it wasn't the only one. In all cases, the retry used to be after a 5-minute wait.

Martouta:
@guicassolato right, but the only reason why we were retrying after 5 minutes was the lock. The other reasons didn't need to retry after 5 minutes and were just a side effect.

Anyway, I am not against it. I approved, so for me feel free to merge if you prefer it that way :smile:

guicassolato:
> the only reason why we were retrying after 5 minutes was the lock

Says who? Is this documented somewhere? Maybe the wait was introduced  for other reason, like to give apisonator some rest in case of connection failure.

In fact, although as part of the same PR, the 5-minute wait to retry was introduced _before_ the lock: https://github.com/3scale/system/commit/91e409f940ad0917f429b7bd4036881b541e990c

PR #1517, opened 2020-01-08
Rails data migrations

guicassolato:
This PR adds https://github.com/OffgridElectric/rails-data-migrations to manage data migrations in porta separately from any DDL code and yet without having to define them as rake tasks.

It also cleans up 4 rake tasks introduced in 2.7, meant to be executed only once, to handle legacy data of Saas and on-prem customers migrating from 2.6. The tasks were at first rewritten as data migration files, but since we won't need any of that to run again in 2.8, then so those files (and corresponding tests) will most likely be removed before we merge this PR. The 4 rake tasks are:
- `services:create_backend_apis`
- `service:update_metric_owners`
- `proxy:update_proxy_rule_owners`
- `data_migration:fix_backend_config_path`

Closes [THREESCALE-3630](https://issues.redhat.com/browse/THREESCALE-3630)

----

We reviewed 3 options of gems to handle DML code. This is the summary of our findings about them:

https://github.com/ilyakatz/data-migrate
- 22 contributors 👍
- Recently updated 👍
- Rails 5+ 😕
- It sorts data migrations mixed to DDL migrations 👎 https://github.com/ilyakatz/data-migrate/blob/4db2dba7d53e73cac06aa9bfab136c10aa38367b/tasks/databases.rake#L357

https://github.com/ka8725/migration_data
- 6 contributors 👎
- ~3 years outdated 👎
- Rails 4 and 5 👎
- DML code goes inside normal `ActiveRecord::Migration`s 👎

https://github.com/OffgridElectric/rails-data-migrations
- 5 contributors 👎
- ~2 years-old 👎
- Rails 4+ 👍
- Data migrations triggered through an independent rake task 👍

----

Useful rake tasks of https://github.com/OffgridElectric/rails-data-migrations:

```
# List data migrations pending to run
rake data:migrate:pending

# Run pending data migrations
rake data:migrate

# Rollback data migrations to a `VERSION`
VERSION=<...> rake data:migrate:down

# Mark all pending migrations as 'complete' (as opposed to 'pending')
rake data:reset
```

To generate a new data migration file:

```
rails generate data_migration my_data_migration
```

### DIFF
--- a/db/data_migrations/20200108105400_create_backend_apis.rb
+++ b/db/data_migrations/20200108105400_create_backend_apis.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+require 'progress_counter'
+
+class CreateBackendApis < ActiveRecord::DataMigration
+  def up
+    services = Service.where.has { not_exists(BackendApiConfig.except(:order).select(:id).by_service(BabySqueel[:services].id)) }
+
+    return puts 'Nothing to do.' if services.empty?
+
+    progress = ProgressCounter.new(services.count)
+
+    services.includes(:backend_api_configs).find_each(batch_size: 100) do |service|
+      progress.call do
+        next if service.backend_api_configs.any?
+
+        service_name = service.name
+        backend_api = service.account.backend_apis.create(
+          system_name: service.system_name,
+          name: "#{service_name} Backend",
+          description: "Backend of #{service_name}",
+          private_endpoint: service.proxy&.[]('api_backend')
+        )
+
+        service.backend_api_configs.create(backend_api: backend_api, path: '/')
+      end
+    end
+  end
+end
--- a/db/data_migrations/20200108171934_update_metric_owners.rb
+++ b/db/data_migrations/20200108171934_update_metric_owners.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+require 'progress_counter'
+
+class UpdateMetricOwners < ActiveRecord::DataMigration
+  def up
+    progress = ProgressCounter.new(Metric.count)
+    Metric.find_each do |metric|
+      metric.update_columns(owner_id: metric.service_id, owner_type: 'Service') unless metric.owner_type?
+      progress.call
+    end
+  end
+end
--- a/db/data_migrations/20200108173336_update_proxy_rule_owners.rb
+++ b/db/data_migrations/20200108173336_update_proxy_rule_owners.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+require 'progress_counter'
+
+class UpdateProxyRuleOwners < ActiveRecord::DataMigration
+  def up
+    progress = ProgressCounter.new(ProxyRule.count)
+    ProxyRule.find_each do |proxy_rule|
+      proxy_rule.update_columns(owner_id: proxy_rule.proxy_id, owner_type: 'Proxy') unless proxy_rule.owner_type?
+      progress.call
+    end
+  end
+end
--- a/db/data_migrations/20200108174841_fix_backend_config_path.rb
+++ b/db/data_migrations/20200108174841_fix_backend_config_path.rb
@@ -1,8 +1,9 @@
 # frozen_string_literal: true
 
-namespace :data_migration do
-  desc 'Backfill Paths for Backend Api Configs'
-  task fix_backend_config_path: :environment do
+require 'progress_counter'
+
+class FixBackendConfigPath < ActiveRecord::DataMigration
+  def up
     query = System::Database.oracle? ? BackendApiConfig.where('path is NULL') : BackendApiConfig.where(path: '')
     progress = ProgressCounter.new(query.count)
     BackendApiConfig.transaction do
--- a/db/migrate/20190805135829_backfill_metric_owners.rb
+++ b/db/migrate/20190805135829_backfill_metric_owners.rb
@@ -1,8 +1,8 @@
-require 'progress_counter'
+require Rails.root.join('db', 'data_migrations', '20200108171934_update_metric_owners')
 
 class BackfillMetricOwners < ActiveRecord::Migration
   def up
     Metric.reset_column_information
-    Rake::Task['services:update_metric_owners'].invoke
+    UpdateMetricOwners.new.up
   end
 end
--- a/test/test_helpers/data_migrations.rb
+++ b/test/test_helpers/data_migrations.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+# Test your data migrations as follows.
+#
+# require 'test_helper'
+#
+# module DataMigrations
+#   class MyDataMigrationTest < DataMigrationTest
+#     test 'does something' do
+#       migration = MyDataMigration.new
+#       assert_something do
+#         migration.up
+#       end
+#     end
+#   end
+# end
+
+module DataMigrations
+  class DataMigrationTest < ActiveSupport::TestCase ; end
+end
+
+RailsDataMigrations::LogEntry.create_table unless ActiveRecord::Base.connection.table_exists? RailsDataMigrations::LogEntry.table_name
+
+Dir[Rails.root.join *%w[db data_migrations *.rb]].each { |file| require file }
--- a/Gemfile.base
+++ b/Gemfile.base
@@ -19,6 +19,7 @@ gem 'protected_attributes', '~> 1.1.4'
 gem 'rails-observers'
 
 gem 'strong_migrations'
+gem 'rails-data-migrations'
 
 group :assets do
   gem 'coffee-rails', '~> 4.2'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -529,6 +529,8 @@ GEM
       bundler (>= 1.3.0, < 2.0)
       railties (= 4.2.11.1)
       sprockets-rails
+    rails-data-migrations (1.2.0)
+      rails (>= 4.0.0)
     rails-deprecated_sanitizer (1.0.3)
       activesupport (>= 4.2.0.alpha)
     rails-dom-testing (1.0.9)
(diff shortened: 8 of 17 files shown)


db/data_migrations/20200108105400_create_backend_apis.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+require 'progress_counter'
+
+class CreateBackendApis < ActiveRecord::DataMigration
```
guicassolato: Not sure if we should still be doing this. Perhaps this migration is outdated.

We originally decided to create one Backend per existing service because of Saas and on-prem customers migrating to 2.7. However, 2.7 also allows a Product to have no Backend at all. So, in on-prem only, where there's a rake task that triggers this migration on every deploy, system will always create Backend records for every Product missing a Backend. This smells like a bug to be, which this PR will also fix by turning this migration again into... well... a migration, rather than a rake task that triggers on every deploy.

db/data_migrations/20200108105400_create_backend_apis.rb
```diff
@@ -0,0 +1,29 @@
...
+    services.includes(:backend_api_configs).find_each(batch_size: 100) do |service|
+      progress.call do
+        next if service.backend_api_configs.any?
+
+        service_name = service.name
+        backend_api = service.account.backend_apis.create(
+          system_name: service.system_name,
+          name: "#{service_name} Backend",
+          description: "Backend of #{service_name}",
+          private_endpoint: service.proxy&.[]('api_backend')
```
guicassolato: Maybe the reason why the "bug" I reported in https://github.com/3scale/porta/pull/1517/files#r364296373 is not causing bigger problems is because of the removal `api_backend` also in 2.7. In the end, only services whose proxy has an `api_backend`, i.e. old services created before 2.7, will get a corresponding Backend created.

db/data_migrations/20200108105400_create_backend_apis.rb
thomasmaas: Valid for on-prem upgrades from 2.6 to 2.7. Plus we already ran it on saas. So, no longer needed?

guicassolato: Indeed, @thomasmaas. That's why it's still a draft :) The reason why I decided to move this rake task only was to provide an example of how data migrations would be written if we introduce the gem.

Now our options for this migration specifically are:
1. remove it completely (together with its original rake task)
2. skip it in the code with an ugly `return` in the beginning as we did a couple times in the past (meh)
3. inject '20200108105400' into the `data_migrations` table to prevent it from running again

thomasmaas: I'd pick 1 but something tells me you would pick 3

guicassolato: Not at all. I'd pick one too. No code is better than any code.

guicassolato: Extracted to a separate commit so we can remove it later. But first I'd like to get some feedback from the team about this is way to handle the data migrations.

test/test_helpers/data_migrations.rb
```diff
@@ -0,0 +1,24 @@
...
+#   end
+# end
+
+module DataMigrations
+  class DataMigrationTest < ActiveSupport::TestCase ; end
+end
+
+RailsDataMigrations::LogEntry.create_table unless ActiveRecord::Base.connection.table_exists? RailsDataMigrations::LogEntry.table_name
+
+Dir[Rails.root.join *%w[db data_migrations *.rb]].each { |file| require file }
```
guicassolato: TBH, we don't _need_ this file.
- Loading the migrations for testing could be done in https://github.com/3scale/porta/blob/master/test/test_helper.rb
- Making sure the `data_migrations` tables exists too
- `DataMigrations::DataMigrationTest` is not essential

The idea of putting it here was to set the convention somewhere. It's open to suggestions anyways.

guicassolato: Alternative

```ruby
RailsDataMigrations::Migrator.migrations(RailsDataMigrations::Migrator.migrations_path).map(&:filename).each { |file| require file }
```

test/test_helpers/data_migrations.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
```
guicassolato: TBH, we don't _need_ this file.
- Loading the migrations for testing could be done in https://github.com/3scale/porta/blob/master/test/test_helper.rb
- Making sure the `data_migrations` tables exists too
- `DataMigrations::DataMigrationTest` is not essential

The idea of putting it here was to set the convention somewhere. It's open to suggestions anyways.

db/data_migrations/20200108171934_update_metric_owners.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+require 'progress_counter'
+
+class UpdateMetricOwners < ActiveRecord::DataMigration
```
guicassolato: We probably don't need this one either.

db/data_migrations/20200108173336_update_proxy_rule_owners.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+require 'progress_counter'
+
+class UpdateProxyRuleOwners < ActiveRecord::DataMigration
```
guicassolato: Same as https://github.com/3scale/porta/pull/1517/files?file-filters%5B%5D=.rb#r364351526

db/data_migrations/20200108174841_fix_backend_config_path.rb
```diff
@@ -1,8 +1,9 @@
 # frozen_string_literal: true
 
-namespace :data_migration do
-  desc 'Backfill Paths for Backend Api Configs'
-  task fix_backend_config_path: :environment do
+require 'progress_counter'
+
+class FixBackendConfigPath < ActiveRecord::DataMigration
```
guicassolato: Same as https://github.com/3scale/porta/pull/1517/files?file-filters%5B%5D=.rb#r364351526

db/data_migrations/20200108105400_create_backend_apis.rb
Martouta: I also prefer the `option 1` for this. We only needed that for 2.7, so in that point in time the code will always be there. But I wouldn't keep it and change it for it not to run XDD I find it pointless and I am definitely not a big fan of keeping code just for the sake of it because it is only useful to ask for problems.

kill it with fire

Martouta:
I like the overall idea :+1: 

And for me the selection of the gem is great, specially for this:
`Data migrations triggered through an independent rake task`

The 22 contributors vs 5 contributors are not like we are talking about hundreds of people anyway... both are small projects and does not really matter how many contributors they have or even when they were updated for the last time because they are very small and unlikely to get out of our hands to change by ourselves if we ever need it.

db/data_migrations/20200108105400_create_backend_apis.rb
duduribeiro: I prefer the 1st also

hallelujah: Just remove it. If I understand it well like migration, it is based on versioning so next new install will not run those data migrations. Past is past: option 1

db/migrate/20190805135829_backfill_metric_owners.rb
```diff
@@ -1,8 +1,8 @@
-require 'progress_counter'
+require Rails.root.join('db', 'data_migrations', '20200108171934_update_metric_owners')
 
 class BackfillMetricOwners < ActiveRecord::Migration
   def up
     Metric.reset_column_information
-    Rake::Task['services:update_metric_owners'].invoke
+    UpdateMetricOwners.new.up
```
hallelujah: No need of this anymore right? If we remove the task?

test/test_helpers/data_migrations.rb
```diff
@@ -0,0 +1,24 @@
...
+#       end
+#     end
+#   end
+# end
+
+module DataMigrations
+  class DataMigrationTest < ActiveSupport::TestCase ; end
+end
+
+RailsDataMigrations::LogEntry.create_table unless ActiveRecord::Base.connection.table_exists? RailsDataMigrations::LogEntry.table_name
```
hallelujah: Is this table the same as our log_entries? It might be confusing or lead to errors

hallelujah:
My question is how to use it and to test it, any guidelines?

Scenario:

1. I am on version 4.0 and want to upgrade to version 4.1
2. I have mixed DDL and DML, because of historical code:

- Added new columns to a model (e.g splitting a column `full_name` into `last_name` and `first_name`)
- Migrated some data, filling up `last_name` and `first_name` by splitting `full_name`
- Removing the column `full_name` only using `first_name` and `last_name` with 
```
def full_name
  [title, last_name, first_name].compact.join(' ')
end
```

So how do we use data migrations in this case?

db/migrate/20190805135829_backfill_metric_owners.rb
guicassolato: This migration won't be merged. It's just an example.

test/test_helpers/data_migrations.rb
guicassolato: No, it's a different one, named `data_migrations`.

guicassolato:
> My question is how to use it and to test it, any guidelines, good practice?

The scenario you described is a tricky one, @hallelujah. I guess we'll have to define our own manual of good practices.

In a case where we have a situation like that (add column B -> move data from a to B -> delete column B), I think we have no other way by run the data migration in between the two DML ones. The good news is that we can always `MyDataMigration.new.up` and when the data migrations task actually runs again in the middle, this specific one will be done and won't run again.

Not ideal, I know, but it's something.

guicassolato:
Closing this as we don't really need it right now. Whoever needs to add a data migration in the future can cherry-pick https://github.com/3scale/porta/pull/1517/commits/9d8aced3099c321f4847bd6d676751b3cad854f6 and continue from there.

PR #1518, opened 2020-01-09
[ApplicationController#check_browser] destroy user session

macejmic:

User is not being logout in `ApplicationController#check_browser`

https://issues.redhat.com/browse/THREESCALE-2632

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -93,7 +93,8 @@ class ApplicationController < ActionController::Base
     return if request_format.xml? || request_format.json?
 
     if current_user && browser_not_modern?
-      logout_keeping_session!
+      logout_killing_session!
+
       flash.now[:error] = "The browser you are using doesn't seem to support the X-Frame-Options header. That means we can't protect you against Cross Frame Scripting and thus not guarantee the security of your session. Please upgrade your browser and sign in again."
       redirect_to provider_admin_path
     end


app/controllers/application_controller.rb
```diff
@@ -93,7 +93,9 @@ def check_browser
     return if request_format.xml? || request_format.json?
 
     if current_user && browser_not_modern?
-      logout_keeping_session!
+      logout_killing_session!
```
Martouta: Aha, I see :smile: 
https://github.com/3scale/porta/blob/1506f97e3023964551c28c4a37d45481dd98b59a/app/lib/authenticated_system.rb#L175-L199

app/controllers/application_controller.rb
```diff
@@ -93,7 +93,9 @@ def check_browser
...
+      destroy_user_session!
```
Martouta: I see the content is this:
https://github.com/3scale/porta/blob/1506f97e3023964551c28c4a37d45481dd98b59a/app/lib/authenticated_system.rb#L66-L71
However, this line is not necessary because it is already called from `reset_session`:
https://github.com/3scale/porta/blob/1506f97e3023964551c28c4a37d45481dd98b59a/app/lib/authenticated_system.rb#L20-L24
which is called from the previous line inside `logout_killing_session!`:
https://github.com/3scale/porta/blob/1506f97e3023964551c28c4a37d45481dd98b59a/app/lib/authenticated_system.rb#L197

app/controllers/application_controller.rb
josemigallas: Why would you keep the cms_token if you want to reset the session?

Martouta: I don't know, it was added inside [a huge PR](https://github.com/3scale/system/commit/1bbc5636e8039a4a724eab44f7580ba1f23513f1) ages ago... without any test :woman_shrugging:

PR #1519, opened 2020-01-09
Removing platform mri 2.3

srijitamukherjee:

> Removing mri_23 - mri_25

We do not have nay open issue for this.

Adding WIP as in case, there are more ruby 2.3 related stuff which needs to be removed, we can do it on the same PR.

### DIFF
--- a/Gemfile.base
+++ b/Gemfile.base
@@ -91,10 +91,9 @@ gem 'analytics-ruby', require: false
 group :development, :test do
   gem 'bootsnap', '~> 1.3'
 
-  platform :mri_23, :mri_24, :mri_25 do
-    gem 'pry-byebug', '>= 3.7.0', require: false
-    gem 'pry-stack_explorer', require: false
-  end
+  gem 'pry-byebug', '>= 3.7.0', require: false
+  gem 'pry-stack_explorer', require: false
+  
 end
 
 gem 'nokogiri', '>= 1.6.8'
--- a/Gemfile.base
+++ b/Gemfile.base
@@ -91,9 +91,10 @@ gem 'analytics-ruby', require: false
 group :development, :test do
   gem 'bootsnap', '~> 1.3'
 
-  gem 'pry-byebug', '>= 3.7.0', require: false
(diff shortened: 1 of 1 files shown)


Gemfile.base
```diff
@@ -91,10 +91,9 @@ gem 'analytics-ruby', require: false
 group :development, :test do
   gem 'bootsnap', '~> 1.3'
 
-  platform :mri_23, :mri_24, :mri_25 do
-    gem 'pry-byebug', '>= 3.7.0', require: false
-    gem 'pry-stack_explorer', require: false
-  end
+  gem 'pry-byebug', '>= 3.7.0', require: false
```
hallelujah: You can keep ruby 2.4 and 2.5 because it does not work on JRuby

srijitamukherjee: Okay, I will add it back

Gemfile.base
```diff
@@ -91,10 +91,10 @@ gem 'analytics-ruby', require: false
 group :development, :test do
   gem 'bootsnap', '~> 1.3'
 
-  platform :mri_23, :mri_24, :mri_25 do
+  platform :mri_24, :mri_25 do
     gem 'pry-byebug', '>= 3.7.0', require: false
     gem 'pry-stack_explorer', require: false
-  end
+end
```
duduribeiro: Just align this `end` with the `platform`

Martouta:
```suggestion
  end
```

Gemfile.base
```diff
@@ -91,10 +91,10 @@ gem 'analytics-ruby', require: false
 group :development, :test do
   gem 'bootsnap', '~> 1.3'
 
-  platform :mri_23, :mri_24, :mri_25 do
+  platform :mri_24, :mri_25 do
```
Martouta: Now we don't support 2.4 anymore so it should be:
```suggestion
  platform :mri_25 do
```

Gemfile.base
srijitamukherjee: Fixed

PR #1521, opened 2020-01-09
Upgrade to ruby2 5

duduribeiro:

Upgrade ruby to version 2.5.
The current version that we use (2.4) EOL is March 31, 2020.

[https://issues.redhat.com/browse/THREESCALE-3900](https://issues.redhat.com/browse/THREESCALE-3900)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM quay.io/3scale/system-builder:ruby24
+FROM quay.io/3scale/system-builder:ruby25
 
 ARG SPHINX_VERSION=2.2.11
 ARG BUNDLER_VERSION=1.17.3
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM quay.io/3scale/system-builder:ruby25
+FROM quay.io/duduribeiro/3scale-system-builder:ruby25
 
 ARG SPHINX_VERSION=2.2.11
 ARG BUNDLER_VERSION=1.17.3
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM quay.io/duduribeiro/3scale-system-builder:ruby25
+FROM quay.io/3scale/system-builder:ruby25
 
 ARG SPHINX_VERSION=2.2.11
 ARG BUNDLER_VERSION=1.17.3
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -21,7 +21,9 @@ make help
 ```
 
 ### Entering a Running Container
+
 Download and build all the images and start a shell session inside the container:
+
 ```bash
 make bash
 ```
@@ -43,6 +45,7 @@ make dev-start
 ```
 
 or, you can run the setup and run with
+
 ```
 make default
 ```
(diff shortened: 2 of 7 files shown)


thomasmaas:
Should we run the oracle build for this one?

duduribeiro:
@thomasmaas yeah, it is already running

INSTALL.md
```diff
@@ -60,7 +60,7 @@ make dev-stop
 
 #### Ruby version
 
-The project supports **Ruby 2.4.x**.
+The project supports **Ruby 2.5.x**.
```
hallelujah: Maybe just say both?

openshift/system/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM centos/ruby-24-centos7
+FROM centos/ruby-25-centos7
```
hallelujah: I would use UBI here

`FROM registry.access.redhat.com/ubi7/ruby-25`

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM quay.io/3scale/system-builder:ruby24
+FROM quay.io/3scale/system-builder:ruby25
```
hallelujah: Maybe system-builder for ruby 2.5 should use UBI
So we are testing against the same base image as the productized version

duduribeiro:
we decided to not use UBI image on CI, cause xdg packages wasn't available for it and we do need it for installing chrome

openshift/system/contrib/scl_enable
```diff
@@ -1,4 +1,4 @@
 # This file contains automatic SCL enablement.
 unset BASH_ENV PROMPT_COMMAND ENV
-source scl_source enable rh-ruby24
+source scl_source enable rh-ruby25
```
hallelujah: Has this been tested? Like `ruby -v` returning version

Dockerfile
Martouta: https://github.com/3scale/porta/pull/1521#issuecomment-573130323

openshift/system/Dockerfile.on_prem
```diff
@@ -1,4 +1,4 @@
-FROM centos/ruby-24-centos7
+FROM centos/ruby-25-centos7
```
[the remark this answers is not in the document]
Martouta: https://github.com/3scale/porta/pull/1521#issuecomment-573130323

openshift/system/Dockerfile
Martouta: https://github.com/3scale/porta/pull/1521#issuecomment-573130323

INSTALL.md
Martouta: It is not a very strong opinion but I think that we shouldn't say that we support 2.4 after the EOL so we can make our life easier. Because if it is officially unsupported and we have to maintain it ourselves and ensure by ourselves that it keeps working... that sounds problematic.

duduribeiro: I agree with @Martouta

openshift/system/contrib/scl_enable
duduribeiro: yeah!

bash-4.2$ ruby -v
ruby 2.5.5p157 (2019-03-15 revision 67260) [x86_64-linux]

PR #1523, opened 2020-01-13
Fix pdf report file name

guicassolato:
Makes PDF reports to be named using the account ID and service ID, instead of account domain and service name.

Screenshot 2020-01-15 at 14 57 00

Closes [THREESCALE-4147](https://issues.redhat.com/browse/THREESCALE-4147)

----

- [x] ~Use service ID or sanitize the name~ _Update:_ use https://github.com/3scale/porta/blob/46f6bdb1963498e8a4e4b49a15ec110f2e01e048/app/lib/pdf/report.rb#L88
- [ ] ~Include the name of the service in the header of the PDF file~ _Update:_ already there: https://github.com/3scale/porta/blob/46f6bdb1963498e8a4e4b49a15ec110f2e01e048/app/lib/pdf/report.rb#L98
- [x] Add tests

### DIFF
--- a/app/lib/pdf/report.rb
+++ b/app/lib/pdf/report.rb
@@ -53,7 +53,7 @@ module Pdf
       metrics
       move_down 3
 
-      @report = @pdf.render_file(Rails.root.join('tmp', "#{@account.domain} #{@service.name}.pdf"))
+      @report = @pdf.render_file(pdf_file_path)
 
       self
     end
@@ -86,7 +86,11 @@ module Pdf
     end
 
     def pdf_file_name
-      [@account.id, @service.id, 'report.pdf'].join '_'
+      ['report', @account.id, @service.id].join('-') + '.pdf'
+    end
+
+    def pdf_file_path
+      Rails.root.join('tmp', pdf_file_name)
     end
 
     def print_period
--- a/app/lib/pdf/report.rb
+++ b/app/lib/pdf/report.rb
@@ -86,7 +86,7 @@ module Pdf
     end
 
     def pdf_file_name
-      ['report', @account.id, @service.id].join('-') + '.pdf'
+      ['report', @account.domain, @service.id].join('-') + '.pdf'
     end
 
     def pdf_file_path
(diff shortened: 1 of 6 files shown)


app/lib/pdf/report.rb
```diff
@@ -53,7 +53,7 @@ def generate
       metrics
       move_down 3
 
-      @report = @pdf.render_file(Rails.root.join('tmp', "#{@account.domain} #{@service.name}.pdf"))
+      @report = @pdf.render_file(Rails.root.join('tmp', "#{@account.domain} #{@service.id}.pdf"))
```
hallelujah: We could also do something like 
```suggestion
      @report = @pdf.render_file(Rails.root.join('tmp', "pdf-report-#{@account.domain}-#{@service.id}.pdf"))
```

This way no spaces ...

Martouta: why we don't use the `pdf_file_name` method?
https://github.com/3scale/porta/blob/46f6bdb1963498e8a4e4b49a15ec110f2e01e048/app/lib/pdf/report.rb#L88-L90

app/lib/pdf/report.rb
```diff
@@ -53,7 +53,7 @@ def generate
       metrics
       move_down 3
 
-      @report = @pdf.render_file(Rails.root.join('tmp', "#{@account.domain} #{@service.name}.pdf"))
+      @report = @pdf.render_file(Rails.root.join('tmp', "pdf-report-#{@account.domain}-#{@service.id}.pdf"))
```
Martouta: https://github.com/3scale/porta/pull/1523#discussion_r365832419

thomasmaas:
@guicassolato can you explain the why?

Martouta:
@thomasmaas because otherwise for 1 customer it raises `Errno::ENOENT` because tries to save something like this:

```
/our_path_of_the_release/tmp/domain https://a_url_here/a_huge_string_here/?a_string_here=another_string.pdf
```

Check [the bugnsag report](https://app.bugsnag.com/3scale-networks-sl/system/errors/5df7803da289100019ac36ca?event_id=5df7803d0054c26670180000&i=jr&m=ci) for more info :smile:

PR #1526, opened 2020-01-13
Upgrade kgio gem from 2.10.0 to 2.11.3

duduribeiro:
### What this PR does

Upgrade kgio gem from 2.10.0 to 2.11.3

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -378,7 +378,7 @@ GEM
       thor (>= 0.14, < 2.0)
     json (1.8.6)
     jwt (1.5.2)
-    kgio (2.10.0)
+    kgio (2.11.3)
     kubeclient (2.5.2)
       http (>= 0.98, < 3)
       recursive-open-struct (~> 1.0.0)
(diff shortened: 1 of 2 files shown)


Martouta:
### Why do we need it

Because otherwise, in Ruby 2.5, when we do `env USER_PASSWORD=p MASTER_PASSWORD=p MASTER_ACCESS_TOKEN=token bundle exec rake db:drop db:create db:setup`, we get:

```
ArgumentError: wrong number of arguments (given 24, expected 0)
/Users/cribeiro/code/porta/app/observers/account_observer.rb:31:in `invalidate_master_id'
/Users/cribeiro/code/porta/app/observers/account_observer.rb:11:in `after_update'
```

it is a problem in dalli gem

```
[3] pry(#<AccountObserver>)> Rails.cache.delete(‘master_account_id’)
ArgumentError: wrong number of arguments (given 24, expected 0)
from /Users/cribeiro/.asdf/installs/ruby/2.5.0/lib/ruby/gems/2.5.0/gems/dalli-2.7.10/lib/dalli/socket.rb:20:in `kgio_wait_readable’
```

PR #1527, opened 2020-01-13
Support websocket protocol in BackendApi

duduribeiro:

This commit allows the BackendApi to accept websocket protocol (ws://
and wss://) in the private_endpoint attribute.

[THREESCALE-916](https://issues.redhat.com/browse/THREESCALE-916)

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -858,7 +858,7 @@ en:
             base:
               cannot_be_destroyed_with_products: cannot be deleted because it is used by at least one Product
             private_endpoint:
-              invalid: "the accepted format is 'scheme://address(:port)(/path). Accepted schemes are http, https, ws and wss'"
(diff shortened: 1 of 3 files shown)


config/locales/en.yml
```diff
@@ -858,7 +858,7 @@ en:
             base:
               cannot_be_destroyed_with_products: cannot be deleted because it is used by at least one Product
             private_endpoint:
-              invalid: "the accepted format is 'http(s)://address(:port)(/path)'"
+              invalid: "the accepted format is 'protocol://address(:port)(/path). Accepted protocols are http and ws'"
```
guicassolato:
```suggestion
              invalid: "the accepted format is 'scheme://address(:port)(/path). Accepted schemes are http, https, ws and wss'"
```

Martouta: I also like more Gui's suggestion :+1: :smile:

PR #1531, opened 2020-01-14
[Refactor] [Tech-debt] Signup logic for provider is sparsed

Martouta:
Closes [THREESCALE-3757](https://issues.redhat.com/browse/THREESCALE-3757)


test/unit/logic/provider_signup_test.rb
```diff
@@ -1,231 +0,0 @@
-require 'test_helper'
-
-class Logic::ProviderSignupTest < ActiveSupport::TestCase
```
Martouta: Just temporarily :smile: I won't remove it all :+1:

app/controllers/provider/signups_controller.rb
```diff
@@ -62,6 +53,13 @@ def create
 
   protected
 
+  def signup_params
+    attributes = params.fetch(:account, {})
```
guicassolato: Does it make sense using this action with this param missing? How about the following? (I haven't tested it BTW.)

```suggestion
    account_params = params.require(:account).merge(sample_data: true)
    user_params = account_params.delete(:user).merge(signup_type: :new_signup, username: 'admin')
```

app/controllers/provider/signups_controller.rb
```diff
@@ -62,6 +53,13 @@ def create
...
+    account_params = attributes.except(:user).merge(sample_data: true)
```
guicassolato:
```suggestion
```

app/controllers/provider/signups_controller.rb
```diff
@@ -62,6 +53,13 @@ def create
...
+    user_params    = attributes.fetch(:user, {}).merge(signup_type: :new_signup, username: 'admin')
```
guicassolato:
```suggestion
```

app/controllers/partners/providers_controller.rb
```diff
@@ -3,27 +3,19 @@ class Partners::ProvidersController < Partners::BaseController
   before_action :find_account, only: [:destroy, :update]
 
   def create
-    signup = Account.master.signup_provider(selected_plan) do |provider, user|
-      @account, @user = provider, user
-      provider.subdomain = "#{params[:subdomain]}-#{@partner.system_name}"
-      provider.org_name = "#{@partner.system_name}-#{params[:org_name]}"
-      provider.sample_data = true
-      provider.settings.monthly_charging_enabled = false
-      provider.extra_fields['partner'] = @partner.system_name
```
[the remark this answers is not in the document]
Martouta: Right! Needed! sorry :see_no_evil:

app/controllers/provider/signups_controller.rb
Martouta: require raises an error and fetch would just return errors of "required params missing" which happens if no param is sent at all. I am a bit scared about doing that because it is changing behavior and I already had a bad experience trying this for another endpoint.

I will try though and I will see if it is a good idea or too risky :thinking:

Martouta: I've tested it, and without this change, it raises an ugly error trying to render the view xD
image


So you are completely right on this one! Great catch!

app/lib/logic/provider_signup.rb
```diff
@@ -10,90 +10,6 @@ def signup_provider_possible?
...
-
-        if ThreeScale.config.onpremises
-          provider.force_upgrade_to_provider_plan!(application_plan)
-        else
-          ::Signup::ProviderAccountManager.set_provider_constraints(provider, application_plan)
-        end
-
-        SignupWorker.enqueue(provider)
-
-        third_party_notifications!(provider, options) unless options[:skip_third_party_notifications]
```
guicassolato: I guess this condition was here only because of the tests, right?
How are the "third party notifications" (user signup tracking) handled now while using `Signup::ProviderAccountManager`?

app/lib/logic/provider_signup.rb
```diff
@@ -10,90 +10,6 @@ def signup_provider_possible?
         ensure_master
         !!(services.default && account_plans.default && services.default.service_plans.default)
       end
-
-      def signup_provider(application_plan = nil, options = {})
-        ensure_master
```
guicassolato: I guess this is covered already at the level of each controller that now uses `Signup::ProviderAccountManager` instead of `Logic::.ProviderSignup#signup_provider`, correct?

Martouta: Now it is ensured by the controllers. Otherwise, it does not have permission for the required request.

app/lib/logic/provider_signup.rb
Martouta: > I guess this condition was here only because of the tests, right?

Right :smile: 

> How are the "third party notifications" (user signup tracking) handled now while using Signup::ProviderAccountManager?

It's a mess :scream: Sometimes it is inside the controller, like in SignupsController#create, and sometimes it is inside `/lib/tracking_notifications.rb`. Before this PR it was also here in ProviderSignup.
And refactoring this is not trivial at all. The params sometimes are different. And even the user initialized in `/lib/tracking_notifications.rb` is the `current_user` because there it mostly tracks the login. Besides, there are some methods related to analytics tracking that I think that we do not use anymore but I didn't touch just in case.

I've just made another commit because you are right that we were introducing a bug by one of the 2 controllers that I've touched but I've only added it in the controller for simplicity and I think that we should create a Jira for refactoring all this analytics tracking mess. Maybe for example in a background job called from the batch of SignupWorker.

Let's talk about this tomorrow :+1:

test/unit/account_test.rb
```diff
@@ -750,64 +750,27 @@ def test_default_service_id
...
-      @provider, @user = provider, user
-      provider.stubs(:provider_can_use?).returns(false)
-      provider.expects(:provider_can_use?).with(:api_as_product).returns(false).at_least_once
-      provider.subdomain = "foo"
-      provider.org_name = "bar"
-      provider.sample_data = false
-      user.password = user.password_confirmation = "foobar"
-      user.email = "foo@example.com"
-    end
-    assert_equal BackendApi.default_api_backend, @provider.first_service.api_backend
```
guicassolato: I remember discussing about this when these couple tests [were first introduced](https://github.com/3scale/porta/pull/1211). Back then the idea was having a way to test that with or without APIAP a default API backend would be always created on signup. IMO, we still need to test for that somehow.

Martouta: We do @guicassolato 
https://github.com/3scale/porta/blob/7c6b46b84bc40dd06e461b33e54cb6d517a303fd/test/unit/signup/account_manager_test.rb#L130-L153

The problem is we were doing the signup through 2 ways, one was the signup_provider (which is what this was testing), and another was the AccountManager which is the only one staying now and the one being tested in what I pasted above.

This is what actually does it:
https://github.com/3scale/porta/blob/7c6b46b84bc40dd06e461b33e54cb6d517a303fd/app/lib/signup/provider_account_manager.rb#L36-L38

guicassolato: Perfect!

PR #1532, opened 2020-01-14
Validation of ActiveDocs based on OAS 3.0.X

guicassolato:
Adds validation to `ApiDocs::Service.specification` based on OpenAPI 3.0.X schema.

Closes [THREESCALE-770](https://issues.redhat.com/browse/THREESCALE-770)

**TODOs**

- [x] Deal with possible multiple servers (?) → picking the first one for now
- [ ] ~Add support to simple base paths (when the `scheme://host(:port)` part is missing)~ We'll need to deal with this once we know what to do with the servers (e.g. create a Backend?)
- [x] Add support to OpenAPI servers using variables (aka "server templating")
- [ ] ~Find a solution to `V30#swagger?`~ Leave for when we touch the UI, because the method is only needed to decide which view template to render in the UI: https://github.com/3scale/porta/blob/60a8465289e332c8ea1ed1a5ffd0311b450c14b6/app/controllers/admin/api_docs/base_controller.rb#L10
- [x] Tests
- [ ] ~Convert yaml spec (?)~ If so, leave it for a future PR when the UI part is in place as well

### DIFF
--- a/app/lib/three_scale/open_api/url_resolver.rb
+++ b/app/lib/three_scale/open_api/url_resolver.rb
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+
+require 'addressable/template'
+
+module ThreeScale
+  module OpenApi
+    class UrlResolver
+      def initialize(specification)
+        @specification = specification
+      end
+
+      def servers
+        @specification.fetch('servers', []).map do |server|
+          url = server['url']
+          variables = server['variables']
+          variables ? resolve_servers_from_variables(url, variables) : url
+        end.flatten
+      end
+
+      def self.resolve_servers_from_variables(server_url, variables = {})
+        variable_names = variables.keys
+        variable_values = variables.values.map { |variable| variable['enum'] || [variable['default']] }
+        value_combinations = variable_values.reduce(&:product).map(&Array.method(:wrap)).map { |values| variable_names.zip(values).to_h }
+        value_combinations.map { |values| Addressable::Template.new(server_url).expand(values).to_s }
+      end
+
+      delegate :resolve_servers_from_variables, to: 'self.class'
+    end
+  end
+end
--- a/app/lib/three_scale/swagger/specification.rb
+++ b/app/lib/three_scale/swagger/specification.rb
@@ -35,10 +35,8 @@ module ThreeScale
 
       class V30 < VBase
         def base_path
-          server = @doc.fetch('servers', []).first&.fetch('url', '').presence
-          # FIXME: What about...?
-          # - simple base paths (missing host and scheme) (They are supported as valid servers in OAS 3.0.X.)
-          # - server "templates"
+          servers = ThreeScale::OpenApi::UrlResolver.new(@doc).servers
+          servers.first
         end
 
         JSON_SCHEMA = {'$ref' => 'https://spec.openapis.org/oas/3.0/schema/2019-04-02'}.freeze
--- a/app/lib/three_scale/swagger/specification.rb
+++ b/app/lib/three_scale/swagger/specification.rb
@@ -1,7 +1,6 @@
 module ThreeScale
   module Swagger
     class Specification
-
       class VBase
         def initialize(spec)
           @doc = spec.doc
@@ -148,28 +147,22 @@ module ThreeScale
 
       # Check if this specification is swagger thus it can be displayed in swagger-ui
       def swagger_1_2?
-        @doc.fetch("swaggerVersion", 0).to_f >= 1.2
+        doc_version.to_f >= 1.2
       end
 
+      alias swagger_1_2_or_newer? swagger_1_2?
+
       def swagger_2_0?
-        @doc.fetch("swagger", 0).to_f >= 2.0
+        doc_version.to_f >= 2.0
       end
 
       def openapi_3_0?
-        @doc.fetch("openapi", 0).to_f >= 3.0
+        doc_version.to_f >= 3.0
       end
 
       # Falls back to "1.0" if version can"t be determined
       def swagger_version
-        if swagger_1_2?
-          @doc["swaggerVersion"]
-        elsif swagger_2_0?
-          @doc["swagger"]
-        elsif openapi_3_0?
-          @doc["openapi"].match(/(3\.0)\..+/)[1] #
-        else
-          "1.0"
-        end
+        swagger_1_2_or_newer? ? doc_version.scan(/\A(\d+\.\d+)(\..+)?\Z/).flatten.first : '1.0'
       end
 
       def as_json
@@ -183,21 +176,26 @@ module ThreeScale
         @errors.empty?
       end
 
+      protected
+
+      def version_attribute
+        %w[openapi swagger swaggerVersion].find(&@doc.method(:has_key?))
+      end
+
+      def doc_version
+        @doc[version_attribute]
+      end
+
       private
 
+      def spec_version_class
+        "ThreeScale::Swagger::Specification::V#{swagger_version.sub(/\./, '')}".constantize
+      rescue NameError
+        VInvalid
+      end
+
       def init_version
-        case swagger_version
-        when '3.0'
-          V30.new(self)
-        when '2.0'
-          V20.new(self)
-        when '1.2'
-          V12.new(self)
-        when '1.0'
-          V10.new(self)
-        else
-          VInvalid.new(self)
-        end
+        spec_version_class.new(self)
       end
     end
   end
--- a/app/lib/three_scale/swagger/specification.rb
+++ b/app/lib/three_scale/swagger/specification.rb
@@ -13,6 +13,10 @@ module ThreeScale
           @doc.fetch("basePath", nil).try(:downcase)
         end
 
+        def servers
+          [base_path]
+        end
+
         def validate!
           raise "#{self.class} should implement #{__method__}"
         end
@@ -34,7 +38,6 @@ module ThreeScale
 
       class V30 < VBase
         def base_path
-          servers = ThreeScale::OpenApi::UrlResolver.new(@doc).servers
           servers.first
         end
 
@@ -48,6 +51,10 @@ module ThreeScale
 
         def swagger?
           true # FIXME: Is it really!?
+        end
+
+        def servers
+          @servers ||= ThreeScale::OpenApi::UrlResolver.new(@doc).servers
         end
       end
 
@@ -143,7 +150,7 @@ module ThreeScale
         @version = init_version
       end
 
-      delegate :base_path, :validate!, :swagger?, to: :@version
+      delegate :base_path, :servers, :validate!, :swagger?, to: :@version
 
       # Check if this specification is swagger thus it can be displayed in swagger-ui
       def swagger_1_2?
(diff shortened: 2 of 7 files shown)


app/lib/three_scale/open_api/url_resolver.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+require 'addressable/template'
+
+module ThreeScale
+  module OpenApi
+    class UrlResolver
```
guicassolato:
```suggestion
    class Parser
```

wdyt?

Martouta: No a big deal for me but imo `Parser` or `Resolver` is better because being inside `ThreeScale::OpenApi` it is already implied that we are talking about a URL :smile:

app/lib/three_scale/swagger/validator.rb
```diff
@@ -35,7 +35,7 @@ def valid_base_path?(record)
         # https://github.com/3scale/system/issues/4080
         return true if record.specification.base_path.blank?
         uri = self.class.parse_uri record.specification.base_path
-        return true if uri && ["https", "http"].include?(uri.scheme)
+        return true if uri && ['https', 'http', 'wss', 'ws'].include?(uri.scheme)
```
Martouta:
```suggestion
        return true if uri && %w[https http wss ws].include?(uri.scheme)
```

app/lib/three_scale/swagger/specification.rb
```diff
@@ -33,9 +33,26 @@ def swagger?
...
+        JSON_SCHEMA = {'$ref' => 'https://spec.openapis.org/oas/3.0/schema/2019-04-02'}.freeze
+
+        def validate!
+          @validator.fully_validate(JSON_SCHEMA).each do |error|
+            @errors.add(:base, error)
+          end
+        end
 
+        def swagger?
+          true # FIXME: Is it really!?
```
Martouta: I guess so :thinking: The name has changed but it keeps being considered `swagger`, I think :thinking:

app/lib/three_scale/swagger/specification.rb
```diff
@@ -33,9 +33,26 @@ def swagger?
         end
       end
 
+      class V30 < VBase
```
Martouta: This will require updating the tests in `test/unit/three_scale/swagger/specification_test.rb`

app/lib/three_scale/swagger/specification.rb
guicassolato: Yeah. That's weird. I think I won't touch that because it's used only to decide which view template to render in the UI. I'll leave for the UI PR to fix it if needed.

app/lib/three_scale/open_api/url_resolver.rb
Martouta: Good :+1: Makes sense :+1: :muscle:

app/lib/three_scale/swagger/specification.rb
```diff
@@ -171,14 +187,16 @@ def valid?
 
       def init_version
         case swagger_version
+        when '3.0'
```
guicassolato: Hmm. So, it's not really "3.0.X". Need to fix this.

app/lib/three_scale/swagger/specification.rb
```diff
@@ -145,12 +155,18 @@ def swagger_2_0?
         @doc.fetch("swagger", 0).to_f >= 2.0
       end
 
+      def openapi_3_0?
+        @doc.fetch("openapi", 0).to_f >= 3.0
```
hallelujah: I think there is no need to use `fetch` in this case

```suggestion
        @doc['openapi'].to_f >= 3.0
```

test/unit/three_scale/open_api/url_resolver_test.rb
```diff
@@ -0,0 +1,94 @@
...
+    expected_servers = %w[http://api.example.com https://api.example.com]
+    assert_equal expected_servers, resolver.servers
+  end
+
+  test 'url with multiple variables' do
+    json = <<~JSON
+      {
+        "servers": [
+          {
+            "url": "{protocol}://{environemnt}.example.com/api",
```
hallelujah:
```suggestion
            "url": "{protocol}://{environment}.example.com/api",
```

test/unit/three_scale/open_api/url_resolver_test.rb
```diff
@@ -0,0 +1,94 @@
...
+            "variables": {
+              "protocol": {
+                "enum": [
+                  "http",
+                  "https"
+                ],
+                "default": "https"
+              },
+              "environemnt": {
```
hallelujah:
```suggestion
              "environment": {
```

app/lib/three_scale/swagger/specification.rb
```diff
@@ -33,9 +33,26 @@ def swagger?
         end
       end
 
+      class V30 < VBase
+        def base_path
+          servers = ThreeScale::OpenApi::UrlResolver.new(@doc).servers
+          servers.first
```
hallelujah: Do we still need this for OAS 3.0?

app/lib/three_scale/swagger/validator.rb
```diff
@@ -35,7 +31,7 @@ def valid_base_path?(record)
         # https://github.com/3scale/system/issues/4080
         return true if record.specification.base_path.blank?
```
hallelujah: We should validate all base_path and not only the first one for OAS 3.0

test/unit/three_scale/swagger/specification_test.rb
```diff
@@ -24,6 +24,21 @@ class ThreeScale::Swagger::SpecificationTest < ActiveSupport::TestCase
...
   end
 
+  test 'petstore oas 3.0' do
+    doc = load_fixture 'pet-store-3.0'
+    specification = ThreeScale::Swagger::Specification.new(doc)
+
+    assert specification.valid?, 'specification should be valid'
+  end
+
+  test 'oas 3.0 base path' do
```
hallelujah: I think that `base_path` does not make anymore sense for OAS 3.0
Validation should happen for all servers, BUT we do not use it yet in our application. We still can validate all servers and picking arbitrarily the first server as base_path

test/unit/three_scale/open_api/url_resolver_test.rb
Martouta: Wow! Good catch! :see_no_evil:

app/lib/three_scale/swagger/specification.rb
guicassolato: I'm afraid we do.

app/lib/three_scale/swagger/validator.rb
guicassolato: We still don't know what do with the multiple servers. Picking the first as _the_ base path was arbitrary.

test/unit/three_scale/swagger/specification_test.rb
guicassolato: https://github.com/3scale/porta/pull/1532#discussion_r369671937

app/lib/three_scale/swagger/validator.rb
guicassolato: I'd rather not touch `base_path` in this iteration, but I've added validation of all servers anyway.

test/unit/three_scale/swagger/specification_test.rb
```diff
@@ -57,12 +72,11 @@ class ThreeScale::Swagger::SpecificationTest < ActiveSupport::TestCase
 
   private
 
+  def load_fixture file
+    File.read Rails.root.join "test", "fixtures", "swagger", "#{file}.json"
```
hallelujah: We can rely on the defined `fixtures_path` and use https://github.com/3scale/porta/blob/2ef2dc9b3addb1b75df2273530a0a24443272b72/test/unit/three_scale/policies/specification_test.rb#L8

guicassolato: I didn't really touch this, but good opportunity to improve it. Thanks.

app/lib/three_scale/swagger/specification.rb
```diff
@@ -133,27 +150,24 @@ def initialize(body)
...
       # Falls back to "1.0" if version can"t be determined
       def swagger_version
-        if swagger_1_2?
-          @doc["swaggerVersion"]
-        elsif swagger_2_0?
-          @doc["swagger"]
-        else
-          "1.0"
-        end
+        (swagger_1_2? || swagger_2_0? || openapi_3_0?) ? doc_version.scan(/\A(\d+\.\d+)(\..+)?\Z/).flatten.first : '1.0'
```
Martouta: The 'else' scenario is lacking a test.

However, I did it locally to check what would happen with a change of code here. I copied and pasted the file like `cp test/fixtures/swagger/pet-store-3.0.json test/fixtures/swagger/pet-store-undefined.json` and I changed the line of `"openapi": "3.0",` for `"openapi": "-",`.
Then I wrote this quick test:
```ruby
test 'pet store default version' do
    doc = load_fixture 'pet-store-undefined'
    specification = ThreeScale::Swagger::Specification.new doc

    assert_equal '1.0', specification.swagger_version
  end
```

And with that I see, that this code works:
```suggestion
       doc_version.scan(/\A(\d+\.\d+)(\..+)?\Z/).flatten.first || '1.0'
```

guicassolato: Try with swagger version "1.1" and you'll see.

Martouta: ooh, I am following you now :smile: In that case, I will think in a nicer solution :thinking:

Martouta: For me even this is less ugly:
```ruby
def swagger_version
  doc_version.to_f >= 1.2 ? doc_version.scan(/\A(\d+\.\d+)(\..+)?\Z/).flatten.first : '1.0'
end
```

Which is exactly the same but doing the `doc_version.to_f >= 1.2` only once... because mathematically does not makes sense to check that a number is bigger than 1.2 OR bigger than 2 OR bigger than 3 ... because if it is bigger than 2 o 3, it is also bigger than 1.2 :confused: And it is also ugly doing `swagger_1_2? ? doc_version.scan(/\A(\d+\.\d+)(\..+)?\Z/).flatten.first : '1.0'` because it is not that understandable for the meaning.

Another possible solution but CHANGES behaviour from this (so I dunno if it suits our needs) would be:
```ruby
def swagger_version
  swagger_version = doc_version.to_f.round(1)
  swagger_version = 1.0 unless swagger_version >= 1.2
  swagger_version.to_s
end
```

However, independently of the code, I still feel that we need a test for this scenario that you mentioned of '1.1' to '1.0'

guicassolato: We do have a test for the "v1.1 not-supported" case: https://github.com/3scale/porta/blob/f5dcf0bf954202a5412523035002daa38cec6eed/test/unit/api_docs/service_test.rb#L370

This check in `ThreeScale::Swagger::Specification#swagger_version` is not strictly the same thing as "we don't support swagger 1.1". The check should be read as "give me the swagger version in the spec, as long as the spec if 1.2 or newer".

I refactored following your suggestion to make this more clear. Thanks, @Martouta !

guicassolato:
> Would this also work with 3.0.1 and 3.0.2?

Yes, @thomasmaas , that's the version we're using to validate the spec. it's 3.0.X.

PR #1536, opened 2020-01-15
Remove the needed for two Gemfiles and Sidekiq Pro

duduribeiro:
We currently have two different Gemfiles, 1 for our production SaaS and
another that is used to on_prem images and for development. This was
needed because of Sidekiq pro. This commit removes this different
Gemfile and also removes SidekiqPro.


hallelujah:
Why do we remove sidekiq-pro?

duduribeiro:
@hallelujah this was one of the motivations that we discussed to remove the both Gemfiles. Since only us need it for SaaS, AFAIR @guicassolato suggested to remove the sidekiq-pro and rely only in sidekiq-batch

Martouta:
@duduribeiro I think that we should link to the Jira [THREESCALE-2117](https://issues.redhat.com/browse/THREESCALE-2117) in the description.

@hallelujah Yes, as he said. It is the latest comment in the Jira issue right now.

hallelujah:
Well the JIRA is about removing it but did we find a replacement for it? Or at least are we sure that the feature we use is not necessary anymore?

This is in the issue description:

> There are no tests for this feature so better to remove it and have a different solution.

Do we have a different solution then?

duduribeiro:
@hallelujah AFAIK, the only thing that we used for the pro is the batching. Sidekiq-batch is the replacement. Also, we run on on_prem clients that do not have the sidekiq-pro with success. cc @guicassolato

guicassolato:
1. This requires extensively testing in preview, especially billing.
2. Not sure if that's the only feature of Sidekiq Pro that we are using for Saas. Other things may include:
    - pause job queues
    - expiration of unprocessed jobs
    - search for jobs in sidekiq web console
    - server reliability on Ruby VM crash
    - client reliability on Redis networking problems

duduribeiro:
@hallelujah @Martouta @guicassolato, agree. So, let's continue using it? We can't cover all these features without it

Martouta:
@duduribeiro told @3scale/system-ruby before the following message, expecting us to give our thoughts about it:

> if we remove sidekiq-pro, we will loose:
> 1. reliable push when redis is down. (currently we don’t have problems for the deployment loose because we do a safe shutdown like @hery explained)
> 2. ACK on messages. In sidekiq-free, if some job is processing, and the sidekiq process suffer an out of memory, the job that was in processing status is lost

@hallelujah clarified:

> for the first point:
> reliable push when redis is down. (currently we don’t have problems for the deployment loose because we do a safe shutdown like @hery explained)
> This is not correct, when redis is down, on sidekiq-pro, the webserver (client) will stack in memory the jobs until it can connect to redis again

I replied:

> we need to know if we can afford that for all our sidekiq jobs
> for the ones that are idempotent and triggered by an event, we probably can afford that because we can retrigger it again by ourselves by the event
> not sure if there are other scenarios when we cannot do that

hallelujah:
@duduribeiro We do not have to cover those features :)
It is about decision making. But it has to be motivated.

duduribeiro:
Closed in favor of https://github.com/3scale/porta/pull/1581

PR #1538, opened 2020-01-16
Fix proxy domains event on background deletion of proxy

guicassolato:
A missing service while a proxy is being deleted in the background is causing a proxy domains changed event to fail to issue. We don't care about the event itself. The idea is preventing the error to be raised at this point.

Closes [THREESCALE-4248](https://issues.redhat.com/browse/THREESCALE-4248)

### DIFF
--- a/app/events/domains/proxy_domains_changed_event.rb
+++ b/app/events/domains/proxy_domains_changed_event.rb
@@ -33,6 +33,7 @@ class Domains::ProxyDomainsChangedEvent < BaseEventStoreEvent
   end
 
   def self.valid?(proxy)
-    !!proxy # TODO: check if deployment_option or domains actually changed
+    # TODO: check if deployment_option or domains actually changed
+    proxy && proxy.provider&.persisted? # Proxy's service or account may have been deleted
   end
 end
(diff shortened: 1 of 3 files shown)


Martouta:
Then I think that we also should change this:
https://github.com/3scale/porta/blob/a5dd5b2e8b18cc3f14437cc0db04002c416af46c/app/events/domains/proxy_domains_changed_event.rb#L35-L37
for:
```ruby
  def self.valid?(proxy)
    !!proxy && !!proxy.provider
  end
```
Because otherwise I am afraid that we will see in Bugsnag the 'invalid event' error raised when the provider is `nil`

app/events/domains/proxy_domains_changed_event.rb
```diff
@@ -33,6 +33,7 @@ def after_commit
   end
 
   def self.valid?(proxy)
-    !!proxy # TODO: check if deployment_option or domains actually changed
+    # TODO: check if deployment_option or domains actually changed
+    proxy && proxy.provider&.persisted? # Proxy's service or account may have been deleted
```
Martouta:
```suggestion
    proxy && proxy.provider.persisted? # Proxy's service or account may have been deleted
```

Small suggestion :smile: Because if `proxy.provider` is `nil`, `nil.persisted?` will be `false`.

PR #1540, opened 2020-01-17
[WIP] Add audit logs for application key changes

Martouta:
Closes [THREESCALE-3733](https://issues.redhat.com/browse/THREESCALE-3733)


app/models/application_key.rb
```diff
@@ -2,6 +2,7 @@ class ApplicationKey < ApplicationRecord
 
   KEYS_LIMIT = 5
 
+  audited
```
hallelujah: Maybe we do not want to audit all the fields

Martouta: which ones we do not want to? check the test to see the ones I am auditing :smile: I guess it is reasonable but I am opened to suggestions.

hallelujah: I think we should whitelist them. Example it does not make sense to audit `tenant_id`

Martouta: Makes sense :+1: :smile:

PR #1541, opened 2020-01-17
Fix AWS S3 options

guicassolato:
Related to [THREESCALE-4052](https://issues.redhat.com/browse/THREESCALE-4052)

----

This PR ensures Paperclip will consider the new configuration options introduced by https://github.com/3scale/porta/pull/1522 while uploading attachments to S3, passing them along to the `Aws::S3::Resource` instance.

- Sets `s3_options` with the `endpoint` value (matching `s3_protocol` and `s3_host_name`) if configured by the user
- Sets `force_path_style: true` by default

----

AWS S3
Screenshot 2020-01-17 at 19 32 35

Minio.io
Screenshot 2020-01-17 at 19 32 10

### DIFF
--- a/openshift/system/config/amazon_s3.yml
+++ b/openshift/system/config/amazon_s3.yml
@@ -8,6 +8,7 @@ s3: &s3
   region:            "<%= ENV['AWS_REGION'] %>"
   hostname:          "<%= ENV['AWS_HOSTNAME'] %>"
   protocol:          "<%= ENV['AWS_PROTOCOL'] %>"
+  force_path_style:  <%= ENV['AWS_PATH_STYLE'].presence || false %>
 
 preview:
   <<: *<%= ENV['FILE_UPLOAD_STORAGE'].presence || 'default' %>
--- a/app/lib/cms/s3.rb
+++ b/app/lib/cms/s3.rb
@@ -11,7 +11,7 @@ module CMS
     end
 
     def enable!
-      @enabled = true if options
+      @enabled = true if config
     end
 
     def disable!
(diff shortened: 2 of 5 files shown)


openshift/system/config/amazon_s3.yml
```diff
@@ -8,6 +8,7 @@ s3: &s3
   region:            "<%= ENV['AWS_REGION'] %>"
   hostname:          "<%= ENV['AWS_HOSTNAME'] %>"
   protocol:          "<%= ENV['AWS_PROTOCOL'] %>"
+  force_path_style:  <%= ENV['AWS_PATH_STYLE'].presence || false %>
```
guicassolato: @eguzki , I'm afraid we'll need to introduce one more env var, `AWS_PATH_STYLE`.

We cannot enforce it by default because path-style endpoints will [lose support on 30 Sep 2020](https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/). Other "API-compatible" services like minio.io, however, may still require it (at least, http://play.minio.io does).

guicassolato: If we decide not to implement this new env var, a workaround is including a dot in the name of the bucket. This will enforce path-style endpoints despite the value of the `force_path_style` configuration option. (Only valid when the protocol is `https`.)

https://github.com/aws/aws-sdk-ruby/blob/9409865777f15c7db2a8b6937fdd571082d12cce/aws-sdk-resources/lib/aws-sdk-resources/services/s3/bucket.rb#L111

eguzki: AWS_PATH_STYLE will be there

eguzki: PR with the new env var: https://github.com/3scale/3scale-operator/pull/306

hallelujah: I am not sure to understand correctly this `force_path_style` configuration.

Why do we need it? I can't see any discussion in the issue or its comments about this
I just want to understand the need of this, or lets document what is missing in the issue (for future archeologist)

guicassolato: This is because https://github.com/aws/aws-sdk-ruby tests if the name of the bucket is "dns-compatible". Whenever it sees a bucket named, for example, "mybucket", it will build an endpoint using the subdomain style: i.e. "https://mybucket.host". On the other hand, if the bucket is named "my.bucket" (i.e. not "dns-compatible"), the endpoint will be built using path style: i.e. "https://host/my.bucket".

http://play.minio.io requires path style always, no matter if the bucket name is "dns compatible" or not. Not having the option to set `force_path_style` would make 3scale incompatible with use cases where the name of the bucket is "dns-compatible" but the endpoint must still be in path-style.

PR #1542, opened 2020-01-20
Swagger ui 3 fix attempt

didierofrivia:

This PR Fixes a couple of error related to CircleCi and its resources. Also, speed up the building process and assets management.

fixes https://issues.redhat.com/browse/THREESCALE-4309

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -439,8 +439,7 @@ jobs:
           name: Precompile assets
           command: |
             cp config/examples/*.yml config/
-            bundle exec rake assets:precompile RAILS_ENV=test
-            bundle exec rake assets:precompile RAILS_ENV=production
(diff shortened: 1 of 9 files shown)


.circleci/config.yml
```diff
@@ -439,8 +439,7 @@ jobs:
           name: Precompile assets
           command: |
             cp config/examples/*.yml config/
-            bundle exec rake assets:precompile RAILS_ENV=test
-            bundle exec rake assets:precompile RAILS_ENV=production
+            bundle exec rake assets:precompile RAILS_ENV=test WEBPACKER_PRECOMPILE=false
```
hallelujah: Why removing this?

didierofrivia: Apparently after some update, the rest of the tests were using the right packs (packs-test) so we don't need to compile both. Also, better to compile webpack assets with our custom `webpack:compile` task that doesn't use yarn

PR #1544, opened 2020-01-20
Swagger-ui: 3scale look & feel

josemigallas:
Adds a CSS "patch" that overrides Swagger-ui styles and makes it look more like dev portal.

Screen Shot 2020-01-20 at 16 42 38

Screen Shot 2020-01-20 at 16 42 46

Screen Shot 2020-01-20 at 16 42 57

### DIFF
--- a/app/javascript/packs/active_docs.js
+++ b/app/javascript/packs/active_docs.js
@@ -2,4 +2,6 @@
 import SwaggerUI from 'swagger-ui'
 import 'swagger-ui/dist/swagger-ui.css'
 
+import 'ActiveDocs/swagger-ui-3-patch.scss'
+
 window.SwaggerUI = SwaggerUI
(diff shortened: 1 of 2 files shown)


thomasmaas:
@josemigallas 

- can we have some before/after screen shots?
- this would be bundled automatically for (new) customers and they can not edit this css, right?

josemigallas:
@thomasmaas 

> * can we have some before/after screen shots?

Screen Shot 2020-01-15 at 17 52 25

> * this would be bundled automatically for (new) customers and they can not edit this css, right?

@didierofrivia can confirm but I think this will affect all customers that use OAS3, not necessarily new ones. It comes from the backend, not the liquid templates, therefore it won't be visible or editable.

thomasmaas:
So its not clear why we need to fix styles as the above screenshot doesn't look broken or am I missing something?

josemigallas:
Oh, I'm sorry I misunderstood what you asked for, this is how swagger-ui 3 looks like:
Screen Shot 2020-01-17 at 17 53 24

What you can see in the comment above is the current state.

josemigallas:
Closing in favor of #1597

PR #1545, opened 2020-01-21
Prevent member users from creating Services

josemigallas:

[THREESCALE-1650: Admin Portal member user with the access "Developer Accounts -- Applications" can create new APIs](https://issues.redhat.com/browse/THREESCALE-1650)

**Verification steps** 

- Log in as a member permission an generate a token under `/p/admin/user/access_tokens`
- Try to create a service via command line, example:
    ```bash
    curl -v  -X POST "http://provider-admin.example.com.lvh.me:3000/admin/api/services.xml" -d 
    'access_token=<token>&name=my- 
    api&deployment_option=self_managed&backend_version=1&system_name=my-api'
    ```

**TODO**:
- [x] add permission check to controller
- [x] add a test case

### DIFF
--- a/app/controllers/admin/api/services_controller.rb
+++ b/app/controllers/admin/api/services_controller.rb
@@ -111,6 +111,7 @@ class Admin::Api::ServicesController < Admin::Api::ServiceBaseController
   protected
 
   def can_create
+    authorize!(:create, Service) if current_user
     head :forbidden unless current_account.can_create_service?
   end
 
--- a/test/integration/admin/api/services_controller_test.rb
+++ b/test/integration/admin/api/services_controller_test.rb
@@ -82,12 +82,46 @@ class Admin::Api::ServicesControllerTest < ActionDispatch::IntegrationTest
       assert_correct_params
     end
 
+    test 'create with provider' do
+      assert_difference(provider_services.method(:count)) do
+        post admin_api_services_path(provider_key: @provider.provider_key, format: :json), permitted_params
+        assert_response :created
+      end
+      assert_correct_params
+    end
+
+    test 'create with unauthorize provider key' do
+      Account.any_instance.stubs(can_create_service?: false)
+      assert_no_difference(provider_services.method(:count)) do
+        post admin_api_services_path(provider_key: @provider.provider_key, format: :json), permitted_params
+        assert_response :forbidden
+      end
+    end
+
+    test 'create with unauthorize admin user' do
+      Account.any_instance.stubs(can_create_service?: false)
+      assert_no_difference(provider_services.method(:count)) do
+        post admin_api_services_path(access_token: access_token_value, format: :json), permitted_params
+        assert_response :forbidden
+      end
+    end
+
     test 'create with errors in the model' do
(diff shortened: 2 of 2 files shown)


test/integration/admin/api/services_controller_test.rb
```diff
@@ -96,6 +96,16 @@ class TenantHostTest < ActionDispatch::IntegrationTest
       assert_correct_params
     end
 
+    test 'a member user cannot create a service' do
+      member = FactoryBot.create(:member, account: provider)
+      member_access_token_value = FactoryBot.create(:access_token, owner: member, scopes: %w[account_management], permission: 'rw').value
+
+      assert_no_difference(provider_services.method(:count)) do
+        post admin_api_services_path(access_token: member_access_token_value, format: :json), permitted_params.merge(forbidden_params)
```
guicassolato:
```suggestion
        post admin_api_services_path(access_token: member_access_token_value, format: :json), permitted_params
```

josemigallas: Why is that if I may ask?

Martouta: I agree with @guicassolato 

The purpose of `forbidden_params` is to see that they are filtered out and not created/updated from the params although the `permitted_params` are updated/created. But that is only for successful response codes.
For a `forbidden` response, nothing should be updated/created. Therefore it is confusing adding the `forbidden_params` as if we expected something different to happen with the different kind of params.

josemigallas: > You're absolutely right, I don't know how I missed that

app/controllers/admin/api/services_controller.rb
```diff
@@ -111,6 +111,7 @@ def destroy
   protected
 
   def can_create
+    authorize!(:create, Service) if current_user
```
guicassolato: This way is perfect for me.

Just for the record, in case one thinks that calling `CanCan::Ability#authorize!` and `head :forbidden` in the next line is not cool, a more sophisticated alternative could be:

```ruby
  def can_create
    authorize!(:create, Service) do
      raise CanCan::AccessDenied unless current_account.can_create_service?
    end
  end

  def authorize!(action, subject, *args)
    super if current_user
    yield if block_given?
    subject
  end
```

It won't break other actions and cover well the case for `Admin::Api::ServicesController#create`.

Martouta: I've commented locally the conditional and still all tests of `test/integration/admin/api/services_controller_test.rb` pass :joy: So imo we are lacking a test for the scenario of the account being able to create but not having a current_user.

josemigallas: That was what I was afraid of, couldn't we get the user directly from the access_token instead of assuming there's always a `current_user`?
@guicassolato

Martouta: we get the user from the access_token but sometimes we get the account from the api_key instead so there is no current_user. That is the scenario that we are lacking to be tested.

PR #1546, opened 2020-01-21
Audit access tokens + Requests through access tokens contain the user_id in their audits

Martouta:
Closes [THREESCALE-421](https://issues.redhat.com/browse/THREESCALE-421)

- [X] Add timestamps to access tokens to make it easier for support to debug.
- [X] Add audits for creating/updating/destroying access tokens.
- [X] When a request that contains audits is done by the API through an access token. That audit also contains the user_id as we do for the requests by the UI.

### DIFF
--- a/app/lib/api_authentication/by_access_token.rb
+++ b/app/lib/api_authentication/by_access_token.rb
@@ -5,7 +5,7 @@ module ApiAuthentication
     extend ActiveSupport::Concern
 
     def current_user
-      @current_user ||= authenticated_token&.owner or defined?(super) && super
+      @current_user ||= User.current = authenticated_token&.owner or defined?(super) && super
     end
 
     included do
--- a/app/models/access_token.rb
+++ b/app/models/access_token.rb
@@ -5,6 +5,10 @@ class AccessToken < ApplicationRecord
 
   serialize :scopes, Array
 
+  audited only: %i[owner_id scopes name permission created_at updated_at]
+
+  delegate :provider_id_for_audits, to: :owner
+
   def self.options_to_hash(options)
     options.map do |key|
       [I18n.t(key, scope: :access_token_options), key]
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -416,6 +416,10 @@ class User < ApplicationRecord
     not scope.without_ids(self.id).exists?(condition)
   end
 
+  def provider_id_for_audits
+    account.try!(:provider_id_for_audits) || provider_account.try!(:provider_id_for_audits)
+  end
+
   private
 
   def archive_as_deleted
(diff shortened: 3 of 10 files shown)


app/models/access_token.rb
```diff
@@ -5,6 +5,10 @@ class AccessToken < ApplicationRecord
 
   serialize :scopes, Array
 
+  audited only: %i[owner_id scopes name permission created_at updated_at]
+
+  delegate :provider_id_for_audits, to: :owner
```
guicassolato: `owner` is a polymorphic association of `AccessToken`, which now seems to only have `User` as type in the db of Saas. Just in case, unless or until we are 100% positive that no other type of entity can own an access token, maybe better start with:

```suggestion
  def provider_id_for_audits
    owner.try(:provider_id_for_audits)
  end
```

app/lib/api_authentication/by_access_token.rb
```diff
@@ -5,7 +5,7 @@ module ByAccessToken
     extend ActiveSupport::Concern
 
     def current_user
-      @current_user ||= authenticated_token&.owner or defined?(super) && super
+      @current_user ||= User.current = authenticated_token&.owner or defined?(super) && super
```
guicassolato: This seems to belong to a different PR.

Martouta: What do you mean? that is one of the purposes of the PR. Because the user of the audit is taken from User.current, which used to be `nil` only when the user was identified by access token, for this.

https://github.com/3scale/porta/blob/e6be7e4c686590fc7f537f723ad6567f8b709298/config/initializers/audited_hacks.rb#L180-L185

app/models/access_token.rb
Martouta: Look at this (the `class_name`):
https://github.com/3scale/porta/blob/e6be7e4c686590fc7f537f723ad6567f8b709298/app/models/access_token.rb#L2

And this:
image

app/lib/api_authentication/by_access_token.rb
guicassolato: I mean auditing the access token changes AND adding the user to the audits in general are 2 different unrelated changes IMO, even though the Jira asks both in one same ticket.

app/models/access_token.rb
guicassolato: I don't know why it's polymorphic then 🤷‍♂️

app/lib/api_authentication/by_access_token.rb
Martouta: Aha, I see :+1:

app/models/access_token.rb
Martouta: Me neither :joy: It is just overcomplicating :thinking: Maybe in the past, it needed to be polymorphic for real but not anymore :woman_shrugging:

PR #1547, opened 2020-01-22
Show hint message when protocol is not https

damianpm:

Hint message for Private Base URL is not shown when protocol is 'http'.

https://issues.redhat.com/browse/THREESCALE-3269

**Verification steps** 

- Go to **API  > Integration > Configuration** (http://provider-admin.example.com.lvh.me:3000/apiconfig/services/2/integration/edit)
- Fill input using 'http://echo-api.3scale.net:443', then 'https://echo-api.3scale.net:443'

bad

edit-ok

### DIFF
--- a/app/assets/javascripts/provider/admin/apiconfig/services/integration.js.coffee
+++ b/app/assets/javascripts/provider/admin/apiconfig/services/integration.js.coffee
@@ -166,8 +166,12 @@ $(document).on 'initialize', '#proxy', ->
       input = $(this)
       undo_button = input.siblings '.undo'
       undo_button.toggle( input_changed_from_default(input) )
-      protocolWarn = input.siblings('.inline-hints').find('.protocol-warn')
-      protocolWarn.toggle( !inputHasSameProtocols(input) )
+
+  toggle_secure_scheme_hint =  ->
+    input = $(this)
+    protocolWarn = input.siblings('.inline-hints').find('.protocol-warn')
+    protocolWarn.toggle( !inputHasSecureProtocol(input) )
+    input.toggleClass('hint-error', !inputHasSecureProtocol(input))
 
   reset_proxy_input = (event) ->
     input = $(this).siblings('input')
@@ -185,12 +189,10 @@ $(document).on 'initialize', '#proxy', ->
   compareHttpHosts = (input) ->
     extractHost(input.val()) == extractHost(input.data('default'))
 
-  inputHasSameProtocols = (input) ->
-    actualValue = document.createElement('a')
-    actualValue.href = input.val()
-    defaultValue = document.createElement('a')
-    defaultValue.href = input.data('default')
-    actualValue.protocol == defaultValue.protocol
+  inputHasSecureProtocol = (input) ->
+    link = document.createElement('a')
+    link.href = input.val()
+    link.protocol == 'https:' || link.protocol == 'wss:'
 
   toggle_reset_buttons()
   toggle_host_header_warning()
@@ -229,6 +231,7 @@ $(document).on 'initialize', '#proxy', ->
   form = $("form.proxy")
   form.on "change keyup", "input, select", toggle_form_changed_ui
   form.on "click", ".undo", reset_proxy_input
+  form.on "change keyup", "#proxy_api_backend", toggle_secure_scheme_hint
 
   #---------------------------------------------------------------------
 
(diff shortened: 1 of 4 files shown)


config/locales/en.yml
```diff
@@ -1245,7 +1245,7 @@ en:
         api_backend_https: "Private address of your API that will be called by the API gateway."
         api_backend_http: |
           Private address of your API that will be called by the API gateway.
-          <span class="warn">For end-to-end encryption your private base URL scheme should be https</span>.
+          <span class="protocol-warn">For end-to-end encryption your private base URL scheme should be https.</span>
```
hallelujah: Hmm.... We are going to support secure websocket so this might change, maybe:

```suggestion
          <span class="protocol-warn">For end-to-end encryption your private base URL should use a secure protocol..</span>
```

app/assets/javascripts/provider/admin/apiconfig/services/integration.js.coffee
```diff
@@ -181,16 +182,15 @@ $(document).on 'initialize', '#proxy', ->
...
-    link.hostname
-
-  compareHttpHosts = (a, b) ->
-    extractHost(a) == extractHost(b)
+  inputHasSameProtocols = (input) ->
+    actualValue = document.createElement('a')
+    actualValue.href = input.val()
+    defaultValue = document.createElement('a')
+    defaultValue.href = input.data('default')
+    actualValue.protocol == defaultValue.protocol
```
hallelujah: Might be better to have a list of secure protocol instead?

Also the `inputHasSameProtocols` name does not look right to me. Change it to `inputProtocolsSecure` ?

damianpm: @hallelujah 
At this moment we are not checking if the protocol is secure or not, only if it's the same as the default data attribute. I'd say, let's just rename the method in the scope of this PR.
For other changes or improvements I'd create a new task. This file has a lot of coffeescript/jQuery spaghetti code, we should refactor into a react component or at least to ES6.

config/locales/en.yml
damianpm: I think we should change the message until websockets are supported

thomasmaas: https://github.com/3scale/porta/pull/1527

thomasmaas:
@damianpm now also works with web socket urls?

didierofrivia:
What I found weird is the highlighted input colour when it's wrong protocol... Before reading the help text below, I'd expect a red colour informing me I did something wrong. Blue is like.. yeah, all cool here.

config/locales/en.yml
```diff
@@ -1242,10 +1242,11 @@ en:
           <span>Make sure to <a href="https://docs.openshift.com/container-platform/latest/dev_guide/routes.html#creating-routes">add the correct route</a></span>
         endpoint_apicast_2_html: >
           Public address of your API gateway in the %{environment_name} environment.
-        api_backend_https: "Private address of your API that will be called by the API gateway."
+        api_backend_https: |
+          Private address of your API that will be called by the API gateway.
+          <span class="protocol-warn">For end-to-end encryption your private base URL scheme should use a secure protocol.</span>
         api_backend_http: |
           Private address of your API that will be called by the API gateway.
```
damianpm: @thomasmaas @hallelujah 
Before rendering the form  (https://github.com/3scale/porta/blob/master/app/views/api/integrations/apicast/configuration_driven/_configuration.html.slim#L27) we are checking if the protocol is 'http' or 'https', but not if 'ws' or 'wss' (https://github.com/3scale/porta/blob/master/app/helpers/api/integrations_helper.rb#L141).

After submitting the form with wrong protocol we get this message:
schemes

I think the hint messages are obsolete now, we should rephrase them, WDYT?

thomasmaas: @damianpm we are working for on prem issue, meaning apiap is enabled, meaning this url is provided at backend edit. I don't understand why it's obsolete?

Screenshot 2020-01-30 11 30 09

damianpm: I'm talking when No-APIAP in API > Integration > configuration
error

config/locales/en.yml
```diff
@@ -1242,10 +1242,11 @@ en:
           <span>Make sure to <a href="https://docs.openshift.com/container-platform/latest/dev_guide/routes.html#creating-routes">add the correct route</a></span>
         endpoint_apicast_2_html: >
           Public address of your API gateway in the %{environment_name} environment.
-        api_backend_https: "Private address of your API that will be called by the API gateway."
+        api_backend_https: |
```
thomasmaas: can't we just use this message in backend edit?

damianpm: Yes, we can

PR #1552, opened 2020-01-23
Disable update of proxy rule position when account is gone

guicassolato:
Deletion of multiple proxy rule objects in parallel, which may happen when an Account or a Backend is being deleted in the background, can deadlock on the update of the "position" of the proxy rules. This PR disables `acts_as_list` position update when a proxy rule is destroyed if (and only if) the proxy rule's owner or the account top of it has been deleted (scheduled for deletion).

Closes [THREESCALE-3845](https://issues.jboss.org/browse/THREESCALE-3845)

### DIFF
--- a/app/events/proxy_configs/affecting_object_changed_event.rb
+++ b/app/events/proxy_configs/affecting_object_changed_event.rb
@@ -12,4 +12,8 @@ class ProxyConfigs::AffectingObjectChangedEvent < ServiceRelatedEvent
       }
     )
   end
+
+  def self.valid?(proxy, *_args)
+    proxy && proxy.account&.persisted? # Proxy's service or account may have been deleted
+  end
 end
--- a/app/models/proxy_rule.rb
+++ b/app/models/proxy_rule.rb
@@ -124,6 +124,18 @@ class ProxyRule < ApplicationRecord
     owner_type == 'BackendApi'
   end
 
+  delegate :account, to: :owner, allow_nil: true
+
+  def destroy
+    if !owner || !account || owner.scheduled_for_deletion?
+      self.class.acts_as_list_no_update do
+        super
+      end
+    else
+      super
+    end
+  end
+
   protected
 
   def querystring_parameter_keys
--- a/app/models/proxy_rule.rb
+++ b/app/models/proxy_rule.rb
@@ -126,14 +126,12 @@ class ProxyRule < ApplicationRecord
 
   delegate :account, to: :owner, allow_nil: true
 
-  def destroy
-    if !owner || !account || owner.scheduled_for_deletion?
-      self.class.acts_as_list_no_update do
-        super
-      end
-    else
-      super
-    end
+  def scheduled_for_deletion?
+    !owner || !account || owner.scheduled_for_deletion?
+  end
+
+  def act_as_list_no_update?
+    super || scheduled_for_deletion?
   end
 
   protected
--- a/test/unit/proxy_rule_test.rb
+++ b/test/unit/proxy_rule_test.rb
@@ -157,6 +157,39 @@ class ProxyRuleTest < ActiveSupport::TestCase
     assert backend_proxy_rule.valid?
   end
 
+  test 'position is not updated when account has been deleted' do
+    provider = FactoryBot.create(:simple_provider)
+    proxy = FactoryBot.create(:service, account: provider).proxy
+    proxy_rule = FactoryBot.create(:proxy_rule, proxy: proxy)
+
+    provider.delete
+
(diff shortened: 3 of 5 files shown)


app/models/proxy_rule.rb
```diff
@@ -124,6 +124,18 @@ def backend_api_owner?
...
+
+  def destroy
+    if !owner || !account || owner.scheduled_for_deletion?
+      self.class.acts_as_list_no_update do
+        super
+      end
+    else
+      super
+    end
+  end
```
guicassolato: A couple alternatives:

1.
```ruby
  def destroy
    return super if owner&.persisted? && account&.persisted? && !owner.scheduled_for_deletion?

    acts_as_list_no_update do
      super
    end
  end
```

2.
```ruby
  def scheduled_for_deletion?
    !owner || !account || owner.scheduled_for_deletion?
  end

  def act_as_list_no_update?
    super || scheduled_for_deletion?
  end
```

test/unit/proxy_rule_test.rb
```diff
@@ -157,6 +157,46 @@ class ProxyRuleTest < ActiveSupport::TestCase
...
+
+    attr_reader :provider
+
+    test 'position is not updated when account has been deleted' do
+      proxy = FactoryBot.create(:service, account: provider).proxy
+      proxy_rule = FactoryBot.create(:proxy_rule, proxy: proxy)
+
+      provider.delete
+
+      proxy_rule.expects(:decrement_positions_on_lower_items).never
```
guicassolato: This is the downside of this approach using `#act_as_list_no_update?`.

app/events/proxy_configs/affecting_object_changed_event.rb
```diff
@@ -8,8 +8,12 @@ def self.create(proxy, object)
       object_type: object.class.name,
       metadata: {
         service_id: proxy.service_id,
-        provider_id: proxy.account.id
+        provider_id: proxy.account&.id
```
hallelujah: Won’t the event be invalid in this case?

hallelujah:
Is the update position inside a callback?
If so it can be skipped

app/models/proxy_rule.rb
```diff
@@ -124,6 +124,16 @@ def backend_api_owner?
     owner_type == 'BackendApi'
   end
 
+  delegate :account, to: :owner, allow_nil: true
+
+  def scheduled_for_deletion?
+    !owner || !account || owner.scheduled_for_deletion?
+  end
+
+  def act_as_list_no_update?
```
hallelujah: OK this is the callback condition

test/unit/proxy_rule_test.rb
hallelujah: I think it is fine
The proxy_rule is going to be deleted anyway no?
Unless you remove the `!account` In scheduled_for_position

app/events/proxy_configs/affecting_object_changed_event.rb
guicassolato: Second time this happens to me in less than a month. First add the nil type check, then define the `.valid?` class method and forget to remove the check. Sorry :/

app/events/proxy_configs/affecting_object_changed_event.rb
```diff
@@ -12,4 +12,8 @@ def self.create(proxy, object)
       }
     )
   end
+
+  def self.valid?(proxy, *_args)
+    proxy && proxy.account&.persisted? # Proxy's service or account may have been deleted
```
hallelujah: Good for me :)

It is more explicit that this event can't be valid if account is not there anymore

Though it reeks code smell by the use of `account&.` The reason is simple, do we generate config change on deletion?

guicassolato: We do issue Affecting Object Changes events on deletion, yes.

hallelujah: Ouch :(
Can't it be fixed when it is scheduled for deletion or maybe?
For the issue you address the PR is good enough but it looks like a can of worm if we want to fix everything.
Good for you to address (if needed) the above issue in another PR?

guicassolato: @hallelujah, thinking about moving the check here to https://github.com/3scale/porta/blob/1e8bdf5d3ab494f09d2db7e471f690860f32b98f/app/lib/proxy_config_affecting_changes.rb#L13?

guicassolato: https://issues.redhat.com/browse/THREESCALE-4322

PR #1557, opened 2020-01-23
Get rid of deprecation warning about :text format in render

hallelujah:
Remove deprecation of render text: 'Something'
Uses render plain: 'Something' instead

### DIFF
--- a/app/controllers/provider/admin/webhooks_controller.rb
+++ b/app/controllers/provider/admin/webhooks_controller.rb
@@ -24,7 +24,7 @@ class Provider::Admin::WebhooksController < Sites::BaseController
         if @webhook
           @ping_response = @webhook.ping
         else
-          render :status => :forbidden, :text => 'Nowhere to ping'
+          render :status => :forbidden, :plain => 'Nowhere to ping'
         end
       end
     end
--- a/lib/developer_portal/app/controllers/developer_portal/cms/new_content_controller.rb
+++ b/lib/developer_portal/app/controllers/developer_portal/cms/new_content_controller.rb
@@ -54,7 +54,7 @@ class DeveloperPortal::CMS::NewContentController < DeveloperPortal::BaseControll
 
       # CMS::Page#mime_type is parsed and valid Mime::Type
       # in case it can't be parsed (nil, something invalid) returns 'text/html'
-      render :layout => false, :content_type => page.mime_type, :text => renderer.content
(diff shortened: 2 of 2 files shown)


app/controllers/provider/admin/webhooks_controller.rb
```diff
@@ -24,7 +24,7 @@ def show
         if @webhook
           @ping_response = @webhook.ping
         else
-          render :status => :forbidden, :text => 'Nowhere to ping'
+          render :status => :forbidden, :plain => 'Nowhere to ping'
```
Martouta: :eyes: 
```suggestion
          render status: :forbidden, plain: 'Nowhere to ping'
```

lib/developer_portal/app/controllers/developer_portal/cms/new_content_controller.rb
```diff
@@ -54,7 +54,7 @@ def show
 
       # CMS::Page#mime_type is parsed and valid Mime::Type
       # in case it can't be parsed (nil, something invalid) returns 'text/html'
-      render :layout => false, :content_type => page.mime_type, :text => renderer.content
+      render :layout => false, :content_type => page.mime_type, :plain => renderer.content
```
Martouta: :eyes: 
```suggestion
      render layout: false, content_type: page.mime_type, plain: renderer.content
```

lib/developer_portal/app/controllers/developer_portal/cms/new_content_controller.rb
```diff
@@ -54,7 +54,7 @@ def show
 
       # CMS::Page#mime_type is parsed and valid Mime::Type
       # in case it can't be parsed (nil, something invalid) returns 'text/html'
```
Martouta: This comment does not match [what it does](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#rendering-content-from-string) anymore.
`render :plain` returns `text/plain` and not `text/html` and `render :text` does. For that response type, the new way would be using `render :body`.

hallelujah: But actually it should not be text/html as it is only text ...

hallelujah: Oh you are right, render :body does not set the mime :)

I will change thanks!

PR #1562, opened 2020-01-24
[Rails 5] silence_stream is removed from rails

macejmic:


test/unit/backend/random_data_generator_test.rb
```diff
@@ -8,13 +8,11 @@ class Backend::RandomDataGeneratorTest < ActiveSupport::TestCase
     cinstance = provider.bought_cinstances.first
     stats = Stats::Client.new(cinstance)
 
-    silence_stream($stdout) do
```
Martouta: ok :thinking: 
https://github.com/rails/rails/blob/dd7bd8c023696657a600ee5dba16bfe5def876bf/activesupport/lib/active_support/core_ext/kernel/reporting.rb#L41-L59

PR #1575, opened 2020-01-24
[Rails 5] Use ApplicationRecord

macejmic:


hallelujah:
In fact because this was already extracted to master when we extracted stuffs from Oracle database support, which was happening at the same time of Rails 5.0 upgrade (unfinished)

https://github.com/3scale/system/pull/9123

Martouta:
@hallelujah I see I should have mentioned here the same I told @macejmic 

There are a few more but they aren't as 'obvious' to see. For example in `test/unit/deadlock_test.rb:6`

PR #1579, opened 2020-01-24
[Rails 5] Fix protected_attributes_continued's evil monkey patch

guicassolato:
Protected attributes gem is a bad citizen. They monkey patch `ActiveRecord::QueryMethods` instead of adding their own module. It doesn't play well with BabySqueel, for example.

This PR reverts the patch and make it right.

### DIFF
--- a/config/initializers/protected_attributes_hacks.rb
+++ b/config/initializers/protected_attributes_hacks.rb
@@ -1,0 +1,25 @@
+# OMG!!!! Such bad citizen!!!
+# https://github.com/westonganger/protected_attributes_continued/blob/master/lib/active_record/mass_assignment_security/relation.rb
+# They just reopened Rails module instead of using their proper module ...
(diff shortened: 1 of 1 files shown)


Martouta:
I'm not sure if I am understanding this correctly so correct me if I am wrong in this:
1. `Protected attributes gem` overrides Active Record methods in a dangerous way
2. here we are overwriting those methods once more for them to contain the code that they had originally in Rails 5

If this is correct... can we get rid of the `Protected attributes gem` ? Not necessarily now because I see that is a lot of work and I understand that this is for 2.8... but this sounds to me like a tech debt that should in a Jira to be addressed soon.

Martouta:
We spoke about this. No idea if we sure end up using strong params only instead and get rid of this gem. But either way, that is not for now. So I will just create a Jira to open that discussion :smile:

PR #1580, opened 2020-01-24
Fix Slack Webhook URL

duduribeiro:
Our Slack isn't receiving any notifications from CI. When you run the
curl command manually, it returns `no_system` error. We added a new
environment variable with the right endpoint.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -741,7 +741,7 @@ jobs:
                 ] \
               } \
               ] \
-            }" $SLACK_WEBHOOK_URL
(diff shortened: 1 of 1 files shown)


Martouta:
This explains a lot :joy:

gsaslis:
could this be related to a recent cleanup of old apps in our slack?

how long has it been since notifications have stopped?

hallelujah:
@gsaslis It never worked AFAIK https://github.com/3scale/porta/commit/546472a2babb18aef04c3355ac5e062fbc15173e

It was wrong from the start, or someone just renamed the variable in circleci settings recently

PR #1581, opened 2020-01-24
Keep production gemfile updated

duduribeiro:
We planned to have our Gemfile.prod.lock automatically updated by circleCI. 

This PR implements this pipeline so we don't need to run it manually everytime we add some library.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -278,7 +278,7 @@ commands: # reusable commands with parameters
   save-gem-cache:
     steps:
       - save_cache:
-          key: v1-bundler-gems-{{ .Environment.DB }}-{{ arch }}-{{ .Branch }}-{{ checksum "Gemfile.lock" }}-{{ checksum "Gemfile.prod.lock" }}
+          key: v1-bundler-gems-{{ .Environment.DB }}-{{ arch }}-{{ .Branch }}-{{ checksum "Gemfile.lock" }}
           paths:
             - ./vendor/bundle
             - ./.bundle/
@@ -287,7 +287,7 @@ commands: # reusable commands with parameters
     steps:
       - restore_cache:
           keys:
-            - v1-bundler-gems-{{ .Environment.DB }}-{{ arch }}-{{ .Branch }}-{{ checksum "Gemfile.lock" }}-{{ checksum "Gemfile.prod.lock" }}
+            - v1-bundler-gems-{{ .Environment.DB }}-{{ arch }}-{{ .Branch }}-{{ checksum "Gemfile.lock" }}
 
   upload-artifacts:
     steps:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -278,7 +278,7 @@ commands: # reusable commands with parameters
   save-gem-cache:
     steps:
       - save_cache:
-          key: v1-bundler-gems-{{ .Environment.DB }}-{{ arch }}-{{ .Branch }}-{{ checksum "Gemfile.lock" }}-{{ checksum "gemfiles/prod/Gemfile.lock" }}
+          key: v1-bundler-gems-{{ .Environment.DB }}-{{ arch }}-{{ .Branch }}-{{ checksum "Gemfile.lock" }}-{{ checksum "Gemfile.prod.lock" }}
           paths:
             - ./vendor/bundle
             - ./.bundle/
@@ -287,7 +287,7 @@ commands: # reusable commands with parameters
     steps:
       - restore_cache:
           keys:
-            - v1-bundler-gems-{{ .Environment.DB }}-{{ arch }}-{{ .Branch }}-{{ checksum "Gemfile.lock" }}-{{ checksum "gemfiles/prod/Gemfile.lock" }}
+            - v1-bundler-gems-{{ .Environment.DB }}-{{ arch }}-{{ .Branch }}-{{ checksum "Gemfile.lock" }}-{{ checksum "Gemfile.prod.lock" }}
 
   upload-artifacts:
     steps:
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -833,6 +833,66 @@ jobs:
...
+      - checkout-with-submodules
+      - restore-gem-cache
+      - run:
+          name: Keep production gemfile up to date
+          command: |
+            echo "keeping production gemfiles up to date"
+            git config user.email "3scale@redhat.com"
+            git config user.name "3scale"
+            unset BUNDLE_FROZEN
+            bundle install --no-deployment
```
hallelujah: I am not sure we need to install the gems. `bundle lock` should use the metadata from the gem spec.

.circleci/config.yml
```diff
@@ -833,6 +833,66 @@ jobs:
...
+            make bundle
+            FILES_CHANGED=$(git status --porcelain | sed s/^...//)
```
hallelujah: I'd add `v1` for the porcelain version and awk to get the second field, relying on the number of character is not sure

```suggestion
            FILES_CHANGED=$(git status --porcelain=v1 | awk '{print $2}')
```

.circleci/config.yml
```diff
@@ -833,6 +833,66 @@ jobs:
...
+            [ -z "$FILES_CHANGED" ] && exit 0
```
hallelujah: Use double brackets to protect your shell, though here it right to use single one


```suggestion
            [[ -z "$FILES_CHANGED" ]] && exit 0
```

.circleci/config.yml
duduribeiro: without this it failed because gems was installed with --deployment option

hallelujah:
Also you will need to revert this https://github.com/3scale/deploy/pull/531
And next SaaS deployment should be done server per server (unicorn reloading issue)

### Unicorn

- stop traffic for this server
- stop unicorn
- deploy only on this server
- start traffic again

### sidekiq

- stop all sidekiq
- deploy

duduribeiro:
@hallelujah I remember that in one of my tries with File.realpath I could get work (only receiving the error of the frozen that we discovered after), I will try to get in this state with the gemfiles folder so we don't need to rever it into the deploy one. wdyt?

.circleci/config.yml
hallelujah: Just use `BUNDLE_DEPLOYMENT` and `BUNDLE_FROZEN` to `"false"` ?
See https://github.com/rubygems/bundler/blob/master/spec/install/deploy_spec.rb#L297-L310

.circleci/config.yml
duduribeiro: the version of git used in the image don't allow to set porcelain version

.circleci/config.yml
```diff
@@ -831,6 +831,64 @@ jobs:
...
+                  } \
+                  ] \
+                } \
+                ] \
+              }" $SLACK_WEBHOOK_SYSTEM
+              exit 1
+            fi
+          environment:
+            BUNDLE_RETRY: 3
+            BUNDLE_JOBS: 3
```
hallelujah: Why this env?
Can we have it based on the processors number?

.circleci/config.yml
```diff
@@ -831,6 +831,64 @@ jobs:
...
+              git commit -m "Update Gemfile.prod.lock"
+              git push origin $(git rev-parse --abbrev-ref HEAD)
+            else
+              echo "More than Gemfile.prod.lock file was changed"
+              curl -X POST -H 'Content-type: application/json' \
+              --data \
+              "{ \
+                \"attachments\": [ \
+                { \
+                  \"fallback\": \"Build failed on $CIRCLE_BRANCH, for $CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME.\", \
```
hallelujah: CIRCLE_CI_BRANCH is always master right?

Gemfile.prod.lock
```diff
@@ -36,7 +36,7 @@ GIT
 
 GIT
   remote: https://github.com/3scale/prawn.git
-  revision: 88aead0d97e230d08e9f51a18711d01317965bfe
+  revision: 6b18589fab63fabef420c0ccb546af431ad46404
```
hallelujah: Are yous sure this change belongs to this commit?

duduribeiro: it is not 😄 will drop it

.circleci/config.yml
duduribeiro: yeah!

Martouta:
There are conflicts with master so you will need to rebase.

However, there is an overlap with https://github.com/3scale/porta/pull/1665 so @duduribeiro and @hallelujah should consider this. Imo we should wait for the Rails 5 PR but I could be wrong

.circleci/config.yml
```diff
@@ -278,7 +278,7 @@ commands: # reusable commands with parameters
   save-gem-cache:
     steps:
       - save_cache:
-          key: v1-bundler-gems-{{ .Environment.DB }}-{{ arch }}-{{ .Branch }}-{{ checksum "Gemfile.lock" }}-{{ checksum "gemfiles/prod/Gemfile.lock" }}
+          key: v1-bundler-gems-{{ .Environment.DB }}-{{ arch }}-{{ .Branch }}-{{ checksum "Gemfile.lock" }}-{{ checksum "Gemfile.prod.lock" }}
```
hallelujah: Oh I think there is no need to checksum the Gemfile.prod.lock anymore. We only run tests against the main Gemfile.lock
```suggestion
          key: v1-bundler-gems-{{ .Environment.DB }}-{{ arch }}-{{ .Branch }}-{{ checksum "Gemfile.lock" }}
```

.circleci/config.yml
```diff
@@ -287,7 +287,7 @@ commands: # reusable commands with parameters
     steps:
       - restore_cache:
           keys:
-            - v1-bundler-gems-{{ .Environment.DB }}-{{ arch }}-{{ .Branch }}-{{ checksum "Gemfile.lock" }}-{{ checksum "gemfiles/prod/Gemfile.lock" }}
+            - v1-bundler-gems-{{ .Environment.DB }}-{{ arch }}-{{ .Branch }}-{{ checksum "Gemfile.lock" }}-{{ checksum "Gemfile.prod.lock" }}
```
hallelujah: Same
```suggestion
            - v1-bundler-gems-{{ .Environment.DB }}-{{ arch }}-{{ .Branch }}-{{ checksum "Gemfile.lock" }}
```

PR #1583, opened 2020-01-24
[Rails 5] Add support for empty Arrays in params

Martouta:
See https://github.com/rack-test/rack-test/commit/ece681de8ffee9d0caff30e9b93f882cc58f14cb

Without the patch, empty arrays will be removed implicitly.
Adding this special case to `build_nested_query` fixes this issue.

### Explanation of the 2nd commit
Explained in https://github.com/3scale/porta/pull/1583#discussion_r371384105

In Rails 4, `put :update, id: @user.id, user: { member_permission_ids: [] }` and `put provider_admin_account_user_path(@user.id), params: {user: { member_permission_ids: [] }}` do not do the same unfortunately. So i've changed the test to `put provider_admin_account_user_path(@user.id), params: {user: { member_permission_ids: [''] }}`.
Because it works and it is because it is what the view actually sends.
The rack test fix can be there or not there and anyway the result does not change in that.


test/functional/provider/admin/account/users_controller_test.rb
```diff
@@ -13,7 +13,7 @@ def setup
   def test_update_blank_member_permission_ids
     assert_equal @default_ids, @user.admin_sections.to_a
 
-    put provider_admin_account_user_path(@user.id), params: {user: { member_permission_ids: [] }}
+    put provider_admin_account_user_path(@user.id), params: {user: { member_permission_ids: [''] }}
```
Martouta: `['']` is the param that is sent from the UI for `member_permission_ids` when it is empty.
image

In the original code, the one currently in master, before this PR, this is what we are testing (functional test):
https://github.com/3scale/porta/blob/837cbb815afda6f881c0fd497701725a96502973/test/functional/provider/admin/account/users_controller_test.rb#L16
Which sends this as params:
image

However, when done as an Integration test, the line `put provider_admin_account_user_path(@user.id), params: {user: { member_permission_ids: [] }}` does not work. Because the param `member_permission_ids` is not sent at all if it is done this way.

However, the line `put provider_admin_account_user_path(@user.id), params: {user: { member_permission_ids: [''] }}` works well.

This is independent of having or not this file: https://github.com/3scale/porta/pull/1583/commits/6ce265a6a05cd27690ddd806effd130baffc1d2c#diff-4b36adfa3f694588ae61b0d05c5ac580R1

PR #1584, opened 2020-01-24
[Rails 5] Use of unsafe parameters in URL generation

Martouta:
Causing this error:

Attempting to generate a URL from non-sanitized request parameters! An attacker can inject malicious data into the generated URL, such as changing the host. Whitelist and sanitize passed parameters to be secure. (ActionView::Template::Error)

### DIFF
--- a/app/lib/three_scale/api/responder.rb
+++ b/app/lib/three_scale/api/responder.rb
@@ -30,6 +30,8 @@ class ThreeScale::Api::Responder < ActionController::Responder
       params.merge!(controller.request.query_parameters)
       params[:id] = resource.id
       params[:action] = :show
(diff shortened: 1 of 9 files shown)


app/lib/three_scale/api/responder.rb
```diff
@@ -30,6 +30,8 @@ def navigation_location
       params.merge!(controller.request.query_parameters)
       params[:id] = resource.id
       params[:action] = :show
+      # Permitting all parameters here but it is not really secure
+      params.permit! if params.respond_to?(:permit!)
```
Martouta:
```suggestion
      params.try(:permit!)
```

PR #1589, opened 2020-01-27
[Bugfix] TypeError: superclass mismatch for class -AnyFunctionalTest-

Martouta:
I wanted to test an issue from the rails 5 branch and I encountered an unrelated error:
image

The problem was that this class was defined twice but with a different superclass so it is not monkey-patching it but defining it again.
image

The problem is that one is the _functional_ and the other one is the _integration_. Functional tests are deprecated so I moved it to integration and I also slightly refactored the test.

This changes works for any Rails version :smile:


test/integration/api/alerts_controller_test.rb
```diff
@@ -19,4 +19,40 @@ class Api::AlertsControllerTest < ActionDispatch::IntegrationTest
...
     end
   end
+
+  test 'get index should filter by account_id, cinstance_id and fulltext' do
+    buyer1, buyer2 = FactoryBot.create_list(:simple_buyer, 2, provider_account: provider)
+    plan = FactoryBot.create(:simple_application_plan, issuer: provider.default_service)
+    cinstance1, cinstance2 = [buyer1, buyer2].map { |buyer| FactoryBot.create(:simple_cinstance, plan: plan, user_account: buyer) }
+    [cinstance1, cinstance2].each { |cinstance|FactoryBot.create(:limit_alert, account: provider, cinstance: cinstance) }
+
+
```
duduribeiro: can you get rid of this multiple spaces?

PR #1590, opened 2020-01-27
[Rails5] Fix UniqueAssociation for Rails 5.0

duduribeiro:


Martouta:
Maybe this should be `only-rails5`

duduribeiro:
yeah, it is. Already added only-rails5 in the doc 😄

PR #1591, opened 2020-01-27
[Rails5] Remove multiple include in FileFixtures

duduribeiro:
Migrated commit 3480e613


hallelujah:
This is only OK for rails 5 :)

duduribeiro:
@hallelujah yeah, I did notice it 😂 will update the doc

PR #1594, opened 2020-01-28
Support OAS3 in Admin Portal (ActiveDocs)

josemigallas:
> ⚠️ Merge only after #1596

Introduces `swagger-ui-react` in ActiveDocs in order to support OAS3 / Swagger 3:

Screen Shot 2020-01-16 at 12 40 52

Screen Shot 2020-01-28 at 10 48 05

[THREESCALE-2725: Start using SwaggerUI 3.18.3 or higher for OAS3 specs](https://issues.redhat.com/browse/THREESCALE-2725)

**Verification steps** 
1. Go to any Product's active docs section: apiconfig/services/2/api_docs/<id>/preview#/
2. Create a new spec (v3) i.e. [Swagger petstore](http://petstore.swagger.io:8080/api/v3/openapi.json)
3. It should display the new UI
4. Opening a spec (v2) should still display the old UI:
    Screen Shot 2020-01-28 at 10 57 47

Dev portal support and proper styling will be handled in a following PR (#1595).

### DIFF
--- a/app/javascript/src/ActiveDocs/components/ActiveDocsSpec.jsx
+++ b/app/javascript/src/ActiveDocs/components/ActiveDocsSpec.jsx
@@ -1,0 +1,18 @@
+// @flow
+
+import React from 'react'
+
+import { createReactWrapper } from 'utilities/createReactWrapper'
+import SwaggerUI from 'swagger-ui-react'
+
+import 'swagger-ui-react/swagger-ui.css'
+
+type ActiveDocsSpecProps = {
+  url: string
+}
+
+const ActiveDocsSpec = ({ url }: ActiveDocsSpecProps) => <SwaggerUI url={url} />
+
+const ActiveDocsSpecWrapper = (props: ActiveDocsSpecProps, id: string) => createReactWrapper(<ActiveDocsSpec {...props} />, id)
+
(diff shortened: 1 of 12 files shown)


app/javascript/src/ActiveDocs/components/ActiveDocsSpec.jsx
```diff
@@ -0,0 +1,18 @@
+// @flow
+
+import React from 'react'
+
+import { createReactWrapper } from 'utilities/createReactWrapper'
+import SwaggerUI from 'swagger-ui-react'
+
+import 'swagger-ui-react/swagger-ui.css'
```
hallelujah: If we could avoid injecting css in the HTML page that would be better
Especially we would like to use the CDN

didierofrivia: Actually, we are now aiming to stop forking and modifying third party assets and providing them through the CDN (which eventually will cease to exist) promoting having everything in our packange.json and extend it through plugins (in the case of swagger ui 3). That style bundle is provided bu the same swagger lib imported 2 lines above, so it's totally OK to provide it in our javascript component. Easy to maintain, test, tweak, track license, etc.

PR #1595, opened 2020-01-28
Support OAS3 in Dev portal

josemigallas:
> ⚠️ Merge only after #1594 

Updates `swagger-ui` to v3 in Dev portal in order to support OAS3 / Swagger 3. Also, `swagger-ui` is no longer added from the cdn but it is included as a webpack asset (node_modules).

Old versions of swagger will still be available for those customers who does not manually update.

Vanilla:
Screen Shot 2020-01-28 at 11 19 32

With Scss patch (final verrsion):
Screen Shot 2020-01-31 at 17 34 30

[THREESCALE-2725: Start using SwaggerUI 3.18.3 or higher for OAS3 specs](https://issues.redhat.com/browse/THREESCALE-2725)

**Verification steps** 
1. Go to any Product's active docs section: apiconfig/services/2/api_docs/<id>/preview#/
2. Create a new spec (v3) i.e. [Swagger petstore](http://petstore.swagger.io:8080/api/v3/openapi.json)
3. Update the following templates at Dev Portal's Content section: /p/admin/cms/templates
    - Documentation
    ```liquid
   <h1>Documentation</h1>

   <p>Use our live documentation to learn about the Echo API</p>

    {% content_for javascripts %}
        {{ 'active_docs.js' | javascript_include_tag }}
    {% endcontent_for %}

    {% include 'shared/swagger_ui' %}
    ```
    - shared/swagger_ui
    ```rb
    -    SwaggerUIBundle({
    +    SwaggerUI({
           url: url,
    -      dom_id: 'swagger-ui-container',
    +      dom_id: '#swagger-ui-container',
           supportedSubmitMethods: ['get', 'post', 'put', 'delete', 'patch'],
           apisSorter: 'alpha', // can also be a function
           operationsSorter: 'method', // can also be 'alpha' or a function
    -      onComplete: function(swaggerApi, swaggerUi) {
    -        $(document.getElementById(swaggerUi.dom_id)).find('pre code').each(function(i, e {hljs.highlightBlock(e)})
    -      },
           docExpansion: 'list',
           transport: function (httpClient, obj) { ApiDocsProxy.execute(httpClient, obj) }
         });
    ```
4. At the dev portal > Documentation: it should correctly display the new (ugly) ui

Admin portal (active docs) support implemented in #1594 
Proper styling will be handled in a following PR (#??).


app/lib/logic/provider_signup.rb
```diff
@@ -256,7 +303,7 @@ def create_default_account_plan!
       end
 
       def create_default_service_plan!
-        plan = service_plans.find_by_name('Default') || service_plans.create!(name: 'Default')
+        plan = service_plans.find_by(name: 'Default') || service_plans.create!(name: 'Default')
```
didierofrivia: isn't `find_by_name` quicker?

features/old/liquid/active_docs_v1.feature
```diff
@@ -0,0 +1,27 @@
...
+      $(function () {
+        ThreeScale.APIDocs.init([]);
+      });
+    </script>
+    """
+    And the cms page "/version-10" has main layout
+
+    And the current domain is "foo.example.com"
+    When I visit "/version-10"
+    Then I should see "Echo"
```
didierofrivia: Maybe asserting the version tag rendered?

lib/developer_portal/app/views/developer_portal/shared/_swagger_ui.html.liquid
```diff
@@ -11,15 +11,18 @@
   (function () {
     var url = "{{provider.api_specs.first.url}}";
 
-    SwaggerUI({
+    window.swaggerUi = new SwaggerUi({
```
didierofrivia: why changing this back to version 2?

app/javascript/packs/active_docs.js
```diff
@@ -0,0 +1,5 @@
+// We can define the 3scale plugins here and export the modified bundle
+import SwaggerUI from 'swagger-ui'
+import 'swagger-ui/dist/swagger-ui.css'
```
hallelujah: Do we really want to inject css to HTML?
Can't we use external script so we can have CDN on it?

features/step_definitions/active_docs_steps.rb
```diff
@@ -41,6 +41,28 @@
   assert_equal 1, evaluate_script("$('.apidocs-param-tips.#{autocomplete}:visible').length")
 end
 
+Then(/^the swagger v3 autocomplete should work for "(.*?)" with "(.*?)"$/) do |input_name, autocomplete|
+  id = 'default' # Could be passed as arg
+  section_id = "#operations-tag-#{id}"
+
+  closed_section = find("#{section_id}[data-is-open='false']")
+  closed_section&.click
```
hallelujah: Why do we need this nil check? That smells like the test is not reliable

features/step_definitions/active_docs_steps.rb
```diff
@@ -41,6 +41,28 @@
...
+
+  within section_id do
+    method_id = "operations-#{id}-get_"
+    closed_method = find(method_id)
+    closed_method&.click
```
hallelujah: https://github.com/3scale/porta/pull/1595/files#r373499401

app/javascript/packs/active_docs.js
didierofrivia: it's not injected in the HTML, it's bundled inside a JS pack. As I commented before, we are aiming to stop providing modifying 3rd party assets from our CDN

app/lib/logic/provider_signup.rb
hallelujah: No and it is deprecated so better use `find_by`

features/old/liquid/active_docs.feature
```diff
@@ -53,7 +53,25 @@ Feature: ActiveDocs
 
     <h3>ActiveDocs version 2.2.10</h3>
 
-    {% include 'shared/swagger_ui' %}
+    <div id="swagger-ui-container" class="swagger-ui-wrap"></div>
```
hallelujah: So we do not have it inside the `'shared/swagger_ui'` anymore?
Do we do this because of liquid template migration?

features/old/liquid/active_docs.feature
```diff
@@ -53,7 +53,25 @@ Feature: ActiveDocs
...
+          url: url,
+          dom_id: 'swagger-ui-container',
+          supportedSubmitMethods: ['get', 'post', 'put', 'delete', 'patch'],
+          apisSorter: 'alpha', // can also be a function
+          operationsSorter: 'method', // can also be 'alpha' or a function
+          onComplete: function(swaggerApi, swaggerUi) {
+            $(document.getElementById(swaggerUi.dom_id)).find('pre code').each(function(i, e) {hljs.highlightBlock(e)})
+          },
+          docExpansion: 'list',
+          transport: function (httpClient, obj) { ApiDocsProxy.execute(httpClient, obj) }
```
hallelujah: This one is really equal to the `shared/swagger_ui` https://github.com/3scale/porta/blob/a903259ffabbf928e7694c94f9463c3737e813f2/lib/developer_portal/app/views/developer_portal/shared/_swagger_ui.html.liquid

app/javascript/packs/active_docs.js
hallelujah: Ah if it is precompiled then we can use a CDN then

didierofrivia: The idea is to eliminate the future use of the CDN, stop providing third party modified assets.

hallelujah: What are you talking about? Cloudfrount is a CDN as Akamai
If you provide compiled assets, then you can distribute them through a CDN
I am not talking about `cdn_assets` but about: injecting assets VS hosting them

injected assets can't be used by a CDN unlike compiled ones

didierofrivia: I meant **our** CDN.

didierofrivia: those packs are treated the same as the rest of Rails assets.

hallelujah: Yes so out of topic :)
Closing

features/old/liquid/active_docs_v1.feature
josemigallas: This one has been extracted from `active_docs.feature`, I'm not sure if I should change it, specially since v1 is (or will be) deprecated.

lib/developer_portal/app/views/developer_portal/shared/_swagger_ui.html.liquid
josemigallas: TLDR: cecause it is not being used in v3.
Some cukes have been failing since `swagger_ui` is still being used for v2. For v3 we updated the parent document directly and it is not using `_swagger_ui.html.liquid` anymore.

features/step_definitions/active_docs_steps.rb
josemigallas: When loading the Spec the methods are collapsed so that they have to be clicked. This is just in case they are loaded expanded already (in which case `closed_section` is `nil`). It's just a way to do `if closed then click it`. Would it be better this way?

features/old/liquid/active_docs.feature
josemigallas: I think so. This way it is easier to upgrade and no need to touch `shared/swagger_ui`.

features/old/liquid/active_docs.feature
josemigallas: `swagger_ui` was modified at the beginning but I reverted it in the last commit, this is wrong and I'm changing it, thanks.

features/step_definitions/active_docs_steps.rb
hallelujah: yes you can pass an options to `find` `visible: :all`

See https://rubydoc.info/github/jnicklas/capybara/master/Capybara/Node/Finders#find-instance_method

didierofrivia:
Maybe fixing the css issues to make codeclimate happy?

features/step_definitions/active_docs_steps.rb
josemigallas: Now that would make the test unreliable.. And we have had false positives in the past just because of that.

josemigallas: I can remove the nil check and assume the section is collapse but I sincerely don't see any problem with checking if its closed first, in order to bypass a possible error.

hallelujah: That's different, I did not mean to make the click but more:

- find the element is there (hidden or visible)
- if visible, click on it / because we want to close it, or the other way around right?
- if hidden, do nothing because it is what we want

Then do your stuff

It is more explicit what to find and more reliable in term of "elements are there, visible or not"

features/step_definitions/active_docs_steps.rb
```diff
@@ -41,6 +41,28 @@
...
+
+    within method_id do
+      input = find("[data-param-name='#{input_name}'] input")
+      click_on 'Try it out'
+      # TODO: do the actual assertion
```
hallelujah: @josemigallas Is this done ?

josemigallas: No, this particular test case is marked as @wip.

features/step_definitions/api_docs_step.rb
```diff
@@ -61,6 +68,35 @@
   end
 end
 
+Then(/^swagger v3 should escape properly the curl string$/) do
+  id = 'default' # Could be passed as arg
+  section_id = "#operations-tag-#{id}"
+
+  closed_section = find("#{section_id}[data-is-open='false']")
+  closed_section&.click
```
hallelujah: Same as https://github.com/3scale/porta/pull/1595/files#r373572166

features/old/providers/active_docs_v3.feature
```diff
@@ -0,0 +1,191 @@
...
+        }
+      }
+    }
+    """
+    And I press "Create Service"
+    Then should see "ActiveDocs Spec was successfully saved."
+    And the swagger v3 autocomplete should work for "user_key" with "user_keys"
+
+  # TODO: not sure what to assert
+  @wip
```
hallelujah: I would avoid committing `@wip` in a PR ready to be merged, please extract to another PR

features/step_definitions/active_docs_steps.rb
josemigallas: It should always be collapsed at first (invisible) so it is always required to click on it. But just in case it someday reders expanded, do not click on it. It makes the cuke more robust IMHO. This doesn't even have to do anything with the assertion, it's just a mechanism to prepare the view in order to test it.

hallelujah: Well when I read: `if` and `someday` that smells like not robust at all ... Uncertainty is worse than certainty
If it breaks in the future, let it break and fix it because we will know why and not just rely on supposition

PR #1601, opened 2020-01-28
[Rails 5] Remove skip_before_action :activate_menu_site_or_settings

Martouta:
The name of the commit is not related to what it currently does. Explanation in https://github.com/3scale/porta/pull/1601#discussion_r373514905


app/controllers/sites/documentations_controller.rb
```diff
@@ -1,6 +1,5 @@
 class Sites::DocumentationsController < Sites::BaseController
   # see ForumsController
-  skip_before_action :activate_menu_site_or_settings
```
hallelujah: @Martouta this method does not exist so rails 5.x will just throw an error

Maybe the content was there but due to many rebases on master it disappeared because already included in master

PR #1525, opened 2020-01-13
Pf4 dark mode

damianpm:
**What this PR does / why we need it**:

PatternFly is switching by default to dark theme.
See:
https://www.patternfly.org/v4/documentation/react/components/page#examples
And:
https://www.patternfly.org/v4/documentation/react/components/nav

**Verification steps** 

`npm run update-dependencies
`

**Special notes for your reviewer**:

**Before:**
before

**After:**
after

Deployed to [preview env](https://multitenant-admin.preview01.3scale.net/p/admin/dashboard)

PR #1605, opened 2020-01-29
Showing exclamation mark when outdated config

damianpm:

We switched to PF4 dark mode in https://github.com/3scale/porta/pull/1525
Now the exclamation mark icon for outdated configuration is missing

https://issues.redhat.com/browse/THREESCALE-4353

outdated

### DIFF
--- a/app/assets/stylesheets/provider/_vertical_nav.scss
+++ b/app/assets/stylesheets/provider/_vertical_nav.scss
@@ -1,10 +1,18 @@
 .pf-c-page__sidebar {
-  color: $white;
   text-align: left;
-  text-decoration: none;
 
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/provider/_vertical_nav.scss
```diff
@@ -1,10 +1,18 @@
 .pf-c-page__sidebar {
-  color: $white;
   text-align: left;
-  text-decoration: none;
```
thomasmaas: why removed?

damianpm: Is not used, never applied

PR #1606, opened 2020-01-29
[Makefile] Produce reproducible docker-compose

hallelujah:
- Adds volume to keep assets and node_modules
- Remove RAILS_ENV=development to allow all db creation
- Remove unused targets

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,11 +34,11 @@ ORACLE_DB_IMAGE := quay.io/3scale/oracle:12.2.0.1-ee
 include wget.mk
 include openshift.mk
 
-.PHONY: assets bash build bundle clean dev-setup dev-start dev-stop help oracle-database oracle-db-setup run schema yarn
+.PHONY: assets bash build bundle clean database dev-setup dev-start dev-stop help oracle-database oracle-db-setup run schema yarn
 .DEFAULT_GOAL := help
 
 # From here on, only phony targets to manage docker compose
-all: dev-setup dev-start
+all: clean dev-start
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -27,104 +27,71 @@ RUBY_ENV += RUBY_GC_OLDMALLOC_LIMIT_GROWTH_FACTOR=1.2
...
-tmp-export: clean-tmp
-	-@ $(foreach dir,$(TMP),docker cp $(IMAGE):/opt/system/$(dir) $(dir) 2>/dev/null;)
-
-clean-tmp: ## Removes temporary files
-	-@ $(foreach dir,$(TMP),rm -rf $(dir);)
+assets: ## Create assets volumes
+assets: yarn
+assets: CMD = rake assets:precompile
+yarn: ## Install JS dependencies
+yarn: CMD = yarn install
```
didierofrivia: I know Yarn is the standard of Rails 5 (I think from Rails 5.1), but this won't work in our current setup. We need to migrate back to Yarn our Webpack entire suite.

hallelujah: Why won't it work?

kevprice83:
@hallelujah this worked for me. I only had an issue with installing all the yarn packages due to network being slow...not sure that was really the case though so I set `--network-timeout=100000` and it was fixed for me, maybe many packages and some quite large? If nobody else has this issue then this can be ignored.

hallelujah:
@kevprice83 I did not have any issue but only on migration where mysql was not completely up so it fails
Re-running dev-start again worked

PR #1607, opened 2020-01-29
[PageTest] fix always_has_liquid_enabled test

macejmic:


Martouta:
We probably should remove the line 39 now :thinking: the `page.save!`

PR #1608, opened 2020-01-29
[Rails5] Fix initializers for rails5

duduribeiro:
Extracted:
60437527
19d3251d
f58e01df

### DIFF
--- a/app/lib/three_scale/filter_arguments.rb
+++ b/app/lib/three_scale/filter_arguments.rb
@@ -14,9 +14,18 @@ module ThreeScale
     end
 
     def filter
-      return FILTER.filter(@arguments) if @arguments.is_a?(Hash)
+      return FILTER.filter(@arguments) if parameters_arguments?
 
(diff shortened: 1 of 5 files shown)


app/lib/three_scale/filter_arguments.rb
```diff
@@ -14,9 +14,18 @@ def initialize(arguments = [])
...
 
     def filter
-      return FILTER.filter(@arguments) if @arguments.is_a?(Hash)
+      return FILTER.filter(@arguments) if parameters_arguments?
 
       @arguments.map { |argument| argument.is_a?(Enumerable) ? FILTER.filter(argument) : argument }
     end
+
+    def parameters_arguments?
+      case @arguments
```
guicassolato: 1-liner alternative:

```ruby
(@arguments.class.ancestors & [ActionController::Parameters, Hash]).present?
```

hallelujah: 👍 nice one but might be complicated to read

PR #1609, opened 2020-01-29
[Rails 5] Refactor Test ExtraFieldsTest

Martouta:
The main goal is to remove the shoulda syntax to be able to run any test in isolation :smile: but the PR also improves the test in other ways :smile:

### DIFF
--- a/test/unit/extra_fields_test.rb
+++ b/test/unit/extra_fields_test.rb
@@ -1,6 +1,5 @@
 require 'test_helper'
 
-
 class ExtraFieldsTest < ActiveSupport::TestCase
 
   setup do
@@ -8,10 +7,10 @@ class ExtraFieldsTest < ActiveSupport::TestCase
     @buyer = FactoryBot.create :buyer_account, :provider_account => @provider
 
     FactoryBot.create(:fields_definition, :account => @provider, :target => "Account",
-            :name => "provider_extra_field", :required => true)
+                       :name => "provider_extra_field", :required => true)
     FactoryBot.create(:fields_definition, :account => Account.master,
-            :target => "Account",
-            :name => "master_extra_field", :required => true)
+                       :target => "Account",
+                       :name => "master_extra_field", :required => true)
 
     @provider.reload
     @buyer.reload
@@ -32,10 +31,10 @@ class ExtraFieldsTest < ActiveSupport::TestCase
 
   test 'to_xml renders extra_fields' do
     FactoryBot.create(:fields_definition,
-            account: @provider,
-            target: "Account",
-            name: "stuff",
-            choices: ["Orange", "Apple", "Banana"])
+                       account: @provider,
+                       target: "Account",
+                       name: "stuff",
+                       choices: ["Orange", "Apple", "Banana"])
     @buyer.reload
     @buyer.extra_fields = { "stuff"=> "Apple" }
     doc = Nokogiri::XML.parse(@buyer.to_xml)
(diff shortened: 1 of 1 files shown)


test/unit/extra_fields_test.rb
```diff
@@ -47,233 +42,213 @@ class ExtraFieldsTest < ActiveSupport::TestCase
...
-      end
-    end # setters
-
-  end # extra fields
-
-  context 'fields definitions' do
-    #TODO: do this test in user and cinstance also or find a way of doing
-    # shared_examples
+  class ExtrafieldsValidations < ActiveSupport::TestCase
+    setup do
```
guicassolato: Perhaps more sets of tests could be wrapped in subclasses of `ExtraFieldsTest` and leave `#setup` to be defined only once at the parent class?

test/unit/extra_fields_test.rb
```diff
@@ -47,233 +42,213 @@ class ExtraFieldsTest < ActiveSupport::TestCase
...
+    test 'Validations BuyerResource resource be done if said so' do
+      @buyer.validate_fields!
+
+      assert false == @buyer.valid?
+      assert @buyer.errors[@provider_field].present?
+    end
+
+  end
+
+  class Validations < FieldsDefinitionTest
```
guicassolato: Do all these 3 classes need to inherit from `FieldsDefinitionTest`?

hallelujah:
Well the shoulda syntax is nice but if we could use rspec instead of test-unit.
Other implementation are buggy to me

test/unit/extra_fields_test.rb
Martouta: @guicassolato Yeah, I do not like it either so I am refactoring it step by step in different commits :smile:

test/unit/extra_fields_test.rb
Martouta: yes, that is the goal @guicassolato https://github.com/3scale/porta/pull/1609#discussion_r374360551
Only not done yet but it is the plan :smile:

Martouta:
@hallelujah I only cherry picked your commit and I am improving it in other commits, but I do not think that I will have time for much more for ER1...

test/unit/extra_fields_test.rb
```diff
@@ -1,279 +1,230 @@
...
-    @provider = FactoryBot.create :provider_account
-    @buyer = FactoryBot.create :buyer_account, :provider_account => @provider
-
-    FactoryBot.create(:fields_definition, :account => @provider, :target => "Account",
-            :name => "provider_extra_field", :required => true)
-    FactoryBot.create(:fields_definition, :account => Account.master,
-            :target => "Account",
-            :name => "master_extra_field", :required => true)
+  def setup
+    @provider = FactoryBot.create(:provider_account, provider_account: master_account)
```
hallelujah: Why do we need to set the accound for provider_account?

test/unit/extra_fields_test.rb
```diff
@@ -1,279 +1,230 @@
...
-    # extra fields are validated on demand only
-    @buyer.validate_fields!
-    User.current = @buyer.admins.first
+    def setup
+      super
 
-    refute @buyer.valid?
-    assert_not_empty @buyer.errors["provider_extra_field"]
-  end
+      @provider_field = Account.optional_fields.first
```
hallelujah: That are some assumptions for first and last

test/unit/extra_fields_test.rb
Martouta: You are right, we don't need it, so I will remove that :smile:

test/unit/extra_fields_test.rb
Martouta: @hallelujah Like before this PR, right? What do you propose instead?

hallelujah: Making it more explicit that optional_fields have multiple entries
Either by a hash or new methods?

Martouta: We spoke about this and decided that I will just make a Jira because it is too much work that does not worth it for now considering that the Rails 5 PR depends on this one :+1:

PR #4, opened 2018-09-18
[WIP] Rails 5.0 upgrade

hallelujah:
Closes #5 

Not ready yet, I want to run the tests first. Make them green
And look at the differences

PR #1613, opened 2020-01-29
[WIP][Rails 5] Add activemodel-serializers-xml as it was removed from Rails 5

Martouta:
Extracted from https://github.com/3scale/porta/pull/4/commits/def3c88adfcd7f32df915e07b02d2eabd94a639c

I dunno if this should work without Rails 5 so we'll see :woman_shrugging:


hallelujah:
I think you should mark it as only-rails-5

Martouta:
Right @hallelujah :+1:

PR #1617, opened 2020-01-30
[Rails5] - Extract some tests changes

duduribeiro:
Extracted:

48a46007
1da0e662
22c90036
d011ce98
3fd650ea
e1c69b38
abf6f9c8
eb439bcb
3f34fd4b
3895cb6d
d95affcb

### DIFF
--- a/test/workers/web_hook_worker_test.rb
+++ b/test/workers/web_hook_worker_test.rb
@@ -3,7 +3,6 @@
 require 'test_helper'
 
 class WebHookWorkerTest < ActiveSupport::TestCase
-  # disable_transactional_fixtures!
 
   setup do
     @worker = WebHookWorker.new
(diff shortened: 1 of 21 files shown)


test/unit/proxy_test.rb
```diff
@@ -629,12 +629,12 @@ def analytics
     # no existing config for staging (sandbox)
     refute proxy.pending_affecting_changes?
 
-    Timecop.travel(1.second.from_now) do
```
Martouta: This change is failing right now for Rails 4. I asked Hery and he said this change is because there was a work done to index some keys in sphinx, which was done after_commit. Rails 4 does not trigger after_commit in tests, but rails 5.0 does.

test/workers/web_hook_worker_test.rb
```diff
@@ -3,10 +3,16 @@
 require 'test_helper'
 
 class WebHookWorkerTest < ActiveSupport::TestCase
+  # disable_transactional_fixtures!
```
guicassolato:
```suggestion
```

PR #1618, opened 2020-01-30
[Rails5] Fix Line Item wrong type

duduribeiro:
Add an error if the Line Item type is wrong

### DIFF
--- a/app/lib/finance/billing.rb
+++ b/app/lib/finance/billing.rb
@@ -14,6 +14,10 @@ module Finance
       li = LineItem.new(params, {without_protection: true})
       li.errors.add(:base, :invalid_invoice_state)
       li
+    rescue ActiveRecord::SubclassNotFound
+      li = LineItem.new(params.except(:type), {without_protection: true})
+      li.errors.add(:base, :invalid_invoice_state)
+      li
     end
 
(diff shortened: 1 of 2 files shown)


app/lib/finance/billing.rb
```diff
@@ -14,6 +14,10 @@ def create_line_item(params)
       li = LineItem.new(params, {without_protection: true})
       li.errors.add(:base, :invalid_invoice_state)
       li
+    rescue ActiveRecord::SubclassNotFound
```
hallelujah: I think this is because of a migration no?

app/lib/finance/billing.rb
```diff
@@ -14,6 +14,10 @@ def create_line_item(params)
...
+      li = LineItem.new(params.except(:type), {without_protection: true})
+      li.errors.add(:base, :invalid_invoice_state)
```
guicassolato:
```suggestion
      li.errors.add(:base, : invalid_line_item_type)
```
?

hallelujah: Not really it is the same as above just removing the `type` in params

hallelujah:
I think you can fix CodeClimate and we are good to go

PR #1619, opened 2020-01-30
Prevent circleci workflows to run automatically for dependabot branches

guicassolato:
This PR prevents `mysql_build` and `javascript_tests` circleci workflows to run automatically on branches created by dependabot. One can still 'approve' such workflows and run them manually, like we normally do for Postgres and Oracle builds.

The way the solution was implemented was by duplicating the `mysql_build` and `javascript_tests` workflows in the circleci config. While one copy requires manual approval, the other has been renamed to `<workflow-name>_auto` and applies a filter to ignore branches whose name match the pattern `/dependabot\/.*/`.

Unfortunately, due to limitations for merging YAML sequences, the list of common jobs among the duplicate workflows had to be repeated.

**TODO** <====== IMPORTANT!
- [ ] Change (if possible) the settings of the repo so either `mysql_build` or `mysql_build_auto` is required to merge the PR, and the same for `javascript_tests`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -129,6 +129,11 @@ only-master-filter: &only-master-filter
     branches:
       only: master
 
+ignore-dependabot-filter: &ignore-dependabot-filter
+  filters:
+    branches:
+      ignore: /dependabot\/.*/
+
 nightly-trigger: &nightly-trigger
   triggers:
     - schedule:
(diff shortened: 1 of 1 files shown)


guicassolato:
It turns out we cannot set the checks the way we'd need to (i.e., either one or other). This is not going to work then.

PR #1620, opened 2020-01-30
[Rails5] Fix ApplicationKey::without scope

duduribeiro:
Extracted: 07bd1431

Refactor scope `without`

### DIFF
--- a/app/models/application_key.rb
+++ b/app/models/application_key.rb
@@ -30,7 +30,7 @@ class ApplicationKey < ApplicationRecord
 
   extend BackendClient::ToggleBackend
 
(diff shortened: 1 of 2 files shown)


app/models/application_key.rb
```diff
@@ -30,6 +30,8 @@ class ApplicationKey < ApplicationRecord
 
   extend BackendClient::ToggleBackend
 
+  scope :without, -> (value) { where(['value <> ?', value])}
```
hallelujah: Fixing rubocop

```suggestion
  scope :without, ->(value) { where(['value <> ?', value])}
```

PR #1622, opened 2020-01-30
AuthenticationProvider::AVAILABLE is an HashWithIndifferentAccess

duduribeiro:
Fixes AuthenticationProvider.available(:provider) #=> nil


hallelujah:
Because you can pass Symbol or String here https://github.com/3scale/porta/blob/265cbc0e7cc3a63753138faf023a37b12ce8ba24/app/models/authentication_provider.rb#L51-L53

PR #1624, opened 2020-01-31
Load countries from other sources

guicassolato:
This PR aims to solve the dependency of Porta on http://download.geonames.org/export/dump/countryInfo.txt to load the list of countries and currencies on `deploy:setup`.

Closes [THREESCALE-4373](https://issues.redhat.com/browse/THREESCALE-4373)

**TODOs**
- [x] Fallback to ~`countries:import:tzinfo`~ static file when `countries:import:geonames` fails
    -- OR --
    ~Remove `countries:import:geonames` altogether~
- [x] ~Fill currencies from different source when `countries:import:tzinfo` is used
    -- OR --
    Consider using https://github.com/hexorx/countries~
- [x] Write test for whatever we decide in the above


lib/tasks/countries.rake
```diff
@@ -1,38 +1,60 @@
+# frozen_string_literal: true
+
 require 'open-uri'
 
-namespace :countries do
+class CountriesImportService
```
guicassolato: Not supposed to be here, of course.

hallelujah:
Maybe we can try the network and if it does not work, fallback to a backup file (no need to do multiple providers)
Countries list and codes are not likely to change often, so maybe just a simple file (no need of a gem unless it will be reused for something else)

guicassolato:
> Maybe we can try the network and if it does not work, fallback to a backup file (no need to do multiple providers)

Agreed 👍

> Countries list and codes are not likely to change often, so maybe just a simple file (no need of a gem unless it will be reused for something else)

Country currencies are more volatile, but I think we can still work with a static file for now. After all, 1) TZInfo does not provide currency info anyway, and 2) a gem can get outdated just the same, so the effort to fix would be same as having an static file.

app/services/import_countries_service.rb
```diff
@@ -0,0 +1,93 @@
+# frozen_string_literal: true
+
+require 'open-uri'
+
+class ImportCountriesService
+  # Ignore these invalid/obsolete countries:
+  INVALID_COUNTRY_CODES = %w[CS XK AQ UM] # Serbia and Montenegro, Kosovo (not there yet), Antartica
+
+  class CountryData < Struct.new(:code, :name, :currency)
```
hallelujah: Avoid inheriting from Struct.new:

- Struct.new creates an anonymous class
- Then you inherit from this anonymous class
- also one struct is equal to another if attributes are equal, they are more like data containers not real object you want to work with.

```ruby
Foo = Struct.new(:bar)
foo1 = Foo.new('bar')
foo2 = Foo.new('bar')

{foo1 => :a, foo2 => :b}.length
#=> 1
```

/cc @3scale/system for your knowledge 🔝 

You might expect something else

PR #1625, opened 2020-02-03
Moving icon to submenu when expanded

damianpm:

Outdated icon in vertical nav bar should be displayed in **Integration** when menu collapsed, then move to **Configuration** when menu expanded

https://issues.redhat.com/browse/THREESCALE-4392

move-icon2

### DIFF
--- a/app/assets/stylesheets/provider/_vertical_nav.scss
+++ b/app/assets/stylesheets/provider/_vertical_nav.scss
@@ -1,3 +1,12 @@
+@mixin outdatedIcon {
+  content: '\f071';
+  font-family: FontAwesome;
+  font-weight: normal;
+  position: absolute;
+  right: line-height-times(3);
+  top: line-height-times(1 / 3);
+}
+
 .pf-c-page__sidebar {
   text-align: left;
 
@@ -7,12 +16,19 @@
   }
 
   // scss-lint:disable SelectorFormat
-  .pf-c-nav__item.outdated-config::before {
-    content: '\f071';
-    font-family: FontAwesome;
-    font-weight: normal;
-    position: absolute;
-    right: line-height-times(3);
-    top: line-height-times(1 / 3);
+  .pf-c-nav__item {
+    &.outdated-config {
+      &:not(.pf-m-expanded)::before {
+        @include outdatedIcon;
+      }
+
+      .pf-c-nav__subnav {
+        // scss-lint:disable SelectorDepth
+        .pf-c-nav__item:first-child::before {
+          @include outdatedIcon;
+          top: line-height-times(2);
+        }
+      }
+    }
   }
 }
--- a/app/assets/stylesheets/provider/_vertical_nav.scss
+++ b/app/assets/stylesheets/provider/_vertical_nav.scss
@@ -1,12 +1,3 @@
-@mixin outdatedIcon {
-  content: '\f071';
-  font-family: FontAwesome;
-  font-weight: normal;
-  position: absolute;
-  right: line-height-times(3);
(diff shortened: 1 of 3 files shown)


josemigallas:
I think it would be better to handle this in JS, firstly by reverting the changes made in #1605 and then updating the predicates with `isOutOfDate && isExpanded`.

The problem with the CSS approach is that it's completely inflexible and make it not anymore extendable (say we needed to add more out-of-date icons to other sections).

WDYT @damianpm @didierofrivia ?

thomasmaas:
I think that would be premature optimization. For me this perfect for we need now.

thomasmaas:
To clarify my last statement, we should just restore the behavior that was in place: https://github.com/3scale/porta/pull/1605/files#diff-f82d17ff82ceac4cebb8068986ecb9e3L245

thomasmaas:
This issue also needs a corresponding jira issue targeted at ER2 as QE might find regression in ER1.

damianpm:
The problem with the js approach is that, previously, we were passing a prop to a react component, because it was our own component, not a PF standard component.
Now, using PF component, it doesn't expect this behavior, not a prop or a child component, to make it work we would need to start hacking the component. The css option is the less invasive.
@josemigallas

josemigallas:
> Now, using PF component, it doesn't expect this behavior, not a prop or a child component, to make it work we would need to start hacking the component. The css option is the less invasive.

It doesn't have to be invasive, `NavItem`s can still be passed a class, then adding the icon via CSS.

didierofrivia:
IMHO, since every PF component accept classes as params, the way to go would be:

1. Computing the warning state (js)
2. Passing it as class in the component class array (js)
3. Let css to manage the visual

app/assets/stylesheets/provider/_vertical_nav.scss
```diff
@@ -7,12 +7,18 @@
...
   // scss-lint:disable SelectorFormat
-  .pf-c-nav__item.outdated-config::before {
-    content: '\f071';
-    font-family: FontAwesome;
-    font-weight: normal;
-    position: absolute;
-    right: line-height-times(3);
-    top: line-height-times(1 / 3);
+  .pf-c-nav__item {
+    &.outdated-config > a::before {
```
josemigallas: If my suggestion below is rejected I think this one might improve readability:
```suggestion
    &.outdated-config:not(.pf-m-expanded) > a::before {
```

app/javascript/src/Navigation/components/VerticalNav.jsx
```diff
@@ -39,9 +40,9 @@ const VerticalNav = ({ sections, activeSection, activeItem }: Props) => (
 const NavSection = ({title, isSectionActive, activeItem, items, outOfDateConfig}) => {
   return (
     <NavExpandable title={title} isActive={isSectionActive} isExpanded={isSectionActive} className={outOfDateConfig ? 'outdated-config' : ''}>
```
josemigallas: Only a suggestion
```suggestion
  const shouldShowIcon = outOfDateConfig && !isSectionActive
  return (
    <NavExpandable title={title} isActive={isSectionActive} isExpanded={isSectionActive} className={shouldShowIcon ? 'outdated-config' : ''}>
```

damianpm: It doesn't work, I already tried `isSectionActive`, but it only changes it's value when you click and navigate to any of the subitems, not when you expand/collapse from other pages

josemigallas: Right, it should be the state not the prop.

PR #1629, opened 2020-02-04
Flash message at top of the page not visible

HemaHG:
Flash message at top of the page not visible when the page is large

https://issues.redhat.com/browse/THREESCALE-4028

**Verification steps**

- Go to **API > Applications > Application Plans**(http://provider-admin.example.com.lvh.me:3000/apiconfig/services/2/application_plans)
-Create an application plan and observe the flash message  

image

### DIFF
--- a/app/assets/stylesheets/provider/_flash.scss
+++ b/app/assets/stylesheets/provider/_flash.scss
@@ -7,7 +7,7 @@
   padding: line-height-times(1 / 2);
   position: fixed;
   right: line-height-times(1);
-  z-index: 1000;
+  z-index: 2000;
 }
 
 .flash-message {
--- a/app/assets/stylesheets/provider/_flash.scss
+++ b/app/assets/stylesheets/provider/_flash.scss
@@ -1,8 +1,13 @@
 #flashWrapper {
-  position: absolute;
+  @include animation(fadeInRight 0.5s ease-in-out);
+  background-color: $white;
+  box-shadow: 0 0 line-height-times(1) 0 $whiteBoxShadowColor;
+  display: block;
+  margin-top: line-height-times(1);
+  padding: line-height-times(1 / 2);
+  position: fixed;
   right: line-height-times(1);
-  text-align: right;
-  @include animation(fadeInLeft 0.5s ease-in-out);
+  z-index: 1000;
 }
 
 .flash-message {
@@ -37,11 +42,11 @@
 }
 
 
-// fade in from left animation
-@include keyframes(fadeInLeft) {
+// fade in from right animation
+@include keyframes(fadeInRight) {
   0% {
     @include opacity(0);
-    @include transform(translate3d(-100%, 0, 0));
+    @include transform(translate3d(100%, 0, 0));
   }
 
   100% {
--- a/app/views/shared/provider/_flash.html.erb
+++ b/app/views/shared/provider/_flash.html.erb
@@ -1,7 +1,7 @@
 <div id="flashWrapper" style="<%= flash.empty? ? "display:none;" : "" %>">
   <div class="ajaxNoticeValid">
     <% flash.each do |type, value| %>
-      <p class="flash-message flash-message--<%= type %>"><%= value %></p>
+      <span class="flash-message flash-message--<%= type %>"><%= value %></span>
       <% flash.discard(type) %>
     <%- end -%>
     <script>
--- a/app/assets/javascripts/flash.js
+++ b/app/assets/javascripts/flash.js
@@ -9,15 +9,15 @@
 
       $.flash.current = message
 
-      var p = $('<p>')
+      var span = $('<span>')
       if (opts && opts.text) {
-        p.text(message)
+        span.text(message)
       } else {
-        p.html(message)
+        span.html(message)
       }
 
       $('.ajaxNoticeValid:first')
-      .html(p.wrap('<div/>').parent().html())
+      .html(span.wrap('<div/>').parent().html())
       .removeClass('flash-message--error')
       .removeClass('flash-message')
       .addClass(clazz)


app/views/shared/provider/_flash.html.erb
```diff
@@ -1,13 +1,14 @@
-<div id="flashWrapper" style="<%= flash.empty? ? "display:none;" : "" %>">
+<div id="flashWrapper" style="position: fixed; margin-top: 2rem">
```
thomasmaas: but should still be display none when flash is empty, no?

damianpm: @thomasmaas is a WIP, I asked @HemaHG to make a PR to be able to test

app/views/shared/provider/_flash.html.erb
```diff
@@ -1,13 +1,14 @@
...
   <div class="ajaxNoticeValid">
     <% flash.each do |type, value| %>
       <p class="flash-message flash-message--<%= type %>"><%= value %></p>
       <% flash.discard(type) %>
     <%- end -%>
-    <script>
+    <!--script>
```
thomasmaas: guess you forgot to uncomment

app/views/shared/provider/_flash.html.erb
Martouta: I agree with @thomasmaas 

```suggestion
<div id="flashWrapper" style="<%= flash.empty? ? "display:none;" : "position: fixed; margin-top: 2rem" %>">
```

damianpm: @HemaHG instead of adding the style inline, you can add it in `app/assets/stylesheets/provider/_flash.scss`
https://github.com/3scale/porta/blob/master/app/assets/stylesheets/provider/_flash.scss

keep the ruby code: `<%= flash.empty? ? "display:none;" : "" %>` 
And put the css in **_flash.scss**

app/views/shared/provider/_flash.html.erb
damianpm: And yes, uncomment here

cntlsn:
@damianpm I would suggest to only add some styling for now.

`#flashWrapper` could be:
```
#flashWrapper {
    display: block;
    position: fixed;
    background-color: #fff;
    padding: 1rem;
    margin-top: 1.5rem;
    -webkit-box-shadow: 0px 0px 20px 0px rgba(0,0,0,0.3);
    -moz-box-shadow: 0px 0px 20px 0px rgba(0,0,0,0.3);
    box-shadow: 0px 0px 20px 0px rgba(0,0,0,0.3);
}
```

And I would make the text content live into a `span` instead of a `p` (in order to lose `p` styling and avoid additional rewrites):
```
<span class="flash-message flash-message--notice">Some text</span>
```

What do you think?

damianpm:
@cntlsn @HemaHG 
OK, but I'd keep the alignment to the right and the animation
```
#flashWrapper {
  @include animation(fadeInLeft 0.5s ease-in-out);
  background-color: #fff;
  box-shadow: 0px 0px 20px 0px rgba(0,0,0,0.3);
  display: block;
  margin-top: 1.5rem;
  padding: 1rem;
  position: fixed;
  right: 0;
}
```
Screenshot 2020-02-04 at 16 59 05

cntlsn:
@damianpm @HemaHG absolutely! I forgot to say more explicitly that my idea was to add those attributes on top of the existing ones!

josemigallas:
@damianpm @cntlsn 
For what is worth I think it looks better with a padding-right, since it's a floating element with the shadow and all.

@HemaHG
I would also use this opportunity to apply conditional rendering. This means you'd check the condition before rendering instead of hiding it with `display: none`.
```ruby
= render 'shared/provider/flash', flash: flash unless flash.empty?
```

thomasmaas:
I think the `display: none;` construct is to make the the flash message available to javascript (`<div class="ajaxNoticeValid">`. Could be that we no longer use that though.

Styling-wise I kind of like the error message sticking to the right but would be good to deploy to preview.

cntlsn:
@josemigallas do you mean margin right?
Just to be clear, if my [styling](https://github.com/3scale/porta/pull/1629#issuecomment-581947055) is added to the current style of `#flashWrapper`, as suggested in a [previous comment](https://github.com/3scale/porta/pull/1629#issuecomment-581987013), it should look like this:
Screenshot 2020-02-05 10 46 57

cc @damianpm @HemaHG

damianpm:
> @josemigallas do you mean margin right?
> Just to be clear, if my [styling](https://github.com/3scale/porta/pull/1629#issuecomment-581947055) is added to the current style of `#flashWrapper`, as suggested in a [previous comment](https://github.com/3scale/porta/pull/1629#issuecomment-581987013), it should look like this:
> Screenshot 2020-02-05 10 46 57
> 
> cc @damianpm @HemaHG

I prefer it aligned to the right

app/assets/stylesheets/provider/_flash.scss
```diff
@@ -1,8 +1,12 @@
 #flashWrapper {
-  position: absolute;
-  right: line-height-times(1);
-  text-align: right;
-  @include animation(fadeInLeft 0.5s ease-in-out);
+  @include animation(fadeInRight 0.5s ease-in-out);
```
damianpm: @HemaHG tests are failing because you changed `fadeInLeft` to `fadeInRight` here, but not in line 45 of this file
https://github.com/3scale/porta/blob/87c5e3253eec562d717e89adc32ff81939ce188b/app/assets/stylesheets/provider/_flash.scss#L45

app/views/layouts/provider.html.slim
```diff
@@ -18,7 +18,7 @@ html[lang="en" class="pf-m-redhat-font"]
       = render partial: '/shared/provider/header'
       = render partial: 'shared/provider/navigation/vertical_nav' unless active_menu == :dashboard
       #border.pf-c-page__main class="self_clear" role="main"
-        = render 'shared/provider/flash', flash: flash
+        = render 'shared/provider/flash', flash: flash unless flash.empty?
```
Martouta: This is dangerous because then if in the 1st page it visits with this layout there isn't a flash and then it goes to another page with the same layout (or even staying in the same page but doing an ajax request) that contains a flash, it will not appear although it should.

thomasmaas: yes, like commented https://github.com/3scale/porta/pull/1629#issuecomment-582312908

damianpm:
@thomasmaas @cntlsn 
It's deployed in preview, if you want to check
https://multitenant-admin.preview01.3scale.net/p/admin

app/assets/stylesheets/provider/_flash.scss
```diff
@@ -1,8 +1,12 @@
...
+  background-color: $white;
+  box-shadow: 0 0 20px 0 $whiteBoxShadowColor;
+  display: block;
+  margin-top: 1.5rem;
+  padding: 1rem;
+  position: fixed;
+  right: 2rem;
```
thomasmaas: It needs a z-index of about 1000 to take care of this:
Screenshot 2020-02-07 12 33 02

cntlsn:
@damianpm @HemaHG 
It looks like the text content is still wrapped in a `p` tag, which breaks the padding:
Screenshot 2020-02-07 15 01 27

Martouta:
@damianpm @HemaHG I remove myself from the reviewers here because I do not think that I am the ideal person to review this :smile:

HemaHG:
@thomasmaas, @cntlsn Updated margins and paddings with `line-height-times'
image

cntlsn:
@HemaHG @damianpm @thomasmaas as I shared in chat already, my feeling is that a more pronounced drop shadow (like the one I posted previously) will make the alert stand out more from the background. This is good UX practice, as it makes the alert more noticeable to users.

thomasmaas:
@cntlsn yes, I already told @HemaHG earlier, it's already fixed in code.

thomasmaas:
@HemaHG I would suggest using half line-height margins like so:
Screenshot 2020-02-11 16 31 28

app/assets/stylesheets/provider/_flash.scss
```diff
@@ -1,8 +1,13 @@
 #flashWrapper {
-  position: absolute;
-  right: line-height-times(1);
-  text-align: right;
-  @include animation(fadeInLeft 0.5s ease-in-out);
+  @include animation(fadeInRight 0.5s ease-in-out);
+  background-color: $white;
+  box-shadow: 0 0 20px 0 $whiteBoxShadowColor;
```
thomasmaas:
```suggestion
  box-shadow: 0 0 line-height-times(1/2) 0 $whiteBoxShadowColor;
```

app/assets/stylesheets/provider/_flash.scss
```diff
@@ -1,8 +1,13 @@
...
+  display: block;
+  margin-top: line-height-times(3 / 2);
+  padding: line-height-times(1);
```
thomasmaas:
```suggestion
  padding: line-height-times(1/2);
```

app/assets/stylesheets/provider/_flash.scss
```diff
@@ -1,8 +1,13 @@
...
+  position: fixed;
+  right: line-height-times(2);
```
thomasmaas:
```suggestion
  right: line-height-times(1/2);
```

app/assets/stylesheets/provider/_flash.scss
```diff
@@ -1,8 +1,13 @@
 #flashWrapper {
-  position: absolute;
-  right: line-height-times(1);
-  text-align: right;
-  @include animation(fadeInLeft 0.5s ease-in-out);
+  @include animation(fadeInRight 0.5s ease-in-out);
+  background-color: $white;
+  box-shadow: 0 0 20px 0 $whiteBoxShadowColor;
+  display: block;
+  margin-top: line-height-times(3 / 2);
```
thomasmaas:
```suggestion
  margin-top: line-height-times(1 / 2);
```

PR #1635, opened 2020-02-04
Fix proxy affecting changes event not issued on Settings update

guicassolato:
This PR ensures that proxy config affecting changes events are issued when the `Service`'s `backend_version` attribute changes. In theory, this should not be needed, because changing `backend_version` also causes the `Proxy`'s `authentication_method` to change, and this one is tracked already as a so called "proxy config affecting change". The problem is that sometimes we change both the service and the proxy together, the service "autosaves" the proxy, but yet the callback fails to catch `authentication_method` among the `previous_changes` and the event is never issued.

With this PR, we will issue redundant events, which is not great resources-wise, but acceptable for now. Important to recall that the event for proxy config affecting changes is idempotent.

In the future, we should replace those `after_commit` callbacks with a more intelligent mechanism to track the changes in the models.

### DIFF
--- a/app/lib/proxy_config_affecting_changes.rb
+++ b/app/lib/proxy_config_affecting_changes.rb
@@ -62,6 +62,21 @@ module ProxyConfigAffectingChanges
     end
   end
 
+  module ServiceExtension
+    extend ActiveSupport::Concern
+
+    included do
+      include ProxyConfigAffectingChanges
+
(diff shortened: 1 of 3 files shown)


guicassolato:
@thomasmaas 

> Do we have a plan for a cleaner implementation?

We do. In fact, I'm working on it right now.

----

@Martouta 

> We probably should create a Jira

Yes, we should. Have we already? Nope.

PR #1636, opened 2020-02-04
[WIP][Rails 5] Remove disable_transactional_fixtures when not needed

Martouta:
In order to speed up tests


Martouta:
Only Rails 5

PR #1638, opened 2020-02-04
OAS3: Cleaning up tests

josemigallas:
Tests were left half done due to time constraints. This PR is to make sure all necessary tests are present and the ones that are not yet supported (due to lack of plugins) are properly marked as `@wip`.

### DIFF
--- a/features/old/liquid/active_docs_v3.feature
+++ b/features/old/liquid/active_docs_v3.feature
@@ -7,8 +7,6 @@ Feature: ActiveDocs
     Given a provider "foo.example.com"
     And the current provider is foo.example.com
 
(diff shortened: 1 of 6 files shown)


hallelujah:
:rocket: Are all wip done?

josemigallas:
@hallelujah @Martouta 
All that is doable, yes. You can see there are still 2 of them that are not applicable, however I left them as a reminder for the future when we introduce the plugins.

Or maybe those test don't make sense anymore, in that case they would be removed. But right now they are present in the OAS2 test suite and I think OAS3 should be backwards compatible in that sense.

PR #1648, opened 2020-02-05
Remove EndUser and EndUserPlans [from code]

josemigallas:

[THREESCALE-2490: Remove “End-user plans” feature (SaaS) - Phase 2](https://issues.redhat.com/browse/THREESCALE-2490)

**TODO**:
- [x] All removed columns defaulted to null
- [x] Ignore columns programmatically
- [x] All code occurrences deleted

Actual removal from DB in #1667.


db/oracle_schema.rb
```diff
@@ -504,14 +504,6 @@
   add_index "deleted_objects", ["object_type", "object_id"], name: "index_deleted_objects_on_object_type_and_object_id"
   add_index "deleted_objects", ["owner_type", "owner_id"], name: "index_deleted_objects_on_owner_type_and_owner_id"
 
-  create_table "end_user_plans", force: :cascade do |t|
```
hallelujah: How did you remove this from schema?

db/postgres_schema.rb
```diff
@@ -503,14 +503,6 @@
   add_index "deleted_objects", ["object_type", "object_id"], name: "index_deleted_objects_on_object_type_and_object_id", using: :btree
   add_index "deleted_objects", ["owner_type", "owner_id"], name: "index_deleted_objects_on_owner_type_and_owner_id", using: :btree
 
-  create_table "end_user_plans", force: :cascade do |t|
```
hallelujah: Same as https://github.com/3scale/porta/pull/1648/files#r377105314

db/schema.rb
```diff
@@ -504,14 +504,6 @@
   add_index "deleted_objects", ["object_type", "object_id"], name: "index_deleted_objects_on_object_type_and_object_id", using: :btree
   add_index "deleted_objects", ["owner_type", "owner_id"], name: "index_deleted_objects_on_owner_type_and_owner_id", using: :btree
 
-  create_table "end_user_plans", force: :cascade do |t|
```
hallelujah: https://github.com/3scale/porta/pull/1648/files#r377105314

db/oracle_schema.rb
josemigallas: I just deleted the line 😳

app/representers/end_user_representer.rb
```diff
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module EndUserRepresenter
```
hallelujah: Remove this as it should not be used anymore

spec/acceptance/api/end_user_spec.rb
```diff
@@ -3,8 +3,7 @@
 resource "EndUser", transactions: false do
```
hallelujah: Remove this file as it should not be used anymore

hallelujah:
We can begin by removing the end_user_plan but we must also remove end_user

app/helpers/vertical_nav_helper.rb
```diff
@@ -195,7 +195,7 @@ def service_nav_sections
     sections << {id: :monitoring,    title: 'Analytics',     items: service_analytics}           if can? :manage, :monitoring
     sections << {id: :applications,  title: 'Applications',  items: service_applications}        if can? :manage, :applications
     sections << {id: :subscriptions, title: 'Subscriptions', items: service_subscriptions}       if can?(:manage, :service_plans) && current_account.settings.service_plans_ui_visible?
-    sections << {id: :end_users,     title: 'End-users',     items: service_end_users}           if can?(:manage, :end_users) && current_account.settings.end_user_plans_ui_visible? && !master_on_premises?
+    sections << {id: :end_users,     title: 'End-users',     items: service_end_users}           if can?(:manage, :end_users) && !master_on_premises?
```
Martouta: @thomasmaas should review this change :smile:

db/oracle_schema.rb
thomasmaas: > # This file is auto-generated from the current state of the database. Instead
> # of editing this file, please use the migrations feature of Active Record to
> # incrementally modify your database, and then regenerate this schema definition.

josemigallas: OK I created a migration here https://github.com/3scale/porta/pull/1667

josemigallas:
@hallelujah I removed all EndUser stuff and also created a migration here https://github.com/3scale/porta/pull/1667.

app/lib/switches.rb
```diff
@@ -288,7 +288,7 @@ def switches
   def globally_denied_switches
     [
       account.master_on_premises? ? :finance : nil,
-      ThreeScale.config.onpremises ? :end_users : nil
+      nil
```
Martouta: remove the line directly :smile: `compact` (the next line) removes all the `nil` values of the array :smile:

thomasmaas:
@josemigallas might be removed but just want to make sure you don't forget about any related API  methods:

Screenshot 2020-02-17 15 30 44

Martouta:
The reason why it is failing it is for this `null: false`:
https://github.com/3scale/porta/blob/f2ec9f6adb88ad2a4e9ffc4957a5f7c06cce3de5/db/schema.rb#L1244

Because it tries to insert a row in the DB and that column is not filled and cannot be null.

And I cannot do a migration with `change_column :settings, :end_users_switch, default: true`  or `change_column :settings, :end_users_switch, null: true` because it is a blocking migration as you can see:

=== Dangerous operation detected #strong_migrations ===

> Changing the type of an existing column requires the entire
> table and indexes to be rewritten. A safer approach is to:
> 
> 1. Create a new column
> 2. Write to both columns
> 3. Backfill data from the old column to the new column
> 4. Move reads from the old column to the new column
> 5. Stop writing to the old column
> 6. Drop the old column

Imo, it is not worth it and we should directly merge the removal of the column instead then. But that is a bit dangerous and we all should decide about it @3scale/system .

macejmic:
re: https://github.com/3scale/porta/pull/1648#issuecomment-587051105

maybe I would start with simple before save callback:

something like:

```

before_save :temp_end_users_switch

def temp_end_users_switch
  self.end_users_switch = true
end

```

and then remove the real DB attribute in the next PR.

Martouta:
woooooooow, strong migrations does not complain about this:

```ruby
class DefaultNilFieldEndUsersSwitch < ActiveRecord::Migration[5.0]
  def change
    change_column_null :settings, :end_users_switch, true
  end
end
```

:dancer: :dancer: :dancer: :dancer: :dancer: :dancer:

config/examples/plan_rules.yml
```diff
@@ -38,7 +38,6 @@ plan_rules: &default
       - account_plans
       - multiple_users
       - groups
-      - end_users
```
Martouta: Right :+1: But this has some implications:
1. Everyone in the team right after merging will have to do `yes | cp -rf config/examples/plan_rules.yml config/plan_rules.yml` or they environment will fail sooner or later (perhaps when resetting the DB).
2. The most important thing now is removing the references to `end_users` also in [the config/plan_rules.yml that we use for production (and preview) in SaaS](https://github.com/3scale/deploy/blob/master/system/data/production/plan_rules.yml), or we will get bad errors.
3. I do not know if the on-prem customers know about this config file, but if so, we should add documentation for upgrading to the next on-prem version saying that they should make this change in their config/plan_rules.yml

josemigallas: Question: isn't the config supposed to be copied from config/example each time?

thomasmaas: I think end-users and end-user-plans were never activated for on-prem but somebody needs to check.

app/controllers/api/usage_limits_controller.rb
```diff
@@ -55,20 +57,14 @@ def destroy
...
 
   def find_plan
-    @plan = find_application_plan || find_end_user_plan
+    @plan = find_application_plan
   end
 
   def find_application_plan
     id = params[:application_plan_id]
     current_account.application_plans.find(id) if id
   end
```
Martouta: I think we should do the following and remove the other method:
```ruby
def find_plan
  @plan = current_account.application_plans.find(params[:application_plan_id])
end
```
So we make this param mandatory and if it isn't there, it will raise a record not found and return a 404, which is better than the ugly error of "method not found for nil class" of not finding the plan because the id is not in the param and then doing `@plan.whatever_method` ass I see it is called for all methods because the plan is mandatory anyway.

app/workers/backend_delete_end_users_worker.rb
```diff
@@ -1,15 +0,0 @@
-# frozen_string_literal: true
-
-class BackendDeleteEndUsersWorker
```
Martouta: Removing this means that either we run this for all services before merging, or Apisonator should remove all end-users of all services on their own (ideally the latter).
Btw, either way, Apisonator and Pisoni should eventually remove the code regarding end-users :smile: and although that is not up to us, we should ensure that they have a Jira for it or we should tell them by a message or something :smile:

lib/tasks/revival.thor
```diff
@@ -29,7 +29,6 @@ class Revival < Thor
   SERVICE_ASSOCIATIONS = [
     { service_plans: [:service_contracts] },
     :application_plans,
-    { end_user_plans: [:plan_limits]},
```
Martouta: I guess this is correct but what do we use this revival.thor for? @3scale/system

lib/system/database/definitions/mysql.rb
```diff
@@ -63,12 +63,6 @@
     SQL
   end
 
-  trigger 'end_user_plans' do
```
Martouta: This means that we must recreate the triggers. In SaaS for sure and probably in on-prem too.

We made it with a migration about 1 year ago in https://github.com/3scale/porta/pull/1047
That would work both for SaaS and for on-prem. But I do not know if the migration is blocking... which by then didn't matter because nobody was using that table anyway, but we cannot block the services table so it is not the same XD **if** it is blocking (I remark the IF because I do not know), we shouldn't make it this way.

We could also open [an issue for Ops](https://github.com/3scale/infrastructure/issues) to recreate triggers in SaaS themselves then :smile: You have an example in https://github.com/3scale/infrastructure/issues/759
But that would be only for SaaS and we should think what to do for on-prem (maybe they should execute `Rake::Task['db:triggers'].invoke`  but probably we should make a migration for them that is not executed in SaaS so it is much more efficient).

Martouta: Sorry, it is in the end_user_plans table, my bad, so we can just do the migration no matter if it is blocking or not because nobody uses that table :wink:

Martouta: We spoke about this more in-depth in private :+1:

app/workers/backend_delete_end_users_worker.rb
josemigallas: @3scale/system-ruby thoughts?

Martouta: @davidor maybe can give us some input :smile:

davidor: I haven't followed all the convesation in this PR, but the code in Apisonator was removed some time ago (2.7?) when someone confirmed that there were no customers using end-users anymore.

doc/active_docs/Service Management API (on-premises).json
```diff
@@ -56,7 +56,7 @@
               "name": "user_id",
               "dataType": "string",
               "paramType": "query",
-              "description": "User id. String identifying an end user. Required only when the application is rate limiting end users. The End User plans feature is not available in all 3scale plans."
+              "description": "User id. String identifying an end user. Required only when the application is rate limiting end users."
```
hallelujah: Shouldn't we remove this?

doc/active_docs/Service Management API (on-premises).json
```diff
@@ -132,7 +132,7 @@
               "name": "user_id",
               "dataType": "string",
               "paramType": "query",
-              "description": "User id. String identifying an end user. Required only when the application is rate limiting end users. The End User plans feature is not available in all 3scale plans."
+              "description": "User id. String identifying an end user. Required only when the application is rate limiting end users. ."
```
hallelujah: https://github.com/3scale/porta/pull/1648/files#r385136206

doc/active_docs/Service Management API (on-premises).json
```diff
@@ -216,7 +216,7 @@
               "name": "user_id",
               "dataType": "string",
               "paramType": "query",
-              "description": "User id. String identifying an end user. Used only when the application is rate limiting end users and the specified token is not associated to a user. The End User plans feature is not available in all 3scale plans."
+              "description": "User id. String identifying an end user. Used only when the application is rate limiting end users and the specified token is not associated to a user. ."
```
hallelujah: https://github.com/3scale/porta/pull/1648/files#r385136206

doc/active_docs/Service Management API (on-premises).json
```diff
@@ -378,7 +378,7 @@
               "name": "user_id",
               "dataType": "string",
               "paramType": "query",
-              "description": "User id. String identifying an end user. Required only when the application is rate limiting end users. The End User plans feature is not available in all 3scale plans."
+              "description": "User id. String identifying an end user. Required only when the application is rate limiting end users. ."
```
hallelujah: https://github.com/3scale/porta/pull/1648/files#r385136206

db/migrate/20200224095152_default_nil_fields_end_user_removal.rb
```diff
@@ -0,0 +1,13 @@
+class DefaultNilFieldsEndUserRemoval < ActiveRecord::Migration[5.0]
```
hallelujah: Shouldn't we do it separately from this PR? Prepare the database so it can accept null values then remove from code the usage? Maybe I am missing missing something 🤔

doc/active_docs/Service Management API (on-premises).json
josemigallas: IDK.. I think it's just a mention of EndUsers but it's not directly related

lib/system/database/definitions/mysql.rb
hallelujah: I agree that normally the table is not accessed but for security it is preferable to have this trigger deleted before this PR, so we in case something bad happens like some unlikely queries with invalid data, it won't just stop the request to finish.

lib/tasks/revival.thor
hallelujah: I made it ... This is to revive some deleted data from a database backup. I did few times to restore some accidental removal of tenants :D 
I think it is outdated now anyway

db/migrate/20200224095152_default_nil_fields_end_user_removal.rb
josemigallas: I agree

josemigallas: Boom #1734

app/controllers/api/usage_limits_controller.rb
```diff
@@ -55,20 +57,10 @@ def destroy
...
 
   def find_plan
-    @plan = find_application_plan || find_end_user_plan
-  end
-
-  def find_application_plan
     id = params[:application_plan_id]
-    current_account.application_plans.find(id) if id
+    @plan = current_account.application_plans.find(id) if id
   end
```
Martouta: Personally I would do this:
```ruby
@plan = current_account.application_plans.find(params[:application_plan_id])
```

Because if that param is not sent, right now it will raise an error with NoMethodError for NilClass and will return an ugly 5XX to the user. But with what I am suggesting, it returns a 404, which is what it should do in this case considering that everything inside this controller now is inside this plan.

doc/active_docs/Service Management API (on-premises).json
Martouta: I agree with Hery. This user_id is to identify and EndUser, which we are removing.

hallelujah: Also removing the parameter for it in the controller

app/controllers/api/metrics_controller.rb
```diff
@@ -136,17 +138,7 @@ def build_metric
   end
 
   def find_plan_and_service
-    @plan = find_application_plan || find_end_user_plan
+    @plan = current_account.application_plans.find(params[:application_plan_id])
```
Martouta: @josemigallas Are you sure that the plan is mandatory for this controller as well? I'm not sure :thinking: And if so, you need to remove the conditional in the next line because if there is no plan, this line will raise an error ActiveRecord::RecordNotFound (that will be rescued and responded with a 404)

doc/active_docs/Service Management API (on-premises).json
```diff
@@ -308,7 +290,7 @@
...
             },
```
Martouta: Remove the whole user_id attribute :smile: Btw, I see that is it not defined anywhere in the code and only in the jsons. I think that you removed it completely in the other json: https://github.com/3scale/porta/pull/1648/files#diff-c693297ce0ff418d3a1e8d2c9fdfa5f8L216
And that is correct, but u should do the same here :smile:

app/controllers/api/metrics_controller.rb
josemigallas: Oops, yeah you're right

lib/system/database/definitions/mysql.rb
josemigallas: Here you go! #1738

app/lib/backend/model_extensions/cinstance.rb
```diff
@@ -39,8 +39,7 @@ def update_backend_application
                                               :state      => state,
                                               :plan_id    => plan.id,
                                               :plan_name  => plan.name,
-                                              :redirect_url => redirect_url,
-                                              :user_required => end_user_required.nil? ? plan.end_user_required : end_user_required  )
```
guicassolato: Is it OK at both pisoni and apisonator if we don't pass this attribute anymore?

Martouta: https://github.com/3scale/porta/pull/1648#discussion_r385117376 @guicassolato

test/factories/account.rb
```diff
@@ -143,7 +143,6 @@
       account.settings.update_attributes!(
         account_plans_ui_visible: true,
         service_plans_ui_visible: true,
```
guicassolato:
```suggestion
        service_plans_ui_visible: true
```

PR #1649, opened 2020-02-05
private define_method in ruby 2.4

hallelujah:
We want to switch to ruby 2.5 at least but for now we need to be compatible with 2.4
See https://bugs.ruby-lang.org/issues/14133
Closes [THREESCALE-4409](https://issues.redhat.com/browse/THREESCALE-4409
)


hallelujah:
> sure we shouldn't be using `class_eval`?

No because we want to work on the instance of the module not the  `Module` class if we do so then all objects (really all 😱 ) will define those methods

PR #1655, opened 2020-02-05
Hello again Yarn

didierofrivia:

Since Rails 5.1, Yarn is adopted as the default node package manager. At it's core, webpacker gem also uses it, making it hacky to use npm as an alternative. We've made many changes in the past to avoid using Yarn because previously there was not a way to install it from npm, and it used Ocaml and we couldn't productize due some licenses issues. Now, Yarn is delivered as a Node package also, which will make it "productizable".

Fixes https://issues.redhat.com/browse/THREESCALE-4000

**Verification steps** 
1. Circle CI should be green.
2. Deploying to both SaaS and OS should go smoothly.

**TODO**
* [x] Fix SaaS deploy - https://github.com/3scale/deploy/pull/641
* [x] Fix OS images


spec/javascripts/Settings/components/Common/__snapshots__/TextInputGroup.spec.jsx.snap
```diff
@@ -1,24 +1,69 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`should render correctly 1`] = `
-<FormGroup
-  fieldId="service_proxy_attributes_dooz_kawa_chanson_input"
-  helperText="Enter your favourite Dooz Kawa tune"
+<TextInputGroup
```
didierofrivia: I know this doesn't quite belong to this PR **BUT** the `package-lock` migration to `yarn.lock` migration resulted to be quite not straightforward with the dependencies versions. It ended upgrading, besides many other packages, Patternfly **minor** version... which it introduces the use of references that React suggest to libraries maintainers to release it as a **major** version upgrade. SO, live with it.

hallelujah: Did you use `yarn import` to generate the yarn.lock?

https://classic.yarnpkg.com/blog/2018/06/04/yarn-import-package-lock/

didierofrivia: Yes, I did

openshift/system/Dockerfile
```diff
@@ -62,7 +63,8 @@ RUN export ${BUNDLER_ENV} >/dev/null \
     && chmod g+w -vfR log tmp public/assets db/sphinx \
     && umask 0002 \
     && cd /opt/system \
-    && npm install --no-progress \
+    && npm install -g yarn \
```
josemigallas: If we're ditching `npm`, I think we should install `yarn` directly.

openshift/system/Dockerfile.on_prem
```diff
@@ -46,13 +46,15 @@ ENV BASH_ENV=/opt/system/etc/scl_enable \
 RUN export ${BUNDLER_ENV} >/dev/null\
     && source /opt/system/etc/scl_enable \
     && gem install bundler --version 1.17.3 \
+
```
josemigallas: Ghost extra line

.gitignore
```diff
@@ -137,3 +137,10 @@ Gemfile.local
 # Typings
 flow-typed
 config/cable.yml
+
+# Yarn
+.yarn/cache
+.yarn/unplugged
+.yarn/build-state.yml
+.pnp.*
+package-lock.json
```
josemigallas: If we're not supposed to have a package-lock, we should not ignore it. If somebody ever commits it, it's a sign they did something wrong.

.circleci/config.yml
```diff
@@ -501,7 +498,7 @@ jobs:
       - run:
           name: Karma tests
           command: |
-            npm test -- --reporters dots,junit --browsers ChromeHeadless
+            yarn test -- --reporters dots,junit --browsers ChromeHeadless
```
josemigallas: I'm not sure but I think this is not required:
```suggestion
            yarn test --reporters dots,junit --browsers ChromeHeadless
```

package.json
```diff
@@ -6,9 +6,8 @@
   },
   "license": "Apache-2.0",
   "scripts": {
+    "install": "yarn check --integrity || yarn install --pure-lockfile --frozen-lockfile --ignore-scripts --skip-integrity-check --prefer-offline --link-duplicates",
```
josemigallas: `yarn check` [will be deprecated](https://classic.yarnpkg.com/en/docs/cli/check/#toc-yarn-check) in yarn 2.0 and can be replaced by `--check-files`.
I think overriding the default script is going to cause us some painful moments.

`frozen-lockfile` is redundant, it already includes `pure-lockfile`.
`ignore-scripts` is a potential constraint, it's better to just not to call it `install`.
`skip-integrity-check` doesn't make sense anymore (because of point made at line 1)
```suggestion
    "install:safe": "yarn install --frozen-lockfile --ignore-scripts --check-files --prefer-offline --link-duplicates",
```

josemigallas: What's actually the reasoning behind `ignore-scripts`?

script/docker.sh
```diff
@@ -27,10 +27,10 @@ bundle config
 echo
 
 echo
-echo "======= NPM ======="
+echo "======= Node ======="
```
josemigallas: Not a big deal but I think we should be explicit and change NPM for Yarn.

spec/javascripts/Settings/components/Common/TextInputGroup.spec.jsx
```diff
@@ -1,5 +1,5 @@
 import React from 'react'
-import { shallow } from 'enzyme'
+import { mount } from 'enzyme'
```
josemigallas: This shouldn't really be in this PR.

openshift/system/Dockerfile.on_prem
```diff
@@ -62,7 +64,8 @@ RUN export ${BUNDLER_ENV} >/dev/null \
     && chmod g+w -vfR log tmp public/assets db/sphinx \
     && umask 0002 \
     && cd /opt/system \
-    && npm install --no-progress \
+    && npm install -g yarn \
+    && yarn install --no-progress \
```
josemigallas: This doesn't seem to do anything 🤔 neither does `--silent` (I tried locally)
```suggestion
    && yarn install \
```

INSTALL.md
```diff
@@ -188,12 +188,19 @@ bundle config --local build.mysql2 "--with-ldflags=-L/usr/local/opt/openssl/lib
...
 
-#### NPM
+#### Node packages
 
-Run [NPM](https://www.npmjs.com/) to install all the required Node modules:
+Install [Yarn](https://yarnpkg.com/) globally:
 
 ```bash
-npm install
+npm install -g yarn
```
josemigallas: I'd suggest homebrew before npm (since this is targeted to MacOS). In any case, I would not suggest anything at all. Providing the link above and leaving the user decide would be better IMHO.
```suggestion
brew install yarn
```

package.json
```diff
@@ -6,9 +6,8 @@
...
     "clean": "rm -rf node_modules/ && npm cache clean --force",
```
josemigallas:
```suggestion
    "clean": "rm -rf node_modules/ && yarn cache clean",
```

spec/javascripts/Settings/components/Common/__snapshots__/TextInputGroup.spec.jsx.snap
josemigallas: Why not coping with the upgrade in another PR? Either install the older version by hard here or upgrade to the latest in master, with the change, and rebase this one.

INSTALL.md
didierofrivia: That's the reason why we ditched it in the first place. The brew package uses ocaml, licenses not quite open source friendly...

spec/javascripts/Settings/components/Common/__snapshots__/TextInputGroup.spec.jsx.snap
didierofrivia: I tried to do it, but it triggered some incompatibility issues with other packages. In the end installing many versions of other dependencies... Was a waste. I think we can totally cope with this.

openshift/system/Dockerfile
didierofrivia: Read the description of the PR and the issue attached.

.gitignore
didierofrivia: We should, like any other file that shouldn't be commited. There are instructions to follow and if there's a change in package.json and not in yarn.lock, we will address in the PR review

script/docker.sh
didierofrivia: In the end we are managing Node packages.

spec/javascripts/Settings/components/Common/TextInputGroup.spec.jsx
didierofrivia: https://github.com/3scale/porta/pull/1655#discussion_r376983311

openshift/system/Dockerfile.on_prem
didierofrivia: Well, it doesn't appear in the logs of circle, OS and capistrano... so seems to be working

spec/javascripts/Settings/components/Common/__snapshots__/TextInputGroup.spec.jsx.snap
josemigallas: Yeah well, it's not a big deal

script/docker.sh
josemigallas: To be completely precise it's npm packages but again, not a big deal 😝

spec/javascripts/Settings/components/Common/TextInputGroup.spec.jsx
josemigallas: Right

INSTALL.md
josemigallas: For the production env I agree with you but for development it's inconsistent that we use this tool with other dependencies (see [here](https://github.com/3scale/porta/blob/master/INSTALL.md#dependencies)) but not with Yarn.

script/docker.sh
didierofrivia: npm stands for Node Package Manager ;)

INSTALL.md
hallelujah: Btw I think there was a confusion when we search for how to productize yarn
It does not use OCaml at all ... (since the beginning) it just have tests that queries the OCaml repository which lead to this confusion ....

yarn can be used to resolve OCaml repositories ... that is why the tests were there ....
But we did not invest much time on it and jump into the conclusion that OCaml is necessary
https://github.com/yarnpkg/yarn/commit/8be4eef75414db89875f5c64b71cf06783b4507a

openshift/system/Dockerfile.on_prem
```diff
@@ -62,7 +63,8 @@ RUN export ${BUNDLER_ENV} >/dev/null \
     && chmod g+w -vfR log tmp public/assets db/sphinx \
     && umask 0002 \
     && cd /opt/system \
-    && npm install --no-progress \
+    && npm install -g yarn \
+    && yarn install:safe --no-progress \
```
hallelujah: Hmmm I have an issue locally

```shell
$ yarn install:safe
yarn run v1.17.3
error Command "install:safe" not found.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

script/docker.sh
hallelujah: Is this file still required??? 😱

hallelujah:
I think we can remove the script/docker.sh completely ...
I do not think it is used anywhere (grep => no result)

openshift/system/Dockerfile.on_prem
hallelujah: My bad I was on `master` branch locally

PR #1658, opened 2020-02-06
Refactoring ProxyConfigAffectingChanges

guicassolato:
**TODOs**

- [x] Add tests for each type of proxy config affecting object (`Service`, `Proxy`, `ProxyRule`, `BackendApi`, `BackendApiConfig`). Make sure to cover the following cases:
    1. Create Product
    2. Add product-level mapping rule
    3. Edit product-level mapping rule
    4. Delete product-level mapping rule
    5. Add Policy config
    6. Edit Policy config
    7. Delete Policy config
    8. Edit Product setting attribute (e.g. authentication method)
    9. Edit Proxy config attribute (e.g. error message)
    10. Stale settings update
    11. Add Backend
    12. Edit Backend path
    13. Edit Backend endpoint
    14. Add backend-level mapping rule 
    15. Edit backend-level mapping rule
    16. Delete product-level mapping rules
    17. Remove Backend
    18. Delete Product
- [x] Add unit test for the `Tracker`
- [x] Add integration test for the controller
- [ ] Fix https://github.com/3scale/porta/blob/53f1027ebc9a1656e5c30d043c245b3ba7f363c6/app/lib/proxy_config_affecting_changes.rb#L77-L80
- [ ] Fix https://issues.redhat.com/browse/THREESCALE-4470

### DIFF
--- a/app/lib/proxy_config_affecting_changes.rb
+++ b/app/lib/proxy_config_affecting_changes.rb
@@ -1,115 +1,154 @@
 # frozen_string_literal: true
 
 module ProxyConfigAffectingChanges
-  module ProxyRuleExtension
+  class TrackedObject
+    def initialize(object)
+      @object = object
+      @initial_state = object.proxy_config_affecting_state.hash
+      @new_record = object.new_record?
+      @proxies = find_proxies(object)
+    end
+
+    attr_reader :object, :initial_state, :new_record
+
+    def ==(other)
+      object == other.object
+    end
+
+    delegate :destroyed?, :persisted?, to: :object
+
+    def created?
+      new_record && persisted?
+    end
+
+    def state_changed?
+      initial_state != object.reload.proxy_config_affecting_state.hash
+    rescue ActiveRecord::RecordNotFound
+      !new_record # Cannot find the model. If the object was a new record, then it was not persisted; otherwise, it was destroyed.
+    end
+
+    def changed?
+      destroyed? || created? || state_changed?
+    end
+
+    def proxies
+      @proxies.presence || find_proxies
+    end
+
+    protected
+
+    def find_object
+      object.class.find_by(id: object.id)
+    end
+
+    def find_proxies(model = find_object)
+      proxy_or_proxies = model.try(:proxy) || model.try(:proxies)
+      [*proxy_or_proxies].flatten.compact
+    end
+  end
+
+  class Tracker
+    def initialize
+      @tracked_objects = []
+    end
+
+    def track(object)
+      tracked_object = TrackedObject.new(object)
+      return if tracking?(tracked_object)
+      @tracked_objects << tracked_object
+    end
+
+    def tracking?(tracked_object)
+      @tracked_objects.include?(tracked_object)
+    end
+
+    def objects_with_affecting_changes
+      @tracked_objects.select(&:changed?)
+    end
+
+    def flush
+      proxies = objects_with_affecting_changes.map(&:proxies).flatten.uniq
+      proxies.each(&method(:issue_proxy_affecting_change_event))
+    ensure
+      @tracked_objects.clear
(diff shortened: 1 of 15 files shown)


app/lib/proxy_config_affecting_changes.rb
```diff
@@ -1,115 +1,158 @@
...
+    def track(object)
+      return unless object
+      tracked_object = TrackedObject.new(object)
+      return if tracking?(tracked_object)
+      @tracked_objects << tracked_object
+    end
 
-      include ProxyConfigAffectingChanges
+    def tracking?(object)
+      tracked_object = object.is_a?(TrackedObject) ? object : TrackedObject.new(object)
```
hallelujah: Why this check?

app/lib/proxy_config_affecting_changes.rb
```diff
@@ -1,115 +1,158 @@
...
+      @tracked_objects.include?(tracked_object)
```
hallelujah: If using a Hash with proper keys we would avoid using `include?` and doing `object == tracked.object` comparison

app/lib/proxy_config_affecting_changes.rb
```diff
@@ -1,115 +1,158 @@
...
-      end
+    protected
 
-      private
+    def issue_proxy_affecting_change_event(proxy)
+      # Sometimes invoked in the context of or concurrently to a hierarchy deletion,
+      # hence the proxy, the service or the account may no longer be available
+      return unless proxy
+      service = proxy.service
+      return unless service&.account
```
hallelujah: If we are using the `&.` operator why not chaining directly as you do not need `service` nor `account` variable

```ruby
return unless proxy&.service&.account
```

app/lib/proxy_config_affecting_changes.rb
```diff
@@ -1,115 +1,158 @@
...
+    def flush
+      proxies = objects_with_affecting_changes.map(&:proxies).flatten.uniq
+      proxies.each(&method(:issue_proxy_affecting_change_event))
+    ensure
+      @tracked_objects.clear
+    end
 
-        issue_proxy_affecting_change_event(self)
-      end
+    # FIXME: This is only so ProxyConfigs::AffectingObjectChangedEvent does not crash
```
hallelujah: Could you describe the issue so we could help a bit?

app/lib/proxy_config_affecting_changes.rb
```diff
@@ -1,115 +1,157 @@
...
+      proxy_or_proxies = model.try(:proxy) || model.try(:proxies)
+      [*proxy_or_proxies].flatten.compact
     end
   end
 
-  module ProxyExtension
-    extend ActiveSupport::Concern
+  class Tracker
+    def initialize
+      @tracked_objects = []
```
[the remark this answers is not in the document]
guicassolato: We cannot use `global_id` because it's only present when the object has an ID.

app/lib/proxy_config_affecting_changes.rb
guicassolato: https://github.com/3scale/porta/blob/4b618c0fe830d85a245a53a9d1f2c7ddb0725c3d/app/events/proxy_configs/affecting_object_changed_event.rb#L7

app/lib/proxy_config_affecting_changes.rb
guicassolato: Sorry. My bad. This was actually to ease the tests. Not cool allowing it to bend the implementation.

app/lib/proxy_config_affecting_changes.rb
guicassolato: The object would have to be the key in the Hash, which just felt wrong to me.

PR #1660, opened 2020-02-06
Require CSV in Csv::Exporter

duduribeiro:
This commit forces the require of the csv lib in the Csv::Exporter
class. This will fix the problem that we have when Sidekiq process this
in background and explodes with the `undefined method `generate' for
CSV:Module` error.


duduribeiro:
Closed in favor of https://github.com/3scale/porta/pull/1659

PR #1661, opened 2020-02-06
Upgrade i18n gem & Remove money gem

duduribeiro:

Our i18n gem is currently in 0.7.0 and this version of i18n has a
vulnerability. This commit upgrades our i18n gem version to fix this.

It also removes the money gem.

[THREESCALE-2922](https://issues.redhat.com/browse/THREESCALE-2922)


Gemfile.base
```diff
@@ -64,7 +64,7 @@ gem 'yabeda-prometheus'
 
 gem 'activemerchant', '~> 1.77.0'
 gem 'active_merchant-adyen12', github: '3scale/active_merchant-adyen12'
-gem 'money', '6.7.1' # Lock to this version we use in production
+gem 'money', '~> 6.0'
```
hallelujah: I'd be more specific `~> 6.13`

Martouta:
That last commit definitely requires updating the description of the PR :stuck_out_tongue_closed_eyes: @duduribeiro

guicassolato:
ActiveMerchant won't support it in the future anyway, so good to remove it if we can: https://github.com/activemerchant/active_merchant/blob/114ec71e9d8a9a51d867a35753d308868a993a32/lib/active_merchant/billing/gateway.rb#L246

We already use Integer value in cents for the amounts, BTW: https://github.com/3scale/porta/blob/518d5868b27842395737d118ac2a306f5a81b7bc/app/models/payment_transaction.rb#L45

duduribeiro:
@Martouta will do 😂.. this is just a testing haha

PR #1662, opened 2020-02-07
[Bugfix] Fix Provider::SignupsController#create & tests.

Martouta:
Closes [THREESCALE-4423](https://issues.redhat.com/browse/THREESCALE-4423)

- Make Provider::SignupsControllerTest runnable by adding the '.each'
- Fields must be built AFTER building user and account. ActionView::Template::Error: undefined method `defined_extra_fields' for nil:NilClass
- Test with invalid email
- Do not test the value of the password
- Ensure variables @Provider and @user even when spam does not pass
- Ensure the correct value of subdomain and singup_mode
- Test signup_type

### DIFF
--- a/app/controllers/provider/signups_controller.rb
+++ b/app/controllers/provider/signups_controller.rb
@@ -27,14 +27,14 @@ class Provider::SignupsController < Provider::BaseController
 
   def create
     @plan = plan
-    @fields = Fields::SignupForm.new(@provider, @user, params[:fields])
     signup_result = Signup::ProviderAccountManager.new(master).create(signup_params) do |result|
-      account = result.account
-      account.signup_mode!
-      break unless spam_check(account)
+      @provider = result.account
+      @user = result.user
+      @provider.signup_mode!
+      @provider.subdomain = account_params[:subdomain]
+      result.add_error(message: 'spam check failed') unless spam_check(@provider)
     end
-    @provider = signup_result.account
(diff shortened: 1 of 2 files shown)


app/controllers/provider/signups_controller.rb
```diff
@@ -27,14 +27,14 @@ def show # original iframe form
...
-    @fields = Fields::SignupForm.new(@provider, @user, params[:fields])
     signup_result = Signup::ProviderAccountManager.new(master).create(signup_params) do |result|
-      account = result.account
-      account.signup_mode!
-      break unless spam_check(account)
+      @provider = result.account
+      @user = result.user
+      @provider.signup_mode!
+      @provider.subdomain = account_params[:subdomain]
+      result.add_error(message: 'spam check failed') unless spam_check(@provider)
```
Martouta: Probably change the message

app/controllers/provider/signups_controller.rb
```diff
@@ -27,14 +27,14 @@ def show # original iframe form
...
     end
-    @provider = signup_result.account
-    @user = signup_result.user
+    @fields = Fields::SignupForm.new(@provider, @user, params[:fields])
 
     return render :show unless signup_result.persisted?
```
Martouta: i guess we should show a flash message or something show the error (maybe we do)

PR #1663, opened 2020-02-10
adds yarn install steps to MacOS setup

kevprice83:
Seems we are using Yarn instead of NPM now? I had to run the commands added in this commit but it also raises the question whether NPM is needed at all here? How to migrate from NPM to Yarn for contributors, are my steps sufficient and correct?

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -183,12 +183,18 @@ bundle config --local build.mysql2 "--with-ldflags=-L/usr/local/opt/openssl/lib 
 
 and run `bundle install` again.
 
-#### NPM
+#### Yarn
 
-Run [NPM](https://www.npmjs.com/) to install all the required Node modules:
+Install [Yarn](https://yarnpkg.com/):
 
 ```bash
-npm install
+npm install -g yarn
+```
+
+Run [Yarn](https://yarnpkg.com/) to install all the required packages:
+
+```bash
+yarn install
 ```
 
 ## Manual setup on Fedora (29)


didierofrivia:
@kevprice83 We are not currently using Yarn just yet. There's a draft addressing the future change, here: https://github.com/3scale/porta/pull/1655
*However*, current images and SaaS **are not supporting Yarn**

kevprice83:
Ah sorry you're right @didierofrivia I should've added this commit to [PR#1606](https://github.com/3scale/porta/pull/1606). I'll close this PR and add it there in case that ever gets merged in the future.

kevprice83:
@didierofrivia @hallelujah I see that #1606 was merged already. Should we be adding the steps I documented in this PR to the INSTALL.md? If not then what needs to change because currently I have to run both `npm install -g yarn` and `yarn install` to get my local environment working.

INSTALL.md
```diff
@@ -183,12 +183,18 @@ bundle config --local build.mysql2 "--with-ldflags=-L/usr/local/opt/openssl/lib
...
 
 ```bash
-npm install
+npm install -g yarn
+```
+
+Run [Yarn](https://yarnpkg.com/) to install all the required packages:
+
+```bash
+yarn install
```
hallelujah: There is another task 

```suggestion
yarn install:safe
```

It would be also be nice to update the `./bin/setup` for yarn install and such

https://github.com/3scale/porta/blob/5c4ce627e6b0a3bd79e5ae55b877d627bdea4c06/bin/setup

Normally we should provide one way boot of the application locally:

* Install necessary development libraries depending on the OS
* Run `./bin/setup` which will do minimal steps to have the project running locally i.e. : install gems, install node packages, start the server

didierofrivia: The Installation guideline is already updated, was part of #1655  https://github.com/3scale/porta/pull/1655/commits/d1608bad472bd975426676859042d94320dbcae9

didierofrivia:
@kevprice83 This PR is no longer valid since it was updated in https://github.com/3scale/porta/pull/1655, hence the conflicting file. However, you could explore what @hallelujah suggested about the `bin/setup`

hallelujah:
Closing

PR #1664, opened 2020-02-10
Adding PF4 styles to change password

damianpm:

Change password is missing styling, adding PF4 to that page

https://issues.redhat.com/browse/THREESCALE-4440

**Verification steps** 

Go to `/p/login?request_password_reset=true`
Request new password
In the console, you will find the link sent:

> Dear xxx,
> 
> You can reset your password by visiting the following link:
> 
> http://provider-admin.example.com/p/password?password_reset_token=xxxxxxxxxxxxxxxx

**Before:**
before

**After:**
after

### DIFF
--- a/app/javascript/packs/ChangePassword.js
+++ b/app/javascript/packs/ChangePassword.js
@@ -1,0 +1,8 @@
+import { ChangePasswordWrapper as ChangePassword } from 'ChangePassword'
+import { safeFromJsonString } from 'utilities/json-utils'
+
+document.addEventListener('DOMContentLoaded', () => {
+  const changePasswordContainer = document.getElementById('pf-login-page-container')
+  const changePasswordProps = safeFromJsonString(changePasswordContainer.dataset.changePasswordProps)
+  ChangePassword(changePasswordProps, 'pf-login-page-container')
+})
--- a/app/javascript/src/ChangePassword/components/ChangePassword.jsx
+++ b/app/javascript/src/ChangePassword/components/ChangePassword.jsx
@@ -1,0 +1,100 @@
+// @flow
+
+import React, { useState } from 'react'
+import { createReactWrapper } from 'utilities/createReactWrapper'
+import { LoginPage, Form, ActionGroup, Button } from '@patternfly/react-core'
+import { HiddenInputs, FlashMessages } from 'LoginPage'
+import { PasswordInput } from 'ChangePassword'
+import type { FlashMessage } from 'Types'
+import { validateForm } from 'LoginPage/utils/formValidation'
+
+import 'LoginPage/assets/styles/loginPage.scss'
+import brandImg from 'LoginPage/assets/images/3scale_Logo_Reverse.png'
+import PF4DownstreamBG from 'LoginPage/assets/images/PF4DownstreamBG.svg'
+
+type Props = {
+  lostPasswordToken: ?string,
+  url: string,
+  errors: (?FlashMessage)[]
+}
+
+const validationConstraints = {
+  'user[password]': {
+    presence: true,
+    length: { minimum: 1 }
+  },
+  'user[password_confirmation]': {
+    presence: true,
+    length: { minimum: 1 },
+    equality: 'user[password]'
+  }
+}
+
+const ChangePassword = ({ lostPasswordToken, url, errors }: Props) => {
+  const [isFormDisabled, setIsFormDisabled] = useState(true)
+  const [passwordsMatch, setPasswordsMatch] = useState(true)
+
+  const onChange = (event) => {
+    const errors = validateForm(event.currentTarget, validationConstraints)
+    const passwordConfirmationError = (errors && errors['user[password_confirmation]']) ? errors['user[password_confirmation]'][0] : ''
+    const doesPasswordsMatch = !passwordConfirmationError.includes('is not equal to')
+    setIsFormDisabled(!!errors)
+    setPasswordsMatch(doesPasswordsMatch)
+  }
+
+  return (
+    <LoginPage
+      brandImgSrc={brandImg}
+      brandImgAlt='Red Hat 3scale API Management'
+      backgroundImgSrc={PF4DownstreamBG}
+      backgroundImgAlt='Red Hat 3scale API Management'
+      loginTitle='Change Password'
+      footer={null}
+    >
+      {errors && <FlashMessages flashMessages={errors}/>}
+      <Form
+        action={url}
+        noValidate
+        id='edit_user_2'
+        acceptCharset='UTF-8'
+        method='post'
+        onChange={onChange}
+      >
+        <input type='hidden' name='_method' value='put' />
+        <HiddenInputs />
+        <PasswordInput
+          isRequired
+          name='password'
+          label='Password'
+          autoFocus='autoFocus'
+        />
+        <PasswordInput
+          isRequired
+          name='password_confirmation'
+          label='Password confirmation'
+        />
+        {!passwordsMatch &&
+          <div className="pf-c-form__helper-text pf-m-error" aria-live="polite">Password and password confirmation must match</div>
+        }
+        {lostPasswordToken &&
+          <input id='password_reset_token' type='hidden' name='password_reset_token' value={lostPasswordToken} />
+        }
+        <ActionGroup>
+          <Button
+            className='pf-c-button pf-m-primary pf-m-block'
+            type='submit'
+            name='commit'
+            isDisabled={isFormDisabled}
+          >
+            Change Password
+          </Button>
+        </ActionGroup>
+      </Form>
+    </LoginPage>
+  )
+}
+
+const ChangePasswordWrapper = (props: Props, containerId: string) =>
+  createReactWrapper(<ChangePassword {...props} />, containerId)
+
+export { ChangePassword, ChangePasswordWrapper }
--- a/app/javascript/src/ChangePassword/components/ChangePassword.jsx
+++ b/app/javascript/src/ChangePassword/components/ChangePassword.jsx
@@ -25,21 +25,16 @@ const validationConstraints = {
   },
   'user[password_confirmation]': {
     presence: true,
-    length: { minimum: 1 },
-    equality: 'user[password]'
+    length: { minimum: 1 }
   }
 }
 
 const ChangePassword = ({ lostPasswordToken, url, errors }: Props) => {
   const [isFormDisabled, setIsFormDisabled] = useState(true)
-  const [passwordsMatch, setPasswordsMatch] = useState(true)
 
   const onChange = (event) => {
     const errors = validateForm(event.currentTarget, validationConstraints)
-    const passwordConfirmationError = (errors && errors['user[password_confirmation]']) ? errors['user[password_confirmation]'][0] : ''
-    const doesPasswordsMatch = !passwordConfirmationError.includes('is not equal to')
     setIsFormDisabled(!!errors)
-    setPasswordsMatch(doesPasswordsMatch)
   }
 
   return (
@@ -73,9 +68,6 @@ const ChangePassword = ({ lostPasswordToken, url, errors }: Props) => {
           name='password_confirmation'
           label='Password confirmation'
         />
-        {!passwordsMatch &&
-          <div className="pf-c-form__helper-text pf-m-error" aria-live="polite">Password and password confirmation must match</div>
-        }
         {lostPasswordToken &&
           <input id='password_reset_token' type='hidden' name='password_reset_token' value={lostPasswordToken} />
         }
--- a/app/javascript/src/ChangePassword/components/PasswordInput.jsx
+++ b/app/javascript/src/ChangePassword/components/PasswordInput.jsx
@@ -1,0 +1,37 @@
+// @flow
+
+import React, { useState } from 'react'
+import { PasswordField, validateSingleField } from 'LoginPage'
+
+type Props = {
+  isRequired?: boolean,
+  name: string,
+  label: string,
+  autoFocus?: string
+}
+
+const PasswordInput = ({ isRequired, name, label, autoFocus }: Props) => {
+  const [value, setValue] = useState('')
+  const [isValid, setIsValid] = useState(undefined)
+
+  const inputProps = {
+    isRequired,
+    name: `user[${name}]`,
+    fieldId: `user_${name}`,
+    label,
+    value,
+    isValid,
+    onChange: (value, event) => {
+      const isValid = validateSingleField(event)
+      setValue(value)
+      setIsValid(isValid)
+    },
+    autoFocus
+  }
+
+  return (
+    <PasswordField inputProps={inputProps} />
+  )
+}
+
+export { PasswordInput }
--- a/app/javascript/src/ChangePassword/index.jsx
+++ b/app/javascript/src/ChangePassword/index.jsx
@@ -1,0 +1,2 @@
+export * from 'ChangePassword/components/ChangePassword'
+export * from 'ChangePassword/components/PasswordInput'
--- a/spec/javascripts/ChangePassword/ChangePassword.spec.jsx
+++ b/spec/javascripts/ChangePassword/ChangePassword.spec.jsx
@@ -1,0 +1,25 @@
+// @flow
+
+import React from 'react'
+import { mount } from 'enzyme'
+import { ChangePassword } from 'ChangePassword'
+
+it('should render itself', () => {
+  const props = {
+    lostPasswordToken: 'foo',
+    url: 'foo/bar',
+    errors: []
+  }
+  const wrapper = mount(<ChangePassword {...props} />)
+  expect(wrapper).toMatchSnapshot()
+})
+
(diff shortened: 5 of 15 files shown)


damianpm:
@thomasmaas font issue addressed
font

app/javascript/src/ChangePassword/index.jsx
```diff
@@ -0,0 +1,84 @@
...
+
+const ChangePassword = ({lostPasswordToken, url, errors}: Props) => {
+  const [password, setPassword] = useState('')
+  const [isPasswordValid, setIsPasswordValid] = useState(undefined)
+  const [passwordConfirmation, setPasswordConfirmation] = useState('')
+  const [isPasswordConfirmationValid, setIsPasswordConfirmationValid] = useState(undefined)
+  const formDisabled = !isPasswordValid || !isPasswordConfirmationValid
+
+  return (
+    <LoginPage
```
josemigallas: Why ChangePassword renders a LoginPage?

app/javascript/src/ChangePassword/index.jsx
```diff
@@ -0,0 +1,84 @@
+// @flow
+
+import React, {useState} from 'react'
```
josemigallas: Use curly spaces consistently, either with a space or without.

app/javascript/packs/ChangePassword.js
```diff
@@ -0,0 +1,8 @@
+import {ChangePasswordWrapper} from 'ChangePassword'
```
josemigallas: I think it's better import it as `ChangePassword` since the wrapper is just a utility to encapsulate the render function. This way when we stop using the wrapper, we don't have to change anything else, only the import sentence.

app/javascript/src/ChangePassword/index.jsx
```diff
@@ -0,0 +1,84 @@
...
+    onChange: (value, event) => {
+      const isValid = validateSingleField(event)
+      setValue(value)
+      setIsValid(isValid)
+    }
+  }
+  return <PasswordField inputProps={props} />
+}
+
+const ChangePassword = ({lostPasswordToken, url, errors}: Props) => {
```
josemigallas: Let's use index only to export components and functions and create a `ChangePassword.jsx` apart.

app/javascript/src/ChangePassword/index.jsx
```diff
@@ -0,0 +1,84 @@
...
+      <Form
+        action={url}
+        noValidate
+        id='edit_user_2'
+        acceptCharset='UTF-8'
+        method='post'
+      >
+        <input type='hidden' name='_method' value='put' />
+        <HiddenInputs />
+        {renderInputWithProps('password', 'Password', password, isPasswordValid, setPassword, setIsPasswordValid, validateSingleField)}
```
josemigallas: IMHO using this function means the PasswordField is not well defined. Also using a function to render a component is an antipattern, since components are functions. Define a new component instead or let's improve PasswordField so that it's useful and doesn't need to redefine "wrapper" functions every time.

app/javascript/src/ChangePassword/index.jsx
damianpm: Because LoginPage is the PF4  level-page-component used when not logged in:
https://www.patternfly.org/v4/documentation/react/components/loginpage

spec/javascripts/ChangePassword/ChangePassword.spec.jsx
```diff
@@ -0,0 +1,23 @@
...
+})
+
+it('should render with errors when present', () => {
+  const props = {
+    lostPasswordToken: 'foo',
+    url: 'foo/bar',
+    errors: [{type: 'error', message: 'Ooops!'}]
+  }
+  const wrapper = mount(<ChangePassword {...props} />)
+  expect(wrapper.find('.pf-c-form__helper-text').text()).toContain('Ooops!')
```
josemigallas:
```suggestion
  expect(wrapper).toMatchSnapshot()
```

app/javascript/src/ChangePassword/components/PasswordInput.jsx
```diff
@@ -0,0 +1,21 @@
...
+  name: string,
+  fieldId: string,
+  label: string,
+  value: string,
+  isValid: boolean,
+  onChange: () => void,
+  autoFocus?: 'string'
+}
+
+const PasswordInput = ({inputProps}: {inputProps: Props}) => {
```
josemigallas: I think this class would be more useful if we moved [this logic](https://github.com/3scale/porta/pull/1664/files#diff-53d51ce55ba27a9ef30bdd59eab90278R20) here instead, the Props being `getInputProps`'s arguments. WDYT?

spec/javascripts/ChangePassword/ChangePassword.spec.jsx
damianpm: I don't want to compare the snapshots, but to be sure that errors are rendered

app/javascript/src/ChangePassword/components/PasswordInput.jsx
damianpm: I think it's OK to pass just props here. The logic in `getInputProps` depends on the state of `ChangePassword` component, and I think is better to keep it separated

spec/javascripts/ChangePassword/ChangePassword.spec.jsx
josemigallas: Snapshots are generated for each set of properties. In this case the snapshot with errors will have a `pf-c-form__helper-text` and in the future is something changes, it will detect it. Same kind of situation [here](https://github.com/3scale/porta/blob/af1095b5a63b3f719f46239b4f4823dcc64c2e18/spec/javascripts/NewService/formElements/ServiceManualListItems.spec.jsx#L20).

app/javascript/src/ChangePassword/components/PasswordInput.jsx
didierofrivia: It seems that logic should be inside here instead, because you are hydrating it with your desired props anyways.

app/javascript/src/ChangePassword/components/ChangePassword.jsx
```diff
@@ -0,0 +1,87 @@
...
+import brandImg from 'LoginPage/assets/images/3scale_Logo_Reverse.png'
+import PF4DownstreamBG from 'LoginPage/assets/images/PF4DownstreamBG.svg'
+
+type Props = {
+  lostPasswordToken: ?string,
+  url: string,
+  errors: (?FlashMessage)[]
+}
+
+const getInpuProps = (name, label, value, isValid, setValue, setIsValid, validateSingleField, autoFocus) => ({
```
didierofrivia: PS: There's a typo in `getInpuProps`

spec/javascripts/ChangePassword/ChangePassword.spec.jsx
didierofrivia: Snapshots are our fastest way to assert any structure given certain props. A cost effective shield that will guard us for any other change in the component

app/javascript/src/ChangePassword/components/PasswordInput.jsx
damianpm: OK, updated

app/javascript/src/ChangePassword/components/ChangePassword.jsx
```diff
@@ -0,0 +1,74 @@
...
+  url: string,
+  errors: (?FlashMessage)[]
+}
+
+const ChangePassword = ({lostPasswordToken, url, errors}: Props) => {
+  const [password, setPassword] = useState('')
+  const [isPasswordValid, setIsPasswordValid] = useState(undefined)
+  const [passwordConfirmation, setPasswordConfirmation] = useState('')
+  const [isPasswordConfirmationValid, setIsPasswordConfirmationValid] = useState(undefined)
+  const formDisabled = !isPasswordValid || !isPasswordConfirmationValid
```
didierofrivia: `isFormDisabled`

app/javascript/src/ChangePassword/components/PasswordInput.jsx
```diff
@@ -0,0 +1,38 @@
...
+const PasswordInput = ({ isRequired, name, label, value, isValid, setValue, setIsValid, autoFocus }: Props) => {
+  const inputProps = {
+    isRequired,
+    name: `user[${name}]`,
+    fieldId: `user_${name}`,
+    label,
+    value,
+    isValid,
+    onChange: (value, event) => {
+      const isValid = validateSingleField(event)
```
didierofrivia: I know this is not from this PR, but it's weird having a fn `validateSingleField` that takes an `event` as argument instead of the field

app/javascript/src/ChangePassword/components/ChangePassword.jsx
```diff
@@ -0,0 +1,74 @@
...
+
+type Props = {
+  lostPasswordToken: ?string,
+  url: string,
+  errors: (?FlashMessage)[]
+}
+
+const ChangePassword = ({lostPasswordToken, url, errors}: Props) => {
+  const [password, setPassword] = useState('')
+  const [isPasswordValid, setIsPasswordValid] = useState(undefined)
```
didierofrivia: Maybe you could save a few lines managing the state of the field valid inside the `PasswordInput`.
Also, I might be missing where the "passwords don't match" check is computed.

app/javascript/src/ChangePassword/components/PasswordInput.jsx
damianpm: I'd address that issue as a separate task, as it will affect all other forms

app/javascript/src/ChangePassword/components/ChangePassword.jsx
```diff
@@ -0,0 +1,74 @@
...
+      <Form
+        action={url}
+        noValidate
+        id='edit_user_2'
+        acceptCharset='UTF-8'
+        method='post'
+      >
+        <input type='hidden' name='_method' value='put' />
+        <HiddenInputs />
+        <PasswordInput isRequired name='password' label='Password'
```
josemigallas: Consistency in code is important for a good readability and I will never quit fighting for this 😄. Let's put properties 1 per line and never starting in the first line, just like Form a couple of lines above.
When properties are short enough, they can be put in one line.

app/javascript/src/ChangePassword/components/ChangePassword.jsx
```diff
@@ -0,0 +1,74 @@
...
+        {lostPasswordToken &&
+          <input id='password_reset_token' type='hidden' name='password_reset_token' value={lostPasswordToken} />
+        }
+        <ActionGroup>
+          <Button className='pf-c-button pf-m-primary pf-m-block'
+            type='submit'
+            name='commit'
+            isDisabled={formDisabled}
+          >
+          Change Password
```
josemigallas: This line has a wrong indentation
```suggestion
            Change Password
```

app/javascript/src/ChangePassword/index.jsx
```diff
@@ -0,0 +1,2 @@
+export { ChangePassword, ChangePasswordWrapper } from 'ChangePassword/components/ChangePassword'
```
josemigallas:
```suggestion
export * from 'ChangePassword/components/ChangePassword'
```

app/javascript/src/ChangePassword/index.jsx
```diff
@@ -0,0 +1,2 @@
...
+export { PasswordInput } from 'ChangePassword/components/PasswordInput'
```
josemigallas:
```suggestion
export * from 'ChangePassword/components/PasswordInput'
```

spec/javascripts/ChangePassword/ChangePassword.spec.jsx
```diff
@@ -0,0 +1,25 @@
+// @flow
+
+import React from 'react'
+import { mount } from 'enzyme'
+import {ChangePassword} from 'ChangePassword'
```
josemigallas: Let's be consistent with the curly space rule: either add a space or don't.
```suggestion
import { ChangePassword } from 'ChangePassword'
```

damianpm:
@didierofrivia @josemigallas @thomasmaas 
I've added validation for matching password and password confirmation
match

app/javascript/src/ChangePassword/components/ChangePassword.jsx
```diff
@@ -0,0 +1,105 @@
...
+          name='password'
+          label='Password'
+          autoFocus='autoFocus'
+        />
+        <PasswordInput
+          isRequired
+          name='password_confirmation'
+          label='Password confirmation'
+        />
+        { !passwordsMatch &&
```
josemigallas:
```suggestion
        {!passwordsMatch &&
```

app/javascript/src/ChangePassword/components/ChangePassword.jsx
```diff
@@ -0,0 +1,105 @@
...
+      <Form
+        action={url}
+        noValidate
+        id='edit_user_2'
+        acceptCharset='UTF-8'
+        method='post'
+        onChange={onChange}
+      >
+        <input type='hidden' name='_method' value='put' />
+        <HiddenInputs />
```
josemigallas: Why not passing here `isPasswordReset`? Is that flag for anything else?

app/javascript/src/ChangePassword/components/ChangePassword.jsx
```diff
@@ -0,0 +1,105 @@
...
+          label='Password confirmation'
+        />
+        { !passwordsMatch &&
+          <div className="pf-c-form__helper-text pf-m-error" aria-live="polite">Password and password confirmation must match</div>
+        }
+        {lostPasswordToken &&
+          <input id='password_reset_token' type='hidden' name='password_reset_token' value={lostPasswordToken} />
+        }
+        <ActionGroup>
+          <Button className='pf-c-button pf-m-primary pf-m-block'
```
josemigallas:
```suggestion
          <Button
            className='pf-c-button pf-m-primary pf-m-block'
```

app/javascript/src/ChangePassword/components/ChangePassword.jsx
damianpm: That flag is need in the Password Reset form, rendered in Login Page, but not here

josemigallas: OK, although it's a bit confusing since this component has the very same name of the flag.

app/javascript/src/ChangePassword/components/ChangePassword.jsx
```diff
@@ -0,0 +1,106 @@
...
+}
+
+const ChangePassword = ({ lostPasswordToken, url, errors }: Props) => {
+  const [isFormDisabled, setIsFormDisabled] = useState(true)
+  const [passwordsMatch, setPasswordsMatch] = useState(true)
+
+  const onChange = (event) => {
+    const errors = validateForm(event.currentTarget, validationConstraints)
+
+    if (errors) {
```
didierofrivia: This whole block is a bit weird. We are validating whether there are errors or not, but then we are just asserting an specific one, that only this one will set the condition of disabling the form. Isn't there any other way that it could be more generic and avoiding this specificity? Also, it's calling the same fn again in the else case, but opposite values, that could be definitely DRY up.

app/javascript/src/ChangePassword/components/ChangePassword.jsx
```diff
@@ -0,0 +1,100 @@
...
+    equality: 'user[password]'
+  }
+}
+
+const ChangePassword = ({ lostPasswordToken, url, errors }: Props) => {
+  const [isFormDisabled, setIsFormDisabled] = useState(true)
+  const [passwordsMatch, setPasswordsMatch] = useState(true)
+
+  const onChange = (event) => {
+    const errors = validateForm(event.currentTarget, validationConstraints)
```
damianpm: @didierofrivia you were right, I've simplified this a bit

damianpm:
@thomasmaas 
Error messages are displayed in separate places because PF4 Input Element comes with it's own hint messages.
About the typing, what do you suggest, validate until the user clicks on submit button?

thomasmaas:
But both of those are validation errors: required and ‘needs to be the same’ so I expect them to show in same place. 

I would prefer no validation until submit indeed.

app/javascript/src/ChangePassword/components/ChangePassword.jsx
```diff
@@ -0,0 +1,92 @@
...
+    presence: true,
+    length: { minimum: 1 }
+  },
+  'user[password_confirmation]': {
+    presence: true,
+    length: { minimum: 1 }
+  }
+}
+
+const ChangePassword = ({ lostPasswordToken, url, errors }: Props) => {
```
didierofrivia: This I'll let it for another PR, but the `ChangePassword` component could be agnostic and independent component. Meaning we can reuse it in different parts of the app, like the ones in  https://github.com/3scale/porta/tree/master/app/javascript/src/Form

PR #1576, opened 2020-01-24
[Rails5] [TEST] javascript cms switch

HemaHG:
Test for JavaScript cms switch

PR #1665, opened 2020-02-10
[rails] Rails 5.0 upgrade

hallelujah:
Closes https://issues.redhat.com/browse/THREESCALE-1606


Gemfile.base
```diff
@@ -10,12 +10,14 @@ gem 'aws-sdk', '~> 2'
 gem 'aws-sdk-rails', '~> 1.0'
 
 gem 'dotenv-rails', '~> 2.7'
-gem 'rails', '4.2.11.1'
-#gem 'rails', '5.0.6'
-#gem 'activesupport', '4.2.9' # somehow paperclip needs to explicitly lock the version to not use 5.0
+#gem 'rails', '4.2.9'
```
guicassolato: Let's be optimistic 😉

```suggestion
```

Rakefile
```diff
@@ -14,6 +14,25 @@ begin
 rescue LoadError
 end
 
+require 'rails/test_unit/runner'
+class Rails::TestUnit::Runner
+  class << self
+    prepend(Module.new do
+
+      # This is a private method, expect this to be broken in the future
```
Martouta: This is fine

TODO-rails-5.md
```diff
@@ -0,0 +1,16 @@
+# List of TODO encountered in the way
+
+## TODO list
+
+* Remove lib/rails5_ssl.rb
+* Check bigint primary while migrating to rails 5.1
+* secure_headers is a rack with some helpers to work on controller. Fix set_x_frame_options and the like
```
Martouta: https://github.com/3scale/porta/pull/1551

Gemfile.base
Martouta: I agree with Gui, besides it feels strange that it is not even the version that we have been using (`4.2.11.1`) but a previous one.

TODO-rails-5.md
```diff
@@ -0,0 +1,16 @@
+# List of TODO encountered in the way
+
+## TODO list
```
Martouta: Personally I prefer having this in Jira just to don't ignore it :smile:

bin/setup
```diff
@@ -0,0 +1,36 @@
...
+  # Add necessary setup steps to this file.
+
+  puts '== Installing dependencies =='
+  system! 'gem install bundler --conservative'
+  system('bundle check') || system!('bundle install')
+
+  # puts "\n== Copying sample files =="
+  # unless File.exist?('config/database.yml')
+  #   cp 'config/database.yml.sample', 'config/database.yml'
+  # end
```
guicassolato: Can we remove this?

config/cable.yml
```diff
@@ -0,0 +1,9 @@
+development:
```
guicassolato: This file should be ignored, no?

Rakefile
```diff
@@ -22,32 +41,35 @@ namespace :test do
...
+  #     puts "TESTS='"
+  #     t.file_list.each do |file|
+  #       print "\t", file, "\n"
+  #     end
+  #     puts "'"
+  #   end
+  #
+  #   t.verbose = verbose
+  #   t.loader = :direct
+  # end
```
guicassolato: Remove?

config/cable.yml
hallelujah: I will clean later if you do not mind

app/models/plan.rb
```diff
@@ -26,7 +27,7 @@ class PeriodRangeCalculationError < StandardError; end
   #  - `#audit_destroy` triggered by audit_destroy
   # I think those gems are fine with their behaviour.
   # The problem is on our side with so many inter-dependent relationships
-  skip_callback :destroy, :before, :lock!, if: -> { !self.class.exists?(id) || destroyed_by_association }
+  # skip_callback :destroy, :before, :lock!, if: -> { !self.class.exists?(id) || destroyed_by_association }
```
Martouta: I am not sure if this is entirely correct. Although it seems so, but I prefer to express my concerns to be sure :smile: 

Part of this has a test in master:
https://github.com/3scale/porta/blob/e2dbb0cf87ed7a340261f681b9444254dd1a199b/test/unit/plan_test.rb#L603-L615
And this is passing :heavy_check_mark: :+1: 

And locally I added a test for the part that was untested there (I am not sure if it is tested somewhere else, but I think not. It was added in https://github.com/3scale/system/pull/9244) :
```ruby
def test_plan_not_locked_if_destroyed_by_association
  plan = FactoryBot.create(:simple_application_plan)
  plan.class.stubs(:exists?).with(plan.id).returns(true)
  plan.class.stubs(:destroyed_by_association).with(plan.id).returns(true)
  plan.expects(:lock!).never
  plan.destroy
end
```

And this also passes when I do `bundle exec rake test TEST=test/unit/plan_test.rb`
So my conclusions are:
1. We should add this test, because it is a scenario that matters.
2. Does this test run callbacks? because otherwise, this is passing for the wrong reasons :joy:

app/models/service.rb
```diff
@@ -217,6 +217,7 @@ def self.all
   def using_proxy_pro?
     provider_can_use?(:proxy_pro) && proxy.self_managed?
   end
+  delegate :service_preffix, to: :provider, allow_nil: true
```
Martouta: I do not understand :confused: 
https://github.com/3scale/porta/blob/e2dbb0cf87ed7a340261f681b9444254dd1a199b/app/models/account.rb#L6-L8

config/cable.yml
Martouta: > This file should be ignored, no?

Regardless of being ignored or not, we already have it inside examples so imo we should do like for database.yml that it is only inside examples but inside config directly is in .gitignore so everybody can customize it was we wish.

> I will clean later if you do not mind

Sure, no problem :+1:

test/integration/finance/charging_invoice_states_test.rb
```diff
@@ -47,8 +47,8 @@ class Finance::ChargingInvoiceStatesTest < ActionDispatch::IntegrationTest
 
       @invoice.buyer_account.bought_plan.update_attributes!(name: 'Paid', cost_per_month: 100)
 
-      ThreeScale::Analytics.expects(:track).with(instance_of(User), 'Charged Invoice',
-                                                 {:plan => 'Paid', :period => 'January  1, 1984 - January 31, 1984', :revenue => 100.0})
+      ThreeScale::Analytics.expects(:track).with(@provider.first_admin, 'Charged Invoice',
```
Martouta:
```suggestion
      ThreeScale::Analytics.expects(:track).with(@provider.first_admin!, 'Charged Invoice',
```

Simply to get a most understandable error if there isn't any because we expect to have it and not to be nil and track nil :smile: 

https://github.com/3scale/porta/blob/e2dbb0cf87ed7a340261f681b9444254dd1a199b/app/models/account.rb#L193-L199

test/integration/finance/charging_invoice_states_test.rb
```diff
@@ -47,8 +47,8 @@ class Finance::ChargingInvoiceStatesTest < ActionDispatch::IntegrationTest
...
+                                                 {:plan => 'Paid', :period => 'January 01, 1984 - January 31, 1984', :revenue => 100.0})
```
Martouta:
```suggestion
                                                 {plan: 'Paid', period: 'January 01, 1984 - January 31, 1984', revenue: 100.0})
```

neither mandatory, nor urgent :wink:

test/unit/backend/random_data_generator_test.rb
```diff
@@ -1,7 +1,7 @@
+# frozen_string_literal: true
```
Martouta: After this line, there should be an empty line :smile: 
But I wouldn't request you to push a new commit only for that :smile: :+1:

test/unit/messengers/cinstance_messenger_test.rb
```diff
@@ -183,7 +183,8 @@ def test_plan_change
       assert messenger.deliver
       assert provider.received_messages.empty?
 
-      message = master.received_messages.last!
+      # Using first! as they are ordered by created_at desc
```
Martouta: This table does not have a `created_at` so this comment kills me :joy: 
https://github.com/3scale/porta/blob/e2dbb0cf87ed7a340261f681b9444254dd1a199b/db/schema.rb#L799-L809

test/unit/messengers/cinstance_messenger_test.rb
```diff
@@ -183,7 +183,8 @@ def test_plan_change
...
+      message = master.received_messages.first!
```
Martouta: image

result of `bundle exec rake test TEST=test/unit/messengers/cinstance_messenger_test.rb TESTOPS="--name= test_plan_change -v"`:
image

And it is true that the `created_at` does not exist for that table, so this comment is wrong.
https://github.com/3scale/porta/blob/e2dbb0cf87ed7a340261f681b9444254dd1a199b/db/schema.rb#L799-L809

And if right before this line I add a `pp master.received_messages.count`, there is only 1 (with mysql, I didn't try with other DBs) so I do not understand why this change but I assume it is for other DBs.

However, let's remember that for Postgres the order is 'random' and that there is ever more than 1, we will get random failures in circleci... so we should set the order manually for example with `message = master.received_messages.order(id: :desc).first!` in order to don't drive us too crazy if we ever see a circleci error in postgres here.

test/unit/services/cms_reset_service_test.rb
```diff
@@ -16,6 +16,7 @@ def test_call
     page = provider.pages.first!
 
     assert @cms_reset.call(provider)
+    provider.reload
```
Martouta: it could also be in the next line as `refute_equal page, provider.reload.pages.first!`, but either way it is fine :smile:

test/unit/services/messages/destroy_all_service_test.rb
```diff
@@ -12,28 +12,33 @@ def setup
...
+      @message.reload
+
+      assert_not_equal nil, @message.reload.deleted_at
     end
+  end
 
-    assert_raise(ActiveRecord::RecordNotFound) { @message.reload }
+  def test_run_with_sidekiq_job
+    perform_enqueued_jobs do
+      Sidekiq::Testing.inline! do
```
Martouta: This is unnecessary now.
Please check this PR to understand why:
https://github.com/3scale/porta/pull/1640

Because this worker now inherits from ApplicationJob now.

And that PR affects this whole test, not just this line.

test/unit/concerns/symbolize_test.rb
```diff
@@ -0,0 +1,82 @@
...
+      self.plan_type ||= 'Plan'
+    end
+  end
+
+  def setup
+    Timecop.freeze
+  end
+
+  def teardown
+    Timecop.return
```
Martouta: Ok, good :+1: I see that the teardown is executed even when there is an error in a test.

image

image

test/unit/cinstance_test.rb
```diff
@@ -1,5 +1,8 @@
 require 'test_helper'
 
+# TODO: Please split this file. It is too huge and takes too long
```
Martouta: I agree :+1: And we should eventually create a jira for it.

test/unit/service_test.rb
```diff
@@ -146,7 +146,7 @@ def test_update_account_default_service
 
   test 'service_token' do
     account = FactoryBot.create(:simple_account)
-    service = Service.new { |s| s.account_id = account.id; s.system_name = 'foo' }
+    service = Service.new { |s| s.account = FactoryBot.create(:simple_account); s.system_name = 'foo' }
```
Martouta: I do not understand what is the previous account created for then (the previous line), or at least I don't see why we save it in a variable that we never use.

test/unit/application_keys_test.rb
```diff
@@ -159,11 +158,10 @@ def test_format
   def test_regenerate_key
     @application_keys.add(key = 'app-key')
     @application.reload
+    Timecop.travel(2.hours.from_now)
     updated_at = @application.updated_at
     @application_keys.regenerate(key)
-    # @application.reload
     assert_not_equal updated_at, @application.updated_at
-
   end
```
Martouta: I propose something like this:

```ruby
def test_regenerate_key
  @application_keys.add(key = 'app-key')
  @application.reload
  updated_at = @application.updated_at
  Timecop.travel(2.hours.from_now) do
    @application_keys.regenerate(key)
  end
  assert_in_delta (updated_at + 2.hours),  @application.updated_at, 1.second
end
```

For the following reasons:
1. Using Timecop without a block and without a return in the teardown scares me because then if there is an error somewhere after the timecop, it might stay like that for the rest of the tests.
2. Using assert_not_equal or expects(:foo).never is ok when we do not really have a choice, but we can assert the exact value that we expect, it is better (less error-prone). Because otherwise it could pass the assert_not_equal for the wrong reasons (like 1 second of different when running the code)
3. Since we are gonna assert_equal, let's keep in mind the possible 1 second of margin of running the code from saving the previous updated_at to making the new update

test/unit/application_keys_test.rb
```diff
@@ -159,11 +158,10 @@ def test_format
   def test_regenerate_key
     @application_keys.add(key = 'app-key')
     @application.reload
+    Timecop.travel(2.hours.from_now)
     updated_at = @application.updated_at
     @application_keys.regenerate(key)
-    # @application.reload
```
Martouta: Good removing this, we do not need it because it is the same instance so it is reloaded

config/initializers/mysql.rb
```diff
@@ -1,15 +1,19 @@
 # frozen_string_literal: true
 
-if defined?(ActiveRecord::ConnectionAdapters::Mysql2Adapter)
+if System::Database.mysql?
+  # This is for Rails 4, I do not know if it is needed
```
Martouta: we should create a jira to figure this out

app/models/cms/email_template.rb
```diff
@@ -82,7 +82,7 @@ def mime_type
 
   def headers=(val)
     @headers = nil
-    self[:options] = val
+    self[:options] = val.respond_to?(:to_unsafe_h) ? val.to_unsafe_h : val
```
Martouta: Although it isn't exactly the same, another option could be (depending on our case, i dunno if this is good or not):
```ruby
self[:options] = val.try(:to_unsafe_h) || val
```
which is actually equivalent to:
```ruby
self[:options] = val.respond_to?(:to_unsafe_h) ? (val.to_unsafe_h || val) : val
```

config/initializers/after_commit_queue_hacks.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+AfterCommitSubscriber::AfterCommitCallback.class_eval do
+  def before_committed!
+    # noop
+  end
+end
```
Martouta: no idea why but i will just believe you on this one :joy:

config/initializers/filter_parameter_logging.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+# Be sure to restart your server when you modify this file.
+
+# Configure sensitive parameters which will be filtered from the log file.
+Rails.application.config.filter_parameters += [:password, :credit_card]
```
Martouta: This list used to be like this in the past but it is much more extense now:
https://github.com/3scale/porta/blob/e2dbb0cf87ed7a340261f681b9444254dd1a199b/config/application.rb#L224-L228

And imo we should define the whole list only once and not twice.

test/integration/admin/cms/switches_controller_test.rb
```diff
@@ -31,14 +31,4 @@ def setup
     assert_response :success
     assert provider.settings.switches[switch_name].visible?
   end
-
-  test "show a  switch" do
```
Martouta: This test was added as an extraction of the Rails 5 branch in https://github.com/3scale/porta/pull/1576, and now it is removed again... :confused: I am very confused with this.

test/unit/services/cms_reset_service_test.rb
Martouta: Btw, this could also be an opportunity to ensure not having the postgres random order problem :smile:

test/unit/messengers/cinstance_messenger_test.rb
guicassolato: Order is ensured by https://github.com/3scale/porta/blob/e2dbb0cf87ed7a340261f681b9444254dd1a199b/app/models/account.rb#L166

Try in the console `MessageRecipient.latest_first`. It uses the `Message`'s `created_at`.

Martouta: Ahaaa, I see :+1: Sorry :+1: Then imo, we should either remove that message or mention that it is the created_at of its message and not its own created_at :smile: Anyway, it is clear now that it isn't a bug so thanks :+1:

app/models/cms/email_template.rb
hallelujah: I think I prefer the `respond_to?` version
Normally this `val` is an `ActionController::Parameters`
A good way to do it is:


```suggestion
self[:options] = val.to_unsafe_h
rescue NoMethodError
  ActiveSupport::Deprecation.warn('Passing a value that does not respond to `to_unsafe_h` to #headers= ')
 val
```

Though I do not want to do that, we might have tons of them ...

app/models/plan.rb
hallelujah: All callbacks since rails 5.0 are called in tests

app/models/cms/email_template.rb
Martouta: good then :+1: :wink:

app/models/service.rb
hallelujah: Right it should be removed 👍

app/models/plan.rb
Martouta: By default yes, but sometimes we change that for some tests :smile: But I see that it isn't the case :+1: Good :+1: I will add the tests in a separate PR.

Martouta: https://github.com/3scale/porta/pull/1666 :smile:

config/initializers/after_commit_queue_hacks.rb
hallelujah: It is because all callbacks class should have `before_committed!` method otherwise Rails will complain but it is already extracted https://github.com/3scale/porta/commit/eebaa23a3228ec75025076d5d34979937a6a492b

test/integration/admin/cms/switches_controller_test.rb
hallelujah: That is why your review is valuable :)

test/unit/application_keys_test.rb
hallelujah: Agree! Though I barely touched this test, only make time travels.

About the teardown stuff 

https://github.com/3scale/porta/blob/e2dbb0cf87ed7a340261f681b9444254dd1a199b/test/test_helper.rb#L58-L61

test/unit/backend/random_data_generator_test.rb
hallelujah:
```suggestion
# frozen_string_literal: true

```

test/unit/service_test.rb
hallelujah: I should have removed it
```suggestion
    service = Service.new { |s| s.account = account; s.system_name = 'foo' }
```

test/unit/services/cms_reset_service_test.rb
hallelujah: Yeah ... but this problem I think we should track it  in JIRA. Maybe there are some options for Postgres to have a default order

PR #1666, opened 2020-02-11
Test Plan not locked if destroyed_by_association

Martouta:
I saw the scenario of this test is lacking when I was reviewing the Rails 5 PR.


Martouta:
I guess that we do not need it anymore :)

PR #1671, opened 2020-02-12
4113-audit logs

macejmic:
https://issues.redhat.com/browse/THREESCALE-4113


db/migrate/20200211080911_add_index_to_audits_auditable_id_auditable_type_version.rb
```diff
@@ -0,0 +1,5 @@
+class AddIndexToAuditsAuditableIdAuditableTypeVersion < ActiveRecord::Migration
+  def change
+    add_index :audits, [:auditable_id, :auditable_type, :version]
```
guicassolato: This can be a bit painful in Saas.

db/oracle_schema.rb
```diff
@@ -503,14 +503,6 @@
   add_index "deleted_objects", ["object_type", "object_id"], name: "index_deleted_objects_on_object_type_and_object_id"
   add_index "deleted_objects", ["owner_type", "owner_id"], name: "index_deleted_objects_on_owner_type_and_owner_id"
 
-  create_table "end_user_plans", force: :cascade do |t|
```
guicassolato: Oops. This is not part of this PR. I'll fix it.

app/models/user.rb
```diff
@@ -32,7 +32,8 @@ def self.columns
     [:notification_preferences, { action: :delete, class_name: 'NotificationPreferences', has_many: false }]
   ].freeze
 
-  audited
+  audited except: %i[salt posts_count janrain_identifier cas_identifier password_digest
+                     authentication_id open_id last_login_at last_login_ip].freeze
```
guicassolato: Should we add `crypted_password` here?

macejmic: we could. will update it.

PR #1672, opened 2020-02-12
add line for MacOS 10.15 in relation to mysql gem

thomasmaas:


INSTALL.md
```diff
@@ -163,6 +163,11 @@ cp config/examples/* config/
 
 #### Bundle
 
+On MacOS 10.15 first configure the bundle config with:
+```shell
+bundle config --global build.mysql2 "--with-opt-dir=/usr/local/opt/openssl"
```
Martouta: Good :+1: But I think that we also should remove what comes right after this speaking about this error.

thomasmaas: But that is still valid for MacOS 10.13-10.14 o?

Martouta: ah, right, sorry!

PR #1678, opened 2020-02-12
Create a loader to deserialize both Hashes and

duduribeiro:
ActionController::Parameter

In Rails5 ActionController::Parameter does not inherits from Hash
anymore, this is causing errors in our application since we have a lot
of stored ActionController::Parameters serialized in our cms_template
table. This commit creates a custom loader that can serialize /
deserialize both hashes and ActionController::Parameters


Martouta:
Closing in favor of https://github.com/3scale/porta/pull/1679

PR #1679, opened 2020-02-12
[hotfix] Allow ActionController::Parameters for email template heades

hallelujah:
Fixes [THREESCALE-4467](https://issues.redhat.com/browse/THREESCALE-4467)
Alternative to https://github.com/3scale/porta/pull/1678

### Issues

I think we are only fixing this bug but not fixing the root cause of the bug
We should pass Hash or HashWithIndifferentAccess to options and deprecate the usage of ActionController::Parameters

But that means, we need to make to cleanup the Database  ....

### DIFF
--- a/app/models/cms/email_template.rb
+++ b/app/models/cms/email_template.rb
@@ -50,7 +50,25 @@ class CMS::EmailTemplate < CMS::Template
   EMAIL_FORMAT = /(?:(#{EMAIL_ADDRESS_FORMAT})|#{EMAIL_WITH_NAME_FORMAT})\s*/i
   EMAILS_FORMAT = /\A#{EMAIL_FORMAT}(,\s*#{EMAIL_FORMAT})*\Z/i
 
-  serialize :options, Hash
+  class HashOrParameters < ::ActiveRecord::Coders::YAMLColumn
+    def load(string)
+      obj = super(string)
+      obj.respond_to?(:to_unsafe_h) ? obj.to_unsafe_h : obj.to_h if obj
+    end
+
+    def dump(object)
+      return object if object.is_a?(String) && assert_valid_value(load(object))
+      obj = object.respond_to?(:to_unsafe_h) ? object.to_unsafe_h : object.to_h if object
+      super(obj)
+    end
+
+    def assert_valid_value(obj)
+      obj.is_a?(Hash) || obj.is_a?(ActionController::Parameters) || super
+    end
+  end
+
+  # TODO: On Rails 5.1+, change in argument contructor ::ActiveRecord::Coders::YAMLColumn.new(attr_name, class_or_coder)
+  serialize :options, HashOrParameters.new(Hash)
 
   attr_accessor :file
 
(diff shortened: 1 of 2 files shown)


app/models/cms/email_template.rb
```diff
@@ -50,7 +50,25 @@ def self.reset_templates_path!
...
 
-  serialize :options, Hash
+  class HashOrParameters < ::ActiveRecord::Coders::YAMLColumn
+    def load(string)
+      obj = super(string)
+      obj.respond_to?(:to_unsafe_h) ? obj.to_unsafe_h : obj.to_h if obj
+    end
+
+    def dump(object)
+      return object if object.is_a?(String) && assert_valid_value(load(object))
```
hallelujah: In fact this line does a lot more than needed
Now `options=` can take a `String` as argument. Before, it rejected anything.

I made this possible because of this test https://github.com/3scale/porta/pull/1679/files#diff-00583319eebb023a454123e16fbd9655R51

Without it I cannot use `update_column :options, '--- some yaml'`, because Rails always tries to serialize the attribute ...
And there is no way I can do it simply, without using low level Rails API ...

app/models/cms/email_template.rb
```diff
@@ -50,7 +50,25 @@ def self.reset_templates_path!
   EMAIL_FORMAT = /(?:(#{EMAIL_ADDRESS_FORMAT})|#{EMAIL_WITH_NAME_FORMAT})\s*/i
   EMAILS_FORMAT = /\A#{EMAIL_FORMAT}(,\s*#{EMAIL_FORMAT})*\Z/i
 
-  serialize :options, Hash
+  class HashOrParameters < ::ActiveRecord::Coders::YAMLColumn
+    def load(string)
+      obj = super(string)
+      obj.respond_to?(:to_unsafe_h) ? obj.to_unsafe_h : obj.to_h if obj
```
hallelujah: This is to load serialized `ActionController::Parameters` as `ActiveSupport::HashWithIndiffferentAccess`

hallelujah: Maybe it should be `.with_indifferent_access` instead of `.to_h`

@3scale/system-ruby thoughts? 🤔

app/models/cms/email_template.rb
```diff
@@ -50,7 +50,25 @@ def self.reset_templates_path!
...
+    end
+
+    def dump(object)
+      return object if object.is_a?(String) && assert_valid_value(load(object))
+      obj = object.respond_to?(:to_unsafe_h) ? object.to_unsafe_h : object.to_h if object
```
hallelujah: Now it is serialized either as `ActionController::Parameters` or `Hash`

app/models/cms/email_template.rb
duduribeiro: I agree with the `with_indifferent_access`

app/models/cms/email_template.rb
duduribeiro: yeah.. but we should find where is the saving point that is storing the serializer as ActionController::Parameters

PR #1685, opened 2020-02-13
[Bugfix] Ensure that all associations of the BackgroundDeletion config can be constantized

Martouta:
This is necessary to be able to do [THREESCALE-4483](https://issues.redhat.com/browse/THREESCALE-4483)
I encountered when testing that Jira issue in preview and seeing:
> DeleteAccountHierarchyWorker Account->associations NameError: uninitialized constant 'class_name'

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -60,22 +60,22 @@ class Account < ApplicationRecord
   self.background_deletion = [
     :users,
     :mail_dispatch_rules,
-    [:api_docs_services, class_name: 'ApiDocs::Service'],
+    [:api_docs_services, { class_name: 'ApiDocs::Service' }],
     :services,
     :contracts,
     :account_plans,
-    [:settings, { action: :destroy, has_many: false }],
+    [:settings, { action: :destroy, class_name: 'Settings', has_many: false }],
     [:payment_detail, { action: :destroy, has_many: false }],
     [:buyer_accounts, { action: :destroy, class_name: 'Account' }],
     [:payment_gateway_setting, { action: :destroy, has_many: false }],
     [:profile, { action: :delete, has_many: false }],
-    [:templates, { action: :delete }],
+    [:templates, { action: :delete, class_name: 'CMS::Template' }],
     [:sections, { action: :delete, class_name: 'CMS::Section' }],
     [:provided_sections, { action: :delete, class_name: 'CMS::Section' }],
-    [:redirects, { action: :delete }],
-    [:files, { action: :delete }],
-    [:builtin_pages, { action: :delete }],
-    [:provided_groups, { action: :delete }]
+    [:redirects, { action: :delete, class_name: 'CMS::Redirect' }],
+    [:files, { action: :delete, class_name: 'CMS::File' }],
+    [:builtin_pages, { action: :delete, class_name: 'CMS::BuiltinPage' }],
+    [:provided_groups, { action: :delete, class_name: 'CMS::Group' }]
   ].freeze
 
   #TODO: this needs testing?
--- a/test/unit/concerns/background_deletion_test.rb
+++ b/test/unit/concerns/background_deletion_test.rb
@@ -40,10 +40,9 @@ module Concerns
     test 'all associations of the config can be constantized' do
       Rails.application.eager_load!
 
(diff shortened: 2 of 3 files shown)


app/models/account.rb
```diff
@@ -69,13 +69,13 @@ def self.columns
     [:buyer_accounts, { action: :destroy, class_name: 'Account' }],
     [:payment_gateway_setting, { action: :destroy, has_many: false }],
     [:profile, { action: :delete, has_many: false }],
-    [:templates, { action: :delete }],
+    [:templates, { action: :delete, class_name: 'CMS::Template' }],
```
hallelujah: Good catch

Martouta: Gold! Thanks! :clap:

Martouta: I've just pushed it applying it to all models :+1:

test/unit/concerns/background_deletion_test.rb
```diff
@@ -36,5 +36,20 @@ def test_delete_integration
       DeleteObjectHierarchyWorker.expects(:perform_later).with(user, anything, 'destroy').never
       DeleteObjectHierarchyWorker.perform_now(double_object)
     end
+
+    test 'all associations of the config can be constantized' do
+      Rails.application.eager_load!
+
+      models = ActiveRecord::Base.descendants - [DoubleObject, DoubleActiveRecordObject, SecondDoubleActiveRecordObject]
```
hallelujah:
```suggestion
      models = ActiveRecord::Base.descendants.select {|model| model.background_deletion.any? } 
```

test/unit/concerns/background_deletion_test.rb
```diff
@@ -36,5 +36,20 @@ def test_delete_integration
...
+
+      models.each do |klass|
+        next unless klass.included_modules.include? BackgroundDeletion
```
hallelujah: They will always include it no? Because it is included in ApplicationRecord

Martouta: Ouch, right!
image

test/unit/concerns/background_deletion_test.rb
Martouta: @hallelujah That wouldn't work because I am removing only the ones that are part of this test file, so it should be then:
models = `ActiveRecord::Base.descendants.select {|model| model.background_deletion.present? && ![DoubleObject, DoubleActiveRecordObject, SecondDoubleActiveRecordObject].include?(model) } `

However, I am not sure if it is worth to do a loop here for filtering and later another loop to treat it. Probably better just do in https://github.com/3scale/porta/pull/1685#discussion_r380162089
directly:
```ruby
ActiveRecord::Base.descendants.each do |klass|
  next if DoubleObject, DoubleActiveRecordObject, SecondDoubleActiveRecordObject].include?(model) || model.background_deletion.blank?
# ...
```

test/unit/concerns/background_deletion_test.rb
```diff
@@ -36,5 +36,19 @@ def test_delete_integration
       DeleteObjectHierarchyWorker.expects(:perform_later).with(user, anything, 'destroy').never
       DeleteObjectHierarchyWorker.perform_now(double_object)
     end
+
+    test 'all associations of the config can be constantized' do
+      Rails.application.eager_load!
+
+      ActiveRecord::Base.descendants.each do |klass|
+        next if [DoubleObject, DoubleActiveRecordObject, SecondDoubleActiveRecordObject].include?(klass) || !klass.respond_to?(:background_deletion)
```
Martouta: Please read these 2 comments/threads:
- https://github.com/3scale/porta/pull/1685#discussion_r380169618
- https://github.com/3scale/porta/pull/1685#discussion_r380170686

Btw I ended up doing `!klass.respond_to?(:background_deletion)` instead of `klass.background_deletion.blank?` for this:
image

hallelujah: You can use `ApplicationRecord.descendants` instead of using ActiveRecord::Base 🤕

hallelujah: Yes OK only one loop using `ApplicationRecord.descendants`  👍

Martouta: Done :sunglasses: :dancer:

PR #1688, opened 2020-02-13
Allow reset password in master provider

duduribeiro:
We currently have a bug in production that does not allow the user to
reset its password in the master provider. It returns a `Wrong domain
for pat /p/password` error.

### DIFF
--- a/app/controllers/provider/passwords_controller.rb
+++ b/app/controllers/provider/passwords_controller.rb
@@ -51,7 +51,7 @@ class Provider::PasswordsController < FrontendController
   # Can't be used for neither Buyer nor Master
   #
   def find_provider
-    unless @provider ||= Account.providers.find_by_self_domain(request.host.to_s)
+    unless @provider ||= Account.providers_with_master.find_by_self_domain(request.host.to_s)
       render_error "Wrong domain '#{request.host}' for path '#{request.path}'", status: 404
       false
     end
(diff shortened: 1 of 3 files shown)


app/controllers/provider/passwords_controller.rb
```diff
@@ -51,7 +51,7 @@ def update
   # Can't be used for neither Buyer nor Master
   #
   def find_provider
-    unless @provider ||= Account.providers.find_by_self_domain(request.host.to_s)
+    unless @provider ||= Account.providers_with_master.find_by_self_domain(request.host.to_s)
```
Martouta: Good catch! :clap:

test/functional/provider/passwords_controller_test.rb
```diff
@@ -41,4 +41,11 @@ def setup
     assert_redirected_to back
   end
 
+  test '#destroy also works for master account' do
```
Martouta: The functional tests (inheriting from `ActionController::TestCase`) are deprecated and we definitely should not add new ones. This test should be inside this other class:
https://github.com/3scale/porta/blob/185db8beaffcfb6ea6db5334d045b2bcb00525f6/test/integration/provider/passwords_test.rb#L3

And should look more or less like this:
```ruby
test '#destroy also works for master account' do
  user = FactoryBot.create(:admin, account: master_account)
  login! master_account

  delete provider_password_path(email: user.email)
  assert_response :found
  assert_match 'A password reset link has been emailed to you.', flash[:notice]
end
```

hallelujah:
I think we did not want to have this ability on master see issue and comments on https://issues.redhat.com/browse/THREESCALE-303

If the intention was to have it please open the correct issue.

PR #1690, opened 2020-02-14
Fix Deadlock destroying plans by association without updating the position within the service

Martouta:
This is necessary to do [THREESCALE-4483](https://issues.redhat.com/browse/THREESCALE-4483)
It fixes the deadlock described in [this comment](https://issues.redhat.com/browse/THREESCALE-4483?focusedCommentId=13971719&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13971719).

We need to update the position of the plans within an issuer when we remove a plan, however, when we are gonna remove the issuer anyway (which means when the plan is destroyed_by_association), we do not need to do that update.

### DIFF
--- a/app/models/plan.rb
+++ b/app/models/plan.rb
@@ -479,6 +479,18 @@ class Plan < ApplicationRecord
 
   private
 
+  def scheduled_for_deletion?
+    !issuer || issuer.deleted? || provider_account&.scheduled_for_deletion?
+  end
+
+  # act_as_list updates the position every time that a plan is destroyed.
+  # But we do not want to do that when its issuer is going to be deleted anyway.
+  # Additionally, we cannot do: skip_callback :destroy, :after, :decrement_positions_on_lower_items, if: -> { destroyed_by_association }
+  # because it is an 'after' callback and by the time it is executed, destroyed_by_association is already 'nil'
+  def act_as_list_no_update?
+    super || scheduled_for_deletion?
+  end
+
   def plan_rule
     @plan_rule ||= PlanRulesCollection.find_for_plan(self)
   end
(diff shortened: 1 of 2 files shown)


app/models/plan.rb
```diff
@@ -479,6 +479,20 @@ def clone_associations(custom)
 
   private
 
+  # act_as_list updates the position every time that a plan is destroyed.
+  # But we do not want to do that when its issuer is going to be deleted anyway.
+  # Additionally, we cannot do: skip_callback :destroy, :after, :decrement_positions_on_lower_items, if: -> { destroyed_by_association }
+  # because it is an 'after' callback and by the time it is executed, destroyed_by_association is already 'nil'
+  def decrement_positions_on_lower_items
```
guicassolato: Perhaps it could be done as in

https://github.com/3scale/porta/blob/d3a59a0164713e681138e808a36f0ec05cf45334/app/models/proxy_rule.rb#L129-L135

hallelujah: Agree :)

app/models/plan.rb
```diff
@@ -479,6 +479,19 @@ def clone_associations(custom)
 
   private
 
+  def scheduled_for_deletion?
+    # try(:account) because issuer could not be a service and could not respond to 'account'
+    !issuer || issuer.deleted? || ((account = issuer.try(:account)) && account.scheduled_for_deletion?)
```
makikata: It think this comment is misleading, looking at the code, polymorphic relation `as: :issuer` is only set on Service and Account which all respond to `#account`

If the issuer is an `Account` then it is a `Provider`, so `issuer.account == Account.master`
If the issuer is a `Service` then `issuer.account` is a `Provider`

But all `Plan` respond_to `provider_account`

```suggestion
    !issuer || issuer.deleted? ||  provider_account&.scheduled_for_deletion?
```

PR #1691, opened 2020-02-14
Update schema files based on latest versions of the gems

guicassolato:
- [x] mysql
- [x] postgres
- [x] oracle

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -1,4 +1,3 @@
-# encoding: UTF-8
 # This file is auto-generated from the current state of the database. Instead
 # of editing this file, please use the migrations feature of Active Record to
 # incrementally modify your database, and then regenerate this schema definition.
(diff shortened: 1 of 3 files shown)


db/schema.rb
```diff
@@ -13,1276 +12,1213 @@
 
 ActiveRecord::Schema.define(version: 20200121142649) do
 
-  create_table "access_tokens", force: :cascade do |t|
-    t.integer  "owner_id",   limit: 8,                      null: false
-    t.string   "owner_type", limit: 255,   default: "User", null: false
+  create_table "access_tokens", id: :bigint, force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_bin" do |t|
+    t.bigint   "owner_id",                                  null: false
```
hallelujah: Bufff do we have this in SaaS too?
If so we should fix it

db/schema.rb
```diff
@@ -13,1276 +12,1213 @@
...
+    t.string   "owner_type",               default: "User", null: false
     t.text     "scopes",     limit: 65535
-    t.string   "value",      limit: 255,                    null: false
-    t.string   "name",       limit: 255,                    null: false
-    t.string   "permission", limit: 255,                    null: false
-    t.integer  "tenant_id",  limit: 8
```
hallelujah: Same https://github.com/3scale/porta/pull/1691/files#r380013269

PR #1693, opened 2020-02-14
[memory-leak] Do not regenerate routes every requests

hallelujah:
Fixes THREESCALE-3397

Big thanks to @guicassolato  who made a deep search on this memory leak
He found out that the routes helpers were increasing every requests

Also read rails/rails#32304 (comment)

### DIFF
--- a/app/models/sso_token.rb
+++ b/app/models/sso_token.rb
@@ -2,7 +2,6 @@ class SSOToken
   include ActiveModel::Validations
   include ActiveModel::MassAssignmentSecurity
   include ActiveModel::Serializers::Xml
-  include Rails.application.routes.url_helpers
 
   attr_accessor   :user_id, :username, :expires_in, :redirect_url, :protocol, :account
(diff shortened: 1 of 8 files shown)


hallelujah:
Left is to clean the use of `Rails.application.routes.url_helpers.send(some_path)` in the tests ... but they are so many ....

app/models/sso_token.rb
```diff
@@ -3,6 +3,7 @@ class SSOToken
   include ActiveModel::MassAssignmentSecurity
   include ActiveModel::Serializers::Xml
   include Rails.application.routes.url_helpers
+  include DeveloperPortal::Engine.routes.url_helpers
```
hallelujah: If we include both, I would prefer not including  any and use the `class_attribute` method

Because both can define the same routes names

hallelujah:
At the end we could end up doing, but I think it is good enough to see some improvements

```ruby
System.main_app
System.developer_portal
```

PR #1694, opened 2020-02-16
🐛 Fix double slashes in Integration Show

josemigallas:

At `Service > Integration > Configuration` the URL may present multiple slashes together (such as `//path/path`). This happens when the Backend API path is `/` or when a method has multiple slashes (this is another bug present, see [THREESCALE-4498](https://issues.redhat.com/browse/THREESCALE-4498))
Screen Shot 2020-02-14 at 17 26 12

[THREESCALE-4499: Double Slash in Mapping rules at APIcast configuration page](https://issues.redhat.com/browse/THREESCALE-4499)

**Verification steps** 

1. Go to Product page and add  Backend with path `/`.
2. At the Configuration page, there should be no double slash in Mapping Rules.
Screen Shot 2020-02-16 at 21 38 14


app/helpers/api/integrations_helper.rb
```diff
@@ -230,7 +230,8 @@ def proxy_rules_preview(owner, path: nil)
     proxy_rules = owner.proxy_rules
     last_rule = proxy_rules.last
     return 'None' unless last_rule
-    code = content_tag(:code) { "#{path}#{last_rule.pattern} => #{last_rule.metric.name}" }
+    preview = "#{path unless path == '/'}#{last_rule.pattern} => #{last_rule.metric.name}"
```
hallelujah: I think you can just use 
```suggestion
    preview = "#{File.join(path.to_s, last_rule.pattern.to_s)} => #{last_rule.metric.name}"
```

guicassolato:
```suggestion
    proxy_rule_text = "#{path unless path == '/'}#{last_rule.pattern} => #{last_rule.metric.name}"
```

app/helpers/api/integrations_helper.rb
```diff
@@ -230,7 +230,8 @@ def proxy_rules_preview(owner, path: nil)
...
+    code = content_tag(:code) { preview }
```
guicassolato:
```suggestion
    code = content_tag(:code) { proxy_rule_text }
```

test/unit/helpers/api/integrations_helper_test.rb
```diff
@@ -230,4 +230,17 @@ class IsHttpTest < self
...
+  class ProxyRulesPreview < self
+    include ProxyRulesHelper
+
+    test 'print only 1 slash when backend_api path is "/"' do
+      backend_api = FactoryBot.create(:backend_api)
+      proxy_rule = FactoryBot.create(:proxy_rule, owner: backend_api, proxy: nil)
+      backend_api_config = FactoryBot.create(:backend_api_config)
+      backend_api_config.stubs(:path).returns('/')
+
+      assert_no_match %r{//}, proxy_rules_preview(backend_api, path: backend_api_config.path)
```
guicassolato: Better to assert the exact text expected, instead of using `assert_no_match`. There are two reasons for that:
1. If the implementation changes, there are more chances of this test to become always green than otherwise, meaning it will be testing nothing
2. `/some//path` is actually is valid proxy rule and would make this test to raise a false-positive.

app/helpers/api/integrations_helper.rb
```diff
@@ -230,10 +230,14 @@ def proxy_rules_preview(owner, path: nil)
     proxy_rules = owner.proxy_rules
     last_rule = proxy_rules.last
     return 'None' unless last_rule
-    code = content_tag(:code) { "#{path}#{last_rule.pattern} => #{last_rule.metric.name}" }
+    code = content_tag(:code) { "#{proxy_rule_url(path, last_rule)} => #{last_rule.metric.name}" }
     code + link_to_more_proxy_rules(proxy_rules, proxy_rules_path_for(owner))
   end
 
+  def proxy_rule_url(path, rule)
```
guicassolato:
```suggestion
  def proxy_rule_uri(path, rule)
```

test/unit/helpers/api/integrations_helper_test.rb
```diff
@@ -230,4 +230,16 @@ class IsHttpTest < self
       refute(is_https?(nil))
     end
   end
+
+  class ProxyRulesPreview < self
```
hallelujah: That will reexecute all tests of the parent class ...

josemigallas: Fixed

PR #1697, opened 2020-02-17
[test] Allow the use of Spring

hallelujah:
Do not eager_load if Spring is defined in Kernel.
Meaning it is run inside of a Spring server fork


config/environments/test.rb
```diff
@@ -11,7 +11,7 @@
   # Do not eager load code on boot. This avoids loading your whole application
   # just for the purpose of running a single test. If you are using a tool that
   # preloads Rails for running tests, you may have to set it to true.
-  config.eager_load = true 
+  config.eager_load = Kernel.const_defined?(:Spring) ? false : true
```
Martouta: config.eager_load = !Kernel.const_defined?(:Spring)

PR #1701, opened 2020-02-17
Disable Reset Password for Master account

duduribeiro:
[THREESCALE-303](https://issues.redhat.com/browse/THREESCALE-303)
We currently enabled reseting password for master account but we
shouldn't have done this. This commit disables the reset password link
for master account.

### DIFF
--- a/app/controllers/provider/passwords_controller.rb
+++ b/app/controllers/provider/passwords_controller.rb
@@ -56,10 +56,11 @@ class Provider::PasswordsController < FrontendController
   # Can't be used for neither Buyer nor Master
   #
   def find_provider
-    unless @provider ||= Account.providers_with_master.find_by_self_domain(request.host.to_s)
-      render_error "Wrong domain '#{request.host}' for path '#{request.path}'", status: 404
-      false
-    end
+    @provider ||= Account.tenants.find_by_self_domain(request.host.to_s)
+    return if @provider
+
+    render_error "Wrong domain '#{request.host}' for path '#{request.path}'", status: 404
+    false
   end
 
   def find_user
--- a/app/controllers/provider/passwords_controller.rb
+++ b/app/controllers/provider/passwords_controller.rb
@@ -56,7 +56,7 @@ class Provider::PasswordsController < FrontendController
   # Can't be used for neither Buyer nor Master
   #
   def find_provider
-    @provider ||= Account.tenants.find_by_self_domain(request.host.to_s)
+    @provider ||= Account.tenants.find_by(self_domain: request.host.to_s)
     return if @provider
 
     render_error "Wrong domain '#{request.host}' for path '#{request.path}'", status: 404
--- a/app/javascript/src/LoginPage/LoginPageWrapper.jsx
+++ b/app/javascript/src/LoginPage/LoginPageWrapper.jsx
@@ -37,6 +37,7 @@ type Props = {
   providerSessionsPath: string,
   redirectUrl: string,
   show3scaleLoginForm: boolean,
+  masterAccount: boolean,
   session: {
     username: ?string
   }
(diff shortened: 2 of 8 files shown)


app/controllers/provider/passwords_controller.rb
```diff
@@ -48,10 +49,14 @@ def update
 
   private
 
+  def can_reset_password?
+    render_error 'Cannot reset password for Master Account. Please contact your system administrator.', status: 422 if domain_account.master?
```
Martouta: imo we should use cancancan for the permissions of this feature :smile:

Martouta: btw this is incompatible with this line:
https://github.com/3scale/porta/pull/1701/files#diff-2208b639adfb8b839f1a0ffb8ea3ef0aR60

Right now, it will NEVER reach that line... and we shouldn't have code that cannot get reached

app/controllers/provider/passwords_controller.rb
```diff
@@ -48,10 +49,14 @@ def update
...
+  end
+
   # Can't be used for neither Buyer nor Master
   #
   def find_provider
-    unless @provider ||= Account.providers_with_master.find_by_self_domain(request.host.to_s)
+    unless @provider ||= Account.providers.find_by_self_domain(request.host.to_s)
```
Martouta: 2 things about this:

1. It is better to do `return if ...` and continue in the next line than doing a method where everything is nested inside a condition.
2. I prefer `Account.tenants` over `Account.providers`. Because `providers` sometimes means just tenants and sometimes means tenants + master. And look at this:
image

Martouta:
This must be related to a Jira. If there isn't one, create it plz.

app/javascript/src/LoginPage/LoginPageWrapper.jsx
```diff
@@ -37,6 +37,7 @@ type Props = {
   providerSessionsPath: string,
   redirectUrl: string,
   show3scaleLoginForm: boolean,
+  masterAccount: boolean,
```
josemigallas: I think it's not ideal to mention "master account". We should make it a more generic feature:
```suggestion
  disablePasswordReset: boolean,
```

spec/javascripts/LoginPage/LoginPageWrapper.spec.jsx
```diff
@@ -39,6 +40,20 @@ it('should render <RequestPasswordForm/> component when formMode state is set to
   expect(wrapper.find(RequestPasswordForm).exists()).toEqual(true)
 })
 
+it('should render <ForgotCredentials/> component when masterAccount is false', () => {
```
josemigallas:
```suggestion
it('should render a reset password button, () => {
```

hallelujah:
We should also modify the routes

https://github.com/3scale/porta/blob/180a6d3f9a9a28da462c9ab100917ba860cc3523/config/routes.rb#L169

Should be moved inside something like, so the routes does not exist for master

```ruby
   namespace :provider, path: 'p', constraints: ProviderDomainConstraint do
        resource :password, :only => %i(new show update destroy)
   end
```

hallelujah:
Lets document how to reset a password for a master admin
Btw member only on master does not make really sense as they can't do anything really (maybe see only stats ...)

hallelujah:
There is https://github.com/3scale/porta/pull/1714 to reset password via console command

guicassolato:
I can still navigate to the reset password page by typing its URL in the address bar:
Screenshot 2020-02-19 at 15 31 37

thomasmaas:
Sorry, didn't see https://github.com/3scale/porta/pull/1701#issuecomment-588263028 in time. Lets fix that for cr1 @duduribeiro

hallelujah:
Answering to #1701 (comment)  Oh because the constraint is not good at all, it accepts master :(
https://github.com/3scale/porta/blob/d1a255f0207ecc85705c281a1009c2fa91867992/lib/routing_constraints.rb#L29
We should add a test for this

PR #1704, opened 2020-02-18
Some improvements

josemigallas:


app/javascript/src/ChangePassword/components/ChangePassword.jsx
```diff
@@ -18,15 +18,18 @@ type Props = {
   errors: (?FlashMessage)[]
 }
 
+const USER_PASS = 'user[password]'
+const USER_PASS_CONFIRMATION = 'user[password_confirmation]'
+
 const validationConstraints = {
-  'user[password]': {
+  USER_PASS: {
```
damianpm: Let's use the full names `USER_PASSWORD` and `USER_PASSWORD_CONFIRMATION`

app/javascript/src/Types/libs/libdefs.js
```diff
@@ -32,7 +32,10 @@ declare module 'swagger-ui-react' {
 }
 
 declare module 'validate.js' {
-  declare module.exports: any;
+  declare type ERRORS = {
```
damianpm: I think this is the problem, validate function accepts more arguments and has several possible return values, take a look at `flow-typed/npm/validate.js_v0.x.x.js`

PR #1705, opened 2020-02-18
oidc auth flow warning message for both apiap and non apiap services …

thomasmaas:
closes https://issues.redhat.com/browse/THREESCALE-3025

Screenshot 2020-02-18 11 22 07

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -655,6 +655,9 @@ en:
     python: Python
     rest: REST
 
+  oidc:
+    authorization_flow_warning: "Any changes to the authorization flow of this product will only be applied to new applications; existing applications will continue to use the original flow."
+
   time:
     formats:
       default: "%d %b %Y %H:%M:%S %Z"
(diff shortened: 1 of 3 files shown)


thomasmaas:
Might be good to point people to solution for their existing app, API? which one?

config/locales/en.yml
```diff
@@ -655,6 +655,9 @@ en:
     python: Python
     rest: REST
 
+  oidc:
+    authorization_flow_warning: "Any changes to the authorization flow will only be applied to new applications; existing applications won't be updated."
```
hallelujah: I'd rephrase adding `default values` => that would work when we will do auth flow per application

```suggestion
    authorization_flow_warning: "New application will take default values from this authorization flow ; changing it will not update existing applications."
```

thomasmaas: I don't understand the word 'default' in this context.

thomasmaas: ping @hallelujah

thomasmaas: we can change the phrase for CR1…lets just merge

josemigallas:
```suggestion
    authorization_flow_warning: "Any changes to the authorization flow will only be applied to new applications; existing applications won't be affected."
```

thomasmaas: Thanks. Based on Hery's feedback I'm now considering:

> Any changes to the authorization flow of this service will only be applied to new applications; existing applications will continue to use the original flow.

PR #1708, opened 2020-02-18
Fix deploy to OCP with PostgreSQL

guicassolato:
Deploying to OCP with PostgreSQL fails with `PG::InsufficientPrivilege: ERROR: permission denied to create database` due to `db:setup` always trying to create the database, even when the database is already there. This PR handles this error.

Closes [THREESCALE-4522](https://issues.redhat.com/browse/THREESCALE-4522)

### DIFF
--- a/openshift/system/config/database.yml
+++ b/openshift/system/config/database.yml
@@ -9,11 +9,11 @@ base: &default
     timezone: 'UTC'
 <% else %>
   adapter: mysql2
+  collation: utf8_bin
 <% end %>
 
   url: <%= ENV['DATABASE_URL'] %>
   encoding: utf8
-  collation: utf8_bin
   pool: <%= ENV.fetch('RAILS_MAX_THREADS', 5) %>
 
 production:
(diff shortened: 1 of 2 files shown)


openshift/system/config/database.yml
```diff
@@ -9,11 +9,11 @@ base: &default
     timezone: 'UTC'
 <% else %>
   adapter: mysql2
+  collation: utf8_bin
 <% end %>
 
   url: <%= ENV['DATABASE_URL'] %>
```
hallelujah: I think we should remove the `:url` parameter
Rails will auto-merge DATABASE_URL with other

In fact it has even `PRODUCTION_DATABASE_URL` `#{Rails.env.to_s.upcase}_DATABASE_URL `
https://github.com/rails/rails/blob/d72f940d1612aa8b905214c06115db48af41e8ec/activerecord/lib/active_record/database_configurations.rb#L264

We can take advantage of this later ;)
Not the purpose of this PR of course

guicassolato: I tried. I doesn't work.

PR #1710, opened 2020-02-18
APIAP small ux improvements

thomasmaas:
Fixes nothing, just tries to:
- make backends in a product and the path they have a bit easier to understand
- be consistent with naming of Private Base URL

(triggered by docs review)

Screenshot 2020-02-18 16 43 38
Screenshot 2020-02-18 16 43 43
Screenshot 2020-02-18 16 44 50

Screenshot 2020-02-18 at 16 46 24

### DIFF
--- a/app/views/api/backend_usages/edit.html.slim
+++ b/app/views/api/backend_usages/edit.html.slim
@@ -1,12 +1,11 @@
-= content_for :sublayout_title, 'Edit Backend usage'
+= content_for :sublayout_title, 'Edit Backend Usage'
 
 = semantic_form_for @backend_api_config, url: admin_service_backend_usage_path(@service, @backend_api_config) do |form|
   = form.inputs do
     li
       = form.label :backend_api_id, 'Backend'
       = form.text_field :backend_api_id, class: 'text', value: @backend_api_config.backend_api.name, disabled: true
-    = form.input :path, hint: 'Changing this path will only affect the behaviour of this Backend within the context of this Product'
+    = form.input :path, hint: 'Changing the public path will only affect this product', label: 'Public Path'
 
   = form.buttons do
     = form.commit_button 'Update'
-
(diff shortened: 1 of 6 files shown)


app/views/api/backend_usages/edit.html.slim
```diff
@@ -1,12 +1,11 @@
-= content_for :sublayout_title, 'Edit Backend usage'
+= content_for :sublayout_title, 'Edit Backend Usage'
 
 = semantic_form_for @backend_api_config, url: admin_service_backend_usage_path(@service, @backend_api_config) do |form|
   = form.inputs do
     li
       = form.label :backend_api_id, 'Backend'
       = form.text_field :backend_api_id, class: 'text', value: @backend_api_config.backend_api.name, disabled: true
-    = form.input :path, hint: 'Changing this path will only affect the behaviour of this Backend within the context of this Product'
+    = form.input :path, hint: 'Changing the public path will only affect this product', label: 'Public Path'
```
josemigallas:
```suggestion
    = form.input :path, hint: 'Changing the public path will only affect this Product', label: 'Public Path'
```

thomasmaas: I did it on purpose as I understand that Docs team doesn't want us to use capitals like that. Not sure, @cntlsn do you know?

app/views/api/backend_usages/index.html.slim
```diff
@@ -1,11 +1,15 @@
-h2 Backends
+h1 Backends
+p.intro For a product to work, it needs to have at least one backend with a Private Base URL (your API). If multiple backends are added to a product, each backend will have a unique public path.
```
didierofrivia: You mentioned in https://github.com/3scale/porta/pull/1710/files#r381168826 that Docs team don't want us to use capitals in that case... should it be consistent here as well?

didierofrivia: It applies to the rest of the changes where the occurrence of `Private Base URL` is found

thomasmaas: yeah, you're right.

PR #1711, opened 2020-02-18
Set compile to true in test environent

damianpm:

When running test (features) locally, webpacker doesn't compile, resulting in errors like:

> Webpacker can't find XXXXX in /your-path-to-file.
> Possible causes:
>       1. You want to set webpacker.yml value of compile to true for your environment
>          unless you are using the `webpack -w` or the webpack-dev-server.
>       2. webpack has not yet re-run to reflect updates.
>       3. You have misconfigured Webpacker's config/webpacker.yml file.
>       4. Your webpack configuration is not creating a manifest.

### DIFF
--- a/config/webpacker.yml
+++ b/config/webpacker.yml
@@ -42,7 +42,7 @@ development:
 
 test:
   <<: *default
-  compile: false
+  compile: true
   webpack_compile_output: true
 
   # Compile test packs to a separate directory


damianpm:
@didierofrivia Is there any reason to not compile in test environment? Test fails locally because of that

didierofrivia:
@damianpm This was set to false, because when the rest of the team running tests not related with UI complained it was taking ages. What you could do now that we are using yarn, profile running a unit, cuke, etc... and see how it goes everytime you run it

hallelujah:
We can just  run `./bin/webpacker-dev-server`
And we got the best of the two worlds :)

damianpm:
OK, closing

didierofrivia:
Also, we are having a different task for compiling test assets in circle, so that would mean doing the same thing twice and it's not cool.

PR #1714, opened 2020-02-19
[rake] Task to reset password of a User of Master tenant

hallelujah:
Relates to https://issues.redhat.com/browse/THREESCALE-303

Usage, in a shell console where the application is running

#### Resetting password

```shell
bundle exec rails 'password:reset:master'
```

You will be prompted for the email or username
Then you will be prompted for the password and confirm it

#### Resetting password providing the username/email

```shell
bundle exec rails 'password:reset:master[john]'
```

Then you will be prompted for the password and confirm it

#### Resetting password providing username/email and password

```shell
bundle exec rails 'password:reset:master[john,123456]'
```

It will set the password of User `john` to `123456`

### DIFF
--- a/lib/tasks/password_reset.rake
+++ b/lib/tasks/password_reset.rake
@@ -1,0 +1,73 @@
+# frozen_string_literal: true
+
+require 'io/console'
+
+namespace :password do
+  namespace :master do
+    desc 'Reset password for a master user'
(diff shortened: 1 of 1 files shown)


hallelujah:
Should I make test?
It is kinda hard with interactive `STDIN.gets`

didierofrivia:
You could fix 4 really easy-to-fix codeclimate issues

PR #1703, opened 2020-02-17
[gem] Update Redis and Sidekiq

hallelujah:
Trying to have libraries to be up to date as much as possible with the upstream

If this gets merge, it will supersede https://github.com/3scale/porta/pull/1700

 Updating the gem resolves all issues raised in the comments/reviews.
See https://github.com/3scale/porta/pull/1703#discussion_r381841759

Also, it closes https://issues.redhat.com/browse/THREESCALE-4495 (again), because redis-rb 4.3.1 now handles `SocketError` as well.

PR #1715, opened 2020-02-20
Test for the RedisHacks

guicassolato:
Introduces a test for the `RedisHacks` module, ensuring that the `RuntimeError` exceptions with message "Name or service not known" are handled when trying attempting to connect to Redis sentinels.

The test assumes too much about the implementation of `Redis::Client`, but I found no other way since it's testing a hack on that class. Maybe we could mock the connection with the sentinel instead.

### DIFF
--- a/config/initializers/redis_hacks.rb
+++ b/config/initializers/redis_hacks.rb
@@ -1,8 +1,15 @@
 # frozen_string_literal: true
 
+# [Patch #1]
 # Monkey patches redis-db so sentinel passwords are supported
 # See https://github.com/redis/redis-rb/pull/856
-# We can remove this once we can upgrade the gem to >= v4.1.2
+# We can remove this specific part of the hack once we can upgrade the gem to >= v4.1.2
+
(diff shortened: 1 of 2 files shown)


hallelujah:
So I was thinking, why not upgrade redis-rb #1703  and hack hiredis-rb

This way we only patch low level errors and not logic errors?

hallelujah:
See https://github.com/redis/hiredis#errors

> REDIS_ERR_OTHER: Any other error. Currently, it is only used when a specified hostname to connect to cannot be resolved.
> In every case, the errstr field in the context will be set to hold a string representation of the error.

guicassolato:
> See https://github.com/redis/hiredis#errors
> 
> > REDIS_ERR_OTHER: Any other error. Currently, it is only used when a specified hostname to connect to cannot be resolved.
> > In every case, the errstr field in the context will be set to hold a string representation of the error.

Does that mean it translates to `Errno::EPERM`?

PR #1718, opened 2020-02-21
Improving validation in change password form

damianpm:

Change password form was updated in https://github.com/3scale/porta/pull/1664 adding PF4 styling
Now we need to improve client-side validation of the form

**Validation criteria** are the following:

- Fields should be validated when loosing focus, even if they are empty
- On typing, validating starts when password confirmation length is equal or greater than password length
- Once validation has started, it shouldn't stop
- When password confirmation is empty, the message is set to "Password confirmation is mandatory"
- When password confirmation is wrong, message is set to "Password and Password confirmation must match"

None, but related to https://issues.redhat.com/browse/THREESCALE-4440


thomasmaas:
Not sure if within scope of this pr but: Screenshot 2020-02-25 18 24 03 gets red the moment you start typing.

damianpm:
> Not sure if within scope of this pr but: Screenshot 2020-02-25 18 24 03 gets red the moment you start typing.

I think is out of scope. But we can create a new task to review the other forms (login, reset password, sign form)

app/javascript/src/ChangePassword/components/ChangePassword.jsx
```diff
@@ -18,25 +18,109 @@ type Props = {
...
+    isValid,
+    onChange,
+    onBlur
+  }
+}
+
+const getPasswordConfirmationErrors = (event) => {
+  const errors = isValidForm(event.currentTarget.parentNode.parentNode)
+  const passwordConfirmationError = (errors && errors[PASSWORD_CONFIRMATION]) && errors[PASSWORD_CONFIRMATION][0]
+  const passwordsMatchError = passwordConfirmationError && passwordConfirmationError.includes('is not equal to')
```
didierofrivia: is it really necessary to discern this error? Can't just be another password confirmation error?

app/javascript/src/ChangePassword/components/ChangePassword.jsx
```diff
@@ -18,25 +18,109 @@ type Props = {
...
+  const [isValid, setIsValid] = useState(undefined)
+  const [passwordDoesntMatch, setPasswordDoesntMatch] = useState(undefined)
+  const [startValidating, setStartValidating] = useState(false)
+
+  const onChange = (value, event) => {
+    setValue(value)
+    const { passwordConfirmationError, passwordsMatchError } = getPasswordConfirmationErrors(event)
+    setPasswordDoesntMatch(passwordsMatchError)
+    const initValidation = shouldInitValidation(value, event)
+    const validationStarted = startValidating || isValid !== undefined
```
didierofrivia: There's quite a lot of logic being stored in order to know if the form needs to be validated or not. I'd say it's better to do validation when the form is not pristine, or not. That kind of check asks for a simpler way to do it

app/javascript/src/ChangePassword/components/ChangePassword.jsx
```diff
@@ -18,25 +18,109 @@ type Props = {
...
+  const errors = isValidForm(event.currentTarget.parentNode.parentNode)
+  const passwordConfirmationError = (errors && errors[PASSWORD_CONFIRMATION]) && errors[PASSWORD_CONFIRMATION][0]
+  const passwordsMatchError = passwordConfirmationError && passwordConfirmationError.includes('is not equal to')
+  return {
+    passwordConfirmationError,
+    passwordsMatchError
+  }
+}
+
+const shouldInitValidation = (value, event) => {
```
didierofrivia: This fn name doesn't quite match what it does, but what it's gonna be used for.

app/javascript/src/ChangePassword/components/ChangePassword.jsx
```diff
@@ -18,25 +18,109 @@ type Props = {
...
+  const passwordLength = event.currentTarget.parentNode.parentNode[PASSWORD].value.length
```
didierofrivia: quite deep it needs to look for what's needed

app/javascript/src/ChangePassword/components/ChangePassword.jsx
```diff
@@ -18,25 +18,109 @@ type Props = {
...
+  const [startValidating, setStartValidating] = useState(false)
+
+  const onChange = (value, event) => {
+    setValue(value)
+    const { passwordConfirmationError, passwordsMatchError } = getPasswordConfirmationErrors(event)
+    setPasswordDoesntMatch(passwordsMatchError)
+    const initValidation = shouldInitValidation(value, event)
+    const validationStarted = startValidating || isValid !== undefined
+    setStartValidating(!validationStarted ? initValidation : undefined)
+    setIsValid((initValidation || validationStarted) ? !passwordConfirmationError : undefined)
```
didierofrivia: This smells, and it's related on my comment above

app/javascript/src/ChangePassword/components/ChangePassword.jsx
damianpm: Well, I agree there is a lot of logic here. But the point is I was explicitly asked (@thomasmaas @cntlsn ) to implement validation not only when pristine/not pristine, but according to several scenarios like:
User tabs and input looses focus (empty)
User tabs and input looses input (filled)
Password confirmation validation should start only after it's length is ` >=` to password length
After initiating validation it shouldn't stop (no matter the length)

app/javascript/src/ChangePassword/components/ChangePassword.jsx
didierofrivia: But can't be both treated and rendered the same way? What would make the latter so specific?

josemigallas: It should be 0/1: either passwords match or they don't. IMO checking for other errors doesn't really make sense.

app/javascript/src/ChangePassword/components/ChangePassword.jsx
josemigallas: Maybe extract this hook and make it more generic (take 2 password forms, not a Node). This way we can simplify the logic and make it easier to follow.

app/javascript/src/ChangePassword/components/ChangePassword.jsx
damianpm: If everybody agree I can leave a single error: `Password and Password confirmation must match`
WDYT @thomasmaas @cntlsn ?

app/javascript/src/LoginPage/loginForms/FormGroups.jsx
```diff
@@ -82,11 +87,17 @@ const EmailField = ({inputProps}: {inputProps: InputProps}) => {
 }
 
 const PasswordField = ({inputProps}: {inputProps: InputProps}) => {
+  const isPasswordConfirmation = inputProps.isPasswordConfirmation
```
josemigallas: This doesn't belong here IMO, it adds new responsibilities that makes a Password field more complex that it should. I think there should be a `ConfirmPasswordField` instead.

app/javascript/src/LoginPage/loginForms/FormGroups.jsx
```diff
@@ -8,14 +8,17 @@ import {
...
+  fieldId: 'string',
+  isValid: boolean,
+  name: string,
+  value: string,
+  onChange?: () => void,
+  onBlur?: () => void,
+  autoFocus?: string,
+  ariaInvalid?: boolean,
+  isPasswordConfirmation?: boolean,
+  passwordDoesntMatch?: boolean
```
josemigallas: This field also seems too restrictive. I think it would be better to pass a string and have more control.

app/javascript/src/ChangePassword/components/ChangePassword.jsx
thomasmaas: need more context

app/javascript/src/LoginPage/utils/formValidation.js
```diff
@@ -18,7 +18,9 @@ const constraintsTypes = {
   }
 }
 
-const validateForm = (form: SyntheticEvent<HTMLFormElement>, constraints: {}) => {
+type HTMLForm = SyntheticEvent<HTMLFormElement> | HTMLFormElement | HTMLElement | null
```
josemigallas: Maybe just a naming problem but it looks weird that `validateForm` validates both forms and events. Is `validate.js` able to distinguish them?

app/javascript/src/ChangePassword/components/ChangePassword.jsx
damianpm: The discussion is if password confirmation field should have 2 possible errors: 
1.- When blank: "Password confirmation is mandatory" 
2.- When it doesn't match with password: "Password and Password confirmation must match"
**OR** just one single error: "Password and Password confirmation must match" for both cases

damianpm: @thomasmaas 
blank
Vs.
wrong

app/javascript/src/LoginPage/utils/formValidation.js
damianpm: `validateForm` accepts a form element or an object, fixing

app/javascript/src/ChangePassword/components/ChangePassword.jsx
damianpm: I'm passing the whole form (formNode) because validate.js is doing the validation of all the inputs, including the equality check. If we pass the 2 inputs separated, we'd need to implement that equality check by ourselves, and the library would be useless.

app/javascript/src/ChangePassword/components/ChangePassword.jsx
```diff
@@ -18,25 +18,111 @@ type Props = {
...
+  const passwordConfirmationError = (errors && errors[PASSWORD_CONFIRMATION]) && errors[PASSWORD_CONFIRMATION][0]
+  const passwordsMatchError = passwordConfirmationError && passwordConfirmationError.includes('is not equal to')
+  return {
+    hasError: !!passwordConfirmationError,
+    errorMessage: passwordsMatchError ? 'mustMatch' : 'isMandatory'
+  }
+}
+
+const compareInputsLength = (value) => {
+  const passwordInput = document.querySelector('input#user_password')
```
josemigallas: Why don't we use a Node here instead of accessing the dom every time the input changes?

app/javascript/src/ChangePassword/components/ChangePassword.jsx
josemigallas: We shouldn't worry about `validate.js`, it's only a tool and a disposable one. We should try to craft out code the most independent / generic the better. In any case, I think the hook should be defined elsewhere so that it can be better tested and reviewd.

app/javascript/src/ChangePassword/components/ChangePassword.jsx
```diff
@@ -18,25 +18,111 @@ type Props = {
...
+    errorMessage,
+    onChange,
+    onBlur
+  }
+}
+
+const ChangePassword = ({ lostPasswordToken, url, errors }: Props) => {
+  const formNode = document.querySelector('form')
+  const form = useFormState(formNode)
+  const password = usePasswordState()
```
josemigallas: Intuitively, this should be something like: `usePasswordState({ value: '', isValid: false })`.

Or perhaps `const { value, isValid } = usePasswordState()`.

app/javascript/src/ChangePassword/components/ChangePasswordHooks.jsx
```diff
@@ -0,0 +1,88 @@
...
+    const initValidation = comparePasswordsLength(event)
+    if (initValidation && !validationStarted) {
+      setValidationStarted(initValidation)
+    }
+    if (initValidation || validationStarted) {
+      setIsPasswordConfirmationValid(!passwordConfirmationError)
+    }
+    setConfirmationErrorMessage(errorMessage)
+    setIsFormDisabled(!!errors)
+  }
```
didierofrivia: I still believe the `validationStarted` is not needed.
This would be the same:

```js
  const onFormChange = (event) => {
    const errors = isFormValid(event.currentTarget)
    const { passwordConfirmationError, errorMessage } = getPasswordConfirmationError(errors)
    const initValidation = comparePasswordsLength(event)
   
    if (initValidation) {
      setIsPasswordConfirmationValid(!passwordConfirmationError)
    }
    setConfirmationErrorMessage(errorMessage)
    setIsFormDisabled(!!errors)
  }
```

app/javascript/src/ChangePassword/components/ChangePasswordHooks.jsx
```diff
@@ -0,0 +1,88 @@
...
+
+  const onPasswordChange = (value, event) => {
+    setPasswordValue(value)
+    setIsPasswordValid(!!validateSingleField(event))
+  }
+
+  const onPasswordBlur = () => setIsPasswordValid(!!isPasswordValid)
```
didierofrivia: Plus you are triggering a change here... so there's an instant error message on blur without changing any of the inputs.

damianpm: That's intentional, because validation should happen if you tab and loose focus, even empty

didierofrivia: My point exactly, you are initiation a validation even if not setting `validationStarted`

app/javascript/src/ChangePassword/components/ChangePasswordHooks.jsx
damianpm: It's not the same because one edge case is missing: 

1. You type password: "abc"
2. You type confirmation: "abcd" // Error is displayed
3. You delete some chars in confirmation: "ab" // Error not displayed anymore, as it's length is < password length

Error message should not disappear once validation has started

app/javascript/src/ChangePassword/components/ChangePasswordHooks.jsx
damianpm: But that's not an error, it's the desired behavior:
On loosing focus you should validate, even empty 
On typing, you should start validating when password confirmation is >= to password

app/javascript/src/ChangePassword/components/ChangePasswordHooks.jsx
cntlsn: @damianpm is it just me or it feels like the description of the edge case is clashing with your last sentence?
I would say: If a confirmation error is triggered, it will stay visible until correct pass confirmation is typed

Does that make sense to you?

damianpm: Yes, that's what I mean
Point 3 is what is we don't want
> You delete some chars in confirmation: "ab" // Error not displayed anymore, as it's length is < password length

PR #1719, opened 2020-02-21
[DEPRECATED - WE WILL TAKE A DIFFERENT APPROACH] [Bugfix][ActiveJob::DeserializationError] Couldn't find Account in BackendDeleteServiceWorker

Martouta:
The 1st idea for [THREESCALE-4620](https://issues.redhat.com/browse/THREESCALE-4620)

What was happening before this PR is:
1. The DeleteAccountHierarchyWorker performs for its associations, including its services
2. The services of the account get destroyed and a `ServiceDeletedEvent` is created for each one of them.
3. As the associations of the account are already deleted, the account gets deleted.
4. The subscriber of `ServiceDeletedEvent` calls `BackendDeleteServiceWorker` sending the event. It tries to deserialize it but it cannot do it because the account has already been deleted.

### DIFF
--- a/app/events/services/service_deleted_event.rb
+++ b/app/events/services/service_deleted_event.rb
@@ -12,7 +12,7 @@ class Services::ServiceDeletedEvent < ServiceRelatedEvent
         provider_id: provider.id
       }
     }
-    data[:provider] = provider if provider.persisted?
+    data[:provider] = provider if provider.persisted? && !provider.scheduled_for_deletion?
 
     new(data)
   end
--- a/test/events/services/service_deleted_event_test.rb
+++ b/test/events/services/service_deleted_event_test.rb
@@ -34,7 +34,7 @@ class Services::ServiceDeletedEventTest < ActiveSupport::TestCase
     assert_can Ability.new(admin), :show, event
   end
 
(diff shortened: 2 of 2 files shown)


app/events/services/service_deleted_event.rb
```diff
@@ -12,7 +12,7 @@ def self.create(service)
         provider_id: provider.id
       }
     }
-    data[:provider] = provider if provider.persisted?
+    data[:provider] = provider if provider.persisted? && !provider.scheduled_for_deletion?
```
hallelujah: I prefer you rescue in the activejob instead
If the object is not found just ignore it

Martouta: @hallelujah I can rescue but I do not think that we should ignore in this case... This ServiceDeletedEvent may be used for more things (or not, I do not know), but for deleting in Backend in particular, we do not need the provider to be still persisted (or even to have its data in the event at all, except for the id, which we already do), and we want it to be removed from Backend anyway.

Martouta:
@hallelujah But that is what it must do... :confused: We want the Service to be removed from Backend when it is deleted from System, and we want to remove the Service (in System and in Backend and in Zync and everywhere) when we remove its Tenant from the DB.

What do you propose otherwise?

PR #1722, opened 2020-02-21
[WIP] Fix the endless select of api docs services of a tenant when removing the whole tenant

Martouta:
Still testing.
Part of [THREESCALE-4483](https://issues.redhat.com/browse/THREESCALE-4483)


Martouta:
This PR was here for testing. It indeed helps, but I rather find the root cause of the issue, because the real difference is only that this selects the ID while the other one selects everything 🤔

PR #1724, opened 2020-02-21
Monkey patch Redis::Connection::Hiredis#connect

guicassolato:
Replaces the more complex `RedisHacks` for a simple hack on `Hiredis::Connection#connect`, allowing redis-rb to be upgraded to 4.1.3 (https://github.com/3scale/porta/pull/1703).

### DIFF
--- a/config/initializers/hiredis_hacks.rb
+++ b/config/initializers/hiredis_hacks.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+# Monkey patches hiredis so connections errors with exception class `RuntimeError` and
+# message "Name or service not known" (i.e. name solving problems) are properly handled
+# as `Redis::BaseConnectionError` errors, which will be caught by `Redis::Client#establish_connection`
+
+module HiredisHacks
+  module Connection
+    def connect(*args)
+      super
+    rescue RuntimeError => exception
+      if exception.message =~ /Name or service not known|nodename nor servname provided, or not known/
+        raise Redis::BaseConnectionError.new("#{exception.message} (#{exception.class})")
+      else
+        raise
+      end
+    end
+  end
+end
+
+Hiredis::Connection.prepend(HiredisHacks::Connection)
--- a/config/initializers/redis_hacks.rb
+++ b/config/initializers/redis_hacks.rb
@@ -1,16 +1,16 @@
 # frozen_string_literal: true
 
-# Monkey patches hiredis so connections errors with exception class `RuntimeError` and
+# Monkey patches redis-rb so connections errors with exception class `RuntimeError` and
 # message "Name or service not known" (i.e. name solving problems) are properly handled
 # as `Redis::BaseConnectionError` errors, which will be caught by `Redis::Client#establish_connection`
 
-module HiredisHacks
-  module Connection
+module RedisHacks
+  module HiredisConnection
     def connect(*args)
       super
     rescue RuntimeError => exception
       if exception.message =~ /Name or service not known|nodename nor servname provided, or not known/
-        raise Redis::BaseConnectionError.new("#{exception.message} (#{exception.class})")
+        raise Redis::BaseConnectionError.new(exception.message)
       else
         raise
       end
@@ -18,4 +18,4 @@ module HiredisHacks
   end
 end
 
-Hiredis::Connection.prepend(HiredisHacks::Connection)
+Redis::Connection::Hiredis.singleton_class.prepend(RedisHacks::HiredisConnection)
--- a/test/unit/hiredis_hacks_test.rb
+++ b/test/unit/hiredis_hacks_test.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class HiredisHacksTest < ActiveSupport::TestCase
+  test 'Name or service not known' do
+    redis_config = ThreeScale::RedisConfig.new(
+      url: 'redis://mymaster',
+      pool_size: 5,
+      pool_timeout: 5,
+      sentinels: 'invalid_host:26380,other_invalid_host:26381',
+      role: :master
+    ).config
+
+    redis = Redis.new(redis_config)
+
+    exception = assert_raises(Redis::CannotConnectError) { redis._client.send(:establish_connection) }
+    assert_equal 'No sentinels available.', exception.message
+  end
+
+  protected
+
+  def build_sentinel_options(redis_client, sentinel)
+    options = redis_client.instance_variable_get(:@options).dup
+    options.delete(:sentinels)
+    options.merge!(
+      host: sentinel[:host],
+      port: sentinel[:port],
+      password: sentinel[:password],
+      reconnect_attempts: 0
+    )
+  end
+end
--- a/test/unit/redis_hacks_test.rb
+++ b/test/unit/redis_hacks_test.rb
@@ -2,7 +2,7 @@
 
 require 'test_helper'
 
-class HiredisHacksTest < ActiveSupport::TestCase
+class RedisHacksTest < ActiveSupport::TestCase
   test 'Name or service not known' do
     redis_config = ThreeScale::RedisConfig.new(
       url: 'redis://mymaster',


config/initializers/hiredis_hacks.rb
```diff
@@ -0,0 +1,17 @@
...
+# Monkey patches hiredis so connections errors with exception class `RuntimeError` and
+# message "Name or service not known" (i.e. name solving problems) are properly handled
+# as `SocketError` errors, which will be caught by `Redis::Client#establish_connection`
+
+module HiredisHacks
+  module Connection
+    def connect(*args)
+      super
+    rescue RuntimeError => exception
+      raise SocketError.new(exception.message) unless exception.message =~ /Name or service not known/
```
guicassolato: Oops. I'll fix this.

config/initializers/hiredis_hacks.rb
```diff
@@ -0,0 +1,21 @@
...
+# Monkey patches hiredis so connections errors with exception class `RuntimeError` and
+# message "Name or service not known" (i.e. name solving problems) are properly handled
+# as `SocketError` errors, which will be caught by `Redis::Client#establish_connection`
+
+module HiredisHacks
+  module Connection
+    def connect(*args)
+      super
+    rescue RuntimeError => exception
+      if exception.message =~ /Name or service not known|nodename nor servname provided, or not known/
```
guicassolato: Perhaps we can find a better way to identify the error here, based on https://github.com/3scale/porta/pull/1715#issuecomment-589210865

hallelujah: Based on https://github.com/redis/hiredis/search?q=REDIS_ERR_OTHER&unscoped_q=REDIS_ERR_OTHER
And the disclaimer in https://github.com/redis/hiredis#errors

> REDIS_ERR_OTHER: Any other error. Currently, it is only used when a specified hostname to connect to cannot be resolved.

Can't we just rescue all ? it is a `Redis::BaseConnectionError`? It is inside the `connect` method too so another point to just rescue.

hallelujah: To identify the error it would be really painful, there is no way from a String to know 100% the error class in Ruby.
And let alone from an error code (errno) in Ruby as it depends heavily on the language implementation.

To do it the hard (maybe right) way:

* You will need to patch `hiredis` directly (and indirectly), either:

By defining more errors here: https://github.com/redis/hiredis/blob/ff4fa45422dc104f3377cdc3bc88d2d24ac2fcf3/read.h#L44-L49
Doing it incrementally when seeing an error on runtime (PR each time)

Or worse solution completely reworking `__redisSetError` to add the error code `errno` inside the struct
https://github.com/redis/hiredis/blob/c96d492215eb3d4c80f4d7344ac41859eced5bab/hiredis.c#L641

Then use the errno in hiredis-rb or any other language wrapper

guicassolato: > To identify the error it would be really painful, there is no way from a String to know 100% the error class in Ruby.
> And let alone from an error code (errno) in Ruby as it depends heavily on the language implementation.
> 
> To do it the hard (maybe right) way:
> 
> You will need to patch `hiredis` directly (and indirectly), either:
> 
> * By defining more errors here: https://github.com/redis/hiredis/blob/ff4fa45422dc104f3377cdc3bc88d2d24ac2fcf3/read.h#L44-L49
>   Doing it incrementally when seeing an error on runtime (PR each time)
> * Or worse solution completely reworking `__redisSetError` to add the error code `errno` inside the struct
>   https://github.com/redis/hiredis/blob/c96d492215eb3d4c80f4d7344ac41859eced5bab/hiredis.c#L641
> 
> Then use the new error or the error code (errno) in hiredis-rb or any other language wrapper

Identifying the error based on the message is what we are doing now anyway: https://github.com/3scale/porta/blob/d248f3221eb3d6dbeb302a433f63b4b80bf5373a/config/initializers/redis_hacks.rb#L34

This is a proposal to remove the other hack, i.e. the part to support sentinel passwords.

guicassolato: > Based on https://github.com/redis/hiredis/search?q=REDIS_ERR_OTHER&unscoped_q=REDIS_ERR_OTHER
> And the disclaimer in https://github.com/redis/hiredis#errors
> 
> > REDIS_ERR_OTHER: Any other error. Currently, it is only used when a specified hostname to connect to cannot be resolved.
> 
> Can't we just rescue all ? it is a `Redis::BaseConnectionError`? It is inside the `connect` method too so another point to just rescue.

I'm afraid that https://github.com/redis/hiredis setting `REDIS_ERR_OTHER` does not cause a `Redis::BaseConnectionError` to be raised. As said in https://github.com/redis/hiredis-rb:

> All other errors (such as a protocol error) result in a `RuntimeError`.

And this is indeed what was experienced and reported in https://issues.redhat.com/browse/THREESCALE-4495. 

Even if you track in the code of https://github.com/redis/redis-rb/tree/v4.1.3, this is more or less the stack for a sentinel connection:

```
Hiredis::Connection#connect                        # not the same in C, but the Ruby implementation gives an idea: https://github.com/redis/hiredis-rb/blob/0134bf1f438767a6b79ef206bb33df8b78e92e0c/lib/hiredis/ruby/connection.rb#L185-L199
Redis::Connection::Hiredis.connect                 # https://github.com/redis/redis-rb/blob/a2c562c002bc8f86d1f47818d63db2da1c5c3d3f/lib/redis/connection/hiredis.rb#L19
Redis::Client#establish_connection                 # https://github.com/redis/redis-rb/blob/a2c562c002bc8f86d1f47818d63db2da1c5c3d3f/lib/redis/client.rb#L349
Redis::Client#connect                              # https://github.com/redis/redis-rb/blob/a2c562c002bc8f86d1f47818d63db2da1c5c3d3f/lib/redis/client.rb#L106
Redis::Client#ensure_connected                     # https://github.com/redis/redis-rb/blob/a2c562c002bc8f86d1f47818d63db2da1c5c3d3f/lib/redis/client.rb#L381
Redis::Client#process                              # https://github.com/redis/redis-rb/blob/a2c562c002bc8f86d1f47818d63db2da1c5c3d3f/lib/redis/client.rb#L231
Redis::Client#call                                 # https://github.com/redis/redis-rb/blob/a2c562c002bc8f86d1f47818d63db2da1c5c3d3f/lib/redis/client.rb#L125
Redis::Client::Connector::Sentinel#sentinel_detect # https://github.com/redis/redis-rb/blob/a2c562c002bc8f86d1f47818d63db2da1c5c3d3f/lib/redis/client.rb#L585
Redis::Client::Connector::Sentinel#resolve_master  # https://github.com/redis/redis-rb/blob/a2c562c002bc8f86d1f47818d63db2da1c5c3d3f/lib/redis/client.rb#L603
Redis::Client::Connector::Sentinel#resolve         # https://github.com/redis/redis-rb/blob/a2c562c002bc8f86d1f47818d63db2da1c5c3d3f/lib/redis/client.rb#L565
Redis::Client#establish_connection                 # https://github.com/redis/redis-rb/blob/a2c562c002bc8f86d1f47818d63db2da1c5c3d3f/lib/redis/client.rb#L344
Redis::Client#connect                              # https://github.com/redis/redis-rb/blob/a2c562c002bc8f86d1f47818d63db2da1c5c3d3f/lib/redis/client.rb#L106
Redis::Client#ensure_connected                     # https://github.com/redis/redis-rb/blob/a2c562c002bc8f86d1f47818d63db2da1c5c3d3f/lib/redis/client.rb#L381
Redis::Client#process                              # https://github.com/redis/redis-rb/blob/a2c562c002bc8f86d1f47818d63db2da1c5c3d3f/lib/redis/client.rb#L231
Redis::Client#call                                 # https://github.com/redis/redis-rb/blob/a2c562c002bc8f86d1f47818d63db2da1c5c3d3f/lib/redis/client.rb#L125
Redis#ping                                         # https://github.com/redis/redis-rb/blob/a2c562c002bc8f86d1f47818d63db2da1c5c3d3f/lib/redis.rb#L153
```

TL;DR: when a sentinel client connection fails, it does not raise a `Redis::BaseConnectionError`.

hallelujah: I meant in your hack just raise a Redis::BaseConnectionError (with appropriate subclass) instead of the generic SocketError

guicassolato: Oh! Sorry. I got that wrong. Sure, we can do that.

guicassolato: Done

hallelujah: Do you really want to filter on the message?
I'd just really not filter at all because it is for the `#connect` method, so no need to have it filtered, because something bad happened while trying to connect and it can be anything.

On our case that still means the sentinel cannot be used

guicassolato: Hmm. I've done it, but thinking again now. Maybe not a good idea. The method is the same for connection with or without sentinels. (The only difference is that for sentinels, it creates one client per sentinel.) So we would be suppressing connection failures in general. We virtually removing `TimeoutError`, `SocketError`, and all sorts of `Errno::`. Better keep the message.

hallelujah:
```suggestion
      raise Redis::BaseConnectionError.new(exception.message)
```

No need to display `exception.class`, we have `#cause`
Also looking at the call stack https://github.com/3scale/porta/pull/1724#discussion_r383189779
All other errors will be rescued before getting into this hack

hallelujah: Oh I get it
Then we should rescue here then https://github.com/redis/redis-rb/blob/41395e9a1923fe72fad8403e60cb98627d515a85/lib/redis/connection/hiredis.rb#L11

guicassolato: This would be hacking https://github.com/redis/redis-rb, not https://github.com/redis/hiredis-rb.

hallelujah: It i s redis-rb hiredis.rb file :)

guicassolato: IMO, it's hiredis-rb (the gem) that should be raising the exception class, not a redis-rb problem. But all right. Adding a commit now.

guicassolato: Done

PR #1731, opened 2020-02-26
Fix Selenium warnings when running cukes

josemigallas:

Trying to remove these deprecation warnings by updating feature-related gems.

[THREESCALE-4551: Fix Selenium Webdriver deprecation warnings](https://issues.redhat.com/browse/THREESCALE-4551)


Gemfile.lock
```diff
@@ -259,22 +261,27 @@ GEM
     crack (0.4.3)
       safe_yaml (~> 1.0.0)
     crass (1.0.6)
-    cucumber (2.99.0)
+    cucumber (3.1.2)
```
hallelujah: Updating cucumber is not as simple
Do you want to work on it?

josemigallas: Probably not but I'm still figuring it out

PR #1732, opened 2020-02-26
Use uri instead of original_request.path in the apiap routing policy

guicassolato:
We currently use `original_request.path` in the config for the APIAP routing policy. This causes the URL Rewriting policy to be a no-op when used in combination with APIAP. This PR replaces `original_request.path` with `uri`, which should have no side-effect for APIAP and re-enable the URL Rewriting policy in combination.

Customers may need to regenerate their proxy configs (i.e. promote) for this change to take effect, but this is OK, because that would be needed anyway in order to append an eventual URL Rewriting policy config.

Closes [THREESCALE-4301](https://issues.redhat.com/browse/THREESCALE-4301)

### DIFF
--- a/app/lib/backend_api_logic/routing_policy.rb
+++ b/app/lib/backend_api_logic/routing_policy.rb
@@ -73,7 +73,7 @@ module BackendApiLogic
         def replace_path
           return {} if config_path.blank?
 
-          { replace_path: "{{original_request.path | remove_first: '#{config_path.path}'}}" }
+          { replace_path: "{{uri | remove_first: '#{config_path.path}'}}" }
         end
       end
       private_constant :Rule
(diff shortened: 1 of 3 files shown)


hallelujah:
As this is not the first time we need to change something in Porta and not in APIcast, do we plan on doing some integration tests?

There are some for APIcast V1 in Porta code base but none for APIcast V2

What I am worried about is that this change is only Porta's side but we do not have regression tests about it, aka we can't be sure if it will break with future APIcast or if it is not breaking something else.

I know this will be hard to achieve, maybe asking @3scale/qe if they can add some integration testing as planned between components to verify this is tracked

guicassolato:
Tested in OpenShift. It works 🙂

PR #1733, opened 2020-02-27
Rake task to reset proxy config change history

guicassolato:
```
# run for all provider accounts
rake task proxy:reset_config_change_history

# run for the proxies of provider whose id is <account_id>
rake task proxy:reset_config_change_history[<account_id>] 
```

Closes [THREESCALE-4588](https://issues.redhat.com/browse/THREESCALE-4588)

### DIFF
--- a/lib/tasks/proxy.rake
+++ b/lib/tasks/proxy.rake
@@ -48,11 +48,14 @@ namespace :proxy do
     reset_date = Time.utc(1900, 1, 1).freeze
     progress = ProgressCounter.new(collection.count)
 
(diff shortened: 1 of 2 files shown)


guicassolato:
Benchmark in preview (all accounts): 53'34''.

Martouta:
Approved but those 53'34' minutes worry me... maybe we should consider releasing the DB (`sleep(0.5)`) after each batch so other processes can update it in the meantime :thinking: or maybe I am being too paranoic and it is fine :smile: but in that case, we should run it when there isn't much traffic :thinking:

PR #1737, opened 2020-02-28
Fix test curl command bad URI

guicassolato:
Closes [THREESCALE-3797](https://issues.redhat.com/browse/THREESCALE-3797) + Refactoring of the `CurlCommandBuilder` class used by the `IntegrationsHelper`.

It also fixes a bug with the test curl command showed for accounts with api as product enabled, when apicast's target environment is `production`. I don't think we really show the example curl command in such case, in a way that the bug was probably never activated. Still, just in case we ever decide to use that option, the `environment` argument was supposed to be passed along and it is not in https://github.com/3scale/porta/blob/3ef93f4c68835705e134cd81c9cb9f7f2c888cbd/app/helpers/api/integrations_helper.rb#L129 It was hard-coded always `:staging`.

### DIFF
--- a/app/helpers/api/integrations_helper.rb
+++ b/app/helpers/api/integrations_helper.rb
@@ -1,132 +1,13 @@
+# frozen_string_literal: true
+
 module Api::IntegrationsHelper
-
-  class CurlCommandBuilder
-    def initialize(proxy)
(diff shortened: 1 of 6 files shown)


hallelujah:
Thank you!

PR #1738, opened 2020-03-02
Remove EndUser and EndUserPlans [triggers]

josemigallas:
Removes the triggers for End User Plans in the 3 DBs. Merge before https://github.com/3scale/porta/pull/1648

[THREESCALE-2490: Remove “End-user plans” feature (SaaS) - Phase 2](https://issues.redhat.com/browse/THREESCALE-2490)

### DIFF
--- a/lib/system/database/definitions/mysql.rb
+++ b/lib/system/database/definitions/mysql.rb
@@ -63,12 +63,6 @@ System::Database::MySQL.define do
     SQL
   end
 
-  trigger 'end_user_plans' do
-    <<~SQL
-      SET NEW.tenant_id = (SELECT tenant_id FROM services WHERE id = NEW.service_id AND tenant_id <> master_id);
-    SQL
-  end
-
   trigger 'features' do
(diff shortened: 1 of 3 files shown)


guicassolato:
> You also need to make a migration to remove the triggers in the DBs that are already created and with data.

Once the table is dropped, so are the related triggers.

guicassolato:
What @Martouta said about the triggers, we actually need to stop creating them

https://github.com/3scale/porta/blob/6dd9108ae44df975f9a5c11ed67094248e57f1bc/lib/system/database/definitions/mysql.rb#L66-L70

https://github.com/3scale/porta/blob/6dd9108ae44df975f9a5c11ed67094248e57f1bc/lib/system/database/definitions/postgres.rb#L66-L70

https://github.com/3scale/porta/blob/6dd9108ae44df975f9a5c11ed67094248e57f1bc/lib/system/database/definitions/oracle.rb#L66-L70

PR #1739, opened 2020-03-02
Use and cache system_url_helpers and cms_url_helpers

hallelujah:
They are initialized just once and reused later

We could use System::DeveloperRoutes and System::MainRoutes

But I think it is not intended to be constantized even though they are
modules

Related to #1693 

How to verify there is no more use of generated/non-cached routes?

```shell
$ rg '^((?!System::UrlHelpers).)*url_helpers' --pcre2 --type ruby
```

### DIFF
--- a/lib/developer_portal/lib/cms/toolbar.rb
+++ b/lib/developer_portal/lib/cms/toolbar.rb
@@ -113,7 +113,7 @@ module CMS::Toolbar
   end
 
   class View < ActionView::Base
-    include ::Rails.application.routes.url_helpers
+    include System::UrlHelpers.system_url_helpers
     include ::DeveloperPortal::CMS::ToolbarHelper
   end
 end
--- a/app/lib/authentication/strategy/base.rb
+++ b/app/lib/authentication/strategy/base.rb
@@ -22,7 +22,7 @@ module Authentication
     end
 
     class Base
-      include DeveloperPortal::Engine.routes.url_helpers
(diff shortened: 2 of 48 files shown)


hallelujah:
@guicassolato Yes, I can even just do module function ... or use plain constants like:

```ruby
System::MainApp = Rails.application.routes.url_helpers
```

or 

```ruby
module RoutesHelpers
  module_function

  def system_url_helpers
    @system_url_helpers ||=Rails.application.routes.url_helpers
  end
end
```

But, first solution is ugly because constants assignments without freeze, I do not know how freeze will behave with the generated module ...

The second solution use module_function which makes all methods private in the including class ... same, the behaviour is unknown

lib/developer_portal/lib/cms/toolbar.rb
```diff
@@ -113,7 +113,9 @@ def cms_toolbar_with_iframe_html
   end
 
   class View < ActionView::Base
-    include ::Rails.application.routes.url_helpers
-    include ::DeveloperPortal::CMS::ToolbarHelper
+    # FIXME: There are shared routes between both
```
hallelujah: @3scale/system-ruby this is something we should never do... there are paths that exist in main app and in the dev portal.

hallelujah: Oh I am totally mistaken here!!!!! Sorry I will fix

PR #1745, opened 2020-03-03
Update package.json and yarn.lock after yarn checks integrity

Martouta:
When I was using my personal laptop, for every action like going to the rails console or running the server, it appeared a message like this one:
image

And I always had to end up updating the package.json and the yarn.lock by it in order to be able to keep working.

I thought it was a problem with my laptop, but now I've done a setup from scratch in the Mac, and I kept having the same problem... so this is what it updates in order to work properly for me.

But I definitely need help from the @3scale/system-ux team because I do not know if this is correct or not and I may be doing something wrong.

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,6 +113,7 @@
     "url-polyfill": "^1.1.3",
     "validate.js": "^0.13.1",
     "virtual-dom": "^2.1.1",
-    "whatwg-fetch": "^3.0.0"
+    "whatwg-fetch": "^3.0.0",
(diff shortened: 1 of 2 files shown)


didierofrivia:
You should install `yarn` with brew or npm but globally before anything to avoid Rails wanting to install it locally. https://github.com/3scale/porta/blob/master/INSTALL.md#node-packages

Martouta:
You are right! it works! thaaaaaanks!! 💃 💃 💃

PR #1748, opened 2020-03-04
Change password validation improve bis

didierofrivia:
There was a pending improvement regarding the code, and this is the PR.
Mar-17-2020 15-21-33

TODO:
* [x] flow


damianpm:
This PR changes the behavior of validation:
- When typing on Password, it shows immediately an error message in Password Confirmation
- Password Confirmation starts validating form the first character

@thomasmaas @cntlsn WDYT?

app/javascript/src/ChangePassword/components/ChangePasswordHooks.jsx
```diff
@@ -1,87 +1,95 @@
...
+
+type PasswordField = 'user[password]'
+type PasswordConfirmationField = 'user[password_confirmation]'
+type FieldName = PasswordField | PasswordConfirmationField
+type Constraints = {[FieldName]: {[string]: {}}}
+type FieldState = {[FieldName]: string}
+type ValidationErrors = {[FieldName]: string[]}
+type FieldErrorsState = ValidationErrors
+
+const PASSWORD: PasswordField = 'user[password]'
```
josemigallas: I think assigning types to constants is a bit excessive! `PASSWORD` exists just to avoid repetition, types should be used with params and variables.

app/javascript/src/ChangePassword/components/ChangePasswordHooks.jsx
```diff
@@ -1,87 +1,95 @@
+// @flow
+
 import { useState } from 'react'
-import { validateForm, validateSingleField } from 'LoginPage/utils/formValidation'
+import validate from 'validate.js' // TODO: including this since validateSingleField and validateForm from LoginPage/utils/formValidation screams for a refactor
+
+type PasswordField = 'user[password]'
+type PasswordConfirmationField = 'user[password_confirmation]'
+type FieldName = PasswordField | PasswordConfirmationField
+type Constraints = {[FieldName]: {[string]: {}}}
```
josemigallas: This could be more specific:
```
type Constraint = 'presence' | 'length' | 'equality'
type Constraints = { [FieldName]: { [Constraint]: {} } }

File: app/javascript/src/ChangePassword/components/ChangePasswordHooks.jsx (hunk 1 above)
Comment (didierofrivia) [reply app/javascript/src/ChangePassword/components/ChangePasswordHooks.jsx]: Well, in Haskell or Elm, when you manage specific strings, those are recommended to be types, not strings. Like in this case, it's not any string the field name, it's a specific string. Any other string won't work, and will crash the app when the form is submited.

File: app/javascript/src/ChangePassword/components/ChangePasswordHooks.jsx (hunk 2 above)
Comment (didierofrivia) [reply app/javascript/src/ChangePassword/components/ChangePasswordHooks.jsx]: Yep, it could definitely be more specific. Will change

**Comment by didierofrivia** [comment]
@damianpm @thomasmaas @cntlsn take a look again, I've added the pristine condition

File: app/javascript/src/ChangePassword/components/ChangePasswordHooks.jsx (hunk 1 above)
Comment (josemigallas) [reply app/javascript/src/ChangePassword/components/ChangePasswordHooks.jsx]: I know what you mean and usually Flow does it for you, it's implied. CONSTANTS (upper case) are meant to avoid repetition, to declare absolute and generic content, they are type agnostic. Types are used to watch for inputs, unexpected inputs. You wouldn't declare:
```
const PORT: number = 3000
```
It doesn't matter what type it is, its value is absolute and unchangeable. What you would do is:
```
function startServer (port: number); // any number will suffice, will throw if PORT wouldn't be a number
function startServer (port: 3000); // it can only be 3000, will throw if PORT is other number than 3000
```
Defining a custom type is redundant:
```
type Port = 3000
const PORT: Port = 3000
```

File: app/javascript/src/ChangePassword/components/ChangePasswordHooks.jsx [hunk 3]
```diff
@@ -1,87 +1,103 @@
...
+import validate from 'validate.js' // TODO: including this since validateSingleField and validateForm from LoginPage/utils/formValidation screams for a refactor
+
+type PasswordField = 'user[password]'
+type PasswordConfirmationField = 'user[password_confirmation]'
+type FieldName = PasswordField | PasswordConfirmationField
+type ValidatorName = 'presence' | 'length' | 'equality'
+type ValidatorOption = {message: string, minimun?: number, attribute?: string}
+type Constraints = {[FieldName]: {[ValidatorName]: ValidatorOption}}
+type FieldState = {[FieldName]: string}
+type ValidationErrors = {[FieldName]: string[]}
```
josemigallas: This should be an optional type, since validate can return nothing (undefined).

app/javascript/src/ChangePassword/components/ChangePasswordHooks.jsx
```diff
@@ -1,87 +1,103 @@
...
-      setValidationStarted(initValidation)
-    }
-    if (initValidation || validationStarted) {
-      setIsPasswordConfirmationValid(!passwordConfirmationError)
-    }
-    setConfirmationErrorMessage(errorMessage)
-    setIsFormDisabled(!!errors)
+  const onFormChange = (event: SyntheticInputEvent<HTMLInputElement>) => {
+    const validationErrors: ValidationErrors = validate(event.currentTarget, validationConstraints)
+    const visibleErrors = !!validationErrors && Object.keys(validationErrors)
```
josemigallas: I'd watch out for the `!!` trick. The correct way would be `validationErrors !== undefined`.

app/javascript/src/ChangePassword/components/ChangePasswordHooks.jsx
```diff
@@ -1,87 +1,103 @@
...
+      .filter(key => !areFieldsPristine[key])
+      .reduce((errors, key) => ({[key]: validationErrors[key]}), {})
+    setFieldErrors(visibleErrors ? {...fieldErrorsTemplate, ...visibleErrors} : fieldErrorsTemplate)
```
josemigallas: `[]` is truthy, maybe `visibleErrors && visibleErrors.length` would be more accurate (although too much verbose maybe)

app/javascript/src/ChangePassword/components/ChangePasswordHooks.jsx
```diff
@@ -1,87 +1,103 @@
...
+    setIsFormDisabled(!!validationErrors)
```
josemigallas:
```suggestion
    setIsFormDisabled(validationErrors !== undefined)
```

app/javascript/src/ChangePassword/components/ChangePasswordHooks.jsx
```diff
@@ -1,87 +1,103 @@
...
-const comparePasswordsLength = (event) => {
-  const passwordInput = event.currentTarget['user_password']
-  const passwordConfirmationInput = event.currentTarget['user_password_confirmation']
-  return passwordConfirmationInput.value.length >= passwordInput.value.length
+const fieldErrorsTemplate: FieldErrorsState = {
+  [PASSWORD]: [''],
+  [PASSWORD_CONFIRMATION]: ['']
 }
 
+const extractFirstErrorMessage = (errorMessageArray: string[]): string => errorMessageArray.slice(0, 1).pop()
```
josemigallas: Why not simply `errorMessageArray[0]`?

app/javascript/src/ChangePassword/components/ChangePasswordHooks.jsx
```diff
@@ -1,87 +1,103 @@
...
+const isFieldValid = (errorMessageArray: string[]): boolean => !errorMessageArray.slice().pop().length
```
josemigallas: Why not `errrorMessageArray.length > 0`?

app/javascript/src/ChangePassword/components/ChangePasswordHooks.jsx
didierofrivia: Well... Flow actually doesn't do it for you. But if we follow the old C convention of symbolic constants, it makes sense. This is defined in our file, we don't get it from external input. Will change

app/javascript/src/ChangePassword/components/ChangePasswordHooks.jsx
didierofrivia: `validate` fn might return ValidationErrors or undefined, better using it in the fn call not the type definition

app/javascript/src/ChangePassword/components/ChangePasswordHooks.jsx
```diff
@@ -1,87 +1,103 @@
...
+  [PASSWORD]: '',
+  [PASSWORD_CONFIRMATION]: ''
 }
 
-const comparePasswordsLength = (event) => {
-  const passwordInput = event.currentTarget['user_password']
-  const passwordConfirmationInput = event.currentTarget['user_password_confirmation']
-  return passwordConfirmationInput.value.length >= passwordInput.value.length
+const fieldErrorsTemplate: FieldErrorsState = {
+  [PASSWORD]: [''],
```
josemigallas: I think the default state is no errors:
```suggestion
  [PASSWORD]: [],
```

app/javascript/src/ChangePassword/components/ChangePasswordHooks.jsx
didierofrivia: I'll add the type, but `visibleErrors` could either be `false` or `ValidationErrors`, and any Object is `truthy`

app/javascript/src/ChangePassword/components/ChangePasswordHooks.jsx
didierofrivia: The "explicit" way you mean

app/javascript/src/ChangePassword/components/ChangePasswordHooks.jsx
didierofrivia: Yes, now it might make more sense to access the array. Thing is in a previous version I had it as part of a piped set of functions accessing the `errors` array

app/javascript/src/ChangePassword/components/ChangePasswordHooks.jsx
didierofrivia: Same as here https://github.com/3scale/porta/pull/1748#discussion_r390863797 hehe

app/javascript/src/ChangePassword/components/ChangePasswordHooks.jsx
```diff
@@ -1,87 +1,103 @@
...
+  const [fieldValues, setFieldValues] = useState(fieldsTemplate)
+  const [fieldErrors, setFieldErrors] = useState(fieldErrorsTemplate)
   const [isFormDisabled, setIsFormDisabled] = useState(true)
-  const [passwordValue, setPasswordValue] = useState('')
-  const [isPasswordValid, setIsPasswordValid] = useState(undefined)
-  const [passwordConfirmationValue, setPasswordConfirmationValue] = useState('')
-  const [isPasswordConfirmationValid, setIsPasswordConfirmationValid] = useState(undefined)
-  const [confirmationErrorMessage, setConfirmationErrorMessage] = useState('isMandatory')
-  const [validationStarted, setValidationStarted] = useState(false)
+  const [areFieldsPristine, setAreFieldsPristine] = useState({[PASSWORD]: true, [PASSWORD_CONFIRMATION]: true})
```
josemigallas: Obviously not a big deal but I wonder why this state doesn't deserver a "template" like the others!
```suggestion
  const pristineTemplate = {
    [PASSWORD]: true,
    [PASSWORD_CONFIRMATION]: true
  }
  const [areFieldsPristine, setAreFieldsPristine] = useState(pristineTemplate)
```

app/javascript/src/ChangePassword/components/ChangePasswordHooks.jsx
damianpm: `visibleErrors` is an object (when errors), `visibleErrors.length` won't work

app/javascript/src/ChangePassword/components/ChangePasswordHooks.jsx
josemigallas: `validate` returns either an Object or `undefined`, the problem is that `!!{} === true` and we can't assure validate will never return that, either by error or by a change in their API.

So IMO here the verbose option is justified.

app/javascript/src/ChangePassword/components/ChangePasswordHooks.jsx
josemigallas: It's not 100% equivlent: https://github.com/3scale/porta/pull/1748#discussion_r391064402

app/javascript/src/ChangePassword/components/ChangePasswordHooks.jsx
didierofrivia: That's why our type system is for, it return either `ValidationErrors` or `undefined`. If they change the API we will know and we are safe since we have our deps locked. If it fails because of that we can update our types and make the changes. 
The only way I'm making this change, if we update the types accordingly.

didierofrivia:
@damianpm @thomasmaas @cntlsn I've added the condition that confirmation needs to start validating not straight away when typing, but when is at least the same length of the password

didierofrivia:
I've rebased in order to fix the unit test failing

PR #1754, opened 2020-03-04
Fix HTTP BasicAuth when using ProviderKey

duduribeiro:

We currently have a bug when apicast tries to fetch the proxy configs
for a service using ProviderKey via BasicAuth. It returns a forbidden
status.

This happened because we changed in #1736 to consider the access_token
when sent in HttpBasicAuth but we only considered if the token was
present.

This commit changes this behave to also consider if the AccessToken is
present, otherwise, it will tries to make the authentication using the
provider key.

[THREESCALE-4625](https://issues.redhat.com/browse/THREESCALE-4625)

### DIFF
--- a/app/lib/api_authentication/by_access_token.rb
+++ b/app/lib/api_authentication/by_access_token.rb
@@ -104,6 +104,7 @@ module ApiAuthentication
 
     def authenticated_token
       return @authenticated_token if instance_variable_defined?(:@authenticated_token)
+
       @authenticated_token = domain_account.access_tokens.find_from_value(access_token) if access_token
     end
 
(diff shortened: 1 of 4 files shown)


app/lib/api_authentication/by_access_token.rb
```diff
@@ -112,7 +113,7 @@ def enforce_access_token_permission
     end
 
     def verify_access_token_scopes
-      return true unless access_token
+      return true unless authenticated_token
```
hallelujah: I think this should not be modified because there is nothing to verify when it is a provider_key

app/lib/api_authentication/by_access_token.rb
```diff
@@ -121,7 +122,7 @@ def verify_access_token_scopes
     end
 
     def verify_write_permission
-      return true unless access_token
+      return true unless authenticated_token
```
hallelujah: Same as https://github.com/3scale/porta/pull/1754/files#r388151992

PR #1757, opened 2020-03-05
Use consistent absolute paths

josemigallas:

`create-react-application` do not support aliasing (that is, giving alias to specific folders) and that means the current path approach we are using for our test suite can't be used in the src.

The solution is to  only use the baseUrl in `ts-config` pointing to `src`. This way we can import everything from their source folders: `assets/`, `components/`, `types/`, `pages/`,  `tests/`. 

[THREESCALE-4663: Use absolute paths in imports](https://issues.redhat.com/browse/THREESCALE-4663)

**Verification steps** 

* `SKIP_PREFLIGHT_CHECK=true yarn build` works
* `SKIP_PREFLIGHT_CHECK=true yarn test` works
* IDE intellisense works


didierofrivia:
I can live without fancy aliases. 👍

PR #1758, opened 2020-03-05
[Sphinx] Do not index Topic for Oracle

Martouta:
Closes [THREESCALE-4652](https://issues.redhat.com/browse/THREESCALE-4652)

Topic has been deprecated for 3 years and it is SaaS-only.
We do not even run tests of Topic for Oracle:
https://github.com/3scale/porta/blob/670411d5bdcba562da6d413446c521bf7e9c598b/test/unit/topic_test.rb#L4-L6

So this PR does not index the topic.

image

image

### DIFF
--- a/app/indices/topic_index.rb
+++ b/app/indices/topic_index.rb
@@ -1,16 +1,21 @@
-ThinkingSphinx::Index.define(:topic,
-                             with: :active_record,
-                             delta: ThinkingSphinx::Deltas::DatetimeDelta,
-                             delta_options: {
-                               threshold: SPHINX_DELTA_INTERVAL,
-                               column: :last_updated_at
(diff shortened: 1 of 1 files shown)


app/indices/topic_index.rb
```diff
@@ -3,14 +3,16 @@
                              delta: ThinkingSphinx::Deltas::DatetimeDelta,
                              delta_options: {
                                threshold: SPHINX_DELTA_INTERVAL,
-                               column: :last_updated_at
-                             }) do
+                             }.merge(System::Database.oracle? ? {} : {column: :last_updated_at})
```
hallelujah: Maybe just skip the indexation for good if Oracle

PR #1761, opened 2020-03-06
[Bugfix & Refactor] Improve validation of URIs.

Martouta:
Closes [THREESCALE-4658](https://issues.redhat.com/browse/THREESCALE-4658)
Fix requested from [ [#1515](https://github.com/3scale/porta/pull/1515) | [THREESCALE-2932](https://issues.redhat.com/browse/THREESCALE-2932) ]

- Add tests for scenarios that were not tested and fix/improve existing tests.
- Respond with 'too_long' message when the label or the host is too long.
- Refactor UriValidator.

TODO in another PR: https://github.com/3scale/porta/pull/1761#discussion_r390632248

### DIFF
--- a/app/validators/uri_validator.rb
+++ b/app/validators/uri_validator.rb
@@ -1,53 +1,102 @@
 # frozen_string_literal: true
 
 class UriValidator < ActiveModel::EachValidator
-  DEFAULT_ACCEPTED_SCHEMES = /^https?$/
-  DEFAULT_OPTIONAL_PARTS = %i[port].freeze
-  DEFAULT_FORBIDDEN_PARTS = %i[userinfo registry path opaque query fragment].freeze
+  DEFAULT_ACCEPTED_SCHEME = /^https?$/
+  DEFAULT_PERMISSIONS_OF_PARTS = {
+    port:     true,
+    userinfo: false,
+    registry: false,
+    path:     false,
+    opaque:   false,
+    query:    false,
+    fragment: false
+  }.freeze
 
-  def validate_each(record, attribute, value)
-    valid = begin
-              uri = URI.parse(value)
-              valid_scheme?(uri.scheme) && valid_host?(uri.host) && !forbidden_part?(record, uri)
-            rescue URI::InvalidURIError
-              false
-            end
-    record.errors.add(attribute, options[:message] || :invalid) unless valid
-    valid
+  def validate_each(record, attribute_name, attribute_value)
+    uri_3scale_compliance_checker = UriThreeScaleComplianceChecker.new(
+      uri: self.class.safe_parse_uri(attribute_value),
+      permissions_parts: permissions_for_record(record),
+      accepted_scheme: options.fetch(:scheme, DEFAULT_ACCEPTED_SCHEME)
+    )
+
+    errors = uri_3scale_compliance_checker.errors
+    errors.each do |error|
+      record.errors.add(attribute_name, error)
+    end
+    errors.empty?
   end
 
-  def valid_scheme?(scheme)
-    accepted_scheme = options.fetch(:scheme, DEFAULT_ACCEPTED_SCHEMES)
-    case accepted_scheme
-    when Regexp
-      scheme =~ accepted_scheme
-    else
-      [*accepted_scheme].include? scheme
+  # This smells :reek:ManualDispatch
+  def permissions_for_record(record)
+    permissions = options.slice(*DEFAULT_PERMISSIONS_OF_PARTS.keys).reverse_merge(DEFAULT_PERMISSIONS_OF_PARTS)
+    permissions.transform_values do |option_value|
+      option_value.respond_to?(:call) ? option_value.call(record) : option_value.presence
     end
   end
 
-  def valid_host?(host)
-    return false if host.blank?
-    (host.size <= 255) && valid_host_labels?(host)
+  def self.safe_parse_uri(value)
+    URI.parse(value)
+  rescue URI::InvalidURIError
+    nil
   end
 
-  def valid_host_labels?(host)
-    host.split('.').map(&:size).none?(&63.method(:'<'))
-  end
+  class UriThreeScaleComplianceChecker
+    MAX_HOST_SIZE = 255
+    MAX_LABEL_SIZE = 63
 
-  def forbidden_part?(record, uri)
-    forbidden_parts = DEFAULT_FORBIDDEN_PARTS.reject { |part| truthy?(record, options[part]) }
-    forbidden_parts += DEFAULT_OPTIONAL_PARTS.select { |part| options.key?(part) && falsy?(record, options[part]) }
-    forbidden_parts.any? { |forbidden_attr| uri.public_send(forbidden_attr).present? }
-  end
+    def initialize(uri:, permissions_parts:, accepted_scheme:)
+      @uri = uri
+      @permissions_parts = permissions_parts
+      @accepted_scheme = accepted_scheme
+    end
 
-  protected
+    attr_reader :uri, :permissions_parts, :accepted_scheme, :generic_error_message
+    delegate :host, :scheme, to: :uri
 
-  def truthy?(record, part)
-    !!(part.respond_to?(:call) ? record.instance_eval(&part) : part.present?)
-  end
+    def errors
+      return [generic_error_message] if uri.blank?
 
-  def falsy?(record, part)
-    !truthy?(record, part)
+      errors_scheme | errors_host | errors_forbidden_parts
+    end
+
+    private
+
+    def errors_scheme
+      valid = case accepted_scheme
+              when Regexp
+                scheme =~ accepted_scheme
+              else
+                [*accepted_scheme].include?(scheme)
+      end
+
+      valid ? [] : [:invalid]
+    end
+
+    def errors_host
+      return [:invalid] if host.blank?
+
+      errors = []
+      errors << I18n.t('errors.messages.too_long', count: MAX_HOST_SIZE) unless valid_host_size?
+      errors << I18n.t('errors.messages.host_label_too_long', count: MAX_LABEL_SIZE) unless valid_host_labels?
+      errors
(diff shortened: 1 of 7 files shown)


app/validators/uri_validator.rb
```diff
@@ -43,6 +51,40 @@ def forbidden_part?(record, uri)
...
+    record.errors[attribute] << :blank and return false if host.blank?
+
+    valid_host_size = valid_host_size?(host)
+    valid_host_labels = valid_host_labels?(host)
+    record.errors[attribute] << I18n.t('activerecord.errors.messages.too_long', count: MAX_HOST_SIZE) unless valid_host_size
+    record.errors[attribute] << I18n.t('activerecord.errors.messages.host_label_too_long', count: MAX_LABEL_SIZE) unless valid_host_labels
+    valid_host_size && valid_host_labels
+  end
+
+  def valid_host_size?(host)
```
josemigallas: Put this before `valid_host_labels` for a better readability.

app/validators/uri_validator.rb
```diff
@@ -4,16 +4,24 @@ class UriValidator < ActiveModel::EachValidator
...
+
+    begin
+      uri = URI.parse(value)
+    rescue URI::InvalidURIError
+      add_generic_failure_message(record, attribute) and return false
+    end
+
+    # It's done this way to get all the errors even if it is already known that is invalid
+    [
+      validate_scheme(record, attribute, uri.scheme),
```
josemigallas: Wouldn't `uri` be `nil` when rescuing `URI::InvalidURIError`?

Martouta: @josemigallas yes, that is why it has a return before reaching here 😄

app/validators/uri_validator.rb
Martouta: @josemigallas i do not want to make this method public unnecessarily... in fact, if we can make the other private, even better... but i didn't have time to check if that is possible as i said in the description

Martouta:
### Old description

I am not entirely proud with the result (too much parameter passing and we would need a new class to make it cleaner + I didn't touch some methods because they are public and I do not know if any other class relies on them) but I am aware that we are in a hurry for CR1.

I also find this a bit risky because it affects more models, plus we need more tests for the error message of each scenario.

### Change of plans

Thanfully, this won't be for CR1 anymore, so I will do it properly instead 😄

app/validators/uri_validator.rb
```diff
@@ -1,53 +1,102 @@
...
+      path:     false,
+      opaque:   false,
+      query:    false,
+      fragment: false
+    }.freeze
+    MAX_HOST_SIZE = 255
+    MAX_LABEL_SIZE = 63
 
-  def valid_host_labels?(host)
-    host.split('.').map(&:size).none?(&63.method(:'<'))
```
guicassolato: This didn't mean that each label could be up to 62-characters long. It called `63 < each_part_of_the_host.size`.

Martouta: Ah right 👍 sorry 🙏 i will fix that 😄

config/locales/en.yml
```diff
@@ -766,6 +766,7 @@ en:
 
   errors:
     messages:
+      host_label_too_long: is too long for one or more labels of the host (maximum is %{count} characters)
```
guicassolato: This message will be completely enigmatic if shown in the creation of a product because the system name is too long.

app/validators/uri_validator.rb
```diff
@@ -1,53 +1,102 @@
...
 
-  def valid_scheme?(scheme)
-    accepted_scheme = options.fetch(:scheme, DEFAULT_ACCEPTED_SCHEMES)
-    case accepted_scheme
-    when Regexp
-      scheme =~ accepted_scheme
-    else
-      [*accepted_scheme].include? scheme
+  # This smells :reek:ManualDispatch
+  def self.options_for_record(options, record)
```
guicassolato:
```suggestion
  def self.options_for_record(options, record)
    return options.slice(:scheme, *UriPartsValidator::DEFAULT_PERMISSIONS_OF_PARTS.keys).transform_values { |value| value.respond_to(:call) ? value.call(record) : value.presence }
```

The `return` is only to say that we don't need the other lines below.

app/validators/uri_validator.rb
```diff
@@ -1,53 +1,102 @@
...
+    options.each_with_object({}) do |(options_name, options_value), opts|
+      opts[options_name] = options_value.respond_to?(:call) ? options_value.call(record) : options_value
     end
   end
 
-  def valid_host?(host)
-    return false if host.blank?
-    (host.size <= 255) && valid_host_labels?(host)
-  end
+  class UriPartsValidator
```
guicassolato: I was thinking... so now `UriValidator` is the actual `ActiveModel::Validator`. It uses this class that does all the hard work – e.g. parsing the value into "parts" of an URI, actually validating each part according to the options and the RFC. Then `UriValidator` only adds the error message. Do we really need two classes? Is it all "URI validation"? 🤔

Martouta: > Do we really need two classes? 

I do not think that we ever really **need** more than one single file for the entire program to work, but to maintain it is another story... same here, **need** is a too strong word and I didn't add it out of necessity but simply because I believe that makes our lives easier.
That class is not mutable and does not have side effects either. It is just one input that implies one output, always, idempotently, no matter how many times it is called. At the same time, not so many parameters being passed everywhere, but just relying on its instance variables, which aren't many.

> Is it all "URI validation"?

I think that this is the problem and what actually makes it confusing... that I didn't write a proper name and it seems like both are "URI validation". I definitely have to fix that :+1:

config/locales/en.yml
Martouta: That's not what the error would be, but because it uses the ServiceCreator.
It should be done in a cleaner way. I will definitely clean that :smile:

app/validators/uri_validator.rb
```diff
@@ -1,53 +1,102 @@
...
-    valid = begin
-              uri = URI.parse(value)
-              valid_scheme?(uri.scheme) && valid_host?(uri.host) && !forbidden_part?(record, uri)
-            rescue URI::InvalidURIError
-              false
-            end
-    record.errors.add(attribute, options[:message] || :invalid) unless valid
-    valid
+  def validate_each(record, attribute_name, attribute_value)
+    errors = UriPartsValidator.validate(attribute_value, self.class.options_for_record(options, record))
```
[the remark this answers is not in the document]
Martouta: I completely agree. For that and because sending a hash with options opens the door to sending almost anything. That is keeping complexity that we do not need. So I will change that :+1:

app/validators/uri_validator.rb
```diff
@@ -1,53 +1,102 @@
...
+      opts[options_name] = options_value.respond_to?(:call) ? options_value.call(record) : options_value
     end
   end
 
-  def valid_host?(host)
-    return false if host.blank?
-    (host.size <= 255) && valid_host_labels?(host)
-  end
+  class UriPartsValidator
+    DEFAULT_ACCEPTED_SCHEMES = /^https?$/
```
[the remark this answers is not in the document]
Martouta: Right :+1:

app/validators/uri_validator.rb
```diff
@@ -1,53 +1,102 @@
...
+    def validate
+      return [generic_error_message] if uri.blank?
+
+      errors_scheme | errors_host | errors_forbidden_parts
+    end
+
+    private
+
+    def errors_scheme
+      accepted_scheme = options.fetch(:scheme, DEFAULT_ACCEPTED_SCHEMES)
```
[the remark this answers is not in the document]
Martouta: https://github.com/3scale/porta/pull/1761#discussion_r390633000

app/validators/uri_validator.rb
```diff
@@ -1,53 +1,102 @@
...
-  def falsy?(record, part)
-    !truthy?(record, part)
+    def validate
+      return [generic_error_message] if uri.blank?
+
+      errors_scheme | errors_host | errors_forbidden_parts
+    end
+
+    private
+
```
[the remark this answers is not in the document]
Martouta: Right but without the `allow_nil: true` :smile:

app/validators/uri_validator.rb
```diff
@@ -1,53 +1,102 @@
...
+                scheme =~ accepted_scheme
+              else
+                [*accepted_scheme].include?(scheme)
+      end
+
+      valid ? [] : [generic_error_message]
+    end
+
+    def errors_host
+      return [generic_error_message] if uri.host.blank?
```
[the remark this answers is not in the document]
Martouta: https://github.com/3scale/porta/pull/1761#discussion_r390633562

app/validators/uri_validator.rb
```diff
@@ -1,53 +1,102 @@
...
+
+      errors = []
+      errors << I18n.t('errors.messages.too_long', count: MAX_HOST_SIZE) unless valid_host_size?
+      errors << I18n.t('errors.messages.host_label_too_long', count: MAX_LABEL_SIZE) unless valid_host_labels?
+      errors
+    end
+
+    def valid_host_size?
+      uri.host.size <= MAX_HOST_SIZE
```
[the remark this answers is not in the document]
Martouta: https://github.com/3scale/porta/pull/1761#discussion_r390633562

app/validators/uri_validator.rb
```diff
@@ -1,53 +1,102 @@
...
+    end
+
+    def valid_host_labels?
+      uri.host.split('.').map(&:size).none?(&MAX_LABEL_SIZE.method(:'<'))
```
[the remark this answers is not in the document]
Martouta: https://github.com/3scale/porta/pull/1761#discussion_r390633562

app/validators/uri_validator.rb
Martouta: Great one :+1: :clap:

PR #1762, opened 2020-03-06
Small Fixes

josemigallas:
* Removes a relative path that wasn't changed in #1757 
~~* Specify props for the `LazyRoute` component, removing "prop spreading" pattern.~~
* Fixes `portafly/src/pages/index.tsx` by importing from default (instead of named) exports.

**Verification steps** 

* yarn start works
* yarn tests is green

### DIFF
--- a/portafly/src/App.tsx
+++ b/portafly/src/App.tsx
@@ -37,8 +37,8 @@ const navItems = [
   }
 ]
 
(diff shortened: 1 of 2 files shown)


josemigallas:
You're right @didierofrivia the changes in `LazyRoute.tsx` could go in the linter PR. The others however fix #1757

PR #1763, opened 2020-03-06
Add linter (Airbnb rules version)

josemigallas:

Adds `eslint` and `stylelint` to watch out our code.

This PR is following [Airbnb](https://github.com/airbnb/javascript)'s recommended rules (the one I am personally more fond of) however there will be another with standardjs rules, so that they can be compared.

Most of this rules can be fixed automatically by eslint.

[THREESCALE-4615: Add an opinionated linter](https://issues.redhat.com/browse/THREESCALE-4615)

**Verification steps** 

* Verify eslint and stylelint rules are used by your IDE

**Notes for the reviewers**
Some rules have been overridden due to team's preferences and familiarity:
- **Single quotes**
- **No semicolon**
- Prefer **relative exports**

### DIFF
--- a/portafly/.eslintrc.js
+++ b/portafly/.eslintrc.js
@@ -1,0 +1,30 @@
+module.exports = {
+  root: true,
+  plugins: [
+    'jest'
+  ],
+  extends: [
+    'airbnb-typescript',
+    'plugin:jest/recommended'
+  ],
+  parserOptions: {
+    project: './tsconfig.json',
+  },
+  rules: {
+    'semi': ['error', 'never'],
+    'comma-dangle': ['error', 'never'],
+    'no-prototype-builtins': 1,
+
+    // import
+    'import/no-extraneous-dependencies': 1, // Because of @testing-library
+    'import/no-default-export': 2,
+    'import/prefer-default-export': 0,
+    'import/no-unresolved': 0, // Need to add a resolver probably https://github.com/benmosher/eslint-plugin-import/blob/master/README.md#resolvers
+
+    // typescript
+    '@typescript-eslint/semi': 0,
+
+    // react
+    'react/prop-types': 0,
+  }
+}
--- a/portafly/package.json
+++ b/portafly/package.json
@@ -33,14 +33,17 @@
     "eslint-plugin-jest": "^23.8.2",
     "eslint-plugin-jsx-a11y": "^6.2.3",
     "eslint-plugin-react": "^7.19.0",
-    "eslint-plugin-react-hooks": "^2.5.0"
+    "eslint-plugin-react-hooks": "^2.5.0",
+    "stylelint": "^13.2.0",
+    "stylelint-config-sass-guidelines": "^7.0.0",
+    "stylelint-config-standard": "^20.0.0"
   },
   "scripts": {
     "start": "react-scripts start",
     "build": "react-scripts build",
     "test": "react-scripts test",
     "test-with-coverage": "yarn test -- --coverage",
-    "lint": "eslint . --ext .js,.jsx,.ts,.tsx",
+    "lint": "stylelint 'src/**/*.?css' --allow-empty-input && eslint . --ext .js,.jsx,.ts,.tsx",
     "eject": "react-scripts eject"
   },
   "jest": {
--- a/portafly/package.json
+++ b/portafly/package.json
@@ -5,6 +5,7 @@
   "dependencies": {
     "@patternfly/react-core": "^3.124.1",
     "@patternfly/react-table": "^2.25.6",
+    "date-fns": "^2.11.0",
     "react": "^16.12.0",
     "react-async": "^10.0.0",
     "react-dom": "^16.12.0",
@@ -26,7 +27,6 @@
     "@types/react-router-dom": "^5.1.3",
     "@typescript-eslint/eslint-plugin": "^2.23.0",
     "@typescript-eslint/parser": "^2.22.0",
-    "date-fns": "^2.9.0",
     "eslint": "^6.8.0",
     "eslint-config-airbnb-typescript": "^7.0.0",
     "eslint-plugin-import": "^2.20.1",
--- a/portafly/package.json
+++ b/portafly/package.json
@@ -24,13 +24,23 @@
     "@types/react-dom": "^16.9.0",
     "@types/react-router": "^5.1.3",
     "@types/react-router-dom": "^5.1.3",
-    "date-fns": "^2.9.0"
+    "@typescript-eslint/eslint-plugin": "^2.23.0",
+    "@typescript-eslint/parser": "^2.22.0",
+    "date-fns": "^2.9.0",
+    "eslint": "^6.8.0",
+    "eslint-config-airbnb-typescript": "^7.0.0",
+    "eslint-plugin-import": "^2.20.1",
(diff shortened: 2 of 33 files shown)


portafly/package.json
```diff
@@ -24,13 +24,26 @@
...
+    "eslint": "^6.8.0",
+    "eslint-config-airbnb-typescript": "^7.0.0",
+    "eslint-plugin-import": "^2.20.1",
+    "eslint-plugin-jest": "^23.8.1",
+    "eslint-plugin-jsx-a11y": "^6.2.3",
+    "eslint-plugin-react": "^7.19.0",
+    "eslint-plugin-react-hooks": "^2.5.0",
+    "stylelint": "^13.2.0",
+    "stylelint-config-sass-guidelines": "^7.0.0",
+    "stylelint-config-standard": "^20.0.0"
```
didierofrivia: So most of these dependencies are not needed for StandardJS?

josemigallas: Stylelint is not Eslint-related (I simply haven't added to the other PR yet) and the rest of dependencies come already with Standard. Some like `jest` and `hooks` are not used by standard.

portafly/src/App.tsx
```diff
@@ -17,60 +19,52 @@ const navItems = [
...
   {
     title: 'Integration',
     to: '/integration',
     items: [
-      {to: '/integration/configuration', title: 'Configuration'}
-    ]
-  }
+      { to: '/integration/configuration', title: 'Configuration' },
+    ],
+  },
```
didierofrivia: Last items in a list (array, obj, etc) should not enforce trailing commas.

portafly/src/App.tsx
```diff
@@ -17,60 +19,52 @@ const navItems = [
...
-        exact
-      />
-    </SwitchWith404>
-  )
-}
+const PagesSwitch = () => (
+  <SwitchWith404>
+    <LazyRoute path="/" exact getComponent={getOverviewPage} />
+    <LazyRoute path="/applications" exact getComponent={getApplicationsPage} />
+    <Redirect path="/overview" to="/" exact />
```
didierofrivia: what's the rule for props here? or is it the standard 80 char long to line break?

portafly/.eslintrc.js
```diff
@@ -0,0 +1,33 @@
+module.exports = {
+  root: true,
+  parser: '@typescript-eslint/parser',
+  plugins: [
+    'jest'
+  ],
+  extends: [
+    'airbnb-typescript',
+    // 'plugin:jest/recommended'
```
damianpm: Is this plugin needed?

portafly/src/tests/components/util/useA11yRoute.test.tsx
```diff
@@ -11,9 +13,9 @@ const SamplePage = () => {
     setFocused(true)
   }
   return (
-    <div id={'test-focus'} tabIndex={-1} onFocus={handleFocus}>
+    <div id="test-focus" tabIndex={-1} onFocus={handleFocus}>
       {focused ? (
-        <div data-testid={'test-focused'}>I'm focused!</div>
+        <div data-testid="test-focused">I&apos;m focused!</div>
```
didierofrivia: Hmmm, maybe better here to use template literals

portafly/.eslintrc.js
josemigallas: Oops, I might have left some comments there

portafly/src/App.tsx
```diff
@@ -1,12 +1,14 @@
...
-import {BrowserRouter as Router, Redirect, useHistory} from 'react-router-dom'
-import { AppLayout, SwitchWith404, LazyRoute } from 'components'
+import { Redirect, BrowserRouter as Router, useHistory } from 'react-router-dom'
+import { Brand } from '@patternfly/react-core'
+import { AppLayout, LazyRoute, SwitchWith404 } from 'components'
 import { LastLocationProvider } from 'react-router-last-location'
 import logo from 'assets/logo.svg'
 
-const Logo = <Brand src={logo} alt={'patternfly logo'}/>
+const Logo = <Brand src={logo} alt="patternfly logo" />
```
damianpm: Is there a reason to not always use single quotes?

portafly/src/App.tsx
```diff
@@ -17,60 +19,52 @@ const navItems = [
...
     ],
   },
   {
     title: 'Applications',
     to: '/applications',
     items: [
-      {to: '/applications', title: 'Listing'},
-      {to: '/applications/plans', title: 'Application Plans'}
+      { to: '/applications', title: 'Listing' },
+      { to: '/applications/plans', title: 'Application Plans' },
```
damianpm: Personally, I don't like this `comma-dangle` rule. 
It looks ugly having unncessary commas all around. The only advantage is when you see the diff.
Is not a big deal, if you both like it I can live with it

portafly/src/App.tsx
josemigallas: The reason is you would use doubles quotes in an HTML file and it's become a standard defacto to considered JSX the same way. There is no practical reason, I think.

portafly/src/App.tsx
josemigallas: The diff is precisely the reason why. I don't particularly like it either.. I think you can achieve the same effect by adding items at the top instead of at the bottom.

damianpm: In that case, I think we can remove it

portafly/src/App.tsx
damianpm: I'm not convinced with this argument of: use double quotes when html, use single with js. For me it's all js as we are using jsx.
But again, I can adapt to both styles

portafly/src/App.tsx
josemigallas: it's 100 chars I think.
1-3 props per line (below 100 chars or 80 whatever), otherwise break and 1 per line

portafly/src/App.tsx
josemigallas: ok

portafly/src/App.tsx
josemigallas: I'm not trying to convince anyone: it's just a standard people use, just like using camelCase instead of PascalCase. It comes from this kind of situation:
```js
<input value="user's choice"/>
```
vs
```js
<input value='user\'s choice'/>
```

portafly/src/App.tsx
josemigallas: OK

portafly/src/App.tsx
damianpm: We could use template strings, but maybe it's too much. OK, let's use double

portafly/src/tests/components/FormatDate.test.tsx
```diff
@@ -1,26 +1,101 @@
...
+
+    rerender(<FormatDate date={dateISOInt} options={options} />)
+    expect(queryByText(result)).toBeInTheDocument()
+
+    rerender(<FormatDate date={dateISODate} options={options} />)
+    expect(queryByText(result)).toBeInTheDocument()
   })
+})
+
+// FIXME: use date-fns-tz OR change library OR fix it somehow
```
didierofrivia: Open an issue about this. Since it's important also for I18N.

PR #1765, opened 2020-03-06
Create custom page for reset password

duduribeiro:

Our current request password reset page is the same as login page, only adding a custom query parameter and handled the display by the frontend. This was making hard for us to block access from Master provider in this page.

This commit creates a new page along with a new pack that renders only the RequestPasswordForm component. This also removes all code handling request password reset from LoginPageWrapper component.

 We  also changed the ProviderDomainConstraint to forbid MasterDomain to access routes only for providers.

[THREESCALE-4619](https://issues.redhat.com/browse/THREESCALE-4619)

Accessing Login at Provider:
image

Accessing request Password Reset at Provider:
image

Accessing Login at Master:
image

Accessing Request Password Reset at Master:
image


app/javascript/packs/ResetPasswordPage.js
```diff
@@ -0,0 +1,52 @@
+import '@babel/polyfill'
```
damianpm: @duduribeiro By convention, we don't render React components directly from the `javascript/packs` directory. In `packs` we only get from rails the props needed and pass them down.
The React components are located in the `javascript/src` directory and  we use a helper called `createReactWrapper`.
I'll create a PR pointing to this one to fix the issues.

app/views/provider/request_password_resets/new.html.slim
```diff
@@ -0,0 +1,8 @@
+= javascript_pack_tag 'ResetPasswordPage'
+- authentication_providers = (@authentication_providers || []).map { |ap| {authorizeURL: ap.authorize_url, humanKind: ap.human_kind} }
+- flash_messages = (flash || []).map { |f| {type: f[0], message: f[1]}}
```
damianpm: Are we expecting any error in this page?

app/views/provider/request_password_resets/new.html.slim
```diff
@@ -0,0 +1,8 @@
...
+- is_master_account = domain_account.master?
```
damianpm: Looks like `is_master_account` is never used

damianpm:
@duduribeiro I'm working on https://github.com/3scale/porta/pull/1766
I still have to fix js tests

app/views/provider/request_password_resets/new.html.slim
duduribeiro: hey @damianpm

yeah.. this page can have errors.

app/views/provider/request_password_resets/new.html.slim
duduribeiro: you are right 💯

app/controllers/provider/request_password_resets_controller.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+class Provider::RequestPasswordResetsController < FrontendController
```
hallelujah: There is already a password controller, why can't we use it?

https://github.com/3scale/porta/blob/8b27ca70d7700d9c275401439e44ea9ff6aa639e/app/controllers/provider/passwords_controller.rb

test/unit/domain_constraints_test.rb
```diff
@@ -83,6 +83,13 @@ class ProviderDomainConstraintTest < DomainConstraintsTest
       assert Account.exists?(self_domain: self_domain)
       assert ProviderDomainConstraint.matches?(request)
     end
+
+    test 'master domain' do
+      master = master_account
+      request = mock
+      request.expects(:host).returns(master.domain)
+      refute ProviderDomainConstraint.matches?(request)
+    end
```
hallelujah: Are you sure this goes in this PR?

app/controllers/provider/request_password_resets_controller.rb
duduribeiro: this passwords controller already has this custom logic for the `new` action that regenerates the token.

test/unit/domain_constraints_test.rb
duduribeiro: it can be in a different PR.. will split it

app/controllers/provider/request_password_resets_controller.rb
hallelujah: I know but can we reuse the same controller?

duduribeiro: @hallelujah I moved it to the already existent controller

app/controllers/provider/passwords_controller.rb
```diff
@@ -1,6 +1,6 @@
 class Provider::PasswordsController < FrontendController
 
-  skip_before_action :login_required, only: %i(destroy show update)
+  skip_before_action :login_required, only: %i(destroy show update request_password_reset)
```
Martouta: Change `(...)` for `[...]` plz 😄 

image

app/controllers/provider/passwords_controller.rb
```diff
@@ -47,6 +47,10 @@ def update
     end
   end
 
+  def request_password_reset
```
Martouta: what about calling it just `reset` ? it's inside the PasswordsController so no need to say that it is the password what is being reset... and the `request` word... all methods here are requests 🤔 it is like naming `update` as `request_password_update`, simply not needed 😄

config/routes.rb
```diff
@@ -165,6 +165,7 @@ def call(env)
 
   namespace :provider, path: 'p', constraints: ProviderDomainConstraint do
     resource :password, :only => %i(new show update destroy)
+    get '/request_password_reset', to: 'passwords#request_password_reset'
```
Martouta: Following https://github.com/3scale/porta/pull/1765/files#r391747885, it would be:
```ruby
resource :password, only: %i[new show update destroy] do
  get 'reset'
 end
```

config/routes.rb
```diff
@@ -164,8 +164,9 @@ def call(env)
   get '/auth/:system_name/bounce' => 'provider/sessions#bounce', constraints: ProviderDomainConstraint, as: :authorization_provider_bounce
 
   namespace :provider, path: 'p', constraints: ProviderDomainConstraint do
-    resource :password, :only => %i(new show update destroy)
-    get '/request_password_reset', to: 'passwords#request_password_reset'
+    resource :password, :only => %i(new show update destroy) do
```
Martouta:
```suggestion
    resource :password, only: %i[new show update destroy] do
```

🙈

hallelujah:
Awesome 🚀

app/javascript/packs/RequestPasswordPage.js
```diff
@@ -0,0 +1,11 @@
+import '@babel/polyfill'
+import 'core-js/es7/object'
+
+import { RequestPasswordWrapper as RequestPassword } from 'LoginPage'
+import { safeFromJsonString } from 'utilities/json-utils'
+
+document.addEventListener('DOMContentLoaded', () => {
+  const container = document.getElementById('pf-request-page-container')
```
josemigallas: Suggestion: extract `pf-request-page-container` into a const.

app/javascript/src/LoginPage/LoginPageWrapper.jsx
```diff
@@ -43,40 +39,21 @@ type Props = {
...
-  }
-
-  componentDidMount () {
-    this.setFormMode(window)
-  }
-
   showForgotCredentials () {
-    const showForgotCredentials = this.state.formMode === formModeTuple[0] && !this.props.disablePasswordReset
-    return showForgotCredentials && <ForgotCredentials providerLoginPath={this.props.providerLoginPath}/>
+    const showForgotCredentials = !this.props.disablePasswordReset
```
josemigallas: Here, use destructuring:
```suggestion
    const { disablePasswordReset, providerLoginPath } = this.props
    return !disablePasswordReset && <ForgotCredentials requestPasswordResetPath={providerLoginPath} />
```

app/javascript/src/LoginPage/RequestPasswordWrapper.jsx
```diff
@@ -0,0 +1,41 @@
+// @flow
```
josemigallas: File should be called `RequestPassword.jsx`.

duduribeiro: @damianpm wdyt ? 🤔

damianpm: We have other files in the same directory with the same 'Wrapper' suffix: `LoginPageWrapper.jsx` and `SignupPageWrapper.jsx` we should keep the same format or change all files.
In the scope of this task, I think we can keep same name, just to not affect files out of scope.

PR #1767, opened 2020-03-10
Removing IE11 code

damianpm:

As we are not supporting IE11 anymore we can delete some styles targeting IE11. We are not maintaining anymore those styles, then it has no sense to keep them

None


damianpm:
After realizing we still have a considerable number of clients using IE11 (even if we don't support it anymore and clients have been notified), we decided to postpone for later.

PR #1768, opened 2020-03-10
Add linter (standardjs version)

josemigallas:

Adds `eslint` and `stylelint` to watch out our code.

This PR is following [standardjs](https://standardjs.com/index.html)'s code style, an inflexible set of eslint rules that are designed to be used with Javascript only and unfortunately does not support very well Typescript and React.

[THREESCALE-4615: Add an opinionated linter](https://issues.redhat.com/browse/THREESCALE-4615)

**Important notes for reviewers**
The advantage of using Standard is that we don't worry about eslint anymore. The problems I've encountered are:
- Doesn't integrate well with VS Code (IDE)
- Doesn't enforce format as well as using Eslint (look for easter eggs in PR..)
- It doesn't support Typescript (it requires a workaround)

### DIFF
--- a/portafly/.eslintrc.js
+++ b/portafly/.eslintrc.js
@@ -1,0 +1,10 @@
+/**
+ * This file should NOT exist. There si however an issue with standard and Typescript that
+ * needs to be worked around: https://standardjs.com/index.html#typescript
+ */
+module.exports = {
+  rules: {
+    'no-unused-vars': 'off',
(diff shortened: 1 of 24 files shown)


didierofrivia:
Is there a reason why this linter added less than half the lines of Airbnb one?

josemigallas:
All the extra lines must come from `Stylelint` and some plugins like `jest` and `react-hooks` that Standard does not require (limiting its linting power)

PR #1772, opened 2020-03-11
Add a mock server for Development and a specialized API Spec

josemigallas:

* ~Introducing a mock server for development. The idea is to mock all possible requests to the backend so that we **analyze what we need**. The server is implemented using [JSON Server](https://github.com/typicode/json-server) for the following reasons:~
  - ~0 code needed for basic usage, only a JSON schema~
  - ~It allow for more complex behavior such as routes, middleware, authorization..~
  - ~It is well maintained and it is extensively used~
  - ~It is super cool~
  - Not anymore!

* Introducing instead [`open-api-mocker`](https://github.com/jormaechea/open-api-mocker) which is able to interpret an Openapi spec and mock an API. This will allow us both to keep developing (while being backend-agnostic) and to define in a clear way our API needs.

* Also introducing a data access layer (DAL) to abstract from the fetch library, from the route and all the network paraphernalia. We should only worry about getting, updating, etc stuff:
  ```javascript
  const { applications, isLoading } = useGetApplications()
  const { application, isLoading } = useGetApplication({ id: 1 })
  const { application, isLoading } = useUpdateApplication(newApplication)
  ```

* Also introducing a way to organize components: **by page**. General use components should be located under `common/` but that' for another moment to decide.

[THREESCALE-4627: Add a mock server to make fake API calls](https://issues.redhat.com/browse/THREESCALE-4627)


didierofrivia:
I'd rather define it with an OAS3 spec, than a plain json file. Thus we can be more flexible and make many other implementations having the spec.

josemigallas:
@didierofrivia I added a very basic spec.yaml and introduced a new library that will read it and act as a server.

damianpm:
Can you add some mock data, to see something different than 'string' in all fields?
apps

josemigallas:
@damianpm I wouldn't spend time on that since @didierofrivia is creating a new Spec that is supposed to replace this one.

portafly/src/tests/pages/Applications.test.tsx
```diff
@@ -0,0 +1,43 @@
+import * as React from 'react'
+
+import { render } from 'tests/setup'
+import { Applications } from 'pages'
+import { useGetApplications } from 'dal/Applications'
+import { getByLabelText } from '@testing-library/react'
```
damianpm, line 6: `getByLabelText` is never used

portafly/package.json
```diff
@@ -24,15 +24,18 @@
...
+    "date-fns": "^2.9.0",
+    "open-api-mocker": "^1.2.4"
   },
   "scripts": {
     "start": "react-scripts start",
     "build": "react-scripts build",
+    "mock-server": "open-api-mocker -s openapi.yml -p 3001 -w",
     "test": "react-scripts test",
     "test-with-coverage": "yarn test -- --coverage",
     "eject": "react-scripts eject"
```
damianpm, line 36: What happened to the "lint" script?

portafly/src/components/applications/ApplicationsTable.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { SimpleEmptyState } from 'components'
+import { IApplication } from 'types'
+
+export interface IApplicationsTable {
+  applications: IApplication[]
+}
+
+const ApplicationsTable: React.FunctionComponent<IApplicationsTable> = ({ applications }) => {
+  if (applications.length === 0) {
+    return <SimpleEmptyState msg="There are no Applications" />
```
damianpm, line 12: "There are no Applications" should come from strings file

portafly/src/components/EmptyState.tsx
```diff
@@ -0,0 +1,23 @@
...
+  EmptyStateVariant,
+  EmptyStateIcon,
+} from '@patternfly/react-core'
+import { CubesIcon } from '@patternfly/react-icons'
+
+export interface ISimpleEmptyStateProps {
+  msg: string
+}
+
+const SimpleEmptyState: React.FunctionComponent<ISimpleEmptyStateProps> = ({ msg }) => (
```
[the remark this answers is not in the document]
josemigallas: Not sure I follow, this `SimpleEmptyState` is shown when a list is empty, the other component `Loading` is for showing a loading state.

josemigallas:
Closing in favor of #1785

PR #1773, opened 2020-03-11
Avoid Master account to access routes with ProviderDomain constraint

duduribeiro:
Extracted from: https://github.com/3scale/porta/pull/1765/

Currently ProviderDomainConstraint has a bug that it also allows Master
to access it. We should only allow provider on this constraint because
we have another one called MasterOrProviderDomainConstraint to be used
when both can access a route.

This commit changes this behave forbidding MasterDomain to access routes
only for providers.

### DIFF
--- a/lib/routing_constraints.rb
+++ b/lib/routing_constraints.rb
@@ -26,7 +26,7 @@ module ProviderDomainConstraint
     request.extend(ThreeScale::DevDomain::Request) if ThreeScale::DevDomain.enabled?
 
     with_deleted = AuthenticatedSystem::Request.new(request).zync?
-    Account.providers.without_deleted(!with_deleted).exists?(:self_domain => request.host)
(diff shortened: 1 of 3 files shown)


lib/routing_constraints.rb
```diff
@@ -26,7 +26,7 @@ def matches?(request)
     request.extend(ThreeScale::DevDomain::Request) if ThreeScale::DevDomain.enabled?
 
     with_deleted = AuthenticatedSystem::Request.new(request).zync?
-    Account.without_deleted(!with_deleted).exists?(:self_domain => request.host)
+    Account.providers.without_deleted(!with_deleted).exists?(:self_domain => request.host)
```
Martouta:
```suggestion
    Account.tenants.without_deleted(!with_deleted).exists?(self_domain: request.host)
```

PR #1775, opened 2020-03-13
Adding a simple first version of an ongoing shaping OAS3 spec

didierofrivia:
In order to not only shaping but documenting the API endpoints we need to hydrate PortaFly, we need to use as base an OAS. This will also serve as a testing tool and development used as mocked data.
This PR also solves the first specification needed for Account > List

This spec is created with Apicurio and it's meant to be edited with it. More info in https://www.apicur.io/

Introduces an Open API Specification and address the specification needs described in https://issues.redhat.com/browse/THREESCALE-4731

https://issues.redhat.com/browse/THREESCALE-4711
https://issues.redhat.com/browse/THREESCALE-4731

BASED ON https://marvelapp.com/55343de/screen/67190787

**How it looks**
Screenshot 2020-03-19 at 14 41 57

**PLEASE, READ THE ISSUES ATTACH TO UNDERSTAND THIS PR, THERE'S DEFINITION OF WHAT'S NEEDED** First @hallelujah comment is obsolete, those issues are meant to cover what he expressed there.

And if you want to see the responses, etc... you might want to check it in online spec docs like redoc or locally with said tool

### DIFF
--- a/portafly/portafly_api.yml
+++ b/portafly/portafly_api.yml
@@ -8,3 +8,95 @@ info:
   license:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0
+paths:
+  /developer/accounts:
+    summary: A list of developer accounts
+    description: A list of developer accounts.
+    get:
+      responses:
+        "200":
+          $ref: '#/components/responses/developerAccountList'
+      operationId: developerAccountList
+      summary: Get developer account list
+    parameters:
+    - name: page
+      description: Page in the paginated list. Defaults to 1.
+      schema:
+        type: integer
+      in: query
+      required: false
+    - name: per_page
+      description: Number of results per page. Default and max is 500.
+      schema:
+        type: integer
+      in: query
+      required: false
+    - name: state
+      description: Account state.
+      schema:
+        type: string
+      in: query
+components:
+  schemas:
+    DeveloperAccount:
+      title: Root Type for developerAccountData
+      description: A simplified format for a developer account data.
+      type: object
+      properties:
+        id:
+          description: The account ID number
+          type: integer
+        created_at:
+          format: date-time
+          description: When it was created
+          type: string
+        updated_at:
+          format: date-time
+          description: When it was last updated
+          type: string
+        state:
+          description: The current state of the account
+          type: string
+        org_name:
+          description: The organization name
+          type: string
+        admin_name:
+          description: The name of the admin user
+          type: string
+        apps_count:
+          description: The total amount of applications
+          type: integer
+      example:
+        id: 1
+        created_at: 2019-10-18T05:13:26Z
+        updated_at: 2019-10-18T05:13:27Z
+        state: approved
+        org_name: Umbrella Corp.
+        admin_name: Oswell E. Spencer
+        apps_count: 3
+  responses:
+    developerAccountList:
+      content:
+        application/json:
+          schema:
+            type: array
+            items:
+              $ref: '#/components/schemas/DeveloperAccount'
+          examples:
+            developerAccountListExample:
+              value:
+              - id: 22
+                created_at: 2018-02-10T09:30Z
+                updated_at: 2018-02-10T09:30Z
+                state: approved
+                org_name: Capsule Corp
+                admin_name: Dr. Brief
+                apps_count: 4
+              - id: 72
+                created_at: 2018-02-10T09:30Z
+                updated_at: 2018-02-10T09:30Z
+                state: approved
+                org_name: FOXHOUND
+                admin_name: Big Boss
+                apps_count: 5
+      description: Developer accountlist
--- a/portafly/portafly_api.yml
+++ b/portafly/portafly_api.yml
@@ -1,0 +1,10 @@
+---
+openapi: 3.0.2
+info:
+  title: PortaFly
+  version: 1.0.0
+  description: |
+    A tuned API for PortaFly, a Patternfly SPA consuming 3scale Porta API MGMT solution.
+  license:
+    name: Apache 2.0
+    url: https://www.apache.org/licenses/LICENSE-2.0


portafly/portafly_api.yml
```diff
@@ -0,0 +1,127 @@
...
+      description: A unique identifier for a `Application`.
+      schema:
+        type: string
+      in: path
+      required: true
+components:
+  schemas:
+    Application:
+      title: Root Type for Application
+      description: A Service Application
```
hallelujah: Right now we have contract instead of application.

| Type | Description | Reference |
| ----- | ----------- | ---------- |
| Contract | Meta class for all types of contracts | Plan (Meta class) |
| Cinstance |  The actual application | ApplicationPlan |
| ServiceContract | A type of contract by Service | ServicePlan |
| AccountContract | A type of contract by Account | AccountPlan |

portafly/portafly_api.yml
```diff
@@ -0,0 +1,127 @@
...
+          format: date-time
+          type: string
+        updated_at:
+          format: date-time
+          type: string
+        state:
+          type: string
+        user_account_id:
+          type: string
+        service_id:
```
hallelujah: It is missing the `plan_id`

portafly/portafly_api.yml
```diff
@@ -0,0 +1,127 @@
...
+        "200":
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/Application'
+              examples:
+                Application List Example:
+                  value:
```
josemigallas: We could benefit from referencing our examples:
```suggestion
                  $ref: '#/components/examples/applicatoinList'
```
Then later:
```yaml
...
components:
  examples:
    applicatoinList:
      summary: 'a list of applications'
      value:
        - id: "1"
        ...
        - id: "2"
        ...
```

josemigallas:
Rebased

portafly/portafly_api.yml
```diff
@@ -0,0 +1,102 @@
...
+      in: query
+      required: false
+    - name: state
+      description: Account state.
+      schema:
+        type: string
+      in: query
+components:
+  schemas:
+    developerAccountData:
```
hallelujah: Why suffixing with `Data`?

Can't it be just `DeveloperAccount`

portafly/portafly_api.yml
```diff
@@ -0,0 +1,102 @@
...
+      title: Root Type for developerAccountData
+      description: A simplified format for a developer account data.
+      type: object
+      properties:
+        id:
+          description: The account ID number
+          type: integer
+        created_at:
+          format: date-time
```
hallelujah: Do not forget: We must make sure that it is formatted correctly: https://tools.ietf.org/html/rfc3339#section-5.6
See https://swagger.io/docs/specification/data-models/data-types/

portafly/portafly_api.yml
```diff
@@ -0,0 +1,102 @@
...
+          type: string
+        apps_count:
+          description: The total amount of applications
+          type: integer
+      example:
+        id: 1
+        created_at: 2019-10-18T05:13:26Z
+        updated_at: 2019-10-18T05:13:27Z
+        state: approved
+        org_name: Umbrella
```
hallelujah:
```suggestion
        org_name: My Company
```

portafly/portafly_api.yml
hallelujah: Also I am wondering if Account is better? Or is it inherited?

Account  -> DeveloperAccount
Account  -> ProviderAccount

portafly/portafly_api.yml
```diff
@@ -0,0 +1,102 @@
...
+          description: When it was last updated
+          type: string
+        state:
+          description: The current state of the account
+          type: string
+        org_name:
+          description: The organization name
+          type: string
+        admin_name:
+          description: The name of the admin user
```
guicassolato: Are you sure you'll only need the name of the user? Maybe a link to the user instead?

didierofrivia: Given the design in https://issues.redhat.com/browse/THREESCALE-4725, only the name is needed. No link for the admin user. It might be needed for the organization link that links to the account, which is not yet created. I thought about expanding the spec as we are building new resources. Does that make sense?

portafly/portafly_api.yml
didierofrivia: You mean I should use a string format and we validate it's a valid date format? 
FYI, the `date-time` format was suggested by Apicurio

portafly/portafly_api.yml
didierofrivia: https://residentevil.fandom.com/wiki/Umbrella_Corporation

portafly/portafly_api.yml
didierofrivia: You are right, no need to sufix with `Data` and should start with Capital letter. I'd say let's leave it as `DeveloperAccount` and when we need to define the `Account` data type I change it. Let's do it a bit by bit. What do you think?

portafly/portafly_api.yml
hallelujah: Yes but enterprise environment is boring

portafly/portafly_api.yml
hallelujah: Well just a reminder for examples, but I saw late they were well formatted 👍

portafly/portafly_api.yml
didierofrivia: But we are not! ;)

josemigallas:
```suggestion
        org_name: Umbrella Corp.
```

portafly/portafly_api.yml
guicassolato: I guess it makes sense, yes.

PR #1776, opened 2020-03-13
Testing Airbnb's polyglot

damianpm:

We need to choose a localization/internazionalization framework for portafly

This PR tests Airbnb's [Polyglot](https://airbnb.io/polyglot.js/)

**Pros:**
Simple, easy to use
Flexible: You can create as many instances as you want and pass a language and set of strings to each one
**Cons:**
It's very basic, we need to take care of things like currency formatting

Note: A more full-featured complex framework is [i18next](https://www.i18next.com/), but it may be over-killing for our needs

**Guidelines:**

1. Are we supporting any other language apart from English?
Not now
2. Do we have plans to add other language(s) in the future?
Not in the short term.
However, we need to prepare an architecture that supports localization. For example, Fuse (the product) is in English and Japanese; while 3scale Docs are in English and also Japanese, we can't dismiss an scenario where 3scale is localized to Japanese in the future.
3. Right now in porta (old UI) we have some strings hardcoded and some others coming from an en.yml file, the new strings are going to be a static file (.yml, .json) ?
We shouldn't have any hardcoded strings, they should come from a single trusted source, it could be a .js, .json or .yml file
4. This file, should live inside porta repo?
Yes, it should be located in portafly subdirectory https://github.com/3scale/porta/tree/master/portafly
5. Who is going to maintain this file ?
    Docs team will create the UI Strings file for the first time, during PatternFly4 adoption.
    Subsequently, we (3scale) will update it.
    Updates will be made via PR's, adding Docs team as reviewers (Docs only will review the texts, not code). We should follow this [3scale Microcopy Guide](https://docs.google.com/document/d/1LBkuJdItYpdwa3-ewLp7gqGsj0VJxDWhp4w1ijKBqdc/edit#)

https://issues.redhat.com/browse/THREESCALE-4685 (parent)
https://issues.redhat.com/browse/THREESCALE-4687 (format of strings files)
https://issues.redhat.com/browse/THREESCALE-4688 (choosing library)

**Verification steps** 

```
cd portafly/
```

```
yarn
```

```
SKIP_PREFLIGHT_CHECK=true PORT=3003 yarn start
```

Small demo:
demo2


portafly/package.json
```diff
@@ -5,6 +5,9 @@
   "dependencies": {
     "@patternfly/react-core": "^3.124.1",
     "@patternfly/react-table": "^2.25.6",
+    "@types/node-polyglot": "^0.4.34",
+    "moment": "^2.24.0",
```
damianpm: this is here only to test some dates, later I'll remove it

portafly/src/i18n/locales/en/strings.js
```diff
@@ -0,0 +1,48 @@
+export const OVERVIEW = {
```
damianpm: We are trying to reach an agreement on the format of the strings file.
At the moment, this is what is been discussed with Docs team.
It may change as the discussion continues

portafly/src/i18n/locales/en/strings.js
```diff
@@ -0,0 +1,48 @@
...
+  page_title: 'Overview | Portafly',
+  page_title_description: 'Title of Overview page for tab browser',
```
damianpm: Every strings needs to have a description, to avoid false friends in translations

portafly/src/pages/Overview-back.tsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react'
```
damianpm: This file is the backup of Overview component, it will be restored once you have checked the demo

portafly/src/pages/Overview.tsx
```diff
@@ -1,32 +1,90 @@
-import React from 'react'
```
damianpm: This file is just a Demo, so you can have an idea of how things may look in the future when new languages are added.

portafly/src/pages/Overview.tsx
```diff
@@ -1,32 +1,90 @@
...
+import { 
+  OVERVIEW as STRINGS_JAP,
+  COMMONS as COMMONS_JAP
+} from 'i18n/locales/jp/strings'
+
 const Overview: React.FunctionComponent = ({children}) => {
+  const [strings, setStrings] = useState(STRINGS_ENG)
+  const [commonStrings, setCommonStrings] = useState(COMMONS_ENG)
+  const [locale, setLocale] = useState('en')
+  const polyglot = useContext(I18nContext)
```
josemigallas: Question: is it necessary to wrap everything in a Context?

portafly/src/pages/Overview.tsx
```diff
@@ -1,32 +1,90 @@
...
+
+  const onClick = (event: React.SyntheticEvent<HTMLButtonElement>) => {
+    setStrings(event.currentTarget.value === 'en' ? STRINGS_ENG : STRINGS_JAP)
+    setCommonStrings(event.currentTarget.value === 'en' ? COMMONS_ENG : COMMONS_JAP)
+    setLocale(event.currentTarget.value)
+  }
+
   useA11yRouteChange()
-  useDocumentTitle('Overview')
+  useDocumentTitle(polyglot.t('page_title'))
```
josemigallas: We could import the `t` only and make it less verbose since it's going to be.. everywhere
```suggestion
  useDocumentTitle(t('page_title'))
```

portafly/src/pages/Overview.tsx
```diff
@@ -1,32 +1,90 @@
...
         <TextContent>
-          <Title size={'3xl'}>Overview</Title>
+          <p>
+            <button value='en' onClick={onClick}>English</button>
+            <button value='jp' onClick={onClick}>日本語</button>
+          </p>
+          <Title size={'3xl'}>{polyglot.t('body_title')}</Title>
           <Text>
-            <b>PortaFly</b> the next gen UI for Porta API MGMT App
+          <b>{polyglot.t('subtitle')}</b>
```
josemigallas: How do we cope with partially-bold phrases like this one?

josemigallas:
I guess not but would it be possible to use the actual phrases as id?
```jsx
<h1>t('Portafly, the best page ever')</h1>
```
instead of
```jsx
<h1>t('title')</h1>
```
I'm thinking.. it's going to be awful when we have to deal with dozens of strings in the same file, and none of them make any sense..

portafly/src/pages/Overview.tsx
damianpm: Not necessary. You can import polyglot directly and create an instance here. But the idea is to use context instead of creating a lot of instances in several files

portafly/src/pages/Overview.tsx
damianpm: In lines 30 and 31 we are passing the `locale` (language) and `phrases`. If we only import the `t` method we can't do that.
We are not adding a second language shortly, so `locale`is not a big problem. But being able to pass the `phrases` is important.

portafly/src/pages/Overview.tsx
damianpm: Polyglot doesn't provide any HTML format support (AFAIK). So, we need to break it into several fields:
`<b>{polyglot.t('subtitle.bold_word')}</b>{polyglot.t('subtitle.normal_word')}`

damianpm:
> I guess not but would it be possible to use the actual phrases as id?
> 
> ```js-jsx
> <h1>t('Portafly, the best page ever')</h1>
> ```
> 
> instead of
> 
> ```js-jsx
> <h1>t('title')</h1>
> ```
> 
> I'm thinking.. it's going to be awful when we have to deal with dozens of strings in the same file, and none of them make any sense..

Yes we can, but it doesn't make sense.
You can pass any string to `t`: `polyglot.t('My fancy title')`
If `'My fancy title'` is not found in the phrases object, it will simply use the string value, polyglot doesn't translate anything, it only helps you to localize properly (e.g plurals) in the  corresponding locale. If we pass the actual text we are hardcoding it and later, when adding a new language or updating a string reused in several places, we'll need to update it by hand.

portafly/package.json
didierofrivia: better do the real thing with `date-fns`, that way you'll proving it works or any other findings

portafly/src/pages/Overview.tsx
didierofrivia: maybe we could import them separately, spreading the import, no?

didierofrivia:
I think we should also POC the currency, since we have a billing module. Also with date-fns locals. Even if the layout is using it, we might decide either to make it work with it or replace it with moment.js

portafly/src/pages/Overview.tsx
josemigallas: That's inconvenient :/

portafly/src/i18n/I18nProvider.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react'
+import Polyglot from 'node-polyglot'
+
+const polyglot = new Polyglot()
+const I18nContext = React.createContext(polyglot)
```
josemigallas: I would call it differently in order to detach from fixed dependencies. Something like `src/localization/LocalizationContext` or something.

portafly/src/pages/Overview.tsx
```diff
@@ -1,32 +1,90 @@
...
+  OVERVIEW as STRINGS_JAP,
+  COMMONS as COMMONS_JAP
+} from 'i18n/locales/jp/strings'
+
 const Overview: React.FunctionComponent = ({children}) => {
+  const [strings, setStrings] = useState(STRINGS_ENG)
+  const [commonStrings, setCommonStrings] = useState(COMMONS_ENG)
+  const [locale, setLocale] = useState('en')
+  const polyglot = useContext(I18nContext)
+  polyglot.locale(locale)
```
josemigallas: This should be called from a setup file I think (the context component's file)

portafly/src/pages/Overview.tsx
```diff
@@ -1,32 +1,90 @@
...
+  COMMONS as COMMONS_JAP
+} from 'i18n/locales/jp/strings'
+
 const Overview: React.FunctionComponent = ({children}) => {
+  const [strings, setStrings] = useState(STRINGS_ENG)
+  const [commonStrings, setCommonStrings] = useState(COMMONS_ENG)
+  const [locale, setLocale] = useState('en')
+  const polyglot = useContext(I18nContext)
+  polyglot.locale(locale)  
+  polyglot.extend({ ...strings, ...commonStrings })
```
josemigallas: And also this

portafly/src/pages/Overview.tsx
josemigallas: You could even import both, I don't se the problem. And initialization code such as `locale` could be move to an initialization file such as the Context (repeating myself https://github.com/3scale/porta/pull/1776/files#r393571503)

portafly/package.json
```diff
@@ -6,6 +6,7 @@
     "@patternfly/react-core": "^3.124.1",
     "@patternfly/react-table": "^2.25.6",
     "@types/node-polyglot": "^0.4.34",
+    "currency.js": "^1.2.2",
```
damianpm: This lib is only to test some currencies, we don't need to keep it

portafly/src/i18n/I18nProvider.tsx
damianpm: I don't get it, what is the problem with i18n ?
`i18n` refers to ` internationalization`, not   to a specific dependency as `i18next`

cntlsn:
> I think we should also POC the currency, since we have a billing module. Also with date-fns locals. Even if the layout is using it, we might decide either to make it work with it or replace it with moment.js

When working with dates, please remember about checking the PatternFly guidelines https://www.patternfly.org/v4/design-guidelines/content/grammar-and-terminology#date-and-time

damianpm:
> When working with dates, please remember about checking the PatternFly guidelines https://www.patternfly.org/v4/design-guidelines/content/grammar-and-terminology#date-and-time

@cntlsn I've updated the format of dates, but don't worry this is just a test

portafly/src/pages/Overview.tsx
damianpm: I moved it to a setup file

portafly/src/i18n/I18nProvider.tsx
josemigallas: You're right, forget about it!

portafly/package.json
```diff
@@ -5,6 +5,9 @@
   "dependencies": {
     "@patternfly/react-core": "^3.124.1",
     "@patternfly/react-table": "^2.25.6",
+    "@types/node-polyglot": "^0.4.34",
```
josemigallas: This goes under dev-dependencies

portafly/src/App.tsx
```diff
@@ -6,6 +6,8 @@ import { AppLayout, SwitchWith404, LazyRoute } from 'components'
 import { LastLocationProvider } from 'react-router-last-location'
 import logo from 'assets/logo.svg'
 
+import { I18nProvider } from 'i18n/I18nProvider'
```
josemigallas:
```suggestion
import { I18nProvider } from 'i18n'
```

portafly/src/i18n/I18nProvider.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+interface Props {
+  children: React.ReactNode
+}
+
+const I18nProvider = (props: Props)=> {
```
josemigallas: No need to declare custom props:
```suggestion
const I18nProvider: React.FunctionComponent = (props)=> {
```

josemigallas: In fact we should discuss if we're using `Props` like we do in porta OR `React.FunctionComponent` like @didierofrivia has been doing.

portafly/src/i18n/locales/en/strings.js
josemigallas: If using a single file, the folder structure could be:
```
i18n
  locales
    en.json
    jp.json

File: portafly/src/i18n/I18nProvider.tsx (hunk 15 above)
Comment (didierofrivia) [reply portafly/src/i18n/I18nProvider.tsx]: I like to define  `Props` when there's more than one property used

File: portafly/src/i18n/setupTranslate.tsx [hunk 16]
```diff
@@ -0,0 +1,8 @@
+import Polyglot from 'node-polyglot'
+
+export const setupTranslate = (polyglot:Polyglot , locale: string, phrases: object) => {
+  const t = polyglot.t.bind(polyglot)
+  polyglot.locale(locale)  
+  polyglot.extend({ ...phrases })
```
josemigallas: I think this is unnecessary:
```suggestion
  polyglot.extend(phrases)
```

portafly/src/i18n/setupTranslate.tsx
```diff
@@ -0,0 +1,8 @@
+import Polyglot from 'node-polyglot'
+
+export const setupTranslate = (polyglot:Polyglot , locale: string, phrases: object) => {
```
josemigallas: Maybe we should define a type for this? I was thinking something like
```
type Phrases = {
  [key: string]: string
}
```
```suggestion
export const setupTranslate = (polyglot:Polyglot , locale: string, phrases: Phrases) => {
```

portafly/package.json
damianpm: true

portafly/src/i18n/setupTranslate.tsx
damianpm: In that case it should be 
```
type Phrases = {
  [key: string]: string | object
}
```

damianpm:
Closed in favor of https://github.com/3scale/porta/pull/1784

PR #1777, opened 2020-03-13
Fix APIAP routing policy rule

guicassolato:
The regex used to match the path of the backend in the APIAP routing policy configuration should only allow paths that _start with_ the chosen value for the path. Currently it matches all paths that _contains_ the value.

This PR fixes it by wrapping the current regex pattern within a `^()` group.

Closes [THREESCALE-4736](https://issues.redhat.com/browse/THREESCALE-4736)

### DIFF
--- a/app/models/config_path.rb
+++ b/app/models/config_path.rb
@@ -10,9 +10,9 @@ class ConfigPath
   end
 
   def to_regex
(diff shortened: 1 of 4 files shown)


eloycoto:
@jsmadis will be good to add this case to the framework ;-)

PR #1778, opened 2020-03-13
Refactoring extended system_names of backend api metrics

guicassolato:
- [x] System name of APIAP Backend metrics no longer hiding the backend ID suffix
    Screenshot 2020-03-13 at 17 47 21
- [x] Prevent metric `system_name` from being updated via UI – now only "hits" and API taken care already
    Screenshot 2020-03-18 at 19 42 09
- [ ] Add an explanation about the `.<backend-id>` suffix to the hint message of the field in the UI (?)
- [x] Remove old API to update the metric that allowed a `name` parameter (deprecated since Feb 2014)
- [x] Show metric's friendly name instead of the `system_name` in the form and index page of the mapping rules
    Screenshot 2020-03-18 at 19 44 01
    Screenshot 2020-03-18 at 19 44 10

Plus a couple fixes:
- [x] Fix link to the mapping rules at the methods & metrics page (product and backend) – now always pointing to "new" in the product section and always pointing to "index" in the backend section :/
- [x] Fix `N+1` queries in index of backend api metrics

Closes [THREESCALE-4015](https://issues.redhat.com/browse/THREESCALE-4015)


app/controllers/admin/api/metrics_base_controller.rb
```diff
@@ -1,24 +1,13 @@
...
-    def metrics
-      @metrics ||= service.metrics
-    end
-
-    def metric_params
-      parameters = params.fetch(:metric)
-
-      if parameters[:name].present?
-        # as of February 2014, notify that this provider is using a deprecated api
-        deprecated_api "`name' parameter for metric/method api was deprecated, `system_name' should be used for provider #{current_account.id}/#{current_account.org_name}"
```
guicassolato: This is a breaking change. OK to proceed?

cc @3scale/product  @3scale/support @3scale/documentation

app/views/provider/admin/backend_apis/metrics/index.html.slim
```diff
@@ -29,7 +29,7 @@ section.Section.u-toggleableBySettingsBox data-state="open"
 
     tbody
       - if @backend_api.method_metrics.any?
-        - @backend_api.method_metrics.each do |method|
+        - @backend_api.method_metrics.includes(:proxy_rules, :parent).each do |method|
```
Martouta: PRobably safer (although maybe there is no need 😄 :

```suggestion
        - @backend_api.method_metrics.includes(:proxy_rules, :parent).find_each do |method|
```

app/views/provider/admin/backend_apis/metrics/index.html.slim
```diff
@@ -79,7 +79,7 @@ section.Section.u-toggleableBySettingsBox data-state="open"
 
     tbody
       - if @backend_api.top_level_metrics.any?
-        - @backend_api.top_level_metrics.each do |metric|
+        - @backend_api.top_level_metrics.includes(:proxy_rules).each do |metric|
```
Martouta: Like https://github.com/3scale/porta/pull/1778/files#r394596230

```suggestion
        - @backend_api.top_level_metrics.includes(:proxy_rules).find_each do |metric|
```

app/views/provider/admin/backend_apis/metrics/index.html.slim
guicassolato: I prefer not to touch it. It's a READ operation for the UI only. If it's too heavy for the database, then so it is for the UI.

app/views/provider/admin/backend_apis/metrics/index.html.slim
guicassolato: https://github.com/3scale/porta/pull/1778#discussion_r394597635

app/views/provider/admin/backend_apis/metrics/index.html.slim
Martouta: I am not as worried for the DB as for the memory it would need at once 😄 
However, this is a very good point: Then it is also too heavy for the UI 👍

app/views/shared/proxy_rules/_table.html.slim
```diff
@@ -33,7 +33,7 @@ table#mapping_rules.data
         td.delta
           = rule.delta
         td.metric
-          = rule.metric_system_name
+          = rule.metric.friendly_name
```
Martouta: Is this "rule" always the decorator? if so, then I think we should do there `delegate :system_name, :friendly_name, to: :metric, prefix: true` and here `rule.metric_friendly_name`

So the less that we spread the "talking to neighbors", the better 😄

features/old/api/metrics/update.feature
```diff
@@ -19,10 +19,9 @@ Feature: Metric update
     And I go to the service definition page
     And I follow "Nukes deployed"
     And I fill in "Friendly name" with "Number of atomic bombs dropped"
-    And I fill in "metric[system_name]" with "bombs"
     And I fill in "Unit" with "drops"
     And I press "Update Metric"
     Then I should see "Number of atomic bombs dropped"
-    And metric "bombs" should have the following:
+    And metric "nukes" should have the following:
       | Friendly name | Number of atomic bombs dropped |
```
Martouta: Does the table stay say "Number of atomic bombs dropped" after this changes? How come? 😱

test/unit/backend_api_logic/metric_extension_test.rb
```diff
@@ -11,17 +11,12 @@ class MetricExtensionTest < ActiveSupport::TestCase
   attr_reader :backend_api, :metric
 
   test 'extends metric system_name with backend api id' do
-    extended_system_name = metric.extended_system_name
+    extended_system_name = metric.send(:extended_system_name)
```
Martouta: I do not think that the response of `extended_system_name` matters enough to be tested on its own anymore considering that now it is a protected method. What truly matters now is what it is saved in `attributes['system_name']` and what `system_name` returns imo 🤔 

So I'd go for testing like this:
```ruby
test 'extends metric system_name with backend api id' do
  assert metric.save
  assert_equal "whatever.#{backend_api.id}", metric.reload.attributes['system_name']
  assert_equal "whatever.#{backend_api.id}", metric.system_name
end
```

app/views/shared/proxy_rules/_table.html.slim
guicassolato: I think it may not be always decorated. What you suggested was the first thing I tried and it did not work. I think `metric_system_name` was actually from https://github.com/3scale/porta/blob/179cd1fd95ff5ecee03618311053f2f81fb698cf/app/models/proxy_rule.rb#L105.

features/old/api/metrics/update.feature
guicassolato: The friendly name [was](https://github.com/3scale/porta/blob/179cd1fd95ff5ecee03618311053f2f81fb698cf/features/old/api/metrics/update.feature#L8) "Nukes deployed" and it was updated to "Number of atomic bombs dropped". The test is correct IMO.

Only thing I changed is no longer to update the `system_name`. That's why instead of searching for a metric whose `system_name` is "bombs", it now searches for "nukes", i.e. the original and unchanged `system_name`.

Martouta: aaaah riiiight sorry 🙈

app/controllers/admin/api/metrics_base_controller.rb
porueesq: @guicassolato Do we need to document this? If so, we need a Doc JIRA.

guicassolato: @porueesq, it's just a heads-up now. I don't know if it needs documentation. Probably the other way around I think, i.e. if we currently have something documented about this (I don't think so), then we need to remove it. Otherwise, it's just "release notes".

PR #1780, opened 2020-03-17
Add validation to trial period, setup fee and cost per month

srijitamukherjee:

This PR adds a validation for parameters which cannot be negative. 

[THREESCALE-1758: Negative value can be set for trial period](https://issues.redhat.com/browse/THREESCALE-1758)

While creating a new application, User(previously) could give negative inputs in the fields like: trial period, setup fee and cost per month. Now these fields ask for a positive value.

**Verification steps** 

- Go to the 'Context Selector'
- Click on the 'API'
- On the left hand side, under the functionality panel, click on 'Applications' and choose 'Application Plans'
- Click on 'Create Application Plan' and these parameters are available on the new page that shows up.

### DIFF
--- a/app/models/plan.rb
+++ b/app/models/plan.rb
@@ -61,8 +61,8 @@ class Plan < ApplicationRecord
 
   validates :name, presence: true
 
-  validates :setup_fee, :cost_per_month, numericality: { allow_nil: false, allow_blank: false }
-
+  validates :setup_fee, :cost_per_month, numericality: { allow_nil: false, allow_blank: false, greater_than_or_equal_to: 0.00 }
+  validates :trial_period_days, numericality: { allow_nil: true, greater_than_or_equal_to: 0 }
   validates :system_name, length: { maximum: 255 }
   validates :name, :rights, :state, :cost_aggregation_rule, :type, :issuer_type,
             length: { maximum: 255 }
--- a/test/unit/plan_test.rb
+++ b/test/unit/plan_test.rb
@@ -589,6 +589,33 @@ class PlanTest < ActiveSupport::TestCase
     end
   end
 
+  test 'setup_fee cannot be negative' do
+    plan = FactoryBot.build_stubbed(:application_plan)
+    plan.setup_fee = -10.00
+    refute plan.valid?
+    assert_equal ['must be greater than or equal to 0.0'], plan.errors[:setup_fee]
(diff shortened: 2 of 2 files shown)


app/models/application_plan.rb
```diff
@@ -17,7 +17,10 @@ class ApplicationPlan < Plan
       where(services: { account_id: provider.id }).joins(:service).references(:service).readonly(false)
     end
   }
-
+  validates :trial_period_days, numericality: { greater_than_or_equal_to: 0 }
+  validates :setup_fee, numericality: { greater_than_or_equal_to: 0.00 }
+  validates :cost_per_month, numericality: { greater_than_or_equal_to: 0.00 }
```
guicassolato: Notice that we already have `numericality` validation defined for `cost_per_month` at https://github.com/3scale/porta/blob/179cd1fd95ff5ecee03618311053f2f81fb698cf/app/models/plan.rb#L64.

Perhaps all these should move to `Plan`?

Also, make sure to include `allow_nil: true` option.

app/models/application_plan.rb
```diff
@@ -17,7 +17,6 @@ class ApplicationPlan < Plan
       where(services: { account_id: provider.id }).joins(:service).references(:service).readonly(false)
     end
   }
-
```
Martouta: ??? 👀

srijitamukherjee: Since all the validations are in plan.rb, Gui suggested to add in the same file and not in application_plan.rb
Hence the change. Thoughts?

Martouta: I agree with Gui, what does not imply removing the empty space between these 2 methods/scopes :stuck_out_tongue_closed_eyes:

Martouta:
What @josemigallas and I are saying is, just do this:

```
git checkout HEAD~2
git add app/models/plan.rb
git add test/unit/plan_test.rb
git commit -m "Add validation to trial period, setup fee and cost per month"
git checkout .
git fetch && git rebase origin/master && git push -f
```

app/models/application_plan.rb
srijitamukherjee: Done

app/models/application_plan.rb
srijitamukherjee: Done

PR #1783, opened 2020-03-19
POC for localization using Polyglot

josemigallas:
This is an alternative POC for localization. It's based on @damianpm's works (#1776).

In this version I've abstracted our components from the `i18n` library and created a hook, `useLocalization` to cope with all the inner logics. Using a `I18nContext` we can keep track of the current locale of the app and by using the hook inside a page we get all the necessary strings:
```jsx
// upper level
<I18nProvider> // default locale is enUS, could be easily customized <I18nProvider locale={...}>
  <App>
    <Component />
  </App>
</I18nProvider>
```
```jsx
// Get a "translator" function from wherever needed
const Component = () => {
  const { t } = useLocalization('component')
  return (
    <h1>{t('title')}</h1>
    <p>{t('description')}</p>
  )
}
```
```jsx
// Change the locale using the hook and everything using it will be updated
const ChangeLocalButton = ({ locale: Locale }) => {
  const { t, setLocale } = useLocalization('change_language_button')
  return (
    <button onClick={() => setLocale(Locale.IT) }>{t('change_language_it)}</button>
    <button onClick={() => setLocale(Locale.ES) }>{t('change_language_es)}</button>
  )
}
```

**IMPORTANT NOTE**: I've decided to take out currency and date internationalization since I don't think it belongs to this PR: neither of them should change depending on the language displayed. Eventually they can be added to the same hook and retrieved alongside `t` but I implemented this and it did not make sense:
```js
const { t, c, date } = useLocalization('overview')
```


portafly/src/i18n/translations/en-us.json
```diff
@@ -0,0 +1,51 @@
+{
```
didierofrivia: This would be much more readable and maintainable if we use YAML.

portafly/src/i18n/useLocalization.tsx
```diff
@@ -0,0 +1,37 @@
...
+ */
+const useLocalization = (page?: Page) => {
+  const { locale, setLocale } = useContext(I18nContext)
+
+  const translations = TRANSLATIONS[locale]
+  let phrases = translations.common
+
+  if (page) {
+    phrases = { ...phrases, ...translations[page] }
+  }
```
didierofrivia:
```suggestion
const phrases = page ? { ...translations.common, ...translations[page] } : translations.common
```

portafly/src/i18n/useLocalization.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+  const polyglot = new Polyglot({
+    locale,
+    phrases
+  })
+
+  return {
+    t: (key: string, options?: any) => polyglot.t(key, options),
```
didierofrivia: Type for options maybe?

portafly/src/App.tsx
```diff
@@ -56,25 +18,16 @@ const PagesSwitch = () => (
...
-      navGroupsStyle="expandable"
-    >
-      <PagesSwitch />
-    </AppLayout>
-  )
-}
+const App = () => (
+  <Router>
+    <LastLocationProvider>
+      <I18nProvider>
```
didierofrivia: In the future we can store the I18N settings either in cookies or bring them from ddbb and pass the locale here, no?

didierofrivia:
Looks good! And makes sense at the moment not to mix it with currencies and TZ. We can either extend or create a separated solution next.

portafly/src/App.tsx
josemigallas: Correct! (it's pointed out in the PR's description, in the snippet)

portafly/src/i18n/translations/en-us.json
josemigallas: True, I chose the JSON because it can be imported by JS "natively" but there must be a loader or something for YAML

didierofrivia: Would be so much better with yml! let's see if we can find a loader/plugin

josemigallas: Turns out is not so simple... since we're using CRA, we can't use a webpack-loader or something. We would need to add more libraries (`js-yaml`+`fs`) to read and parse the yaml.. or eject.. or jail-break it using something like `react-app-rewired`..

Another solution could be convert it from YAML to JSON to be consumed by the app, not ideal either.

josemigallas:
We're choosing react-i18next

PR #1785, opened 2020-03-20
Adds mock server for proposed OAS spec

josemigallas:

* Using [Prism](https://github.com/stoplightio/prism) for the mock server. It reads our YAML and can be run for development.

* Implementing a DAL (data access layer) with React Hooks to handle all network requests.

* Implementing a dummy Accounts page.

* And also, tests.

[THREESCALE-4627: Add a mock server to make fake API calls](https://issues.redhat.com/browse/THREESCALE-4627)

**Verification steps**
1. `yarn && yarn mock-server` in a terminal.
2. In another terminal `yarn start` and navigate to "Accounts" page in the browser.


portafly/package.json
```diff
@@ -41,11 +42,13 @@
   "scripts": {
     "start": "react-scripts start",
     "build": "react-scripts build",
+    "mock-server": "prism mock portafly_api.yml",
```
hallelujah: 🚀

portafly/src/i18n/locales/en/accounts.json
```diff
@@ -0,0 +1,12 @@
+{
+  "page_title": "Accounts | Portafly",
```
damianpm, line 2: We need a description for every field
```
page_title_description
body_title_description
```
etc

portafly/src/components/accounts/AccountsTable.tsx
```diff
@@ -0,0 +1,42 @@
...
+  const rows: string[][] = accounts.map((a) => [
+    a.org_name,
+    a.admin_name,
+    a.created_at,
+    a.apps_count.toString(),
+    a.state
+  ])
+
+  return (
+    <Table aria-label="Accounts list" cells={COLUMNS} rows={rows}>
```
damianpm, line 35: "Accounts list" should be in the strings file

portafly/src/dal/accounts/getDeveloperAccounts.ts
```diff
@@ -0,0 +1,17 @@
...
+import { IAccounts } from 'types'
+
+/**
+ * Get all Applications asynchronously
+ * @returns An object containing: { accounts, error, isPending }
+ */
+function useGetDeveloperAccounts() {
+  const { data, error, isPending } = useFetch<[IAccounts]>(
+    '/developer/accounts',
+    { headers: { Accept: 'application/json' } }
```
damianpm, line 11: does it accept .yml?

portafly/src/pages/Accounts.tsx
```diff
@@ -0,0 +1,38 @@
...
+    <>
+      <PageSection variant={PageSectionVariants.light}>
+        <TextContent>
+          <Text component="h1">{t('body_title')}</Text>
+        </TextContent>
+      </PageSection>
+
+      <PageSection>
+        {isPending && <Loading />}
+        {error && <Alert variant="danger" title={error.message} />}
```
damianpm, line 29: Should "danger" be included in the strings file , maybe in shared?

portafly/src/i18n/locales/en/accounts.json
josemigallas: I wouldn't worry too much about it right now, it's just a dummy example of Accounts page.

portafly/src/dal/accounts/getDeveloperAccounts.ts
josemigallas: I don't think so, the API is supposed to be JSON, or is it?

portafly/src/pages/Accounts.tsx
josemigallas: It is a fixed input controlled by `Alert`:
```jsx
variant?: 'success' | 'danger' | 'warning' | 'info' | 'default';
```
Not really a string susceptible to the locale, so IMO it's supposed to be "hardcoded".

portafly/src/dal/accounts/getDeveloperAccounts.ts
damianpm: I was thinking in our strings file, not the api, you are right

portafly/src/tests/examples/accounts.json
```diff
@@ -0,0 +1,20 @@
+[
```
didierofrivia, line 1: These kind of objects are gonna be built with factories, covered in https://github.com/3scale/porta/pull/1806

portafly/src/tests/pages/Accounts.test.tsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react'
+
+import { render } from 'tests/custom-render'
+import { Accounts } from 'pages'
+import { useGetDeveloperAccounts } from 'dal/accounts'
+import accounts from 'tests/examples/accounts.json'
```
didierofrivia, line 6: So instead of importing that example you'd build the necessary data like https://github.com/3scale/porta/pull/1806/files#diff-9f4f2ee10599b0500704e7b16b78965bR5-R9

josemigallas: There would still be repetition, right? For this particular case we could use the factory in `tests/examples/accounts` and import them in `Accounts.test` and `AccountsTable.test`.
WDYT?

josemigallas: See my [latest commit](https://github.com/3scale/porta/pull/1785/commits/3096438f638b963c297f981443d7d8609266084a).

portafly/src/tests/examples/accounts.json
josemigallas: Moving discussion to https://github.com/3scale/porta/pull/1785#discussion_r402928169

portafly/src/tests/pages/Accounts.test.tsx
didierofrivia: The idea would be remove the "tests/examples/*.json" favoring the factories

josemigallas: Exactly. What I mean is to call the factories in `tests/examples/*` so that we don't repeat ourselves.

portafly/src/dal/accounts/getDeveloperAccounts.ts
```diff
@@ -0,0 +1,17 @@
+import { useFetch } from 'react-async'
+import { IDeveloperAccount } from 'types'
+
+/**
+ * Get all Applications asynchronously
+ * @returns An object containing: { accounts, error, isPending }
```
didierofrivia, line 6: What would you say about normalizing the response of the DAL layer with as follows:

`payload`:  would be of the desired data type (I.e: IDeveloperAccount, IAccountSettings, etc) OR error type.
`error`: Optional that  **may** be set to boolean `true` if the fetch action presented an error
`isPending`: Boolean, self explanatory.

portafly/src/pages/DeveloperAccounts.tsx
```diff
@@ -0,0 +1,38 @@
...
+      <PageSection variant={PageSectionVariants.light}>
+        <TextContent>
+          <Text component="h1">{t('body_title')}</Text>
+        </TextContent>
+      </PageSection>
+
+      <PageSection>
+        {isPending && <Loading />}
+        {error && <Alert variant="danger" title={error.message} />}
+        {accounts && <DeveloperAccountsTable accounts={accounts} />}
```
didierofrivia, line 30: Instead of having https://github.com/3scale/porta/pull/1785/files#diff-8d46996cf806d696795c4eadfc6f2a05R14-R16 I'd rather have here a ternary operator loading `DeveloperAccountsTable` or `SimpleEmptyState`

josemigallas: From my point of view, an empty state always belong to the list/table, since it features other elements such as a toolbar and pagination, and the EmptyState would replace everything. But I'd like to confirm this with @cntlsn first because there is no empty state in the page's mockups and my assumption might be wrong.

portafly/src/dal/accounts/getDeveloperAccounts.ts
josemigallas: I think I'd rather leave it like it is to be honest. I'm no sure we should mix network errors with http error codes.. We would have to check payload every time for an error to set `error = true`...

didierofrivia: Why would you think we would be mixing network errors with http codes? And we wouldn't be checking the payload, just the error boolean. I thing this way the types would be unified and  our implementation consistent. We would only need to have a custom fn encapsulating `fetch` that we will be passing only the endpoint, and it'll be returning `payload`, `error` and `isPending`. We also will be avoiding repetition of passing the same headers, most probably AuthN/AuthZ, etc. Give it a second thought and maybe we could explore the interface together

portafly/src/pages/DeveloperAccounts.tsx
didierofrivia: Ok, makes sense. Would be good then to avoid repetition of the conditional check in all the pages to make it part of the Table and list components if possible

portafly/src/pages/DeveloperAccounts.tsx
```diff
@@ -0,0 +1,38 @@
...
+      </PageSection>
+    </>
+  )
+}
+
+// Default export needed for React.lazy
```
damianpm, line 36: If we are disabling this rule so often, maybe we could just remove it. I know it's out of scope of this PR, just keep it in mind.

josemigallas: The rest of the project should follow the rule so I would not disable it, it's only skipped in `pages/`. We could however add an `eslint` file inside then folder that overrides the rule, to avoid verbosity.

PR #1786, opened 2020-03-21
[WIP - Just testing][Bugfix] captcha :auto renders when object might be spam even if the captcha is not configured

Martouta:
Closes [THREESCALE-4531 - Dev portal - password reset email does not get sent](https://issues.redhat.com/browse/THREESCALE-4531)

Related to [#1466 - Add Captcha to password reset form in Dev Portal](https://github.com/3scale/porta/pull/1466)

The problem is that [we are not rendering the captcha](https://github.com/3scale/porta/blob/179cd1fd95ff5ecee03618311053f2f81fb698cf/app/lib/three_scale/spam_protection/protector.rb#L73-L88) (because rendering the 'reset password form' is a get request) although it is detected as possible spam, but when the form is sent (post request), [we detect it as spam as well and try to verify the captcha](https://github.com/3scale/porta/blob/179cd1fd95ff5ecee03618311053f2f81fb698cf/app/lib/three_scale/spam_protection/integration/controller.rb#L22-L25) (which could not be sent in the request because we didn't render it in the first place).


app/lib/three_scale/spam_protection/protector.rb
```diff
@@ -60,7 +60,7 @@ def captcha_required?
       end
 
       def captcha_needed?
-        captcha_required? || (!http_method.get? && spam?)
+        captcha_required? || spam?
```
Martouta: This is not correct, because it tries to render the captcha even if it is not configured, but it is like this in the code for now because I am only testing 😄

PR #1788, opened 2020-03-23
Ejecting react-scripts

didierofrivia:

We are no longer using `react-scripts` dependency, we've "ejected" and all the dependencies, scripts and config files are included explicitly in our codebase.

fixes https://issues.redhat.com/browse/THREESCALE-4769

**Verification steps** 
Run `yarn start` and `yarn build` and we should still be in business  

In this PR just the following has been done:
1. Update react-scripts
2. run the `eject` script which imports everything from the dependency to our project
3. Run both yarn start and yarn build to make sure everything works.

That been said, don't be surprise about codeclimate going bananas ;) 

**TODO**
* [x] Sort duplicated deps and move dev only dependencies to devDependencies section
* [x] Fix tests
* [x] Linting only our code
* [x] Remove from lint and codeclimate `react-scripts` scripts and config


didierofrivia:
@3scale/system-ux question: Should be fix linting issues on react script imported files? That'd mean in the future when we want to "update" them from their last version, more differences. Another alternative would be exclude these files

portafly/src/react-app-env.d.ts
```diff
@@ -1,2 +1,69 @@
+/* eslint import/no-default-export: 0 */
+// / <reference types="node" />
```
josemigallas: Watch out!, it's been formatted by eslint automatically and it will cause problems with TS. There should be no space: `/// <reference..`
I say we could esling-ignore the whole `/config` folder... if not, disable the rule in this file like it was before: https://github.com/3scale/porta/blob/4581223c9d11afeb04aace285ce8bcc8f7664612/portafly/src/react-app-env.d.ts#L1-L2

portafly/package.json
```diff
@@ -6,16 +6,23 @@
...
     "@patternfly/react-table": "^2.25.6",
     "date-fns": "^2.11.0",
+    "dotenv": "8.2.0",
+    "dotenv-expand": "5.1.0",
+    "fs-extra": "^8.1.0",
+    "history": "^4.10.1",
     "react": "^16.12.0",
+    "react-app-polyfill": "^1.0.6",
     "react-async": "^10.0.0",
+    "react-dev-utils": "^10.2.1",
```
josemigallas: Does "dev-utils" really go under dependencies or was it overlooked?

PR #1789, opened 2020-03-23
[Proxy] read_and_parse_policies_config!

macejmic:

### DIFF
--- a/app/models/proxy.rb
+++ b/app/models/proxy.rb
@@ -608,15 +608,15 @@ class Proxy < ApplicationRecord
     end
   end
 
-  def read_and_parse_policies_config(raise_error: false)
-    attr_policies_config = read_attribute(:policies_config)
-    attr_policies_config.blank? ? [] : Array(JSON.parse(attr_policies_config))
-  rescue JSON::ParserError => error
-    raise_error ? raise(error) : []
+  def read_and_parse_policies_config
+    read_and_parse_policies_config!
+  rescue JSON::ParserError
+    []
   end
 
   def read_and_parse_policies_config!
-    read_and_parse_policies_config(raise_error: true)
+    attr_policies_config = read_attribute(:policies_config)
+    attr_policies_config.blank? ? [] : Array(JSON.parse(attr_policies_config))
   end
 
   def policies_config_structure
--- a/app/models/proxy.rb
+++ b/app/models/proxy.rb
@@ -135,8 +135,7 @@ class Proxy < ApplicationRecord
   end
 
   def policies_config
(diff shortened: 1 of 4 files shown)


test/integration/api/policies_controller_test.rb
```diff
@@ -28,6 +28,13 @@ def setup
     assert_redirected_to edit_admin_service_policies_path(@service)
   end
 
+  test 'invalid config - does not update policies' do
+    invalid_config = 'invalid-config'.to_json
+    put admin_service_policies_path(@service), proxy: { policies_config: invalid_config }
+    # Checking flash won't work anymore in rails 5+
+    assert_equal 'The policies cannot be saved', flash[:error]
```
Martouta:
```suggestion
    assert_selector "#flashWrapper", text: 'The policies cannot be saved'
```

app/models/proxy.rb
```diff
@@ -609,12 +608,26 @@ def policies_configs_are_correct
...
   def policies_config_structure
-    policies_object = PoliciesConfig.new(policies_config)
+    parsed_config = read_and_parse_policies_config!
+    policies_object = PoliciesConfig.new(parsed_config)
     return if policies_object.valid?
     policies_object.errors.each do |attribute, message|
       errors.add(:policies_config, errors.full_message(attribute, message).downcase)
     end
+  rescue JSON::ParserError
+    errors.add(:policies_config, "has invalid format. The Correct format is: [{ \"name\": \"apicast\"... }]")
```
Martouta: we should use i18n 😄

app/models/proxy.rb
```diff
@@ -609,12 +608,26 @@ def policies_configs_are_correct
...
+  def read_and_parse_policies_config(raise_error: false)
+    attr_policies_config = read_attribute(:policies_config)
+    attr_policies_config.blank? ? [] : Array(JSON.parse(attr_policies_config))
+  rescue JSON::ParserError => error
+    raise_error ? raise(error) : []
+  end
+
+  def read_and_parse_policies_config!
+    read_and_parse_policies_config(raise_error: true)
+  end
```
Martouta: Imo the simple fact of having one bang method and one non-bang method should already imply raising the error or not, without that extra param.

So I propose something like this:

```suggestion
  def read_and_parse_policies_config
    read_and_parse_policies_config!
  rescue JSON::ParserError
    []
  end

  def read_and_parse_policies_config!
    attr_policies_config = read_attribute(:policies_config)
    attr_policies_config.blank? ? [] : Array(JSON.parse(attr_policies_config))
  end
```

macejmic: ok, that's super nice!

PR #1790, opened 2020-03-23
[Refactor] Tests for DeveloperPortal::Admin::Account::PasswordsController in the same file

Martouta:
In order to make it easier to work with it by not having it spread across different files with the same class name.


test/integration/developer_portal/passwords_controller_test.rb
```diff
@@ -0,0 +1,84 @@
+# frozen_string_literal: true
```
macejmic: IMO we should stop writing these "smart" tests using inheritance and other object-oriented mechanisms as our juniors can't really understand them and it's hard to maintain them.

Martouta: > other object-oriented mechanisms as our juniors can't really understand them

sorry if this is too rude but I believe that if someone is beginner enough like to not understand basic object-oriented mechanisms such as this kind of inheritance, what we should do is help them learn it instead of not using it for it... using it has a purpose and the benefit of helping them understand it instead of not using it is higher...
In general, following the rule of not using very helpful and widely used stuff just to don't learn it is too harmful IMHO. Because then that would imply also not using Docker, not using FactoryBot, not using i18n, not using Git, etc... and that would complicate things even more.


> it's hard to maintain them

mmmm... in this case i guess that this is fair enough... so i will use methods for the buyer, provider, etc. instead 😄

test/integration/developer_portal/passwords_controller_test.rb
```diff
@@ -0,0 +1,84 @@
...
+
+require 'test_helper'
+
+class DeveloperPortal::PasswordsControllerTest < ActionDispatch::IntegrationTest
```
Martouta: https://github.com/3scale/porta/pull/1790/files#r396498302

image

PR #1791, opened 2020-03-24
[oracle] Migration for all timestamps

hallelujah:
Because of https://github.com/rsim/oracle-enhanced/blob/def13fc382bd4ad75125921d60ce94dd6f392b31/README.md#upgrade-rails-42-or-older-version-to-rails-5

Fixes https://issues.redhat.com/browse/THREESCALE-4767

TODO:

- [x] Make it work with Postgres:

```

PG::DatatypeMismatch: ERROR:  column "started_at" cannot be cast automatically to type timestamp without time zone
HINT:  You might need to specify "USING started_at::timestamp without time zone".
: ALTER TABLE "line_items" ALTER COLUMN "started_at" TYPE timestamp
/Users/hery/Devel/3scale/porta/db/migrate/20200324094424_change_date_to_timestamp.rb:59:in `block in change'
/Users/hery/Devel/3scale/porta/db/migrate/20200324094424_change_date_to_timestamp.rb:3:in `change'
/Users/hery/.rbenv/versions/2.5/bin/bundle:23:in `load'
/Users/hery/.rbenv/versions/2.5/bin/bundle:23:in `<main>'

Caused by:
ActiveRecord::StatementInvalid: PG::DatatypeMismatch: ERROR:  column "started_at" cannot be cast automatically to type timestamp without time zone
HINT:  You might need to specify "USING started_at::timestamp without time zone".
: ALTER TABLE "line_items" ALTER COLUMN "started_at" TYPE timestamp
/Users/hery/Devel/3scale/porta/db/migrate/20200324094424_change_date_to_timestamp.rb:59:in `block in change'
/Users/hery/Devel/3scale/porta/db/migrate/20200324094424_change_date_to_timestamp.rb:3:in `change'
/Users/hery/.rbenv/versions/2.5/bin/bundle:23:in `load'
/Users/hery/.rbenv/versions/2.5/bin/bundle:23:in `<main>'

Caused by:
PG::DatatypeMismatch: ERROR:  column "started_at" cannot be cast automatically to type timestamp without time zone
HINT:  You might need to specify "USING started_at::timestamp without time zone".
/Users/hery/Devel/3scale/porta/db/migrate/20200324094424_change_date_to_timestamp.rb:59:in `block in change'
/Users/hery/Devel/3scale/porta/db/migrate/20200324094424_change_date_to_timestamp.rb:3:in `change'
/Users/hery/.rbenv/versions/2.5/bin/bundle:23:in `load'
/Users/hery/.rbenv/versions/2.5/bin/bundle:23:in `<main>'
Tasks: TOP => db:migrate
```


db/migrate/20200324094424_change_date_to_timestamp.rb
```diff
@@ -0,0 +1,182 @@
...
+      change_column :cms_groups, :created_at, :datetime
+      change_column :cms_groups, :updated_at, :datetime
+      change_column :cms_files, :attachment_updated_at, :datetime
+      change_column :cms_files, :created_at, :datetime
+      change_column :cms_files, :updated_at, :datetime
+      change_column :alerts, :timestamp, :datetime
+      change_column :line_items, :created_at, :datetime
+      change_column :line_items, :updated_at, :datetime
+      change_column :line_items, :finished_at, :datetime
+      change_column :line_items, :started_at, :datetime, System::Database.postgres? ? {using: 'current_date + started_at'} : {}
```
hallelujah: I think this is incorrect ... I will try to check

hallelujah:
```
hery@Tiburon ~/Devel/3scale/porta (oracle/fix-timestamps●)$ rg started_at                                       [2.5]
migration.rb
56:change_column :line_items, :started_at, :datetime

db/postgres_schema.rb
719:    t.datetime "started_at"

db/oracle_schema.rb
756:    t.datetime "started_at",    precision: 6

db/schema.rb
720:    t.datetime "started_at"

app/models/line_item.rb
10:  attr_accessible :name, :description, :cost, :finished_at, :quantity, :started_at

db/migrate/20200324094424_change_date_to_timestamp.rb
59:      change_column :line_items, :started_at, :datetime, System::Database.postgres? ? {using: 'started_at::timestamp without time zone'} : {}

db/migrate/20101026173617_fill_in_new_finance_fields.rb
14:      item.started_at = item.created_at

db/migrate/20101014164141_add_new_finance_fields.rb
6:    add_column :line_items, :started_at, :time
12:    remove_columns :line_items, :started_at
```

It seems that `started_at` is not used at all ....
We might need to remove this column @Martouta

db/migrate/20200324094424_change_date_to_timestamp.rb
hallelujah: So in Postgres `time` does not convey dates ... that is why we add `current_date` to be able to transform it to `timestamp` (datetime)

https://stackoverflow.com/questions/24930570/postgresql-convert-to-timestamp

Looks OK then

hallelujah:
At the end I removed all non `datetime` columns (we do not want to change column types)

- invoices.last_charging_retry (date)
- invoices.due_on (date)
- invoices.period (date)
- line_items.started_at

db/migrate/20200324094424_change_date_to_timestamp.rb
hallelujah: At then end, we should not change them

db/oracle_schema.rb
```diff
@@ -639,13 +639,13 @@
...
     t.datetime "updated_at",                       precision: 6
-    t.datetime "due_on",                           precision: 6
+    t.date     "due_on"
     t.string   "pdf_file_name"
     t.string   "pdf_content_type"
     t.integer  "pdf_file_size",                    precision: 38
     t.datetime "pdf_updated_at",                   precision: 6
-    t.datetime "period",                           precision: 6
-    t.datetime "issued_on",                        precision: 6
+    t.date     "period"
```
hallelujah: Those are really stored as DB `DATE`  in MySQL, Oracle and Postgres ....
Oracle adapter was just buggy, as it could not differentiate `DATE` and `TIMESTAMPS` before 1.7

No need to convert them for oracle as they are already in the right type`DATE`.

db/migrate/20200324130825_change_date_to_timestamp.rb
```diff
@@ -0,0 +1,198 @@
+class ChangeDateToTimestamp < ActiveRecord::Migration
+  def up
+    safety_assured do
+      change_column :access_tokens, :created_at, :datetime
+      change_column :access_tokens, :updated_at, :datetime
+      change_column :accounts, :created_at, :datetime
+      change_column :accounts, :updated_at, :datetime
+      change_column :accounts, :paid_at, :datetime
+      change_column :accounts, :deleted_at, :datetime if column_exists?(:accounts, :deleted_at)
```
hallelujah: Yeah ... somehow Postgres has this column already removed ... 🤦‍♂

Martouta: I know, I am dealing with that in the columns removal PR :joy:

Martouta: Anyway that column is not used at all anymore so it wouldn't be a drama we don't change the column type for this one.

db/migrate/20200324130825_change_date_to_timestamp.rb
```diff
@@ -0,0 +1,198 @@
+class ChangeDateToTimestamp < ActiveRecord::Migration
+  def up
+    safety_assured do
+      change_column :access_tokens, :created_at, :datetime
```
Martouta: At first, I thought you would do something more dynamic, for example like this:
```ruby
class ChangeDateToTimestamp < ActiveRecord::Migration[5.0]
  def up
    safety_assured do
      ApplicationRecord.connection.tables.each do |table_name|
        ApplicationRecord.connection.columns(table_name).each do |column|
          next if column.type != 'date' # or perhaps compare the column.limit here :)
          change_column table_name, column.name, :datetime
        end
      end
    end
  end
end
```

But anyway, I trust your list 👍

hallelujah: I did it :) But I want to verify the list :)

```ruby
require './config/environment'

Rails.application.eager_load!

changes = ActiveRecord::Base.descendants.each_with_object({}) do |model, memo|
  next if model.abstract_class?

  columns = model.columns.select do |c|
    c.type == :datetime
  end

  memo[model.table_name] = columns.map(&:name)
end

changes = changes.sort

File.open('migration.rb', 'w+') do |f|
  changes.each do |table, names|
    names.each do |name|

      f.puts "change_column :#{table}, :#{name}, :datetime"
    end
  end
end
```

hallelujah: You will miss some tables if you just use ApplicationRecord btw ;)

db/migrate/20200324130825_change_date_to_timestamp.rb
hallelujah: Btw we see from this list some tables and columns that are not used anymore, candidates for the big removal of columns.

db/migrate/20200324130825_change_date_to_timestamp.rb
Martouta: Gold! 👏

PR #1792, opened 2020-03-24
[WIP] [Bugfix] THREESCALE-4531 - Dev portal - password reset email does not get sent

Martouta:
Closes [THREESCALE-4531 - Dev portal - password reset email does not get sent](https://issues.redhat.com/browse/THREESCALE-4531)
Related to [ [THREESCALE-1135 - Add Captcha element in the password reset form](https://issues.redhat.com/browse/THREESCALE-1135) | https://github.com/3scale/porta/pull/1466 ]

The captcha loads when is configured in the settings.yml and the tenant settings spam_protection_level is either :captcha or :auto but the object is considered spam?
If the captcha is not configured but the object is still considered spam, then we display an error message informing of that.

### Verification steps
For :auto in the settings spam_protection_level. With recaptcha configured.

The captcha loads
image

#### If I don't fill it
image
image

#### If I fill it
image

image

### DIFF
--- a/lib/developer_portal/app/controllers/developer_portal/admin/account/passwords_controller.rb
+++ b/lib/developer_portal/app/controllers/developer_portal/admin/account/passwords_controller.rb
@@ -10,17 +10,15 @@ class DeveloperPortal::Admin::Account::PasswordsController < ::DeveloperPortal::
   before_action :find_user, :only => [:show, :update]
 
   def create
-    if !spam_check(buyer)
-      flash[:error] = 'Spam protection failed.'
-      redirect_to_password_reset_url
-    elsif (user = @provider.buyer_users.find_by_email(params[:email]))
-      user.generate_lost_password_token!
-      flash[:notice] = 'A password reset link has been emailed to you.'
-      redirect_to login_url
-    else
-      flash[:error] = 'Email not found.'
-      redirect_to_password_reset_url
-    end
+    return password_reset_error('Spam protection failed.') unless spam_check(buyer)
+
+    user = @provider.buyer_users.find_by_email(params[:email])
+    return password_reset_error('Email not found.') unless user
+
+    user.generate_lost_password_token!
+
+    flash[:notice] = 'A password reset link has been emailed to you.'
+    redirect_to login_url
   end
 
   def new; end
@@ -43,7 +41,8 @@ class DeveloperPortal::Admin::Account::PasswordsController < ::DeveloperPortal::
 
   private
 
-  def redirect_to_password_reset_url
+  def password_reset_error(message)
+    flash[:error] = message
     redirect_to new_admin_account_password_url(request_password_reset: true) # keep hash for retrocompatibility
   end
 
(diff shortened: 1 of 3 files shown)


test/integration/developer_portal/passwords_controller_test.rb
```diff
@@ -59,6 +69,37 @@ def test_update_password
...
+
+  test 'create sends the email when captcha passes and finds the email' do
+    provider.settings.update_attributes(spam_protection_level: :auto)
+    ThreeScale::SpamProtection::Protector.any_instance.stubs(spam?: true)
+    Recaptcha::Verify.stubs(skip?: true)
+
+    post developer_portal.admin_account_password_path(email: user.email)
+    assert_equal 'A password reset link has been emailed to you.', flash[:notice]
+    assert_redirected_to developer_portal.login_path
+    assert user.reload.lost_password_token
```
josemigallas: Asserting something is defined could be dangerous (maybe here, it isn't). Would it be a good idea/possible to stub the token and assert it is exactly the one we expect? Something like:
```
post developer_portal.admin_account_password_path(email: user.email, token: 123)
...
assert_equal user.reload.lost_password_token, 123

Comment (Martouta) [reply test/integration/developer_portal/passwords_controller_test.rb]: Unfortunately, I cannot check for any specific value because it is always randomly generated:
https://github.com/3scale/porta/blob/7645462e3696bcdd2b05704573c23ea344bded1c/app/models/user.rb#L234-L252

However, I still can improve the test by ensuring:
1. The lost_password_token does not only contain a value at this point in time, but also it does not contain it before the request
2. The lost_password_token_generated_at has the current date
3. It also sends the email 👀

File: test/integration/developer_portal/passwords_controller_test.rb [hunk 2]
```diff
@@ -59,6 +69,40 @@ def test_update_password
...
+  end
+
+  test 'create sends the email when captcha passes and finds the email' do
+    provider.settings.update_attributes(spam_protection_level: :auto)
+    ThreeScale::SpamProtection::Protector.any_instance.stubs(spam?: true)
+    Recaptcha::Verify.stubs(skip?: true)
+
+    UserMailer.expects(:lost_password).returns(mock('mail', deliver_now: true)).once
+
+    assert_change of: lambda { user.reload.lost_password_token.present? }, from: false, to: true do
```
Martouta: https://github.com/3scale/porta/pull/1792#discussion_r400891703

lib/developer_portal/app/controllers/developer_portal/admin/account/passwords_controller.rb
```diff
@@ -10,15 +10,16 @@ class DeveloperPortal::Admin::Account::PasswordsController < ::DeveloperPortal::
   before_action :find_user, :only => [:show, :update]
 
   def create
-    return redirect_to new_admin_account_password_url unless spam_check(buyer)
-
-    if user = @provider.buyer_users.find_by_email(params[:email])
+    if !spam_check(buyer)
```
guicassolato, line 13:
```suggestion
    return password_reset_error('Spam protection failed.') unless spam_check(buyer)
    user = @provider.buyer_users.find_by_email(params[:email])
    return password_reset_error('Email not found.') unless user

    user.generate_lost_password_token!
    flash[:notice] = 'A password reset link has been emailed to you.'
    redirect_to login_url
  end
  
  def password_reset_error(message)
    flash[:error] = message
    redirect_to new_admin_account_password_url(request_password_reset: true) # keep hash for retrocompatibility
  end
```

Martouta:
Closing but not removing the branch yet :smile: 
@macejmic will carry on these days as he considers appropriate :smile:

PR #1793, opened 2020-03-24
Analytics for APIAP Backend metrics

guicassolato:
Introduces analytics to APIAP backends.

Screenshot 2020-03-25 at 12 46 39

Closes [THREESCALE-3159](https://issues.redhat.com/browse/THREESCALE-3159).

Contains commits of https://github.com/3scale/porta/pull/1778.

----

#### Backend (porta querying apisonator's Redis)
- [x] `Stats::BackendApi` (usage)
- [ ] ~Stats::BackendApi (top applications)~

#### Admin API
Screenshot 2020-03-25 at 15 18 04

- [x] New endpoint: Backend Traffic by Metric
- [x] "Autocomplete" for backend api metrics

#### Admin Portal (UI)
- [x] Backend > Analytics > Usage (index)
- [x] ~[UI] Backend > Analytics > Top Applications (index)~
- [x] Wire UI and API together → make `statsUsage` polymorphic (service or backend)
    https://github.com/3scale/porta/blob/4581223c9d11afeb04aace285ce8bcc8f7664612/app/javascript/src/Stats/provider/stats_usage.jsx#L15 

#### Filter Product Analytics by API Backend metrics
- [x] UI
    Screenshot 2020-03-25 at 12 54 05
- [x] API
    Screenshot 2020-03-25 at 15 19 11

#### Rename “Usage” → “Traffic” (UI and API)
- [x] Product Analytics
- [x] Application Analytics
- [x] Backend Analytics


thomasmaas:
@gui Is there any way we can group those metrics by backend in the select (and making that clear using `optgroup`)?

app/javascript/src/Stats/lib/metrics_list.jsx
```diff
@@ -7,8 +7,9 @@ export class StatsMetrics {
   }
 
   static getSelectedMetrics ({selectedMetricName, list}) {
-    let selectedMetric = list.metrics.find(metric => metric.systemName === selectedMetricName)
-    let methodsOfSelectedMetric = (list.methods || []).filter(method => method.parentId === selectedMetric.id)
+    let listMethods = (list.methods || [])
+    let selectedMetric = (list.metrics.concat(listMethods)).find(metric => metric.systemName === selectedMetricName)
+    let methodsOfSelectedMetric = listMethods.filter(method => method.parentId === selectedMetric.id)
```
didierofrivia: We could use `const` here instead of `let` to define those.

app/javascript/src/Stats/lib/stats.jsx
```diff
@@ -51,7 +55,7 @@ export const Stats = function ({ChartManager, Chart, Sources}) {
       let statsState = new StatsState(store, Object.assign({}, DEFAULT_STATE, selectedState))
 
       metrics.then(list => {
-        let groupedMetrics = settings.hasGroupedMethods ? list.metrics : [list.metrics[0], ...list.methods, ...list.metrics.slice(1)]
+        let groupedMetrics = settings.hasGroupedMethods ? list.metrics : groupMetrics(list)
```
didierofrivia: here `const` too. I know previously it's been defined with a `let` but since we are changing it now we could use the oportunity

app/javascript/src/Stats/lib/metrics_list.jsx
```diff
@@ -7,8 +7,9 @@ export class StatsMetrics {
   }
 
   static getSelectedMetrics ({selectedMetricName, list}) {
-    let selectedMetric = list.metrics.find(metric => metric.systemName === selectedMetricName)
-    let methodsOfSelectedMetric = (list.methods || []).filter(method => method.parentId === selectedMetric.id)
+    let listMethods = (list.methods || [])
```
didierofrivia: no need of parentheses

app/javascript/src/Stats/lib/metric.jsx
```diff
@@ -4,10 +4,15 @@ export class StatsMetric {
   }
 
   _buildMetric (metric) {
+    let systemName = metric.system_name
+    let parentId = metric.parent_id
```
didierofrivia: `const` here too

app/javascript/src/Stats/lib/metric.jsx
```diff
@@ -4,10 +4,15 @@ export class StatsMetric {
...
+
     this.id = metric.id
     this.name = metric.friendly_name
     this.serviceId = metric.service_id
-    this.systemName = metric.system_name
-    this.isMethod = metric.parent_id != null
+    this.systemName = systemName
+    this.parentId = parentId
+    this.isMethod = parentId != null
+    this.isHits = !!systemName.match(/hits(\.\d+)?/)
```
didierofrivia: I've seen this RegEx used in many other files, maybe we could inside this file also export a `const` with it... like `HITS_METRIC` or something

app/javascript/src/Stats/lib/series.jsx
```diff
@@ -80,8 +80,9 @@ export class StatsSeries {
...
 
   _findMetricHits (response) {
-    if (response.metric) {
-      return response.metric.system_name === 'hits'
+    let metric = response.metric
+    if (metric) {
+      return metric.system_name.match(/hits(\.\d+)?/)
     } else {
       return null
     }
```
didierofrivia:
```suggestion
    const metric = response.metric
    return metric ? metric.system_name.match(/hits(\.\d+)?/) : null
```

app/controllers/admin/api/metrics_base_controller.rb
```diff
@@ -1,24 +1,13 @@
 class Admin::Api::MetricsBaseController < Admin::Api::ServiceBaseController
+  include MetricParams
 
   protected
 
-    def metric
-      raise NotImplementedError.new "#{self.class.name} should implemented metric method"
-    end
+  def metric
+    raise NotImplementedError.new "#{self.class.name} should implemented metric method"
```
hallelujah: NotImplementedError is not the correct class. It has nothing to do with system calls.

app/controllers/api/metrics_controller.rb
```diff
@@ -126,15 +128,16 @@ def find_service
...
+    @metric = find_metric_by(params[:id])
   end
 
-  def build_metric
-    @metric = if params[:metric_id]
-                @service.metrics.find(params[:metric_id]).children
-              else
-                @service.metrics
-              end.build(params[:metric] || {})
+  def find_metric_by(id)
```
hallelujah: The `by` suggests some key

```suggestion
  def find_metric(id)
```

app/controllers/stats/data/backend_apis_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  private
+
+  def set_source
+    backend_apis = current_account.backend_apis.accessible
+    @backend_api = backend_apis.find(params[:backend_api_id])
+    @source  = Stats::BackendApi.new(@backend_api)
+
+    authorize!(:show, @backend_api) if current_user
```
hallelujah: Why don't we have `current_user` always?
Is it because of `provider_key`?

app/controllers/stats/data/backend_apis_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+  rescue ActiveRecord::RecordNotFound
+    render_error 'Backend not found', status: :not_found
+  end
+
+  def slice_and_use_defaults(params, *allowed)
+    options = super(params, *allowed)
+
+    return options unless options.key?(:metric_name)
+
+    extended_system_name = Metric.build_extended_system_name(options[:metric_name], owner_id: params[:backend_api_id])
```
hallelujah: What if the `metric_name` sent is the right one (already extended?)
Can't we send it already with the extended name?

app/controllers/stats/usage_controller.rb
```diff
@@ -7,8 +7,7 @@ class Stats::UsageController < Stats::ServiceBaseController
   liquify if: :buyer_domain?
 
   def index
-    @methods = @service.method_metrics
-    @metrics = @service.metrics.top_level
+    @methods, @metrics = @service.all_metrics.partition(&:method_metric?)
```
hallelujah: If there are many metrics and methods it might bloat the memory

app/controllers/stats/usage_controller.rb
```diff
@@ -22,8 +21,7 @@ def index
 
   def top_applications
     activate_menu :serviceadmin, :monitoring, :top_applications
-    @metrics = @service.metrics.top_level
-    @methods = @service.method_metrics
+    @methods, @metrics = @service.all_metrics.partition(&:method_metric?)
```
hallelujah: If there are many metrics and methods it might bloat the memory

app/decorators/proxy_rule_decorator.rb
```diff
@@ -9,9 +9,8 @@ def pattern
     File.join('/', backend_api_path, pattern_value)
   end
 
-  def metric_system_name
-    object.metric.extended_system_name
-  end
+  delegate :metric, to: :object
```
hallelujah: Hmmm if it is a decorator, why isn't it already delegated to the object?

app/lib/api_docs/provider_data.rb
```diff
@@ -61,12 +61,21 @@ def service_ids
...
+      @account.metrics.top_level
+    end
+
+    def backend_api_metrics
+      @account.backend_api_metrics.top_level
+    end
+
     def metrics
-      @metrics ||= @account.metrics.includes(:service).top_level
+      @metrics ||= service_metrics + backend_api_metrics
```
hallelujah: You are transforming them into Array though they are scopes.
Maybe you can use Arel::Nodes::Union? 


```suggestion
      @metrics ||= Metric.from(service_metrics.union(backend_api_metrics).to_sql)
```

app/controllers/stats/data/backend_apis_controller.rb
guicassolato: `Metric.build_extended_system_name` takes  care of it.

app/controllers/stats/data/backend_apis_controller.rb
guicassolato: I think so. That's is basically the same as done in https://github.com/3scale/porta/blob/9fd53ef353a69c314c6572403f0266b65757e35e/app/controllers/stats/data/services_controller.rb#L74

app/controllers/admin/api/metrics_base_controller.rb
guicassolato: This file was touched in a different PR whose commits are included here (https://github.com/3scale/porta/pull/1778). Even so, this specific line only appears in the diff because of indentation fix.

app/controllers/api/metrics_controller.rb
guicassolato: Belongs to https://github.com/3scale/porta/pull/1778

app/controllers/stats/usage_controller.rb
guicassolato: This is a problem we already have now.

app/decorators/proxy_rule_decorator.rb
guicassolato: Belongs to https://github.com/3scale/porta/pull/1778

app/lib/api_docs/provider_data.rb
guicassolato: Too complicated to use `Arel::Nodes::Union` here. Exactly like suggested, it results in

```sql
SELECT `metrics`.* FROM ( SELECT `metrics`.* FROM `metrics` INNER JOIN `services` ON `metrics`.`service_id` = `services`.`id` WHERE `services`.`account_id` = ? AND `metrics`.`parent_id` IS NULL ORDER BY `metrics`.`id` ASC UNION SELECT `metrics`.* FROM `metrics` INNER JOIN `backend_apis` ON `metrics`.`owner_id` = `backend_apis`.`id` WHERE `backend_apis`.`account_id` = ? AND `metrics`.`owner_type` = ? AND `metrics`.`parent_id` IS NULL ORDER BY `metrics`.`id` ASC ) ORDER BY `metrics`.`id` ASC"
```

This fails in MySQL in at least 3 different ways:
1. Does not evaluate the binds – need to call `#to_sql` on each component individually or pass the binds explicitly all over again
2. Misuse of `ORDER BY` – could try with `#reorder`, but it starts to complicate stuff too much
3. Missing alias for the inner table (see https://github.com/rails/arel/issues/428)

DavidRajnoha:
Hi,
can I ask what are the endpoints to get the backend analytics, so QE can write tests against them? 
Thanks in advance

guicassolato:
@DavidRajnoha, you can check the new endpoint for fetching APIAP backend usage/traffic stats in the diff: https://github.com/3scale/porta/pull/1793/files?file-filters%5B%5D=.json#diff-0d89d8d7ae29d093d1edf2b1b5d3aba7

app/controllers/provider/admin/backend_apis/stats/base_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+      format.json { render json: metrics_with_methods }
+    end
+  end
+
+  protected
+
+  # TODO: ensure provider domain
+  def authorize_monitoring
+    authorize! :manage, :monitoring
+  end
```
Martouta, line 29: The permission to `manage` `monitoring` is defined here:
https://github.com/3scale/porta/blob/c67d5d8227a1f64c0ff2c0d91f5ea39a00cb9fb9/config/abilities/master_admin.rb#L43
https://github.com/3scale/porta/blob/c67d5d8227a1f64c0ff2c0d91f5ea39a00cb9fb9/config/abilities/provider_admin.rb#L58
https://github.com/3scale/porta/blob/c67d5d8227a1f64c0ff2c0d91f5ea39a00cb9fb9/config/abilities/provider_member.rb#L45-L46

This means that either master admin or tenant admin or tenant member with `monitoring` can access.

So to do `TODO: ensure provider domain`, there are numerous ways to do it, but one of them could be:
```ruby
def authorize_monitoring
  authorize! :manage, :monitoring
  raise CanCan::AccessDenied if current_account.master?
end
```

app/controllers/provider/admin/backend_apis/stats/base_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  def find_backend_api
+    @backend_api = collection.find(params[:backend_api_id])
+    authorize! :show, @backend_api
```
Martouta, line 33: Wait, if both master and tenant can manage backend apis:
image

Then I do not understand the TODO comment that I spoke about in https://github.com/3scale/porta/pull/1793/files#r402297728

doc/active_docs/Analytics API.json
```diff
@@ -266,12 +266,276 @@
...
+                "values": [
+                  "month",
+                  "day",
+                  "hour"
+                ],
+                "valueType": "LIST"
+              }
+            },
+            {
+              "description": "Time zone for calculations.",
```
Martouta, line 364: I do not think that this belongs to this PR so we should probably run `bundle exec rake doc:swagger:generate:all` before this PR 😂

Martouta: My bad! It DOES belong to this PR! 👍

app/controllers/stats/data/base_controller.rb
```diff
@@ -24,6 +24,8 @@ class Stats::Data::BaseController < ApplicationController
   ##
   ##~ @parameter_service_id = {:name => "service_id", :description => "ID of the service.", :dataType => "int", :required => true, :paramType => "path", :threescale_name => "service_ids"}
   ##
+  ##~ @parameter_backend_api_id_by_id_name = { :name => "backend_api_id", :description => "ID of the Backend.", :dataType => "int", :required => true, :paramType => "path" }
```
Martouta, line 27: wow, I see that there is some repetition between `Stats::Data::BaseController` (this class) and `Admin::Api::BaseController`:
https://github.com/3scale/porta/blob/c67d5d8227a1f64c0ff2c0d91f5ea39a00cb9fb9/app/controllers/admin/api/base_controller.rb#L14-L15
https://github.com/3scale/porta/blob/c67d5d8227a1f64c0ff2c0d91f5ea39a00cb9fb9/app/controllers/admin/api/base_controller.rb#L132

but definitely not your fault 👍

app/lib/api_docs/provider_data.rb
```diff
@@ -61,12 +61,21 @@ def service_ids
       end
     end
 
+    def service_metrics
+      @account.metrics.top_level
```
Martouta, line 65:
```suggestion
      @account.metrics.includes(:owner).top_level
```

Because I see an N+1 coming below for this:
```ruby
METRIC_NAME = ->(metric) { "#{metric.friendly_name} | #{metric.owner.name}" }
	
def metric_names(metrics = self.metrics)
  metrics.map do |metric|
     { :name  => METRIC_NAME.call(metric),
```

app/lib/api_docs/provider_data.rb
```diff
@@ -61,12 +61,21 @@ def service_ids
...
+    end
+
+    def backend_api_metrics
+      @account.backend_api_metrics.top_level
```
Martouta, line 69: Same as https://github.com/3scale/porta/pull/1793/files#r402309363 but in this case it would be:
```suggestion
      @account.backend_api_metrics.includes(:owner).top_level
```

app/lib/api_docs/provider_data.rb
Martouta: Btw, instead of `.metrics.top_level`, we can also do directly `.top_level_metrics` 😄

test/integration/provider/admin/api_docs/account_data_controller_test.rb
```diff
@@ -6,6 +6,10 @@ class Provider::Admin::ApiDocs::AccountDataControllerTest < ActionDispatch::Inte
   def setup
     @provider = FactoryBot.create(:provider_account)
     @service = provider.default_service
+
+    provider.backend_apis.last.delete # FIXME: provider_account has 2 backend apis!
```
Martouta, line 10: image

app/lib/api_docs/provider_data.rb
Martouta: This method is only called by its own class.

image

Those 2 classes respond with the 'json' data directly, having done the loop through this just as it is. So, if we make this method `private`, this 'issue' won't matter 😄

app/controllers/stats/data/backend_apis_controller.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+class Stats::Data::BackendApisController < Stats::Data::BaseController
+  before_action :set_source
+
+  ##~ sapi = source2swagger.namespace("Analytics API")
+  ##
+  ##~ e = sapi.apis.add
+  ##~ e.path = "/stats/backend_apis/{backend_api_id}/usage.{format}"
```
Martouta, line 9: We shouldn't support XML for new API endpoints 👀 only JSON allowed 💃

app/controllers/stats/data/backend_apis_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+  private
+
+  def set_source
+    backend_apis = current_account.backend_apis.accessible
+    @backend_api = backend_apis.find(params[:backend_api_id])
+    @source  = Stats::BackendApi.new(@backend_api)
+
+    authorize!(:show, @backend_api) if current_user
+  rescue ActiveRecord::RecordNotFound
+    render_error 'Backend not found', status: :not_found
```
Martouta, line 42: If this is removed, and I add this test in `test/integration/stats/data/backend_apis_controller_test.rb`:

```ruby
test 'when the backend api does not exist' do
    params = { metric_name: metric.system_name, period: 'day', since: Time.utc(2020, 3, 19).to_date, timezone: ActiveSupport::TimeZone['UTC'].name, skip_change: false }
    get usage_stats_data_backend_apis_path(backend_api_id: backend_api.id + 20, format: :json, access_token: access_token.value), params

    assert_response :not_found
    pp response.body
end
```
It works perfectly well and the response is just like in the other APIs: `"{\"status\":\"Not found\"}"`

I guess this custom error message is for the UI but I do not think that it is necessary... if it does a request to backend_apis and the response status is 404, it definitely means that the Backend is not found.

app/controllers/stats/data/backend_apis_controller.rb
Martouta, line 40: I think that this is lacking a test 👀

test/integration/provider/admin/api_docs/account_data_controller_test.rb
Martouta: Fixed in https://github.com/3scale/porta/pull/1805 😄

guicassolato: Let's get that one merged first and then I'll rebase this one and remove this line.

app/lib/backend_api_logic/metric_extension.rb
```diff
@@ -32,18 +32,26 @@ def hits_extended_system_name_as_sql
       def hits_extended_system_name_regex
         /\Ahits(#{Regexp.escape(SYSTEM_NAME_SUFFIX_SEPARATOR)}\d+)?\z/
       end
+
+      def build_extended_system_name(system_name, owner_id:)
```
Martouta, line 36: This probably should have a unit test of its own 😄

app/controllers/stats/data/backend_apis_controller.rb
guicassolato: It reuses the implementation of `Stats::Data::BaseController` (same as `Stats::Data::ServicesController`), where XML is supported. No need to override only to remove for the API backend, IMO. Once we remove XML from the base, it’s going to be removed for all endpoints.

Martouta: I am aware of that and that's why I didn't say anything in the routes.rb or in any other part of the controller. But at least we shouldn't add the documentation for XML for the new endpoint as we are doing in this line. That is like inviting the customer to use it when we do not want that.
So here it should be just:
`##~ e.path = "/stats/backend_apis/{backend_api_id}/usage.json"`

app/controllers/provider/admin/backend_apis/stats/base_controller.rb
guicassolato: This TODO has been copied from https://github.com/3scale/porta/blob/edf4cfcb170c6a6613b652796f6fab6f70674d91/app/controllers/stats/service_base_controller.rb#L18

Removed.

app/controllers/provider/admin/backend_apis/stats/base_controller.rb
guicassolato: In the end, I think everything related to `BackendApi` is going to be removed for master, including this ability.

app/controllers/stats/data/backend_apis_controller.rb
guicassolato: Done.

app/lib/backend_api_logic/metric_extension.rb
guicassolato: Done.

app/controllers/stats/data/backend_apis_controller.rb
guicassolato: Done.

PR #1798, opened 2020-03-27
Prevent creating application with forbidden plan in UI

josemigallas:

Restrain the list of plans returned to a user, depending on his permissions. [This line](https://github.com/3scale/porta/blob/031493986866f026054a06c7b575c325cf8d446f/app/controllers/buyers/applications_controller.rb#L61) probably.

[THREESCALE-3268: Admin Portal member user with the access "Developer Accounts -- Applications" is able to create new Applications using the UI for Services they don't have permissions
](https://issues.redhat.com/browse/THREESCALE-3268)

**Verification steps**
See JIRA ticket.

### DIFF
--- a/app/controllers/buyers/applications_controller.rb
+++ b/app/controllers/buyers/applications_controller.rb
@@ -12,8 +12,6 @@ class Buyers::ApplicationsController < FrontendController
   before_action :find_cinstance, :except => [:index, :create, :new]
   before_action :find_provider,  only: %i[new create update]
 
-  before_action :find_application_plan,          :only => :create
-
   activate_menu :buyers
 
   layout 'provider'
@@ -24,11 +22,8 @@ class Buyers::ApplicationsController < FrontendController
     activate_menu :audience, :applications, :listing
 
     @states = Cinstance.allowed_states.collect(&:to_s).sort
-    accessible_services = current_account.accessible_services
-    @services = accessible_services.includes(:application_plans)
     @search = ThreeScale::Search.new(params[:search] || params)
-    @application_plans = current_account.application_plans.stock
-    @stock_and_custom_application_plans = current_account.application_plans.size
+    @application_plans = accessible_plans
 
     if params[:service_id]
       @service = accessible_services.find params[:service_id]
@@ -36,7 +31,7 @@ class Buyers::ApplicationsController < FrontendController
     end
 
     if params[:application_plan_id]
-      @plan = current_account.application_plans.find params[:application_plan_id]
+      @plan = @application_plans.find params[:application_plan_id]
       @search.plan_id = @plan.id
       @service ||= @plan.service
     end
@@ -58,7 +53,7 @@ class Buyers::ApplicationsController < FrontendController
   def new
     @cinstance = @buyer.bought_cinstances.build
     extend_cinstance_for_new_plan
-    @plans = @provider.application_plans.stock
+    @plans = accessible_plans.stock
 
     if params[:account_id]
       @account = current_account.buyers.find params[:account_id]
@@ -68,11 +63,13 @@ class Buyers::ApplicationsController < FrontendController
 
   # TODO: this should be done by buy! method
   def create
+    @plans = accessible_plans.stock
+    application_plan = @plans.find plan_id
     service_plan = if service_plan_id = params[:cinstance].delete(:service_plan_id)
-                     @application_plan.service.service_plans.find(service_plan_id)
+                     application_plan.service.service_plans.find(service_plan_id)
                    end
 
-    @cinstance = current_account.provider_builds_application_for(@buyer, @application_plan, params[:cinstance], service_plan)
+    @cinstance = current_account.provider_builds_application_for(@buyer, application_plan, params[:cinstance], service_plan)
     @cinstance.validate_human_edition!
 
     if @cinstance.save
@@ -80,7 +77,6 @@ class Buyers::ApplicationsController < FrontendController
       redirect_to(admin_service_application_path(@cinstance.service, @cinstance))
     else
       @cinstance.extend(AccountForNewPlan)
-      @plans = @provider.application_plans
       render :action => :new
     end
   end
@@ -126,7 +122,7 @@ class Buyers::ApplicationsController < FrontendController
   def change_plan
     # there is no need to query available_application_plans as we already have a validation
     service = @cinstance.service
-    new_plan = service.application_plans.stock.find(params[:cinstance][:plan_id])
+    new_plan = accessible_plans.stock.find(plan_id)
     @cinstance.provider_changes_plan!(new_plan)
     flash[:notice] = "Plan changed to '#{new_plan.name}'."
     redirect_to admin_service_application_url(service, @cinstance)
@@ -191,8 +187,18 @@ class Buyers::ApplicationsController < FrontendController
     @provider = current_account
   end
 
-  def find_application_plan
-    @application_plan = @provider.application_plans.find params[:cinstance][:plan_id]
+  def accessible_services
+    @accessible_services ||= current_user.accessible_services.includes(:application_plans)
+  end
+
+  helper_method :accessible_services
+
+  def accessible_plans
+    current_account.application_plans.where(issuer: accessible_services)
+  end
+
+  def plan_id
+    @plan_id ||= params.require(:cinstance).permit(:plan_id).tap { |plan_params| plan_params.require(:plan_id) }[:plan_id]
   end
 
   def authorize_partners
--- a/app/views/buyers/applications/_listing.html.slim
+++ b/app/views/buyers/applications/_listing.html.slim
@@ -18,7 +18,7 @@ div id="bulk-operations" style="display: none;"
     = javascript_include_tag 'bulk_operations.js'
     = javascript_include_tag 'vendor/jquery.tipsy.js'
 
-- show_application_plans = @stock_and_custom_application_plans > 1 && !master_on_premises?
+- show_application_plans = @application_plans.size > 1 && !master_on_premises?
 
 table class="data"
   thead
@@ -66,13 +66,13 @@ table class="data"
 
           - if display_view_portion?(:service)
             th
-              = s.collection_select :service_id, @services, :id, :name,
+              = s.collection_select :service_id, accessible_services, :id, :name,
                                       { include_blank: true },
                                       { disabled: @service }
 
           - if show_application_plans
-            - if @services
-              th = s.grouped_collection_select :plan_id, @services,
+            - if display_view_portion?(:service)
+              th = s.grouped_collection_select :plan_id, accessible_services,
                                               :application_plans, :name, :id, :name,
                                               { include_blank: true },
                                               { disabled: @plan }
--- a/test/integration/buyers/applications_controller_test.rb
+++ b/test/integration/buyers/applications_controller_test.rb
@@ -22,8 +22,7 @@ class Buyers::ApplicationsTest < ActionDispatch::IntegrationTest
     def setup
       @provider = FactoryBot.create(:provider_account)
 
-      host! @provider.admin_domain
-      provider_login_with @provider.admins.first.username, "supersecret"
+      login! provider
 
       #TODO: dry with @ignore-backend tag on cucumber
       stub_backend_get_keys
@@ -31,19 +30,93 @@ class Buyers::ApplicationsTest < ActionDispatch::IntegrationTest
       stub_backend_utilization
     end
 
+    attr_reader :provider
+
     test 'index shows the services column when the provider is multiservice' do
-      @provider.services.create!(name: '2nd-service')
-      assert @provider.reload.multiservice?
+      provider.services.create!(name: '2nd-service')
+      assert provider.reload.multiservice?
       get admin_buyers_applications_path
       page = Nokogiri::HTML::Document.parse(response.body)
       assert page.xpath("//tr").text.match /Service/
     end
 
     test 'index does not show the services column when the provider is not multiservice' do
-      refute @provider.reload.multiservice?
+      refute provider.reload.multiservice?
       get admin_buyers_applications_path
       page = Nokogiri::HTML::Document.parse(response.body)
       refute page.xpath("//tr").text.match /Service/
+    end
+
+    test 'index shows an application of a custom plan' do
+      service = provider.default_service
+      buyer = FactoryBot.create(:buyer_account, provider_account: provider)
+      app_plan = FactoryBot.create(:application_plan, issuer: service)
+      custom_plan = app_plan.customize
+      cinstance = FactoryBot.create(:cinstance, user_account: buyer, plan: custom_plan, name: 'my custom cinstance')
+
+      get admin_buyers_applications_path
+      assert_response :success
+      page = Nokogiri::HTML::Document.parse(response.body)
+      assert page.xpath("//td").text.match /my custom cinstance/
+    end
+
+    test 'member cannot create an application when lacking access to a service' do
+      forbidden_service = FactoryBot.create(:service, account: provider)
+      forbidden_plan = FactoryBot.create(:application_plan, issuer: forbidden_service)
+      forbidden_service_plan = FactoryBot.create(:service_plan, issuer: forbidden_service)
+
+      authorized_service = FactoryBot.create(:service, account: provider)
+      authorized_plan = FactoryBot.create(:application_plan, issuer: authorized_service)
+      authorized_service_plan = FactoryBot.create(:service_plan, issuer: authorized_service)
+
+      buyer = FactoryBot.create(:buyer_account, provider_account: provider)
+
+      member = FactoryBot.create(:member, account: provider,
+                                          member_permission_ids: [:partners])
+      member.activate!
+      FactoryBot.create(:member_permission, user: member,
+                                            admin_section: :services,
+                                            service_ids: [authorized_service.id])
+
+      login_provider provider, user: member
+
+      post admin_buyers_account_applications_path(account_id: buyer.id), cinstance: {
+        plan_id: forbidden_plan.id,
+        name: 'Not Allowed!',
+        service_plan_id: forbidden_service_plan.id
+      }
+
+      assert_response :not_found
+
+      post admin_buyers_account_applications_path(account_id: buyer.id), cinstance: {
+        plan_id: authorized_plan.id,
+        name: 'Allowed',
+        service_plan_id: authorized_service_plan.id
+      }
+
+      assert_response :found
+    end
+
+    test 'buying a stock plan is allowed but buying a custom plan is not' do
+      service = provider.default_service
+      initial_plan = FactoryBot.create(:application_plan, issuer: service)
+      buyer = FactoryBot.create(:buyer_account, provider_account: provider)
+      cinstance = FactoryBot.create(:cinstance, user_account: buyer, plan: initial_plan)
+
+      app_plan = FactoryBot.create(:application_plan, issuer: service)
+      custom_plan = app_plan.customize
+
+
+      put change_plan_admin_buyers_application_path(account_id: buyer.id, id: cinstance.id), cinstance: {plan_id: custom_plan.id}
+
+      assert_response :not_found
+      assert_equal initial_plan.id, cinstance.reload.plan_id
+
+
+      put change_plan_admin_buyers_application_path(account_id: buyer.id, id: cinstance.id), cinstance: {plan_id: app_plan.id}
+
+      assert_redirected_to admin_service_application_path(service, cinstance)
+      assert_equal app_plan.id, cinstance.reload.plan_id
     end
   end
 end


app/controllers/buyers/applications_controller.rb
```diff
@@ -12,7 +12,8 @@ class Buyers::ApplicationsController < FrontendController
   before_action :find_cinstance, :except => [:index, :create, :new]
   before_action :find_provider,  only: %i[new create update]
 
-  before_action :find_application_plan,          :only => :create
+  before_action :find_accessible_services,
+                :find_accessible_plans,    only: %i[index new create change_plan]
```
Martouta, line 16: Why not for `update` and `edit`?
image

Btw, no idea what these `bulk` actions are and if they affect us (if we should apply that restriction for those or not):
image

app/controllers/buyers/applications_controller.rb
```diff
@@ -191,8 +191,12 @@ def find_provider
     @provider = current_account
   end
 
-  def find_application_plan
-    @application_plan = @provider.application_plans.find params[:cinstance][:plan_id]
+  def find_accessible_services
+    @services = current_user.accessible_services
+  end
```
Martouta, line 196: We should still have the `includes(:application_plans)` for the `index` case as we used to,  in order to avoid a N+1.

https://github.com/3scale/porta/blob/c1dc253ac89c9b2e72dc064260ea9ae740c096ca/app/controllers/buyers/applications_controller.rb#L28

Something like this:
```ruby
def find_accessible_services
    @services = current_user.accessible_services
    @services = @services.includes(:application_plans) if action_name == 'index'
end
```

Martouta, line 196: We should still have the `includes(:application_plans)` for the `index` case as we used to,  in order to avoid a N+1.

https://github.com/3scale/porta/blob/c1dc253ac89c9b2e72dc064260ea9ae740c096ca/app/controllers/buyers/applications_controller.rb#L28

Something like this:
```ruby
def find_accessible_services
    @services = current_user.accessible_services
    @services = @services.includes(:application_plans) if action_name == 'index'
end
```

app/controllers/buyers/applications_controller.rb
```diff
@@ -126,7 +126,7 @@ def resume
   def change_plan
     # there is no need to query available_application_plans as we already have a validation
     service = @cinstance.service
-    new_plan = service.application_plans.stock.find(params[:cinstance][:plan_id])
+    new_plan = @application_plans.stock.find(params[:cinstance][:plan_id])
```
Martouta, line 129: https://github.com/3scale/porta/pull/1798/files#r402593556

```ruby
    new_plan = @application_plans.stock.find(params.require(:cinstance).permit(:plan_id))
```

app/controllers/buyers/applications_controller.rb
```diff
@@ -68,19 +65,22 @@ def new
 
   # TODO: this should be done by buy! method
   def create
+    # If not an accessible plan, it returns 404. TODO: it should redirect to :new with flash[:error]
+    application_plan = @application_plans.find params[:cinstance][:plan_id]
```
Martouta, line 69: Then, something like this:
```ruby
plan_id = @application_plans.find params.fetch(:cinstance, {}).permit(:plan_id)
application_plan = @application_plans.find_by(id: plan_id) if plan_id
unless application_plan
  flash[:error] = 'Application Plan not found' # even better if you do it by i18n ;)
  return redirect_to(new_admin_buyers_account_application_path)
end
```

app/controllers/buyers/applications_controller.rb
josemigallas: > update & edit
It's not possible to edit the plan_id; it needs to go tru `change_plan`.

> bulk actions
Not sure about this but I'd say it also goes thru `change_plan`

Martouta: Awesome, makes sense 👍

app/controllers/buyers/applications_controller.rb
josemigallas: Right. Although I think it would be better to call it in `index`?

Martouta: Yes, makes sense 😄

test/integration/buyers/applications_controller_test.rb
```diff
@@ -45,5 +45,17 @@ def setup
...
     end
+
+    test 'user do not have permission to access the service' do
+      plan = @provider.services.first.application_plans.create(name: 'plan')
+      buyer = FactoryBot.create(:buyer_account, provider_account: @provider)
+
+      User.any_instance.stubs(accessible_services: [])
+
+      # post admin_api_account_applications_path(account_id: @buyer.id, plan_id: @plan.id, format: :xml)
+      post admin_buyers_applications_path(account_id: buyer.id, plan_id: plan.id, format: :xml)
```
Martouta, line 56: Remove that format because it is for the UI and not the API so it won't be XML

josemigallas: Moving the test to this file leads to very weird errors 😕 
```
# User.any_instance.stubs(accessible_services: [])

NoMethodError: undefined method `where_values_hash' for []:Array
```

```
# post admin_buyers_applications_path(account_id: buyer.id, plan_id: plan.id)

ActionController::UrlGenerationError: No route matches {:action=>"new", :controller=>"buyers/applications"} missing required keys: [:account_id]
from /Users/jgallaso/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/actionpack-5.0.7.2/lib/action_dispatch/journey/formatter.rb:51:in `generate'
```

app/controllers/buyers/applications_controller.rb
```diff
@@ -12,7 +12,8 @@ class Buyers::ApplicationsController < FrontendController
   before_action :find_cinstance, :except => [:index, :create, :new]
   before_action :find_provider,  only: %i[new create update]
 
-  before_action :find_application_plan,          :only => :create
+  before_action :find_application_plan, only: :create
+  before_action :accessible_plans,      only: :index
```
guicassolato, line 16: You don't need this.

```suggestion
```

app/controllers/buyers/applications_controller.rb
```diff
@@ -24,19 +25,16 @@ def index
     activate_menu :audience, :applications, :listing
 
     @states = Cinstance.allowed_states.collect(&:to_s).sort
-    accessible_services = current_account.accessible_services
-    @services = accessible_services.includes(:application_plans)
     @search = ThreeScale::Search.new(params[:search] || params)
-    @application_plans = current_account.application_plans.stock
-    @stock_and_custom_application_plans = current_account.application_plans.size
+    @services = accessible_services.includes(:application_plans)
```
guicassolato, line 29: It looks like `#accessible_services` already sets `@service`. You may want to move the `includes(:application_plans)` to that method and remove this line.

app/views/buyers/applications/_listing.html.slim
```diff
@@ -18,7 +18,7 @@ div id="bulk-operations" style="display: none;"
     = javascript_include_tag 'bulk_operations.js'
     = javascript_include_tag 'vendor/jquery.tipsy.js'
 
-- show_application_plans = @stock_and_custom_application_plans > 1 && !master_on_premises?
+- show_application_plans = @application_plans.size > 1 && !master_on_premises?
```
guicassolato, line 21: https://github.com/3scale/porta/pull/1798/files#r404884739

Now I see why the `before_filter`. Maybe rename the method in the controller then.

app/controllers/buyers/applications_controller.rb
```diff
@@ -24,19 +25,16 @@ def index
...
 
     if params[:service_id]
-      @service = accessible_services.find params[:service_id]
+      @service = @services.find params[:service_id]
```
guicassolato, line 32: After https://github.com/3scale/porta/pull/1798/files#r404886489, you may want to keep

```suggestion
      @service = accessible_services.find params[:service_id]
```

test/integration/buyers/applications_controller_test.rb
```diff
@@ -31,19 +31,59 @@ def setup
...
+      assert_response :forbidden
+
+      post admin_buyers_account_applications_path(account_id: buyer.id), cinstance: {
+        plan_id: authorized_plan.id,
+        name: 'Allowed',
+        service_plan_id: authorized_service_plan.id
+      }
+
+      assert_response :found
+    endesponse :forbidden
```
guicassolato, line 86:
```suggestion
```

🙂

app/controllers/buyers/applications_controller.rb
josemigallas: I added it because `@application_plans` is used in the view. Is there a more preferable way to do it?

test/integration/buyers/applications_controller_test.rb
josemigallas: 🤦‍♂

app/views/buyers/applications/_listing.html.slim
josemigallas: Like it was before, `find_application_plans`? I also added an alias to make it look cleaner 🦐

guicassolato: Personally, I like to use the `before_filter` when at least one of the the following conditions apply:
1. more than one action depends on what it does;
2. it can abort the request, raising an immediate response (e.g. 404, 403, etc) before getting into the details of the specific action requested, i.e. before starting the actual transaction that could cause the state of the application to mutate. Not necessarily because the payload is invalid (this would be too much for a controller to decide), but because the request can be rejected right away.

I think your `find_accessible_plans` (aliases `accessible_plans`) is somehow in the middle. It doesn't seem to fit any of the conditions above entirely. It won't raise a 404 or 403 in any case, so condition `#2` is completely out. But even condition `#1` is not fully satisfied. The instance variable `@application_plans` is only needed for the `#index` action, while other actions do need to fetch "accessible plans" as well but without setting the instance variable.

I see no action referring to `accessible_plans` more than once. So this is what I would do. I would remove the before filter and the memoization. I'd make the `#index` action to set `@applications_plans = accessible_plans` and then use `@applications_plans` instead of `accessible_plans` further down. As for the other actions, just keep `accessible_plans` as is.

app/controllers/buyers/applications_controller.rb
```diff
@@ -24,19 +22,17 @@ def index
     activate_menu :audience, :applications, :listing
 
     @states = Cinstance.allowed_states.collect(&:to_s).sort
-    accessible_services = current_account.accessible_services
-    @services = accessible_services.includes(:application_plans)
+    @services = accessible_services
     @search = ThreeScale::Search.new(params[:search] || params)
-    @application_plans = current_account.application_plans.stock
-    @stock_and_custom_application_plans = current_account.application_plans.size
+    @application_plans = accessible_plans
```
guicassolato, line 27:
```suggestion
    @application_plans = accessible_plans.stock
```

app/controllers/buyers/applications_controller.rb
```diff
@@ -191,8 +188,12 @@ def find_provider
     @provider = current_account
   end
 
-  def find_application_plan
-    @application_plan = @provider.application_plans.find params[:cinstance][:plan_id]
+  def accessible_services
+    current_user.accessible_services.includes(:application_plans)
```
guicassolato, line 192: I think this is worth memoizing. `#index` will invoke it twice, for example. Plus, by doing this, you can save a couple lines of changes in the `#index` action too.

app/views/buyers/applications/_listing.html.slim
```diff
@@ -77,7 +77,7 @@ table class="data"
                                               { include_blank: true },
                                               { disabled: @plan }
             - else
-              th = s.collection_select :plan_id, @application_plans, :id, :name,
+              th = s.collection_select :plan_id, @application_plans.stock, :id, :name,
```
guicassolato, line 80:
```suggestion
              th = s.collection_select :plan_id, @application_plans, :id, :name,
```

app/controllers/buyers/applications_controller.rb
guicassolato: I think this comment is still valid.

app/controllers/buyers/applications_controller.rb
```diff
@@ -68,19 +64,20 @@ def new
 
   # TODO: this should be done by buy! method
   def create
+    @plans = accessible_plans
+    application_plan = @plans.find(params[:cinstance][:plan_id])
```
Martouta, line 68: I think that we should enforce those params to be present 😄 
```suggestion
    application_plan = @plans.find params.require(:cinstance).permit(:plan_id).tap do |plan_params|
      plan_params.require(:plan_id)
    end
```

app/controllers/buyers/applications_controller.rb
Martouta: I wonder why the `.stock` isn't inside the `accessible_plans` method in this controller 🤔

app/controllers/buyers/applications_controller.rb
hallelujah: Well it is a Relation, memoizing or not does not change it as it has not been queried yet.
But yes good remark

app/controllers/buyers/applications_controller.rb
```diff
@@ -24,19 +22,17 @@ def index
     activate_menu :audience, :applications, :listing
 
     @states = Cinstance.allowed_states.collect(&:to_s).sort
-    accessible_services = current_account.accessible_services
-    @services = accessible_services.includes(:application_plans)
+    @services = accessible_services
```
hallelujah, line 25: Is `@services` used in the view? In this case why not making an helper in the view? `method_helper :accessible_services`
That way we explicitly tell in the view that it is about accessible_services and not any services.
So the view cannot be used by anything else though it will break SRP ...
Dunno as you wish

app/controllers/buyers/applications_controller.rb
josemigallas: I don't know what `.stock` is for but it has been moved to the view I think

Martouta:
I will open new PR(s) 😉

app/controllers/buyers/applications_controller.rb
```diff
@@ -24,19 +22,16 @@ def index
     activate_menu :audience, :applications, :listing
 
     @states = Cinstance.allowed_states.collect(&:to_s).sort
-    accessible_services = current_account.accessible_services
-    @services = accessible_services.includes(:application_plans)
     @search = ThreeScale::Search.new(params[:search] || params)
-    @application_plans = current_account.application_plans.stock
-    @stock_and_custom_application_plans = current_account.application_plans.size
+    @application_plans = accessible_plans # .stock
```
Martouta, line 26: `features/old/api/plans/customize_application_plan.feature:55` fails with stock

app/controllers/buyers/applications_controller.rb
```diff
@@ -126,7 +122,7 @@ def resume
   def change_plan
     # there is no need to query available_application_plans as we already have a validation
     service = @cinstance.service
-    new_plan = service.application_plans.stock.find(params[:cinstance][:plan_id])
+    new_plan = accessible_plans.find(plan_id) # .stock
```
Martouta, line 125: test pass both with and without stock so I should add a test whatever it should be.
the decision of which kinda depends on https://github.com/3scale/porta/pull/1798/files#r441727410

app/controllers/buyers/applications_controller.rb
hallelujah: We should not list custom plans in this screen, I think your test fails for something else

image

Martouta: sorry @hallelujah but I do not understand since those 2 sentences seem "incompatible" with each other.

> I think your test fails for something else

I'm not sure if we are understanding each other... The test current passes if we do not have the `.stock` (just as you are seeing it now in the review) but fails if the code adds the `.stock`.

You just said that the page should not contain stock, right? If that is the case, the test is correct, and it is passing right now correctly.

> We should not list custom plans in this screen

Okidoki, but why we have this in the `master` branch?
https://github.com/3scale/porta/blob/f7ff1025c4c37570082c4a35623a90dd33d85bc0/app/controllers/buyers/applications_controller.rb#L30

And another question:
Why you don't have the list of services in that screenshot? are you sure that we are on the same page?

hallelujah: ` @application_plans = current_account.application_plans.stock` means no custom plans
The screenshot is a list of application plans

I got it from `http://provider-admin.example.com/apiconfig/services/{service_id}/application_plans`

hallelujah: OK I got confused, it should be the application plan of an account not the defined plans in general
Sorry

Martouta: `http://provider-admin.example.com/apiconfig/services/{service_id}/application_plans` is not the view of this controller 😄 the view of this controller is `http://provider-admin.example.com//buyers/applications`

I know what the scope `stock` is 😉 
https://github.com/3scale/porta/blob/f7ff1025c4c37570082c4a35623a90dd33d85bc0/app/models/plan.rb#L130-L131

My issue now is:
1. How come in the `master` branch we have here the `.stock` but with this change... the test passes without `.stock` but not without it?
2. Should we show custom plans in that specific view or not? According to the `master` branch, I'd say that we do not, but according to the test I'd say that we do... I'm confused 😕

app/controllers/buyers/applications_controller.rb
hallelujah: I think it is related to https://github.com/3scale/porta/blob/master/test/integration/user-management-api/buyers_application_plans_test.rb#L113

You cannot buy a custom plan so changing to a custom one does not make sense. As a custom one is tied to only one application

Martouta: Makes sense 👍  I will add a test 😉

app/controllers/buyers/applications_controller.rb
Martouta: Thomas told me that the applications page should definitely show all applications and that if they are on a custom plan is irrelevant 👍

app/controllers/buyers/applications_controller.rb
```diff
@@ -191,8 +187,18 @@ def find_provider
     @provider = current_account
   end
 
-  def find_application_plan
-    @application_plan = @provider.application_plans.find params[:cinstance][:plan_id]
+  def accessible_services
+    @accessible_services ||= current_user.accessible_services.includes(:application_plans)
+  end
+
+  helper_method :accessible_services
```
macejmic, line 194: a bit weird that `accessible_services` is a helper method and `accessible_plans` not..

app/controllers/buyers/applications_controller.rb
```diff
@@ -191,8 +187,18 @@ def find_provider
...
+
+  def accessible_plans
+    current_account.application_plans.where(issuer: accessible_services)
+  end
+
+  def plan_id
+    @plan_id ||= params.require(:cinstance).permit(:plan_id).tap { |plan_params| plan_params.require(:plan_id) }[:plan_id]
```
macejmic, line 201: so how the params should look like? `cintances[:plan_id][:plan_id]` ?

Martouta: `{cinstance: {plan_id: 1}}` for example.
Look at the test 😄 but it is exactly the same format as before, just that now the param is mandatory and otherwise raises the proper error instead of trying to find by nil.
look at https://github.com/3scale/porta/pull/1798#discussion_r408688711

app/controllers/buyers/applications_controller.rb
Martouta: because accessible_services is used from the view, look:
https://github.com/3scale/porta/blob/c10f20c670c91c4ed63eb43bf601cbc1fe0fd028/app/controllers/buyers/applications_controller.rb#L55
https://github.com/3scale/porta/pull/1798/files#diff-35521f5b3d5085d49b2eac64c87211feR69
https://github.com/3scale/porta/pull/1798#discussion_r409777996

app/controllers/buyers/applications_controller.rb
hallelujah: Why not just `params.require(:cinstance).require(:plan_id)`?

Martouta: That doesn't work @hallelujah . It always needs a permit at the end of the chain.
Look at [the documentation of ActionController::Parameters#require](https://api.rubyonrails.org/classes/ActionController/Parameters.html#method-i-require)
image

guicassolato, line 201: Is `tap` really needed here? Does it work simply like this?

```suggestion
    @plan_id ||= params.require(:cinstance).require(:plan_id)
```

Martouta: @guicassolato same as https://github.com/3scale/porta/pull/1798#discussion_r443405932

hallelujah: I do not get it, why is it safer?

hallelujah: OK got it, imagining people sending a has as `:plan_id`

```ruby

params = ActionController::Parameters.new(person: {name: {a: :b}})
params.require(:person).require(:name)
# => <ActionController::Parameters {"a" => "b"} permitted: false>

params.require(:person).permit(:name)
# => <ActionController::Parameters {} permitted: false>
```

PR #1800, opened 2020-03-27
Add types and encapsulates translations

josemigallas:
Proposal for some improvements in the base PR:
* Creates types for the translations: it's very important to be able to check what "pages" we are requesting.
* Encapsulates translations (JSON) inside their modules (folders). This way no matter if we use one or multiple jsons for each language, the code stays consistent.
* Creates types for the locales (languages really): i18next handles simple strings.

One thing to keep in mind: I think we're starting to mix up _locales_ with _languages_. A locale takes localization into account whereas language does not. That is basically why I defined a `ISupportedLanguages` instead of `ILocales`, because we're using `en` instead of `en-US`.

### DIFF
--- a/portafly/src/i18n/i18n.tsx
+++ b/portafly/src/i18n/i18n.tsx
@@ -1,7 +1,7 @@
 import i18n, { FormatFunction } from 'i18next'
 import LanguageDetector from 'i18next-browser-languagedetector'
 import { initReactI18next } from 'react-i18next'
-import { EN } from 'i18n/locales'
+import { ITranslationsPages, Translations, EN } from 'i18n'
 
 const formatFn: FormatFunction = (value, format) => {
   if (format === 'uppercase') return value.toUpperCase()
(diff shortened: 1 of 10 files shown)


portafly/src/i18n/i18n.tsx
```diff
@@ -1,36 +1,30 @@
...
-import { EN } from 'i18n/locales'
+import { ITranslationsPages, Translations, EN } from 'i18n'
 
 const formatFn: FormatFunction = (value, format) => {
   if (format === 'uppercase') return value.toUpperCase()
   if (format === 'lowercase') return value.toLowerCase()
   return value
 }
 
+const sections: Array<ITranslationsPages> = ['shared', 'overview', 'analytics', 'applications', 'integration']
```
thomasmaas: so this will be a growing list, right? That probably needs to be repeated somewhere else, like for navigation? Maybe we can define it once somewhere and take it from there? (or maybe that doesn't make sense at all)

josemigallas: There is a way to auto-define this list but it's kind a ugly and doesn't even have type checking. This list coincides with `ItranslationsPage` really and it comes from reading the keys from the translations JSON.
TLDR: `sections = Object.keys(Translations.en)` and we don't need to manage any array but... it's ugly

damianpm:
@josemigallas I've rebased to fix some conflicts, yo may rebase too

portafly/src/i18n/locales/index.ts
```diff
@@ -1,15 +1,11 @@
...
+import * as en from 'i18n/locales/en'
 
-const EN = {
-  SHARED: shared,
-  OVERVIEW: overview,
-  ANALYTICS: analytics,
-  APPLICATIONS: applications,
-  INTEGRATION: integration
+export type ITranslationsPages = keyof typeof en
+export type ITranslations = { [P in ITranslationsPages]: any}
```
damianpm:
```suggestion
export type ITranslations = { [P in ITranslationsPages]: any }
```

josemigallas: This should have been fixed automagically... There seems to be [a bug](https://github.com/typescript-eslint/typescript-eslint/issues/1784) in eslint/typescript 😩

PR #1806, opened 2020-04-03
Introducing factories for better testing

didierofrivia:

Aiming to have more solid and realistic examples when testing, an object factory specific for testing is added. [Fishery](https://github.com/thoughtbot/fishery)  is used to tackle this, from Thoughtbot, same company that produced Factory Bot currently used in Porta.

fixes https://issues.redhat.com/browse/THREESCALE-4646

**Verification steps** 
Run within portafly directory: `yarn test src/tests/pages/DeveloperAccountList.test.tsx`
And it should be green. Also, feel free to `console.log` and analyze the object created.

### DIFF
--- a/portafly/src/tests/factories/developer-account.ts
+++ b/portafly/src/tests/factories/developer-account.ts
@@ -1,0 +1,14 @@
+import { Factory } from 'fishery'
+import { IDeveloperAccount } from 'types'
+
+const DeveloperAccount = Factory.define<IDeveloperAccount>(({ sequence }) => ({
+  admin_name: 'Oswell E. Spencer',
+  apps_count: 3,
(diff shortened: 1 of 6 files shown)


portafly/src/tests/factories/developer-account.ts
```diff
@@ -0,0 +1,14 @@
+import { Factory } from 'fishery'
+import { IDeveloperAccount } from 'types'
+
+const DeveloperAccount = Factory.define<IDeveloperAccount>(({ sequence }) => ({
```
damianpm, line 4: Can we have another type of account, a not-developer account ?

didierofrivia: We eventually will, and when we do, we could add the factory

PR #1809, opened 2020-04-06
Removing errorception code

damianpm:
We will use Bugsnag to track runtime javascript errors so we should get rid of any reference to Errorception in our cod

https://issues.redhat.com/browse/THREESCALE-4355

### DIFF
--- a/app/assets/javascripts/ajax_events.js
+++ b/app/assets/javascripts/ajax_events.js
@@ -9,7 +9,5 @@
     .on('ajax:error', function(event, xhr, status, error){
       alert('Request failed - ' + status);
       console.error(error);
-      // report error to errorception in async manner
-      setTimeout(function(){ throw(error) }, 1);
     });
 }());
--- a/app/javascript/src/Dashboard/index.jsx
+++ b/app/javascript/src/Dashboard/index.jsx
@@ -17,16 +17,7 @@ export function widget (url) {
       return
     }
 
-    let errorception = window._errs
     let exception = new Error(`Failed to load ${url} with ${error} (${status})`)
-
-    if (errorception) {
-      let meta = errorception.meta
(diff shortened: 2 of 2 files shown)


app/assets/javascripts/ajax_events.js
```diff
@@ -6,10 +6,4 @@
     .on('ajaxComplete ajax:complete pjax:end', function(event, xhr, status){
       $('#ajax-in-progress').remove();
     })
-    .on('ajax:error', function(event, xhr, status, error){
-      alert('Request failed - ' + status);
-      console.error(error);
-      // report error to errorception in async manner
-      setTimeout(function(){ throw(error) }, 1);
```
josemigallas, line 13: Besides the fact that the comment mentions it, how's this `alert` and `throw(error)` related to errorception? 🤔Maybe it's still used by bugsnag?

damianpm: I'm not sure, it's the 1st time I deal with either bugsnag or errorception, but I don't think is used by bugsnag, otherwise tests like [this](https://github.com/3scale/porta/blob/master/test/unit/events/importers/alert_importer_test.rb#L99) or [this](https://github.com/3scale/porta/blob/edf4cfcb170c6a6613b652796f6fab6f70674d91/test/unit/event_store/repository_test.rb#L104) would fail

damianpm: And I think the `alert` is not related to anything, is just for the user , I guess

damianpm: Maybe @didierofrivia knows

hallelujah: Only the `setTimeout` (line 13) needs to be removed.
Basically bugsnag or errorception catch the error thrown, so here it will catch a failed ajax request.

damianpm: Thanks @hallelujah

PR #1812, opened 2020-04-08
Fix/threescale 4531 captcha auto 2

macejmic:
Reset password email is not being send when spam protection is set to suspicious. Also, the captcha should be shown when the previous request was considered as possible spam. 

Which issue(s) this PR fixes
https://issues.redhat.com/browse/THREESCALE-4531

### DIFF
--- a/app/lib/three_scale/spam_protection/protector.rb
+++ b/app/lib/three_scale/spam_protection/protector.rb
@@ -35,7 +35,7 @@ module ThreeScale::SpamProtection
 
       delegate :template, to: :form
       delegate :logged_in?, to: :template, allow_nil: true
-      delegate :spam?, :checks, to: :protector
+      delegate :checks, to: :protector
       delegate :captcha_configured?, to: Recaptcha
 
       def initialize(form, protector)
@@ -60,11 +60,11 @@ module ThreeScale::SpamProtection
       end
 
       def captcha_needed?
-        captcha_required? || (!http_method.get? && spam?)
+        captcha_required? || possible_spam?
       end
 
       def enabled?
-        not logged_in? and level != :none
+        !logged_in? && level != :none
       end
 
       def to_str
@@ -89,6 +89,28 @@ module ThreeScale::SpamProtection
       end
 
       alias to_s to_str
+
+      private
+
+      def request_session
+        return {} if template.controller.blank?
+
+        template.controller.request.session
+      end
+
+      def possible_spam?
+        http_method.get? ? marked_as_possible_spam? : mark_possible_spam
+      end
+
+      def marked_as_possible_spam?
+        request_session[:marked_as_possible_spam_until].to_i > Time.now.utc.to_i
+      end
+
+      def mark_possible_spam
+        return false unless protector.spam?
+
+        request_session[:marked_as_possible_spam_until] = (Time.now.utc + 5.minutes).to_i
+      end
     end
 
     def form(form)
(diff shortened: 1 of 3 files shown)


app/lib/three_scale/spam_protection/protector.rb
```diff
@@ -60,11 +60,11 @@ def captcha_required?
       end
 
       def captcha_needed?
-        captcha_required? || (!http_method.get? && spam?)
+        captcha_required? || spam_post_request? || spam_get_request?
```
guicassolato, line 63: So in the end it does not matter if it's `http_method.get?` or `!http_method.get?`? It looks like this can be reduced to 

```suggestion
        captcha_required? || considered_as_spam? || was_considered_as_spam?
```

though I'd suggest for readability

```suggestion
        captcha_required? || considered_spam?
```

and then

```ruby
def considered_spam?
  considered_as_spam? || was_considered_as_spam?
end

File: app/lib/three_scale/spam_protection/protector.rb [hunk 2]
```diff
@@ -89,6 +89,33 @@ def to_str
...
+
+      def spam_post_request?
+        !http_method.get? && considered_as_spam?
+      end
+
+      def spam_get_request?
+        http_method.get? && was_considered_as_spam?
+      end
+
+      def was_considered_as_spam?
```
guicassolato, line 109:
```suggestion
      def has_been_considered_as_spam?
```

The condition seems to suggest it has been and it is _still_ considered a possible spam.

app/lib/three_scale/spam_protection/protector.rb
guicassolato: I get it now. We cannot invoke both!

guicassolato: Maybe just wrap `spam_post_request? || spam_get_request?` into a method then:

```suggestion
        captcha_required? || spam?
```

```ruby
def spam?
  spam_post_request? || spam_get_request?
end
```

app/lib/three_scale/spam_protection/protector.rb
```diff
@@ -89,6 +89,33 @@ def to_str
...
+        template.controller.request.session
+      end
+
+      def spam_post_request?
+        !http_method.get? && considered_as_spam?
+      end
+
+      def spam_get_request?
+        http_method.get? && was_considered_as_spam?
+      end
```
guicassolato, line 107: At least for me, I think these the whole thing could be simpler to read like the following:

```ruby
def captcha_needed?
  captcha_required? || possible_spam?
end

private

def possible_spam?
  if http_method.get?
    marked_as_possible_spam?
  else
    mark_possible_spam!
  end
end

def marked_as_possible_spam?
  request_session[:marked_as_possible_spam_until].to_i > Time.now.utc.to_i
end

def mark_possible_spam!
  return unless protector.spam?
  request_session[:marked_as_possible_spam_until] = (Time.now.utc + 5.minutes).to_i
end
```

app/lib/three_scale/spam_protection/protector.rb
```diff
@@ -89,6 +89,28 @@ def to_str
...
+      private
+
+      def request_session
+        return {} if template.controller.blank?
+
+        template.controller.request.session
+      end
+
+      def possible_spam?
+        http_method.get? ? marked_as_possible_spam? : mark_possible_spam
```
guicassolato, line 102: Much better than what I had suggested! Nice.

Martouta:
Awesome! 👏

PR #1813, opened 2020-04-09
[apiap] Fix routing policy's replace_path rule with Lua magic characters

guicassolato:
Routing paths that include [Lua magic characters](https://www.lua.org/pil/20.2.html) will cause apicast to fail removing the pattern from the backend request before redirect, essentially making the `remove_first` directive injected by porta in the APIAP routing policy rules a no-op in such cases. So the magic characters such as the dash sign (`-`), the plus sign (`+`) and others need to be escaped with a `%` character.

Closes https://issues.redhat.com/browse/THREESCALE-4937

### DIFF
--- a/app/lib/backend_api_logic/routing_policy.rb
+++ b/app/lib/backend_api_logic/routing_policy.rb
@@ -73,7 +73,7 @@ module BackendApiLogic
         def replace_path
           return {} if config_path.blank?
 
-          { replace_path: "{{uri | remove_first: '#{config_path.path}'}}" }
+          { replace_path: "{{uri | remove_first: '#{LuaUtility.escape(config_path.path)}'}}" }
         end
       end
       private_constant :Rule
--- a/app/lib/lua_utility.rb
+++ b/app/lib/lua_utility.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+module LuaUtility
+  MAGIC_CHARACTERS = %w@( ) . % + - * ? [ ^ $@.freeze
+  ESCAPE_CHARACTER = '%'
+
+  module_function
+
+  def escape(string)
+    string.gsub(/./) do |char|
+      MAGIC_CHARACTERS.include?(char) ? [ESCAPE_CHARACTER, char].join : char
+    end
+  end
+end
--- a/test/unit/backend_api_logic/routing_policy_test.rb
+++ b/test/unit/backend_api_logic/routing_policy_test.rb
@@ -80,7 +80,10 @@ module BackendApiLogic
           [{ private_endpoint: 'https://safe-second-api.io/v2', path: '/' }, nil],
           [{ private_endpoint: 'http://actual-api.behind.com/ns/', path: '/hey' }, "{{uri | remove_first: '/hey'}}"],
           [{ private_endpoint: 'https://safe-second-api.io', path: '/ho' }, "{{uri | remove_first: '/ho'}}"],
-          [{ private_endpoint: 'https://safe-second-api.io/v2', path: '/lets-go' }, "{{uri | remove_first: '/lets-go'}}"]
+          [{ private_endpoint: 'https://safe-second-api.io', path: '/foo/bar' }, "{{uri | remove_first: '/foo/bar'}}"],
+          [{ private_endpoint: 'https://safe-second-api.io/v2', path: '/lets-go' }, "{{uri | remove_first: '/lets%-go'}}"],
+          [{ private_endpoint: 'https://safe-second-api.io/v2', path: '/let/us+go' }, "{{uri | remove_first: '/let/us%+go'}}"],
+          [{ private_endpoint: 'https://safe-api-api.io/', path: '/path%20with%20percent' }, "{{uri | remove_first: '/path%%20with%%20percent'}}"],
         ].each do |config, replace_path_value|
           expected_replace_path = replace_path_value ? { replace_path: replace_path_value } : {}
           assert_equal expected_replace_path, rule_class.new(stub(config)).replace_path
--- a/test/unit/lua_utility_test.rb
+++ b/test/unit/lua_utility_test.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class LuaUtilityTest < ActiveSupport::TestCase
+  test '#escape' do
+    assert_equal 'mystring', LuaUtility.escape('mystring')
+    assert_equal 'my%%string', LuaUtility.escape('my%string')
+    assert_equal 'my%-string', LuaUtility.escape('my-string')
+    assert_equal 'my%+string', LuaUtility.escape('my+string')
+    assert_equal 'my%(string', LuaUtility.escape('my(string')
+    assert_equal 'my%(string%)', LuaUtility.escape('my(string)')
+    assert_equal 'my%*string', LuaUtility.escape('my*string')
+    assert_equal 'my%[string]', LuaUtility.escape('my[string]')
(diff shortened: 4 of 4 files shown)


guicassolato:
Deployed and tested on OpenShift with a config the includes:

```json
        "policy_chain": [
            {
                "name": "routing",
                "version": "builtin",
                "enabled": true,
                "configuration": {
                    "rules": [
                        {
                            "url": "https://echo-api.3scale.net:443/a-cid-service/v1/aCids",
                            "owner_id": 2,
                            "owner_type": "BackendApi",
                            "condition": {
                                "operations": [
                                    {
                                        "match": "path",
                                        "op": "matches",
                                        "value": "^(/a-cid-service/v1/aCids/.*|/a-cid-service/v1/aCids/?)"
                                    }
                                ]
                            },
                            "replace_path": "{{uri | remove_first: '/a--cid--service/v1/aCids'}}"
                        }
                    ]
                }
            },
            {
                "name": "apicast",
                "version": "builtin",
                "configuration": {}
            }
        ],
        "jwt_claim_with_client_id": null,
        "jwt_claim_with_client_id_type": null,
        "proxy_rules": [
            {
                "id": 3,
                "proxy_id": null,
                "http_method": "GET",
                "pattern": "/a-cid-service/v1/aCids/",
                "metric_id": 7,
                "metric_system_name": "hits.2",
                "delta": 1,
                "tenant_id": null,
                "created_at": "2020-04-09T13:01:48Z",
                "updated_at": "2020-04-09T13:01:48Z",
                "redirect_url": null,
                "position": 1,
                "last": false,
                "owner_id": 2,
                "owner_type": "BackendApi",
                "parameters": [],
                "querystring_parameters": {}
            }
        ]
```

Result:
```shell
$ curl -k "https://api-3scale-apicast-staging.806eda9.apps.dev-eng-ocp4-3.dev.3sca.net:443/a-cid-service/v1/aCids/?user_key=[FILTERED]"
{
  "method": "GET",
  "path": "/a-cid-service/v1/aCids/",
  "args": "user_key=[FILTERED]",
  "body": "",
  "headers": {...},
  "uuid": "f0252fb0-6d4d-4412-bab4-d4cdbb25fc92"
}
```

app/lib/backend_api_logic/routing_policy.rb
```diff
@@ -73,7 +73,13 @@ def as_json
...
           return {} if config_path.blank?
 
-          { replace_path: "{{uri | remove_first: '#{config_path.path}'}}" }
+          { replace_path: "{{uri | remove_first: '#{liquid_path}'}}" }
+        end
+
+        private
+
+        def liquid_path
+          config_path.path.gsub(/([+-])/, '\1\1')
```
thomasmaas, line 82: maybe add a comment so some time from now we know why we are doing this?

guicassolato: Done

app/lib/backend_api_logic/routing_policy.rb
```diff
@@ -73,7 +73,16 @@ def as_json
...
+          { replace_path: "{{uri | remove_first: '#{liquid_path}'}}" }
+        end
+
+        private
+
+        # The replace_path rule is a liquid template. We need to escape some special characters,
+        # otherwise they won't match and won't get removed from the backend uri
+        # The percent sign (%) is also required to be escaped, but we do not accept it in an URI path, so it's OK
+        def liquid_path
+          config_path.path.gsub(/([+-])/, '\1\1')
```
hallelujah, line 85: :warning: `%` also is needed

guicassolato: `%` is not accepted in the path.

```ruby
[1] pry(main)> URI.parse('/my/path').path
=> "/my/path"
[2] pry(main)> URI.parse('/my-path').path
=> "/my-path"
[3] pry(main)> URI.parse('/my+path').path
=> "/my+path"
[4] pry(main)> URI.parse('/my.path').path
=> "/my.path"
[5] pry(main)> URI.parse('/my_path').path
=> "/my_path"
[6] pry(main)> URI.parse('/my%path').path
URI::InvalidURIError: bad URI(is not URI?): /my%path
from /Users/guilherme.cassolato/.rbenv/versions/2.5.1/lib/ruby/2.5.0/uri/rfc3986_parser.rb:67:in `split'
```

hallelujah:
```
irb(main):001:0> URI.parse('/my%20hello').path
=> "/my%20hello"

```

guicassolato: Oops. You are right. I see it now.

hallelujah: FYI readers: https://www.w3schools.com/tags/ref_urlencode.ASP

guicassolato:
> Also Lua Magic Characters should be escaped by `%` instead of doubling them
> 
> @eloycoto @guicassolato
> 
> I think to be exhaustive you need to use % to escape all the magic characters, i.e. See https://www.lua.org/pil/20.2.html
> 
> This can be a Lua escape utility to make it generic instead of putting it in the routing policy
> 
> The magic characters are:
> 
> ```
>     ( ) . % + - * ? [ ^ $
> ```
> 
> * `(` by `%(`
> * `)` by `%)`
> * `.` by `%.`
> * `%` by `%%`
> * `+` by `%+`
> * `-` by `%-`
> * `*` by `%*`
> * `?` by `%?`
> * `[` by `%[`
> * `^` by `%^`
> * `$` by `%$`
> 
> https://issues.redhat.com/browse/THREESCALE-4937?focusedCommentId=14029139&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14029139
> 
> https://www.lua.org/pil/20.2.html

Well spotted, @hallelujah. Thank you! I'll make the changes.

app/lib/backend_api_logic/routing_policy.rb
```diff
@@ -73,7 +73,15 @@ def as_json
...
-          { replace_path: "{{uri | remove_first: '#{config_path.path}'}}" }
+          { replace_path: "{{uri | remove_first: '#{liquid_path}'}}" }
+        end
+
+        private
+
+        # The replace_path rule is a liquid template. We need to escape some special characters,
+        # otherwise they won't match and won't get removed from the backend uri
+        def liquid_path
+          config_path.path.gsub(/([%+-])/, '\1\1')
```
hallelujah, line 84: Can be a Lua utility:

```
# frozen_string_literal: true
class LuaUtility
  MAGIC_CHARACTERS = %w@( ) . % + - * ? [ ^  $ @
  ESCAPE_CHARACTER = "%"

  def self.escape(string)
    string.gsub(/./) do |char| 
      MAGIC_CHARACTERS.include?(char) ? "%#{char}" : char
  end
end
```

guicassolato: I hadn't seen this. I will rename the module. `LuaUtility` looks nice.

app/lib/liquid/lua_utility.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module Liquid
```
hallelujah, line 3: This is fine but it has nothing to do with the Liquid used in CMS ;)

guicassolato: Moved away from the `Liquid`.

guicassolato:
It turns out this was [fixed in APIcast](https://github.com/3scale/liquid-lua/pull/10) side (so it does not bend other types of integrations).

PR #1815, opened 2020-04-16
[Bugfix] application is already destroyed in destroy_backend_value of AppKey/RefFilter

Martouta:
Closes [THREESCALE-4806](https://issues.redhat.com/browse/THREESCALE-4806
)

Explained the problem and the solution in the Jira issue.

### TODO
- https://github.com/3scale/porta/pull/1815#discussion_r412789927
- Extract to smaller PRs


app/lib/archive_deletion_belonging_to_application.rb
```diff
@@ -0,0 +1,14 @@
...
+module ArchiveDeletionBelongingToApplication
+  extend ActiveSupport::Concern
+  included do
+    after_destroy :archive_as_deleted, if: :destroyed_by_association
+  end
+
+  private
+
+  def archive_as_deleted
+    ::DeletedObject.create!(object: self, owner: application, metadata: {value: value})
```
macejmic, line 12: so if `create!` would fail, it would roll back the whole transaction.. is that desired?

Martouta: Yes, IMO, it should be part of the destruction of the object, and if this fails, the destroy should fail.
Because if this fails silently, we are not cleaning it up from Apisonator.
WDYT @macejmic ?

macejmic: ok, fair enough.. but let's put it to a models/concerns.. that's the right directory for sharing logic between the models.. and maybe some other name? SaveAssociation(For)ApplicationDestroy or something like that..

app/models/deleted_object.rb
```diff
@@ -23,4 +25,9 @@ class DeletedObject < ApplicationRecord
       & (created_at <= 1.week.ago)
     end
   end
+
+  def object_instance
+    obj_attrs = {id: object_id}.merge(metadata)
+    object_type.constantize.new(obj_attrs, without_protection: true)
```
macejmic, line 31: i would call `readonly!` on this instance, just to make it sure that it won't be saved..

Martouta: Right! Good one! 👍

app/workers/backend_delete_application_worker.rb
```diff
@@ -0,0 +1,48 @@
...
+    System::ErrorReporting.report_error(exception, parameters: {event_id: event_id})
+  end
+
+  private
+
+  attr_reader :application
+
+  def build_application(event)
+    app_attrs = {id: event.application.id, user_key: event.user_key, application_id: event.application_id}
+    Cinstance.new(app_attrs, without_protection: true).tap do |app|
```
macejmic, line 29: building mocked objects just to call a method on them sounds like the right time for extracting that logic to somewhere else IMO.

Martouta: Makes sense 🤔 Those Apisonators calls inside the model look like candidates for service objects 🤔

app/models/deleted_object.rb
hallelujah: And even freeze it so internal variables won't be modified

app/events/applications/application_deleted_event.rb
```diff
@@ -4,12 +4,16 @@ class Applications::ApplicationDeletedEvent < ApplicationRelatedEvent
 
   # @param [Cinstance] application
   def self.create(application)
+    service_id = application.service_id
     new(
       application: MissingModel::MissingApplication.new(id: application.id),
+      service_id: service_id,
+      user_key: application.user_key,
```
hallelujah, line 11: Do we need the user_key?

Martouta: Unfortunately, it seems so. This is from the master branch:
https://github.com/3scale/porta/blob/be996712fbab0ca0f762a6e3676c9916204d17b6/app/lib/backend/model_extensions/cinstance.rb#L73-L76
https://github.com/3scale/porta/blob/be996712fbab0ca0f762a6e3676c9916204d17b6/app/lib/backend/model_extensions/cinstance.rb#L14

app/models/deleted_object.rb
Martouta: Good 👍 I will 👍

Martouta:
Closing to extract it into smaller PRs and make some changes there 😄

PR #1825, opened 2020-04-20
Log in with Auth tokens

didierofrivia:
Screenshot 2020-05-06 at 11 55 13

We need a way to communicate with Porta API to work on the real implementation and not leaving the page components using the OAS examples. This is not a production ready AuthN solution, just a temporary way of login and secure the calls.

fixes https://issues.redhat.com/browse/THREESCALE-4992

**Verification steps** 

1. checkout the branch
2. `cd portafly`
3. `yarn install`
4. `yarn start`
5. Open browser `http://localhost:3003` and you should hit the log in page

Note: This steps will be evolving when introducing new features.

- Since we are not sure how we are going to store the access token just yet, and not even know how we are going to get it, as a temporary measure it'll be stored in `localStorage` (not recommended because it could be target of a XSS and leaked). However, if in the future we work with JWT, it might be Ok since it can't be forged, changing a JWT payload invalidates the JWT as a whole and auth servers won't accept them.

- ~~There's a possibility to "check" the validity of the token with a call to the API, then store it or show an error message.~~ NO

- ~~Should "private" pages links be hidden in navigation? Or should the entire navigation be hidden and we should show only a login page?~~ Not logged in -> login page, Logged in -> App

**TODO**
* [x] Tests


guicassolato:
Screenshot 2020-04-23 at 11 58 02

Maybe adding a big yellow ribbon "NOT MEANT FOR PRODUCTION!!!"?

portafly/src/pages/Login.tsx
```diff
@@ -0,0 +1,89 @@
...
+import brandImg from 'assets/logo.svg'
+
+interface ILoginForm {
+  token: string
+  onChange: (v: string) => void
+  onClick: () => void
+  t: (v: string) => string
+}
+
+const LoginForm = (
```
josemigallas, line 23:
```suggestion
const LoginForm: React.FunctionComponent<ILoginForm> = (
```

portafly/src/pages/Login.tsx
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  return (
+    <LoginPage
+      brandImgSrc={brandImg}
+      brandImgAlt={t('brand_img_alt')}
+      loginTitle={t('login_title')}
+      loginSubtitle={t('login_subtitle')}
+    >
+      <LoginForm
```
josemigallas, line 77: I would rather define this as a const inside `Login` and not passing props (specially t looks weird)

portafly/src/pages/Login.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const { setAuthToken } = useAuth()
+  const { t } = useTranslation('login')
+  useDocumentTitle(t('page_title'))
+
+  function postLogin() {
+    setAuthToken(token)
+    setLoggedIn(true)
+  }
+
+  if (isLoggedIn) {
```
josemigallas, line 66: Minimal performant improvement: move this to the component's first line

portafly/src/auth/AuthContext.tsx
```diff
@@ -0,0 +1,28 @@
...
+    setAuthToken(token)
+  }
+  return (
+    <AuthContext.Provider value={{ authToken, setAuthToken: setToken }}>
+      {children}
+    </AuthContext.Provider>
+  )
+}
+
+export { AuthProvider, AuthContext }
```
josemigallas, line 28: I'm not sure about this but I think the Context itself should not be exported, if we're going to have a hook. Instead, to promote encapsulation this what I would do:
```js
// AuthContext.tsx
const AuthContext = createContext(...)
const AuthProvider = () => ...
const useAuth = () => ..

export { AuthProvider, useAuth }
```
We end up with a single file to export from and we restraint AuthContext to a single instance (ContextProvider). This is what I'm doing for the `AlertsContext` ([link](https://github.com/3scale/porta/blob/2d65f4f0357ef025d3617d715c189f996d5cef9f/portafly/src/components/util/AlertsContext.tsx#L1)).

portafly/src/pages/Login.tsx
didierofrivia: It was like that before, was trying to comply with the 25 lines per function rule, still failed.

portafly/src/pages/Login.tsx
didierofrivia: The flow it's not definitive, take this as a WIP since there are many questions that need to be answered. See description

portafly/src/auth/AuthContext.tsx
didierofrivia: Could be in the same file as it is now, yeah.

portafly/src/pages/Login.tsx
didierofrivia: You need also to bare in mind that eventually we'll have at least 2 tokens as part of the AuthZ/AuthN process

didierofrivia: Along with scopes and some claims ;)

portafly/src/pages/Login.tsx
didierofrivia: I agree that passing `t` is weird... I could instantiate the hook again inside that component. WDYT? any other solution would be weirder (or go back to a 44+ lines function)

portafly/src/pages/Login.tsx
didierofrivia: I could, however, export more specific and usable hooks when we have a established Auth flow

josemigallas: I'm just saying, from an architectural perspective, there's no need to define both a local state and a Context, since the latter already contains that state.

portafly/src/pages/Login.tsx
josemigallas: > It was like that before, was trying to comply with the 25 lines per function rule, still failed.

Am I missing something or since when is this an issue? I mean, it's defo not a problem to extract consts and functions in the same file, as long as they make sense and they cannot be extracted. But for this particular case that the _inner_ component relies on internal variables/functions I think it's better so simply put it inside. I believe it's going to be literally impossible to maintain a 25 lines limit per function.

josemigallas: > I agree that passing `t` is weird... I could instantiate the hook again inside that component. WDYT? any other solution would be weirder (or go back to a 44+ lines function)

I think instantiating the hook inside each component is the way to go, although I still don't know if that's going to be a problem performance-wise. It's what I'm doing for my PR.

portafly/src/pages/Login.tsx
didierofrivia: Oh that, yeah. There are some stuff from Patternfly login page example I need to get rid of

portafly/src/pages/Login.tsx
```diff
@@ -0,0 +1,98 @@
...
+import brandImg from 'assets/logo.svg'
+import BackgroundImageSrcLg from 'assets/images/pfbg_1200.jpg'
+import BackgroundImageSrcSm from 'assets/images/pfbg_768.jpg'
+import BackgroundImageSrcSm2x from 'assets/images/pfbg_768@2x.jpg'
+import BackgroundImageSrcXs from 'assets/images/pfbg_576.jpg'
+import BackgroundImageSrcXs2x from 'assets/images/pfbg_576@2x.jpg'
+
+interface ILoginForm {
+  token: string
+  onChange: (v: string) => void
```
damianpm, line 24: Why not `value`instead of `v`?

portafly/src/pages/Login.tsx
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  return authToken
+    ? <Redirect to="/" />
+    : (
+      <LoginPage
+        brandImgSrc={brandImg}
+        brandImgAlt={t('brand_img_alt')}
+        backgroundImgSrc={backgroundImages}
+        backgroundImgAlt="Images"
```
damianpm, line 82: "Background images" ?

portafly/src/i18n/locales/en/login.yml
```diff
@@ -0,0 +1,14 @@
+---
+page_title: Login | Portafly
+page_title_description: Log in to your account
```
damianpm, line 3:
```suggestion
page_title_description: Page title of the Login page
```
The descriptions should be useful for the translators and technical writers, giving context

portafly/src/i18n/locales/en/login.yml
```diff
@@ -0,0 +1,14 @@
...
+body_title: Login
+body_title_description: Login page
```
damianpm, line 5:
```suggestion
body_title_description: Body title of the Login page
```

portafly/src/i18n/locales/en/login.yml
```diff
@@ -0,0 +1,14 @@
...
+subtitle: This is the Login page
+subtitle_description: Subtitle of the Login page
+brand_img_alt: Red Hat logo
```
damianpm, line 8: And description is needed for all fields

portafly/src/i18n/locales/en/login.yml
didierofrivia: Hmm... I didn't put much of effort since indeed those are gonna change. BUT still, it needs to be reviewed first for others in the team that need some other kind of visual cues

portafly/src/i18n/locales/en/login.yml
didierofrivia: I think it pretty much says it all in the name of the key. no?

portafly/src/pages/Login.tsx
didierofrivia: Even better: will include it in the locales yml

portafly/src/i18n/locales/en/login.yml
damianpm: I know they are gonna change, but we should get used to add them. Besides is a compromise we made with Docs team

portafly/src/i18n/locales/en/login.yml
```diff
@@ -0,0 +1,15 @@
...
+body_title: Body title of the Login page
+body_title_description: Body title description of the Login page
+subtitle: This is the Login page
+subtitle_description: Subtitle of the Login page
+brand_img_alt: Brand image alt text for 3scale Red Hat logo
+background_img_alt: Background images alt text
+login_title: Log in to your account
+login_subtitle: Please, use your access token in order to log in
+login_submit: Submit
+login_cancel: Cancel
```
damianpm, line 13: Submit and Cancel may live in `src/i18n/locales/en/shared.yml`

portafly/src/pages/Login.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+const backgroundImages = {
+  [BackgroundImageSrc.lg]: BackgroundImageSrcLg,
+  [BackgroundImageSrc.sm]: BackgroundImageSrcSm,
+  [BackgroundImageSrc.sm2x]: BackgroundImageSrcSm2x,
+  [BackgroundImageSrc.xs]: BackgroundImageSrcXs,
+  [BackgroundImageSrc.xs2x]: BackgroundImageSrcXs2x
+}
+
+const LoginForm: React.FunctionComponent<ILoginForm> = (
```
damianpm, line 36: Maybe extract `LoginForm` to it's own file, instead of having 2 components in the same file?

didierofrivia: I don't think it's worthy, really. This, as the description says, is a temporary solution. We might get rid of the entire file, better to have everything in the same place

damianpm: OK, I get it

portafly/src/tests/auth/AuthContext.test.tsx
```diff
@@ -0,0 +1,25 @@
...
+  )
+}
+
+const SampleApp = () => <AuthProvider><SamplePage /></AuthProvider>
+
+test('should change the document title', () => {
+  const { container, getByTestId } = render(<SampleApp />)
+  const login = getByTestId('login')
+
+  expect(getNodeText(container.querySelector('p'))).toBe('Log in?')
```
josemigallas, line 21:
```suggestion
  expect(getByText('Log in?')).toBeInDocument()
```

portafly/src/tests/pages/Login.test.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react'
+
+import { render } from 'tests/custom-render'
+import { Login } from 'pages'
+
+it('should render the Login page', () => {
+  const wrapper = render(<Login />)
+  expect(wrapper).toMatchSnapshot()
```
josemigallas, line 8: wrapper
```suggestion
  expect(wrapper.container.firstChild).toMatchDiffSnapshot()
```

PR #654, opened 2019-02-28
Removes the “Latest transactions” page for good

josemigallas:
**What this PR does / why we need it**:

Removes files associated with `apiconfig/transactions` route.

**Which issue(s) this PR fixes** 

[THREESCALE-1767: Remove the “Latest transactions” page](https://issues.jboss.org/browse/THREESCALE-1767)

PR #19, opened 2018-09-19
Delete Backend Stats when Service is destroyed

Martouta:
Part of [THREESCALE-1164 - Delete objects in Backend when the Provider is completely deleted for real](https://issues.jboss.org/browse/THREESCALE-1164)
Delete stats of a Service when the Service is completely removed from System 😄

PR #1826, opened 2020-04-22
Updating pisoni 1.29

HemaHG:
Updating pisoni 1.27 to 1.29

### DIFF
--- a/Gemfile.base
+++ b/Gemfile.base
@@ -38,7 +38,7 @@ gem 'bcrypt', '~> 3.1.7'
 gem 'i18n'
 
 # Apisonator client
-gem 'pisoni', '~> 1.26'
+gem 'pisoni', '~> 1.29'
 
 # 3scale fork that allows OPTIONS passthrough
 gem 'font_assets', git: 'https://github.com/3scale/font_assets.git', ref: 'da97b8601528ee189795cc94b953ec9a30f47e83', groups: [:production, :preview]
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -476,7 +476,7 @@ GEM
       excon
       faraday (>= 0.9)
     pg (0.21.0)
-    pisoni (1.27.0)
+    pisoni (1.29.0)
       faraday (>= 0.9.1)
       injectedlogger (>= 0.0.13)
       json (>= 1.8.1)
@@ -886,7 +886,7 @@ DEPENDENCIES
   paperclip (~> 5.3.0)
   percy-capybara (~> 2.5.1)
   pg (~> 0.21.0)
-  pisoni (~> 1.26)
+  pisoni (~> 1.29)
   prawn-core!
   prawn-format (= 0.2.1)
   prawn-layout (= 0.2.1)
--- a/Gemfile.prod.lock
+++ b/Gemfile.prod.lock
@@ -477,7 +477,7 @@ GEM
       addressable
       excon
       faraday (>= 0.9)
-    pisoni (1.27.0)
+    pisoni (1.29.0)
       faraday (>= 0.9.1)
       injectedlogger (>= 0.0.13)
       json (>= 1.8.1)
@@ -887,7 +887,7 @@ DEPENDENCIES
   open_id_authentication
   paperclip (~> 5.3.0)
   percy-capybara (~> 2.5.1)
-  pisoni (~> 1.26)
+  pisoni (~> 1.29)
   prawn-core!
   prawn-format (= 0.2.1)
   prawn-layout (= 0.2.1)
--- a/app/models/metric.rb
+++ b/app/models/metric.rb
@@ -2,6 +2,7 @@
 class Metric < ApplicationRecord
   include Backend::ModelExtensions::Metric
   include SystemName
+  include SaveDestroyForServiceAssociation
   include BackendApiLogic::MetricExtension
 
   self.background_deletion = %i[pricing_rules usage_limits plan_metrics proxy_rules]
--- a/app/models/metric.rb
+++ b/app/models/metric.rb
@@ -2,7 +2,6 @@
 class Metric < ApplicationRecord
   include Backend::ModelExtensions::Metric
   include SystemName
-  include SaveDestroyForServiceAssociation
   include BackendApiLogic::MetricExtension
 
   self.background_deletion = %i[pricing_rules usage_limits plan_metrics proxy_rules]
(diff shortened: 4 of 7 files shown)


Martouta:
Also this:
https://github.com/3scale/porta/blob/be996712fbab0ca0f762a6e3676c9916204d17b6/Gemfile.base#L40-L41
Should become:
```ruby
# Apisonator client
gem 'pisoni', '~> 1.29'
```

hallelujah:
Please read https://bundler.io/ to know more about bundler

hallelujah:
Can you also run the `javascript_tests` and `mysql_build` checks please?

Martouta:
Good job so far :)

Martouta:
According to their changelog of [1.28](https://github.com/3scale/pisoni/blob/master/CHANGELOG.md#1280---2019-12-20) and [1.29](https://github.com/3scale/pisoni/blob/master/CHANGELOG.md#1290---2020-03-17), these are the changes:
1. **Removed**: Support for end-users, a feature that's no longer supported in recent Apisonator.
✅ That is what you already fixed here 😄 👏 We also removed support completely and the only reference we had left was in a test helper.
2. **Removed**: "Latest transactions" functionality.
✅ That was already fixed by Josemi from our side in [ [THREESCALE-1767](https://issues.redhat.com/browse/THREESCALE-1767) | #654 ]
3. **Changed**: `Service.delete_stats` has been adapted to the changes in Apisonator. Now it only needs to receive the service ID as a param and deletes all the stats for that service. The second param of the function, "delete_job" is ignored now.
❗ This can be cleaned up inside this PR or in another one after merging this one, but this means that we need to change some code to stop storing and sending stuff that nobody needs anymore 😄 
This is the original PR that we made for deleting stats in Apisonator through the use of the Gem Pisoni: #19
And some of that won't be needed anymore.
But this is probably too complicated and I would suggest merging this PR without doing this cleanup 🤔 but as you wish, and if you want to do it, I can explain more in-depth in a call what u need to do exactly and why.

Martouta:
This is a great 1st step to accomplish this 😄 👏 
But this also implies many other changes from what was introduced in #19

### Explanation of what was introduced back then
When we first made the deletion of stats in Apisonator, we did this:
1. When a Metric is deleted, save its ID in DeletedObject, and also save which service it belongs to (that is this line: https://github.com/3scale/porta/pull/19/files#diff-2721d36caf860d77df259a3f95dc7437R5)
2. When a Cinstance is deleted, save its ID in DeletedObject, and also save which service it belongs to (that is this line: https://github.com/3scale/porta/pull/19/files#diff-e0c7a7924a6d0dda4394a5ba0142904dR2)

Although now it has a different naming: https://github.com/3scale/porta/pull/1827

3. When a Service is destroyed, we have an event called `Services::ServiceDeletedEvent`. This event triggers other actions through a subscriber. Following a [pub/sub pattern](https://en.wikipedia.org/wiki/Publish%E2%80%93subscribe_pattern).
This is the subscriber:
https://github.com/3scale/porta/blob/34221313be7db9fba745d1cbfd2d9a2e97917cb0/app/subscribers/service_deleted_subscriber.rb#L3-L10
And this is the line that says "when this event is created, inform this subscriber":
https://github.com/3scale/porta/blob/14eadb1030fd3b2e2e61b3572396dbd6464bff62/app/lib/event_store/repository.rb#L165
Then the subscriber sends this event to delete the service in Apisonator through a background sidekiq worker:
https://github.com/3scale/porta/blob/14eadb1030fd3b2e2e61b3572396dbd6464bff62/app/subscribers/service_deleted_subscriber.rb#L6
Then, what this worker does this, in this code: https://github.com/3scale/porta/blob/14eadb1030fd3b2e2e61b3572396dbd6464bff62/app/workers/backend_delete_service_worker.rb#L10-L26
What is doing now is this:
1. Remove the stats
2. Remove the service itself AFTER the stats are destroyed
That is what is doing now, but a while ago what it was doing was:
1. Remove the stats
2. At the same time, remove the end users of this service
3. Remove the service itself AFTER the stats AND the end users are destroyed
But remember that we removed the functionality of end users 😄 
As for deleting the stats, what we were doing before is this:
https://github.com/3scale/porta/blob/14eadb1030fd3b2e2e61b3572396dbd6464bff62/app/workers/backend_delete_stats_worker.rb#L6-L24
which means:
take all the associations we previously saved in DeletedObject belong to this service, and then send the request to Apisonator in the format:
```ruby
{ deletejobdef: {
      applications: deleted_associations.contracts.pluck(:object_id),
      metrics: deleted_associations.metrics.pluck(:object_id),
      users: [],
      from: Time.parse(event.service_created_at).to_i,
      to: Time.now.utc.to_i
}}
```
that is the 2nd para of the request, it contains the end users, the applications ids, the metrics ids, the from which is the service_created_at saved in the event, and the "Time.now".

### What needs to change now IMO
I won't mention the changes in the tests for this comment, but they must change accordingly 😄 
1. We do not need the 2nd param to delete stats, so we can remove the method and replace the param for an empty hash:
https://github.com/3scale/porta/blob/34221313be7db9fba745d1cbfd2d9a2e97917cb0/app/workers/backend_delete_stats_worker.rb#L14-L24
https://github.com/3scale/porta/blob/34221313be7db9fba745d1cbfd2d9a2e97917cb0/app/workers/backend_delete_stats_worker.rb#L9
2. We do not need to save the created_at in the event anymore:
https://github.com/3scale/porta/blob/34221313be7db9fba745d1cbfd2d9a2e97917cb0/app/events/services/service_deleted_event.rb#L8
3. We do not need to keep saving the destroyed metrics belonging to a service:
https://github.com/3scale/porta/blob/34221313be7db9fba745d1cbfd2d9a2e97917cb0/app/models/metric.rb#L5
4. Although we need to keep saving the applications/cinstances belonging to a service, the reason is not the stats anymore, but deleting the application and its associations in Apisonator. Which means that although we need to keep saving the applications, the reason now is only to be able to clean up old DeletedObject that we do not need anymore, hence what needs to change is this:
https://github.com/3scale/porta/blob/34221313be7db9fba745d1cbfd2d9a2e97917cb0/app/models/deleted_object.rb#L22-L27
For this:
```ruby
scope :stale, -> do
  where.has do
    ((id.in DeletedObject.deleted_owner.select(:id)) | (object_type == Contract.name)) \
    & (created_at <= 1.week.ago)
  end
end
```
5. We do not need batches anymore for doing this, since we deleted the support for end users, and now even less:
https://github.com/3scale/porta/blob/34221313be7db9fba745d1cbfd2d9a2e97917cb0/app/workers/backend_delete_service_worker.rb#L10-L26


As I told you before, I am aware that this is complicated, and it is up to you to decide if you wanna do it. But if you want to, you can count on all the time and help that you need 😄

PR #1830, opened 2020-04-23
When a Cinstance is destroyed => We delete in Apisonator the Cinstance and its associations saved in DeletedObject

Martouta:
Extracted from #1815 as part of [THREESCALE-4806](https://issues.redhat.com/browse/THREESCALE-4806
)

Depends on #1829 (already merged 💃 )

### DIFF
--- a/app/events/applications/application_deleted_event.rb
+++ b/app/events/applications/application_deleted_event.rb
@@ -4,12 +4,16 @@ class Applications::ApplicationDeletedEvent < ApplicationRelatedEvent
 
   # @param [Cinstance] application
   def self.create(application)
+    service_id = application.service_id
     new(
       application: MissingModel::MissingApplication.new(id: application.id),
+      service_id: service_id,
+      user_key: application.user_key,
+      application_id: application.application_id,
       metadata: {
         provider_id: application.provider_account_id || application.tenant_id,
         zync: {
-          service_id: application.service_id,
+          service_id: service_id,
           proxy_id: application.service.try(:proxy).try(:id)
         }
       }
--- a/app/events/applications/application_deleted_event.rb
+++ b/app/events/applications/application_deleted_event.rb
@@ -8,7 +8,6 @@ class Applications::ApplicationDeletedEvent < ApplicationRelatedEvent
     new(
       application: MissingModel::MissingApplication.new(id: application.id),
       service_id: service_id,
-      user_key: application.user_key,
       application_id: application.application_id,
       metadata: {
         provider_id: application.provider_account_id || application.tenant_id,
--- a/app/workers/backend_delete_application_worker.rb
+++ b/app/workers/backend_delete_application_worker.rb
@@ -10,7 +10,7 @@ class BackendDeleteApplicationWorker < ApplicationJob
 
     [ApplicationKey, ReferrerFilter].each(&method(:delete_backend_associations))
 
-    delete_backend_application
+    application.delete_backend_application
   rescue ActiveRecord::RecordNotFound => exception
     System::ErrorReporting.report_error(exception, parameters: {event_id: event_id})
   end
@@ -20,7 +20,7 @@ class BackendDeleteApplicationWorker < ApplicationJob
   attr_reader :application
 
   def build_application(event)
-    app_attrs = {id: event.application.id, user_key: event.user_key, application_id: event.application_id}
+    app_attrs = {id: event.application.id, application_id: event.application_id}
     Cinstance.new(app_attrs, without_protection: true).tap do |app|
       app.service = Service.new({id: event.service_id}, without_protection: true)
     end
@@ -34,10 +34,5 @@ class BackendDeleteApplicationWorker < ApplicationJob
     DeletedObject.public_send(klass.to_s.underscore.pluralize).where(owner: application).order(:id).find_each do |deleted_object|
       build_association_object(deleted_object).destroy_backend_value
     end
-  end
-
-  def delete_backend_application
-    application.delete_backend_user_key_to_application_id_mapping
-    application.delete_backend_application
   end
 end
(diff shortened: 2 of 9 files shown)


app/events/applications/application_deleted_event.rb
```diff
@@ -4,12 +4,16 @@ class Applications::ApplicationDeletedEvent < ApplicationRelatedEvent
 
   # @param [Cinstance] application
   def self.create(application)
+    service_id = application.service_id
     new(
       application: MissingModel::MissingApplication.new(id: application.id),
+      service_id: service_id,
+      user_key: application.user_key,
```
Martouta, line 11: https://github.com/3scale/porta/pull/1815#discussion_r412804008

image

app/workers/backend_delete_application_worker.rb
```diff
@@ -0,0 +1,48 @@
...
+    System::ErrorReporting.report_error(exception, parameters: {event_id: event_id})
+  end
+
+  private
+
+  attr_reader :application
+
+  def build_application(event)
+    app_attrs = {id: event.application.id, user_key: event.user_key, application_id: event.application_id}
+    Cinstance.new(app_attrs, without_protection: true).tap do |app|
```
Martouta, line 29: I will refactor in the following PR to use service objects instead of building the models to call to apisonator through them.

app/workers/backend_delete_application_worker.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+class BackendDeleteApplicationWorker < ApplicationJob
+
+  queue_as :low
+
+  def self.enqueue(event)
```
Martouta, line 7: I guess I can remove this method :)

app/events/applications/application_deleted_event.rb
Martouta: Waaaait, i think that we may not need this anymore 🤔 but i will check more in-depth

app/workers/backend_delete_application_worker.rb
```diff
@@ -0,0 +1,48 @@
...
+      app.service = Service.new({id: event.service_id}, without_protection: true)
+    end
+  end
+
+  def build_association_object(deleted_object)
+    deleted_object.object_instance { |object| object.application = application }
+  end
+
+  def delete_backend_associations(klass)
+    DeletedObject.public_send(klass.to_s.underscore.pluralize).order(:id).find_each do |deleted_object|
```
Martouta, line 39: where owner is this app!

app/events/applications/application_deleted_event.rb
Martouta: RIght, we do not need it. The application is removed using the ID, hence it does not need any other extra removal.

app/workers/backend_delete_application_worker.rb
```diff
@@ -0,0 +1,38 @@
...
+      app.service = Service.new({id: event.service_id}, without_protection: true)
+    end
+  end
+
+  def build_association_object(deleted_object)
+    deleted_object.object_instance { |object| object.application = application }
+  end
+
+  def delete_backend_associations(klass)
+    DeletedObject.public_send(klass.to_s.underscore.pluralize).where(owner: application).order(:id).find_each do |deleted_object|
```
guicassolato, line 34: This works now because we know it's a `has_many`, not a `has_one`. Do you think we it's worth extracting `klass.to_s.underscore.pluralize` to a separate method? Ideally, we would resolve the association name from the name of the class without assuming much, but I think it may already help a lot whoever reads the code realise the importance of this and, in case of a `has_one` association of `Cinstance` added later to line 11, where to fix it. Plus, you could alternatively map the list of classes into association names (using this new method) and be more assertive with `#delete_backend_associations` by passing the name of the association instead of the name of the class.

app/workers/backend_delete_application_worker.rb
```diff
@@ -0,0 +1,38 @@
...
+  rescue ActiveRecord::RecordNotFound => exception
+    System::ErrorReporting.report_error(exception, parameters: {event_id: event_id})
+  end
+
+  private
+
+  attr_reader :application
+
+  def build_application(event)
+    app_attrs = {id: event.application.id, application_id: event.application_id}
```
guicassolato, line 23: I guess this is failure proof because `application` is a `MissingModel::MissingApplication`, right? Not an actual `Cinstance` and therefore there's no risk of the application having been deleted already, correct?

app/workers/backend_delete_application_worker.rb
Martouta: @guicassolato this is removed in the following PR 😉

app/workers/backend_delete_application_worker.rb
Martouta: Exactly 😄

app/workers/backend_delete_application_worker.rb
Martouta: I mean #1832

app/workers/backend_delete_application_worker.rb
```diff
@@ -0,0 +1,38 @@
...
+    Cinstance.new(app_attrs, without_protection: true).tap do |app|
+      app.service = Service.new({id: event.service_id}, without_protection: true)
```
guicassolato, line 25: This is a bit fragile, IMO. I'm sure it works now, but I confess to be worried with the layer between the service's id and `Service#backend_id`. As you know, the deletion in apisonator actually happens using `Service#backend_id`, which now reduces to `service.id.to_s`. This cannot change or it might be hard to spot the bug it would cause here.

https://github.com/3scale/porta/blob/34221313be7db9fba745d1cbfd2d9a2e97917cb0/app/models/application_key.rb#L124

https://github.com/3scale/porta/blob/34221313be7db9fba745d1cbfd2d9a2e97917cb0/app/lib/backend/model_extensions/service.rb#L12

https://github.com/3scale/porta/blob/34221313be7db9fba745d1cbfd2d9a2e97917cb0/app/models/service.rb#L272

Martouta: @guicassolato This is not meant to stay here. See this:
https://github.com/3scale/porta/pull/1830#discussion_r414215489

PR #1832, opened 2020-04-23
[Refactor] Service objects for destroy_backend_value in associations of Application

Martouta:
I'm not sure if the way I've done it is a good idea. I definitely need some reviews with suggestions 😄 

This PR intends to improve the code from the reviews I've received in the other PRs of [THREESCALE-4806](https://issues.redhat.com/browse/THREESCALE-4806)

In particular, this is what it does:
1. Add a test to ensure that the ref filters are also removed only belonging to that Application
2. Remove unused method 'delete_backend_cinstance'
3. Use Service Objects instead of rebuilding the model only to call to Pisoni
4. Use service.backend_id instead of service.id.to_s

### DIFF
--- a/app/services/application_association_backend_service.rb
+++ b/app/services/application_association_backend_service.rb
@@ -2,7 +2,7 @@
 
 module ApplicationAssociationBackendService
   def delete_all(application_id:, service_backend_id:, application_backend_id:)
-    DeletedObject.public_send(reflection_name).where(owner_type: Contract, owner_id: application_id).order(:id).find_each do |deleted_object|
+    DeletedObject.public_send(reflection_name).where(owner_type: Contract.name, owner_id: application_id).order(id: :asc).find_each do |deleted_object|
       delete(service_backend_id: service_backend_id, application_backend_id: application_backend_id, value: deleted_object.metadata[:value])
     end
   end
--- a/app/services/application_association_backend_service.rb
+++ b/app/services/application_association_backend_service.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+module ApplicationAssociationBackendService
+  def delete_all(application_id:, service_backend_id:, application_backend_id:)
+    DeletedObject.public_send(reflection_name).where(owner_type: Contract, owner_id: application_id).order(:id).find_each do |deleted_object|
+      delete(service_backend_id: service_backend_id, application_backend_id: application_backend_id, value: deleted_object.metadata[:value])
+    end
+  end
+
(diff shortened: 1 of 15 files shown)


app/services/application_association_backend_service.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module ApplicationAssociationBackendService
+  def delete_all(application_id:, service_backend_id:, application_backend_id:)
+    DeletedObject.public_send(reflection_name).where(owner_type: Contract, owner_id: application_id).order(:id).find_each do |deleted_object|
```
macejmic, line 5: what's the default for `order(:id)` i would rather to see in here asc/desc.

if ower_type could be a string, `Contract` would be better.. using classes could end up with random errors.

app/services/application_key_backend_service.rb
```diff
@@ -0,0 +1,14 @@
...
+module ApplicationKeyBackendService
+  module_function
+  extend ApplicationAssociationBackendService
+
+  def reflection_name
+    :application_keys
+  end
+
+  def pisoni_class
+    ThreeScale::Core::ApplicationKey
```
macejmic, line 12: also in here, better to define the class in a string way and then constantine the string to a class.

it could end up failing on a pre-loading class error etc. this should be a standard so let's ask @guicassolato and @hallelujah too.

app/services/application_association_backend_service.rb
Martouta: the default is :asc . But I do not really care about that as long as it keeps an order... but in postgres it would be 'random' otherwise.

> if ower_type could be a string, Contract would be better.. using classes could end up with random errors.

You are totally right! Will do! In fact, it is deprecated doing it this way and in future versions of Ruby (I think already in 5.1), it will always fail for sure! 😱 good catch!

app/services/application_key_backend_service.rb
Martouta: > it could end up failing on a pre-loading class error etc

In this case, isn't it a good thing? because we would notice right away as soon as we make any change in Pisoni (like upgrading it) that it is not working 🤔

Martouta: Although maybe you are right and it is better to make it fail only at that moment 🤔 I dunno 🤔

macejmic: ok, leave it as it is :) we can discuss it later.

app/services/application_association_backend_service.rb
Martouta: Done 😎

PR #1836, opened 2020-04-28
[2.5] Do not recreate OIDCConf object from UI IntegrationsController

Martouta:
2.5 part of [THREESCALE-5103](https://issues.redhat.com/browse/THREESCALE-5103)


Martouta:
It seems that we won't patch for 2.5
and that branch that I just created is not the real 2.5.1 anyway but something custom instead :)

PR #1837, opened 2020-04-28
Adds a Provider to handle Toast Alerts in all Portafly

josemigallas:

Introduces a Provider that handles toast-like Alerts. By means of a `useAlertsContext` hook, it is possible to pop an alert up from anywhere in the app, using the same AlertGroup. Each Alert has a lifespan of 8 seconds and can be closed manually.

TODO?: theoretically it should be possible to maintain an alert alive as long as the mouse is on top of it.

**Verification steps** 

A test button has been added to `http://localhost:3000/applications` in order to verify/demo this functionality.

### DIFF
--- a/portafly/src/components/util/AlertsContext.tsx
+++ b/portafly/src/components/util/AlertsContext.tsx
@@ -1,0 +1,68 @@
+import React, { useContext } from 'react'
+import {
+  AlertGroup,
+  AlertActionCloseButton,
+  Alert,
+  AlertVariant
+} from '@patternfly/react-core'
+import { useTranslation } from 'i18n/useTranslation'
+
+interface IAlert {
+  id: string
+  variant: keyof typeof AlertVariant
+  title: string
+}
+
+type IAlertsContext = { addAlert: (alert: IAlert) => void }
+const AlertsContext = React.createContext<IAlertsContext>({ addAlert: () => {} })
+
+const TIMEOUT = 8000
+
+const AlertsProvider: React.FunctionComponent = ({ children }) => {
+  const { t } = useTranslation('shared')
+
+  type AlertsState = Array<IAlert & { timeout: NodeJS.Timeout }>
+  const [alerts, setAlerts] = React.useState<AlertsState>([])
+
+  const removeAlert = (id: string) => (
+    setAlerts((prevAlerts) => prevAlerts.filter((a) => a.id !== id))
+  )
+
+  const addAlert = (alert: IAlert) => {
+    const timeout = setTimeout(() => removeAlert(alert.id), TIMEOUT)
+    setAlerts((prevAlerts) => [...prevAlerts, { ...alert, timeout }])
+  }
+
+  const CloseButton = ({ id, timeout }: { id: string, timeout: NodeJS.Timeout}) => (
+    <AlertActionCloseButton
+      title={t('alerts.close_button')}
+      onClose={() => {
+        removeAlert(id)
+        clearTimeout(timeout)
+      }}
+    />
+  )
+
+  return (
+    <AlertsContext.Provider value={{ addAlert }}>
+      <AlertGroup isToast>
+        {alerts.map(({
+          id, variant, title, timeout
+        }) => (
+          <Alert
+            key={id}
+            isLiveRegion
+            variant={variant}
+            title={title}
+            action={<CloseButton id={id} timeout={timeout} />}
+          />
+        ))}
+      </AlertGroup>
+      {children}
+    </AlertsContext.Provider>
+  )
+}
+
+const useAlertsContext = () => useContext(AlertsContext)
+
+export { AlertsProvider, useAlertsContext }
--- a/portafly/src/tests/components/util/AlertContext.test.tsx
+++ b/portafly/src/tests/components/util/AlertContext.test.tsx
@@ -1,0 +1,22 @@
+import React from 'react'
+import { AlertsProvider, useAlertsContext } from 'components'
+import { render } from 'tests/custom-render'
+import { waitFor } from '@testing-library/dom'
+
+const AlertComponent = () => {
+  const { addAlert } = useAlertsContext()
+  React.useEffect(() => addAlert({ id: 'id', title: 'title', variant: 'danger' }), [])
+  return <></>
+}
+
+it('shows an alert that disappears after 8 seconds', async () => {
+  jest.setTimeout(8001)
+  const { getByText, queryByText } = render(
+    <AlertsProvider>
+      <AlertComponent />
+    </AlertsProvider>
+  )
+
+  await waitFor(() => expect(getByText('title')).toBeInTheDocument(), { timeout: 7999 })
+  waitFor(() => expect(queryByText('title')).not.toBeInTheDocument(), { timeout: 8001 })
+})
--- a/portafly/src/App.tsx
+++ b/portafly/src/App.tsx
@@ -5,7 +5,8 @@ import { AuthProvider } from 'auth'
 import {
   SwitchWith404,
   LazyRoute,
-  Root
+  Root,
+  AlertsProvider
 } from 'components'
 import { LastLocationProvider } from 'react-router-last-location'
 
@@ -26,9 +27,11 @@ const App = () => (
   <AuthProvider>
     <Router>
       <LastLocationProvider>
-        <Root>
-          <PagesSwitch />
-        </Root>
+        <AlertsProvider>
+          <Root>
+            <PagesSwitch />
+          </Root>
+        </AlertsProvider>
       </LastLocationProvider>
     </Router>
   </AuthProvider>
--- a/portafly/src/components/util/index.tsx
+++ b/portafly/src/components/util/index.tsx
@@ -1,3 +1,4 @@
 export * from './useA11yRoute'
 export * from './useBreadcrumb'
 export * from './useDocumentTitle'
+export * from './AlertsContext'
--- a/portafly/src/i18n/locales/en/shared.yml
+++ b/portafly/src/i18n/locales/en/shared.yml
@@ -33,4 +33,6 @@ format:
   lowercase: "{{text, lowercase}}"
   lowercase_description: Formats a text as lowercase
   format_description: Utilities to format text
-
+alerts:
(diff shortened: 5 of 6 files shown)


portafly/src/components/util/AlertsContext.tsx
```diff
@@ -0,0 +1,55 @@
...
+const AlertsProvider: React.FunctionComponent = ({ children }) => {
+  const [alerts, setAlerts] = React.useState<IAlert[]>([])
+
+  const removeAlert = (key: string) => {
+    setAlerts((prevAlerts) => prevAlerts.filter((a) => a.key !== key))
+  }
+
+  const addAlert = (alert: IAlert) => {
+    setAlerts((prevAlerts) => [...prevAlerts, alert])
+    setTimeout(() => removeAlert(alert.key), 8000)
```
damianpm, line 26: 5000 ?

portafly/src/components/util/AlertsContext.tsx
```diff
@@ -0,0 +1,55 @@
...
+  AlertVariant
+} from '@patternfly/react-core'
+
+interface IAlert {
+  key: string
+  variant: keyof typeof AlertVariant
+  title: string
+}
+
+const AlertsContext = React.createContext<({ addAlert?: (alert: IAlert) => void })>({})
```
damianpm, line 15: `addAlert` is optional, that's causing the error in
https://github.com/3scale/porta/pull/1837/files#diff-7be801d10bf78bb72b655d866926c909R75

portafly/src/components/util/AlertsContext.tsx
josemigallas: According to @cntlsn / UX team, 8 seconds is the way to go.

portafly/src/components/util/AlertsContext.tsx
josemigallas: I see.. I'll fix it

portafly/src/components/util/AlertsContext.tsx
```diff
@@ -0,0 +1,55 @@
+import React, { useContext } from 'react'
```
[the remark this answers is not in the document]
josemigallas: Weird!! I didn't notice.. at least it was working for me. I'll fix it.

portafly/src/components/util/AlertsContext.tsx
```diff
@@ -0,0 +1,55 @@
...
+  return (
+    <AlertsContext.Provider value={{ addAlert }}>
+      <AlertGroup isToast>
+        {alerts.map(({ key, variant, title }) => (
+          <Alert
+            key={key}
+            isLiveRegion
+            variant={variant}
+            title={title}
+            action={(
```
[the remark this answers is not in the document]
josemigallas: I could make it more compact, like this:
```jsx
  const CloseButton = ({ key, timeout }: Pick<IAlert, 'key' | 'timeout'>) => (
    <AlertActionCloseButton
      title="button title"
      variantLabel="variant alert"
      onClose={() => {
        removeAlert(key)
        clearTimeout(timeout)
      }}
    />
  )

  return (
    <AlertsContext.Provider value={{ addAlert }}>
      <AlertGroup isToast>
        {alerts.map(({
          key, variant, title, timeout
        }) => (
          <Alert
            key={key}
            isLiveRegion
            variant={variant}
            title={title}
            action={<CloseButton key={key} timeout={timeout} />}
          />
        ))}
      </AlertGroup>
      {children}
    </AlertsContext.Provider>
  )
```
But it kinda looks the same. This is also possible but seems like too much repetition IMO:
```jsx
const AutocloseAlert = ({
    key,
    variant,
    title,
    timeout
  }: IAlert) => (
    <Alert
      key={key}
      isLiveRegion
      variant={variant}
      title={title}
      action={(
        <AlertActionCloseButton
          title="button title"
          variantLabel="variant alert"
          onClose={() => {
            removeAlert(key)
            clearTimeout(timeout)
          }}
        />
      )}
    />
  )

  return (
    <AlertsContext.Provider value={{ addAlert }}>
      <AlertGroup isToast>
        {alerts.map((alert) => (
          <AutocloseAlert
            key={alert.key}
            title={alert.title}
            variant={alert.variant}
            timeout={alert.timeout}
          />
        ))}
      </AlertGroup>
      {children}
    </AlertsContext.Provider>
  )
```

portafly/src/components/Root.tsx
```diff
@@ -53,13 +61,35 @@ const Root: React.FunctionComponent = ({ children }) => {
     }),
     [history]
   )
+
+  const pageToolbar = (
```
didierofrivia, line 65: Can we move this example to Overview page (or an "example" page)... Or are you planning to remove this before merge?

portafly/src/components/util/AlertsContext.tsx
```diff
@@ -0,0 +1,64 @@
...
+} from '@patternfly/react-core'
+
+interface IAlert {
+  id: string
+  variant: keyof typeof AlertVariant
+  title: string
+  timeout: NodeJS.Timeout
+}
+
+type IAlertsContext = { addAlert: (alert: Omit<IAlert, 'timeout'>) => void }
```
didierofrivia, line 16: I feel like when using `Omit` to remove some properties, and it's not for defining a new type or interface, I'm missing the "why" or "what"... specially when it's used more than once. Maybe declaring an `Alert` type, and extend it in the interface? Utility types are a double edge sword sometimes

portafly/src/components/Root.tsx
josemigallas: I'll create an example page

portafly/src/components/util/AlertsContext.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const [alerts, setAlerts] = React.useState<IAlert[]>([])
+
+  const removeAlert = (id: string) => {
+    setAlerts((prevAlerts) => prevAlerts.filter((a) => a.id !== id))
+  }
+
+  const addAlert = (alert: Omit<IAlert, 'timeout'>) => {
+    const timeout = setTimeout(() => removeAlert(alert.id), 8000)
+    setAlerts((prevAlerts) => [...prevAlerts, { ...alert, timeout }])
+  }
```
didierofrivia, line 29: Not super important, but for the sake of composition, something like...

```suggestion
const TIMEOUT = 8000
...
const alertTimeout = (id: string) => setTimeout(() => removeAlert(id), TIMEOUT)

const addAlert (alert: Omit<IAlert, 'timeout'>) => (
  setAlerts((prevAlerts) => [...prevAlerts, { ...alert, alertTimeout(alert.id) }])
)
```

portafly/src/components/util/AlertsContext.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  const CloseButton = ({ id, timeout }: Pick<IAlert, 'id' | 'timeout'>) => (
+    <AlertActionCloseButton
+      title="button title"
+      variantLabel="variant alert"
```
didierofrivia, line 34: These will be located eventually, right?

portafly/src/tests/components/util/AlertContext.test.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+it('shows an alert that disappears after 8 seconds', async () => {
+  jest.setTimeout(8001)
+  const { getByText, queryByText } = render(
+    <AlertsProvider>
+      <AlertComponent />
+    </AlertsProvider>
+  )
+
+  await waitFor(() => expect(getByText('title')).toBeInTheDocument(), { timeout: 7999 })
```
didierofrivia, line 20: This means the test will be idle for 8 seconds, right? If so, maybe we should look for an alternative with time travel / time freezing / etc

portafly/src/components/util/AlertsContext.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+type IAlertsContext = { addAlert: (alert: Omit<IAlert, 'timeout'>) => void }
+const AlertsContext = React.createContext<IAlertsContext>({ addAlert: () => {} })
+
+const AlertsProvider: React.FunctionComponent = ({ children }) => {
+  const [alerts, setAlerts] = React.useState<IAlert[]>([])
+
+  const removeAlert = (id: string) => {
+    setAlerts((prevAlerts) => prevAlerts.filter((a) => a.id !== id))
+  }
```
didierofrivia, line 24: 💄 
```suggestion
  const removeAlert = (id: string) => (
    setAlerts((prevAlerts) => prevAlerts.filter((a) => a.id !== id))
  )
```

portafly/src/components/util/AlertsContext.tsx
josemigallas: Well the why is because it doesn't make sense to define 2 types for Alerts. `timeout` is only relevant for the Provider so as you mention the `Alert` shouldn't have this. But I didn't do it at first because it turned out to be very verbose:
```ts
React.useState<(IAlert & { timeout: NodeJS.Timeout})[]>([])
```

portafly/src/tests/components/util/AlertContext.test.tsx
josemigallas: Well I don't know exactly why but it doesn't wait for it. The test takes about 50ms. The reason why I put await here is for jest to not complain about having multiple async calls.

When putting a second `await` (in the next line), then the test does take 8 seconds.

portafly/src/components/util/AlertsContext.tsx
josemigallas: If you mean localized 😁, yes I missed it

portafly/src/components/util/AlertsContext.tsx
josemigallas: I extracted `TIMEOUT` as a constant but I don't see the value of having a function for `alertTimeout`. It belongs only to `addAlert` IMO.

portafly/src/tests/components/util/AlertContext.test.tsx
didierofrivia: We could address this in the future with a library like timecop in ruby

PR #1838, opened 2020-04-28
[2.8] Do not recreate OIDCConf object from UI IntegrationsController

Martouta:
2.8 patch for [THREESCALE-5103](https://issues.redhat.com/browse/THREESCALE-5103)

### Verification steps
Follow these steps to test it locally.
1. Add this to `config/rolling_updates.yml` (I think that only the 2nd one is necessary for this, but just in case):
```
apicast_v2: true
apicast_oidc: true
```
2. Make webpack work properly for development `bundle exec rake webpack:dev`
3. Run the rails server `UNICORN_WORKERS=8 bundle exec rails server -b 0.0.0.0`
4. You need to ensure that your service has a service token. If you are not sure of that, open sidekiq with the default queue `bundle exec sidekiq -q default -L log/sidekiq.log` and run `bundle exec rake fixes:create_missing_service_tokens`. Then you won't need sidekiq anymore for this so feel free to close it after this 😄 
5. Go to `http://provider-admin.example.com.lvh.me:3000/apiconfig/services/2/integration`,  `Edit authentication settings` and select the _Authentication_ `OpenID Connect`
image
image
6. In `http://provider-admin.example.com.lvh.me:3000/apiconfig/services/2/integration`, `add the base URL of your API and save the configuration`, scroll down to `Authentication Settings`, change the OIDC config, and click to `Update`.
image
image
7. Go to the rails console with `bundle exec rails console` and check the OIDCConfig with `Proxy.find(2).oidc_configuration`. Ensure that it has the configuration that you saved and that it contains an ID.
8. Repeat _step 6_ with a different configuration. Then repeat _step 7_ and see that the new configuration is saved with the data you sent and that the ID is the same one as before.
As a bonus, do `OIDCConfiguration.all` and see that there is only one.

It is recommended to do these steps with this branch and also in a branch without these changes so you can see the difference.

### DIFF
--- a/app/controllers/api/integrations_controller.rb
+++ b/app/controllers/api/integrations_controller.rb
@@ -240,7 +240,7 @@ class Api::IntegrationsController < Api::BaseController
       :error_headers_auth_missing, :error_auth_missing, :error_status_no_match,
       :error_headers_no_match, :error_no_match, :error_status_limits_exceeded, :error_headers_limits_exceeded, :error_limits_exceeded,
       :api_test_path, :policies_config, proxy_rules_attributes: %i[_destroy id http_method pattern delta metric_id
-                                                                   redirect_url position last], oidc_configuration_attributes: OIDCConfiguration::Config::ATTRIBUTES
+                                                                   redirect_url position last], oidc_configuration_attributes: OIDCConfiguration::Config::ATTRIBUTES + [:id]
     ]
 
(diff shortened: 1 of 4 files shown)


guicassolato:
@Martouta , does it also affect https://github.com/3scale/porta/blob/62f27e830b521413b56c984f0f586aec542b3093/app/controllers/api/services_controller.rb#L104 maybe?

Martouta:
U were right, @guicassolato . Good catch! 👏

PR #1840, opened 2020-04-29
[Bugfix] LineItem does not raise an error if the cost cannot be converted to BigDecimal

Martouta:
Closes [THREESCALE-4953](https://issues.redhat.com/browse/THREESCALE-4953)

`BigDecimal('something-wrong')` in Ruby 2.3 returns `BigDecimal(0)`
`BigDecimal('something-wrong')` in Ruby 2.4 raises `ArgumentError`

We are using Ruby 2.4 in SaaS since a couple of months ago approx.
This PR makes the easy/quick fix, which is making the code behave like few months ago, since it wasn't failing back then.
A better solution would be returning an invalid error instead, but that is not trivial in our case.

### DIFF
--- a/app/models/line_item.rb
+++ b/app/models/line_item.rb
@@ -44,6 +44,8 @@ class LineItem < ApplicationRecord
 
   def cost=(value)
     self[:cost] = BigDecimal((value || 0).to_s).round(DECIMALS)
+  rescue ArgumentError
+    self[:cost] = 0
   end
 
   def custom?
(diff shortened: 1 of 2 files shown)


app/models/line_item.rb
```diff
@@ -44,6 +44,8 @@ def type
 
   def cost=(value)
     self[:cost] = BigDecimal((value || 0).to_s).round(DECIMALS)
+  rescue ArgumentError
```
macejmic, line 47: so $5.50 is a problem (for example) ?

Martouta: yes @macejmic because there is another attribute called `currency` that is the one for the dolar. `cost` is only the number/value.

PR #1841, opened 2020-04-29
[Bugfix][Task] Create default metrics for the backend apis that do not have it

Martouta:
Closes [THREESCALE-5137](https://issues.redhat.com/browse/THREESCALE-5137)

Problem explained in https://issues.redhat.com/browse/THREESCALE-5137?focusedCommentId=14066849&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-14066849

The reason is that this backend api does not have metrics. Since [THREESCALE-3355](https://issues.redhat.com/browse/THREESCALE-3355), we create this metric for all new backend apis. But [this was made the 9th of September](https://github.com/3scale/porta/commit/6be289ac6ce93b1a83aa99fd6d4bc1f3793b9167), but this backend api was created before that... so we should create it for all the older backend apis.

### DIFF
--- a/test/unit/tasks/backend_api_test.rb
+++ b/test/unit/tasks/backend_api_test.rb
@@ -2,26 +2,28 @@ require 'test_helper'
 
 module Tasks
   class BackendApiTest < ActiveSupport::TestCase
-    setup do
-      @backend_api = FactoryBot.create(:backend_api)
-      FactoryBot.create(:backend_api_config, backend_api: @backend_api)
-      @orphan_backend_api = FactoryBot.create(:backend_api)
-    end
+    class DestroyOrphans < ActiveSupport::TestCase
+      setup do
+        @backend_api = FactoryBot.create(:backend_api)
+        FactoryBot.create(:backend_api_config, backend_api: @backend_api)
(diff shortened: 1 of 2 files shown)


test/unit/tasks/backend_api_test.rb
```diff
@@ -2,26 +2,41 @@
...
+    test 'create_default_metrics' do
+      backend_apis_without_metrics = FactoryBot.create_list(:backend_api, 2)
+      backend_apis_without_metrics.each { |backend_api| backend_api.metrics.delete_all }
+      _backend_apis_with_metrics = FactoryBot.create_list(:backend_api, 2)
+
+      execute_rake_task 'backend_api.rake', 'backend_api:create_default_metrics'
 
-      execute_rake_task 'backend_api.rake', 'backend_api:destroy_orphans'
+      BackendApi.all.each do |backend_api|
+        assert_equal 1, backend_api.metrics.count
```
guicassolato, line 37: Other tests can leak and raise false positives here. Better running only through the backend apis created in this test.

Martouta: Right, makes sense 👍

PR #1843, opened 2020-04-30
Remove requests about an Application to Apisonator through its user_key as the identifier param

Martouta:
Closes [THREESCALE-5151](https://issues.redhat.com/browse/THREESCALE-5151)

Please review each person from @3scale/system-ruby **very carefully**.

### Motivation
I encountered last week a bunch of "repeated" call to Apisonator/Pisoni. We were saving/deleting/updating Cinstances both identifying them by its `ID` and identifying them by its `user_key`. It turns our that this is not only unnecessary, but also those endpoints identifying them through `user_key` have been deprecated for many years, and they only introduced them again 6 years ago with the message "this is deprecated" because in System/Porta side we had not deleted that code yet and it was raising errors in multitenant for them deleting it before us.

### Solution
Let's remove all calls identifying the application by `user_key` and let's keep the already existing calls identifying them by ID. Thus, we only use ID as the identifier, but we keep updating the user_key in the Service when it changes.
The calls we already use and will keep using instead of the old ones are the following.
https://github.com/3scale/porta/blob/356bc620ffaaf0557006ec6347a7aa413f930f84/app/lib/backend/model_extensions/cinstance.rb#L44-L51
https://github.com/3scale/porta/blob/356bc620ffaaf0557006ec6347a7aa413f930f84/app/lib/backend/model_extensions/cinstance.rb#L28-L42
https://github.com/3scale/porta/blob/356bc620ffaaf0557006ec6347a7aa413f930f84/app/lib/backend/model_extensions/cinstance.rb#L87-L97

### DIFF
--- a/app/lib/backend/model_extensions/cinstance.rb
+++ b/app/lib/backend/model_extensions/cinstance.rb
@@ -8,9 +8,7 @@ module Backend
           before_destroy :preload_used_associations
           before_validation :set_application_id, :on => :create
 
-          after_commit :update_provider_backend_service_if_user_key_changed, :on => :update
-
-          after_commit :update_backend_user_key_to_application_id_mapping, unless: :destroyed?
+          after_commit :update_provider_backend_service_if_user_key_changed, on: :update
 
           after_commit :update_backend_application, unless: :destroyed?
 
@@ -21,7 +19,6 @@ module Backend
       def delete_backend_cinstance
         application_keys.each(&:destroy_backend_value)
         referrer_filters.each(&:destroy_backend_value)
-        delete_backend_user_key_to_application_id_mapping
         delete_backend_application
       end
 
@@ -47,28 +44,6 @@ module Backend
         else
           Rails.logger.warn("Cinstance id: #{id}, application_id: #{application_id} cannot be deleted from backend")
         end
-        true
-      end
-
-      def update_backend_user_key_to_application_id_mapping
-        user_key_was, current_user_key = previous_changes[:user_key]
-
-        if previously_changed?(:user_key) && service.id.present? && user_key_was.present?
-          ThreeScale::Core::Application.delete_id_by_key(service.backend_id, user_key_was)
-        end
-
-        ## save no matter what, even if not changed, it's safe. Required for backend rake task that are
-        ## unaware of previous changes
-        if !service.nil? && user_key.present?
-          ThreeScale::Core::Application.save_id_by_key(service.backend_id, user_key, application_id)
-        end
-
-        true
-      end
-
-      def delete_backend_user_key_to_application_id_mapping
-        return true if !service || !service.id || user_key.blank?
-        ThreeScale::Core::Application.delete_id_by_key(service.backend_id, user_key)
         true
       end
 
--- a/app/lib/backend/model_extensions/cinstance.rb
+++ b/app/lib/backend/model_extensions/cinstance.rb
@@ -32,6 +32,7 @@ module Backend
                                               :state      => state,
                                               :plan_id    => plan.id,
                                               :plan_name  => plan.name,
+                                              :user_key   => user_key,
                                               :redirect_url => redirect_url )
         end
 
--- a/app/lib/backend/storage_rewrite.rb
+++ b/app/lib/backend/storage_rewrite.rb
@@ -25,7 +25,7 @@ module Backend
 
       Cinstance.find_each do |cinstance|
         cinstance.send(:update_backend_application)
-        cinstance.send(:update_backend_user_key_to_application_id_mapping)
+        cinstance.send(:update_provider_backend_service_if_user_key_changed)
 
         # to ensure that there is a 'backend_object' - there are
         # invalid data floating around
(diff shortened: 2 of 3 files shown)


app/lib/backend/model_extensions/cinstance.rb
```diff
@@ -83,18 +56,6 @@ def preload_used_associations
       def set_application_id
         self.application_id ||= SecureRandom.hex(4)
       end
-
-      def update_provider_backend_service_if_user_key_changed
-        if previously_changed?(:user_key)
-          user_key_was, user_key = previous_changes[:user_key]
-
-          if user_account && user_account.provider? && user_key_was.present? && user_account.services.present?
-            ThreeScale::Core::Service.change_provider_key!(user_key_was, user_key)
```
hallelujah, line 92: It looks to me this is legit call,

In pisoni master it is not marked as deprecate https://github.com/3scale/pisoni/blob/9904bce5dfb482a49a49fb5684f6262af7591ada/lib/3scale/core/service.rb#L26-L38

Or should it? Ref: https://issues.redhat.com/browse/THREESCALE-5152

Martouta: Makes sense 👍

app/lib/backend/model_extensions/cinstance.rb
```diff
@@ -50,28 +45,6 @@ def delete_backend_application
...
-        if !service.nil? && user_key.present?
-          ThreeScale::Core::Application.save_id_by_key(service.backend_id, user_key, application_id)
-        end
-
-        true
-      end
-
-      def delete_backend_user_key_to_application_id_mapping
-        return true if !service || !service.id || user_key.blank?
-        ThreeScale::Core::Application.delete_id_by_key(service.backend_id, user_key)
```
hallelujah, line 71: It looks like we do not pass the `user_key` in

https://github.com/3scale/porta/blob/70cd39602a28c9d451b1aaaa3c4656145cb7466e/app/lib/backend/model_extensions/cinstance.rb#L25

Referring to https://github.com/3scale/porta/blob/70cd39602a28c9d451b1aaaa3c4656145cb7466e/app/lib/backend/model_extensions/cinstance.rb#L46

app/lib/backend/model_extensions/cinstance.rb
hallelujah: This should not be changed

app/lib/backend/model_extensions/cinstance.rb
```diff
@@ -50,28 +45,6 @@ def delete_backend_application
         true
       end
 
-      def update_backend_user_key_to_application_id_mapping
```
hallelujah, line 53: So this method used the `user_key`, deleting the application in backend and recreating it

We should not remove it but change its content using the correct API or is it replaced by `update_backend_application`?

app/lib/backend/storage_rewrite.rb
```diff
@@ -25,7 +25,6 @@ def rewrite_all
 
       Cinstance.find_each do |cinstance|
         cinstance.send(:update_backend_application)
-        cinstance.send(:update_backend_user_key_to_application_id_mapping)
```
hallelujah, line 28: We should not remove this or is it replaced by `update_backend_application`?

Martouta: @hallelujah What this does is: if the user_key changed, remove the whole application_plan, and create it again with the new user_key. We do not need that anymore.
`update_backend_application` updates all attributes, and it does not contain user_key...  and afaik, we shouldn't add it there. However, we should call `update_provider_backend_service_if_user_key_changed` instead imo.

app/lib/backend/storage_rewrite.rb
```diff
@@ -91,7 +90,6 @@ def rewrite_provider(id)
 
       update_cinstance = ->(cinstance) do
         cinstance.send(:update_backend_application)
-        cinstance.send(:update_backend_user_key_to_application_id_mapping)
```
Martouta, line 94: Same as https://github.com/3scale/porta/pull/1843/files#r419308523 & https://github.com/3scale/porta/pull/1843/files#r419911421

app/lib/backend/model_extensions/cinstance.rb
Martouta: exactly 😄

app/lib/backend/model_extensions/cinstance.rb
Martouta: u r right! 🙏 💪

app/lib/backend/model_extensions/cinstance.rb
Martouta: same as https://github.com/3scale/porta/pull/1843/files#r419911421
We should just use `update_backend_application` and `update_provider_backend_service_if_user_key_changed` instead, which is what we are doing

app/lib/backend/storage_rewrite.rb
guicassolato: I think you meant "remove the whole application", not "application_plan".

Martouta:
@macejmic Everyone is 😄 I will wait for the 3 of you to agree on this one because it is a risky one and everyone's brain is helpful 💪

app/lib/backend/storage_rewrite.rb
Martouta: right hahahahaha 🙈 😂

Martouta:
Closing because we need [this](https://issues.redhat.com/browse/THREESCALE-5152?focusedCommentId=14077159&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-14077159) done before.

PR #1845, opened 2020-04-30
Do not notify AppKey destroyed if Cinstance is about to be destroyed

Martouta:
Closes [THREESCALE-5139](https://issues.redhat.com/browse/THREESCALE-5139)

We didn't want to send a message when the application is destroyed, but we weren't keeping in mind destroying in background 😄

### DIFF
--- a/app/models/application_key.rb
+++ b/app/models/application_key.rb
@@ -23,7 +23,7 @@ class ApplicationKey < ApplicationRecord
   after_commit :push_webhook_key_destroyed, :on => :destroy
 
   after_commit :notify_after_create, :on => :create, :if => :should_notify?
-  after_commit :notify_after_destroy, :on => :destroy, :if => :should_notify?
+  after_commit :notify_after_destroy, on: :destroy, if: -> { should_notify? && !destroyed_by_association }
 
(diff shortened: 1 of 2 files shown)


app/models/application_key.rb
```diff
@@ -23,7 +23,7 @@ class ApplicationKey < ApplicationRecord
   after_commit :push_webhook_key_destroyed, :on => :destroy
 
   after_commit :notify_after_create, :on => :create, :if => :should_notify?
-  after_commit :notify_after_destroy, :on => :destroy, :if => :should_notify?
+  after_commit :notify_after_destroy, :on => :destroy, if: proc { should_notify? && !destroyed_by_association }
```
macejmic, line 26: let's use `->` in here, instead of proc. better to stick with one way and not confuse juniors.

Martouta: @macejmic please use `suggestion` so I can commit from here 😎

macejmic:
```suggestion
after_commit :notify_after_destroy, :on => :destroy, if: -> { should_notify? && !destroyed_by_association }
```

PR #1847, opened 2020-04-30
Configuration of min_infix_len to address THREESCALE-5123 - Improve Admin Portal Search feature

samu91gn:
Improve the Admin Portal User search feature in order to find users by email address even in case the address contains two-character strings separated by dots (e.g. test@mail.co.uk)
This isn't currently possible due to `min_infix_len` being set to `3` in `thinking_sphinx.yml`

[THREESCALE-5123](https://issues.redhat.com/browse/THREESCALE-5123) (RFE), THREESCALE-632 (bug)

**Verification steps** 
1. Go to the Admin Portal and create a user with email `test@aaa.bb.com`
2. Force a reindex (or wait for it to happen automatically)
3. Make sure the user can be found via the seach feature by inserting `test@aaa.bb.com` in the search bar

### DIFF
--- a/app/controllers/api/integrations_controller.rb
+++ b/app/controllers/api/integrations_controller.rb
@@ -240,7 +240,7 @@ class Api::IntegrationsController < Api::BaseController
       :error_headers_auth_missing, :error_auth_missing, :error_status_no_match,
       :error_headers_no_match, :error_no_match, :error_status_limits_exceeded, :error_headers_limits_exceeded, :error_limits_exceeded,
       :api_test_path, :policies_config, proxy_rules_attributes: %i[_destroy id http_method pattern delta metric_id
-                                                                   redirect_url position last], oidc_configuration_attributes: OIDCConfiguration::Config::ATTRIBUTES
+                                                                   redirect_url position last], oidc_configuration_attributes: OIDCConfiguration::Config::ATTRIBUTES + [:id]
     ]
 
     if Rails.application.config.three_scale.apicast_custom_url || @proxy.saas_configuration_driven_apicast_self_managed?
--- a/app/controllers/api/services_controller.rb
+++ b/app/controllers/api/services_controller.rb
@@ -101,7 +101,7 @@ class Api::ServicesController < Api::BaseController
   end
 
   def proxy_params
-    oidc_params = %i[oidc_issuer_type oidc_issuer_endpoint jwt_claim_with_client_id jwt_claim_with_client_id_type] + [{oidc_configuration_attributes: OIDCConfiguration::Config::FLOWS}]
+    oidc_params = %i[oidc_issuer_type oidc_issuer_endpoint jwt_claim_with_client_id jwt_claim_with_client_id_type] + [{oidc_configuration_attributes: OIDCConfiguration::Config::FLOWS + [:id]}]
     permitted_params = oidc_params + %i[
       auth_user_key auth_app_id auth_app_key credentials_location hostname_rewrite secret_token
       error_status_auth_failed error_headers_auth_failed error_auth_failed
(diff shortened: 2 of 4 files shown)


Martouta:
Why not changing this?
https://github.com/3scale/porta/blob/37d27c7812de550d8b779e6d1e693c6816305a4e/config/thinking_sphinx.yml#L7

It feels strange having a different default number everywhere and 2 different configs.

Martouta:
It would also be good to have a test here:
https://github.com/3scale/porta/blob/37d27c7812de550d8b779e6d1e693c6816305a4e/test/unit/account/search_test.rb#L3

samu91gn:
> Why not changing this?
> https://github.com/3scale/porta/blob/37d27c7812de550d8b779e6d1e693c6816305a4e/config/thinking_sphinx.yml#L7
> 
> It feels strange having a different default number everywhere and 2 different configs.

That's what I did at first, then... the idea was to only configure this for the Account object so to avoid increasing the indexes amount and possibly impacting performances for all the other objects, does that make sense?

Good point about the test.

Martouta:
ah, makes sense 👍

app/indices/account_index.rb
```diff
@@ -15,6 +15,7 @@
   indexes users.id,                            as: :user_id
 
   set_property field_weights: { name: 2 }
+  set_property min_infix_len: ENV.fetch('ACCOUNT_MIN_INFIX_LEN', 2)
```
guicassolato, line 18: Oh. Please let's try not to introduce new env vars, especially in the middle of the code. If this really needs to be configurable, it should not be here, but using https://github.com/3scale/porta/blob/master/config/thinking_sphinx.yml#L7 instead.

Martouta:
Ideally, [sign the commit](https://github.com/3scale/porta/blob/master/CONTRIBUTING.md#is-there-any-other-requirement-to-open-a-pr).

app/indices/account_index.rb
kevprice83: @guicassolato Is there any reason why directly referencing an env var in the yaml instead of here is better? If we do it in the yaml you still have to reference it here no? The same as 
```delta_options: { threshold: SPHINX_DELTA_INTERVAL })```

Or are you saying there is another and better way of doing this?

guicassolato: The reason is maintainability, @kevprice83. Everything that can be configured ideally should be documented; eventually it needs to be supported in the operators/templates too. Keeping track of all those env vars is a PITA.

It's not the same as https://github.com/3scale/porta/blob/37d27c7812de550d8b779e6d1e693c6816305a4e/app/indices/account_index.rb#L6 because `SPHINX_DELTA_INTERVAL` is not configurable. It's hardcoded here:
https://github.com/3scale/porta/blob/37d27c7812de550d8b779e6d1e693c6816305a4e/config/initializers/sphinx.rb#L1

If you do in the yaml, then you have two options:
1. apply to all types of records, which @samu91gn [said](https://github.com/3scale/porta/pull/1847#issuecomment-621942878) he didn't want to do
2. significantly refactor that config file and the way we load it so one can set the option per model

My recommendation is: you are now changing the default value for the `Account` model from `3` to `2`. I assume this is acceptable then. Does this value have to be configurable? If no, just set fixed to `2` in https://github.com/3scale/porta/blob/c64aa9ac1398349e0a6466ad2aa200d02ce7a27b/app/indices/account_index.rb#L18 No env variable, and done.

samu91gn:
Closed as originally opened from wrong branch, I'm going to reopen with the suggested changes, thanks for all the tips.

app/indices/account_index.rb
thomasmaas: My vote for no env var

kevprice83: @thomasmaas this PR was closed, new one is #1851

PR #1851, opened 2020-05-01
Improve the Admin Portal User search - THREESCALE-632

samu91gn:
[Reopening from right branch, previous PR: [here](https://github.com/3scale/porta/pull/1847)]

Improve the Admin Portal User search feature in order to find users by email address even in case the address contains two-character strings separated by dots (e.g. test@mail.co.uk)
This isn't currently possible due to min_infix_len being set to 3 in thinking_sphinx.yml

[THREESCALE-632](https://issues.redhat.com/browse/THREESCALE-632) 

**Verification steps**
Go to the Admin Portal and create a user with email test@mail.co.uk
Force a reindex (or wait for it to happen automatically)
Make sure the user can be found via the seach feature by inserting test@mail.co.uk in the search bar

### DIFF
--- a/app/indices/account_index.rb
+++ b/app/indices/account_index.rb
@@ -12,7 +12,8 @@ ThinkingSphinx::Index.define(:account, with: :real_time) do
   indexes sphinx_user_ids, as: :user_id
 
   set_property field_weights: { name: 2 }
-
+  set_property charset_table: "0..9, A..Z->a..z, a..z, U+23, U+25, U+27, U+2A..U+2C, U+2E, U+3A, U+3B, U+3F, U+5F, U+60, U+7B, U+7D"
+  
   has :provider_account_id, type: :integer
   has :tenant_id, type: :integer
   has :state, type: :string
--- a/test/unit/account/search_test.rb
+++ b/test/unit/account/search_test.rb
@@ -65,6 +65,22 @@ class Account::SearchTest < ActiveSupport::TestCase
     assert_equal 'fo\/o', Account.search_ids('fo/o').query
   end
 
+  test 'search with query by email' do
+    ThinkingSphinx::Test.rt_run do
+      perform_enqueued_jobs(only: SphinxIndexationWorker) do
+        provider = FactoryBot.create(:simple_provider)
+        buyers = FactoryBot.create_list(:simple_buyer, 2, provider_account: provider)
+        ['foo@bar.co.example.com', 'foo@example.org'].each_with_index do |email, buyer_index|
+          FactoryBot.create(:admin, account: buyers[buyer_index], email: email)
+        end
+
+        buyers = provider.buyer_accounts.scope_search(:query => 'foo@bar.co.example.com')
+        expected_buyer = User.find_by!(email: 'foo@bar.co.example.com').account
+        assert_equal [expected_buyer], buyers
+      end
+    end
+  end
+
   test 'search_ids options' do
     ThinkingSphinx::Test.rt_run do
       assert Account.buyers.search_ids('foo').populate


guicassolato:
I think this is great. Just a small reminder about https://github.com/3scale/porta/pull/1847#issuecomment-621940432 before we can approve the PR.

kevprice83:
@samu91gn we can take a look at the unit tests for account search together. Should be straight forward enough.

Martouta:
@kevprice83 @samu91gn Any news?

samu91gn:
> Any news?

Hey Marta, not sure the test I wrote is correct, I will try to find some minutes to commit it tomorrow and will let you know because I have some doubts. Thanks!

samu91gn:
Updated the test @Martouta . Thanks

test/unit/account/search_test.rb
```diff
@@ -65,6 +65,22 @@ class Account::SearchTest < ActiveSupport::TestCase
     assert_equal 'fo\/o', Account.search_ids('fo/o').query
   end
 
+  test 'search with query by email' do
+    ThinkingSphinx::Test.run do
+     provider = FactoryBot.create(:provider_account)
```
Martouta, line 70:
```suggestion
     provider = FactoryBot.create(:simple_provider)
```
For efficiency 😄 Because `simple_provider` creates a provider but `provider_account` creates much more that you do not need for this test ;) like contracts and stuff that you are not using here.

test/unit/account/search_test.rb
```diff
@@ -65,6 +65,22 @@ class Account::SearchTest < ActiveSupport::TestCase
...
+     buyer = FactoryBot.create(:simple_buyer, :provider_account => provider)
+     buyer2 = FactoryBot.create(:simple_buyer, :provider_account => provider)
```
Martouta, line 72:
```suggestion
     buyer, buyer2 = FactoryBot.create_list(:simple_buyer, 2, provider_account: provider)
```

test/unit/account/search_test.rb
```diff
@@ -65,6 +65,22 @@ class Account::SearchTest < ActiveSupport::TestCase
...
+     buyer.users << FactoryBot.build(:admin, :account => buyer, :username => "uname1", :email => "foo@bar.co.uk")
+     buyer2.users << FactoryBot.build(:admin, :account => buyer2, :username => "uname2", :email => "foo@bar.com")
```
Martouta, line 74: I don't know if we are allowed to use `foo@bar.co.uk` and `foo@bar.com` because they might belong to somebody 👀 I feel that it is safer to use `foo@bar.co.example.com` and `foo@example.org` or anything that ends with `example.*` 😄

Martouta: Another detail: for hashes, plz try to use `foo: 'bar'` instead of `:foo => 'bar'` 😄

test/unit/account/search_test.rb
```diff
@@ -65,6 +65,22 @@ class Account::SearchTest < ActiveSupport::TestCase
...
+  
+     Indexer.run_after_index do
+      buyers = provider.buyer_accounts.scope_search(:query => 'foo@bar.co.uk')
+      assert_equal 1, buyers.count
+      assert_equal buyer, buyers.first
```
Martouta, line 79:
```suggestion
      assert_equal [buyer], buyers
```

So you get the count and the content in 1 single line 😄

test/unit/account/search_test.rb
Martouta, line 74: You can even make this stuff and what comes before much shorter like:
```ruby
buyers = FactoryBot.create_list(:simple_buyer, 2, provider_account: provider)
['foo@bar.co.example.com', 'foo@example.org'].each_with_index do |email, buyer_index|
  FactoryBot.create(:admin, account: buyers[buyer_index], email: email)
end
```

test/unit/account/search_test.rb
```diff
@@ -65,6 +65,22 @@ class Account::SearchTest < ActiveSupport::TestCase
...
+  test 'search with query by email' do
+    ThinkingSphinx::Test.run do
+     provider = FactoryBot.create(:provider_account)
+     buyer = FactoryBot.create(:simple_buyer, :provider_account => provider)
+     buyer2 = FactoryBot.create(:simple_buyer, :provider_account => provider)
+     buyer.users << FactoryBot.build(:admin, :account => buyer, :username => "uname1", :email => "foo@bar.co.uk")
+     buyer2.users << FactoryBot.build(:admin, :account => buyer2, :username => "uname2", :email => "foo@bar.com")
+  
+     Indexer.run_after_index do
+      buyers = provider.buyer_accounts.scope_search(:query => 'foo@bar.co.uk')
```
Martouta, line 77: Adapting this to what I mentioned above, of course 😄

test/unit/account/search_test.rb
Martouta: If you apply what I mentioned above, this should be:
```ruby
expected_buyer = User.find_by!(email: 'foo@bar.co.example.com').account
assert_equal [expected_buyer], buyers 
```

samu91gn:
that should be it @Martouta hope I didn't miss any suggestion :smile: thanks

app/indices/account_index.rb
```diff
@@ -15,6 +15,7 @@
   indexes users.id,                            as: :user_id
 
   set_property field_weights: { name: 2 }
+  set_property min_infix_len: 2
```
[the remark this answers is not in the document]
kevprice83: what do you think @hallelujah? @samu91gn does this still give the same results as `min_infix_len: 2`?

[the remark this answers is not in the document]
hallelujah: Did you try `index_exact_words=1` and run the tests?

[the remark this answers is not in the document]
kevprice83: @samu91gn can comment what different settings he has tested but I read the documentation and the following description about `index_exact_words` seems to suggest that the index size will be significantly impacted:

> When enabled, index_exact_words forces indexer to put the raw keywords in the index along with the stemmed versions. That, in turn, enables exact form operator in the query language to work. This impacts the index size and the indexing time. However, searching performance is not impacted at all. 

Whereas `dict = keywords` is enabled by default and hence adding `.` to the indexed charset looks to solve the issue and have the least impact on our current configuration, especially as we can restrict that addition to the charset to the account index only.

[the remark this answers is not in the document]
samu91gn: @hallelujah I've tested with `index_exact_words: 1` and `expand_keywords: 1` but the test never succeeded until I added the dot to the `charset_table` as explained above. On the other hand, adding the dot (and possibly other special characters accepted in an email like `set_property charset_table: "0..9, A..Z->a..z, a..z, ., -, _, +, %"`) is enough to pass the tests.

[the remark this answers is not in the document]
hallelujah: So the dot character is not in the charset_table, we should also add the `@` no?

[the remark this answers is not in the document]
samu91gn: @hallelujah I think `@` is escaped here: https://github.com/3scale/system/blob/1eeb2bc8e6bf575b176f1cc2c11e7006b7c7a07d/app/lib/three_scale/search.rb#L105 
In fact, if I add it to the charset_table, the entire email never matches.

app/indices/account_index.rb
```diff
@@ -15,7 +15,7 @@
   indexes users.id,                            as: :user_id
 
   set_property field_weights: { name: 2 }
-
+  set_property charset_table: "0..9, A..Z->a..z, a..z, U+23, U+25, U+27, U+2A..U+2C, U+2E, U+3A, U+3B, U+3F, U+5F, U+60, U+7B, U+7D"
```
hallelujah, line 18: They  are alredy defined in https://github.com/3scale/porta/blob/91c63772aab5068952293011060f142a03e30635/config/thinking_sphinx.yml#L17 and https://github.com/3scale/porta/blob/91c63772aab5068952293011060f142a03e30635/config/docker/thinking_sphinx.yml#L17

Please modify them there

samu91gn: @hallelujah - @kevprice83 and I thought it would have a smaller impact to restrict this change to the charset in the account index only by editing `account_index.rb`, isn't that how it works? I'll wait for your confirmation before applying the changes to `thinking_sphinx.yml` instead. Thanks!

hallelujah: @samu91gn @kevprice83 Yes that is better indeed!

samu91gn:
> Also please rebase on master, we have made some changes on sphinx lately
> 
> #1963

done & updated the test to use real time indexing

PR #1856, opened 2020-05-04
[Dashboard] optimization

macejmic:

Performance problems with the dashboard.

https://issues.redhat.com/browse/THREESCALE-3592

Tested with 1K products and backends -> time to load page reduced from 34 seconds(on 3scale 2.8.2) to 14 secconds
Tested with 1.5K backends -> time to load page reduced from 29 seconds(on 3scale 2.8.2) to 9 secconds
Tested with 1.5K products -> time to load page reduced from more than 1 minute/or internal error displayed due to long loading time (on 3scale 2.8.2) to 19 secconds

### DIFF
--- a/app/views/provider/admin/dashboards/_service_navigation.html.slim
+++ b/app/views/provider/admin/dashboards/_service_navigation.html.slim
@@ -23,11 +23,10 @@ nav.DashboardNavigation
                                        admin_service_applications_path(service),
                                        icon_name: 'cubes'
 
-    - unread_alerts = current_account.buyer_alerts.by_service(service).unread
-    - if can?(:manage, :monitoring) && unread_alerts.any?
+    - if can?(:manage, :monitoring) && service.unread_alerts_count > 0
       li.DashboardNavigation-list-item
         = dashboard_collection_link 'Alert',
-                                    unread_alerts,
+                                    service.unread_alerts_count,
                                     admin_service_alerts_path(service),
                                     icon_name: 'exclamation-triangle',
                                     notice: true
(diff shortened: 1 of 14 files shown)


app/views/provider/admin/dashboards/_service_navigation.html.slim
```diff
@@ -37,21 +36,21 @@ nav.DashboardNavigation
...
        = dashboard_collection_link 'ActiveDoc',
-                                    service.api_docs_services,
+                                    service.api_docs_services_count,
                                     admin_service_api_docs_path(service),
                                     icon_name: 'file-text'
     // Integration
     - if can?(:manage, :plans)
       li.DashboardNavigation-list-item
-        - if service.has_traffic? || (apiap? && service.proxy.proxy_configs.sandbox.any?)
+        - if service.cinstances_with_traffic_count > 0 || (apiap? && service.sandbox_proxy_configs_count > 0)
```
Martouta, line 53: Are you sure that this performs better than this?:
```ruby
        - if service.cinstances_with_traffic.exists? || (apiap? && service.sandbox_proxy_configs.exists?)
```
Because the `exists?` works with `LIMIT 1` which implies that instead of counting them all, it only checks until it find the first one if there is any 😄

app/views/provider/admin/dashboards/_service_navigation.html.slim
```diff
@@ -23,11 +23,10 @@ nav.DashboardNavigation
                                        admin_service_applications_path(service),
                                        icon_name: 'cubes'
 
-    - unread_alerts = current_account.buyer_alerts.by_service(service).unread
-    - if can?(:manage, :monitoring) && unread_alerts.any?
+    - if can?(:manage, :monitoring) && service.unread_alerts_count > 0
```
Martouta, line 26: Same as https://github.com/3scale/porta/pull/1856#discussion_r420406862

app/views/provider/admin/dashboards/_service_navigation.html.slim
macejmic: so i finally have an asnwer: https://issues.redhat.com/browse/THREESCALE-3592?focusedCommentId=14116380&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-14116380

PR #1861, opened 2020-05-04
4836-Account_Users main and modal pages

porueesq:

> Lists UI Strings for Accounts - Users

> `fixes #<4836>`

**Verification steps** 

> None

* UI Strings Jira: [THREESCALE-4836](https://issues.redhat.com/browse/THREESCALE-4836)
* Most of the strings contain text similar to the mockup.
* For strings different to the mockup, you will see this tag: **[PROPOSED]**

### DIFF
--- a/portafly/src/i18n/locales/en/accounts.yml
+++ b/portafly/src/i18n/locales/en/accounts.yml
@@ -23,7 +23,23 @@ invitations_tab_description: Label for the tab that contains the list of all inv
 ## Toolbar
 #-------# Bulk selector: available in shared.yml
 #-------# Delete button: available in shared.yml
-team_members_filter: Name
+team_members_filter: #----PROPOSED
+  by_name: Name
+  by_email: Email
+  by_sso_authentication: SSO authentication
+   yes: Yes
+   no: Not yet
+  by_role: Role
+   admin: Admin
+   member: Member
+  by_permission_group: Permission group
+   developer_portal: Developer Portal
+   billing: Billing
+   settings: Settings
+   accounts_applications: Accounts – Applications
+   analytics: Analytics
+   configuration_application_plans: Configuration and application plans
+   policy_registry: Policy registry
 team_members_filter_description: List containing the name criterion to filter team members
 team_members_typeahead: Filter by name #----PROPOSED
 team_members_typeahead_description: Default text to show when the user wants to filter by typing the text
@@ -52,6 +68,19 @@ team_members_table:
   role_header_value_description: Container with possible values of the roles of each team member
   permission_groups_header: Permission groups
   permission_groups_header_description: Column listing the permisions mapped to each team member
+  #-------# Actions header plural: available in shared.yml
+  #-------# Edit button: available in shared.yml
+  #-------# Delete button: available in shared.yml
+
+# Modal: Team member deletion
+team_member_deletion:
+  page: Delete selected team members #----PROPOSED: Approved
+  page_description: Body title of modal dialog to delete the selected team members from the table
+  admonition: This action will irreversibly delete the selected team members. Please confirm below to proceed with deleting them. #----PROPOSED
+  admonition_description: Admonition text to explain that the removal of the selected team members cannot be undone.
+  #-------# Delete button: available in shared.yml
+  #-------# Cancel button: available in shared.yml
+
 
 ## Tab: Invitations
 ## Toolbar
@@ -69,7 +98,7 @@ invitations_typeahead_description: Default text to show when the user filters by
 invitations_table:
   aria_label: Invitations table
   aria_label_description: Aria label of the table of invitations
-  empty_state: Currently, there are no invitations sent. Start sending invitations to your team members. #----PROPOSED
+  empty_state: Currently, there are no invitations sent. Start sending invitations to your team members. #----PROPOSED: placeholder
   empty_state_description: Message to show when there are no invitations to show, and prompt the user to send some invites
   recipient_header: Recipient
   recipient_header_description: Column listing the email addresses of invitation recipients
@@ -81,8 +110,8 @@ invitations_table:
     yes_with_time: Yes – {{utc_time}}
     not_yet: Not yet
   acceptance_values_description: Container with the possible responses to whether the invitation has been accepted or not.
-  actions_header: Actions
-  actions_header_description: Column listing the available actions to perform with each invitation
+  #-------# Action header: available in shared.yml
+  #-------# Delete button: available in shared.yml
 
 # Modal: Invitation deletion
 invitation_deletion:
--- a/portafly/src/i18n/locales/en/accounts.yml
+++ b/portafly/src/i18n/locales/en/accounts.yml
@@ -23,15 +23,15 @@ invitations_tab_description: Label for the tab that contains the list of all inv
 ## Toolbar
 #-------# Bulk selector: available in shared.yml
 #-------# Delete button: available in shared.yml
-team_members_filter: [Name]
-team_members_filter_description: List containing the criteria to filter team members
-[PROPOSED] team_members_typeahead: Filter by name
-team_members_typeahead_description: Filter by typing ahead the text
+team_members_filter: Name
+team_members_filter_description: List containing the name criterion to filter team members
+team_members_typeahead: Filter by name #----PROPOSED
+team_members_typeahead_description: Default text to show when the user wants to filter by typing the text
 
 team_members_table:
   aria_label: Team members table
   aria_label_description: Aria label of the table of team members
-  [PROPOSED] empty_state: Currently, there are no team members. Add team members.
+  empty_state: Currently, there are no team members. Add team members. #----PROPOSED
   empty_state_description: Message to show when there are no team members to show, and prompt the user to add members
   name_header: Name
   name_header_description: Column listing names of team members
@@ -39,10 +39,15 @@ team_members_table:
   email_header_description: Column listing the email addresses of team members
   sso_authentication_heaer: SSO authentication
   sso_authentication_header_description: Column listing if SSO authentication has been configured for team members
-  [PROPOSED] sso_authentication_value: [Yes, Not yet]
+  sso_authentication_value:
+    yes: Yes
+    not_yet: Not yet
   role_header: Role
   role_header_description: Column listing the roles associated to each team member
-  [PROPOSED] role_header_value: [Admin (you), Admin, Member]
+  role_header_value: #----PROPOSED
+    admin_you: Admin (you)
+    admin: Admin
+    member: Member
   permission_groups_header: Permission groups
   permission_groups_header_description: Column listing the permisions mapped to each team member
 
@@ -50,31 +55,37 @@ team_members_table:
 ## Toolbar
 #-------# Bulk selector: available in shared.yml
 #-------# Delete button: available in shared.yml
-[PROPOSED] invitations_filter: [Recipient, Acceptance]
+invitations_filter:  #----PROPOSED
+ by_recipient: Recipient
+ by_acceptance: Acceptance
 invitations_filter_description: List containing the criteria to filter invitations
-invitations_typeahead: [Filter by recipient, Filter by acceptance]
-invitations_typeahead_description: Filter by recipient or acceptance, as the user types the text.
+invitations_typeahead:
+  filter_by_recipient: Filter by recipient
+  filter_by_acceptance: Filter by acceptance
+invitations_typeahead_description: Default text to show when the user filters by recipient or acceptance.
 
 invitations_table:
   aria_label: Invitations table
   aria_label_description: Aria label of the table of invitations
-  [PROPOSED] empty_state: Currently, there are no invitations sent. Start sending invitations to your team members.
+  empty_state: Currently, there are no invitations sent. Start sending invitations to your team members. #----PROPOSED
   empty_state_description: Message to show when there are no invitations to show, and prompt the user to send some invites
   recipient_header: Recipient
   recipient_header_description: Column listing the email addresses of invitation recipients
-  [PROPOSED] sent_header: Sent on UTC time
+  sent_header: Sent on UTC time #----PROPOSED
   sent_header_description: Column listing the date when invitations were sent
-  [PROPOSED] acceptance_header: Accepted on UTC time
+  acceptance_header: Accepted on UTC time #----PROPOSED
   acceptance_header_description: Column listing the date when the invitations were accepted
-  [PROPOSED] acceptance_values: [Yes - UTC time, Not yet]
+  acceptance_values: #----PROPOSED
+    yes_with_time: Yes - UTC time
+    not_yet: Not yet
   actions_header: Actions
   actions_header_description: Column listing the available actions to perform with each invitation
 
 # Modal: Invitation deletion
 invitation_deletion:
-  [PROPOSED] invitation_deletion_page: Delete selected invitations
+  invitation_deletion_page: Delete selected invitations #----PROPOSED
   invitation_deletion_page_description: Body title of modal dialog to delete the selected invitations from the table
-  [PROPOSED] admonition: This action will irreversibly delete the selected invitations. Please confirm below to proceed with the deletion.
+  admonition: This action will irreversibly delete the selected invitations. Please confirm below to proceed with the deletion. #----PROPOSED
   admonition_description: Admonition text to explain that the removal of the selected invitations cannot be undone.
   #-------# Delete button: available in shared.yml
   #-------# Cancel button: available in shared.yml
@@ -85,9 +96,9 @@ invitation_sending:
   invitation_sending_page_description: Body title of modal dialog to send invitation to a new team member
   invitation_sending_label: Send invitation to *
   invitation_sending_label_description: Label prompting the user to specify the email address of the invitation recipient
-  [PROPOSED] invitation_sending_validation: Please indicate a valid email address of the colleague or friend you want to invite to your team.
+  invitation_sending_validation: Please indicate a valid email address of the colleague or friend you want to invite to your team. #----PROPOSED
   invitation_sending_validation_description: Validation message to prompt user to specify a valid email address; to prevent an empty value, or bad format email address.
-  [PROPOSED] invitation_sending_helper: Email address of a colleague or friend you want to invite to your team
+  invitation_sending_helper: Email address of a colleague or friend you want to invite to your team #----PROPOSED
   invitation_sending_helper_description: Text to prompt the user to specify an email address.
   #-------# Send button: available in shared.yml
   #-------# Cancel button: available in shared.yml
--- a/portafly/src/i18n/locales/en/accounts.yml
+++ b/portafly/src/i18n/locales/en/accounts.yml
@@ -45,7 +45,7 @@ team_members_table:
   sso_authentication_value_description: Container of possible values about users having SSO configured
   role_header: Role
   role_header_description: Column listing the roles associated to each team member
-  role_header_value: #----PROPOSED
+  role_header_value: #----PROPOSED: Approved
     admin_you: Admin (you)
     admin: Admin
     member: Member
@@ -57,7 +57,7 @@ team_members_table:
 ## Toolbar
 #-------# Bulk selector: available in shared.yml
 #-------# Delete button: available in shared.yml
-invitations_filter:  #----PROPOSED
+invitations_filter:  #----PROPOSED: Approved
  by_recipient: Recipient
  by_acceptance: Acceptance
 invitations_filter_description: List containing the criteria to filter invitations
@@ -78,7 +78,7 @@ invitations_table:
   acceptance_header: Accepted on UTC time #----PROPOSED
   acceptance_header_description: Column listing the date when the invitations were accepted
   acceptance_values: #----PROPOSED
-    yes_with_time: Yes - {{utc_time}}
+    yes_with_time: Yes – {{utc_time}}
     not_yet: Not yet
   acceptance_values_description: Container with the possible responses to whether the invitation has been accepted or not.
   actions_header: Actions
@@ -86,9 +86,9 @@ invitations_table:
 
 # Modal: Invitation deletion
 invitation_deletion:
-  page: Delete selected invitations #----PROPOSED
+  page: Delete selected invitations #----PROPOSED: Approved
   page_description: Body title of modal dialog to delete the selected invitations from the table
-  admonition: This action will irreversibly delete the selected invitations. Please confirm below to proceed with the deletion. #----PROPOSED
+  admonition: This action will irreversibly delete the selected invitations. Please confirm below to proceed with deleting them. #----PROPOSED
   admonition_description: Admonition text to explain that the removal of the selected invitations cannot be undone.
   #-------# Delete button: available in shared.yml
   #-------# Cancel button: available in shared.yml
--- a/portafly/src/i18n/locales/en/accounts.yml
+++ b/portafly/src/i18n/locales/en/accounts.yml
@@ -42,12 +42,14 @@ team_members_table:
   sso_authentication_value:
     yes: Yes
     not_yet: Not yet
+  sso_authentication_value_description: Container of possible values about users having SSO configured
   role_header: Role
   role_header_description: Column listing the roles associated to each team member
   role_header_value: #----PROPOSED
     admin_you: Admin (you)
     admin: Admin
     member: Member
+  role_header_value_description: Container with possible values of the roles of each team member
   permission_groups_header: Permission groups
   permission_groups_header_description: Column listing the permisions mapped to each team member
 
@@ -62,7 +64,7 @@ invitations_filter_description: List containing the criteria to filter invitatio
 invitations_typeahead:
   filter_by_recipient: Filter by recipient
   filter_by_acceptance: Filter by acceptance
-invitations_typeahead_description: Default text to show when the user filters by recipient or acceptance.
+invitations_typeahead_description: Default text to show when the user filters by recipient or acceptance
 
 invitations_table:
   aria_label: Invitations table
@@ -76,8 +78,9 @@ invitations_table:
   acceptance_header: Accepted on UTC time #----PROPOSED
   acceptance_header_description: Column listing the date when the invitations were accepted
   acceptance_values: #----PROPOSED
-    yes_with_time: Yes - UTC time
+    yes_with_time: Yes - {{utc_time}}
     not_yet: Not yet
+  acceptance_values_description: Container with the possible responses to whether the invitation has been accepted or not.
   actions_header: Actions
   actions_header_description: Column listing the available actions to perform with each invitation
 
--- a/portafly/src/i18n/locales/en/accounts.yml
+++ b/portafly/src/i18n/locales/en/accounts.yml
@@ -21,9 +21,8 @@ invitations_tab_description: Label for the tab that contains the list of all inv
 
 ## Tab: Listing
 ## Toolbar
-bulk_selector: {{number}} selected
-bulk_selector_description: Label for the bulk selector to indicate the number of chosen team members. This label is visible only when team members were selected.
-# Delete button: available in shared.yml
+#-------# Bulk selector: available in shared.yml
+#-------# Delete button: available in shared.yml
 team_members_filter: [Name]
 team_members_filter_description: List containing the criteria to filter team members
 [PROPOSED] team_members_typeahead: Filter by name
@@ -49,13 +48,12 @@ team_members_table:
 
 ## Tab: Invitations
 ## Toolbar
-bulk_selector: {{number}} selected
-bulk_selector_description: Label for the bulk selector to indicate the number of chosen invitations. This label is visible only when invitations were selected.
+#-------# Bulk selector: available in shared.yml
 #-------# Delete button: available in shared.yml
 [PROPOSED] invitations_filter: [Recipient, Acceptance]
 invitations_filter_description: List containing the criteria to filter invitations
 invitations_typeahead: [Filter by recipient, Filter by acceptance]
-invitations_typeahead_description: Filter by typing ahead the text.
+invitations_typeahead_description: Filter by recipient or acceptance, as the user types the text.
 
 invitations_table:
   aria_label: Invitations table
--- a/portafly/src/i18n/locales/en/accounts.yml
+++ b/portafly/src/i18n/locales/en/accounts.yml
@@ -111,6 +111,8 @@ invitations_table:
     not_yet: Not yet
   acceptance_values_description: Container with the possible responses to whether the invitation has been accepted or not.
   #-------# Action header: available in shared.yml
+  resend_button: Resend
+  resend_button_description: Button link to send invitations again
   #-------# Delete button: available in shared.yml
 
 # Modal: Invitation deletion
--- a/portafly/src/i18n/locales/en/accounts.yml
+++ b/portafly/src/i18n/locales/en/accounts.yml
@@ -83,8 +83,8 @@ invitations_table:
 
 # Modal: Invitation deletion
 invitation_deletion:
-  invitation_deletion_page: Delete selected invitations #----PROPOSED
-  invitation_deletion_page_description: Body title of modal dialog to delete the selected invitations from the table
+  page: Delete selected invitations #----PROPOSED
+  page_description: Body title of modal dialog to delete the selected invitations from the table
   admonition: This action will irreversibly delete the selected invitations. Please confirm below to proceed with the deletion. #----PROPOSED
   admonition_description: Admonition text to explain that the removal of the selected invitations cannot be undone.
   #-------# Delete button: available in shared.yml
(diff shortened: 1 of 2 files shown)


portafly/src/i18n/locales/en/accounts.yml
```diff
@@ -1,20 +1,95 @@
...
+team_members_filter: [Name]
+team_members_filter_description: List containing the criteria to filter team members
+[PROPOSED] team_members_typeahead: Filter by name
+team_members_typeahead_description: Filter by typing ahead the text
+
+team_members_table:
+  aria_label: Team members table
+  aria_label_description: Aria label of the table of team members
+  [PROPOSED] empty_state: Currently, there are no team members. Add team members.
+  empty_state_description: Message to show when there are no team members to show, and prompt the user to add members
```
dfennessy, line 36:
```suggestion
  empty_state_description: The message when there are no team members to show, and prompt the user to add members
```

portafly/src/i18n/locales/en/accounts.yml
```diff
@@ -1,20 +1,95 @@
...
+
+## Tab: Invitations
+## Toolbar
+bulk_selector: {{number}} selected
+bulk_selector_description: Label for the bulk selector to indicate the number of chosen invitations. This label is visible only when invitations were selected.
+#-------# Delete button: available in shared.yml
+[PROPOSED] invitations_filter: [Recipient, Acceptance]
+invitations_filter_description: List containing the criteria to filter invitations
+invitations_typeahead: [Filter by recipient, Filter by acceptance]
+invitations_typeahead_description: Filter by typing ahead the text.
```
dfennessy, line 58: Just a punctuation question for consideration.

Should full stops (periods) be used or not be used or either way is okay?

dfennessy: Also, I'm not sure if the sentence makes sense.

Filter by typing....

portafly/src/i18n/locales/en/accounts.yml
```diff
@@ -1,20 +1,95 @@
...
+  [PROPOSED] role_header_value: [Admin (you), Admin, Member]
+  permission_groups_header: Permission groups
+  permission_groups_header_description: Column listing the permisions mapped to each team member
+
+## Tab: Invitations
+## Toolbar
+bulk_selector: {{number}} selected
+bulk_selector_description: Label for the bulk selector to indicate the number of chosen invitations. This label is visible only when invitations were selected.
+#-------# Delete button: available in shared.yml
+[PROPOSED] invitations_filter: [Recipient, Acceptance]
```
damianpm, line 55: In the cases where we have several options like [a, b, c] it's better to have a single entry for every option, in this way we don't need to implement logic to extract the value.
Only when the literal value to show is `[a, b, c]` it's OK to have a single entry, but when those are possible values (or a, or b, or c) is better having an entry for each option

portafly/src/i18n/locales/en/accounts.yml
```diff
@@ -1,20 +1,95 @@
...
+invitations_filter_description: List containing the criteria to filter invitations
+invitations_typeahead: [Filter by recipient, Filter by acceptance]
```
damianpm, line 57: Same here, better 2 items

damianpm:
> **What this PR does / why we need it**:
> 
> > Lists UI Strings for Accounts - Users
> 
> **Which issue(s) this PR fixes**
> 
> > `fixes #<4836>`
> 
> **Verification steps**
> 
> > None
> 
> **Special notes for your reviewer**:
> 
>     * Most of the strings contain text similar to the mockup.
> 
>     * For strings different to the mockup, you will see this tag: **[PROPOSED]**

@porueesq next time, can you paste the link to the Jira issues, not the ticket number, please?
Thanks, it's looking great!

portafly/src/i18n/locales/en/accounts.yml
thomasmaas: > The message when there are no team members to show, and prompt the user to add members

Shouldn't this be replaced by actual meaningful text?

portafly/src/i18n/locales/en/accounts.yml
```diff
@@ -1,20 +1,95 @@
...
+  [PROPOSED] invitation_deletion_page: Delete selected invitations
+  invitation_deletion_page_description: Body title of modal dialog to delete the selected invitations from the table
+  [PROPOSED] admonition: This action will irreversibly delete the selected invitations. Please confirm below to proceed with the deletion.
+  admonition_description: Admonition text to explain that the removal of the selected invitations cannot be undone.
+  #-------# Delete button: available in shared.yml
+  #-------# Cancel button: available in shared.yml
+
+# Modal: Invitation to a new member
+invitation_sending:
+  invitation_sending_page: Invite a new team member
```
josemigallas, line 86: I think including the "section" prefix is a bit redundant. Either use prefixes:
```suggestion
invitation_sending_page: Invite a new team member
```
Or organize by sections:
```suggestion
invitation_sending:
  page: Invite a new team member
```

portafly/src/i18n/locales/en/accounts.yml
porueesq: @thomasmaas 
The actual text is proposed in `empty_state`
`empty_state_description` is aimed at translators or localizers who in the future might need contextual information about the UI string.

portafly/src/i18n/locales/en/accounts.yml
porueesq: @dfennessy In general, punctuation is not used, unless full sentences (SVO) and questions. For `xxx_description`, since they are for internal use, we can use punctuation as we like.

_Filter by typing_: you're right. It should be "filter as you type...", "filter as the user types...". Applying this change.

portafly/src/i18n/locales/en/accounts.yml
```diff
@@ -1,20 +1,95 @@
...
-  col_admin_description: Title of the Admin column
-  col_signup: Signup date
-  col_signup_description: Title of the Signup date column
-  col_apps: Apps
-  col_apps_description: Title of the Apps column
-  col_state: State
-  col_state_description: Title of the State column
+# Account settings > Users
+
+title_page: Accounts | Team members
```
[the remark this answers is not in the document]
porueesq: @damianpm The [mockup](https://marvelapp.com/ab13h96/screen/68426948)  in [THREESCALE-4765](https://issues.redhat.com/browse/THREESCALE-4765) uses *Team members*  as the page title.

portafly/src/i18n/locales/en/accounts.yml
```diff
@@ -1,20 +1,95 @@
...
+# Delete button: available in shared.yml
+team_members_filter: [Name]
+team_members_filter_description: List containing the criteria to filter team members
+[PROPOSED] team_members_typeahead: Filter by name
+team_members_typeahead_description: Filter by typing ahead the text
+
+team_members_table:
+  aria_label: Team members table
+  aria_label_description: Aria label of the table of team members
+  [PROPOSED] empty_state: Currently, there are no team members. Add team members.
```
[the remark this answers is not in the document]
porueesq: Thanks for the suggestion @damianpm . We'll discuss this option in our next standup.

portafly/src/i18n/locales/en/accounts.yml
```diff
@@ -1,20 +1,95 @@
...
+  empty_state_description: Message to show when there are no team members to show, and prompt the user to add members
+  name_header: Name
+  name_header_description: Column listing names of team members
+  email_header: Email
+  email_header_description: Column listing the email addresses of team members
+  sso_authentication_heaer: SSO authentication
+  sso_authentication_header_description: Column listing if SSO authentication has been configured for team members
+  [PROPOSED] sso_authentication_value: [Yes, Not yet]
```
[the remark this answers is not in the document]
porueesq: @damianpm How can I portray more than one option? For example, does this work for you?

```
sso_authentication_value_yes: Yes
sso_authentication_value_notyet: Not yet
```

portafly/src/i18n/locales/en/accounts.yml
```diff
@@ -1,20 +1,95 @@
...
+  role_header: Role
+  role_header_description: Column listing the roles associated to each team member
+  [PROPOSED] role_header_value: [Admin (you), Admin, Member]
```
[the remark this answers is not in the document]
porueesq: Pending answer from https://github.com/3scale/porta/pull/1861#discussion_r421706093

portafly/src/i18n/locales/en/accounts.yml
```diff
@@ -1,20 +1,95 @@
...
+  permission_groups_header: Permission groups
+  permission_groups_header_description: Column listing the permisions mapped to each team member
+
+## Tab: Invitations
+## Toolbar
+bulk_selector: {{number}} selected
+bulk_selector_description: Label for the bulk selector to indicate the number of chosen invitations. This label is visible only when invitations were selected.
```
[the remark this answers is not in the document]
porueesq: Sure @damianpm . I'll move this to the `shared.yml` file.

portafly/src/i18n/locales/en/accounts.yml
porueesq: Pending answer from #1861 (comment)

portafly/src/i18n/locales/en/accounts.yml
porueesq: Pending answer from #1861 (comment)

portafly/src/i18n/locales/en/accounts.yml
porueesq: I'll test the alternatives and submit the changes.

portafly/src/i18n/locales/en/accounts.yml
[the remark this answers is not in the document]
damianpm: Yes, that works fine!

portafly/src/i18n/locales/en/accounts.yml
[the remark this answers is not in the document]
josemigallas: @porueesq I would take this as an option, since brackets can't be used this way in YAML (they are reserved for arrays). It has to me moved in a comment o written w/o them like this:
```suggestion
  PROPOSED empty_state: Currently, there are no team members. Add team members.
```

portafly/src/i18n/locales/en/accounts.yml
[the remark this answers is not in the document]
josemigallas: Either that or:
```suggestion
  sso_authentication_value:
    yes: Yes
    not_yet: Not yet
```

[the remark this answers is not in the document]
porueesq: Thanks @damianpm and @josemigallas for the suggestions. I'll review the items with options we currently have and submit the changes.

portafly/src/i18n/locales/en/accounts.yml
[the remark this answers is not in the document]
porueesq: @josemigallas We'll be using `#---PROPOSED` at the end of the line. Ideally, you should receive files with all the strings approved, so this temporary tag will be removed by when you start coding.

portafly/src/i18n/locales/en/accounts.yml
porueesq: @josemigallas changes are ready for your review: https://github.com/3scale/porta/pull/1861/commits/a723442a74fce9ed9f67177c84a3079603041854

portafly/src/i18n/locales/en/accounts.yml
```diff
@@ -1,20 +1,107 @@
...
+invitations_table:
+  aria_label: Invitations table
+  aria_label_description: Aria label of the table of invitations
+  empty_state: Currently, there are no invitations sent. Start sending invitations to your team members. #----PROPOSED
+  empty_state_description: Message to show when there are no invitations to show, and prompt the user to send some invites
+  recipient_header: Recipient
+  recipient_header_description: Column listing the email addresses of invitation recipients
+  sent_header: Sent on UTC time #----PROPOSED
+  sent_header_description: Column listing the date when invitations were sent
+  acceptance_header: Accepted on UTC time #----PROPOSED
```
cntlsn, line 78: @porueesq same as above

portafly/src/i18n/locales/en/accounts.yml
```diff
@@ -1,20 +1,107 @@
...
+  acceptance_header_description: Column listing the date when the invitations were accepted
+  acceptance_values: #----PROPOSED
```
cntlsn, line 80: @porueesq looks good. Just a note, after the "Yes" there should be a en dash "–", not an hyphen "-".

portafly/src/i18n/locales/en/accounts.yml
```diff
@@ -1,20 +1,107 @@
...
+    yes_with_time: Yes - {{utc_time}}
+    not_yet: Not yet
+  acceptance_values_description: Container with the possible responses to whether the invitation has been accepted or not.
+  actions_header: Actions
+  actions_header_description: Column listing the available actions to perform with each invitation
```
cntlsn, line 85: @porueesq I think we should create a UI string for the action too – I think the only option for actions is actually the "Resend" link button. @thomasmaas can you please double check?

portafly/src/i18n/locales/en/accounts.yml
```diff
@@ -1,20 +1,107 @@
...
+
```
cntlsn, line 86: @porueesq UI string is missing for the button link in the contextual menu triggered by the kebab menu:
* Delete

portafly/src/i18n/locales/en/accounts.yml
```diff
@@ -1,20 +1,107 @@
...
+# Modal: Invitation deletion
+invitation_deletion:
+  page: Delete selected invitations #----PROPOSED
+  page_description: Body title of modal dialog to delete the selected invitations from the table
+  admonition: This action will irreversibly delete the selected invitations. Please confirm below to proceed with the deletion. #----PROPOSED
+  admonition_description: Admonition text to explain that the removal of the selected invitations cannot be undone.
+  #-------# Delete button: available in shared.yml
+  #-------# Cancel button: available in shared.yml
+
```
cntlsn, line 95: @porueesq UI strings for the Members deletion modal should be added. The modal is not present in the mockups but it's described in the ux requirements in https://issues.redhat.com/browse/THREESCALE-4765

portafly/src/i18n/locales/en/accounts.yml
```diff
@@ -1,20 +1,107 @@
...
+  acceptance_values: #----PROPOSED
+    yes_with_time: Yes - {{utc_time}}
+    not_yet: Not yet
+  acceptance_values_description: Container with the possible responses to whether the invitation has been accepted or not.
+  actions_header: Actions
+  actions_header_description: Column listing the available actions to perform with each invitation
+
+# Modal: Invitation deletion
+invitation_deletion:
+  page: Delete selected invitations #----PROPOSED
```
cntlsn, line 89: @porueesq looks good!

portafly/src/i18n/locales/en/accounts.yml
```diff
@@ -1,20 +1,107 @@
...
+  page_description: Body title of modal dialog to delete the selected invitations from the table
+  admonition: This action will irreversibly delete the selected invitations. Please confirm below to proceed with the deletion. #----PROPOSED
```
cntlsn, line 91: @porueesq I'm not sure about "the deletion". By looking at https://www.merriam-webster.com/dictionary/deletion it seems like the article is only needed for the plural form?

portafly/src/i18n/locales/en/accounts.yml
```diff
@@ -1,20 +1,107 @@
...
+  #-------# Delete button: available in shared.yml
+  #-------# Cancel button: available in shared.yml
+
+# Modal: Invitation to a new member
+invitation_sending:
+  page: Invite a new team member
+  page_description: Body title of modal dialog to send invitation to a new team member
+  label: Send invitation to *
+  label_description: Label prompting the user to specify the email address of the invitation recipient
+  validation: Please indicate a valid email address of the colleague or friend you want to invite to your team. #----PROPOSED
```
cntlsn, line 102: @porueesq I think that for technical reasons it's better to use "Account" here. The user is inviting somebody to use their same Account, and Account is a concept in 3scale, team is not. I understand the confusion, which makes me think that a paragraph above the form explaining what's the output of this action would be beneficial. @thomasmaas what do you think?

portafly/src/i18n/locales/en/accounts.yml
```diff
@@ -1,20 +1,107 @@
...
+  validation_description: Validation message to prompt user to specify a valid email address; to prevent an empty value, or bad format email address.
+  helper: Email address of a colleague or friend you want to invite to your team #----PROPOSED
```
cntlsn, line 104: @porueesq same as above

portafly/src/i18n/locales/en/shared.yml
```diff
@@ -29,4 +29,14 @@ format:
...
   format_description: Utilities to format text
-
+shared_elements:
+  bulk_selector: {{number}} selected
+  bulk_selector_description: Label for the bulk selector to indicate the number of selected items. This label is visible only when items are selected.
+  send_button: Send
+  send_button_description: Label to perform the sending of items
+  delete_button: Delete
+  delete_button_description: Label to perform the deletion of selected items
+  delete_button_tooltip: Select at least an item to enable the option to delete.
```
cntlsn, line 39: @porueesq "delete option" instead of "the option to delete" looks like a more common UI string, what do you think?

portafly/src/i18n/locales/en/accounts.yml
```diff
@@ -1,20 +1,107 @@
...
+listing_tab: Listing
+listing_tab_description: Label for the tab that contains the list of all team members, selection criteria and actions that can be applied on them.
+invitations_tab: Invitations
+invitations_tab_description: Label for the tab that contains the list of all invitations, selection criteria and actions that can be applied on them.
+
+## Tab: Listing
+## Toolbar
+#-------# Bulk selector: available in shared.yml
+#-------# Delete button: available in shared.yml
+team_members_filter: Name
```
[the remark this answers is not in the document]
damianpm: It's true, we are missing `Recipient` and `Accepted?`

portafly/src/i18n/locales/en/accounts.yml
```diff
@@ -1,20 +1,107 @@
...
+invitations_typeahead_description: Default text to show when the user filters by recipient or acceptance
+
+invitations_table:
+  aria_label: Invitations table
+  aria_label_description: Aria label of the table of invitations
+  empty_state: Currently, there are no invitations sent. Start sending invitations to your team members. #----PROPOSED
+  empty_state_description: Message to show when there are no invitations to show, and prompt the user to send some invites
+  recipient_header: Recipient
+  recipient_header_description: Column listing the email addresses of invitation recipients
+  sent_header: Sent on UTC time #----PROPOSED
```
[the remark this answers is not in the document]
damianpm: @cntlsn I'm not sure, but it's possible to have other timezones, not only UTC ?

portafly/src/i18n/locales/en/accounts.yml
```diff
@@ -1,20 +1,107 @@
...
+#-------# Delete button: available in shared.yml
+team_members_filter: Name
+team_members_filter_description: List containing the name criterion to filter team members
+team_members_typeahead: Filter by name #----PROPOSED
+team_members_typeahead_description: Default text to show when the user wants to filter by typing the text
+
+team_members_table:
+  aria_label: Team members table
+  aria_label_description: Aria label of the table of team members
+  empty_state: Currently, there are no team members. Add team members. #----PROPOSED
```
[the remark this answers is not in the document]
porueesq: Yes, adding this as a placeholder. At some point we'll need to review this string.

portafly/src/i18n/locales/en/accounts.yml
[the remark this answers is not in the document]
porueesq: @cntlsn Due to accessibility, special characters such as parentheses are discouraged. I'd simply place "Sent" , and explain in the docs that it uses UTC time. In this way, if other time zone systems are used, we update the docs instead of the UI. WDYT?

portafly/src/i18n/locales/en/accounts.yml
porueesq: Updated.

portafly/src/i18n/locales/en/accounts.yml
porueesq: Rewording this.

portafly/src/i18n/locales/en/accounts.yml
porueesq: @cntlsn If the title of the page is **Team members**, but the concept refers to a group of people sharing the same account, we'll need to add a paragraph to explain this.

Also, we'll need to find a way to distinguish from an _user account_ (the one used to login to 3scale) versus an _account_ defined from a technical perspective. Let's avoid semantic overload.

portafly/src/i18n/locales/en/shared.yml
porueesq: Yes, it's simpler. I'll modify it.

portafly/src/i18n/locales/en/accounts.yml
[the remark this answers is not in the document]
porueesq: @damianpm what you mentioned is the criteria to filter invitations. You can find them in:
L60 >> `invitations_filter:  #----PROPOSED`
Proposal accepted: https://github.com/3scale/porta/pull/1861/files#r423672387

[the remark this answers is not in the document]
porueesq: Adding the missing filters for team members:

    Name > text input field
    Email > text input field
    SSO authentication (if enabled) > Dropdown menu with single selection ('yes', 'not yet')
    Role > Dropdown menu with single selection ('admin', 'member')
    Permission groups > Dropdown menu with multi-selection ('Developer portal', 'Billing', 'Settings', 'Accounts – Applications', 'Analytics', 'Integration & Application plans', 'Policy registry')

portafly/src/i18n/locales/en/accounts.yml
```diff
@@ -1,20 +1,107 @@
...
+  role_header: Role
+  role_header_description: Column listing the roles associated to each team member
+  role_header_value: #----PROPOSED
+    admin_you: Admin (you)
+    admin: Admin
+    member: Member
+  role_header_value_description: Container with possible values of the roles of each team member
+  permission_groups_header: Permission groups
+  permission_groups_header_description: Column listing the permisions mapped to each team member
+
```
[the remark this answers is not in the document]
porueesq: In this case, I'll propose an **Actions** column, for accessibility purposes.

portafly/src/i18n/locales/en/accounts.yml
porueesq: Commented in https://github.com/3scale/porta/pull/1861#discussion_r423784270

portafly/src/i18n/locales/en/accounts.yml
porueesq: Created two strings in the shared file: **Actions** and **Action**

portafly/src/i18n/locales/en/accounts.yml
porueesq: Added as a placeholder, using _Team members_, instead of "accounts". See https://github.com/3scale/porta/pull/1861#discussion_r423792110 for more details

portafly/src/i18n/locales/en/accounts.yml
porueesq: See https://github.com/3scale/porta/pull/1861#discussion_r423792110

portafly/src/i18n/locales/en/accounts.yml
cntlsn: I meant to say that a UI string maybe should be created for the possible actions listed in the column. And from what I recall the only possible action button displayed in the Actions column is "Resend"

portafly/src/i18n/locales/en/accounts.yml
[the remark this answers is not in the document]
cntlsn: @porueesq  the contextual row menu triggered by the kebab button is a Patternfly standard. Exposing the actions would mean featuring two more columns in this case, and to keep consistency it would eventually mean many more columns in scenarios where multiple actions are available for a row -- which makes it a non-viable option.

portafly/src/i18n/locales/en/accounts.yml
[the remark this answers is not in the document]
cntlsn: @damianpm it is possible to select a different time zone, but user needs to go to settings and change it at the Account level (not at the personal level).
@porueesq for the reason I mentioned above I think it is hard for a user to understand exactly what the date is referred to (I'm thinking of an edge case where multiple members in different time zones share the same account). for this reason if we don't want to display the timezone in the table header I think we should at least display a help icon that would trigger a popover including:
* info about the current time zone
* info about how to change it (if user has permission)

portafly/src/i18n/locales/en/accounts.yml
cntlsn: @porueesq I agree with adding a paragraph in the page header to explain this!
About the differentiation between Account (the 3scale entity) and User account (linked to a specific user), I agree that this is misleading, I'm just not sure this can be solved in this context.
Here is my thinking: users are familiar with the concept of a User account (because that is everywhere on the internet), but they are not with the concept of an Account as used by 3scale, which by the way is linked to the Group/Organization, not to a specific user (unless only one user exists). So to make things easy for users, I would keep the concept of a User account (or simple "Account") intact, and change the wording for the 3scale Account to something like Organization or Team, which is actually a collection of user accounts at the Company level.
But, it is possible that this would be a breaking change, so I'm not sure we want to make this change at this point...
@thomasmaas comments are welcome.

portafly/src/i18n/locales/en/accounts.yml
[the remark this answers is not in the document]
damianpm: @porueesq I don't think adding the UTC info in the docs is enough, the user need to see that info in the UI. I'd keep it in the table by now. 

The other option @cntlsn mentioned (the help icon) could be a good idea, but I say let's keep it like this in the scope of this PR, and later we could add the help icon if needed (I think we need to do more research, for example to see if we are going to show in the table other timezones for an edge case)

portafly/src/i18n/locales/en/accounts.yml
thomasmaas: I would probably use 'organization' and reserve 'teams' for when we might one day introduce something that lets you further organize your org into teams.

cntlsn: @thomasmaas @porueesq 
So, to summarize and make sure we are all on the same page.
* What is now a 3scale "Account" concept, will become an "Organization":
  * There will always be two types of Organizations (one API Provider organization, multiple API Consumer organizations) and will both be called with the same "Organization" term
* Users of the Organizations will be simply called "Users", and all actions related to them will refer to them as such (eg. "Add user", "Delete user")
* Current logged-in Admin user of the API Provider organization will be able to:
  * manage its organization
  * manage its personal user account, which we will simply call "Account" instead of "Personal" (ie. a related action could be "Edit account details")
  * manage its organization users, which we will simply call "Users" instead of "Team members", and all actions related to them will refer to them as such (eg. "Add user", "Delete user")

I see two scenarios where this terminology could be confusing ([visual schema](https://app.lucidchart.com/invitations/accept/1d56541c-bcbc-4fc2-9802-2bc938e653b4)):
* For an Admin user of an API Provider Organization it could be confusing to find the same terminology used to refer to different concepts:
  * "Organization" will both refer to her own API Provider organization and to her customers' API Consumer organizations
  * Same for the term "Users"
  * I think this scenario invites us to differentiate terms:
    * one to refer to the API Provider organization, one to refer to the API Consumer organizations
    * one to refer to the API Provider organization's users, one to refer to the API Consumer organizations' users
* For an Admin who is the only user of an Organization, her personal Account and the Organization would practically mean the same thing (despite technically being two separate things)
  * In this scenario referring to the organization as something like a "Company" would recall concepts familiar also to one-people-ventures (ie. the company is a separate entity from the physical person)

To conclude, I can't seem to come up with a set of terms that would satisfy the meaning of the concepts and the contexts of use at the same time (eg. it would be strange to call an API consumer organization a company if they are internal to the same API provider company). Any feedback and/or input in this sense is very much welcome.

portafly/src/i18n/locales/en/accounts.yml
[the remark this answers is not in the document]
porueesq: @cntlsn @damianpm 
We need an alternative to describe that the last column contains all the possible actions.This is a matter of accessibility. 
A screen reader goes from left to right, and if it finds an empty cell in the header row, it won't report anything. Then, when the reader moves along the next rows, it will find the available actions, hence creating confusion to the user with the reader.
Some alternatives include, but not limited to: 

- an **Actions** header 
- a tooltip describing that the column contains actions to be applied to the relevant row.

Reference: [Accessibility principles](https://www.w3.org/WAI/fundamentals/accessibility-principles/#alternatives)

portafly/src/i18n/locales/en/accounts.yml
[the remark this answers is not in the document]
porueesq: @cntlsn @damianpm Two columns might require the popover with info about the current time zone and info on how to change it: Sent and Accepted.
On one hand, adding a popover for each column implies duplication. I'd go with a message on top of the table that can be dismissed by the user. WDYT?

portafly/src/i18n/locales/en/accounts.yml
porueesq: @cntlsn Can you please update the mockups with this new naming arrangement so we have a clear visibility of the changes?

portafly/src/i18n/locales/en/accounts.yml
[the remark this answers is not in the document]
damianpm: In [patternfly tables](https://www.patternfly.org/v4/documentation/react/components/table#actions) I don't see any option of adding a header or a tooltip for the Actions column.
Maybe we can add a header but hide it with css, but maybe is not a best practice...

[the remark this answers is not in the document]
cntlsn: @porueesq Thank you for raising this, I understand your point.
About the "Actions" header, I think this should be discussed with the PatternFly team. Please refer to PF [design guidelines](https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/lists-and-tables) and [documentation](https://www.patternfly.org/v4/documentation/react/components/table) for tables: no "Actions" column header is ever mentioned, albeit actions and kebab menus being present in some of the examples.
I'm not saying it is wrong, I'm just saying that this change would have consequences in some scenarios. For instance, please have a look at [this mockup](https://marvelapp.com/ab13h96/screen/67417389) where I have added the "Actions" column header after your suggestion. Since both a button and a kebab menu are included in each row, shall we have two "Actions" column headers? It seems a bit too redundant.
My suggestion would be to include an "Actions" column header for columns that include buttons, but not for the ones that include kebab icons.

Sure we can – and probably should – add a tooltip for kebab icons, but this would only partially solve the problem, as tooltips can't be triggered on touchscreen devices. Unless screenreaders will be able to detect tooltips on touchscreen devices, which could be unlikely according to:
* https://inclusive-components.design/tooltips-toggletips/
* https://ux.stackexchange.com/questions/71950/help-tooltips-and-screen-readers

@damianpm WDYT?

portafly/src/i18n/locales/en/accounts.yml
cntlsn: @porueesq This terminology change would have consequences on pages for which mockups haven't been designed yet. Most affected pages:
**Account settings**
* Sidebar navigation
* Overview page
* Edit account details page
* Personal details page
* Users Listing / Invitations page

**Audience**
* Sidebar navigation
* Accounts listing page
* Create new account page
* Account dashboard page
* Edit account details page
* Account applications list page

portafly/src/i18n/locales/en/accounts.yml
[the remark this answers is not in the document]
porueesq: @cntlsn Yes, to fulfil the guidelines and keep simplicity, I'm with your suggestion:
> My suggestion would be to include an "Actions" column header for columns that include buttons, but not for the ones that include kebab icons.

@damianpm Is this also fine with you?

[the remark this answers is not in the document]
damianpm: I agree we need to discuss this with PF team, in the meantime we can keep it without the additional header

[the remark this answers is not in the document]
porueesq: Adding this thread to the agenda of our next standup meeting, to agree on action points.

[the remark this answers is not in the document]
porueesq: As a temporary measure, we'll stick to the suggestion proposed by @cntlsn
> My suggestion would be to include an "Actions" column header for columns that include buttons, but not for the ones that include kebab icons.

portafly/src/i18n/locales/en/accounts.yml
porueesq: Marking this conversation as resolved, although there might be future changes due to ongoing discussions about UI terminology.

PR #1873, opened 2020-05-11
Fix CSRF vulnerability in Login Dev Portal

Martouta:
Fixes [THREESCALE-2922](https://issues.redhat.com/browse/THREESCALE-2922)
For reference: https://github.com/3scale/system/commit/504d8a75d15e6dbf12a235c0f673d2fa6366d795
But most of the work was already done before this PR 😄

### DIFF
--- a/app/controllers/concerns/error_handling.rb
+++ b/app/controllers/concerns/error_handling.rb
@@ -27,6 +27,11 @@ module ErrorHandling
       handle_error(exception, :forbidden)
     end
 
+    def handle_forgery_protection(exception)
+      System::ErrorReporting.report_error(exception)
+      handle_error(exception, :forbidden)
+    end
+
     def log_error(exception, status)
       logger.error "Handling Exception: #{exception} with status #{status}"
       logger.debug exception.backtrace.join("\n") if exception&.backtrace
--- a/features/support/capybara.rb
+++ b/features/support/capybara.rb
@@ -36,6 +36,10 @@ Before '@chrome' do
   Capybara.current_driver = :chrome
 end
 
+Around '@security' do |scenario, block|
+  with_forgery_protection(&block)
+end
+
 # monkeypatch to fix
 # not opened for reading (IOError)
 # /cucumber-1.3.20/lib/cucumber/formatter/interceptor.rb:33:in `each'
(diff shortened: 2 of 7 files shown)


features/support/capybara.rb
```diff
@@ -36,6 +36,14 @@
   Capybara.current_driver = :chrome
 end
 
+Before '@security' do
```
josemigallas, line 39: Is having the same before hook twice intentional?

Martouta: Wow, you are right! 😱 

image

image

Thanks for the catch! 🙏

Martouta: hahahahaha I see 😂 Nooope, I wanted to do "Before" and "After", but I see that I do not need any of them 😳

Martouta: Good catch! 😳 🙈

features/support/capybara.rb
```diff
@@ -36,6 +36,10 @@
   Capybara.current_driver = :chrome
 end
 
+Around '@secure' do |scenario, block|
```
guicassolato, line 39:
```suggestion
Around '@security' do |scenario, block|
```

?

Martouta: :see_no_evil:

app/controllers/concerns/error_handling.rb
```diff
@@ -27,6 +28,11 @@ def handle_access_denied(exception)
       handle_error(exception, :forbidden)
     end
 
+    def handle_forgery_protection(exception)
+      System::ErrorReporting.report_error(exception)
+      handle_error(exception, :forbidden)
```
Martouta, line 33: Forbidden because, as @thomasmaas told me: https://stackoverflow.com/questions/23478370/what-response-should-be-sent-back-a-when-cross-site-request-forgery-csrf-is-de

thomasmaas: although it's not an accepted answer…I think it makes sense.

Martouta:
Tests fixed (and explained in https://github.com/3scale/porta/pull/1878). Tested both together in https://github.com/3scale/porta/pull/1881

PR #1820, opened 2020-04-16
Implements Accounts Listing

josemigallas:
**What this PR does / why we need it**:

Implements "Accounts > Listing" page for Portafly.

**Which issue(s) this PR fixes** 

[THREESCALE-4732: Create component for Account listing page](https://issues.redhat.com/browse/THREESCALE-4732)

**Verification steps** 
* To work with mock data, open a separate terminal run `yarn mock-server`.
* To work with real data from preview or your local cluster, create an access token there and use it to log in Portafly.

Once the data source is sorted out, simply run `yarn start`

**Future steps** 
This PR focuses on the Table logic, there are still a bunch of tasks and improvements related to this page (see [Epic: THREESCALE-4724](https://issues.redhat.com/browse/THREESCALE-4724)).

PR #1876, opened 2020-05-13
Action for Resetting Pagination

josemigallas:

When being in a distant page (say the last one) and adding filters, the collection is updated but not the Pagination state. It is necessary to reset back to the initial state in order to show the filtered results, for example:
```ts
useEffect(resetPagination, [filters])
```

**Verification steps**:
It can be teste and verified in #1820

### DIFF
--- a/portafly/src/components/data-list/reducers/PaginationReducer.ts
+++ b/portafly/src/components/data-list/reducers/PaginationReducer.ts
@@ -12,6 +12,13 @@ export type PaginationState = {
   endIdx?: number
 }
 
+const defaultPagination: PaginationState = {
+  page: 1,
+  perPage: 10,
+  startIdx: 0,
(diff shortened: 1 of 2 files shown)


portafly/src/components/data-list/reducers/PaginationReducer.ts
```diff
@@ -31,7 +38,8 @@ const usePagination = ({ state, dispatch }: IUsePagination) => ({
   perPage: state.pagination.perPage,
   startIdx: state.pagination.startIdx,
   endIdx: state.pagination.endIdx,
+  resetPagination: () => dispatch({ type: 'SET_PAGINATION', payload: defaultPagination }),
```
didierofrivia, line 41: we might start defining action names in `const`

PR #1879, opened 2020-05-14
[Portafly] Porta-wiring refactor

didierofrivia:

This is refactor that will ease the use of http async requests and managing responses from Porta API

We were using hooks inside the data layer functions that was an anti pattern. Also redefining `Response` type and using async hooks inside components. 

**Verification steps** 
Launch the app with `yarn start` and porta Rails server,  login with a valid token and navigate to dev accounts, it should render a list of dev accounts. It varies depending on the data in your local porta

### DIFF
--- a/portafly/src/auth/AuthContext.tsx
+++ b/portafly/src/auth/AuthContext.tsx
@@ -1,6 +1,5 @@
 import React, { createContext, useContext, useState } from 'react'
-
-type AuthToken = string | null
+import { getToken, setToken, AuthToken } from 'utils'
 
 interface IAuthContext {
   authToken: AuthToken,
@@ -13,15 +12,15 @@ const AuthContext = createContext<IAuthContext>({
 })
 
 const AuthProvider: React.FunctionComponent = ({ children }) => {
-  const existingToken = localStorage.getItem('token')
-  const [authToken, setAuthToken] = useState(existingToken && JSON.parse(existingToken))
+  const existingToken = getToken()
+  const [authToken, setAuthToken] = useState(existingToken)
 
-  const setToken = (token: AuthToken) => {
-    localStorage.setItem('token', JSON.stringify(token))
+  const saveToken = (token: AuthToken) => {
+    setToken(token)
     setAuthToken(token)
   }
   return (
-    <AuthContext.Provider value={{ authToken, setAuthToken: setToken }}>
+    <AuthContext.Provider value={{ authToken, setAuthToken: saveToken }}>
       {children}
     </AuthContext.Provider>
   )
--- a/portafly/src/pages/DeveloperAccounts.tsx
+++ b/portafly/src/pages/DeveloperAccounts.tsx
@@ -1,4 +1,5 @@
 import React from 'react'
+import { useAsync } from 'react-async'
 import { useDocumentTitle, Loading, DeveloperAccountsTable } from 'components'
 import {
   Alert,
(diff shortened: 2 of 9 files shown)


portafly/src/auth/AuthContext.tsx
```diff
@@ -13,15 +14,15 @@ const AuthContext = createContext<IAuthContext>({
 })
 
 const AuthProvider: React.FunctionComponent = ({ children }) => {
-  const existingToken = localStorage.getItem('token')
-  const [authToken, setAuthToken] = useState(existingToken && JSON.parse(existingToken))
+  const existingToken = getToken()
+  const [authToken, setAuthToken] = useState(existingToken && existingToken)
```
josemigallas, line 18:
```suggestion
  const [authToken, setAuthToken] = useState(existingToken)
```

portafly/src/pages/DeveloperAccounts.tsx
```diff
@@ -7,14 +8,14 @@ import {
   TextContent,
   Text
 } from '@patternfly/react-core'
-import { useGetDeveloperAccounts } from 'dal/accounts'
+import { getDeveloperAccounts } from 'dal/accounts'
 import { useTranslation } from 'i18n/useTranslation'
 
 const DeveloperAccounts: React.FunctionComponent = () => {
+  const { data: accounts, error, isPending } = useAsync(getDeveloperAccounts)
```
josemigallas, line 15: 👏

portafly/src/utils/index.ts
```diff
@@ -19,4 +20,37 @@ const combineReducers = (reducerSlices: Record<string, any>) => (
     prevState
   ))
 
-export { createReducer, combineReducers }
+// # Auth
+// Token is stored in JSON and not string since in the future we will be working with JWT
+const getToken = () => JSON.parse(<string>localStorage.getItem('token'))
```
josemigallas, line 25: Cast is redundant here. Also I would define `AuthToken` in this file instead:
```suggestion
const getToken: AuthToken = () => JSON.parse(localStorage.getItem('token'))
```

portafly/src/utils/index.ts
```diff
@@ -19,4 +20,37 @@ const combineReducers = (reducerSlices: Record<string, any>) => (
...
+const setToken = (token: string | null) => localStorage.setItem('token', JSON.stringify(token))
```
josemigallas, line 26: I should not save `"null"` but call `localStorage.removeItem` instead. The input should only be `JWT` only (or at this point, Object I guess)
```suggestion
const setToken = (token: AuthToken) => localStorage.setItem('token', JSON.stringify(token))
```

portafly/src/auth/AuthContext.tsx
```diff
@@ -1,4 +1,5 @@
 import React, { createContext, useContext, useState } from 'react'
+import { getToken, setToken } from 'utils'
 
 type AuthToken = string | null
```
josemigallas, line 4: If the AuthToken is going to be the result of `JSON.parse` we should change this to object (eventually JWT):
```suggestion
type AuthToken = Object | null
```

portafly/src/utils/index.ts
didierofrivia: I've changed the fn signature, but still needs the cast since `JSON.parse` can't parse `null`. A more robust way would be having a fn like in Porta UI. `SafeParse`

josemigallas: Ah you're right, I don't know why I thought that.

A thought tho: I think this kind of casting is deceitful, because it is used only to make TS happy. `getItem` _can return_ `null` and therefore it is a case to be considered. It's not because the code won't work in this particular case (it is 100% good) but because we're being dishonest with TS and we should always avoid that.
```js
// we could return an empty JWT / Object when the token is null
JSON.parse(localStorage.getItem('token') || '{}')

// or we could check
item = localStorage.getItem('token')
return item ? JSON.parse(item) : {}

// or we could do a "real" casting
JSON.parse(String(localStorage.getItem('token'))
```

didierofrivia: Yeah, it's quite a double edge sword... actually more than that, since sometimes we just can't be sure

didierofrivia: Still, we are not sure how the flow will be, so maybe I'll go for checking or do a "real casting"

PR #1882, opened 2020-05-15
[portafly] BulkSelectorWidget

josemigallas:
Adds a widget and introduces selections and table data into DataListContext

bulk

### DIFF
--- a/portafly/src/components/data-list/BulkSelectorWidget.tsx
+++ b/portafly/src/components/data-list/BulkSelectorWidget.tsx
@@ -9,13 +9,14 @@ import { useTranslation } from 'i18n/useTranslation'
 import { useDataListTable, useDataListPagination } from 'components/data-list'
 
 interface Props {
-  filteredRows: any[]
+  // TODO: try make it independent from table data
+  filteredRows: { id: number }[]
 }
 
 const BulkSelectorWidget: React.FunctionComponent<Props> = ({
   filteredRows
 }) => {
-  const { t } = useTranslation('accounts')
+  const { t } = useTranslation('shared')
   const { selectPage, selectAll, selectedRows } = useDataListTable()
   const { startIdx, endIdx } = useDataListPagination()
 
@@ -32,26 +33,18 @@ const BulkSelectorWidget: React.FunctionComponent<Props> = ({
   // eslint-disable-next-line no-nested-ternary
   const isChecked = (selectedCount === 0) ? false
     : (selectedCount === allCount) ? true
+    // For the indeterminate state, PF React needs to be passed null
       : null
-  // OR
-  // let isChecked: boolean | null
-  // if (selectedCount === 0) {
-  //   isChecked = false
-  // } else if (selectedCount === allCount) {
-  //   isChecked = true
-  // } else {
-  //   isChecked = null
-  // }
 
   const dropdownItems = [
     <DropdownItem key="0" onClick={() => selectAll(false)}>
-      {t('accounts_table.data_toolbar.bulk_selector.none')}
+      {t('bulk_selector.none')}
     </DropdownItem>,
     <DropdownItem key="1" onClick={() => selectPage(visibleRows)}>
-      {t('accounts_table.data_toolbar.bulk_selector.page', { count: pageCount })}
+      {t('bulk_selector.page', { count: pageCount })}
     </DropdownItem>,
     <DropdownItem key="2" onClick={() => selectAll(true, filteredRows)}>
-      {t('accounts_table.data_toolbar.bulk_selector.all', { count: allCount })}
+      {t('bulk_selector.all', { count: allCount })}
     </DropdownItem>
   ]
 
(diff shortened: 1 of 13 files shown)


portafly/src/components/data-list/BulkSelectorWidget.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const [isOpen, setIsOpen] = useState(false)
+
+  const selectedCount = selectedRows.length
+
+  const { startIdx, endIdx } = useDataListPagination()
+  const visibleRows = filteredRows.slice(startIdx, endIdx)
+  const pageCount = visibleRows.length
+  const allCount = filteredRows.length
+
+  const onSelect = () => {
```
didierofrivia, line 30:
```suggestion
  const onSelect = () => setIsOpen((current) => !current)
```

portafly/src/components/data-list/BulkSelectorWidget.tsx
```diff
@@ -0,0 +1,81 @@
...
+    setIsOpen((current) => !current)
+  }
+
+  // FIXME: Setting isChecked null creates an error in the console for uncontrolled React prop
```
didierofrivia, line 34: Is it mandatory that the prop of `DropdownToggleCheckbox` have 3 possible values? (true, false and null)

portafly/src/components/data-list/reducers/TableReducer.ts
```diff
@@ -0,0 +1,84 @@
...
+    const visibleRows = action.payload as any[]
+    const newRows = state.initialRows.map((r) => {
+      const isRowVisible = visibleRows.find((vR) => vR.id === r.id)
+      return { ...r, selected: isRowVisible }
+    })
+    // OR
+    // const newRows = state.initialRows.map((r) => ({
+    //   ...r,
+    //   selected: visibleRows.find((vR) => vR.id === r.id)
+    // }))
```
didierofrivia, line 36: I personally fancy more this option

portafly/src/components/data-list/BulkSelectorWidget.tsx
```diff
@@ -0,0 +1,81 @@
...
+  let isChecked: boolean | null
+  if (selectedCount === 0) {
+    isChecked = false
+  } else if (selectedCount === allCount) {
+    isChecked = true
+  } else {
+    isChecked = null
+  }
```
didierofrivia, line 42: Something like the following might be a bit better:

```suggestion
  const isChecked = (selectedCount === 0) ? false : (selectedCount === allCount) ? true : null 
```

portafly/src/components/data-list/BulkSelectorWidget.tsx
didierofrivia: What is puzzling is the fact that something called `isChecked` that suggest boolean type, can have a third one...  I know it's from PF, but something like `checkStatus` might've been more adequate.

josemigallas: > What is puzzling is the fact that something called isChecked that suggest boolean type
Indeed. I would have used a prop `indeterminate` instead.

We have to live with this I'm afraid.

portafly/src/components/data-list/reducers/TableReducer.ts
josemigallas: My subconscious probably left it for you 😂

portafly/src/components/data-list/BulkSelectorWidget.tsx
josemigallas: 🙄 🤣

josemigallas: There's an Eslint rule preventing us from nesting ternary operators... which I agree with, it can lead to errors or very hard to read sentences. Since the file is small enough it doesn't hurt to have some verbose if/else blocks, I think. On the other hand your suggestion is not so complex. Hopefully @damianpm can give us the casting vote.

didierofrivia: I know it's not super nice either, but it could be crafted even as a fn if you want...thing is, I'd rather have a nested ternary instead of a variable that changes depending on the result of a nested if-else.

portafly/src/components/data-list/BulkSelectorWidget.tsx
```diff
@@ -0,0 +1,83 @@
...
+import { useDataListTable, useDataListPagination } from 'components/data-list'
+
+interface Props {
+  filteredRows: any[]
+}
+
+const BulkSelectorWidget: React.FunctionComponent<Props> = ({
+  filteredRows
+}) => {
+  const { t } = useTranslation('accounts')
```
josemigallas, line 18: Move to shared translations

portafly/src/components/data-list/DataListContext.tsx
```diff
@@ -26,11 +37,29 @@ export interface IDataListContext {
 
 const DataListContext = createContext<IDataListContext>({ state: initialState, dispatch: () => {} })
 
-const DataListProvider: React.FunctionComponent = ({ children }) => {
+type Props = {
+  data: {
```
josemigallas, line 41:
```suggestion
  initialState: {
```

portafly/src/components/data-list/DataListContext.tsx
```diff
@@ -35,29 +34,25 @@ export interface IDataListContext {
...
 
 type Props = {
-  data: {
-    columns: DataListCol[]
-    rows: DataListRow[]
+  initialState: {
+    table: {
+      columns: DataListCol[]
+      rows: DataListRow[]
+    }
```
didierofrivia, line 44: Maybe directly:
```suggestion
type Props = {
    table: {
      columns: DataListCol[]
      rows: DataListRow[]
    },
    pagination: ... // future
    filters: ... // future
```

Or are we expecting something else than an initial state?

PR #1883, opened 2020-05-15
[Test] API Account endpoints fire webhook

Martouta:
Closes [THREESCALE-5207](https://issues.redhat.com/browse/THREESCALE-5207)

It was accidentally implemented for 2.8 when we added https://github.com/3scale/porta/pull/1546/commits/e2e44c797f41784565b8b0b424567b6cd6e4b81e
This PR is to add tests.

### Tasks
- [X] Test Create Buyers through API fires webhook
- [X] Test Update/Delete Buyers through API fires webhook

### DIFF
--- a/test/integration/admin/api/accounts_controller_test.rb
+++ b/test/integration/admin/api/accounts_controller_test.rb
@@ -123,6 +123,50 @@ class Admin::Api::AccountsControllerTest < ActionDispatch::IntegrationTest
     end
   end
 
+  class WebHooksTest < Admin::Api::AccountsControllerTest
+    disable_transactional_fixtures!
+
+    test 'update by access token fires webhooks' do
+      provider.settings.allow_web_hooks!
+      FactoryBot.create(:webhook, account: provider, account_updated_on: true, active: true)
+
+      assert_difference(WebHookWorker.jobs.method(:size)) do
+        put admin_api_account_path(buyer, format: :json), { monthly_billing_enabled: true, access_token: token.value }
+        assert_response :success
+      end
+    end
+
+    test 'update by provider key does not fire webhooks' do
+      provider.settings.allow_web_hooks!
+      FactoryBot.create(:webhook, account: provider, account_updated_on: true, active: true)
+
+      assert_no_difference(WebHookWorker.jobs.method(:size)) do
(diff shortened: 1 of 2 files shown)


mkudlej:
@Martouta  @secretsurf I don't think this is an integration test. This is functional test on unit test level. It uses directly functions from Porta code-base and it disables some features in 3scale. What deployment is used for running this test? Does it run on fully deployed 3scale(operator/template based)? How frequently will it run? Will it be tested with all supported configurations?
I think this test is probably enough to test code deployed in SaaS but it is not enough for closing related Jira issue for On Prem deployment.
Please consider to write tests like this in QE testsuite, like @eloycoto does. It is easy and QE doesn't need to test this again.

secretsurf:
@mkudlej, i'm confused. So QE would no longer write tests for issues done by engineering? i mean, if we could split some work and you guys write some code in engineering's codebase and engineering writes some code in your codebase, i'd totally get this. But we don't have a setup like that (even though we would probably all be happy to have it) and we have a bottleneck problem on the ruby side that is even worse than the QE bottleneck. Is this something that we can implement later on and share our respective short-handedness between teams? Because i don't think this can work right now, especially if the ruby team has to shoulder the entire burden.

mkudlej:
@secretsurf I undertand what you've written but I don't understand why @Martouta done this test when development team has a lot of work to do. The issue/change will be tested by QE anyway.

Martouta:
I didn't make the decision. But it is beneficial for us to have our own test suit of everything we do so we get 2 things out of it:
1) If any change breaks it, we realize right away before merging.
2) Documentation :) So every person can understand more easily what it is intended from the code.

Martouta:
At the same time, I do not think that it is worth it to write it for all APIs for us :sweat:

Martouta:
I am not strongly against closing this without merging either btw :smile:

Martouta:
> The issue/change will be tested by QE anyway.
According to that logic, we shouldn't test anything except for Billing xD

You should check https://www.youtube.com/watch?v=iSKCb1J7KLM
You are not gonna ever check all options by Integration tests at all XDD Besides, too much repetition and lack of resources.
So, as a general rule, it is not a bad idea that we try to keep our own healthy "unit" test suit as well.

PR #1884, opened 2020-05-15
Allow to configure custom list of currencies

Martouta:
Closes [THREESCALE-3076](https://issues.redhat.com/browse/THREESCALE-3076)
Depends on https://github.com/3scale/deploy/pull/662

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -72,6 +72,7 @@ C:\\nppdf32Log\\debuglog.txt
 /config/web_hooks.yml
 /config/rolling_updates.yml
 /config/paperclip.yml
+/config/currencies.yml
 /config/plan_rules.yml
 /config/service_discovery.yml
 /config/smtp.yml
--- a/app/lib/currencies_loader.rb
+++ b/app/lib/currencies_loader.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+module CurrenciesLoader
+  REQUIRED_CURRENCIES = {
+    'USD - American Dollar' => 'USD',
+    'EUR - Euro' => 'EUR'
+  }.freeze
+
+  module_function
+
+  def load_config
+    REQUIRED_CURRENCIES.reverse_merge(ThreeScale.config.currencies.stringify_keys)
+  end
+end
--- a/app/models/finance/billing_strategy.rb
+++ b/app/models/finance/billing_strategy.rb
@@ -15,18 +15,7 @@ class Finance::BillingStrategy < ApplicationRecord
 
   attr_reader :failed_buyers
 
-  CURRENCIES = {
-    'USD - American Dollar' => 'USD',
-    'EUR - Euro'=> 'EUR',
-    'GBP - British Pound' => 'GBP',
-    'NZD - New Zealand dollar' => 'NZD',
-    'CNY - Chinese Yuan Renminbi' => 'CNY',
-    'CAD - Canadian Dollar' => 'CAD',
-    'AUD - Australian Dollar' => 'AUD',
-    'JPY - Japanese Yen' => 'JPY',
-    'CHF - Swiss Franc' => 'CHF',
-    'SAR - Saudi Riyal' => 'SAR'
-  }.freeze
+  CURRENCIES = CurrenciesLoader.load_config.freeze
 
   belongs_to :account
   alias_attribute :provider, :account
@@ -34,7 +23,7 @@ class Finance::BillingStrategy < ApplicationRecord
   attr_protected :account_id, :tenant_id, :audit_ids
 
   accepts_nested_attributes_for :account
-  validates :currency, inclusion: { :in => CURRENCIES.values }
+  validates :currency, inclusion: { in: CURRENCIES.values, message: :invalid }
 
   # TODO: uncomment when factories are fixed
   # validates_presence_of :account
--- a/config/application.rb
+++ b/config/application.rb
@@ -186,6 +186,9 @@ module System
     config.three_scale.plan_rules = ActiveSupport::OrderedOptions.new
     config.three_scale.plan_rules.merge!(try_config_for(:plan_rules) || {})
 
+    config.three_scale.currencies = ActiveSupport::OrderedOptions.new
+    config.three_scale.currencies.merge!(try_config_for(:currencies) || {})
+
     config.three_scale.features = ActiveSupport::OrderedOptions.new
     config.three_scale.features.merge!(try_config_for(:features) || {})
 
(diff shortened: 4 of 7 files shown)


app/lib/currencies_loader.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module CurrenciesLoader
+  DEFAULT_CURRENCIES = {
+    'USD - American Dollar' => 'USD'
```
Martouta, line 5: https://github.com/3scale/porta/blob/204d1f00c9885dbcc7bfa21301f427cfe7413005/db/schema.rb#L264

app/lib/currencies_loader.rb
```diff
@@ -0,0 +1,13 @@
...
+  }.freeze
+
+  module_function
+
+  def load_config
+    ThreeScale.config.currencies || DEFAULT_CURRENCIES
```
Martouta, line 11: This means that it only goes to the default if the currencies config is not set, otherwise, if they set the config and the dollar is not in their list, it won't be there... not sure if it is a good idea like this or not 🤷‍♀️

config/examples/currencies.yml
```diff
@@ -0,0 +1,13 @@
+currencies: &default
+ USD - American Dollar: 'USD'
+ EUR - Euro: 'EUR'
```
Martouta, line 3: For this:
https://github.com/3scale/porta/blob/46ac8701c641320d608777761f7bb9c793bfd640/test/factories/billing_strategy.rb#L3

Not sure if I should keep this example or remove it and change the factory to use USD.

app/lib/currencies_loader.rb
```diff
@@ -0,0 +1,13 @@
...
+
+module CurrenciesLoader
+  DEFAULT_CURRENCIES = {
+    'USD - American Dollar' => 'USD'
+  }.freeze
+
+  module_function
+
+  def load_config
+    ThreeScale.config.currencies.presence || DEFAULT_CURRENCIES
```
Martouta, line 11: This means that it only goes to the default if the currencies config is not set, otherwise, if they set the config and the dollar is not in their list, it won't be there... not sure if it is a good idea like this or not 🤷‍♀️

guicassolato: Plus https://github.com/3scale/porta/blob/46ac8701c641320d608777761f7bb9c793bfd640/app/models/account.rb#L378 and https://github.com/3scale/porta/blob/46ac8701c641320d608777761f7bb9c793bfd640/app/models/account/provider_methods.rb#L171

Martouta: OMG! Right! Thanks!! 👏

guicassolato: This may be helpful as well, @Martouta 
https://github.com/3scale/porta/blob/46ac8701c641320d608777761f7bb9c793bfd640/app/controllers/finance/provider/settings_controller.rb#L16-L18

guicassolato: Also, not having USD in the list may not be an option now: https://github.com/3scale/porta/blob/46ac8701c641320d608777761f7bb9c793bfd640/app/lib/switches.rb#L233

Martouta: Right! I will make 'EUR' and 'USD' to be ALWAYS in the list, no matter it is in the config or not 💪

app/lib/currencies_loader.rb
guicassolato: A funny one for you
https://github.com/3scale/porta/blob/46ac8701c641320d608777761f7bb9c793bfd640/db/schema.rb#L899

app/lib/currencies_loader.rb
guicassolato: Yeah. I think that would be the easiest for now.

And please make sure to use the `Account::ProviderMethods.DEFAULT_CURRENCY` in https://github.com/3scale/porta/blob/46ac8701c641320d608777761f7bb9c793bfd640/app/lib/switches.rb#L233

app/lib/currencies_loader.rb
Martouta: This is fine GIF

app/lib/currencies_loader.rb
Martouta: Right 😂 👍

guicassolato: Never mind about `Account::ProviderMethods::DEFAULT_CURRENCY`. Different currencies. 🤯

app/lib/currencies_loader.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module CurrenciesLoader
+  DEFAULT_CURRENCIES = {
```
guicassolato, line 4:
```suggestion
  REQUIRED_CURRENCIES = {
```

Martouta: Right, love it! ❤️ 👏

.gitignore
```diff
@@ -72,6 +72,7 @@ C:\\nppdf32Log\\debuglog.txt
 /config/web_hooks.yml
 /config/rolling_updates.yml
 /config/paperclip.yml
+/config/currencies.yml
```
hallelujah, line 75: There is nothing in config/examples/currencies.yml?
Would be good to add what you removed there.

Martouta: @hallelujah Lacking intentionally 😝 the only ones that we require are USD and EUR and they are here... examples is just an example, no? why do we need an example for this?

And about this comment:
image

I disagree that we will lose them, for this: https://github.com/3scale/deploy/pull/662

hallelujah: But on-premises needs it anyway or is it an intentional breaking change? Like next release upgrade will make billing fail on a unknown currency?

Martouta: We spoke in private and I will add them only to `openshift/system/config`

app/models/finance/billing_strategy.rb
```diff
@@ -15,26 +15,15 @@ class << self
...
-    'GBP - British Pound' => 'GBP',
-    'NZD - New Zealand dollar' => 'NZD',
-    'CNY - Chinese Yuan Renminbi' => 'CNY',
-    'CAD - Canadian Dollar' => 'CAD',
-    'AUD - Australian Dollar' => 'AUD',
-    'JPY - Japanese Yen' => 'JPY',
-    'CHF - Swiss Franc' => 'CHF',
-    'SAR - Saudi Riyal' => 'SAR'
-  }.freeze
+  CURRENCIES = CurrenciesLoader.load_config
```
hallelujah, line 18: Maybe freeze it? So it cannot be modified on runtime

Martouta: Right, done 😎

PR #1886, opened 2020-05-18
Destroy a default service through a rake task

Martouta:
Tested in Preview 
`RAILS_ENV=preview bundle exec rake "services:destroy_default_service[account_id_here service_id_here]"`

We were requested to destroy a default service in [THREESCALE-5238](https://issues.redhat.com/browse/THREESCALE-5238)
That is a very old and deprecated flag. We do not add it anymore in Porta, but in Apisonator it is mandatory and they usually add it for the first service of the provider.
We did all this in the rails console for it is dangerous and we should have a rake task instead.

### DIFF
--- a/app/lib/backend/model_extensions/service.rb
+++ b/app/lib/backend/model_extensions/service.rb
@@ -45,6 +45,7 @@ module Backend
 
       def make_default_backend_service
         ThreeScale::Core::Service.make_default(backend_id)
+        account.update!({default_service_id: id}, without_protection: true)
       end
     end
   end
--- a/lib/tasks/services.rake
+++ b/lib/tasks/services.rake
@@ -12,4 +12,16 @@ namespace :services do
   task :destroy_marked_as_deleted => :environment do
     DestroyAllDeletedObjectsWorker.perform_later(Service.to_s)
   end
+
+  desc 'Destroy a service in the Background as long as the provider has at least 1 more service. Whether it is a default one or not'
+  task :destroy_service, %i[account_id service_id] => [:environment] do |_task, args|
+    account = Account.providers.find(args[:account_id])
+    service_id = args[:service_id]
+    service = account.services.find(service_id) # So it will raise an error if this service does not belong to this account :)
+    if service.default?
+      new_default_service = account.services.where.not(id: service_id).first! # It will raise an error if it does not have another service
+      new_default_service.send(:make_default_backend_service)
+    end
+    service.reload.mark_as_deleted!
+  end
 end
--- a/app/models/service.rb
+++ b/app/models/service.rb
@@ -276,12 +276,6 @@ class Service < ApplicationRecord
     application_plans.published
   end
 
-  # use this method to destroy the default service with all callbacks firing
-  def destroy_default
-    @destroy_default_allowed = true
-    destroy
-  end
-
   def stop_destroy_if_last_or_default
     return if destroyable?
     errors.add :base, 'This service cannot be removed'
@@ -553,8 +547,7 @@ class Service < ApplicationRecord
   end
 
   def destroyable?
-    return true if destroyed_by_association
-    !default_or_last? || @destroy_default_allowed
+    destroyed_by_association || !default_or_last?
   end
 
   def create_default_proxy
--- a/test/unit/account_test.rb
+++ b/test/unit/account_test.rb
@@ -99,21 +99,6 @@ class AccountTest < ActiveSupport::TestCase
     assert_raise(ActiveRecord::RecordNotFound) { metric.reload }
   end
 
-  def test_default_service_id
-    service = FactoryBot.create(:simple_service)
-    account = FactoryBot.create(:simple_account, services: [service], default_service_id: service.id)
-
-    assert service.default?
-    assert_equal service.id, account.default_service_id
-
-    service.destroy_default
-
-    account.reload
-
-    assert_raises(ActiveRecord::RecordNotFound) { service.reload }
-    assert_nil account.default_service_id
-  end
-
   test '#trashed_messages' do
     account  = FactoryBot.build_stubbed(:simple_account)
     other    = FactoryBot.build_stubbed(:simple_account)
--- a/test/unit/service_test.rb
+++ b/test/unit/service_test.rb
@@ -99,29 +99,6 @@ class ServiceTest < ActiveSupport::TestCase
     assert_raise(ActiveRecord::RecordNotFound) { service.reload }
   end
 
-  def test_update_account_default_service
-    default_service = FactoryBot.create(:simple_service)
-    other_service   = FactoryBot.create(:simple_service)
-    account         = FactoryBot.create(:simple_account,
-                                         services: [default_service, other_service],
-                                         default_service_id: default_service.id)
-
-    assert_equal account.default_service_id, default_service.id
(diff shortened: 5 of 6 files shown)


lib/tasks/services.rake
```diff
@@ -12,4 +12,14 @@ namespace :services do
...
     DestroyAllDeletedObjectsWorker.perform_later(Service.to_s)
   end
+
+  task :destroy_default_service, %i[account_id service_id] => [:environment] do |_task, args|
+    account = Account.providers.find(args[:account_id])
+    service_id = args[:service_id]
+    service = account.services.find(service_id) # So it will raise an error if this service does not belong to this account :)
+    new_default_service = account.services.where.not(id: service_id).select(:id).first! # It will raise an error if it does not have another service
+    account.update!({default_service_id: nil}, without_protection: true) # We removed this flag and should never add it anymore ;) we only keep it for the old ones who already have it
+    ThreeScale::Core::Service.save!(id: new_default_service.id, default_service: true)
```
guicassolato, line 22:
```suggestion
    ThreeScale::Core::Service.make_default(new_default_service.id)
```

Martouta: wow, true! good catch! 👏

lib/tasks/services.rake
```diff
@@ -12,4 +12,14 @@ namespace :services do
   task :destroy_marked_as_deleted => :environment do
     DestroyAllDeletedObjectsWorker.perform_later(Service.to_s)
   end
+
+  task :destroy_default_service, %i[account_id service_id] => [:environment] do |_task, args|
```
Martouta, line 16: Missing description

duduribeiro:
looks good 😛 

kidding.. miss u folks 💔

Martouta:
@duduribeiro we miss you too! 😭 💔 
I hope that you are doing great! 💪

lib/tasks/services.rake
```diff
@@ -12,4 +12,15 @@ namespace :services do
...
     DestroyAllDeletedObjectsWorker.perform_later(Service.to_s)
   end
+
+  desc 'Destroy a default service.'
+  task :destroy_default_service, %i[account_id service_id] => [:environment] do |_task, args|
+    account = Account.providers.find(args[:account_id])
+    service_id = args[:service_id]
+    service = account.services.find(service_id) # So it will raise an error if this service does not belong to this account :)
+    new_default_service = account.services.where.not(id: service_id).select(:id).first! # It will raise an error if it does not have another service
+    account.update!({default_service_id: nil}, without_protection: true) # We removed this flag and should never add it anymore ;) we only keep it for the old ones who already have it
```
Martouta, line 22: The opinion of each and every person of @3scale/system-ruby is pure gold for this 😉

macejmic: what about this compromise:

```
default_service_id = account.default_service_id.present? ? new_default_service.id  : nil
account.update!({default_service_id: default_service_id}, without_protection: true)
```

so we wouldn't add default_service_id for new accounts and keep it for the old ones.

Martouta: mmmm the goal of this task is to be called only if it is a default service :thinking: but you are right that it apparently can be called if it is not default, but then, as it is not what it is for, it would mess up badly even in Apisonator XD so... oki, I will change it to work for both default and non-default services :wink:

lib/tasks/services.rake
```diff
@@ -12,4 +12,17 @@ namespace :services do
...
     DestroyAllDeletedObjectsWorker.perform_later(Service.to_s)
   end
+
+  desc 'Destroy a service in the Background as long as the provider has at least 1 more service. Whether it is a default one or not'
+  task :destroy_default_service, %i[account_id service_id] => [:environment] do |_task, args|
+    account = Account.providers.find(args[:account_id])
+    service_id = args[:service_id]
+    service = account.services.find(service_id) # So it will raise an error if this service does not belong to this account :)
+    if service.default?
+      new_default_service = account.services.where.not(id: service_id).select(:id).first! # It will raise an error if it does not have another service
```
guicassolato, line 22:
```suggestion
      new_default_service = account.services.where.not(id: service_id).first! # It will raise an error if it does not have another service
```

`.select(:id)` is ok here because you know `Service#backend_id` only depends on this attribute of the model, but this is breaking the abstraction and we should not limit this way.

app/lib/backend/model_extensions/service.rb
```diff
@@ -40,12 +40,6 @@ def delete_backend_service
         System::ErrorReporting.report_error e
         raise e
       end
-
-      private
-
-      def make_default_backend_service
```
guicassolato, line 46: The task clearly shows that we still need this, so maybe making it public instead?

On the other hand, I guess you decided to remove it to make it clear that making a service the "default" service is not something related to porta per se anymore, right?

Martouta: > I guess you decided to remove it to make it clear that making a service the "default" service is not something related to porta per se anymore

Exactly :smile:

lib/tasks/services.rake
Martouta: Right :+1:

app/lib/backend/model_extensions/service.rb
guicassolato: I see. I think we should keep it for the moment though. As I said, make it public and add a comment pointing out that we may need this because until we no longer support `provider_key` as an option to authenticate.

Martouta: okidoki :+1:

guicassolato: In fact, making a service the default in backend/apisonator should also be reflected in `Account.default_service_id`. I know we are trying to no longer support default services, but apisonator still need them and we may be breaking consistency if we don't follow for now.

lib/tasks/services.rake
```diff
@@ -12,4 +12,17 @@ namespace :services do
...
+
+  desc 'Destroy a service in the Background as long as the provider has at least 1 more service. Whether it is a default one or not'
+  task :destroy_default_service, %i[account_id service_id] => [:environment] do |_task, args|
+    account = Account.providers.find(args[:account_id])
+    service_id = args[:service_id]
+    service = account.services.find(service_id) # So it will raise an error if this service does not belong to this account :)
+    if service.default?
+      new_default_service = account.services.where.not(id: service_id).first! # It will raise an error if it does not have another service
+      account.update!({default_service_id: nil}, without_protection: true)
+      ThreeScale::Core::Service.make_default(new_default_service.backend_id)
```
guicassolato, line 24: https://github.com/3scale/porta/pull/1886/files#r428142874

I think these 2 lines should be abstracted to
```ruby
new_default_service.make_default_backend_service
```

app/lib/backend/model_extensions/service.rb
Martouta: Ok but we also need to be careful to ONLY do it for those providers that already have a default service and not end up introducing it for new ones... for that I wonder if I should make as you said in https://github.com/3scale/porta/pull/1886#discussion_r428144393 but making the method private to only access it from here 🤔

guicassolato: Good point. Let's keep it private then.

PR #1889, opened 2020-05-19
Add support for node-sass

josemigallas:
Portafly doesn't have `node-sass` and the build fails locally.

**Verification**
* removes all node_modules from `porta` and `portafly`
* `yarn && yarn start`

### DIFF
--- a/portafly/.stylelintrc.json
+++ b/portafly/.stylelintrc.json
@@ -1,3 +1,6 @@
 {
-  "extends": "stylelint-config-sass-guidelines"
+  "extends": "stylelint-config-sass-guidelines",
+  "rules": {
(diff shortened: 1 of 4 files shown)


portafly/.stylelintrc.json
```diff
@@ -1,3 +1,6 @@
 {
-  "extends": "stylelint-config-sass-guidelines"
+  "extends": "stylelint-config-sass-guidelines",
+  "rules": {
+    "selector-class-pattern": null
```
josemigallas, line 4: This is to fix the error: https://github.com/3scale/porta/pull/1889/files#diff-94b2cd13981c12108120c7f72bdbe5c3L1

PR #1893, opened 2020-05-19
Highlight SSO Integrations in vertical navigation

srijitamukherjee:
Closes [THREESCALE-3548](https://issues.redhat.com/browse/THREESCALE-3548)

**Verification steps** 

The title: SSO Integration, in the vertical navigation, under Developer Portal -> Setting-> SSO Integrations, should be highlight when checked.

### DIFF
--- a/app/controllers/provider/admin/authentication_providers_controller.rb
+++ b/app/controllers/provider/admin/authentication_providers_controller.rb
@@ -3,7 +3,7 @@
 class Provider::Admin::AuthenticationProvidersController < FrontendController
   before_action :authorize_settings
 
-  activate_menu :audience, :cms, :oauth2
+  activate_menu :audience, :cms, :sso_integrations
   sublayout 'sites/developer_portals'
 
   before_action :find_authentication_provider, only: %i[show edit update publish_or_hide destroy]


app/controllers/provider/admin/authentication_providers_controller.rb
```diff
@@ -3,7 +3,7 @@
 class Provider::Admin::AuthenticationProvidersController < FrontendController
   before_action :authorize_settings
 
-  activate_menu :audience, :cms, :oauth2
+  activate_menu :audience, :cms, :sso_integrations
```
josemigallas, line 6: @thomasmaas 
There are 2 authentication provider controllers: one calls `:sso_integrations` and this one `oauth2`.

In `vertical_nav_helper` both menus have same id `:sso_integration`. Do you know if this ID matters?

We could change it here or in `vertical_nav_helper`:
https://github.com/3scale/porta/blob/b376edcebf7bf56673201ab30c091031aaba5ac2/app/helpers/vertical_nav_helper.rb#L151

hallelujah: There is no `:oauth2` in this helper in the activate_sidebar (third parameter is sidebar

https://github.com/3scale/porta/blob/master/app/lib/menu_system.rb#L62-L64

I think @srijitamukherjee is right

josemigallas: > There is no :oauth2 in this helper in the activate_sidebar (third parameter is sidebar

Yep I'm aware, I did that. But I could have been wrong, I'm asking if `oauth2` is relevant for something else than the sidebar. Just in case. @thomasmaas wrote that line so maybe he knows.

thomasmaas: oauth2 has been deprecated

thomasmaas: and will not make it to portafly

Martouta:
Right 😄 
image

I would appreciate a test but in this case, I can also accept not having it.

hallelujah:
Yes test case not necessary ... active menu is tested but its usage is not correct or it was and we failed to spot this on oauth2 removal

PR #1894, opened 2020-05-20
[PoliciesController] InvalidPoliciesConfigError

macejmic:

A policies_config parameter should be an array. If it's an object, it fails on 500. We should validate this scenario and return 422 if possible.

### DIFF
--- a/app/controllers/admin/api/services/proxy/policies_controller.rb
+++ b/app/controllers/admin/api/services/proxy/policies_controller.rb
@@ -56,8 +56,9 @@ class Admin::Api::Services::Proxy::PoliciesController < Admin::Api::Services::Ba
 
   def proxy_params
     proxy_params = params.require(:proxy).dup
+
     proxy_params.permit(:policies_config).tap do |whitelisted|
-      whitelisted[:policies_config] = proxy_params[:policies_config].tap{|h| h.try(:each, &:permit!)}
+      whitelisted[:policies_config] = PoliciesConfigParams.new(proxy_params[:policies_config]).call
     end
   end
 
--- a/app/services/policies_config_params_wrapper.rb
+++ b/app/services/policies_config_params_wrapper.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+class PoliciesConfigParams
+
+  attr_reader :policies_config_params
+
+  def initialize(policies_config_params)
+    @policies_config_params = policies_config_params
+  end
+
+  def call
+    return policies_config_params if json_param?
+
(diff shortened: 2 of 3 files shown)


app/controllers/admin/api/services/proxy/policies_controller.rb
```diff
@@ -56,6 +56,11 @@ def authorize_rolling_update
 
   def proxy_params
     proxy_params = params.require(:proxy).dup
+    policies_config_params = proxy_params[:policies_config]
+
+    if policies_config_params && policies_config_params.is_a?(ActionController::Parameters) && policies_config_params.to_unsafe_h.is_a?(Hash)
```
hallelujah, line 61: This makes it too much responsibility

You can create a wrapper to transform the `policies_config_params` to an object you know the behaviour ;)

Martouta: Some extra info from Hery:

> this check should not be in the controller action itself
> you can create an object that makes those check for you and spit out the desired format (Array or Hash)
> And then you do not even need a is_a? I think :slightly_smiling_face:

app/services/policies_config_params_wrapper.rb
```diff
@@ -0,0 +1,28 @@
...
+class PoliciesConfigParamsWrapper
+
+  attr_reader :policies_config_params
+
+  def initialize(policies_config_params)
+    @policies_config_params = policies_config_params
+  end
+
+  def wrap
+    return policies_config_params if json_param?
```
hallelujah, line 12: You return a String? No need to parse it?

Then it you could leverage the update in the model

macejmic: `policies_config` is being saved in a JSON.. so to make it "smarter" I would have to decode the array to JSON and then send it to a proxy object. so I would say better to keep it this way..

app/services/policies_config_params_wrapper.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+class PoliciesConfigParamsWrapper
```
hallelujah, line 3: Maybe Wrapper is not the right word, my fault, I induced you to write this with my comment. Trying to find a replacement...

How about simply `PoliciesConfigParams` and use `call` instead of `wrap`?

macejmic: sure, done.

Martouta:
image
Easy to fix but I won't force you to push only for this 😝

PR #1898, opened 2020-05-21
Remove credentials from Dockerfile

hallelujah:
They are no use anyway, not working
And it is a leak adding them


hallelujah:
@Martouta why do we remove the entire Dockerfile?
I believe she is removing the docker-compose.mk not the Dockerfile ;)

Martouta:
ah right 🙈 _lapsus_ 😊

PR #1899, opened 2020-05-21
Do not report error to Bugsnag if Authentication Token is empty

Martouta:
If someone tries to do `POST /session` to login to the dev portal without the `authentication token` it will fail the request as `forbidden` for https://github.com/3scale/porta/pull/1873
It will also log the error in Rails (because `handle_error` does it 😄 ), but we should additionally notify to Bugsnag when it might be a bug on our side, which is not when the authentication token is sent empty.
Related to what I explained [in this comment in JIRA](https://issues.redhat.com/browse/THREESCALE-5240?focusedCommentId=14109051&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-14109051).

### DIFF
--- a/app/controllers/concerns/error_handling.rb
+++ b/app/controllers/concerns/error_handling.rb
@@ -28,7 +28,7 @@ module ErrorHandling
     end
 
     def handle_forgery_protection(exception)
-      System::ErrorReporting.report_error(exception)
+      System::ErrorReporting.report_error(exception) if params[:authenticity_token].present?
       handle_error(exception, :forbidden)
     end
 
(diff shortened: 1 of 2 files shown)


app/controllers/concerns/error_handling.rb
```diff
@@ -28,7 +28,7 @@ def handle_access_denied(exception)
     end
 
     def handle_forgery_protection(exception)
-      System::ErrorReporting.report_error(exception)
+      System::ErrorReporting.report_error(exception) if params[:authenticity_token].present?
       handle_error(exception, :forbidden)
```
Martouta, line 32: This contains adding logs in Rails.logger :)

hallelujah: Maybe only if the exception is of type `ActionController::InvalidAuthenticityToken` ?

PR #1902, opened 2020-05-22
Adds Sorting to Data List

josemigallas:
Adds sorting to Table's context and implements it in the Overview.

`sortBy` format and behavior is defined by Patternfly's `Table`.

⚠️ A workaround is needed. When a table is "selectable", the sortBy takes into account the first column (checkboxes) but not the row items. The workaround is to add 1 to the sorting index.

### DIFF
--- a/portafly/src/components/data-list/reducers/TableReducer.ts
+++ b/portafly/src/components/data-list/reducers/TableReducer.ts
@@ -17,18 +17,18 @@ export type TableState = {
 }
 
 // Action Handlers
-const SORT_BY = 'SORT_BY'
+const SET_SORT_BY = 'SET_SORT_BY'
 const SELECT_ONE = 'SELECT_ONE'
 const SELECT_PAGE = 'SELECT_PAGE'
 const SELECT_ALL = 'SELECT_ALL'
 
 const tableActionHandlers: ActionHandlers<TableState, any> = {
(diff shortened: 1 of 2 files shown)


portafly/src/components/data-list/reducers/TableReducer.ts
```diff
@@ -5,18 +5,41 @@ import {
...
+      if (prev > next) return 1
+      return 0
+    })
+
+    if (direction === SortByDirection.desc) {
+      sortedRows.reverse()
+    }
+
+    return { ...state, rows: sortedRows, sortBy: action.payload }
+  },
```
didierofrivia, line 42: Careful, both `sort` and `reverse` change the original array, so here you are mutating the state. 
You could `slice` the original array before the sort, and that would be enough in this case. Also, better to export that logic into a separated function to make it clear the sorting, something like `sortByIndexCorrection` or `withIndexCorrection`

josemigallas: > Careful, both `sort` and `reverse` change the original array, so here you are mutating the state.

👍 

> Also, better to export that logic into a separated function to make it clear the sorting, something like `sortByIndexCorrection` or `withIndexCorrection`

Now that I think about it, I don't suppose this belongs to the reducer. I'm taking it out and figure the _correction_ out besides the Table.

PR #1904, opened 2020-05-25
Prepare removing monkey patch on request.host

hallelujah:

Do not make features worry about if we are in production or in preview.
We need to prepare of the removal of ThreeScale::Middleware::DevDomain and remove all monkey patching of `request.host`

This first step is to use `request.internal_host` when it is possible
It will deprecate the use of `Account#domain` and `Account#self_domain`
They will be private later

Relates to https://issues.redhat.com/browse/THREESCALE-5265
Partially closes it.

Example:

When enabled you can point your browser to `http://provider-admin.stg-saas.localhost` and it will use the `Account` which `self_domain` is `provider-admin.example.com`

### DIFF
--- a/test/unit/site_account_support_test.rb
+++ b/test/unit/site_account_support_test.rb
@@ -21,14 +21,21 @@ class SiteAccountSupportTest < ActiveSupport::TestCase
     p2 = FactoryBot.create(:provider_account)
 
     object = Params.new
-    object.request = mock(host: p2.self_domain)
-    object.request.expects(:params).returns({ provider_key: p2.api_key }).at_least_once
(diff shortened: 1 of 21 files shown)


test/unit/site_account_support_test.rb
```diff
@@ -21,14 +21,21 @@ def helper_method(*args)
     p2 = FactoryBot.create(:provider_account)
 
     object = Params.new
-    object.request = mock(host: p2.self_domain)
-    object.request.expects(:params).returns({ provider_key: p2.api_key }).at_least_once
+    mock_request = ActionDispatch::TestRequest.create(
```
hallelujah, line 24: Use a "real" mock request :)

PR #1905, opened 2020-05-25
Extract filterRows from props

josemigallas:
This is probably a middle step between passing filteredRows as props and storing them in the context.

As a side-effect, `categories` has also to be stored in the context.

Managing it inside the context is complex and rises some questions about its fundamental architecture. The complexity lies in the fact that in order to calculate the filtered rows, both `dataListTable` and `dataListFilters` contexts need to be used. It would be much more easier if we had a unique context or a way to add reducers that use the whole state instead of state slices.


portafly/src/components/data-list/DataListContext.tsx
```diff
@@ -28,7 +28,9 @@ const defaultState: State = {
     rows: [],
     columns: []
   },
-  filters: {},
+  filters: {
+    categories: []
```
didierofrivia, line 32: Hmmm, I'm not convinced about this. Previously, filters were `Record<string, string[]>` were the first string corresponds to a `Category`. If you want to include the categories in the state, I'd rename the `filters` to something else to include both `filters` and `categories`. Also, those could be in different slices of the state if it makes sense.

portafly/src/components/data-list/utils/filterRows.ts
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    newRows = newRows.filter((r) => {
+      const value = (r.cells)[colIdx].toLowerCase()
+      return filters[categoryName].some((f) => value.indexOf(f.toLowerCase()) > -1)
+    })
+  })
+
+  return newRows
+}
```
didierofrivia, line 36:
```suggestion
const filter = (rows) => (colIdx) =>(filtersInCategory) => (
  rows.filter((r) => filtersInCategory.some((f) => (r.cells)[colIdx].toLowerCase().indexOf(f.toLowerCase()) > -1))
)
// It even could be a custom find fn in utils where returns the element or raise error
const findCategory = (categories) => (categoryName) => {
  const category = categories.find((c) => c.name === categoryName)
  if (!category) throw new Error('You are trying to filter a category that doesn\'t exist in your DataList: %o', categoryName)
  return category
}
const findColIdx = (columns) => (categoryHumanName) => {
  const index = columns.findIndex((c) => c.title === categoryHumanName)
  if (index === -1) throw new Error('You are trying to filter a category that doesn\'t correspond to any column in your DataList')
  return index
}

const filterRows = (
  rows: DataListRow[],
  filters: Record<string, string[]>,
  categories: Category[],
  columns: ICell[]
) => (
  Object.keys(filters).map((categoryName) => {
    const category = findCategory(categories)(categoryName)
    const colIdx = findColIdx(columns)(category.categoryHumanName)
    return (!!rows.length || !!filters[categoryName].length)
      ? filter(rows)(colIdx)(filters[categoryName])
      : rows
  })
)

export { filterRows }

```

portafly/src/components/data-list/DataListContext.tsx
josemigallas: Well using categories is a middle step to find the column index so I think it could be a more elegant solution to add an "id" to the columns, where such id is the name of the category (see last commit).

portafly/src/components/data-list/utils/filterRows.ts
```diff
@@ -0,0 +1,29 @@
...
```
didierofrivia, line 29: This could be more declarative instead of imperative as I've shown in the previous review ;)

josemigallas: Yeah.. but I really don't like chaining so many arrow functions. It makes it harder to debug and read IMO.
I think this way it's more readable at least, even if it's a bit verbose. You think there's something wrong with it or is it only preference?

didierofrivia: Besides the verbosity, the use of `let`, `forEach` and 2 conditionals? just the code being imperative. The functions being named help for the readability, the name is telling you what's doing, and/or what it is. Using that kind of composition makes it easier to debug and maintain, each tiny function is easier than 20+ lines one. 
That being said and this being tested, right now it's ok to merge... Most probably it will be follow by a refactor more declarative in the near future

PR #1906, opened 2020-05-25
Clean up Portafly Login tests

josemigallas:
* Removes Snapshots since they're a bit overkilling
* Cleanup tests


josemigallas:
@didierofrivia The milestone branch is broken. Here's the fix https://github.com/3scale/porta/pull/1912

PR #1909, opened 2020-05-26
Fix ActiveDocs 2.0 Liquid Oracle CircleCI failures & Speed it up

Martouta:
Part of [THREESCALE-5261](https://issues.redhat.com/browse/THREESCALE-5261)
- [X] Fix ActiveDocs 2.0 Liquid

- [X] InitialData reuses master_account and has one single Before block.
This is faster and it can be reused comfortably when doing ssh to CircleCI 😄

### DIFF
--- a/features/support/initial_data.rb
+++ b/features/support/initial_data.rb
@@ -1,4 +1,5 @@
-#TODO: join all these into one single Before for speed
+# frozen_string_literal: true
+
 Before do
   countries = {'ES' => 'Spain',
                'US' => 'United States of America'}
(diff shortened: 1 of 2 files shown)


features/support/initial_data.rb
```diff
@@ -1,16 +1,13 @@
...
 
   countries.each do |code, name|
     FactoryBot.create(:country, :name => name, :code => code) unless Country.exists?(:code => code)
   end
-end
 
-Before do
   ThreeScale.config.stubs(superdomain: 'example.com')
-  FieldsDefinition.create_defaults! FactoryBot.create(:master_account)
+  FieldsDefinition.create_defaults! master_account
```
hallelujah, line 10: <3

features/step_definitions/cms/page_template_steps.rb
```diff
@@ -12,6 +12,7 @@ def fill_draft(text)
     page.execute_script <<-JS
       $('#cms_template_draft').data('codemirror').setValue(#{text.inspect});
     JS
+    wait_for_requests
```
hallelujah, line 15: So setting the data attribute `codemirror` triggers an AJAX request?

Martouta: My bad 🙈

PR #1910, opened 2020-05-26
[Portafly] Fixing wrong strings in accounts and shared pages

damianpm:

This is a follow up of https://github.com/3scale/porta/pull/1861, where we introduced bad formatted yaml strings

### DIFF
--- a/portafly/src/i18n/locales/en/accounts.yml
+++ b/portafly/src/i18n/locales/en/accounts.yml
@@ -27,19 +27,28 @@ team_members_filter:
   by_name: Name
   by_email: Email
   by_sso_authentication: SSO authentication
-   yes: Yes
-   no: Not yet
+  by_sso_authentication_description: Label for the filter criterion based on SSO authentication
+  by_sso_authentication_options:
+    yes: Yes
+    no: Not yet
+  by_sso_authentication_options_description: Group name that contains the options of the SSO authentication criterion
   by_role: Role
-   admin: Admin
-   member: Member
+  by_role_description: Label for the filter criterion based on Role
+  by_role_options:
+    admin: Admin
+    member: Member
+  by_role_options_description: Group name that contains the options of the Role criterion
   by_permission_group: Permission group
-   developer_portal: Developer Portal
-   billing: Billing
-   settings: Settings
-   accounts_applications: Accounts – Applications
-   analytics: Analytics
-   configuration_application_plans: Configuration and application plans
-   policy_registry: Policy registry
+  by_permission_group_description: Label for the filter criterion based on Permission group
+  by_permission_group_options:
+    developer_portal: Developer Portal
+    billing: Billing
+    settings: Settings
+    accounts_applications: Accounts – Applications
+    analytics: Analytics
+    configuration_application_plans: Configuration and application plans
+    policy_registry: Policy registry
+  by_permission_group_options_description: Group name that contains the options of the Permission group criterion
 team_members_filter_description: List containing the name criterion to filter team members
 team_members_typeahead: Filter by name
 team_members_typeahead_description: Default text to show when the user wants to filter by typing the text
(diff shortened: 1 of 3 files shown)


portafly/src/i18n/locales/en/accounts.yml
```diff
@@ -27,19 +27,22 @@ team_members_filter:
   by_name: Name
   by_email: Email
   by_sso_authentication: SSO authentication
-   yes: Yes
-   no: Not yet
+  by_sso_authentication_options:
+    yes: Yes
+    no: Not yet
```
damianpm, line 32: @porueesq @dfennessy 
In yaml format, where you have indentation it means the indented items are children.
In this case `yes` and `no` are children of `by_sso_authentication`
The parent `by_sso_authentication`, can't have a value when it has children

portafly/src/i18n/locales/en/shared.yml
```diff
@@ -30,7 +30,7 @@ format:
   lowercase_description: Formats a text as lowercase
   format_description: Utilities to format text
 shared_elements:
-  bulk_selector: {{number}} selected
+  bulk_selector: "{{number}} selected"
```
damianpm, line 33: @porueesq @dfennessy 
When we use interpolation (variables like `{{number}}`), we have to use quotes, other way the library is unable to recognize the vars

portafly/src/i18n/locales/en/accounts.yml
porueesq: @damianpm 
> In this case yes and no are children of by_sso_authentication
Exactly, by_sso_authentication is a filter criterion that has two possible values: Yes and Not yet

> The parent by_sso_authentication, can't have a value when it has children
Not sure what you mean.

portafly/src/i18n/locales/en/shared.yml
porueesq: Understood.

portafly/src/i18n/locales/en/accounts.yml
damianpm: @porueesq 
Yes, we need an item for the label (by_sso_authentication), and one for the options (by_sso_authentication_options).
The second one, as it has children, can't have a value, it means that you can't  do this:
```
  by_sso_authentication: SSO authentication
    yes: Yes
    no: Not yet
```
In other words, you can't assign "SSO authentication" to "by_sso_authentication", that's possible only when the item has no children

To give you a better idea, it will look like this
filters

portafly/src/i18n/locales/en/accounts.yml
```diff
@@ -27,19 +27,25 @@ team_members_filter:
...
+  by_sso_authentication_options_description: List of possible values of by_sso_authentication_options
```
damianpm, line 33: @porueesq I've added the descriptions

portafly/src/i18n/locales/en/accounts.yml
porueesq: @damianpm Got you. Then these lines should be available in the file:

```
by_sso_authentication: SSO authentication
by_sso_authentication_description: Label for the filter criterion based on SSO authentication

by_sso_authentication_options: 
by_sso_authentication_options_description: Group name that contains the option of the SSO authentication criterion
 yes: Yes
 no: Not yet
```

portafly/src/i18n/locales/en/accounts.yml
porueesq: @damianpm To avoid dependencies, we don't use variables in the descriptions. I've suggested the contents and descriptions here: https://github.com/3scale/porta/pull/1910/files#r430329124

portafly/src/i18n/locales/en/accounts.yml
```diff
@@ -27,19 +27,28 @@ team_members_filter:
   by_name: Name
   by_email: Email
   by_sso_authentication: SSO authentication
-   yes: Yes
-   no: Not yet
+  by_sso_authentication_description: Label for the filter criterion based on SSO authentication
```
damianpm, line 30: @porueesq I've added the descriptions

PR #1914, opened 2020-05-26
Handle ActionDispatch::ParamsParser::ParseError

Martouta:
Closes [THREESCALE-5280](https://issues.redhat.com/browse/THREESCALE-5280)

### DIFF
--- a/test/unit/three_scale/middleware/handle_parse_error_test.rb
+++ b/test/unit/three_scale/middleware/handle_parse_error_test.rb
@@ -1,0 +1,62 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class ThreeScale::Middleware::HandleParseErrorTest < ActiveSupport::TestCase
+  test 'correct request params' do
+    app = proc { perform_post_request(raw_post_data: '{ "foo" : "bar" }') }
+    middleware = ThreeScale::Middleware::HandleParseError.new(app)
+
+    expected_response = {'foo' => 'bar'}
+    assert_equal expected_response, middleware.call({})
+  end
+
+  test 'invalid request params' do
+    Rails.logger.expects(:error).with('Handling Exception: \'ActionDispatch::ParamsParser::ParseError\' with status 422')
+
+    app = proc { perform_post_request(raw_post_data: '{ wrong: "invalid", ') }
+    middleware = ThreeScale::Middleware::HandleParseError.new(app)
+
+    assert_equal [422, {}, [nil]], middleware.call({})
+  end
+
+  private
+
+  def perform_post_request(raw_post_data:)
+    env = {
+      'rack.input' => 'foo',
+      'CONTENT_TYPE' => 'application/json',
+      'CONTENT_LENGTH' => 9,
+      'RAW_POST_DATA' => raw_post_data
+    }
+    request = ActionDispatch::Request.new(env)
+    request.POST
+  end
+
+  class RequestIntegrationTest < ActionDispatch::IntegrationTest
+    include ApiRouting
+
+    def setup
+      provider = FactoryBot.create(:simple_provider) # TODO: We do not even need it to have users or services or anything
+      host! provider.admin_domain
+    end
+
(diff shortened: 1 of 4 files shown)


lib/three_scale/middleware/multitenant.rb
```diff
@@ -111,9 +111,27 @@ def call(env)
       def _call(env)
         Thread.current[:multitenant] = TenantChecker.new(@attribute,@app,env)
         @app.call(env)
+      rescue ActionDispatch::ParamsParser::ParseError => error
```
hallelujah, line 114: I do not think it is the responsibility of this middleware. Rather create another one that is called just before ActionDis

It should be caught earlier in the chain, create another middleware if needed, so you will disable it on test and development but safely rescue in production

Martouta: > so you will disable it on test and development but safely rescue in production
I do not think that disabling it by default in `test` is a good idea 😄 We should make valid format requests in `test`, OR test that it is behaving as we desire in production when the params are not correct.

>I do not think it is the responsibility of this middleware. Rather create another one that is called just before Rack::Sendfile
> It should be caught earlier in the chain, create another middleware if needed

Gotcha 👍

hallelujah: Raising in development and test is better no? In production you fallback to a response.

That's how rails work with 404 and `ActiveRecord::RecordNotFound` at least. Make it configurable with sane defaults

Martouta: For everybody else, we spoke about this in private, and these are the conclusions:

We never raise it again for RecordNotFound and we won't raise it either for this one 😄 

https://github.com/3scale/porta/blob/7b55693bcedfed8cd707550126c1444a0187111d/app/controllers/application_controller.rb#L11
https://github.com/3scale/porta/blob/7b55693bcedfed8cd707550126c1444a0187111d/app/controllers/concerns/error_handling.rb#L8
https://github.com/3scale/porta/blob/7b55693bcedfed8cd707550126c1444a0187111d/app/controllers/concerns/error_handling.rb#L22-L24
https://github.com/3scale/porta/blob/7b55693bcedfed8cd707550126c1444a0187111d/app/controllers/concerns/error_handling.rb#L40-L65

Martouta, line 114: `/Users/marta/.rvm/gems/ruby-2.5.5/gems/actionpack-5.0.7.2/lib/action_dispatch/middleware/params_parser.rb`

👀

test/integration/admin/api/buyers_users_controller_test.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class Admin::Api::BuyersUsersControllerTest < ActionDispatch::IntegrationTest
```
hallelujah, line 5: I do not think it is related to this controller

Please create a more generic one to test it ;)
Btw no need to test in controller just make a rack test is enough

Martouta: Awesome, will do 👍

test/unit/three_scale/middleware/handle_parse_error_test.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class ThreeScale::Middleware::HandleParseErrorTest < ActionDispatch::IntegrationTest
```
hallelujah, line 5: You do not need (but you can have ...) this integration test, as long you have the corresponding unit test
See:
https://github.com/3scale/porta/blob/3d07b9e15c22ae607dad7abc3aebca4bc0770e8a/test/unit/sidekiq_logging_middleware_test.rb
and
https://github.com/3scale/porta/blob/3d07b9e15c22ae607dad7abc3aebca4bc0770e8a/test/unit/three_scale/middleware/dev_domain_test.rb

PR #1916, opened 2020-05-26
[Fix Oracle CircleCI] Scroll to button if not visible at 1st

Martouta:
Let's give credit to the [SO hero](https://stackoverflow.com/questions/17623075/auto-scroll-a-button-into-view-with-capybara-and-selenium) if this works 😄

### DIFF
--- a/features/step_definitions/web_steps.rb
+++ b/features/step_definitions/web_steps.rb
@@ -17,6 +17,17 @@ module WithinHelpers
 end
 World(WithinHelpers)
 
+module JavascriptDriver
+  def scroll_to(element)
+    script = <<-JS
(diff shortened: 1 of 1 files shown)


features/step_definitions/web_steps.rb
```diff
@@ -17,6 +17,17 @@ def with_scope(locator)
 end
 World(WithinHelpers)
 
+module JavascriptDriver
+  def scroll_to(element)
```
hallelujah, line 21: This is already implemented in Capybara

Martouta:
Closing because it is failing, and it is not worth it investing time in that approach... instead it is better to go directly for the cucumber 3 and native scroll_to approach :wink:

PR #1922, opened 2020-05-27
Fix ensure_provider callback returning false

hallelujah:
Fixes rubocop warnnig

Notes: `throw(:abort)` only works for ActiveRecord and ActiveModel

### DIFF
--- a/lib/developer_portal/app/controllers/developer_portal/cms/new_content_controller.rb
+++ b/lib/developer_portal/app/controllers/developer_portal/cms/new_content_controller.rb
@@ -3,6 +3,7 @@ class DeveloperPortal::CMS::NewContentController < DeveloperPortal::BaseControll
   skip_before_action :login_required
 
   before_action :ensure_can_view_content, :if => :check_permissions?
+  append_before_action :ensure_buyer_domain, only: [:show]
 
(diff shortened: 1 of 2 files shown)


lib/developer_portal/app/controllers/developer_portal/cms/new_content_controller.rb
```diff
@@ -3,6 +3,7 @@ class DeveloperPortal::CMS::NewContentController < DeveloperPortal::BaseControll
   skip_before_action :login_required
 
   before_action :ensure_can_view_content, :if => :check_permissions?
+  append_before_action :ensure_buyer_domain, only: [:show]
```
Martouta, line 6: wait... now we are running the `show` anymore even if it is not buyer domain... no?

hallelujah: It is related to this @Martouta https://github.com/3scale/porta/pull/1922/files#diff-1c011907b2542a5292030b656bc59903L26

Martouta: I see, sorry 🙈 👍

PR #1927, opened 2020-05-28
Custom ThreeScale::Deprecation

hallelujah:
Just a subclass of `ActiveSupport::Deprecation`
But it has its own `#behavior` and `#silenced` methods

To modify its configurations, edit ./config/initializers/deprecation.rb

## Usage

See `ActiveSupport::Deprecation` for full usage

```ruby
deprecate :some_method,
  another_method: "use #replacement_method instead",
  deprecator: ThreeScale::Deprecation.new
```


hallelujah:
@dependabot merge

PR #1930, opened 2020-05-28
Use domain substitution in mailers

hallelujah:
Do not use account.domain or account.admin_domain anymore in mailers
Other url in drops will have to be fixed

Relates to https://issues.redhat.com/browse/THREESCALE-5265

### DIFF
--- a/app/mailers/provider_user_mailer.rb
+++ b/app/mailers/provider_user_mailer.rb
@@ -37,7 +37,7 @@ class ProviderUserMailer < ActionMailer::Base
 
   def domain(account)
     raise "Using ProviderUserMailer for buyer account #{account.name}(#{account.id})" unless account.master? || account.provider?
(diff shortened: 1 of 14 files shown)


app/mailers/provider_user_mailer.rb
```diff
@@ -82,7 +82,7 @@ def domain(account)
   # private
 
   # def admin_url(user)
-  #   admin_dashboard_url(:host => user.account.provider_account.domain)
+  #   admin_dashboard_url(:host => user.account.provider_account.external_admin_domain)
   # end
```
Martouta, line 86: If this is not used, perhaps we could directly remove it 😄

hallelujah: I prefer letting it there, not the purpose of this PR
Maybe somebody will take a look at it

PR #1933, opened 2020-05-29
Update i18n strings

josemigallas:
### What / Why
* Removal of all aria-labels that are not being tracked in our translations. They will be added properly when we figure out what we need.
* Proposal for namespaces: since we have so many files inside our translations folder, I came up with the idea of defining a namespace for each page so usage of `t` is less verbose. We no longer have to define `sections` (renamed to `namespaces`) manually because it uses the keys of the exports directly.

### DIFF
--- a/portafly/src/components/Loading.tsx
+++ b/portafly/src/components/Loading.tsx
@@ -7,12 +7,16 @@ import {
   Title,
   Spinner
 } from '@patternfly/react-core'
+import { useTranslation } from 'i18n/useTranslation'
 
-export const Loading: React.FunctionComponent = () => (
-  <PageSection aria-label="Loading Content Container">
-    <EmptyState variant={EmptyStateVariant.full}>
-      <EmptyStateIcon variant="container" component={Spinner} />
-      <Title headingLevel="h2" size="lg">Loading</Title>
-    </EmptyState>
-  </PageSection>
-)
+export const Loading: React.FunctionComponent = () => {
+  const { t } = useTranslation('shared')
+  return (
+    <PageSection aria-label={t('loading.aria_label')}>
+      <EmptyState variant={EmptyStateVariant.full}>
+        <EmptyStateIcon variant="container" component={Spinner} />
+        <Title headingLevel="h2" size="lg">{t('loading.title')}</Title>
+      </EmptyState>
+    </PageSection>
+  )
+}
--- a/portafly/src/components/data-list/SearchWidget.tsx
+++ b/portafly/src/components/data-list/SearchWidget.tsx
@@ -15,6 +15,7 @@ import {
 } from '@patternfly/react-core'
 import { FilterIcon, SearchIcon } from '@patternfly/react-icons'
 import { useDataListFilters } from 'components/data-list/'
+import { useTranslation } from 'i18n/useTranslation'
 import { Category, CategoryOption } from 'types'
 
 import './searchWidget.scss'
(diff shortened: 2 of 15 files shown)


josemigallas:
> * why removing aria labels and not adding comments next to those stating that we're gonna deal with them later, so
>   
>   * we're sure we don't forget about them
>   * we don't have to do more work later about locating where those aria labels initially where

I assume the existing aria labels are wrong / insufficient, therefore I would do it from scratch.

> * would the difference in the way the UI strings are located by namespaces suggest a different folder structure of those UI strings? or that doesn't really matter?

No, "namespaces" is just a Javascript thing, it doesn't matter how the translations files are structured.

cntlsn:
> > * why removing aria labels and not adding comments next to those stating that we're gonna deal with them later, so
> >   
> >   * we're sure we don't forget about them
> >   * we don't have to do more work later about locating where those aria labels initially where
> 
> I assume the existing aria labels are wrong / insufficient, therefore I would do it from scratch.

Let's make sure to have a conversation about this with PF team and agree on what is the desired way to deal with aria labels, before we embark on making something from scratch!

> 
> > * would the difference in the way the UI strings are located by namespaces suggest a different folder structure of those UI strings? or that doesn't really matter?
> 
> No, "namespaces" is just a Javascript thing, it doesn't matter how the translations files are structured.

Ok, this means we can decide on the best way to organize the strings for the people who will have to work on them (Docs team) independently from their implementation, correct?!

portafly/src/components/Loading.tsx
```diff
@@ -7,12 +7,17 @@ import {
...
-      <EmptyStateIcon variant="container" component={Spinner} />
-      <Title headingLevel="h2" size="lg">Loading</Title>
-    </EmptyState>
-  </PageSection>
-)
+export const Loading: React.FunctionComponent = () => {
+  const { t } = useTranslation('shared')
+  return (
+  // FIXME: missing translation string
+    <PageSection aria-label={t('loading_aria_label')}>
```
damianpm, line 16: Instead of leaving a nonexistent string here, why not creating the string in the lang file, using the current value, and add the TODO/FIXME comment ?
In shared.yml, we can add something like
```
aria_labels:
  loading: Loading Content Container
```

portafly/src/components/Loading.tsx
```diff
@@ -7,12 +7,17 @@ import {
...
+      <EmptyState variant={EmptyStateVariant.full}>
+        <EmptyStateIcon variant="container" component={Spinner} />
+        <Title headingLevel="h2" size="lg">Loading</Title>
```
damianpm, line 19: And "Loading" could also be moved to strings file

portafly/src/pages/Overview.tsx
```diff
@@ -98,7 +98,7 @@ const DataListTable = () => {
   return (
     <>
       <Table
-        aria-label="data-list-table"
+        aria-label="dummy-aria-label"
```
damianpm, line 101: TODO: remove dummy stuff

portafly/src/components/data-list/SearchWidget.tsx
```diff
@@ -97,6 +100,7 @@ interface ISearchBar {
 }
 
 const SearchBar = ({ textInputRef, category, onSearchClick }: ISearchBar) => {
+  const { t } = useTranslation('audienceAccountsListing')
```
didierofrivia, line 103: I think here this is too specific, since the SearchBar component will be used in many other data lists than accounts. If we can't find a more generic set of translations and scope, we might want to pass it in the props maybe?

josemigallas: I agree, however that is something we should address in another PR (a restructure of translations). Certainly these strings belong in "data-list" or at least "shared". Maybe I could do that change as part of the DevAccounts list. These strings should not have been merged to the milestone (but the accounts listing PR) in the first place.

PR #1934, opened 2020-05-29
'ActionDispatch::ParamsParser::ParseError' is a 400 instead of 422

Martouta:
My bad 🙈 
Fixes the status of https://github.com/3scale/porta/pull/1914. Part of [THREESCALE-5280](https://issues.redhat.com/browse/THREESCALE-5280)

[This is why](https://www.restapitutorial.com/httpstatuscodes.html):

-  **400 Bad Request**
The request could not be understood by the server due to malformed syntax. The client SHOULD NOT repeat the request without modifications.

- **422 Unprocessable Entity (WebDAV)**
The 422 (Unprocessable Entity) status code means the server understands the content type of the request entity (hence a 415(Unsupported Media Type) status code is inappropriate), and the syntax of the request entity is correct (thus a 400 (Bad Request) status code is inappropriate) but was unable to process the contained instructions. For example, this error condition may occur if an XML request body contains well-formed (i.e., syntactically correct), but semantically erroneous, XML instructions.


hallelujah:
👍  Thank you!

PR #1863, opened 2020-05-06
[POC] Autocomplete OAS3 with 3scale account data - 1

damianpm:
**What this PR does / why we need it**:

We need to deliver active docs like features for OAS3

This POC implements **autocomplete** using our `/p/admin/api_docs/account_data.json` endpoint. The following fields have autocomplete: 

- account_id
- application_id
- user_key 
- Not in this PR, but all non-empty fields in the response of account_data.json may work

**Which issue(s) this PR fixes** 

https://issues.redhat.com/browse/THREESCALE-5181

**Verification steps** 

- Go to API > ActiveDocs 
  http://provider-admin.example.com.lvh.me:3000/apiconfig/services/2/api_docs
- Create a new OAS 3 spec
- Go to the preview of the OAS 3 spec

**Special notes for your reviewer**:

You need a valid **OAS3 spec** in json format, I have created a basic spec to test, you can find it [here](https://gist.github.com/damianpm/aca68c763223ee93aed4b6e9c2562cfe)

autocomplete

PR #1935, opened 2020-05-29
Implement Autocomplete of OAS3 with 3scale account data

damianpm:

Implements autocomplete of OAS 3 using 3scale account data

https://issues.redhat.com/browse/THREESCALE-4857

**Notes**

This PR is an alternative to https://github.com/3scale/porta/pull/1863

https://github.com/3scale/porta/pull/1863 uses the [plugins system](https://swagger.io/docs/open-source-tools/swagger-ui/customization/overview/) of swagger ui to wrap and modify the components.

This PR uses a different approach: **intercepts the response**, injects the data to autocomplete and leaves swagger-ui do the rendering.
See `requestInterceptor` in https://swagger.io/docs/open-source-tools/swagger-ui/usage/configuration/

**PROS:**

- Cleaner code
- No DOM manipulation
- No need to maintain wrapped components
- It uses swagger default ui components

**CONS:**

- If some examples are already in the specification, they will be overwritten.
I don't think it's a great problem, as we are autocompleting only the parameters using 3scale acccount data (from `/p/admin/api_docs/account_data.json` endpoint.).

autocomplete

**Steps to reproduce**

- Go to a Service's active docs section: apiconfig/services/2/api_docs
- Create a new spec using OAS v3 (you can use [this example](https://gist.github.com/damianpm/aca68c763223ee93aed4b6e9c2562cfe))
- Update the following templates at Dev Portal's Content section: /p/admin/cms/templates
**Documentation**
```
<h1>Documentation</h1>

<p>Use our live documentation to learn about the Echo API</p>

{% content_for javascripts %}
    {{ 'active_docs.js' | javascript_include_tag }}
{% endcontent_for %}

{% include 'shared/swagger_ui' %}
```
____
**shared/swagger_ui:**

Replace the `<script></script>` section at the end of the document with:

```
<script type="text/javascript">
  (function () {
    var url = "{{provider.api_specs.first.url}}";
    var accountDataUrl = '/api_docs/account_data.json'


    SwaggerUI({
      url: url,
      responseInterceptor: function(response){return autocompleteOAS3(response, accountDataUrl)},
      dom_id: '#swagger-ui-container',
      supportedSubmitMethods: ['get', 'post', 'put', 'delete', 'patch'],
      apisSorter: 'alpha', // can also be a function
      operationsSorter: 'method', // can also be 'alpha' or a function
      docExpansion: 'list',
      transport: function (httpClient, obj) { ApiDocsProxy.execute(httpClient, obj) }
      });
  }());
</script>
```


didierofrivia:
Could you point this to master in order to see all the changes?

damianpm:
> Could you point this to master in order to see all the changes?

@didierofrivia 
Done!

app/javascript/packs/service_active_docs.js
```diff
@@ -1,8 +1,19 @@
...
+    dataUrl: {
+      provider: '/p/admin/api_docs/account_data.json',
+      buyer: '/api_docs/account_data.json'
+    }
+  }
+  const { url, baseUrl, accountType } = document.getElementById(containerId).dataset
+  const dataUrl = accountType ? AUTOCOMPLETE_CONFIG.dataUrl[accountType] : 'buyer'
 
-  ActiveDocsSpec({ url }, containerId)
+  const data = await fetchData(`${baseUrl}${dataUrl}`)
```
josemigallas, line 15: Could this block the UI for a long time?

app/javascript/src/ActiveDocs/components/ActiveDocsSpec.jsx
```diff
@@ -4,15 +4,84 @@ import React from 'react'
...
+  return (
+    <SwaggerUI
+      url={url}
+      responseInterceptor={(response) => injectAccountDataToResponse(response, accountData)}
+    />)
+}
 
-const ActiveDocsSpecWrapper = (props: ActiveDocsSpecProps, id: string) => createReactWrapper(<ActiveDocsSpec {...props} />, id)
+const ActiveDocsSpecWrapper = (props: ActiveDocsSpecProps, id: string) =>
+  createReactWrapper(<ActiveDocsSpec {...props} />, id)
```
josemigallas, line 85:
```suggestion
const ActiveDocsSpecWrapper = (props: ActiveDocsSpecProps, id: string) => (
  createReactWrapper(<ActiveDocsSpec {...props} />, id)
)
```

app/javascript/src/ActiveDocs/components/ActiveDocsSpec.jsx
```diff
@@ -4,15 +4,84 @@ import React from 'react'
...
+  return responseWithAutocomplete
+}
+
+const ActiveDocsSpec = ({ url, accountData }: ActiveDocsSpecProps) => {
+  return (
+    <SwaggerUI
+      url={url}
+      responseInterceptor={(response) => injectAccountDataToResponse(response, accountData)}
+    />)
+}
```
josemigallas, line 82:
```suggestion
const ActiveDocsSpec = ({ url, accountData }: ActiveDocsSpecProps) => (
  <SwaggerUI
    url={url}
    responseInterceptor={(response) => injectAccountDataToResponse(response, accountData)}
  />
)
```

app/javascript/src/ActiveDocs/components/ActiveDocsSpec.jsx
```diff
@@ -4,15 +4,84 @@ import React from 'react'
...
+  const paramName = AUTOCOMPLETE_PARAMS_MAP[param.name]
+  const paramHasAutocompleteData = (accountData[paramName].length > 0) &&
+    accountData[paramName].every(param => param.value !== '')
+  if (!paramHasAutocompleteData) {
+    return param
+  }
+  const autocompleteData = accountData[paramName]
+  const examples = { '0': { summary: 'Select an option to autocomplete', value: '-' } }
+  autocompleteData.forEach((item, index) => {
+    examples[`${index + 1}`] = {
```
josemigallas, line 47: Q: does it have to be a string?
```suggestion
    examples[index + 1] = {
```

app/views/admin/api_docs/base/swagger.html.erb
```diff
@@ -3,7 +3,9 @@
 
 <div class="swagger-section">
   <div id="message-bar" class="swagger-ui-wrap">&nbsp;</div>
-  <div id="swagger-ui-container" class="swagger-ui-wrap" data-url="<%= spec_url(@api_docs_service) %>"></div>
+  <div id="swagger-ui-container" class="swagger-ui-wrap" data-url="<%= spec_url(@api_docs_service) %>"
+    data-base-url="<%= base_url %>" data-account-type="provider">
+  </div>
```
josemigallas, line 8:
```suggestion
  <div
    id="swagger-ui-container"
    class="swagger-ui-wrap"
    data-url="<%= spec_url(@api_docs_service) %>"
    data-base-url="<%= base_url %>"
    data-account-type="provider"
  >
  </div>
```

damianpm:
@josemigallas @didierofrivia 
then you agree this is a better option than https://github.com/3scale/porta/pull/1863 ?

josemigallas:
> @josemigallas @didierofrivia
> then you agree this is a better option than #1863 ?

It looks simpler and better

app/javascript/packs/service_active_docs.js
damianpm: Well, trying in preview with cache disabled it takes about 1 second, I think it's acceptable 

Screenshot 2020-06-03 at 18 09 57

app/javascript/src/ActiveDocs/components/ActiveDocsSpec.jsx
```diff
@@ -4,15 +4,84 @@ import React from 'react'
...
+      summary: item.name,
+      value: item.value
+    }
+  })
+  const paramWithExamples = {
+    ...param,
+    examples
+  }
+  return paramWithExamples
```
didierofrivia, line 56: You might want to add types first, I doubt you need to create an object for examples with an index, can't just be an Array of Objects with `summary` and `value`?

app/javascript/src/ActiveDocs/components/ActiveDocsSpec.jsx
```diff
@@ -4,15 +4,84 @@ import React from 'react'
...
+  for (let key in paths) {
+    pathsWithAutocomplete[key] = paths[key]
+    const parametersWithAutocomplete = paths[key].parameters
+      .map(param => getParam(param, accountData))
+    pathsWithAutocomplete[key].parameters = parametersWithAutocomplete
+  }
+  responseWithAutocomplete.body.paths = pathsWithAutocomplete
+  responseWithAutocomplete.data = JSON.stringify(responseWithAutocomplete.body)
+  responseWithAutocomplete.text = JSON.stringify(responseWithAutocomplete.body)
+  return responseWithAutocomplete
```
didierofrivia, line 73: So the entire idea is to replace the response with the autocomplete values. What's the type of this? I'm sure there's a way to simplify the logic of crafting the response with autocomplete

didierofrivia:
I think this is the right path to follow, you can close https://github.com/3scale/porta/pull/1863 if you want 👍

app/javascript/src/ActiveDocs/components/ActiveDocsSpec.jsx
damianpm: That's because this old flow issue: https://github.com/facebook/flow/issues/380
`Non-string literal property keys not supported.`

app/javascript/src/ActiveDocs/components/ActiveDocsSpec.jsx
damianpm: True, the index is not necessary

app/javascript/src/ActiveDocs/components/ActiveDocsSpec.jsx
```diff
@@ -1,18 +1,127 @@
...
+  'access_token': 'access_token'
+}
+
+type Primitive = string | number | boolean
+
+type AccountData = {
+  [string]: Array<{name: string, value: string}> | []
+}
+
+type SwaggerResponse = {
```
didierofrivia, line 35: Can't this type be imported from Swagger flow types?

app/javascript/src/ActiveDocs/components/ActiveDocsSpec.jsx
```diff
@@ -1,18 +1,127 @@
...
+  ok: boolean,
+  status: number,
+  statusText: string,
+  text: string,
   url: string
 }
 
-const ActiveDocsSpec = ({ url }: ActiveDocsSpecProps) => <SwaggerUI url={url} />
+type ActiveDocsSpecProps = {
+  accountData: AccountData | {},
```
didierofrivia, line 50: The problem here is that it accepts AccountData or any Object... can't this be AccountData optional?

app/javascript/src/ActiveDocs/components/ActiveDocsSpec.jsx
```diff
@@ -1,18 +1,127 @@
...
+  url: string,
+}
+
+type ParamArraySchema = {
+  type: 'array',
+  items: { type: Primitive }
+}
+type ParamEnumSchema = {
+  enum: Array<Primitive> | [],
```
didierofrivia, line 59: Here I don't think you need the second part, just 
```suggestion
  enum: Array<Primitive>,
```
does the trick

app/javascript/src/ActiveDocs/components/ActiveDocsSpec.jsx
damianpm: The problem is both
`flow-typed search swagger-ui`
And
`flow-typed search swagger-ui-react`
return
`No definitions found, sorry!`

app/javascript/src/ActiveDocs/components/ActiveDocsSpec.jsx
```diff
@@ -1,18 +1,127 @@
...
+      summary: item.name,
+      value: item.value
+    })
+  })
+  const paramWithExamples = {
+    ...param,
+    examples
+  }
+  return paramWithExamples
+}
```
didierofrivia, line 97: Something like the following would more elegant

```suggestion
const addAutocompleteToParam = (param: Param, accountData: AccountData): Param => {
  const paramName = AUTOCOMPLETE_PARAMS_MAP[param.name]
  const autocompleteData = accountData[paramName]
  const paramHasAutocompleteData = !!autocompleteData && autocompleteData.every(param => param.value !== '')
    return paramHasAutocompleteData
      ? { ...param, { examples: autocompleteData.reduce((examples, item) => ([...examples, {summary: item.name, value: item.value}]})
      , [{ summary: 'Select an option to autocomplete', value: '-' }]) } }
      : param
}
```

app/javascript/src/ActiveDocs/components/ActiveDocsSpec.jsx
```diff
@@ -1,18 +1,127 @@
...
+  for (let key in paths) {
+    pathsWithAutocomplete[key] = paths[key]
+    const parametersWithAutocomplete = paths[key].parameters
+      .map(param => getParam(param, accountData))
+    pathsWithAutocomplete[key].parameters = parametersWithAutocomplete
+  }
+  responseWithAutocomplete.body.paths = pathsWithAutocomplete
+  responseWithAutocomplete.data = JSON.stringify(responseWithAutocomplete.body)
+  responseWithAutocomplete.text = JSON.stringify(responseWithAutocomplete.body)
+  return responseWithAutocomplete
```
didierofrivia, line 113: You can follow the logic behind https://github.com/3scale/porta/pull/1935/files#r435275310 as well here

app/javascript/src/ActiveDocs/components/ActiveDocsSpec.jsx
```diff
@@ -1,18 +1,123 @@
...
+    pathsWithAutocomplete[key] = paths[key]
+    const parametersWithAutocomplete = paths[key].parameters
+      .map(param => addAutocompleteToParam(param, accountData))
+    pathsWithAutocomplete[key].parameters = parametersWithAutocomplete
+  }
+  responseWithAutocomplete.body.paths = pathsWithAutocomplete
+  responseWithAutocomplete.data = JSON.stringify(responseWithAutocomplete.body)
+  responseWithAutocomplete.text = JSON.stringify(responseWithAutocomplete.body)
+  return responseWithAutocomplete
+}
```
didierofrivia, line 108: You could do something similar to what I suggested in `addAutocompleteToParam` here.

app/javascript/packs/service_active_docs.js
josemigallas: But the thing is, it could slow it down more. Try using a slow 3G connection for instance (although I admit it's probably not the most common situation).
Using the promise instead would work around this [potential] problem.

spec/javascripts/ActiveDocs/ActiveDocsSpec.spec.jsx
```diff
@@ -16,7 +16,5 @@ it('should render itself', () => {
 
 it('should render swagger-ui', () => {
   const wrapper = mount(<ActiveDocsSpec url={url} />)
-  console.log(wrapper.debug())
-
   expect(wrapper).toMatchSnapshot()
```
josemigallas, line 19: Instead of adding the snapshot, since the rendered result it not very meaningful, I would play with `url`:
```suggestion
  expect(container.querySelector(`[url="${url}"]`)).toBeInTheDocument()
```

damianpm: We are not using `@testing-library/react` in `porta`, just in `portafly`

app/javascript/packs/service_active_docs.js
damianpm: I will move the fetch inside `ActiveDocsSpec`, that way `swagger-ui` will show the loading spinner

spec/javascripts/ActiveDocs/ActiveDocsSpec.spec.jsx
josemigallas: Well even better, since we're not limited to `container.query`.
```suggestion
  expect(wrapper.find(SwaggerUI).hasProp('url')).toBe(true)
```
I think the only thing we should care about is the actual specification is rendered. I think we can _find_ the methods and resources from our spec. The problem is `foo.example.com/api_docs/account_data.json` is an empty file so we would need to mock it somehow.

But what is clear to me is that the snapshot doesn't really count as a test.

lib/developer_portal/app/views/developer_portal/shared/_swagger_ui.html.liquid
```diff
@@ -10,19 +10,18 @@
 <script type="text/javascript">
   (function () {
     var url = "{{provider.api_specs.first.url}}";
+    var accountDataUrl = '/api_docs/account_data.json'
```
damianpm, line 13: @Martouta this is the url I want to get from ruby

lib/developer_portal/app/views/developer_portal/shared/_swagger_ui.html.liquid
```diff
@@ -10,19 +10,18 @@
...
+    {% comment %} var accountDataUrl = '{{api_docs_account_data_url}}' {% endcomment %}
```
damianpm, line 14: And this is not working

app/javascript/packs/OAS3Autocomplete.js
```diff
@@ -0,0 +1,76 @@
...
+          if (pathParameters) {
+            paths[key] = {
+              ...response.body.paths[key],
+              parameters: pathParameters.map(param => addAutocompleteToParam(param, accountData))
+            }
+          }
+          return paths
+        }
+        , {})
+    }
```
didierofrivia, line 68: You can extract this to another function if you want.

app/javascript/packs/service_active_docs.js
```diff
@@ -1,8 +1,18 @@
 import { ActiveDocsSpecWrapper as ActiveDocsSpec } from 'ActiveDocs/components/ActiveDocsSpec'
+import { autocompleteOAS3 } from './OAS3Autocomplete'
 
-document.addEventListener('DOMContentLoaded', () => {
+document.addEventListener('DOMContentLoaded', async () => {
```
josemigallas, line 4: Do we still need `async` here?

damianpm: Good catch, forgot to delete it

app/javascript/src/ActiveDocs/components/ActiveDocsSpec.jsx
```diff
@@ -1,18 +1,26 @@
...
+ accountDataUrl: string,
+ autocompleteOAS3: (response: SwaggerResponse, accountDataUrl: string) => Promise<SwaggerResponse>,
+ url: string,
 }
 
-const ActiveDocsSpec = ({ url }: ActiveDocsSpecProps) => <SwaggerUI url={url} />
+const ActiveDocsSpec = ({ url, accountDataUrl, autocompleteOAS3 }: ActiveDocsSpecProps) => (
+  <SwaggerUI
+    url={url}
+    responseInterceptor={(response) => autocompleteOAS3(response, accountDataUrl)}
```
josemigallas, line 18: Looks like `accountDataUrl` and `autocompleteOAS3` don't have to be passed. I would reduce props by passing `responseInterceptor` directly. This way `ActiveDocsSpec` and `SwaggerUI` maintain the same signature.

app/javascript/src/Types/SwaggerTypes.jsx
```diff
@@ -0,0 +1,50 @@
+export type Primitive = string | number | boolean
+
+export type AccountData = {
+ [string]: Array<{name: string, value: string}> | []
```
josemigallas, line 4: I think empty array is implicit:
```suggestion
 [string]: Array<{name: string, value: string}>
```

thomasmaas:
@damianpm be carefull to reference the correct jira in branch name. as now parent issue was moved to ready for testing

PR #1936, opened 2020-05-29
[portafly] Refactor: filterRows fn

didierofrivia:
* A more declarative approach
* If a wrong/invalid filter category is passed, it means the context
state was tampered or no longer valid

### DIFF
--- a/portafly/src/components/data-list/utils/filterRows.ts
+++ b/portafly/src/components/data-list/utils/filterRows.ts
@@ -1,44 +1,32 @@
 import { DataListRow, DataListCol, Filters } from 'types/data-list'
 
-const getFilterCallbackFn = (colIdx: number, terms: string[]) => (r: DataListRow) => {
-  const value = (r.cells)[colIdx].toLowerCase()
-  return terms.some((f) => value.indexOf(f.toLowerCase()) > -1)
-}
+const filterByTerm = (colIdx: number, terms: string[]) => (row: DataListRow): boolean => (
+  terms.length
+    ? terms.some((f) => (row.cells)[colIdx].toLowerCase().indexOf(f.toLowerCase()) > -1)
+    : true
+)
 
+const findColIdx = (columns: DataListCol[]) => (categoryName: string): number => {
+  const index = columns.findIndex((c) => c.categoryName === categoryName)
+  // TODO: In the future we could show the error in a toast. Event driven approach maybe.
+  if (index === -1) throw new Error('You are trying to filter a category that doesn\'t correspond to any column in your DataList')
+  return index
+}
 /**
  * Returns a new array with the provided rows filtered.
  * @param rows DataListRow[] - The rows to be filtered
  * @param filters Filters - An object containing different categories and filtering terms
  * @param columns DataListCol[] - The columns of the Table being filtered
  */
-export function filterRows(
+
+const filterRows = (
   rows: DataListRow[],
   filters: Filters,
   columns: DataListCol[]
-) {
-  if (rows.length === 0) {
-    return []
-  }
+): DataListRow[] => (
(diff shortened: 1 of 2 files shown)


portafly/src/components/data-list/utils/filterRows.ts
```diff
@@ -1,44 +1,32 @@
...
-  }
-
-  let newRows = [...rows]
-
-  Object.keys(filters).forEach((categoryName) => {
-    // Check category is valid
-    const colIdx = columns.findIndex((c) => c.categoryName === categoryName)
+): DataListRow[] => (
+  Object.keys(filters).reduce((acc, categoryName) => (
+    acc.filter(filterByTerm(findColIdx(columns)(categoryName), filters[categoryName]))
```
josemigallas, line 28: The problem with this approach is that it works "by magic". Ideally it should be understood on a quick read but instead this requires to look into each method. The use of `reduce` is a neat improvement and throwing an error is as valid as ignoring a wrong filters (neither should happen in practice anyway). But I think it's not worth it to have such a condensed method if it doesn't read well. I'd rather have multiple, meaningful lines that keeps the reader clear of going thru each function individually.
```suggestion
  Object.keys(filters).reduce((filteredRows, categoryName) => {
    const colIdx = findColIdxForCategoryName(columns, categoryName)
    const terms = filters[categoryName]
    const filterFn = filterRowByTerms(terms, colIdx)
    return filteredRows.filter(filterFn)
  }, rows)
```
Something like this suggestion may seem verbose or less performant but it really isn't. It's a way to explain step by step what's happening. On the other hand I think concatenating arrow functions is a call for problems, at least in JS. This what I understand when I read it for the first time:
```
findColIdx(columns, categoryName) // Find the index of a column using a categoryName
findColIdx(columns)(categoryName) // Find the index of a column (somehow) then do something with a categoryName
```
The second option can mean anything really, without looking into what `findColIdx` does.

Summing up, I usually put explicitness before obfuscation and this case is particularly unusual due to having 3 arguments and the complex models behind.

didierofrivia: I agree with you that code needs to be understood, and in this case I think just with some renaming of the aux filter functions and the reducer accumulator will make it neat. Thanks for pointing it out :)

However, we can agree that absolutely nothing (in SW engineer at least) works "by magic", and if there are software engineers out there that don't understand what a reducer applying a filter to list does, well, I'd be very concerned for the future of their careers.

Also, if they would be more into functional programming (languages like Haskell, Elm, or simply following the paradigm in JS), then they will know that currying by default has many advantages and the only thing that would be surprising or seemed weird is passing more than 1 param to a fn. 

Finally, they'll feel reassured when they see, on an unexpected path that was not supposed to happen and otherwise would fail silently, a nasty error was thrown.. then, since errors are reported, they could proceed to write another case to improve the UX for that unexpected path.

josemigallas: Tell that to [Magic Software](www.magicsoftware.com) 😉 
Jokes beside, about the error, I think it's only for us developers. If it's thrown it means we've defined Categories wrong (or Columns) I don't think the user could possible trigger it 🤔 So we shouldn't mix it with the UI I guess.

didierofrivia: That's true, the user wouldn't trigger it unless deliberately with dev tools. However, I added the error that way because I've been in a similar situation in the past with an stale state (cached, and not refreshed) and new data coming from an API hydrating a component props

josemigallas: Yes, sorry that was a bit misleading. The error makes sense but not the TODO comment about adding a toast.

PR #1939, opened 2020-06-02
Add applications count and admin name to GET /accounts response

josemigallas:

For Portafly's Accounts Listing we need 2 more fields:
- The count of applications for each account
- The name of the admin

After adding the new fields in the response, we need to update the DAL in portafly to match the same name (`apps_count` and `admin_name`).

Tasks:
- [x] Add a new column to Accounts for the counter: #1967
- [x] Back fill Accounts using a rake task
- [x] Add bought_cinstances_count to the response
- [ ] Tests in spec/acceptance/api/account_spec.rb
- [x] Tests in test/unit/tasks
- [x] Add admin_name field to the response
- [x] Update DAL in Portafly to include the new fields (verify it renders in Dev Accounts Table)
- [x] Update Account index in Porta (and other places probably) to benefit from the new column

[THREESCALE-5328: Get the applications count for each account](https://issues.redhat.com/browse/THREESCALE-5328)

**Verification steps:**
Try it locally with curl:
```
$ curl -v  -X GET "http://provider-admin.example.com.lvh.me:3000/admin/api/accounts.json?access_token=<TOKEN>
```
The response must contain the new fields.

### DIFF
--- a/app/lib/backend/storage_rewrite.rb
+++ b/app/lib/backend/storage_rewrite.rb
@@ -63,7 +63,7 @@ module Backend
       metrics = provider.metrics.includes(:service, :parent)
       usage_limits = provider.usage_limits.includes(plan: :service)
 
-      total_count = services.size + cinstances.size + metrics.size + usage_limits.size + bought_cinstances.size
+      total_count = services.size + cinstances.size + metrics.size + usage_limits.size + provider.bought_cinstances_count
 
       index = 0
       progress = -> do
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -581,6 +581,10 @@ class Account < ApplicationRecord
 
   private
 
+  def reset_counter_cache
+    update_column(:bought_cinstances_count, bought_cinstances.size) if persisted?
+  end
+
   def validate_timezone
     tz = ActiveSupport::TimeZone.new(timezone)
 
@@ -613,6 +617,4 @@ class Account < ApplicationRecord
 
     errors.add :master, 'can be only one' if scope.exists?(master: true)
   end
-
-  protected
 end
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -102,7 +102,12 @@ class Account < ApplicationRecord
 
   before_validation(on: :create, if: :provider?) { generate_s3_prefix }
   before_validation(on: :create, if: :provider?) { generate_domains }
+  before_validation(on: :create) { default_bought_cinstances_count }
   before_create :generate_site_access_code
+
+  def default_bought_cinstances_count
+    self.bought_cinstances_count = 0
+  end
 
   attr_protected :master, :provider, :buyer, :from_email, :vat_rate, :sample_data, :default_service_id, :s3_prefix,
                  :provider_account_id, :paid_at, :paid, :signs_legal_terms, :tenant_id, :default_account_plan_id,
--- a/app/models/contract.rb
+++ b/app/models/contract.rb
@@ -39,7 +39,7 @@ class Contract < ApplicationRecord
   validates :user_key, length: { maximum: 256 }
 
   # TODO: rename to buyer_account and remove alias
-  belongs_to :user_account, class_name: 'Account', autosave: false, counter_cache: :bought_cinstances_count
+  belongs_to :user_account, class_name: 'Account', autosave: false
 
   alias buyer_account user_account
   alias buyer         user_account
@@ -296,7 +296,7 @@ class Contract < ApplicationRecord
   private
 
   def reset_counter_cache_for
-    [:plan].freeze
+    [:plan, :account].freeze
   end
 
   def update_counter_cache?(association_name)
--- a/app/models/contract.rb
+++ b/app/models/contract.rb
@@ -296,7 +296,7 @@ class Contract < ApplicationRecord
   private
 
   def reset_counter_cache_for
-    [:plan, :account].freeze
+    %i[plan account].freeze
   end
 
   def update_counter_cache?(association_name)
--- a/app/models/contract.rb
+++ b/app/models/contract.rb
@@ -96,7 +96,7 @@ class Contract < ApplicationRecord
 
   def self.by_plan_type(type)
 
-    plans = Plan.unscoped.uniq.joins { pricing_rules.outer }
+    plans = Plan.unscoped.distinct.joins { pricing_rules.outer }
 
     plan_type = case type.to_s
                 when 'free'
--- a/app/models/contract.rb
+++ b/app/models/contract.rb
@@ -250,15 +250,6 @@ class Contract < ApplicationRecord
 
   protected
 
-  def update_counter_cache?(association_name)
-    case association_name
-    when :plan
-      !provider_account&.scheduled_for_deletion? && !issuer&.deleted?
-    else
-      true
-    end
-  end
-
   def correct_plan_subclass?
     if plan && (not plan.is_a?(Plan))
       errors.add(:plan, 'wrong plan subclass')
--- a/app/models/contract.rb
+++ b/app/models/contract.rb
@@ -39,7 +39,7 @@ class Contract < ApplicationRecord
   validates :user_key, length: { maximum: 256 }
 
   # TODO: rename to buyer_account and remove alias
-  belongs_to :user_account, class_name: 'Account', autosave: false
+  belongs_to :user_account, class_name: 'Account', autosave: false, counter_cache: :bought_cinstances_count
 
   alias buyer_account user_account
   alias buyer         user_account
--- a/app/representers/account_representer.rb
+++ b/app/representers/account_representer.rb
@@ -54,6 +54,8 @@ module AccountRepresenter
   end
 
   property :state
+  property :bought_cinstances_count
+  property :admin_display_name
 
   link :self do
     admin_api_account_url(self) unless provider?
@@ -67,5 +69,8 @@ module AccountRepresenter
     credit_card_stored?
   end
 
+  delegate :display_name, to: :first_admin, allow_nil: true
   delegate :monthly_charging_enabled, :monthly_billing_enabled, to: :settings, allow_nil: true
+
+  alias admin_display_name display_name
 end
--- a/app/views/buyers/accounts/_search_results.html.slim
+++ b/app/views/buyers/accounts/_search_results.html.slim
@@ -50,7 +50,7 @@ table class="data" id="buyer_accounts"
           td class="plan" = h account.bought_account_plan.try(:name)
 
         - if can? :see, :multiple_applications
-          td = link_to (current_user.accessible_cinstances.bought_by(account).count), admin_buyers_account_applications_path(account)
+          td = link_to account.bought_cinstances_count, admin_buyers_account_applications_path(account)
 
         td class="state" = account_states_info(account)
         td class="actions operations"
--- a/app/views/buyers/applications/_widget.html.erb
+++ b/app/views/buyers/applications/_widget.html.erb
@@ -1,5 +1,4 @@
-<%- cinstances = @account.bought_cinstances -%>
-<%- cinstance = cinstances.first! if cinstances.count == 1 -%>
+<%- cinstance = @account.bought_cinstances.first! if @account.bought_cinstances_count == 1 -%>
 
 <%- if cinstance && can?(:read, cinstance) -%>
   <div class="dashboard_bubble round">
--- a/lib/tasks/accounts.rake
+++ b/lib/tasks/accounts.rake
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 namespace :accounts do
 
   desc "Count bought cinstnaces for each account"
--- a/lib/tasks/accounts.rake
+++ b/lib/tasks/accounts.rake
@@ -2,7 +2,7 @@
 
 namespace :accounts do
 
-  desc "Count bought cinstnaces for each account"
+  desc "Reset the counter of bought_cinstances for all accounts"
   task reset_bought_cinstances_count: :environment do
     accounts = Account.select(:id)
     progress = ProgressCounter.new(accounts.count)
--- a/lib/tasks/accounts.rake
+++ b/lib/tasks/accounts.rake
@@ -4,8 +4,12 @@ namespace :accounts do
 
   desc "Count bought cinstnaces for each account"
   task count_bought_cinstances: :environment do
-    Account.select(:id).find_each do |account|
+    accounts = Account.select(:id)
+    progress = ProgressCounter.new(accounts.count)
+
+    accounts.find_each do |account|
       Account.reset_counters(account.id, :bought_cinstances)
+      progress.call
     end
   end
 
--- a/lib/tasks/accounts.rake
+++ b/lib/tasks/accounts.rake
@@ -1,0 +1,10 @@
+namespace :accounts do
+
+  desc "Count bought cinstnaces for each account"
+  task count_bought_cinstances: :environment do
+    Account.select(:id).find_each do |account|
+      Account.reset_counters(account.id, :bought_cinstances)
+    end
+  end
+
+end
--- a/lib/tasks/accounts.rake
+++ b/lib/tasks/accounts.rake
@@ -3,7 +3,7 @@
 namespace :accounts do
 
   desc "Count bought cinstnaces for each account"
-  task count_bought_cinstances: :environment do
+  task reset_bought_cinstances_count: :environment do
     accounts = Account.select(:id)
     progress = ProgressCounter.new(accounts.count)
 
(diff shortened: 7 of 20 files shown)


app/representers/account_representer.rb
```diff
@@ -18,6 +18,7 @@ module AccountRepresenter
   end
 
   with_options(if: ->(*) { provider? }) do
+    # TODO: property :admin_name
```
josemigallas, line 21: This will go in the next PR

app/representers/account_representer.rb
```diff
@@ -54,6 +55,8 @@ module AccountRepresenter
   end
 
   property :state
+  property :bought_cinstances_count
+  property :admin_name
```
Martouta, line 59:
```suggestion
  property :admin_user_name
```
WDYT? because `admin` is not clear that it is refering to `admin user`, specially when there are account attributes called `admin_domain` and `admin_base_url`

Martouta, line 59: These 2 need to be tested in:
https://github.com/3scale/porta/blob/23c541044bf72ea2e9632129877ef0d393e9e32c/spec/acceptance/api/account_spec.rb#L5

db/migrate/20200615130642_add_bought_cinstances_count_to_accounts.rb
```diff
@@ -0,0 +1,22 @@
+class AddBoughtCinstancesCountToAccounts < ActiveRecord::Migration[5.0]
```
Martouta, line 1: This needs to be migrated for the 3 schemas 😝

app/representers/account_representer.rb
josemigallas: The problem with `admin_user_name` is that it can be confused with the admin's `user_name`. It could be `admin_display_name`?

db/migrate/20200615130642_add_bought_cinstances_count_to_accounts.rb
Martouta: And please ensure that it works for the 3 DBs also when it contains data for the affected tables/columns

app/representers/account_representer.rb
Martouta: @josemigallas yes that's good enough I guess

app/representers/account_representer.rb
```diff
@@ -67,5 +70,9 @@ def credit_card_stored
     credit_card_stored?
   end
 
+  def admin_name
+    account.admins.first.try(:display_name)
```
Martouta, line 74: You may want to use one of these methods instead of doing `.admins.first` here 😄 
https://github.com/3scale/porta/blob/23c541044bf72ea2e9632129877ef0d393e9e32c/app/models/account.rb#L192-L198

Martouta, line 74: `account.` is not necessary 😄 you are kinda already inside the Account 😉 
So it would be just `first_admin.try!(:display_name)`, but it is even better if you delegate IMO 😄 
like:
```
delegate :display_name, to: :first_admin, allow_nil: true
alias admin_display_name display_name
```

db/migrate/20200615130642_add_bought_cinstances_count_to_accounts.rb
```diff
@@ -0,0 +1,22 @@
...
+  def change
+    safety_assured do
+      add_column :accounts, :bought_cinstances_count, :integer, :null => false, :default => 0
```
Martouta, line 4: This is a blocking migration so how do we plan to run this for SaaS?

hallelujah: And we should not add anymore column to accounts, this table is terribly huge and is used every time

app/models/contract.rb
```diff
@@ -39,7 +39,7 @@ class Contract < ApplicationRecord
   validates :user_key, length: { maximum: 256 }
 
   # TODO: rename to buyer_account and remove alias
-  belongs_to :user_account, class_name: 'Account', autosave: false
+  belongs_to :user_account, class_name: 'Account', autosave: false, counter_cache: true
```
Martouta, line 42:
```suggestion
  belongs_to :user_account, class_name: 'Account', autosave: false, counter_cache: :bought_cinstances_count
```

db/migrate/20200615130642_add_bought_cinstances_count_to_accounts.rb
macejmic: @hallelujah and where you would create that counter attribute?

lib/tasks/accounts.rake
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+namespace :accounts do
+
+  desc "Count bought cinstnaces for each account"
```
Martouta, line 5:
```suggestion
  desc "Reset the counter of bought_cinstances for all accounts"
```

lib/tasks/accounts.rake
```diff
@@ -0,0 +1,12 @@
...
+  task count_bought_cinstances: :environment do
```
Martouta, line 6:
```suggestion
  task reset_bought_cinstances_count: :environment do
```
or:
```ruby
  task reset_bought_cinstances_counter: :environment do
```

Martouta, line 6: Add a test in [/test/unit/tasks/...](https://github.com/3scale/porta/tree/master/test/unit/tasks)

test/unit/tasks/accounts_test.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+module Tasks
+  class AccountsTest < ActiveSupport::TestCase
+    test 'accounts:reset_bought_cinstances_count' do
+      account = FactoryBot.create(:account)
+      account.update(bought_cinstances_count: nil)
```
Martouta, line 9: This works:
```ruby
# frozen_string_literal: true

require 'test_helper'

module Tasks
  class AccountsTest < ActiveSupport::TestCase
    test 'accounts:reset_bought_cinstances_count' do
      account = FactoryBot.create(:simple_buyer)
      FactoryBot.create(:cinstance, user_account: account)

      sql_update_nil = <<-SQL
        UPDATE  accounts
        SET     bought_cinstances_count = NULL
        WHERE   accounts.id = #{account.id}
      SQL
      ActiveRecord::Base.connection.execute(sql_update_nil)

      assert_nil account.reload.bought_cinstances_count

      execute_rake_task 'accounts.rake', 'accounts:reset_bought_cinstances_count'

      account.reload
      assert_equal account.bought_cinstances_count, account.bought_cinstances.count
    end
  end
end
```

lib/tasks/accounts.rake
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+require 'progress_counter'
+
+namespace :accounts do
+
+  desc "Reset the counter of bought_cinstances for all accounts"
+  task reset_bought_cinstances_count: :environment do
+    accounts = Account.where.not(state: :scheduled_for_deletion)
```
Martouta, line 9:
```suggestion
    accounts = Account.without_deleted
```
Using:
https://github.com/3scale/porta/blob/26657694321c1d922879efc69b033d8c9c6b1e9f/app/models/account/states.rb#L96

app/models/account.rb
```diff
@@ -102,8 +102,13 @@ def self.columns
 
   before_validation(on: :create, if: :provider?) { generate_s3_prefix }
   before_validation(on: :create, if: :provider?) { generate_domains }
+  before_validation(on: :create) { default_bought_cinstances_count }
```
Martouta, line 105: This scenario should have a test too 😄

app/views/buyers/accounts/_search_results.html.slim
```diff
@@ -50,7 +50,7 @@ table class="data" id="buyer_accounts"
           td class="plan" = h account.bought_account_plan.try(:name)
 
         - if can? :see, :multiple_applications
-          td = link_to (current_user.accessible_cinstances.bought_by(account).count), admin_buyers_account_applications_path(account)
+          td = link_to account.bought_cinstances_count, admin_buyers_account_applications_path(account)
```
Martouta, line 53: I am not sure if it is the same 🤔 `account.bought_cinstances_count` is the number of bought cinstances. `current_user.accessible_cinstances.bought_by(account).count)` is the number of bought **accessible** cinstances. What I do not know is a non-accessible cinstances can be bought at any point in time.

test/unit/account_test.rb
```diff
@@ -864,4 +864,23 @@ def test_smart_destroy_master
     assert master_account.reload
     refute master_account.reload.scheduled_for_deletion?
   end
+
+  test 'bought_cinstances_count when buying a cinstance' do
+    account = FactoryBot.create(:simple_buyer, provider_account: master_account)
+
+    assert_difference -> { account.bought_cinstances_count }, 1 do
+      account.buy!(master_account.default_service.published_plans.first)
```
Martouta, line 872:
```suggestion
      account.buy!(master_account.default_service.published_plans.first!)
```

test/unit/account_test.rb
```diff
@@ -864,4 +864,23 @@ def test_smart_destroy_master
     assert master_account.reload
     refute master_account.reload.scheduled_for_deletion?
   end
+
+  test 'bought_cinstances_count when buying a cinstance' do
+    account = FactoryBot.create(:simple_buyer, provider_account: master_account)
```
Martouta, line 869:
```suggestion
    account = FactoryBot.create(:simple_provider, provider_account: master_account)
```
Because a tenant is a buyer of master which is what you are doing 🤔

test/unit/account_test.rb
```diff
@@ -864,4 +864,23 @@ def test_smart_destroy_master
...
+
+    assert_difference -> { account.bought_cinstances_count }, 1 do
+      account.buy!(master_account.default_service.published_plans.first)
+      account.reload.bought_cinstances_count
+    end
+  end
+
+  test 'bought_cinstances_count when deleting a cinstance' do
+    account = FactoryBot.create(:simple_buyer, provider_account: master_account)
+    cinstance = account.buy!(master_account.default_service.published_plans.first)
```
Martouta, line 879: same as https://github.com/3scale/porta/pull/1939/files#r442742867

test/unit/account_test.rb
```diff
@@ -864,4 +864,23 @@ def test_smart_destroy_master
...
+    account = FactoryBot.create(:simple_buyer, provider_account: master_account)
+
+    assert_difference -> { account.bought_cinstances_count }, 1 do
+      account.buy!(master_account.default_service.published_plans.first)
+      account.reload.bought_cinstances_count
+    end
+  end
+
+  test 'bought_cinstances_count when deleting a cinstance' do
+    account = FactoryBot.create(:simple_buyer, provider_account: master_account)
```
Martouta, line 878: same as https://github.com/3scale/porta/pull/1939/files#r442743238

db/migrate/20200615130642_add_bought_cinstances_count_to_accounts.rb
Martouta: About this: we decided that now 'arbitrarily' splitting the accounts table by this criteria of just being a new attribute will just make the problem worse, so we should add this column to accounts + make a Jira issue to split the accounts table under a well-thought criteria if we end up doing so.

app/views/buyers/accounts/_search_results.html.slim
josemigallas: That's a bit worrying because the whole point of the new column is to provide Portafly with this very value.

Martouta: I think that it is correct but I am not 100% sure so I left the comment in case that someone can be sure of that 😄 
I believe that only accessible plans can be bought, what I do not know for sure if a plan that has been accessible, and has bought plans can become inaccessible later on. I hope not for our mental health 😛

app/lib/backend/storage_rewrite.rb
```diff
@@ -63,7 +63,7 @@ def rewrite_provider(id)
       metrics = provider.metrics.includes(:service, :parent)
       usage_limits = provider.usage_limits.includes(plan: :service)
 
-      total_count = services.size + cinstances.size + metrics.size + usage_limits.size + bought_cinstances.size
+      total_count = services.size + cinstances.size + metrics.size + usage_limits.size + provider.bought_cinstances_count
```
hallelujah, line 66: I wonder, `services.size`, `cinstances.size` etc ... they are loading the records but `bought_cinstances_count` is not loading `bought_cinstances`

Why this change? Don't we need `bought_cinstances` later on? If we do then I think there is no need to use the cached counter

app/models/contract.rb
```diff
@@ -296,7 +296,7 @@ def change_plan_internal(new_plan, &block)
   private
 
   def reset_counter_cache_for
-    [:plan].freeze
+    [:plan, :account].freeze
```
Martouta, line 299: First, RuboMarta strikes 😝 :
```suggestion
    %i[plan account].freeze
```
Now my "questions/suggestions":
1. Is it `account` or `user_account` ? why?
2. All this needs a test 😄 
3. Shouldn't the next method (`update_counter_cache?`) have this same association in the first `when` ? if not, why? because I believe that it shouldn't update this cache if the account is being deleted 😕

app/lib/backend/storage_rewrite.rb
hallelujah: Also the counter cache is made so you do not have to use `#bought_cinstances_count` but plainly do `bought_cinstances.size`

> The method “size” will use the counter_cache if its present, where as using “.count” itself would do the actual sql count.

app/views/buyers/accounts/_search_results.html.slim
guicassolato: Strictly, it's not the same. The access is granted to a user for a given service (or set of services), not for buyer accounts or applications. Consider the following:
- current_user has access to services A and B, but not C
- account X bought cinstances of A and C (1 of each)
- `current_user.accessible_cinstances.bought_by(account_x).count` should be 1
- `account_x. bought_cinstances_count` will return 2

app/models/contract.rb
josemigallas: @macejmic

app/views/buyers/applications/_widget.html.erb
```diff
@@ -1,5 +1,4 @@
-<%- cinstances = @account.bought_cinstances -%>
-<%- cinstance = cinstances.first! if cinstances.count == 1 -%>
+<%- cinstance = @account.bought_cinstances.first! if @account.bought_cinstances_count == 1 -%>
```
Martouta, line 1: I wonder why not `<%- cinstance = @account.bought_cinstances.first -%>` 🤔 because I guess that internally the DB should use the index we have on cinstances user_account_id and it would quickly see that it doesn't have any 🤔

app/models/account.rb
```diff
@@ -576,6 +581,10 @@ def independent_mapping_rules_enabled?
 
   private
 
+  def reset_counter_cache
+    update_column(:bought_cinstances_count, bought_cinstances.size) if persisted?
```
Martouta, line 585: why is this necessary? 😕 shouldn't it do this automatically without overriding this method?

josemigallas: @macejmic

app/views/buyers/applications/_widget.html.erb
josemigallas: We should ask @macejmic 
https://github.com/3scale/system/blame/master/app/views/buyers/applications/_widget.html.erb

macejmic: `<%- cinstance = @account.bought_cinstances.first -%>` looks good to me.

app/models/account.rb
macejmic: `counter_cache` is not defined for `bought_cinstances` so yes, it has to be done manually, that's the concept of the manual counter cache.

app/models/contract.rb
macejmic: `alias account user_account` so it does not matter, if it's `account` or `user_account`.

macejmic: 3. https://github.com/3scale/porta/blob/master/app/models/contract.rb#L302

so the counter is not being updated when the provider account is being deleted.

macejmic: 2. this is an "integration" unit test that covers the main use case:

```
test 'bought_cinstances_count when deleting a cinstance' do
    account = FactoryBot.create(:simple_provider, provider_account: master_account)
    cinstance = account.buy!(master_account.default_service.published_plans.first!)

    assert_difference -> { account.reload.bought_cinstances_count }, -1 do
      cinstance.destroy
      account.reload.bought_cinstances_count
    end
  end
```

all other are already covered..

josemigallas:
Apps no longer part of this screen so it's reduced to admin_name. I'll leave this branch for future reference but close the PR in favor of #1976

PR #1944, opened 2020-06-03
Add empty state generic components

josemigallas:
Adding the EmptyState components that will be used first by Accounts Listing but then for any page/data-list page.

The Page version is meant to occupy the whole page.
The Table version is meant to be inside a table after loading/filtersing and to have a button like "Try again", "Refresh", "Clear all filters", etc.


didierofrivia:
Probably not for this PR, but we should get rid of `portafly/src/components/developer-accounts/DeveloperAccountsTable.tsx`, anything related to a page, should not be in `components` directory

PR #1949, opened 2020-06-05
[portafly] Patch dev accounts collection

didierofrivia:
This is the PortaFly OAS for the update of a developer accounts collection
The definition can be found in https://issues.redhat.com/browse/THREESCALE-5330

Fixes https://issues.redhat.com/browse/THREESCALE-5348

### DIFF
--- a/portafly/portafly_api.yml
+++ b/portafly/portafly_api.yml
@@ -18,6 +18,27 @@ paths:
           $ref: '#/components/responses/developerAccountList'
       operationId: developerAccountList
       summary: Get developer account list
+    patch:
+      requestBody:
+        content:
+          application/json:
+            schema:
+              type: array
+              items:
+                $ref: '#/components/schemas/DeveloperAccount'
+            examples:
+              Patch the state of 2 Dev Accounts:
+                value:
+                - id: 37
+                  state: active
+                - id: 18
+                  state: active
+        required: true
+      responses:
+        "204":
+          description: Resource updated correctly
+      summary: Partially updates developer accounts
+      description: Applies partial modifications to the Developer Accounts collection
     parameters:
     - name: page
       description: Page in the paginated list. Defaults to 1.
--- a/portafly/portafly_api.yml
+++ b/portafly/portafly_api.yml
@@ -89,8 +89,8 @@ components:
           type: integer
       example:
         id: 1
-        created_at: 2019-10-18T05:13:26Z
-        updated_at: 2019-10-18T05:13:27Z
+        created_at: 2019-10-18T05:13:26.000Z
+        updated_at: 2019-10-18T05:13:27.000Z
         state: approved
         org_name: Umbrella Corp.
         admin_name: Oswell E. Spencer


portafly/portafly_api.yml
```diff
@@ -18,6 +18,27 @@ paths:
...
+            examples:
+              Patch the state of 2 Dev Accounts:
+                value:
+                - id: 37
+                  state: active
+                - id: 18
+                  state: active
+        required: true
+      responses:
+        "204":
```
josemigallas, line 38: If the procedure is async (which I believe it is), it probably should return `200`.

didierofrivia: Hmm I would be quick to agree about the send email endpoint which would trigger a worker... But this update I'm not so sure. WDYT @hallelujah @guicassolato ?

portafly/portafly_api.yml
```diff
@@ -18,6 +18,27 @@ paths:
...
+          description: Resource updated correctly
+      summary: Partially updates developer accounts
```
hallelujah, line 40: What does mean "Partially" it is not clear.
To my point of view update is changing some state so it can't be partial (like maybe it has changed or maybe not)

portafly/portafly_api.yml
hallelujah: Well I can't really answer because of the unclear summary but if it is an update then it should return 200 with some content. I really doubt about 204 no content.

portafly/portafly_api.yml
didierofrivia: Partial in the sense that you are not sending the entire record back, just the fields you want to update, not the entire record. https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PATCH

portafly/portafly_api.yml
didierofrivia: Usually, when you do a PATCH request, the successful response code is 204. More info: https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PATCH

portafly/portafly_api.yml
hallelujah: Not true. They just use an example and even explained it:

> A successful response is indicated by any 2xx status code.
>
> In the example below a 204 response code is used, because the response does not > carry a payload body. A 200 response could have contained a payload body.

portafly/portafly_api.yml
hallelujah: Not true. They just use an example and even explained it:

> A successful response is indicated by any 2xx status code.
>
> In the example below a 204 response code is used, because the response does not carry a payload body. A 200 response could have contained a payload body.

portafly/portafly_api.yml
hallelujah: Why should we ever do this? It looks broken design to me

hallelujah: Scenario: What happens if one record PATCHING is successful and the other not?

didierofrivia: I'm not saying it's the only way, just following practices and matching our use case since we have the changes in our state, if it's 204 back, we don't need any data from the server. And if one fails and the other not, you receive an error message back from the server, we show it that it failed... not big deal, maybe even in that case update the list.
SO if we want instead of having a 200 and return the records, should we return the records changed or all of them? If it fails, we should return the failed ones, the good ones, just an error message?

josemigallas: You send only the field you want to update, then the backend returns the whole collection (updated). Is that what you mean @hallelujah ?

portafly/portafly_api.yml
josemigallas: If we go with optimism, we update everything locally and return 204, this is more React-ty because the update is almost immediate. If we wanna make it super reliable (but also slower and heavier) we return 200 with the updated collection.

hallelujah: React does not have to take into account the body of the response if you do not want but an API has the responsibility of covering all use case. And for sure not sending a response is not the good way to do it

portafly/portafly_api.yml
hallelujah: I mean, by default you send all the resource.

First scenario:

- You update multiple resources (which is an anti pattern, but lets live with that for the sake of the argument)
- one of the resource fail to update, the other are updated. What is the correct status code if there is no body to tell you what really happens

Another scenario:

- all the resource updated but another process (outside of this React component) updated another attribute of this resource. Not sending back the resource in the body will lead for sure of a desync.

hallelujah: https://blog.ploeh.dk/2013/04/30/rest-lesson-learned-avoid-204-responses/

portafly/portafly_api.yml
hallelujah: API should not be aware of how it is used. Your React component decide to use or not the Body.

If we want a better design, React should send a Header to tell the server no need of the body: e.g. `Accept: none` BUT I do not know if this exist in HTTP spec

PR #1956, opened 2020-06-08
Extract isMultitenant as an Env variable

josemigallas:

Right now `isMultitenant` is a hardcoded value.

[THREESCALE-5326: Distinguish multitenant accounts](https://issues.redhat.com/browse/THREESCALE-5326)

### DIFF
--- a/portafly/src/components/pages/accounts/utils/accountsTableDataFactory.tsx
+++ b/portafly/src/components/pages/accounts/utils/accountsTableDataFactory.tsx
@@ -2,9 +2,10 @@ import React from 'react'
 import { sortable } from '@patternfly/react-table'
 import { IDeveloperAccount } from 'types'
 import { TFunction } from 'i18next'
-import { ActionButtonImpersonate } from 'components'
+import { ActionButtonImpersonate } from 'components/pages/accounts'
 
-const generateRows = (accounts: IDeveloperAccount[], isMultitenant = false) => {
+const generateRows = (accounts: IDeveloperAccount[]) => {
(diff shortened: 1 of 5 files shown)


didierofrivia:
Could you change the base to `master` in order to see just the changes that don't involve accounts page?

portafly/src/components/pages/accounts/utils/accountsTableDataFactory.tsx
```diff
@@ -2,9 +2,10 @@ import React from 'react'
 import { sortable } from '@patternfly/react-table'
 import { IDeveloperAccount } from 'types'
 import { TFunction } from 'i18next'
-import { ActionButtonImpersonate } from 'components'
+import { ActionButtonImpersonate } from 'components/pages/accounts'
 
-const generateRows = (accounts: IDeveloperAccount[], isMultitenant = false) => {
+const generateRows = (accounts: IDeveloperAccount[]) => {
+  const isMultitenant = process.env.REACT_APP_MULTITENANT === 'true'
```
didierofrivia, line 8:
```suggestion
  const isMultitenant = process.env.REACT_APP_MULTITENANT
```

josemigallas: The problem I see with this is if I wrongly set my var to `REACT_APP_MULTITENANT="fals"` it's going to be read as `true`.

didierofrivia: That problem would denote a poorly setup/orchestration. Specially when this will be managed by OpenShift or any other K8s solution

PR #1960, opened 2020-06-09
🏗 Stop using Listing when it's an Index

josemigallas:

We need to stop using "Listing" in our codebase, it's the index page.

Part of [THREESCALE-4724](https://issues.redhat.com/browse/THREESCALE-4724).

It's pointing to its base branch but it could be merge into master after.


josemigallas:
> AccountsIndexPage feels a bit double to me (AccountsPage would do I think) but this works too.

That was my first option but I finally choose the current name for consistency with the other pages:
* AccountsIndexPage
* AccountsShowPage
* AccountsEditPage

PR #1962, opened 2020-06-10
[PortaFly] Export as CSV button

didierofrivia:

Adds the export as CSV feature

fixes https://issues.redhat.com/browse/THREESCALE-5324

**Verification steps** 
1. Go to Accounts -> Listing
2. Click "Export all accounts"
3. It should have downloaded an `accounts.csv` file listing the table data


portafly/src/components/pages/accounts/AccountsIndexPage.tsx
```diff
@@ -43,13 +48,7 @@ const AccountsIndexPage: React.FunctionComponent = () => {
...
           <FlexItem breakpointMods={[{ modifier: FlexModifiers['align-right'] }]}>
             <CreateAccountButton />
-            <Button
-              variant="link"
-              icon={<ExportIcon />}
-              aria-label={t('export_accounts_button_aria_label')}
-            >
-              {t('export_accounts_button')}
-            </Button>
+            {accounts && <ExportAccountsButton data={accounts} />}
```
josemigallas, line 51: How about disabling instead of hiding?
@cntlsn opinion from UX perspective?

portafly/src/i18n/locales/en/audience/accounts/listing.yml
```diff
@@ -58,6 +58,8 @@ accounts_table:
  applications_header_description: Column listing applications linked to accounts, on a 1:1 relationship
  state_header: State
  state_header_description: Column listing the state of the accounts
+ created_header: Created at
+ updated_header: Updated at
```
josemigallas, line 62:
```suggestion
 created_header: Created at
 created_header_description: ...
 updated_header: Updated at
 updated_header_description: ...
```

josemigallas: But I think we should export the data as is, with the property names instead of translations.

portafly/src/tests/components/pages/accounts/AccountsIndexPage.test.tsx
```diff
@@ -21,11 +21,22 @@ it('should have a button to create an account', () => {
 })
 
 it('should be able to export all accounts', () => {
-  const { getByRole } = setup({})
-  const exportAccountsBtn = getByRole('button', { name: 'export_accounts_button_aria_label' })
+  const { getByRole } = setup(
```
josemigallas, line 24:
```suggestion
  const { getByRole } = setup({ data: developerAccounts })
```

portafly/src/components/pages/accounts/AccountsIndexPage.tsx
didierofrivia: That makes total sense, what do you think @cntlsn ?

portafly/src/tests/components/pages/accounts/AccountsIndexPage.test.tsx
```diff
@@ -21,11 +21,22 @@ it('should have a button to create an account', () => {
...
+        created_at: '2019-10-18T05:13:26Z',
+        updated_at: '2019-10-18T05:13:27Z'
+      }]
+    }
+  )
+  const exportAccountsBtn = getByRole('link', { name: 'export_accounts_button_aria_label' })
   expect(exportAccountsBtn).toHaveTextContent('export_accounts_button')
-
-  // TODO: test export?
+  expect(exportAccountsBtn.getAttribute('href')).toContain('"ID","accounts_table.admin_header","accounts_table.group_header","accounts_table.state_header","accounts_table.applications_header","accounts_table.created_header","accounts_table.updated_header"\n"1","Root","Linux","active","42","2019-10-18T05:13:26Z",""')
```
josemigallas, line 39: Does this mean if we have 100 accounts, the href will contain all?

portafly/src/i18n/locales/en/audience/accounts/listing.yml
didierofrivia: this too, @thomasmaas @cntlsn, any comments?

portafly/src/tests/components/pages/accounts/AccountsIndexPage.test.tsx
didierofrivia: Nope, it eventually converts it to a blob

portafly/src/components/pages/accounts/ExportAccountsButton.tsx
```diff
@@ -0,0 +1,34 @@
...
+    { label: t('accounts_table.created_header'), key: 'created_at' },
+    { label: t('accounts_table.updated_header'), key: 'update_at' }
+  ]
+
+  return (
+    <CSVLink
+      filename="accounts.csv"
+      headers={exportHeaders}
+      data={data}
+      className="pf-c-button pf-m-link"
```
josemigallas, line 25: I must discourage this, let's try to wrap the CSVDownload inside a PF Button instead.

didierofrivia: I've changed to use the CSVDownload component, however, we are losing some features like naming the download file. Also, it glitches the UI when clicked if not set to `_blank`, which opens a new tab.

portafly/src/components/pages/accounts/ExportAccountsButton.tsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react'
+import { useTranslation } from 'i18n/useTranslation'
+import { ExportIcon } from '@patternfly/react-icons'
+import { CSVDownload } from 'react-csv'
+import { IDeveloperAccount } from 'types'
+import { Button } from '@patternfly/react-core'
+import { render } from '@testing-library/react'
+
+const ExportAccountsButton = ({ data }: {data: IDeveloperAccount[] | undefined}) => {
```
josemigallas, line 9:
```suggestion
interface Props {
  data?: IDeveloperAccount[]
}

const ExportAccountsButton: React.FunctionComponent<Props> = ({ data }) => {
```

portafly/src/i18n/locales/en/audience/accounts/listing.yml
didierofrivia: We also need to consider that this kind of CSV files will be used by people who are not technical or are expecting to see the same as in the UI.

portafly/src/components/pages/accounts/ExportAccountsButton.tsx
```diff
@@ -0,0 +1,44 @@
...
+    { label: t('accounts_table.updated_header'), key: 'update_at' }
+  ]
+
+  return (
+    <Button
+      isDisabled={!data}
+      variant="link"
+      icon={<ExportIcon />}
+      aria-label={t('export_accounts_button_aria_label')}
+      onClick={() => !!data && render(
```
josemigallas, line 31: This is very smelly, no offense. Instead of double bang use a casting, since the type is well-known at this point `data as IDeveloperAccount[]`. And I am not at all convinced about using a testing render method... The correct way is to render it as a sibling.

But after running it locally and playing a bit with it, everything points to using <CSVLink> instead :/ I think it's the lesser evil after all.

didierofrivia: Shit. It really smells, I didn't want to use that particular one TBH. I think it was auto-completed and auto-imported by VS. Thanks for spotting it!

didierofrivia: it doesn't smell, it stinks! a strong offensive odor right there

didierofrivia: I think the lesser evil here would be to use the `CSVLink` component

josemigallas: > I think the lesser evil here would be to use the `CSVLink` component

That's what I meant exactly! 😅  But I realize now I left a gap:
> [...] everything points to using _`CSVLink`_  instead

josemigallas: Screen Shot 2020-06-17 at 09 04 26
Jeez, it's there but it's invisible. I forgot to close the tag I guess!

portafly/src/i18n/locales/en/audience/accounts/listing.yml
josemigallas: That's a good point

portafly/src/components/pages/accounts/ExportAccountsButton.tsx
```diff
@@ -0,0 +1,44 @@
...
+  const { t } = useTranslation('accountsIndex')
+
+  const exportHeaders = [
+    { label: 'ID', key: 'id' },
+    { label: t('accounts_table.admin_header'), key: 'admin_name' },
+    { label: t('accounts_table.group_header'), key: 'org_name' },
+    { label: t('accounts_table.state_header'), key: 'state' },
+    { label: t('accounts_table.applications_header'), key: 'apps_count' },
+    { label: t('accounts_table.created_header'), key: 'created_at' },
+    { label: t('accounts_table.updated_header'), key: 'update_at' }
```
josemigallas, line 22: Typo!
```suggestion
    { label: t('accounts_table.updated_header'), key: 'updated_at' }
```

portafly/src/tests/components/pages/accounts/AccountsIndexPage.test.tsx
```diff
@@ -21,11 +23,20 @@ it('should have a button to create an account', () => {
 })
 
 it('should be able to export all accounts', () => {
-  const { getByRole } = setup({})
+  const accounts = [factories.DeveloperAccount.build()]
```
josemigallas, line 26: Remember we have already the examples available, no need to build them again:
`import { developerAccounts } from 'tests/examples'`

didierofrivia: Yeah, but I wanted to have a smaller set, instead of the one in examples. Also, when I introduced factories, it was to craft the examples in the tests using the factories templates.

portafly/src/components/pages/accounts/ExportAccountsButton.tsx
```diff
@@ -0,0 +1,50 @@
...
+      <CSVLink
+        filename="accounts.csv"
+        headers={exportHeaders}
+        data={data}
+        className="pf-c-button pf-m-link"
+        aria-label={t('export_accounts_button_aria_label')}
+      >
+        <span className="pf-c-button__icon pf-m-start"><ExportIcon /></span>
+        {t('export_accounts_button')}
+      </CSVLink>
```
josemigallas, line 35: Instead of rendering the different button I think we could [override the onClick](https://github.com/react-csv/react-csv#--onclick-props) and add a disabled style:
```suggestion
// Button should be disabled also when there are no accounts (empty)
// const isDisabled = data && data.length
      <CSVLink
        filename="accounts.csv"
        headers={exportHeaders}
        data={data}
        className={`pf-c-button pf-m-link ${isDisabled ? 'pf-m-disabled'} : ''`}
        aria-label={t('export_accounts_button_aria_label')}
        onClick={() => !isDisabled}
      >
        <span className="pf-c-button__icon pf-m-start"><ExportIcon /></span>
        {t('export_accounts_button')}
      </CSVLink>
```

portafly/src/i18n/locales/en/audience/accounts/listing.yml
josemigallas: But still we're missing the descriptions.

portafly/src/tests/components/pages/accounts/AccountsIndexPage.test.tsx
```diff
@@ -21,11 +22,15 @@ it('should have a button to create an account', () => {
...
+  const { getByText } = setup({ data: [factories.DeveloperAccount.build()] })
+  const exportAccountsLink = getByText('export_accounts_button')
+  expect(exportAccountsLink.getAttribute('href')).toContain('"ID","accounts_table.admin_header","accounts_table.group_header","accounts_table.state_header","accounts_table.applications_header","accounts_table.created_header","accounts_table.updated_header"\n"2","Oswell E. Spencer","Umbrella Corp.","approved","3","2019-10-18T05:13:26Z","2019-10-18T05:13:27Z"')
+})
 
-  // TODO: test export?
+it('should disable the export all button when no data is provided', () => {
+  const { getByRole } = setup({ data: undefined })
+  const exportAccountsBtn = getByRole('button', { name: 'export_accounts_button_aria_label' })
+  expect(exportAccountsBtn).toHaveAttribute('disabled')
```
josemigallas, line 33: This seems ambiguous, perhaps;
```suggestion
  expect(exportAccountsBtn.getAtttribute('disabled')).toBe(true)
```

portafly/src/i18n/locales/en/audience/accounts/listing.yml
didierofrivia: I'll leave it to the docs team

portafly/src/components/pages/accounts/ExportAccountsButton.tsx
```diff
@@ -0,0 +1,40 @@
...
+    { label: t('accounts_table.updated_header'), key: 'updated_at' }
+  ]
+
+  const isDisabled = !(data && data.length)
+
+  return (
+    <CSVLink
+      filename="accounts.csv"
+      headers={exportHeaders}
+      data={data || exportHeaders}
```
didierofrivia, line 29: if we cast here `IDeveloperAccount[]` when first renders is `undefined`, making it crash. I think it was just ok to pass it first the headers, then the full thing. Alternatively, an empty string/array would do. Either way, it won't be ever exported if empty.

PR #1963, opened 2020-06-11
Sphinx Real time indexing

hallelujah:
Accounts are now indexed in real time (asynchronously in background) when:

- Account get created, updated
- User get created, updated, deleted
- Cinstance get created, updated, deleted

CMS::Page are now indexed in real time when:
- CMS::Page get created, updated

Topic are now indexed in real time when:
- Topic get created, updated
- Post get created, updated, deleted

## SaaS

- removed Sphinx cron jobs
- Need to delete all sphinx files and rebuild the index first:
```shell
./bin/remote system preview01 system:ts:stop
```

```shell
sudo kill $(pidof searchd)
sudo rm -rf /mnt/sphinx_index/*
cd system_preview/current
BUNDLE_GEMFILE=Gemfile.prod RAILS_ENV=preview ruby2.4 /usr/local/bin/bundle exec rails ts:configure
```

```shell
./bin/remote system preview01 system:ts:start
```

```shell
# on the server in tmux because it will be long running task
cd system_preview/current
BUNDLE_GEMFILE=Gemfile.prod RAILS_ENV=preview ruby2.4 /usr/local/bin/bundle exec rails ts:rt:index
```

## QE

There is no changes in the pods. Follow upgrade documentation 

## Upgrade

1. Remove the content of the mounted volume `system-sphinx-database` in `system-sphinx` pod. You can run `rake ts:clear` inside the `system-sphinx` pod after upgrade to achieve this step.
2. Run `rake ts:rt:rebuild` in `system-sphinx` pod to rebuild the indices


app/models/cinstance.rb
```diff
@@ -391,6 +392,10 @@ def app_plan_change_should_request_credit_card?
 
   protected
 
+  def account_for_sphinx
```
guicassolato, line 395: `alias_method` maybe?

hallelujah: hmm I want it not to be public but yes why not
Thing is I do not want people to use this outside of this class

app/lib/three_scale/domain.rb
```diff
@@ -2,7 +2,7 @@ module ThreeScale
   module Domain
     def self.current_endpoint(request, host = request.host)
       host_and_port = if Rails.env.development?
-                        request.host_with_port("#{host}:#{request.port}")
```
hallelujah, line 5: This has nothing to do with this PR

features/old/buyers/accounts/searching.feature
```diff
@@ -1,4 +1,4 @@
-@search @no-txn
```
hallelujah, line 1: Yes it does not need transaction anymore :)

test/test_helpers/sphinx.rb
```diff
@@ -0,0 +1,19 @@
...
+    class << self
+
+      def real_time_run
+        init
+        start index: false
+        yield
+      ensure
+        stop
+      end
+      alias_method :rt_run, :real_time_run
```
guicassolato, line 15: I'd prefer not having the abbreviated name, if you don't mind.

Actually, I'd even propose breaking the API by doing the following this:

```ruby
  def run(realtime: false, &block)
    return super(&block) unless realtime
    realtime_run(&block)
  end

  protected

  def realtime_run
    init
    start index: false
    yield
  ensure
    stop
  end
```

I will understand of course if you prefer not to break ThinkingSphinx original API, in which case you can just ignore this suggestion.

app/indices/account_index.rb
```diff
@@ -1,31 +1,94 @@
...
+  module ForAccount
+
+    extend ActiveSupport::Concern
+
+    included do
+      after_save :index_account
+    end
+
+    def sphinx_usernames
+      users.pluck(:username).join(' ')
```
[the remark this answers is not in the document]
hallelujah: Well it does nothing in fact, pluck will just select the passed attributes

app/indices/account_index.rb
```diff
@@ -1,31 +1,94 @@
...
+      bought_cinstances.pluck(:application_id).join(' ')
+    end
+
+    def sphinx_app_names
+      bought_cinstances.pluck(:name).join(' ')
+    end
+
+    def sphinx_user_ids
+      user_ids.join(' ')
+    end
```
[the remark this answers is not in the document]
hallelujah: It is more complicated to read :)

test/test_helpers/sphinx.rb
hallelujah: Let be good citizen ;) Adding new methods are ok, in fact we should not even add it to ThinkingSphinx::Test but have our own class ... Can be done later as refactoring (imagine using another indexer than Sphinx like ElasticSearch or Postgres)

app/indices/topic_index.rb
```diff
@@ -1,21 +1,53 @@
...
+
+module TopicIndex
+  extend ActiveSupport::Concern
+
+  included do
+    after_save :sphinx_index
+  end
+
+  def sphinx_post_bodies
+    posts.pluck(:body).join(" ")
```
Martouta, line 26:
```suggestion
    posts.pluck(:body).join(' ')
```

app/workers/sphinx_indexation_worker.rb
```diff
@@ -0,0 +1,19 @@
...
+
+
+# SphinxIndexationWorker updates sphinx index for the provided model.
+# It is enqueued when:
+# - Account gets created, updated (deletion is done automatically as per https://freelancing-gods.com/thinking-sphinx/v3/real_time.html
+# - User gets created, updated, deleted
+# - Cinstance gets created, updated, deleted
+class SphinxIndexationWorker < ApplicationJob
+
+  rescue_from(ActiveJob::DeserializationError) do |exception|
```
Martouta, line 11: In which case can this happen? As far as I understood, the perform is for the model, and in what moment can it not exist anymore?

hallelujah: Imagine you update the model, sidekiq workers are down, then some process deletes the record, and sidekiq up again
It will be a GlobalId deserialization error

Martouta: ok then when this is legitimately called for a Cinstance/User that has been deleted, can we be sure that it will always be called using [the missing model](https://github.com/3scale/porta/blob/acc77228aa4f1d1ac226c78cae0e421588de6990/app/lib/missing_model.rb#L4-L26) or any other way where we won't have this DeserializationError as a real bug for those cases?

config/schedule.rb
```diff
@@ -45,17 +45,12 @@
...
 end
-
-every 1.day, :at => '6:00 am', roles: [:sphinx_server] do
-  ThreeScale::Jobs::SPHINX_INDEX.each { |task| instance_exec(task, &job_proc) }
-end
-
-# every hour would put at minute 0 which is quite busy with tasks already
-every '42 * * * *', roles: [:sphinx_server] do
-  ThreeScale::Jobs::SPHINX_DELTA.each { |task| instance_exec(task, &job_proc) }
-end
```
guicassolato, line 61: Need to remove from https://github.com/3scale/porta/blob/1ef7f18809ec074bebe396ded6436a2f4c1da64f/openshift/system/config/sidekiq_schedule.yml#L39-L51 too?

app/indices/account_index.rb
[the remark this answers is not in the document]
macejmic: agree with @hallelujah on this one.

app/workers/sphinx_indexation_worker.rb
hallelujah: The only params sent is either `Account`, `Topic` or `CMS::Page` so no worries.

It will not be called using missing model.
When it gets destroyed ThinkingSphinx automatically remove the record from the index https://freelancing-gods.com/thinking-sphinx/v3/real_time.html#model-callbacks

app/indices/account_index.rb
```diff
@@ -1,31 +1,89 @@
...
+  indexes sphinx_user_keys, as: :user_key
+  indexes sphinx_app_ids, as: :app_id
+  indexes sphinx_app_names, as: :app_name
+  indexes :id, as: :account_id
+  indexes sphinx_user_ids, as: :user_id
 
   set_property field_weights: { name: 2 }
 
-  set_property(group_concat_max_len: 32_000 - 1) if System::Database.mysql?
+  # set_property(group_concat_max_len: 32_000 - 1) if System::Database.mysql?
```
guicassolato, line 16:
```suggestion
```

features/support/sphinx.rb
```diff
@@ -2,15 +2,17 @@
...
   ::ThinkingSphinx::Test.stop
-  output = ::ThinkingSphinx::Test.start_with_autostop
+  ::ThinkingSphinx::Test.autostop
+  output = ::ThinkingSphinx::Test.start index: false
   assert ::ThinkingSphinx::Test.config.controller.running?, "thinking sphinx should be running: #{output}"
 end
 
 AfterStep '@search' do
-  ::ThinkingSphinx::Test.index
+  # ::ThinkingSphinx::Test.index
```
guicassolato, line 15:
```suggestion
```

PR #1964, opened 2020-06-11
Real time index for Topic and CMS::Page

hallelujah:
It normally also fixes issue with Oracle though not needed


test/workers/delete_account_hierarchy_worker_test.rb
```diff
@@ -43,19 +43,19 @@ def setup
...
+    perform_enqueued_jobs(only: DeleteAccountHierarchyWorker) { DeleteAccountHierarchyWorker.perform_now(provider) }
   end
 
   test 'does not perform if wrong state' do
     provider.update_column(:state, 'approved')
 
-    perform_enqueued_jobs { DeleteAccountHierarchyWorker.perform_now(provider) }
+    perform_enqueued_jobs(only: DeleteAccountHierarchyWorker) { DeleteAccountHierarchyWorker.perform_now(provider) }
 
     assert provider.reload
```
Martouta, line 54: This is cheating 😝 It should be:
```
provider.update_column(:state, 'approved')

perform_enqueued_jobs(only: [DeleteAccountHierarchyWorker, DeletePlainObjectWorker]) { DeleteAccountHierarchyWorker.perform_now(provider) }

assert provider.reload
```

test/workers/delete_account_hierarchy_worker_test.rb
```diff
@@ -43,19 +43,19 @@ def setup
...
 
-    perform_enqueued_jobs { DeleteAccountHierarchyWorker.perform_now(provider) }
+    perform_enqueued_jobs(only: DeleteAccountHierarchyWorker) { DeleteAccountHierarchyWorker.perform_now(provider) }
   end
 
   test 'does not perform if wrong state' do
     provider.update_column(:state, 'approved')
 
-    perform_enqueued_jobs { DeleteAccountHierarchyWorker.perform_now(provider) }
+    perform_enqueued_jobs(only: DeleteAccountHierarchyWorker) { DeleteAccountHierarchyWorker.perform_now(provider) }
```
Martouta, line 52: This is cheating 😝 It should be:
```
provider.update_column(:state, 'approved')

perform_enqueued_jobs(only: [DeleteAccountHierarchyWorker, DeletePlainObjectWorker]) { DeleteAccountHierarchyWorker.perform_now(provider) }

assert provider.reload
```

Martouta: Or even better if it is `perform_enqueued_jobs(except: SphinxIndexationWorker) { DeleteAccountHierarchyWorker.perform_now(provider) }` 🤔 It really has to be the same as the one in the test below 😄

test/workers/delete_account_hierarchy_worker_test.rb
```diff
@@ -43,19 +43,19 @@ def setup
     DeletePaymentSettingHierarchyWorker.expects(:perform_later).with(provider.payment_gateway_setting, anything, 'destroy')
     DeleteObjectHierarchyWorker.expects(:perform_later).with(api_docs_service, anything, 'destroy')
 
-    perform_enqueued_jobs { DeleteAccountHierarchyWorker.perform_now(provider) }
+    perform_enqueued_jobs(only: DeleteAccountHierarchyWorker) { DeleteAccountHierarchyWorker.perform_now(provider) }
```
Martouta, line 46: In this case, this is good 👍 😄

test/workers/delete_account_hierarchy_worker_test.rb
```diff
@@ -43,19 +43,19 @@ def setup
...
 
     assert provider.reload
   end
 
   test 'the account ends up destroyed after the hierarchy' do
-    perform_enqueued_jobs { DeleteAccountHierarchyWorker.perform_now(provider) }
+    perform_enqueued_jobs(except: SphinxIndexationWorker) { DeleteAccountHierarchyWorker.perform_now(provider) }
```
Martouta, line 58: This and the one above must run the same way, otherwise they are not testing the same thing.
https://github.com/3scale/porta/pull/1964/files#r440077985
IMO, this one is correct and the test above is wrong right now 😄

PR #1965, opened 2020-06-15
Adds HTTP timeout errors to fallback conditions of ImportCountriesService

guicassolato:
Right now `ImportCountriesService#call!` only rescues from `OpenURI::HTTPError` exceptions, which include 404 but not time out errors, for example. Without including time out errors, the step to import the list of countries using the Geonames API will fail to fallback to the static list of countries shipped with the code in some scenarios of disconnected deployment. This PR addresses this issue by including HTTP timeout errors to the fallback conditions of `ImportCountriesService`.

Closes [THREESCALE-4373](https://issues.redhat.com/browse/THREESCALE-4373)

### DIFF
--- a/test/unit/services/import_countries_service_test.rb
+++ b/test/unit/services/import_countries_service_test.rb
@@ -111,9 +111,19 @@ class ImportCountriesServiceTest < ActiveSupport::TestCase
     refute Country.find_by(code: 'XX')
   end
 
-  test 'falls back to local static file' do
+  test 'falls back to local static file on 404' do
     GeonamesLoader.unstub(:load_countries)
     stub_request(:get, 'http://download.geonames.org/export/dump/countryInfo.txt').to_return(status: 404) # Causes OpenURI::HTTPError to be raised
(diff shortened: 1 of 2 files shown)


test/unit/services/import_countries_service_test.rb
```diff
@@ -121,6 +121,16 @@ def geonames_response
     end
   end
 
+  test 'falls back to local static file on timeout' do
+    GeonamesLoader.unstub(:load_countries)
+    stub_request(:get, 'http://download.geonames.org/export/dump/countryInfo.txt').to_raise(Net::OpenTimeout.new)
```
Martouta, line 126: I'm not sure if having this link hardcoded everywhere is ideal 🤔 
https://github.com/3scale/porta/blob/7ca690acc92c3dea43779a0f600bd9b99613494a/app/services/import_countries_service.rb#L44

It would be better if it was customizable, but as I know that it is more work than desired, a simple improvement would be simply having it inside a constant and always referring to that same constant from all places instead of copy-pasting the hardcoded url everywhere.

Anyway, I know that this is out of the intended scope of the issue so I will approve 😄

PR #1966, opened 2020-06-15
[sphinx] Remove thread in openshift:thinking_sphinx:start

hallelujah:
It does not create a thread for indexing anymore

All indexing is done real time in background with Sidekiq

To initialize an indexation, run `rake ts:clear ts:rebuild`

### DIFF
--- a/app/indices/account_index.rb
+++ b/app/indices/account_index.rb
@@ -19,7 +19,7 @@ ThinkingSphinx::Index.define(:account, with: :real_time) do
   has :tenant_id, type: :integer
   has :state, type: :string
 
-  scope { Account.where(master: false).includes(:users, :bought_cinstances) }
+  scope { Account.not_master.includes(:users, :bought_cinstances) }
 end
 
 module AccountIndex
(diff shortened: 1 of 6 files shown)


app/indices/account_index.rb
```diff
@@ -19,12 +19,7 @@
   has :tenant_id, type: :integer
   has :state, type: :string
 
-  # if System::Database.oracle?
-  #   # Need to add the group by otherwise it will complain about ORA-00979 not a Group By function error
-  #   group_by "accounts.state"
-  # end
-
-  # where sanitize_sql(['COALESCE(accounts.master, ?) = ?', false, false])
+  scope { Account.where(master: false).includes(:users, :bought_cinstances) }
```
Martouta, line 22: Why not this?
```suggestion
  scope { Account.not_master.includes(:users, :bought_cinstances) }
```
And if you are sure that master can never be `nil` and it is always `false` instead, then we can also simplify:
https://github.com/3scale/porta/blob/23c541044bf72ea2e9632129877ef0d393e9e32c/app/models/account.rb#L91

PR #1967, opened 2020-06-17
Adds new column bought_cinstances_count to Accounts table

josemigallas:
First part of [THREESCALE-5328: Get the applications count for each account](https://issues.redhat.com/browse/THREESCALE-5328)

It's blocking #1939


db/migrate/20200617105046_add_new_column_to_accounts_bought_c_instances_count.rb
```diff
@@ -0,0 +1,7 @@
+class AddNewColumnToAccountsBoughtCInstancesCount < ActiveRecord::Migration[5.0]
+  def change
+    safety_assured do
```
Martouta, line 3: This shouldn't be necessary for this migration, is it? 👀 Because if it is, we have a problem 😝

PR #1971, opened 2020-06-18
Refactor (mini) 2 methods in ProviderMethods to make it easier/quicker to ever change them or debug them.

Martouta:
This couple of small changes makes it easier to change and easier to debug https://github.com/3scale/porta/pull/1939

### DIFF
--- a/app/models/account/provider_methods.rb
+++ b/app/models/account/provider_methods.rb
@@ -190,7 +190,7 @@ module Account::ProviderMethods
 
   def provider_key
     ensure_provider
-    self.bought_cinstances.first!.user_key
+    bought_cinstances.first!.user_key
   end
 
   def api_key?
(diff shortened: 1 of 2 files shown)


app/models/account/provider_methods.rb
```diff
@@ -338,13 +338,11 @@ def create_service(attrs)
   protected
 
   def create_first_service
-    return if @signup_mode
+    return if @signup_mode || has_bought_cinstance?
```
josemigallas, line 341:
```suggestion
    return if @signup_mode || self.bought_cinstances.exists?
```

hallelujah, line 341: That is not the same, in this case you will never run the ServiceCreator

app/models/account/provider_methods.rb
```diff
@@ -338,13 +338,11 @@ def create_service(attrs)
...
 
-    unless self.bought_cinstances.exists?
-      application_plans = Account.master.accessible_services.default.application_plans
-      plan = application_plans.default || application_plans.first!
-      plan.create_contract_with!(self)
-    end
+    application_plans = Account.master.accessible_services.default.application_plans
+    plan = application_plans.default || application_plans.first!
+    plan.create_contract_with!(self)
```
hallelujah, line 345: You can extract this in its new method

app/models/account/provider_methods.rb
Martouta: I know, that's why I left it as "WIP". sorry that I only spoke about it in private to another person 😞

app/models/account/provider_methods.rb
Martouta: better yes, done

PR #1972, opened 2020-06-18
Fix spec/acceptance/api/account_spec.rb 

Martouta:
This makes the test pass 😄

### DIFF
--- a/spec/support/api/context.rb
+++ b/spec/support/api/context.rb
@@ -26,7 +26,7 @@ end
 shared_context "provider api", provider: true do
   let(:master) { provider && master_account }
   let(:provider) { FactoryBot.create(:provider_account, self_domain: 'example.org') }
-  let(:provider_key) { provider.provider_key }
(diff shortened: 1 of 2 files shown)


spec/support/api/context.rb
```diff
@@ -26,7 +26,7 @@
 shared_context "provider api", provider: true do
   let(:master) { provider && master_account }
   let(:provider) { FactoryBot.create(:provider_account, self_domain: 'example.org') }
-  let(:provider_key) { provider.provider_key }
+  let(:provider_key) { provider.api_key? ? provider.provider_key : provider.reload.provider_key }
```
guicassolato, line 29: Can we always reload?

Martouta: I do not know but I didn't try out of efficiency... should we always reload? why?

guicassolato: I just thought about it for simplicity (of the code). Of course, it's another select to the db.

PR #1973, opened 2020-06-18
Injecting 3scale server to OAS3 swagger-ui

damianpm:

Dynamically injects the **server** to swagger-ui v3, using the `responseInterceptor` previously used to autocomplete params in https://github.com/3scale/porta/pull/1935.
OAS v3 use keyword `servers` replacing the `host`, `basePath` and `schemes` keywords of v2, see official [docs](https://swagger.io/docs/specification/api-host-and-base-path/): 

https://issues.redhat.com/browse/THREESCALE-4858
https://issues.redhat.com/browse/THREESCALE-4212 (parent task)

**Verification steps** 

- Go to a Service's active docs section: apiconfig/services/2/api_docs
- Create a new spec using OAS v3 (you can use [this example](https://gist.github.com/damianpm/aca68c763223ee93aed4b6e9c2562cfe))

If your OAS doesn't include any server, then a 3scale detected server will be injected.
If your OAS already includes one or more servers, they will be replaced by 3scale proxy endpoint

**Before:**
before

**After:**
after


app/javascript/packs/OAS3Autocomplete.js
```diff
@@ -53,25 +53,36 @@ const updateResponseBody = (response: SwaggerResponse, accountData: AccountData)
...
-)
+export const autocompleteOAS3 = async (response: SwaggerResponse, accountDataUrl: string, serviceEndpoint: string): SwaggerResponse => {
+  if (isSpecFetched(response)) {
+    return response
+  }
+
+  let body = injectServerToResponseBody(response.body, serviceEndpoint)
+
+  const accountData = await fetchData(accountDataUrl).then(data => data)
+  if (accountData.results) {
```
guicassolato, line 78: Perhaps

```suggestion
  if (accountData && accountData.results) {
```

damianpm:
@guicassolato @thomasmaas I've deployed to preview, if you want to test

app/javascript/packs/OAS3Autocomplete.js
```diff
@@ -53,25 +53,36 @@ const updateResponseBody = (response: SwaggerResponse, accountData: AccountData)
   }
 )
 
+const injectServerToResponseBody = (responseBody: ResponseBody, serviceEndpoint: string): ResponseBody => {
+  const originalServers = responseBody.servers ? responseBody.servers : []
```
josemigallas, line 57:
```suggestion
  const originalServers = responseBody.servers || []
```

app/views/admin/api_docs/base/swagger.html.erb
```diff
@@ -25,7 +26,7 @@
 
   <% content_for :javascripts do -%>
     <%= javascript_include_tag "/_cdn_assets_/swagger-ui/2.2.10/swagger-ui.js" -%>
-    <%= javascript_include_tag "swagger-ui2/api_docs_proxy" %>
+		<%= javascript_include_tag "swagger-ui2/api_docs_proxy" %>
```
josemigallas, line 29:
```suggestion
    <%= javascript_include_tag "swagger-ui2/api_docs_proxy" %>
```

app/javascript/packs/OAS3Autocomplete.js
```diff
@@ -53,25 +53,36 @@ const updateResponseBody = (response: SwaggerResponse, accountData: AccountData)
...
+  if (isSpecFetched(response)) {
+    return response
+  }
+
+  let body = injectServerToResponseBody(response.body, serviceEndpoint)
+
+  const accountData = await fetchData(accountDataUrl).then(data => data)
+  if (accountData.results) {
+    body = injectAutocompleteToResponseBody(body, accountData.results)
+  }
```
didierofrivia, line 80:
```suggestion
  // The params of `injectServerToResponseBody` and `injectAutocompleteToResponseBody` should be curried
  
  const injectServer = injectServerToResponseBody(serviceEndpoint)

  const body = await fetchData(accountDataUrl).then(data => (
   data.results 
     ? [injectServer, injectAutocompleteToResponseBody(data.results)].reduce((body, fn) => fn(body), response.body) 
     : injectServer(response.body)
  ))
```

app/javascript/packs/OAS3Autocomplete.js
```diff
@@ -53,25 +53,36 @@ const updateResponseBody = (response: SwaggerResponse, accountData: AccountData)
...
-      })
-)
+export const autocompleteOAS3 = async (response: SwaggerResponse, accountDataUrl: string, serviceEndpoint: string): SwaggerResponse => {
+  if (isSpecFetched(response)) {
+    return response
+  }
+
+  let body = injectServerToResponseBody(response.body, serviceEndpoint)
+
+  const accountData = await fetchData(accountDataUrl).then(data => data)
```
josemigallas, line 77:
```suggestion
  const accountData = await fetchData(accountDataUrl)
```

app/javascript/packs/OAS3Autocomplete.js
damianpm: There is something wrong with this, it breaks the flow
I think it's because `injectServerToResponseBody` is getting the wrong number of args, I will take a look later and we can make an inprovement later, by now we can keep what is working

app/views/admin/api_docs/base/swagger.html.erb
```diff
@@ -8,6 +8,7 @@
     class="swagger-ui-wrap"
     data-url="<%= spec_url(@api_docs_service) %>"
     data-base-url="<%= base_url %>"
+    data-service-endpoint="<%= @api_docs_service.service&.proxy&.endpoint %>"
```
Martouta, line 11: The [Law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter) says that one should not speak to strangers, and this is following a whole chain of strangers 😱

Martouta: You should at least delegate instead 😉

lib/developer_portal/lib/liquid/drops/api_spec.rb
```diff
@@ -21,6 +21,13 @@ def service
         return unless (service = @spec.service)
         Drops::Service.new(service)
       end
+
+      desc 'Returns the public url of the service of the spec if it has any or `nil` otherwise'
+      def endpoint
```
Martouta, line 26: The description is kinda clear but the name isn't. We have many "endpoint" and the name should let obvious which one it is.
For example `service_proxy_endpoint` (which would be easy with the delegation 😄 )

lib/developer_portal/app/views/developer_portal/docs.html.liquid
```diff
@@ -1,21 +1,20 @@
-<h1>Documentation</h1>
+{% content_for javascripts %}
+    {{ 'active_docs.js' | javascript_include_tag }}
+{% endcontent_for %}
 
-<p>Use our live documentation to learn about the Echo API</p>
+{% assign spec = provider.api_specs.first %}
```
Martouta, line 5: If the provider must have at least one api_spec to reach this view, then this should be:
```suggestion
{% assign spec = provider.api_specs.first! %}
```

app/views/admin/api_docs/base/swagger.html.erb
damianpm: @guicassolato suggested this approach
Any suggestion @Martouta ?

lib/developer_portal/lib/liquid/drops/api_spec.rb
guicassolato: That's actually _the_ endpoint of the API product. Perhaps it's the other "endpoints" that require better names 🙂 

@3scale/system-ux, how would you like to call it?

app/views/admin/api_docs/base/swagger.html.erb
guicassolato: Yeah. We'd better define a method in `ApiDocs::Service`.

lib/developer_portal/app/views/developer_portal/docs.html.liquid
guicassolato: They don't.

app/views/admin/api_docs/base/swagger.html.erb
Martouta: @damianpm yes, using delegate allowing nil. Perhaps also with the prefix. For example:
https://github.com/3scale/porta/blob/0a1016acf6876b578964251944d354e577617b9d/app/models/contract.rb#L50

lib/developer_portal/lib/liquid/drops/api_spec.rb
guicassolato: A few options:
- `production_endpoint`
- `production_service_endpoint`
- `production_api_endpoint`
- `production_api_product_endpoint`
- `api_product_endpoint`

damianpm: For me it's ok `service_proxy_endpoint`, but I don't know the names of other endpoints

damianpm: I haven't seen your list, then I like `api_product_endpoint`

guicassolato: We went with `api_product_production_public_base_url` after suggestion from @thomasmaas.

app/models/api_docs/service.rb
```diff
@@ -101,6 +101,11 @@ def specification
     @_spec ||= ThreeScale::Swagger::Specification.new(self.body)
   end
 
+  def api_product_production_public_base_url
+    return unless service&.proxy
+    service.proxy.endpoint
+  end
```
Martouta, line 107: I love the name ❤️ 
The content can be polished later knowing that this is for 2.9 😉

PR #1976, opened 2020-06-22
Remove "Admin" from getAccounts

josemigallas:
~**What this PR does / why we need it**:~

~Once `admin_user_display_name` is returned by porta, it has to be included in Accounts > Index table.~
~**Which issue(s) this PR fixes**~

### EDIT:
The admin's name will be included in the future due to implementation complexities. It's a TODO for now.

### DIFF
--- a/portafly/src/dal/accounts/getDeveloperAccounts.ts
+++ b/portafly/src/dal/accounts/getDeveloperAccounts.ts
@@ -53,8 +53,7 @@ const parseAccounts = (accounts: BuyersAccount[]) => accounts.map(({ account }) 
   createdAt: account.created_at,
   updatedAt: account.updated_at,
   orgName: account.org_name,
-  // TODO: Porta should return admin_name (username of first user role admin)
-  adminName: account.billing_address?.company,
+  adminName: 'TODO', // TODO: include admin_display_name in the response body
   state: account.state as State
 }))
 


spec/acceptance/api/account_spec.rb
```diff
@@ -14,6 +14,8 @@
     account
   end
 
+  # FIXME: admin_display_name should be always present, test should pass?
+  # let(:expected_provider_fields) { %w[admin_domain domain admin_base_url base_url from_email support_email finance_support_email site_access_code admin_display_name] }
```
Martouta, line 18: Problems with this:
1. This variable is used for testing both the JSON and the XML responses, but for the JSON it is using the AccountRepresenter which you edited, as for the XML, it is using the Account#to_xml method. To make it work for XML, you should add to that method something like `xml.admin_display_name display_name` or `xml.admin_display_name first_admin&.display_name` or something like that...
2. We have 2 different concepts for "display name" of an Account now, which is confusing like hell... you want the `first_admin.display_name`, which is:
https://github.com/3scale/porta/blob/acc77228aa4f1d1ac226c78cae0e421588de6990/app/models/user.rb#L317-L320
And, the other concept of Account#display_name is the one in the model which does not return the same thing (it uses the `username` of the user and never its name):
https://github.com/3scale/porta/blob/acc77228aa4f1d1ac226c78cae0e421588de6990/app/models/account.rb#L361-L364
3. What `expected_provider_fields` is used for below is to do this (simplified):

- For **JSON**: `should have_properties('admin_display_name').from(resource)`. In here:
https://github.com/3scale/porta/blob/acc77228aa4f1d1ac226c78cae0e421588de6990/spec/acceptance/api/account_spec.rb#L168

- For **XML**: `it { should have_tags('admin_display_name').from(resource) }` In here:
https://github.com/3scale/porta/blob/acc77228aa4f1d1ac226c78cae0e421588de6990/spec/acceptance/api/account_spec.rb#L218

What this does internally is [here](https://github.com/3scale/porta/blob/3122cd7781c421b330433a413b9ca54ea665ec2e/spec/support/api/matchers.rb#L1-L46), but as a simplified idea, this is what it does:
`assert resource.public_send(attribute) == json[attribute]` (`resource` being the model of account in this case). `resource.public_send(attribute)` cannot work for  `admin_display_name` because there isn't a method with that name inside this model

Martouta: I have a few possible solutions in mind but I do not like any 😓 : 
1. Remove the `delegate :display_name, to: :first_admin, allow_nil: true` line to use the `display_name` of `Account`, just as it is, without changing it. This means that you would always have the `username`, and never the `name` of the user.
2. Option 1 but changing the `Account` model `display_name` to return always the `display_name` of the user instead of the `username`.
3. Keep the `display_name` as it is in the models, remove the `delegate :display_name, to: :first_admin, allow_nil: true` from the representer, and add a new `admin_display_name` that returns actually what u want

hallelujah: We don't care about XML, we should deprecate it: Old API who have XML still works but newer APIs or response might not be compatible

Martouta: That's not the biggest issue at all IMO 😄 that it solved in the test by not adding the `expected_provider_fields` and instead just doing directly `it { should have_properties('admin_display_name').from(resource) }` or even directly an equality comparison there.
But I've mentioned much more above that should be addressed. And all my proposed solutions above have drawbacks.

hallelujah: `display_name` should never be on the model (nothing to do with database) but in the representer

And if there is conflict with `user.display_name` or anything else then we should address it already

Martouta:
Closing because this is not the approach that we are following. I'm actively working on it tho 👍

josemigallas:
Reopening since this is required. Let's change the approach if it's not adequate or close the JIRA first.

Martouta:
Ok, clarified: I am doing the backend side and Josemi will do the frontend side.

didierofrivia:
@Martouta what's the PR associated with this one? (The backend work)

Martouta:
@didierofrivia It is a lot of complicated work in the Backend side to do it well. The Jira issue of the job that you are requesting, in particular, is [THREESCALE-5815](https://issues.redhat.com/browse/THREESCALE-5815), but that is the last of the list of tasks of the father issue, which is [THREESCALE-5541](https://issues.redhat.com/browse/THREESCALE-5541).
The hardest part of what @macejmic and I have been fighting about is [THREESCALE-5814](https://issues.redhat.com/browse/THREESCALE-5814), for the billing address and for other issues.
I plan to record a video explaining why it is so hard to do it well and the possible options that I have in mind (one of the options is a 'make it ugly, make it work' and do the cleanup later, and another option is just not doing this at all if it is not needed and @thomasmaas has suspected). I initially wanted to write the whole thing but it is too long, so I will just explain it in a video.

didierofrivia:
@josemigallas Maybe we could just remove it for the time being, close this one (or change it removing the field)

josemigallas:
@didierofrivia OK I'll change the PR and leave the JIRA open for the future.

PR #1977, opened 2020-06-22
[PortaFly] Removing unused features from Account > Index

didierofrivia:

Removes features not needed in our first version of the account index.
https://github.com/3scale/porta/pull/1985 covers the missing tests.

Fixes: https://issues.redhat.com/browse/THREESCALE-5430

Screenshot 2020-06-22 at 11 51 52

### DIFF
--- a/portafly/src/components/pages/accounts/AccountsTable.tsx
+++ b/portafly/src/components/pages/accounts/AccountsTable.tsx
@@ -16,7 +16,6 @@ import {
   PaginationWidget,
   SendEmailModal,
   ChangeStateModal,
-  BulkActionsWidget,
   filterRows
 } from 'components'
 import { useTranslation } from 'i18n/useTranslation'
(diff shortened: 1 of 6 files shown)


thomasmaas:
Not sure if that’s true

didierofrivia:
I think we should not remove it. The actions column will apply to the given record and won't be triggering a batch update as how it was meant with bulk actions.

portafly/src/components/pages/accounts/AccountsTable.tsx
```diff
@@ -72,21 +57,11 @@ const AccountsTable: React.FunctionComponent = () => {
     setSortBy(index, direction, true)
```
josemigallas, line 57: Sorting needs the correction no more:
```suggestion
    setSortBy(index, direction)
```

damianpm: Well, technically, that line is not part of this PR

josemigallas: Precisely why I pointed it out, it must be changed because of https://github.com/3scale/porta/pull/1977/files#diff-89d3f403e9ab019e878ee0565aefb328L110.

PR #1988, opened 2020-06-23
Portafly - Applications index

josemigallas:

The index page for Applications. The navigation should be:
*  Product > Applications

[THREESCALE-5440: Create component for Application listing page](https://issues.redhat.com/browse/THREESCALE-5440)

Screen Shot 2020-07-06 at 13 55 01


portafly/src/components/pages/applications/ApplicationsTable.tsx
```diff
@@ -0,0 +1,114 @@
...
+  const { filters } = useDataListFilters()
+
+  // @ts-ignore
+  const planOptions = applications.reduce(
+    (plans, { plan }) => (plans.find((p) => p.id === plan.id) ? plans : [...plans, plan]),
+    [] as IPlan[]
+  ).map((plan) => ({
+    name: String(plan.id),
+    humanName: plan.name
+  }))
```
didierofrivia, line 49: Here we are relying on the plans gathered from the application list, which means there are potentially more plans available but we won't show them in the filter list. Which it might be fine, but @cntlsn and @thomasmaas should be aware of this behaviour

didierofrivia:
You need to rebase with master to remove latest type commits from https://github.com/3scale/porta/pull/2027

portafly/src/components/pages/applications/ApplicationsTable.tsx
josemigallas: I think the intended behavior is to take into account existing plans only better to have it confirmed.

cntlsn: I think what is important from a UX standpoint is to only show in the filter list the plans that are actually present in the table at that point in time. How does that sound?

thomasmaas: Fully agree with @cntlsn, this is about filtering applications by plan(s) so any plan that is not relevant for the collection of applications is irrelevant.

thomasmaas: I just love commenting stuff that has already been merged

PR #1993, opened 2020-06-25
Implement links in Accounts Table rows

josemigallas:

Some cells in the table not text but actual button Elements (links really). We should start using elements like this instead of strings (even if the target href leads to 404).

This collides with our current implementation for filtering and sorting doesn't work. Trying to filter throws an error and sorting doesn't do anything. The best example for this is the test: https://github.com/3scale/porta/pull/1993/files?file-filters%5B%5D=.tsx#diff-59165f5a444eb6a3db208329f63ecc87

Ideally we would simply get the `innerTex` of the element but what's stored in the State is an actual JS object, not a JSX element.

No JIRA

**Verification steps** 

Go to accounts and try filter and sorting by group/org.


portafly/src/components/pages/accounts/utils/accountsTableDataFactory.tsx
```diff
@@ -1,18 +1,21 @@
...
+import { ActionButtonImpersonate, AccountOverviewLink } from 'components/pages/accounts'
 
-const generateRows = (accounts: IDeveloperAccount[]) => {
+const generateRows: DataListRowGenerator = (accounts: IDeveloperAccount[]) => {
   const isMultitenant = process.env.REACT_APP_MULTITENANT
   // Rows and Columns must have the same order
   const mapAccountToRowCell = (account: IDeveloperAccount) => [
-    account.org_name,
+    {
+      filterableString: account.org_name,
```
didierofrivia, line 11: This is up to interpretation, but what I get from `filterableString` (the construction of the adj + string) is that one can (is able to) filter that particular string, that string is filterable. While something like `filterString` (noun + string) is that string _is_ the filter string, it's used for filtering purposes. 
Again, it's interpretation... Maybe someone else can give us some feedback as well.

portafly/src/components/shared/data-list/reducers/TableReducer.ts
```diff
@@ -20,8 +20,16 @@ export type TableState = {
...
   const next = b.cells[index]
-  if (prev < next) return -1
-  if (prev > next) return 1
+
+  const prevTarget = typeof prev === 'object'
+    ? prev.filterableString.replace(/<[^>]+>/g, '')
+    : prev
+  const nextTarget = typeof next === 'object'
+    ? next.filterableString.replace(/<[^>]+>/g, '')
+    : next
```
didierofrivia, line 29:
```suggestion
  const buildTarget = (target) => typeof target === 'object' ? target.filterableString.replace(/<[^>]+>/g, '') : target
  const prevTarget = buildTarget(prev) 
  const nextTarget = buildTarget(next)
```

portafly/src/components/pages/accounts/utils/accountsTableDataFactory.tsx
josemigallas: I agree with you, I don't like the name and have been looking for a better one. I thought about implementing `toString` instead:
```suggestion
      toString: () => account.org_name,
```
Other options `computedString`, `stringValue`, `value`....

PR #1997, opened 2020-06-26
Using swagger-ui instead of swagger-ui-react

damianpm:

ActiveDocs is using `swagger-ui-react`, which is having a bug not updating the state in inputs
Switching to `swagger-ui` package

https://issues.redhat.com/browse/THREESCALE-5463


app/javascript/packs/service_active_docs.js
```diff
@@ -9,5 +10,9 @@ document.addEventListener('DOMContentLoaded', () => {
 
   const responseInterceptor = (response) => autocompleteOAS3(response, accountDataUrl)
 
-  ActiveDocsSpec({ url, responseInterceptor }, containerId)
+  window.swaggerUi = new SwaggerUI({
```
josemigallas, line 13: I think you don't need to export it to window, in the liquid template it is called directly:
```suggestion
  SwaggerUI({
```

PR #2000, opened 2020-06-29
Refactor LazyRoute to support passing props better

josemigallas:

Some pages such as Account Overview need being passed props (la the account's own id). The current implementation of `LazyRoute` is too obscured and also its types not well defined.

Here's the aforementioned example implemented https://github.com/3scale/porta/pull/1998/files#diff-e4adad5bca8019703fcc2feecd1730c6R23-R30.

`useMemo` is also not really necessary and finally `React.Suspense` could be moved outside.

The component `LazyRoute` is therefore not really needed, but I preferred doing one step at a time. If the team agrees this PR can do the whole refactorization.

**Verification steps** 

`yarn start` and navigate around, everything should work.
Looking at browser's react tools, only **one** route should be rendered at any time.
Screen Shot 2020-06-29 at 17 14 53

### DIFF
--- a/portafly/src/components/LazyRoute.tsx
+++ b/portafly/src/components/LazyRoute.tsx
@@ -2,19 +2,10 @@ import * as React from 'react'
 import { RouteProps, Route } from 'react-router-dom'
 import { Loading } from 'components'
 
-export interface IDynamicImportProps extends RouteProps {
-  getComponent: () => Promise<{ default: React.ComponentType }>
-}
-
-export function LazyRoute({ path, exact, getComponent }: IDynamicImportProps) {
-  const LazyComponent = React.useMemo(() => React.lazy(getComponent), [
-    getComponent
-  ])
+export function LazyRoute({ path, exact, render }: RouteProps) {
   return (
-    <Route path={path} exact={exact}>
-      <React.Suspense fallback={<Loading />}>
-        <LazyComponent />
-      </React.Suspense>
-    </Route>
+    <React.Suspense fallback={<Loading />}>
+      <Route path={path} exact={exact} render={render} />
+    </React.Suspense>
   )
 }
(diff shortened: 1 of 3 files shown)


portafly/src/components/LazyRoute.tsx
```diff
@@ -2,19 +2,10 @@ import * as React from 'react'
 import { RouteProps, Route } from 'react-router-dom'
 import { Loading } from 'components'
 
-export interface IDynamicImportProps extends RouteProps {
-  getComponent: () => Promise<{ default: React.ComponentType }>
-}
-
-export function LazyRoute({ path, exact, getComponent }: IDynamicImportProps) {
-  const LazyComponent = React.useMemo(() => React.lazy(getComponent), [
```
didierofrivia, line 10: So we are losing memoization?

josemigallas: It didn't make sense in the first place AFAIK.

didierofrivia: In this case it seems it's good it seems. While visiting each page, you are lazy loading them and memoizing, second time you visit that page, and the props haven't changed, it's super fast.

josemigallas: The theory is valid but `useMemo` doesn't work like that. It is meant to save in _expensive_ computations (which we don't have) when **re-rendering** a component, but we aren't re-rendering either. The React.lazy trick will have in the DOM only one Route at a time, therefore when visiting a second time a page, it will render from scratch. Which again, it's super fast anyway.

You can check it yourself by adding a log:
```jsx
const LazyComponent = React.useMemo(() => {
    console.log('useMemo')
    return React.lazy(getComponent)
  }, [getComponent])
```

PR #2005, opened 2020-06-30
Update Portafly OAS Spec with /applications

josemigallas:

We need the endpoint for Applications index page.

[THREESCALE-5479: Update OAS spec with /applications endpoint](https://issues.redhat.com/browse/THREESCALE-5479)

### DIFF
--- a/portafly/doc/ux_specs/index.md
+++ b/portafly/doc/ux_specs/index.md
@@ -22,6 +22,7 @@ Please refer to the single page files together with mockups linked in each page 
 ##### Product
 * [Overview](./product/overview.md)
 * [Create product](./product/create_product.md)
+* [Edit product](./product/edit_product.md)
 * Applications
   * [Applications index](./product/applications/index.md)
 
--- a/portafly/doc/ux_specs/product/edit_product.md
+++ b/portafly/doc/ux_specs/product/edit_product.md
@@ -1,0 +1,25 @@
+# Edit Product
+
+[Back to index](../index.md)
+
+* [Mockups](https://marvelapp.com/prototype/ee82j74/screen/70939421)
+* [Mockups JIRA ticket](https://issues.redhat.com/browse/THREESCALE-4523)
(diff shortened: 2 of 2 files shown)


portafly/portafly_api.yml
```diff
@@ -36,95 +36,240 @@ paths:
...
+          example:
+            id: 1
+            created_at: 2019-10-18T05:13:26Z
+            updated_at: 2019-10-18T05:13:27Z
+            state: approved
+            org_name: Umbrella Corp.
+            admin_name: Oswell E. Spencer
+            apps_count: 3
   responses:
+    getApplications:
```
damianpm, line 166: It seems like this response is never used
And why not use `ApplicationsList` to keep it consistent, instead of `getApplications` that suggest an action

portafly/portafly_api.yml
```diff
@@ -36,95 +36,240 @@ paths:
...
         type: string
       in: query
+  /applications:
+    summary: A list of all applications
+    description: A list of all applications
+    get:
+      responses:
+        "200":
+          description: A list of applications
+          content:
```
damianpm, line 46: Maybe you wanted to use the `getApplications` here ?

portafly/portafly_api.yml
josemigallas: You're absolutely right

damianpm:
@josemigallas there is something wrong with the spec, try it in our ActiveDocs preview
no-operations

portafly/portafly_api.yml
```diff
@@ -36,95 +36,234 @@ paths:
       schema:
         type: string
       in: query
+  /applications:
+    summary: A list of all applications
+    description: A list of all applications
+    get:
+      responses:
+        "200":
```
damianpm, line 44: We are using single quotes in the other responses:
https://github.com/3scale/porta/blob/master/portafly/portafly_api.yml#L17

PR #2008, opened 2020-06-30
[sphinx] Fix cron hour for sphinx enqueue

hallelujah:
Put it at 13:00 UTC for safety, monitoring queue growth in SaaS

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -45,7 +45,7 @@ every :day, :at => '08:00', roles: [:cron] do
   ThreeScale::Jobs::BILLING.each { |task| instance_exec(task, &job_proc) }
 end
 
-every 1.day, :at => '6:00 am', roles: [:sphinx_server] do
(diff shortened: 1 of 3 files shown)


Martouta:
Yes @guicassolato because this is only for the ops folks to be able to monitor it in SaaS 😄

hallelujah:
I can put both at 13:00 but I aim at giving the customer more flexibility by changing those cron scheduler to real Openshift CronJob (it is beta so ...)


In fact today it ran at 6:00 and it was fine

hallelujah:
So ops want it at 1pm UTC every Tuesday

hallelujah:
@guicassolato you mean for doc people?

guicassolato:
> @guicassolato you mean for doc people?

Yes, @hallelujah. Although sphinx indexation is supposed to be transparent, some customers may be used to seeing a daily job. Now the new process will run once a week.

PR #2020, opened 2020-07-02
Upgrade @patternfly/react-core

josemigallas:

Upgrades `@patternly/react-core` to the latest version and fixes broken component.

### DIFF
--- a/portafly/src/components/AppLayout.tsx
+++ b/portafly/src/components/AppLayout.tsx
@@ -20,11 +20,7 @@ export const AppLayoutContext = React.createContext<IAppLayoutContext>({
   setBreadcrumb: () => 0
 })
 
-export interface IAppLayoutProps
-  extends Pick<PageHeaderProps, 'logo'>,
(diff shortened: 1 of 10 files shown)


portafly/src/components/AppLayout.tsx
```diff
@@ -20,11 +20,7 @@ export const AppLayoutContext = React.createContext<IAppLayoutContext>({
   setBreadcrumb: () => 0
 })
 
-export interface IAppLayoutProps
-  extends Pick<PageHeaderProps, 'logo'>,
-  Pick<PageHeaderProps, 'logoProps'>,
-  Pick<PageHeaderProps, 'avatar'>,
```
damianpm, line 26: What happens with the avatar? screenshot?

josemigallas: The Avatar goes now in `headerTools`. But we don't have an Avatar so nothing changes for us.

PR #2022, opened 2020-07-03
Bugfix: autocomplete params

damianpm:
Autocomplete of OAS 3 parameters fails, depending on the format of spec
`parameters` can be placed inside or outside the `get` object, but we are handling only the second scenario
```
{
  "openapi": "3.0.2",
  "info": {},
  "paths": {
      "/example/path": {
          "parameters": []
      }
  }
}
```

```
{
  "openapi": "3.0.2",
  "info": {},
  "paths": {
      "/example/path": {
          "get": {
              "parameters": []
          }
      }
  }
}
```

https://issues.redhat.com/browse/THREESCALE-5516

### DIFF
--- a/app/javascript/packs/OAS3Autocomplete.js
+++ b/app/javascript/packs/OAS3Autocomplete.js
@@ -32,12 +32,17 @@ const addAutocompleteToParam = (param: Param, accountData: AccountData): Param =
     : param
 }
 
+const getPathParameters = (path) => {
+  const hasParamsInGet = path.get && path.get.parameters
+  return hasParamsInGet ? path.get.parameters : path.parameters
+}
+
 const injectAutocompleteToResponseBody = (responseBody: ResponseBody, accountData: AccountData): ResponseBody => (
   {
     ...responseBody,
     paths: Object.keys(responseBody.paths).reduce(
       (paths, key) => {
-        const pathParameters = responseBody.paths[key].get.parameters
+        const pathParameters = getPathParameters(responseBody.paths[key])
         if (pathParameters) {
           paths[key] = {
             get: {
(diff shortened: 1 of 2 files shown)


app/javascript/packs/OAS3Autocomplete.js
```diff
@@ -32,12 +32,17 @@ const addAutocompleteToParam = (param: Param, accountData: AccountData): Param =
     : param
 }
 
+const getPathParameters = (path) => {
+  const hasParamsinGet = path.get && path.get.parameters
+  return hasParamsinGet ? path.get.parameters : path.parameters
```
josemigallas, line 37:
```suggestion
  const hasParamsInGet = path.get && path.get.parameters
  return hasParamsInGet ? path.get.parameters : path.parameters
```

app/javascript/packs/OAS3Autocomplete.js
```diff
@@ -32,12 +32,17 @@ const addAutocompleteToParam = (param: Param, accountData: AccountData): Param =
     : param
 }
 
+const getPathParameters = (path) => {
```
josemigallas, line 35: Are we sure `path` is always defined as well?

app/javascript/packs/OAS3Autocomplete.js
damianpm: Is not the same ?

app/javascript/packs/OAS3Autocomplete.js
damianpm: yes, is the name of the path

app/javascript/packs/OAS3Autocomplete.js
didierofrivia: `camelCased`

app/javascript/packs/OAS3Autocomplete.js
didierofrivia: checking `ResponseBody` type, it seems it is, but not its contents

PR #2024, opened 2020-07-03
[Refactor] Move display methods to the decorator for User and Account

Martouta:

### TODO in the following PR

The views should receive the user/account already decorated by the controller.
Too risky to do it in this same PR.
It will be in https://github.com/3scale/porta/pull/2043

### DIFF
--- a/app/decorators/account_decorator.rb
+++ b/app/decorators/account_decorator.rb
@@ -1,13 +1,11 @@
 # frozen_string_literal: true
 
 class AccountDecorator < ApplicationDecorator
-  def admin_user_display_name
-    admin_user.decorate.display_name
-  end
+  delegate :display_name, to: :admin_user, prefix: true
 
   private
 
   def admin_user
-    @admin_user ||= (super || User.new)
+    @admin_user ||= (super || User.new).decorate
   end
 end
--- a/app/decorators/account_decorator.rb
+++ b/app/decorators/account_decorator.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+class AccountDecorator < ApplicationDecorator
+  def admin_user_display_name
+    admin_user.decorate.display_name
+  end
+
+  private
+
+  def admin_user
+    @admin_user ||= (super || User.new)
+  end
+end
--- a/app/helpers/buyers/accounts_helper.rb
+++ b/app/helpers/buyers/accounts_helper.rb
@@ -6,7 +6,7 @@ module Buyers::AccountsHelper
   end
 
   def account_title account
-    [ h(account.org_name), h(account.admin_user.try!(:display_name)) ].compact.join(" &mdash; ").html_safe
+    [ h(account.org_name), h(account.decorate.admin_user_display_name) ].compact.join(" &mdash; ").html_safe
   end
 
   def link_to_buyer_or_deleted( buyer, path_method = :admin_buyers_account_path)
--- a/app/lib/csv/buyers_exporter.rb
+++ b/app/lib/csv/buyers_exporter.rb
@@ -16,6 +16,7 @@ class Csv::BuyersExporter < ::Csv::Exporter
   end
 
   def values_for_account(account)
+    user = (account.admin_user || User.new).decorate
     [
       account.id,
       account.state,
@@ -24,10 +25,10 @@ class Csv::BuyersExporter < ::Csv::Exporter
       account.bought_account_plan.try!(:name),
       account.created_at.to_s(:db),
       account.bought_cinstances.count,
-      account.first_admin.try!(:display_name),
-      account.first_admin.try!(:email),
+      user.display_name,
+      user.email,
       account.extra_fields.to_json,
-      account.first_admin.try!(:extra_fields).to_json
+      user.extra_fields.to_json
     ]
   end
 
--- a/app/lib/csv/buyers_exporter.rb
+++ b/app/lib/csv/buyers_exporter.rb
@@ -16,7 +16,7 @@ class Csv::BuyersExporter < ::Csv::Exporter
   end
 
   def values_for_account(account)
-    user = (account.admin_user || User.new).decorate
+    user = account.admin_user
     [
       account.id,
       account.state,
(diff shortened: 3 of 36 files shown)


app/lib/csv/buyers_exporter.rb
```diff
@@ -16,6 +16,7 @@ def generate
   end
 
   def values_for_account(account)
+    user = (account.admin_user || User.new).decorate
```
guicassolato, line 19: Hmm. If we ever implement some defaults to attributes of `User`, this can lead to some quite unexpected CSV outputs. I think better ensuring `account.admin_user` is present or the user columns are `nil`. WDYT?

app/mailers/notification_mailer.rb
```diff
@@ -72,12 +72,12 @@ def application_created(event, receiver)
...
     @account          = event.account
-    @receiver         = receiver
-    @user             = event.user
+    @receiver         = receiver.decorate
+    @user             = event.user.decorate
     @event            = event
 
     mail to: @receiver.email,
-         subject: "#{@user.informal_name} from #{@account.name} signed up"
+         subject: "#{@user.decorate.informal_name} from #{@account.name} signed up"
```
guicassolato, line 80: It should be decorated already, no?

```suggestion
         subject: "#{@user.informal_name} from #{@account.name} signed up"
```

app/lib/csv/buyers_exporter.rb
Martouta: Makes sense 👍

app/mailers/notification_mailer.rb
Martouta: Right, my bad 😄

app/lib/csv/buyers_exporter.rb
```diff
@@ -24,10 +25,10 @@ def values_for_account(account)
       account.bought_account_plan.try!(:name),
       account.created_at.to_s(:db),
       account.bought_cinstances.count,
-      account.first_admin.try!(:display_name),
-      account.first_admin.try!(:email),
+      account.decorate.admin_user_display_name,
+      user&.email,
       account.extra_fields.to_json,
-      account.first_admin.try!(:extra_fields).to_json
+      (user&.extra_fields || {}).to_json
```
guicassolato, line 31: I think this is not the same as before.

_Before:_ if there was no `account.first_admin`
```ruby
account.first_admin.try!(:extra_fields).to_json
=> "null"
```

_Now:_ if there's no `user`
```ruby
(user&.extra_fields || {}).to_json
=> "{}"
```

Martouta: Right 👍 And I will also add a test for that 😄

app/lib/csv/buyers_exporter.rb
```diff
@@ -24,10 +25,10 @@ def values_for_account(account)
       account.bought_account_plan.try!(:name),
       account.created_at.to_s(:db),
       account.bought_cinstances.count,
-      account.first_admin.try!(:display_name),
-      account.first_admin.try!(:email),
+      account.decorate.admin_user_display_name,
+      user&.email,
       account.extra_fields.to_json,
-      account.first_admin.try!(:extra_fields).to_json
+      user&.extra_fields.to_json
```
hallelujah, line 31: user should not be `nil`, you probably meant

```suggestion
      user.extra_fields.to_json
```

app/lib/csv/buyers_exporter.rb
```diff
@@ -24,10 +25,10 @@ def values_for_account(account)
       account.bought_account_plan.try!(:name),
       account.created_at.to_s(:db),
       account.bought_cinstances.count,
-      account.first_admin.try!(:display_name),
-      account.first_admin.try!(:email),
+      account.decorate.admin_user_display_name,
+      user&.email,
```
hallelujah, line 29: `user` is not nil, it is either a record or a User.new

app/decorators/account_decorator.rb
```diff
@@ -0,0 +1,13 @@
...
+
+class AccountDecorator < ApplicationDecorator
+  def admin_user_display_name
+    admin_user.decorate.display_name
+  end
+
+  private
+
+  def admin_user
+    @admin_user ||= (super || User.new)
```
hallelujah, line 11: Why not decorating directly the user?

app/helpers/buyers/accounts_helper.rb
```diff
@@ -6,7 +6,7 @@ def public_domain(account)
   end
 
   def account_title account
-    [ h(account.org_name), h(account.admin_user.try!(:display_name)) ].compact.join(" &mdash; ").html_safe
+    [ h(account.org_name), h(account.decorate.admin_user_display_name) ].compact.join(" &mdash; ").html_safe
```
hallelujah, line 9: I think we could use the decorated account in the controller instead of calling decorate in the helpers.

Martouta:
So how decorators should work:

instantiate them in the controller.
use them in the views / helpers
Calling #decorate in the helpers or views is not the way to go.



@hallelujah You are right but doing that in this same PR is even riskier, because it is too easy to skip a controller that is using it in the view... WDYT? worth the risk in this same PR or do it in the next one?

app/lib/csv/buyers_exporter.rb
Martouta: `admin_user` can return `nil`, this is why before this PR it was doing `try!(...`.

At first I did `user = (account.admin_user || User.new).decorate` but @guicassolato suggested that this could lead to wrong values:
https://github.com/3scale/porta/pull/2024#discussion_r451386790

I've recently added a test for this scenario in https://github.com/3scale/porta/pull/2040

app/decorators/account_decorator.rb
Martouta: Good point 👍

app/lib/csv/buyers_exporter.rb
Martouta: let's move this conversation to https://github.com/3scale/porta/pull/2024/files#r453490457

app/helpers/buyers/accounts_helper.rb
macejmic, line 9: what would happen what you would call decorate on a decorated object? nothing?

Martouta: It is not the case here, but I did it by accident in another place (fixed later), Gui spotted it, and it was working anyway XD: https://github.com/3scale/porta/pull/2024#discussion_r451387424

Look at this:
image


But anyway, it is not the case here, and I will move the `.decorate` to controllers instead of views/helpers in the following PR.

PR #2030, opened 2020-07-07
Add UX specs for Create product page

cntlsn:

Add UX specs for Create product page
Add link to Create product page UX specs file in table of content

https://issues.redhat.com/browse/THREESCALE-5548


thomasmaas:
I'm guessing the greyed out text in the description is meant as a placeholder? Not sure if that's needed.

cntlsn:
@josemigallas @thomasmaas thank you for your comments. I think placeholder text could be helpful to support users during the first use, but we would need to agree on a _theme_, for instance the PetStore example. In that case, the placeholder text could be:
**Name**
eg. PetStore API

**System name**
eg. pet-store-api

**Description**
eg. This API product includes the Pet API backend and the Food supply API backend

What do you think?

josemigallas:
> @josemigallas @thomasmaas thank you for your comments. I think placeholder text could be helpful to support users during the first use, but we would need to agree on a _theme_, for instance the PetStore example. In that case, the placeholder text could be:
> **Name**
> eg. PetStore API
> 
> **System name**
> eg. pet-store-api
> 
> **Description**
> eg. This API product includes the Pet API backend and the Food supply API backend
> 
> What do you think?

I think we might me talking about different things. I'm referring to the grayed text any input has to inform the user what kind of data goes there. For instance the input for Name would have a placeholder like _The name of your API_.

I'm not sure what you mean by _theme_ though.

thomasmaas:
@josemigallas palceholders are not meant to be labels. They are meant to contain an actual example of a valid input

deernano:
Updated to Marvel: https://marvelapp.com/prototype/ee82j74/screen/71008634

portafly/doc/ux_specs/product/create_product.md
```diff
@@ -0,0 +1,27 @@
+# Create Product
+
+[Back to index](../../index.md)
```
damianpm, line 3:
```suggestion
[Back to index](../index.md)
```

portafly/doc/ux_specs/product/create_product.md
```diff
@@ -0,0 +1,27 @@
...
+* [Mockups](https://marvelapp.com/prototype/ee82j74/screen/70939425)
+* [Mockups JIRA ticket](https://issues.redhat.com/browse/THREESCALE-5548)
+* [Parent JIRA ticket](https://issues.redhat.com/browse/THREESCALE-5547)
+
+##### Form
+* Input fields
+  * Name > text input [required]
+    * Validation errors:
+      * Required field on form submit (if left blank)
+      * Name already exists on _onblur_
```
damianpm, line 14: Maybe is better to validate if name already exists also on submit.
That way we do the validation (for both cases) only once.
WDYT @josemigallas @didierofrivia ?

cntlsn: @damianpm I was taking that for granted, but I will add it to the spec

cntlsn: added in https://github.com/3scale/porta/pull/2030/commits/57c24917c89b46f6f05e65b634503c44d2421125

PR #2036, opened 2020-07-08
[PortaFly] Storing API data in State

didierofrivia:

We need to provide other components with the retrieved API data.

https://issues.redhat.com/browse/THREESCALE-5564


portafly/src/components/shared/data-list/reducers/DataReducer.ts
```diff
@@ -0,0 +1,33 @@
...
+}
+
+// Reducer
+const dataReducer = createReducer(filtersActionHandlers)
+
+// Hook
+interface IUseData {
+  state: Record<'data', DataState>
+  dispatch: React.Dispatch<Action<DataState>>
+}
```
josemigallas, line 27: Instead of a union type for `DataSet`, I think we should use generics:
```suggestion
interface IUseData<T> {
  state: Record<'data', <T>>
  dispatch: React.Dispatch<Action<T>>
}
```

portafly/src/components/shared/data-list/DataListContext.tsx
```diff
@@ -14,17 +14,23 @@ import {
...
+  data: DataState
   table: TableState
   filters: FiltersState
   pagination: PaginationState
   bulkActions: BulkActionsState
 }
 
 const defaultState: State = {
+  data: [],
+  // TODO: We won't need to store this when PF updates the way they craft tables
```
josemigallas, line 33: I'm not so sure of this after all. The logic behind our table data is pretty heavy (rendering components, translations, predicates..) and it could have a great impact in performance, due to the high amount of re-renders a Table has.

portafly/src/components/shared/data-list/reducers/DataReducer.ts
```diff
@@ -0,0 +1,33 @@
...
+const useData = ({ state, dispatch }: IUseData) => ({
```
josemigallas, line 28:
```suggestion
const useData = <T>({ state, dispatch }: IUseData<T>) => ({
```

portafly/src/tests/components/shared/data-list/DataListContext.test.tsx
```diff
@@ -54,3 +56,59 @@ describe('DataListContext: useDataFilters', () => {
     expect(filters.innerHTML).toBe('')
   })
 })
+
+describe('DataListContext: DataReducer', () => {
```
josemigallas, line 60:
```suggestion
describe('DataListContext: useDataListData', () => {
```

portafly/src/tests/components/shared/data-list/DataListContext.test.tsx
```diff
@@ -54,3 +56,59 @@ describe('DataListContext: useDataFilters', () => {
...
+      .forEach((child, i) => expect(child.innerHTML).toBe(firstRenderAdminNames[i]))
+
+    fireEvent.change(input, { target: { value: 'Ringo' } })
+    fireEvent.click(addButton)
+    fireEvent.change(input, { target: { value: 'George' } })
+    fireEvent.click(addButton)
+
+    const addedAdminNames = ['John', 'Ringo', 'George']
+    Object.values(accounts.children)
+      .forEach((child, i) => expect(child.innerHTML).toBe(addedAdminNames[i]))
```
josemigallas, line 112:
```suggestion
      .forEach((child, i) => expect(getByText(child)).toBeInTheDocument())
```

portafly/src/tests/components/shared/data-list/DataListContext.test.tsx
```diff
@@ -54,3 +56,59 @@ describe('DataListContext: useDataFilters', () => {
...
+
+    return (
+      <>
+        <ul data-testid="accounts">
+          { data.map((account: IDeveloperAccount) => (
+            <li key={account.adminName}>{account.adminName}</li>
+          )) }
+        </ul>
+        <input data-testid="adminName" onChange={onChange} value={adminName} />
+        {/* eslint-disable-next-line react/button-has-type */}
```
josemigallas, line 82: Why not adding a type instead? 😂

portafly/src/tests/components/shared/data-list/DataListContext.test.tsx
```diff
@@ -54,3 +56,59 @@ describe('DataListContext: useDataFilters', () => {
     expect(filters.innerHTML).toBe('')
   })
 })
+
+describe('DataListContext: DataReducer', () => {
+  const ExampleComponent = () => {
+    const { data, setData } = useDataListData()
```
josemigallas, line 62:
```suggestion
    const { data, setData } = useDataListData<IDeveloperAccount[]>()
```

portafly/src/tests/components/shared/data-list/DataListContext.test.tsx
```diff
@@ -54,3 +56,59 @@ describe('DataListContext: useDataFilters', () => {
...
+    const [adminName, setAdminName] = useState('')
+    const onChange = (ev: React.ChangeEvent<HTMLInputElement>) => (
+      setAdminName(ev.currentTarget.value)
+    )
+
+    const addAccount = () => (
+      setData(
+        [...data, ...[factories.DeveloperAccount.build({ adminName })]] as IDeveloperAccount[]
```
josemigallas, line 70: I think this would not be needed if we use generics
```suggestion
        [...data, ...[factories.DeveloperAccount.build({ adminName })]]
```

portafly/src/components/shared/data-list/DataListContext.tsx
didierofrivia: I'll remove the comment. We'll need to give a second look to the table, and aim to remove those rows and columns (which store components) from the state.

portafly/src/tests/components/shared/data-list/DataListContext.test.tsx
didierofrivia: The thing is, that won't assert the specific value that comes from the data stored, which is the whole point of the test.

portafly/src/components/shared/data-list/reducers/DataReducer.ts
didierofrivia: Good thing of generics is less code and we don't need to define what to expect... but the bad thing is it could accept everything that is passed to the interface. If we think we shouldn't constrain what data is ok to be stored, then it's ok to go for generics.

portafly/src/tests/components/shared/data-list/DataListContext.test.tsx
didierofrivia: well, just because it's a test hehe

portafly/src/components/shared/data-list/reducers/DataReducer.ts
josemigallas: I think we don't have any constraints 🤔 What defines a DataList? A list of data, where data is a collection of anything, really.

portafly/src/tests/components/shared/data-list/DataListContext.test.tsx
josemigallas: My bad, what I really meant is:
```ts
addedAdminNames.forEach(name => expect(getText(name)).toBeInTheDocument())
```

portafly/src/components/shared/data-list/reducers/DataReducer.ts
didierofrivia: In our data list, it has some specifics. There's no usability without id and at least a matching row - column

portafly/src/tests/components/shared/data-list/DataListContext.test.tsx
didierofrivia: Doesn't apply it seems.

josemigallas: It's related to https://github.com/3scale/porta/pull/2036#discussion_r452087240

portafly/src/components/shared/data-list/DataListContext.tsx
```diff
@@ -69,11 +76,13 @@ const useDataListTable = () => useTable(useContext(DataListContext))
 const useDataListFilters = () => useFilters(useContext(DataListContext))
 const useDataListPagination = () => usePagination(useContext(DataListContext))
 const useDataListBulkActions = () => useBulkActions(useContext(DataListContext))
+const useDataListData = () => useData(useContext(DataListContext))
```
josemigallas, line 79:
```suggestion
const useDataListData = <T>() => useData<T>(useContext(DataListContext))
```

portafly/src/components/shared/data-list/reducers/DataReducer.ts
```diff
@@ -0,0 +1,31 @@
+import {
+  ActionHandlers,
+  Action,
+  createReducer
+} from 'utils'
+
+export type DataState = { id: number }[]
```
didierofrivia, line 7: We could add more fields when we understand better the needs for all the data sets from our API

portafly/src/components/shared/data-list/DataListContext.tsx
```diff
@@ -14,17 +14,23 @@ import {
...
+  data: DataState
   table: TableState
   filters: FiltersState
   pagination: PaginationState
   bulkActions: BulkActionsState
 }
 
 const defaultState: State = {
+  data: [],
+  // TODO: We should not store the table in our state, we should construct it out from data.
```
josemigallas, line 33: I'm still not convinced by this statement

portafly/src/components/shared/data-list/reducers/DataReducer.ts
josemigallas: Maybe it's better to use an `interface` for this 🤔 so that we can extend it.

PR #2037, opened 2020-07-08
"Create Product" page

josemigallas:

Adds a page component for Products > New.

Screen Shot 2020-08-12 at 13 14 15

Screen Shot 2020-07-29 at 12 50 04

Screen Shot 2020-07-29 at 12 49 10 1

[THREESCALE-5636: Component for "Create product" page](https://issues.redhat.com/browse/THREESCALE-5636)

**Verification steps** 

Go to `/products/new`
1. Check validation works as expected (onblur effects, etc)
2. Submit wrong data and check errors are shown (no product is created)
3. Submit correct data and check product is created in porta (also it should navigate to `products/:id`)

### DIFF
--- a/portafly/src/components/pages/product/CreateProductPage.tsx
+++ b/portafly/src/components/pages/product/CreateProductPage.tsx
@@ -1,0 +1,197 @@
+import React, {
+  useState,
+  FormEventHandler,
+  FocusEventHandler,
+  useEffect
+} from 'react'
+
+import {
+  PageSection,
+  TextContent,
+  Text,
+  Card,
+  CardBody,
+  Form,
+  FormGroup,
+  TextInput,
+  TextArea,
+  ActionGroup,
+  Button
+} from '@patternfly/react-core'
+import { useTranslation } from 'i18n/useTranslation'
+import { useHistory, Redirect } from 'react-router'
+import { createProduct, NewProduct } from 'dal/products'
+import { useAsync } from 'react-async'
+import { useAlertsContext, useDocumentTitle } from 'components/util'
+import { ValidationException, Validator } from 'utils'
+
+type Validations = Record<string, {
+  validation: 'default' | 'success' | 'error',
+  errors?: string[]
+}>
+
+const CreateProductPage = () => {
+  const { t } = useTranslation('product')
+  useDocumentTitle(t('create.pagetitle'))
+  const { goBack } = useHistory()
+  const { addAlert } = useAlertsContext()
+
+  const [name, setName] = useState('')
+  const [systemName, setSystemName] = useState('')
+  const [description, setDescription] = useState('')
+  const [validations, setValidations] = useState<Validations>({
+    name: { validation: 'default' },
+    system_name: { validation: 'default' },
+    description: { validation: 'default' }
+  })
+
+  const {
+    isPending, error, run, data
+  } = useAsync({ deferFn: createProduct })
+
+  useEffect(() => {
+    if (error) {
+      if (Object.prototype.hasOwnProperty.call(error, 'validationErrors')) {
+        const { validationErrors } = (error as unknown as ValidationException)
+        const newValidations: Validations = {}
+        Object.keys(validationErrors).forEach((id) => {
+          newValidations[id] = {
+            validation: 'error',
+            errors: validationErrors[id]
+          }
+        })
+        setValidations({ ...validations, ...newValidations })
+      } else {
+        addAlert({ id: String(Date.now()), title: error.message, variant: 'danger' })
+      }
+    }
+  }, [error])
+
+  if (data) {
+    const { service } = data as NewProduct
+    return <Redirect to={`/products/${service.id}`} />
+  }
+
+  const isValid = validations.name.validation === 'success' && validations.system_name.validation !== 'error'
+
+  const validator = Validator()
+    .for('name', () => (name.length > 0 ? 'success' : 'error'))
+    // eslint-disable-next-line no-nested-ternary
+    .for('system_name', () => (!systemName ? 'default' : (systemName.length > 0 ? 'success' : 'error')))
+    .for('description', () => (description.length > 0 ? 'success' : 'default'))
+
+  const onBlur: FocusEventHandler = (ev) => {
+    const { name: inputName } = ev.currentTarget as HTMLInputElement
+
+    const newValidations = { ...validations }
+
+    newValidations[inputName] = {
+      validation: validator.validate(inputName)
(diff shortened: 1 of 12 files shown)


portafly/src/components/pages/product/CreateProductPage.tsx
```diff
@@ -0,0 +1,182 @@
...
+          newValidations[id] = {
+            validation: 'error',
+            errors: validationErrors[id]
+          }
+        })
+        setValidations({ ...validations, ...newValidations })
+      } else {
+        addAlert({ id: String(Date.now()), title: error.message, variant: 'danger' })
+      }
+    }
```
didierofrivia, line 64: So `validationErrors` are not provided in the `data` object, but within `error` and not always. We might want to leverage this inside the http module. Then, would be easy to check that if `error`is true, data will be `validationErrors`. Since the parsing of this can't be done without modifying `useAsync`

portafly/src/components/pages/product/CreateProductPage.tsx
```diff
@@ -0,0 +1,182 @@
...
+        return name.length > 5 ? 'success' : 'error'
+      case 'system_name':
+        // eslint-disable-next-line no-nested-ternary
+        return !systemName ? 'default' : (systemName.length > 5 ? 'success' : 'error')
+      case 'description':
+        return description.length > 0 ? 'success' : 'default'
+      default:
+        return 'default'
+    }
+  }
```
didierofrivia, line 86: We could have a similar approach to our `ActionHandlers` and `reducers`, so we can define the case, here would be the input, and the fn to execute or the value to return.

portafly/src/utils/http.ts
```diff
@@ -1,35 +1,60 @@
...
+  if (response.status === 422) {
+    const responseBody = await response.json()
+    throw { validationErrors: responseBody.errors } as ValidationException
+  }
+
+  if (response.status === 201) {
+    const created = await response.json() as T
+
+    return created
+  }
```
didierofrivia, line 38: This too, a more declarative way would be good when we need to extend it, or pass it around. Check my comment https://github.com/3scale/porta/pull/2037/files#r455059864

portafly/src/utils/http.ts
```diff
@@ -1,35 +1,60 @@
...
+
+  throw new Error(response.statusText)
 }
 
 const craftRequest = (path: string, params?: URLSearchParams) => {
   const authToken = getToken()
 
-  const url = new URL(`${process.env.REACT_APP_API_HOST || '/'}${path}?${params?.toString()}`)
+  const url = new URL(`${process.env.REACT_APP_API_HOST || '/'}${path}?${String(params || '')}`)
```
didierofrivia, line 46: Is it not possible to add the params with the `URL` instance methods?

portafly/src/utils/http.ts
```diff
@@ -1,35 +1,60 @@
...
   url.searchParams.append('access_token', authToken as string)
 
   return new Request(
     url.toString(),
     {
-      headers: { Accept: 'application/json' }
+      headers: {
+        Accept: 'application/json',
+        'Content-Type': 'application/x-www-form-urlencoded'
+      }
```
didierofrivia, line 55: We could define the default ones in a constant.

portafly/src/components/pages/product/CreateProductPage.tsx
josemigallas: No, error can be anything, including a validation error we're gonna need to check for it either way. Consider as if `Object.prototype.hasOwnProperty.call(error, 'validationErrors')` were `typeof error === 'ValidationException'`...
There's no way to work around this. It's either check the error for `validationErrors` or the data, but I think it belongs to error.

portafly/src/utils/http.ts
josemigallas: This will be coped with in another PR

portafly/src/components/pages/product/CreateProductPage.tsx
```diff
@@ -0,0 +1,197 @@
...
+} from '@patternfly/react-core'
+import { useTranslation } from 'i18n/useTranslation'
+import { useHistory, Redirect } from 'react-router'
+import { createProduct, NewProduct } from 'dal/products'
+import { useAsync } from 'react-async'
+import { useAlertsContext, useDocumentTitle } from 'components/util'
+import { ValidationException, Validator } from 'utils'
+
+type Validations = Record<string, {
+  validation: 'default' | 'success' | 'error',
```
didierofrivia, line 29: maybe we can simplify the API with something like:

```suggestion
  validation?: boolean,
```

thus, if `undefined` means is not yet validated, and the boolean will define if it was success or error.

portafly/src/components/pages/product/CreateProductPage.tsx
```diff
@@ -0,0 +1,197 @@
...
+    return <Redirect to={`/products/${service.id}`} />
+  }
+
+  const isValid = validations.name.validation === 'success' && validations.system_name.validation !== 'error'
+
+  const validator = Validator()
+    .for('name', () => (name.length > 0 ? 'success' : 'error'))
+    // eslint-disable-next-line no-nested-ternary
+    .for('system_name', () => (!systemName ? 'default' : (systemName.length > 0 ? 'success' : 'error')))
+    .for('description', () => (description.length > 0 ? 'success' : 'default'))
```
didierofrivia, line 81: I like it more than the switch case version, good!

I was thinking that maybe we can pass the `Validator` factory a validation rules object like:

```suggestion
  const validator = Validator({
    name: () => (name.length > 0 ? 'success' : 'error'),
    'system_name': () => (!systemName ? 'default' : (systemName.length > 0 ? 'success' : 'error')),
    description: () => (description.length > 0 ? 'success' : 'default'))
  })
   ```

then in the implementation you could assert the case with `Object.prototype.hasOwnProperty` and execute the fn

portafly/src/components/pages/product/CreateProductPage.tsx
josemigallas: It's give by PF I'm afraid https://www.patternfly.org/v4/documentation/react/components/textinput#props

portafly/src/components/pages/product/CreateProductPage.tsx
josemigallas: But isn't that how it works already? Only with a sugar coating on top of it https://github.com/3scale/porta/pull/2037/files#diff-9d8e17e1d9797344cd149b6d58dc6377

PR #2039, opened 2020-07-08
[ci] Adding firefox non-headless

hallelujah:
This is because new Chrome has some incompatibilities with the current version of Selenium
https://groups.google.com/g/ruby-capybara/c/xR77bvMCGUo

If a button is inside a scrollable container and it is outside of the viewport it cannot be clicked
https://app.circleci.com/pipelines/github/3scale/porta/12745/workflows/4b5b052b-3ecc-4e7d-bccf-1e5bacb596a0/jobs/177980

Upgrading selenium and capybara might help, but it requires a fair amount of work

- Using a new system-builder so we need to update the cache


features/support/capybara.rb
```diff
@@ -7,7 +7,7 @@
 #  Capybara::Selenium::Driver.new(app, :browser => :firefox)
 #end
 
-DEFAULT_JS_DRIVER = :headless_chrome
+DEFAULT_JS_DRIVER = :headless_firefox
 # in case firefox is needed!
```
guicassolato, line 11: Perhaps we could invert this comment now to "in case chrome is needed"

Martouta: Or just remove the comment 😄

.circleci/config.yml
```diff
@@ -34,7 +34,7 @@ save-flow-typed-cache: &save-flow-typed-cache
     paths:
       - ./flow-typed
 
-npm-cache-key: &npm-cache-key node-v10-{{ checksum "yarn.lock" }}
+npm-cache-key: &npm-cache-key node-v10-{{ checksum "yarn.lock" }}-3
```
josemigallas, line 37: Why 3? Why should format this string properly, `v10` should say the version of the cache, not node.. So I would change it altogether to
```suggestion
npm-cache-key: &npm-cache-key node-cache-{{ checksum "yarn.lock" }}-v2
```

PR #2041, opened 2020-07-09
Update HTTP utils

josemigallas:

We need to update HTTP utils to support POST requests.

There's also a bug that prints `undefined` in the URL when `URLSearchParams` is undefined.


portafly/src/utils/index.ts
```diff
@@ -1,3 +1,4 @@
 export * from './auth'
 export * from './http'
 export * from './reducers'
+export * from './useCustomAsync'
```
josemigallas, line 4: @didierofrivia @damianpm 
This file is an idea of implementation, it will not be included in the commit as it. Ideas and suggestions welcome.

PR #2043, opened 2020-07-13
[WIP][DO NOT REVIEW] Decorate from Controllers instead of Views/Helpers

Martouta:


Martouta:
At first, I intended to push all the changes here, but it is too big and risky, so I will push it in smaller PRs instead.

PR #2044, opened 2020-07-13
(WIP) Update README with steps to set up PortaFly environment

cntlsn:

Add steps to set up the PortaFly environment

https://issues.redhat.com/browse/THREESCALE-5629

**Notes for the reviewers**
Please follow the steps in the README to set up the environment and run PortaFly on your machine.

### DIFF
--- a/portafly/README.md
+++ b/portafly/README.md
@@ -18,7 +18,7 @@ A React SPA crafted with the ambitious purpose of replacing Porta UI, currently 
 4. Allow CORS requests in your browser (Disclaimer: only needed to run the app in Development mode. Please note that allowing CORS requests makes your browser more vulnerable, do it at your own risk):
   * Install the CORS browser extension ([Google Chrome extension](https://chrome.google.com/webstore/detail/allow-cors-access-control/lhobafahddgcelffkeicbaginigeejlf), [Firefox add-on](https://addons.mozilla.org/en-US/firefox/addon/cors-everywhere/))
   * Click on the extension icon in the browser  toolbar and activate it by clicking on its logo
-5. Run PortaFly:
+5. Run PortaFly in development mode (please check the [following section](https://github.com/3scale/porta/blob/master/portafly/README.md#available-scripts) for all the available scripts):
   * Open a terminal session
   * Navigate to `/portafly`
   * Run `yarn install`
--- a/portafly/README.md
+++ b/portafly/README.md
@@ -1,6 +1,31 @@
 # PortaFly
 A React SPA crafted with the ambitious purpose of replacing Porta UI, currently residing inside the Rails App.
 
+## Getting up and running
+1. Clone the `porta` repository to your local machine
+2. Create a `.env` file:
+  * Create a new file in the `/portafly` root folder, and name it `.env`
+  * Paste `REACT_APP_API_HOST="https://multitenant-admin.preview01.3scale.net"` into it, and Save
+3. Create an Access token:
+  * Access your local Porta build (or access [Preview](https://multitenant-admin.preview01.3scale.net/))
+  * Navigate to Settings > Personal > Tokens
+  * Click the "Add Access Token" button
+  * Give your access token a Name
+  * Select all scopes
+  * Set permissions to Read & Write
+  * Create the access token
+  * Copy the access token
+4. Allow CORS requests in your browser:
+  * Install the CORS browser extension ([Google Chrome extension](https://chrome.google.com/webstore/detail/allow-cors-access-control/lhobafahddgcelffkeicbaginigeejlf))
+  * Click on the extension icon in the browser  toolbar and activate it by clicking on its logo
+5. Run PortaFly: 
+  * Open a terminal session
+  * Navigate to `/portafly`
+  * Run `yarn install`
+  * Run `yarn start`
+  * Open [http://localhost:3003](http://localhost:3003) (if it hasn't opened automagically)
+  * Enter your access token in the input field
+
 ## Available Scripts
 
 Inside the `portafly` directory, you can run:
--- a/portafly/README.md
+++ b/portafly/README.md
@@ -23,7 +23,7 @@ A React SPA crafted with the ambitious purpose of replacing Porta UI, currently 
   * Navigate to `/portafly`
   * Run `yarn install`
   * Run `yarn start`
-  * Open [http://localhost:3003](http://localhost:3003) (if it hasn't opened automagically)
+  * If the home page hasn't been displayed, open [http://localhost:3003](http://localhost:3003) in your browser.
   * Enter your access token in the input field
 
 ## Available Scripts
--- a/portafly/README.md
+++ b/portafly/README.md
@@ -15,10 +15,10 @@ A React SPA crafted with the ambitious purpose of replacing Porta UI, currently 
   * Set permissions to Read & Write
   * Create the access token
   * Copy the access token
-4. Allow CORS requests in your browser:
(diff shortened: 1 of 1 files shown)


portafly/README.md
```diff
@@ -1,6 +1,31 @@
 # PortaFly
 A React SPA crafted with the ambitious purpose of replacing Porta UI, currently residing inside the Rails App.
 
+## Getting up and running
+1. Clone the `porta` repository to your local machine
+2. Create a `.env` file:
```
damianpm, line 6: Creating an `.env` file is optional, you could create the file or simply add the Environment variables in the terminal, before `yarn start`, look [here](https://github.com/3scale/porta/blob/master/portafly/README.md#environment-variables)

portafly/README.md
```diff
@@ -1,6 +1,31 @@
...
+  * Create a new file in the `/portafly` root folder, and name it `.env`
+  * Paste `REACT_APP_API_HOST="https://multitenant-admin.preview01.3scale.net"` into it, and Save
```
damianpm, line 8: This is only needed of you are creating your access token in the preview environment, you could create it in your local environment

portafly/README.md
```diff
@@ -1,6 +1,31 @@
...
+3. Create an Access token:
```
damianpm, line 9: This step is right (by now), but it's not the definitive way to authenticate, just a temporary solution. IDK if we should add this section, as this is going to change in the future...

portafly/README.md
```diff
@@ -1,6 +1,31 @@
...
+  * Access your local Porta build (or access [Preview](https://multitenant-admin.preview01.3scale.net/))
+  * Navigate to Settings > Personal > Tokens
+  * Click the "Add Access Token" button
+  * Give your access token a Name
+  * Select all scopes
+  * Set permissions to Read & Write
+  * Create the access token
+  * Copy the access token
+4. Allow CORS requests in your browser:
```
damianpm, line 18: We may explain that allowing CORS is only needed in development mode, and add a disclaimer that if you decide to add an extension to bypass CORS, is at your own risk

portafly/README.md
```diff
@@ -1,6 +1,31 @@
...
+  * Install the CORS browser extension ([Google Chrome extension](https://chrome.google.com/webstore/detail/allow-cors-access-control/lhobafahddgcelffkeicbaginigeejlf))
+  * Click on the extension icon in the browser  toolbar and activate it by clicking on its logo
+5. Run PortaFly: 
+  * Open a terminal session
+  * Navigate to `/portafly`
+  * Run `yarn install`
```
damianpm, line 24: This is right, but is repeated [here](https://github.com/3scale/porta/blob/master/portafly/README.md#yarn-start)

portafly/README.md
Martouta, line 8: This is very unstable. We use Preview for testing and many times this API host won't be available. I think that it is important to point this out.

Martouta: Btw, it should be possible to use the local porta master/tenant account as API host as well. In fact, that will be more stable.

portafly/README.md
```diff
@@ -1,6 +1,31 @@
 # PortaFly
 A React SPA crafted with the ambitious purpose of replacing Porta UI, currently residing inside the Rails App.
 
+## Getting up and running
+1. Clone the `porta` repository to your local machine
+2. Create a `.env` file:
+  * Create a new file in the `/portafly` root folder, and name it `.env`
+  * Paste `REACT_APP_API_HOST="https://multitenant-admin.preview01.3scale.net"` into it, and Save
+3. Create an Access token:
+  * Access your local Porta build (or access [Preview](https://multitenant-admin.preview01.3scale.net/))
```
Martouta, line 10: This should be in the same host used in the previous step.

portafly/README.md
```diff
@@ -1,6 +1,31 @@
...
+  * Navigate to Settings > Personal > Tokens
+  * Click the "Add Access Token" button
+  * Give your access token a Name
+  * Select all scopes
+  * Set permissions to Read & Write
+  * Create the access token
+  * Copy the access token
+4. Allow CORS requests in your browser:
+  * Install the CORS browser extension ([Google Chrome extension](https://chrome.google.com/webstore/detail/allow-cors-access-control/lhobafahddgcelffkeicbaginigeejlf))
```
Martouta, line 19: It is ok to reference to the chrome extension, but imo, we should also reference to [the firefox one](https://addons.mozilla.org/en-US/firefox/addon/cors-everywhere/).
specially now that we are considering running our cukes with firefox: https://github.com/3scale/porta/pull/2039

portafly/README.md
didierofrivia: yeah, there are options like adding it to your .zshrc / .bashrc, explicitely declare them before executing the command, etc

portafly/README.md
didierofrivia: true, maybe a way to describe the necessity of having an instance running of Porta project. Also, our preview env shouldn't be public.

didierofrivia: we shouldn't make our preview env public

portafly/README.md
```diff
@@ -1,6 +1,31 @@
...
+  * Click on the extension icon in the browser  toolbar and activate it by clicking on its logo
+5. Run PortaFly: 
+  * Open a terminal session
+  * Navigate to `/portafly`
+  * Run `yarn install`
+  * Run `yarn start`
+  * Open [http://localhost:3003](http://localhost:3003) (if it hasn't opened automagically)
```
porueesq, line 26:
```suggestion
  * If the home page hasn't been displayed, open [http://localhost:3003](http://localhost:3003) in your browser.
```

portafly/README.md
cntlsn: what would be our suggested way?

portafly/README.md
cntlsn: so we suggest running Porta locally first in order to be able to generate the access token, correct?

portafly/README.md
cntlsn: you wouldn't be able to access portafly without an access token right now, so I think it is a necessary step, although temporary. we will remove it when things change.

portafly/README.md
cntlsn: I agree with adding the disclaimer.
What other modes besides development mode are available?

portafly/README.md
cntlsn: I know, but that doesn't look like a step by step guide on how to run portafly. It rather looks like a list of commands you can eventually run. We can link to that section for further details.

portafly/README.md
cntlsn: agreed, thank you for linking that!

portafly/README.md
Martouta: yeah I prefer we use it locally only

portafly/README.md
```diff
@@ -1,6 +1,31 @@
...
+  * Select all scopes
+  * Set permissions to Read & Write
+  * Create the access token
+  * Copy the access token
+4. Allow CORS requests in your browser:
+  * Install the CORS browser extension ([Google Chrome extension](https://chrome.google.com/webstore/detail/allow-cors-access-control/lhobafahddgcelffkeicbaginigeejlf))
+  * Click on the extension icon in the browser  toolbar and activate it by clicking on its logo
+5. Run PortaFly: 
+  * Open a terminal session
+  * Navigate to `/portafly`
```
josemigallas, line 23: Use commands directly, less explaining more doing IMO:
`$ cd portafly`

PR #2046, opened 2020-07-13
[PortaFly] Refactor: removing table from state

didierofrivia:

We are storing table components inside our data list state. This is an anti pattern.
Ideally we will be storing just data and states, not pieces or full components.
There are also some "extra-step" components that are not quite needed.

TBD

**Verification steps** 

TBD

TBD


hallelujah:

PR #2050, opened 2020-07-14
Fixed Branding issues in Dev Portal and SaaS Auth Provider Create and Update APIs

srijitamukherjee:
Problems: 

1. UI : When a user sets the credential for the GitHub SSO Integrations, the credential gets set in the system but the branding field does not get updated to 'custom_branding'. The branding information was not available in the interface either.

Before Adding Client-Id and Client-Secret:
Screenshot from 2020-07-22 14-00-52
Screenshot from 2020-07-22 14-01-01
After fixing visibility of the branding state:
Before
Before1
After adding **Client-Id** and **Client-Secret:**
github
After1
After

2. SaaS API: The _branding_state_event_ parameter is missing in the **Authentication Provider Developer Portal Update** API. And there is no Delete API available. Thus, user is stuck with the default and cannot update it.

**Look for the Branding Key in all the images below:**
The branding parameter was **branding_state_event**. Now it's, **branding_state**
**Authentication Provider Developer Portal Create**

Previously:
beforeCreateAPI
Now:
AfterCreateAPI

**Authentication Provider Developer Portal Update**
Previously:
BeforeUpdateAPI
Now:
After-API

3. Saas API: Even after setting the right value for the _branding_state_event_ (It can be either 'brand_as_threescale' (the default one) or 'custom_brand') parameter in the **Authentication Provider Developer Portal Create** API, it always is set as threescale_branded.

The Screenshots from Point-1 justifies the solution to this problem.

 
https://issues.redhat.com/browse/THREESCALE-4144

**Verification steps**

### DIFF
--- a/app/controllers/admin/api/authentication_providers_controller.rb
+++ b/app/controllers/admin/api/authentication_providers_controller.rb
@@ -155,14 +155,14 @@ class Admin::Api::AuthenticationProvidersController < Admin::Api::BaseController
       :name, :system_name, :client_id, :client_secret,
       :token_url, :user_info_url, :authorize_url, :site,
       :identifier_key, :username_key, :trust_email, :kind,
-      :published, :branding_state_event, :skip_ssl_certificate_verification,
+      :published, :branding_state, :skip_ssl_certificate_verification,
       :automatically_approve_accounts
     )
   end
 
   def authentication_provider_update_params
     params.require(:authentication_provider).permit(
-      :client_id, :client_secret, :published, :site,
+      :client_id, :client_secret, :published, :branding_state, :site,
       :skip_ssl_certificate_verification, :automatically_approve_accounts
     )
   end
--- a/app/controllers/provider/admin/authentication_providers_controller.rb
+++ b/app/controllers/provider/admin/authentication_providers_controller.rb
@@ -93,12 +93,13 @@ class Provider::Admin::AuthenticationProvidersController < FrontendController
   end
 
   UPDATE_PARAMS = %i[client_id client_secret automatically_approve_accounts site
-                     realm skip_ssl_certificate_verification branding_state_event
+                     realm skip_ssl_certificate_verification branding_state
                      token_url authorize_url user_info_url identifier_key
                      username_key trust_email].freeze
 
   def update_params
-    params.require(:authentication_provider).permit(UPDATE_PARAMS)
+    defaults = { branding_state: 'custom_branded' }
+    params.require(:authentication_provider).permit(UPDATE_PARAMS).reverse_merge(defaults)
   end
 
   def create_params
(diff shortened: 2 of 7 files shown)


thomasmaas:
Doesn't seem to solve the issue? In jira it says that if you enable custom branding, we still use 3scale github to authenticate while we should use the customer data.

srijitamukherjee:
> Doesn't seem to solve the issue? In jira it says that if you enable custom branding, we still use 3scale github to authenticate while we should use the customer data.

Possibly I am missing something here.
Here is what I did: 
1) Created an oauth app with my github. Added google.com as the callback and homepage URL.
2) Entered the client ID and secret in dev-portal
3) Selected the Custom branded in branding
4) Tested the auth flow which took me to google.

Screencast from 07-15-2020 01_45_09 PM

app/controllers/provider/admin/authentication_providers_controller.rb
```diff
@@ -93,12 +93,13 @@ def authorize_authentication_provider(action = action_name)
   end
 
   UPDATE_PARAMS = %i[client_id client_secret automatically_approve_accounts site
-                     realm skip_ssl_certificate_verification branding_state_event
+                     realm skip_ssl_certificate_verification branding_state
```
hallelujah, line 96: This controller is for the Admin SSO, the issue is about the developer portal SSO


BTW, You should not modify the parameter in the API, it will break many customer integrations

`branding_state_event` is an action which can take 2 values:

- [brand_as_threescale](https://github.com/3scale/porta/blob/6acea5d7722fb2c3319d12066200019b240b0892/app/models/authentication_provider/github.rb#L23)
- [custom_brand](https://github.com/3scale/porta/blob/6acea5d7722fb2c3319d12066200019b240b0892/app/models/authentication_provider/github.rb#L19)

app/controllers/admin/api/authentication_providers_controller.rb
```diff
@@ -155,14 +155,14 @@ def authentication_provider_create_params
...
-      :published, :branding_state_event, :skip_ssl_certificate_verification,
+      :published, :branding_state, :skip_ssl_certificate_verification,
       :automatically_approve_accounts
     )
   end
 
   def authentication_provider_update_params
     params.require(:authentication_provider).permit(
-      :client_id, :client_secret, :published, :site,
+      :client_id, :client_secret, :published, :branding_state, :site,
```
hallelujah, line 165: It should be

```suggestion
      :client_id, :client_secret, :published, :branding_state_event, :site,
```

Martouta:
@srijitamukherjee When u push to do changes from the review above, plz also rebase the branch because now it got too outdated.

hallelujah:

PR #2052, opened 2020-07-15
Updating jquery in Dev Portal

damianpm:

Updates jquery from 1.7 to 3.x in the developer portal, for security reasons.
It affects only new instances of the developer portal

https://issues.redhat.com/browse/THREESCALE-5381

**Verification steps** 

To create a new instance of the developer portal, you can reset the database
`bundle exec rake db:reset MASTER_PASSWORD=p USER_PASSWORD=p`


lib/developer_portal/app/views/developer_portal/javascripts/3scale.js
```diff
@@ -25,7 +25,6 @@ var hljs=new function(){function e(e){return e.replace(/&/gm,"&amp;").replace(/<
  *
  * Requires jQuery 1.6.0 or later.
  * https://github.com/rails/jquery-ujs
-
```
didierofrivia, line 28: Avoid these kind of changes

app/assets/javascripts/provider/cms_toolbar_v2.js
```diff
@@ -0,0 +1,51 @@
...
+    }
+  })
+  $('#hide-side-bar').on('click', event => {
+    event.preventDefault()
+    toggleValues()
+    if (storedToolbarState() === 'hidden') {
+      saveToolbarState('visible')
+    } else {
+      saveToolbarState('hidden')
+    }
```
josemigallas, line 49: Should we modernize the code as well?
```suggestion
    saveToolbarState(storedToolbarState() === 'hidden' ? 'visible' : 'hidden')
```

damianpm: But then why "modernizing" this line and not the others? The whole file can be "modernized", but I'd do it as a separate task.

josemigallas: This was only an example, but yes let's leave it for later!

PR #2053, opened 2020-07-15
API Account find without params should not try to find a 'nil' email

Martouta:
Fix 1st error reported by QE in [THREESCALE-5390](https://issues.redhat.com/browse/THREESCALE-5390)

Explained in [this comment in Jira](https://issues.redhat.com/browse/THREESCALE-5390?focusedCommentId=14242400&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-14242400)

I do not know how the email can be nil according to our validations but anyway it sounds like a good idea to me performance-wise to not do a search in the DB of empty user emails when the param has not been sent.

### DIFF
--- a/app/controllers/admin/api/accounts_controller.rb
+++ b/app/controllers/admin/api/accounts_controller.rb
@@ -266,7 +266,7 @@ class Admin::Api::AccountsController < Admin::Api::BaseController
       buyer_accounts.first_by_provider_key!(provider_key, error: ActiveRecord::RecordNotFound)
     when current_account.master? && service_token = params[:buyer_service_token]
       buyer_accounts.find_by_service_token!(service_token)
-    when email = params[:email]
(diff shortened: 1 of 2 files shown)


app/controllers/admin/api/accounts_controller.rb
```diff
@@ -266,8 +266,10 @@ def find_buyer_account
       buyer_accounts.first_by_provider_key!(provider_key, error: ActiveRecord::RecordNotFound)
     when current_account.master? && service_token = params[:buyer_service_token]
       buyer_accounts.find_by_service_token!(service_token)
+    when email = params[:email]
```
macejmic, line 269: maybe even `params[:email].presence` ?

Martouta: better, right 👍

PR #2054, opened 2020-07-15
Fixes Portafly build script

josemigallas:
`yarn build` is broken!

I took a look at the latest RCA and compared the package.json and configuration.

**TODO**

- [x] Add a step in our CircleCI pipeline to also run `build`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -828,7 +828,7 @@ jobs:
           path: tmp/capybara
           destination: capybara
 
-  portafly:
+  portafly-tests:
     resource_class: small
     docker:
       - image: circleci/node:12
@@ -841,8 +841,24 @@ jobs:
           command: |
             cd portafly
             yarn --version
-            yarn install
+            yarn install --frozen-lockfile
             yarn test --maxWorkers=3
+
+  portafly-build:
+    resource_class: small
+    docker:
+      - image: circleci/node:12
+    parameters:
+    steps:
+      - checkout-with-submodules
+      - run:
+          name: Run PortaFly Build
+          root: .
+          command: |
+            cd portafly
+            yarn --version
+            yarn install --frozen-lockfile
+            yarn build
 
   check_production_gemfile:
     parameters:
@@ -1139,16 +1155,19 @@ workflows:
             - assets_precompile
           <<: *only-master-filter
 
-  portafly_build:
+  portafly:
     jobs:
       - manual_approval:
           type: approval
       - notify_start:
           requires:
             - manual_approval
-      - portafly:
+      - portafly-tests:
           requires:
             - manual_approval
+      - portafly-build:
+          requires:
+            - portafly-tests
 
   production_gemfile-master:
     jobs:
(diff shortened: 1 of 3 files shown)


portafly/package.json
```diff
@@ -88,10 +88,10 @@
     "ts-pnp": "1.2.0",
     "typescript": "~3.8.3",
     "url-loader": "4.1.0",
-    "webpack": "4.43.0",
```
damianpm, line 91: Why it's needed to downgrade this packages?

josemigallas: The config file supports the old version, it's better to be aligned with CRA releases.

In other words, our config doesn't work with 4.43

.circleci/config.yml
```diff
@@ -1139,16 +1155,19 @@ workflows:
...
-  portafly_build:
+  portafly:
     jobs:
       - manual_approval:
           type: approval
       - notify_start:
           requires:
             - manual_approval
-      - portafly:
+      - portafly-tests:
```
didierofrivia, line 1165: You should follow convention and align with the rest of the tasks and use underscores, snake_cased names.

.circleci/config.yml
```diff
@@ -858,7 +858,7 @@ jobs:
             cd portafly
             yarn --version
             yarn install --frozen-lockfile
-            yarn build
+            CI=false yarn build
```
didierofrivia, line 861: Isn't there any other way to disable that behaviour? I'm not sure about this, specially when we don't know what else uses the `CI` env...

josemigallas: Nothing else should be affected since the env var is changed for this line only, isn't it?

.circleci/config.yml
josemigallas: I don't know if there's such convention whatsoever 😂 I've followed the most common format in the config file: 51 kebab vs 21 snake

didierofrivia: I agree that most artifacts are kebab, but the name of the workflows/jobs are snake: `mysql_build`, `postgres_build`, `oracle_build`, `assets_precompile`, `dependencies_npm`etc...

.circleci/config.yml
didierofrivia: In theory a build shouldn't depend on `CI`, that's true, but we are triggering from circle and since the problem is that warnings are been treated as errors, I'd expect a different flag/config to adjust the level, not completely disabled it

josemigallas: I can't disagree but it's way it is 🤷‍♂️ 
More info: https://github.com/facebook/create-react-app/issues/3657

I don't like having warnings anyway, maybe we should linter-treat them as errors and use `eslint-ignore` more often 🤷‍♂️

didierofrivia: We should, yeah. Will leave the circle official example here for future reference:
https://circleci.com/docs/2.0/sample-config/

josemigallas: OK PR coming soon!

PR #2057, opened 2020-07-16
(WIP) [UX specs] Network (and success) messages

cntlsn:

Add UX specs for Network and Success messages in the form of toast and inline alerts

### DIFF
--- a/portafly/doc/ux_specs/global_components/alerts.md
+++ b/portafly/doc/ux_specs/global_components/alerts.md
@@ -1,0 +1,19 @@
+# Alerts UX/UI specs
+
+[Back to index](../index.md)
+
+* [PatternFly Alerts design guidelines](https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/alerts-and-notifications)
+* [PatternFly-react Alerts docs](https://www.patternfly.org/v4/documentation/react/components/alert)
+
+### Network error alerts
+* Network errors are displayed in **toast alerts**, unless the current state of the UI is blocking users' view of the page layout (eg. modal dialogs)
+  * In this case, the errors will be displayed as **inline alerts**, making sure the positioning is as close as possible to the trigger (eg. button) of the action that launched the network request.
+* Network error alerts are persistent and always displays a dismiss button (`fa-times` icon)
+* Network error alerts variant is `warning`
+* Network error types:
+  * Generic network error
+  * Authentication error
+    * Wrong token
+    * Session expired
+  * Server error (there’s a bug, server is down..)
+  * Server too slow (timeout)
--- a/portafly/doc/ux_specs/global_components/alerts.md
+++ b/portafly/doc/ux_specs/global_components/alerts.md
@@ -8,7 +8,7 @@
 ### Network error alerts
 * Network errors are displayed in **toast alerts**, unless the current state of the UI is blocking users' view of the page layout (eg. modal dialogs)
   * In this case, the errors will be displayed as **inline alerts**, making sure the positioning is as close as possible to the trigger (eg. button) of the action that launched the network request.
-* Network error alerts are persistent and always displays a dismiss button (`fa-times` icon)
+* Network error alerts are persistent and always display a dismiss button (`fa-times` icon)
 * Network error alerts variant is `warning`
 * Network error types:
   * Generic network error
@@ -20,5 +20,5 @@
 
 ### Success alerts (bulk actions, delete object...)
 * Success messages are displayed in **toast alerts**
-* Success messages are not persistent, will be dismissed after 8 seconds and always displays a dismiss button (`fa-times` icon)
+* Success messages are not persistent, will be dismissed after 8 seconds and always display a dismiss button (`fa-times` icon)
 * Success alerts variant is `success`
--- a/portafly/doc/ux_specs/global_components/alerts.md
+++ b/portafly/doc/ux_specs/global_components/alerts.md
@@ -12,11 +12,16 @@
 * Network error alerts variant is `warning`
 * Network error types:
   * Generic network error
+  * Secure connection failed
+  * Timed out
+  * Cannot find host
+  * Not connected to Internet
+  * Cannot connect to host
+  * Not found
   * Authentication error
     * Wrong token
     * Session expired
-  * Server error (there’s a bug, server is down..)
-  * Server too slow (timeout)
+  * Server internal error
 
 ### Success alerts (bulk actions, delete object...)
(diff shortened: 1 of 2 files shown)


portafly/doc/ux_specs/global_components/alerts.md
```diff
@@ -0,0 +1,24 @@
...
+
+* [PatternFly Alerts design guidelines](https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/alerts-and-notifications)
+* [PatternFly-react Alerts docs](https://www.patternfly.org/v4/documentation/react/components/alert)
+
+### Network error alerts
+* Network errors are displayed in **toast alerts** ([PF specs](https://www.patternfly.org/v4/documentation/react/components/alert#types)), unless the current state of the UI is blocking users' view of the page layout (eg. modal dialogs)
+  * In this case, the errors will be displayed as **inline alerts** ([PF specs](https://www.patternfly.org/v4/documentation/react/components/alert#inline-types)), making sure the positioning is as close as possible to the trigger (eg. button) of the action that launched the network request.
+* Network error alerts are persistent and always display a dismiss button (`fa-times` icon)
+* Network error alerts variant is `warning`
+* Network error types:
```
cntlsn, line 13: @josemigallas @damianpm could you please elaborate on that? Or shall we better spec out the type of error in the page or component file where the alert is displayed?

portafly/doc/ux_specs/global_components/alerts.md
```diff
@@ -0,0 +1,24 @@
+# Alerts UX/UI specs
+
+[Back to index](../index.md)
+
+* [PatternFly Alerts design guidelines](https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/alerts-and-notifications)
+* [PatternFly-react Alerts docs](https://www.patternfly.org/v4/documentation/react/components/alert)
+
+### Network error alerts
+* Network errors are displayed in **toast alerts** ([PF specs](https://www.patternfly.org/v4/documentation/react/components/alert#types)), unless the current state of the UI is blocking users' view of the page layout (eg. modal dialogs)
+  * In this case, the errors will be displayed as **inline alerts** ([PF specs](https://www.patternfly.org/v4/documentation/react/components/alert#inline-types)), making sure the positioning is as close as possible to the trigger (eg. button) of the action that launched the network request.
```
cntlsn, line 10: @josemigallas is this even a possible scenario? What is it exactly that is dismissing the modal, clicking on the button that submits the form or that the request returns successfully? In the latter scenario, what if the request takes some time, will the user be stuck in the modal?

portafly/doc/ux_specs/index.md
```diff
@@ -8,6 +8,7 @@ Please refer to the single page files together with mockups linked in each page
 ### Global components
 * [Toolbar](./global_components/toolbar.md) -- the toolbar is a collection of UI components that allow a user to manage and manipulate a data set. A toolbar will almost always be present in pages that include a table, data list etc.
 * [Modal](./global_components/modal.md) -- the modal is an overlay dialog triggered to allow users taking contextual bulk actions or confirming destructive actions.
+* [Alerts](./global_components/alerts.md) -- Alerts are used to notify the user about a change in system status or other event that needs attention.
```
porueesq, line 11: Are alerts a mean to notify only, or also to prompt users to perform an action? 
Examples:
* Notify that the number of hits to an endpoint has reached its limit (no action required)
* To create an application, an application plan is required. Inform the user about this, and prompt the creation of an application plan (action required)

Perhaps we should distinguish between:
* Notifications: informative messages
* Alerts: informative messages requiring action from the user.

portafly/doc/ux_specs/global_components/alerts.md
josemigallas: It's a perfectly valid scenario IMO. The model would be dismissed after the action is successful (server responded with an OK).

The user should be stuck in the modal if there's an error to avoid losing information (forms) and having to open the same modal again (which I as an user find annoying).

portafly/doc/ux_specs/global_components/alerts.md
josemigallas: A network error related to the connection, not the server. They can be:

- Secure connection failed
- Timed out
- Cannot find host
- Not connected to Internet
- Cannot connect to host

On the other hand, server errors include:

- Not found
- Authentication error
- Server internal error
etc

portafly/doc/ux_specs/index.md
josemigallas: I think `Alert` refers to a [Patternfly component](https://www.patternfly.org/v4/documentation/react/components/alert) and yes it is possible to include a prompt for the user. They call it `AlertAction`.

I personally would keep it simple and use _Alert_. For informative messages requiring action I would use _Alert with an action_.

portafly/doc/ux_specs/global_components/alerts.md
cntlsn: @josemigallas cool, thank you! I will update the spec.
about server errors, aren't those usually displayed in full pages (thinking of 404 pages for instance). Could they be triggered by bulk actions as well?

portafly/doc/ux_specs/global_components/alerts.md
cntlsn: @josemigallas it's tricker than that. locking the user into the modal is bad UX. for instance, what if for some reason the server takes minutes to respond? the user will be stuck in the modal view, whereas they could instead submit a bulk action on a large number of objects and then navigate away to their next task while the bulk action is processed in the background.

at the same time, yes, I agree that we don't want them to eventually lose the content they entered in the modal before we're sure the server responded successfully.

what do you think about the following proposals?
- The modal is dismissed as soon as the request to the server is **sent**. 
- A toast alert will be displayed showing the status of the request.
- User-generated content is temporarily saved in the fields and will be deleted **only** when the response from the server is OK.
- In case an error is triggered, the toast alert will update to display a button that says something like "Try again". Clicking on the button will reopen the modal in the last state before it was dismissed (pre-selected items, user-generated content) – If user dismisses the toast alert the user-generated content will be deleted as well.

portafly/doc/ux_specs/index.md
cntlsn: @porueesq thank you for your comment. I agree with @josemigallas, and for the record, I have used `Alert` because that is how the component is called in PF, where there is no differentiation between an 'alert' and a 'notification'

portafly/doc/ux_specs/global_components/alerts.md
damianpm: Yes, server errors are displayed in it's own page, for example a 404 needs its own full page because there is no other page where you can insert an alert.

PR #2058, opened 2020-07-16
[portafly] Dockerfile and openshift definitions

guicassolato:
- Adds a Dockerfile to build Portafly independently from Porta
- Adds examples of Openshift ImageStream, DeploymentConfig, Service and (provider) Route to deploy Portafly

Depends on https://github.com/3scale/porta/tree/portafly/fix_http_url

----

Deployed to Openshift:
Screenshot 2020-07-16 at 17 57 42

### DIFF
--- a/portafly/Dockerfile
+++ b/portafly/Dockerfile
@@ -1,0 +1,13 @@
+FROM node:10-alpine as builder
+WORKDIR /tmp/portafly
+COPY . ./
+ENV PUBLIC_URL="/portafly"
+RUN yarn install \
+    && yarn build --production --frozen-lockfile
+
+FROM node:10-alpine
+RUN npm install -g local-web-server
+WORKDIR /opt/portafly
+COPY --from=builder /tmp/portafly/build ./
+EXPOSE 5000
+CMD ws -p 5000 --rewrite '/portafly/(.*)->$1' -d /opt/portafly --spa index.html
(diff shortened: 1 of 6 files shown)


portafly/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM node:10-alpine as builder
+WORKDIR /tmp/portafly
+COPY . ./
+ENV PUBLIC_URL="/portafly"
+RUN yarn install \
```
josemigallas, line 5: This will install only production dependencies, according to our yarn.lock file
```suggestion
RUN yarn install --production --frozen-lockfile \
```

portafly/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM node:10-alpine as builder
```
hallelujah, line 1: It would be nice to use CentOS to be the closest possible to productized build
But it is OK for now

portafly/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+WORKDIR /tmp/portafly
+COPY . ./
+ENV PUBLIC_URL="/portafly"
+RUN yarn install \
+    && yarn build --production --frozen-lockfile
+
+FROM node:10-alpine
```
hallelujah, line 8: Oh it is a multi stage build.
Be aware It is not supported by Red Hat productization build

PR #2060, opened 2020-07-17
Update width in tables

cntlsn:

* Add Width modifiers to define column distribution rules for Accounts index page
* Add Width modifiers to define column distribution rules for Applications index page
* Add column distribution rules in UX specs files for pages that include a table


cntlsn:
> Maybe clean up the commits? There are 2 commits with the same name, might be just 1?

Same commit name, but two diff files!

PR #2064, opened 2020-07-20
[PoliciesConfig] validate json

macejmic:

### DIFF
--- a/app/models/proxy.rb
+++ b/app/models/proxy.rb
@@ -136,6 +136,7 @@ class Proxy < ApplicationRecord
 
   def policies_config
     parsed_config = read_and_parse_policies_config
+    return parsed_config if errors[:policies_config].present?
 
     if parsed_config.detect { |c| c['name'] == DEFAULT_POLICY['name'] }
       parsed_config
@@ -560,11 +561,12 @@ class Proxy < ApplicationRecord
     include ActiveModel::Validations
 
     attr_accessor :name, :version, :configuration, :enabled
+
     validates :name, :version, presence: true
     validate :configuration_is_object
 
     def initialize(attributes = {})
-      self.attributes = attributes
+      self.attributes = attributes.is_a?(Hash) ? attributes : {}
     end
 
     def attributes=(attributes)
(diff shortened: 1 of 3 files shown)


test/integration/user-management-api/services/proxy/policies_test.rb
```diff
@@ -58,6 +58,12 @@ def test_invalid_policies_config
     assert_response :unprocessable_entity
   end
 
+  def test_invalid_json_policies_config
+    put admin_api_service_proxy_policies_path(valid_params.merge({ proxy: { policies_config: { name: 'echo '}.to_json }}))
+    binding.pry
```
Martouta, line 63: 👀

test/integration/user-management-api/services/proxy/policies_test.rb
```diff
@@ -58,6 +58,11 @@ def test_invalid_policies_config
     assert_response :unprocessable_entity
   end
 
+  def test_invalid_json_policies_config
+    put admin_api_service_proxy_policies_path(valid_params.merge({ proxy: { policies_config: { name: 'echo '}.to_json }}))
+    assert_response :unprocessable_entity
+  end
```
Martouta, line 64: This should also test that the response.body contains the proper error message in policies_config

macejmic: done.

app/models/proxy.rb
```diff
@@ -559,12 +560,16 @@ def proxy
   class PolicyConfig
     include ActiveModel::Validations
 
+    attr_reader :init_attributes
     attr_accessor :name, :version, :configuration, :enabled
-    validates :name, :version, presence: true
-    validate :configuration_is_object
+
+    validates :name, :version, presence: true, if: :attributes_are_hash?
```
hallelujah, line 566: Why not raise an error if passed a wrong argument?

app/models/proxy.rb
```diff
@@ -559,12 +560,16 @@ def proxy
...
+    validate :configuration_is_object, if: :attributes_are_hash?
+    validate :attributes_are_hash
 
     def initialize(attributes = {})
-      self.attributes = attributes
+      @init_attributes = attributes
```
hallelujah, line 571: `#initialize` should check the argument passed instead of validating them, wdyt?

app/models/proxy.rb
macejmic: QE is complaining about 500.. so that wouldn't solve the problem.

app/models/proxy.rb
macejmic: okok, that could be also the way.

PR #2066, opened 2020-07-20
🐛 Fix HTTP utils

josemigallas:
`craftRequest` should work regardless of API_HOST being defined. For local for instance we need it but for a production environment such as Openshift we don't.

😓 
Screen Shot 2020-07-21 at 09 25 17

### DIFF
--- a/portafly/src/utils/http.ts
+++ b/portafly/src/utils/http.ts
@@ -18,14 +18,19 @@ const fetchData = async <T>(request: Request): Promise<T> => {
   return response.parsedBody as T
 }
 
+const getStringUrl = (path: string, params: URLSearchParams) => {
+  const host = process.env.REACT_APP_API_HOST
+  // REACT_APP_API_HOST will be undefined in Openshift and URL is not compatible with relative paths
+  const url = host ? new URL(path, host) : path
+  return `${url.toString()}?${params.toString()}`
+}
+
 const craftRequest = (path: string, params = new URLSearchParams()) => {
   const authToken = getToken()
   params.append('access_token', authToken as string)
 
-  const url = `${process.env.REACT_APP_API_HOST || '/'}${path}?${params.toString()}`.replace('//', '/')
-
   return new Request(
-    url,
+    getStringUrl(path, params),
     {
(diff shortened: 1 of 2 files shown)


portafly/src/utils/http.ts
```diff
@@ -22,10 +22,12 @@ const craftRequest = (path: string, params = new URLSearchParams()) => {
   const authToken = getToken()
   params.append('access_token', authToken as string)
 
-  const url = `${process.env.REACT_APP_API_HOST || '/'}${path}?${params.toString()}`.replace('//', '/')
+  const host = process.env.REACT_APP_API_HOST
+  const url = host ? new URL(path, host) : path
+  const fullUrl = `${url.toString()}?${params.toString()}`
```
didierofrivia, line 27: Since here we are using the `URL` instance invoked the line above, and being an exception to the rest of the functional paradigm, we could go further and use the defined properties and methods , in this case the `search` one. We could also define localhost as default

```suggestion
  const host = process.env.REACT_APP_API_HOST || LOCALHOST
  const url = new URL(path, host)
  url.search = params
  return new Request(
    url.toString(),
    ...
```

josemigallas: > ```
> // outside the fn
> const LOCALHOST = //127.0.0.1
> ```

This won't work, if target is localhost then it should be passed `API_HOST="http://127.0.0.1"`. It's imperative to make it possible to have a blank host for Openshift (where we don't know the domain) and `URL` doesn't support relative paths. It's not ideal I know but it's required.

didierofrivia: But OS will assign the env var `API_HOST` no? Then it will not take `//127.0.0.1`? Anyways, you could still apply the second part of the change I suggested

josemigallas: No, `API_HOST` it's only useful for local development, in Openshift it is undefined and therefore `new URL()` will throw an error.
By leaving it blank, `fetch` can point to the current domain regardless of what's porta domain.
In summary: no need to define LOCALHOST and can't use only `URL`.

PR #2068, opened 2020-07-21
Fix email validation

josemigallas:

Change the validation to not allow blank.

[THREESCALE-5653: /admin/api/signup.json doesn't require email and creates user without it](https://issues.redhat.com/browse/THREESCALE-5653)

**Verification steps** 
This should return an error `Email can't be blank`:
```
HOST="..." \
TOKEN="..." \
curl -v "$HOST/admin/api/signup.json" -d 'access_token=$TOKEN&name=Peter&username=Griffin&org_name=Petoria'
```

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -83,9 +83,8 @@ class User < ApplicationRecord
                                  :message => MSG_LOGIN_BAD }
 
   #OPTIMIZE: use the same format validations in user, account and invitation, add TESTS!
-  validates :email, length: { :within => 6..255, :allow_blank => true } #r@a.wk
-  validates :email, format: { :with => RE_EMAIL_OK, :allow_blank => false,
-                              :message => MSG_EMAIL_BAD, :unless => :minimal_signup? }
+  validates :email, length: { :within => 6..255, :allow_blank => false } #r@a.wk
+  validates :email, format: { :with => RE_EMAIL_OK, :allow_blank => false, :message => MSG_EMAIL_BAD }
 
   #TODO: this needs tests
   validates :password, length: { :minimum => 6, :allow_blank => true,
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -14,6 +14,7 @@ ActiveRecord::Base.transaction do
   master_plan = ENV.fetch('MASTER_PLAN', 'Master Plan')
   master_domain = ENV['MASTER_DOMAIN'].presence
   master_access_code = ENV.fetch('MASTER_ACCESS_CODE', '')
+  master_email = ENV.fetch('MASTER_EMAIL', 'master@example.com')
 
   master_login = ENV.fetch('MASTER_USER', 'master')
   master_password = ENV.fetch('MASTER_PASSWORD') { SecureRandom.base64(32) }
@@ -65,6 +66,7 @@ ActiveRecord::Base.transaction do
   master_user = master.users.create!(username: master_login, password: master_password, password_confirmation: master_password) do |user|
     user.signup_type = 'minimal'
     user.role = :admin
+    user.email = master_email
   end
   master_user.activate!
 
--- a/test/integration/admin/api/signups_controller_test.rb
+++ b/test/integration/admin/api/signups_controller_test.rb
@@ -19,7 +19,7 @@ class Admin::Api::SignupsControllerTest < ActionDispatch::IntegrationTest
 
       assert_difference(provider.buyers.method(:count)) do
         assert_difference(WebHookWorker.jobs.method(:size)) do
-          post(admin_api_signup_path, format: :json, access_token: token.value, org_name: 'company', username: 'person')
+          post(admin_api_signup_path, format: :json, access_token: token.value, org_name: 'company', username: 'person', email: 'person@example.com')
           assert_response :created
         end
       end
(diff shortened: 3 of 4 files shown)


test/unit/signup/account_manager_test.rb
```diff
@@ -116,6 +116,15 @@ class ProviderAccountManagerTest < Signup::AccountManagerTest
         assert_includes signup_errors, 'Account Organization/Group Name can\'t be blank'
       end
 
+      test 'create provider with wrong params does not create correctly without email' do
+        signup_result = signup_account_manager.create(signup_params(different_account_params: { email: '' }))
```
Martouta, line 120:
```suggestion
        signup_result = signup_account_manager.create(signup_params(different_user_params: { email: '' }))
```

test/unit/signup/account_manager_test.rb
```diff
@@ -116,6 +116,15 @@ class ProviderAccountManagerTest < Signup::AccountManagerTest
...
+
+        refute signup_result.valid?
+        refute signup_result.persisted?
+        signup_errors = signup_result.errors.full_messages
+        assert_includes signup_errors, 'Email can\'t be blank'
```
Martouta, line 125:
```suggestion
        assert_includes signup_errors, 'Email is too short'
```

app/models/user.rb
```diff
@@ -83,9 +83,9 @@ def moderatable
                                  :message => MSG_LOGIN_BAD }
 
   #OPTIMIZE: use the same format validations in user, account and invitation, add TESTS!
-  validates :email, length: { :within => 6..255, :allow_blank => true } #r@a.wk
+  validates :email, length: { :within => 6..255, :allow_blank => false } #r@a.wk
   validates :email, format: { :with => RE_EMAIL_OK, :allow_blank => false,
-                              :message => MSG_EMAIL_BAD, :unless => :minimal_signup? }
+                              :message => MSG_EMAIL_BAD }
```
Martouta, line 88: I think that @macejmic is the best person to review this 😄 

Btw, this might break our own `db/seeds.rb` as well 👀 Because I think that we create the master account through minimal signup and with empty email xD

test/unit/signup/account_manager_test.rb
```diff
@@ -258,6 +267,15 @@ class DeveloperAccountManagerTest < Signup::AccountManagerTest
...
       end
 
+      test 'create developer without email' do
+        signup_result = signup_account_manager.create(signup_params(different_account_params: { email: '' }))
+
+        refute signup_result.valid?
+        refute signup_result.persisted?
+        signup_errors = signup_result.errors.full_messages
+        assert_includes signup_errors, 'Email can\'t be blank'
+      end
```
Martouta, line 277: I'm not sure if writing the same test twice is a good idea. Plus, the change has been done in the User model, so perhaps we should only have a test in that model (in `test/unit/user_test.rb` )

app/models/user.rb
josemigallas: I guess as long as it is something internal.. it can stay

test/unit/signup/account_manager_test.rb
josemigallas: Should it go outside of the class? I guess there should be one per "block", no? `ProviderAccountManagerTest` and `DeveloperAccountManagerTest`?

app/models/user.rb
```diff
@@ -83,9 +83,8 @@ def moderatable
                                  :message => MSG_LOGIN_BAD }
 
   #OPTIMIZE: use the same format validations in user, account and invitation, add TESTS!
-  validates :email, length: { :within => 6..255, :allow_blank => true } #r@a.wk
-  validates :email, format: { :with => RE_EMAIL_OK, :allow_blank => false,
-                              :message => MSG_EMAIL_BAD, :unless => :minimal_signup? }
+  validates :email, length: { :within => 6..255, :allow_blank => false } #r@a.wk
+  validates :email, format: { :with => RE_EMAIL_OK, :allow_blank => false, :message => MSG_EMAIL_BAD }
```
Martouta, line 87: This allow_blank false should have a test here:  https://github.com/3scale/porta/blob/2ead029a58701b3ff3e196c5338177f4f25a97f6/test/unit/user_test.rb#L3

something like:
```ruby
test 'email cannot be blank' do
  user = FactoryBot.build(:admin)
  user.email = nil
  refute user.valid?
  assert_matches user.errors[:email].to_s, 'too short'
end
```

hallelujah, line 87: That's OK to leave like that but you can use Ruby 1.9+ Hash style instead of Hash rockets

app/models/user.rb
```diff
@@ -83,9 +83,8 @@ def moderatable
                                  :message => MSG_LOGIN_BAD }
 
   #OPTIMIZE: use the same format validations in user, account and invitation, add TESTS!
-  validates :email, length: { :within => 6..255, :allow_blank => true } #r@a.wk
-  validates :email, format: { :with => RE_EMAIL_OK, :allow_blank => false,
-                              :message => MSG_EMAIL_BAD, :unless => :minimal_signup? }
```
hallelujah, line 88: minimal signup is allowed on Signup express API and impersonation admin 
https://github.com/3scale/porta/blob/2ead029a58701b3ff3e196c5338177f4f25a97f6/app/controllers/admin/api/signups_controller.rb#L39-L41

https://github.com/3scale/porta/blob/2ead029a58701b3ff3e196c5338177f4f25a97f6/app/lib/signup/impersonation_admin_builder.rb#L18

And we do not need email for that.

Do you intend to modify those behaviour?

Issues:

1. We modify the behaviour of signup express that our customers  might use
2. What will be the default email of the impersonation admin?

josemigallas: I would like to avoid changing the "minimal signup" but I was under the impression that it was required. What I want is really just fix the associated [JIRA](https://issues.redhat.com/browse/THREESCALE-5653).

PR #2076, opened 2020-07-24
Better error messages for creating a Buyer Account from the admin portal UI

Martouta:
Fixes issues 1 & 2 of [THREESCALE-5678](https://issues.redhat.com/browse/THREESCALE-5678)

### DIFF
--- a/app/controllers/buyers/accounts_controller.rb
+++ b/app/controllers/buyers/accounts_controller.rb
@@ -52,6 +52,7 @@ class Buyers::AccountsController < Buyers::BaseController
       redirect_to admin_buyers_account_path(@buyer)
     else
       @user = signup_result.user
+      flash.now[:error] = signup_result.errors.messages.without(:account, :user).values.join('. ')
       render action: :new
     end
   end
--- a/app/lib/logic/signup.rb
+++ b/app/lib/logic/signup.rb
@@ -4,7 +4,7 @@ module Logic
   module Signup
     module Provider
       def create_buyer_possible?
-        account_plans.stock.present?
+        account_plans.default_or_nil.present?
       end
 
       def signup_enabled?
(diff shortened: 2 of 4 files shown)


app/lib/logic/signup.rb
```diff
@@ -4,7 +4,7 @@ module Logic
   module Signup
     module Provider
       def create_buyer_possible?
-        account_plans.stock.present?
+        account_plans.default_or_nil.present?
```
Martouta, line 7: image

safe change 😉

app/controllers/buyers/accounts_controller.rb
```diff
@@ -52,6 +52,7 @@ def create
       redirect_to admin_buyers_account_path(@buyer)
     else
       @user = signup_result.user
+      flash[:error] = signup_result.errors.messages.without(:account, :user).values.join(', ')
```
macejmic, line 55: `full_messages.to_sentence` cannot be used?

hallelujah, line 55: Flash is stored into cookies to persist across redirect unless used.
It is better to use `flash.now` if you know it will be used in the same request

Martouta: It is possible but redundant because the errors of `account` and `user` already appear right below the attribute that has the error.

app/controllers/buyers/accounts_controller.rb
```diff
@@ -52,6 +52,7 @@ def create
       redirect_to admin_buyers_account_path(@buyer)
     else
       @user = signup_result.user
+      flash.now[:error] = signup_result.errors.messages.without(:account, :user).values.join(', ')
```
guicassolato, line 55: It would be nice to have a test for the case when there's more than one message joint.

guicassolato, line 55:
```suggestion
      flash.now[:error] = signup_result.errors.messages.without(:account, :user).values.join('. ')
```

Martouta:
@thomasmaas Exactly 😉

PR #2077, opened 2020-07-24
add cross console

christiemolloy:

Adds the cross-console navigation to the masthead. (It does not work yet, that will come in a later PR)

https://issues.redhat.com/browse/APPDUX-121?jql=text%20~%20%22cross-console%203scale%22

**Verification steps**

1. Go to any page and test that the cross-console nav works / is style correctly.
Note: There is no server-side functionality that works yet.

Screen Shot 2020-07-24 at 10 47 05 AM
Screen Shot 2020-07-24 at 10 47 15 AM
Screen Shot 2020-07-24 at 10 47 24 AM
la


package.json
```diff
@@ -83,6 +83,7 @@
     "@patternfly/react-core": "^3.120.5",
     "@patternfly/react-icons": "^3.14.18",
     "@rails/webpacker": "^4.2.2",
+    "@rh-uxd/integration-core": "1.0.4",
```
damianpm, line 86: You should also commit the changes in `yarn.lock`, as you are adding a new dependency

christiemolloy:
updated @damianpm

damianpm:
> updated @damianpm

Hi @christiemolloy thanks for the update.
Can you fix the [errors found by codeclimate](https://codeclimate.com/github/3scale/porta/pull/2077) in the css files, please?
Most of them are simply because properties should be ordered alphabetically.
And for that errors about selector names, like `Selector pf-c-page__header-brand-toggle should match regex` that's because Patternfly uses a different convention to name selectors. We may change that rule in the future, but for now is enough if you ask scss-lint to ignore that rule with a comment: `// scss-lint:disable SelectorFormat` like [here](https://github.com/3scale/porta/blob/master/app/assets/stylesheets/provider/_patternfly_tabs_fake.scss#L1)

Another thing, I'm going to change the base branch from `master` to `patternfly-cross-console-navigation` that way we can run some tests in our preview environment

app/views/shared/provider/_header.html.slim
```diff
@@ -1,15 +1,32 @@
...
 
   .pf-c-page__header-nav
     = render 'shared/api_selector'
 
+  .app-context-selector-wrapper
+    .pf-c-context-selector
+      button.pf-c-context-selector__toggle.PopNavigation-trigger.u-toggler
+        span.pf-c-context-selector__toggle-text
+          | 3scale API Management
+          span.pf-c-context-selector__toggle-icon[aria-hidden="true"]
```
josemigallas, line 17:
```suggestion
          span.pf-c-context-selector__toggle-icon aria-hidden="true"
```

hallelujah:

PR #2081, opened 2020-07-27
OAS autocomplete: Supporting all methods, not just get

damianpm:

In a previous PR (https://github.com/3scale/porta/pull/2022) we added support for `get` methods in paths, but other methods as `post`, `put`, or `delete` are missing.

https://issues.redhat.com/browse/THREESCALE-5516

**Verification steps** 

Go to ActiveDocs and create a new OAS 3 spec, you can use this example: https://github.com/OAI/OpenAPI-Specification/blob/master/examples/v3.0/petstore-expanded.json
All paths should work fine
Also tests with our Echo Api default example, autocomplete should still work

### DIFF
--- a/app/javascript/packs/OAS3Autocomplete.js
+++ b/app/javascript/packs/OAS3Autocomplete.js
@@ -20,22 +20,36 @@ const addAutocompleteToParam = (param: Param, accountData: AccountData): Param =
   const xDataKey = param[X_DATA_ATTRIBUTE]
   const autocompleteData = accountData[xDataKey]
   const paramHasAutocompleteData = autocompleteData && autocompleteData.length > 0 &&
-   autocompleteData.every(param => param.name !== '')
+    autocompleteData.every(param => param.name !== '')
 
   return paramHasAutocompleteData
     ? {
       ...param,
       examples: autocompleteData.reduce((examples, item) => (
-        [...examples, {summary: item.name, value: item.value}]
-      ), [{summary: X_DATA_PARAMS_DESCRIPTIONS[xDataKey], value: '-'}])
+        [...examples, { summary: item.name, value: item.value }]
+      ), [{ summary: X_DATA_PARAMS_DESCRIPTIONS[xDataKey], value: '-' }])
     }
     : param
 }
 
-const getPathParameters = (path) => {
-  const hasParamsInGet = path.get && path.get.parameters
-  return hasParamsInGet ? path.get.parameters : path.parameters
-}
+const getPathParameters = (path) => (
+  Object.keys(path).reduce((params, item) => {
+    params[item] = path[item].parameters
+    return params
+  }, {})
+)
+
+const injectParametersToPath = (path, pathParameters, accountData) => (
+  Object.keys(pathParameters).reduce((parameters, item) => {
+    parameters[item] = {
+      ...path[item],
+      parameters: pathParameters[item]
+        ? pathParameters[item].map(param => X_DATA_ATTRIBUTE in param ? addAutocompleteToParam(param, accountData) : param)
+        : []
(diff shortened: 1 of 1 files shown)


app/javascript/packs/OAS3Autocomplete.js
```diff
@@ -20,22 +20,36 @@ const addAutocompleteToParam = (param: Param, accountData: AccountData): Param =
...
-      ), [{summary: X_DATA_PARAMS_DESCRIPTIONS[xDataKey], value: '-'}])
+        [...examples, { summary: item.name, value: item.value }]
+      ), [{ summary: X_DATA_PARAMS_DESCRIPTIONS[xDataKey], value: '-' }])
     }
     : param
 }
 
-const getPathParameters = (path) => {
-  const hasParamsInGet = path.get && path.get.parameters
-  return hasParamsInGet ? path.get.parameters : path.parameters
```
josemigallas, line 37: How about this case when `path.get` was undefined? Is `path.parameters` still possible?

app/javascript/packs/OAS3Autocomplete.js
```diff
@@ -20,22 +20,36 @@ const addAutocompleteToParam = (param: Param, accountData: AccountData): Param =
...
+  Object.keys(pathParameters).reduce((parameters, item) => {
+    parameters[item] = {
+      ...path[item],
+      parameters: pathParameters[item]
+        ? pathParameters[item].map(param => X_DATA_ATTRIBUTE in param ? addAutocompleteToParam(param, accountData) : param)
+        : []
+    }
+    return parameters
+  }, {})
+)
```
didierofrivia, line 52:
```suggestion
const injectParametersToPath = (path, pathParameters, accountData) => (
  Object.keys(pathParameters).reduce((parameters, item) => (
     { ...parameters, [item]: path[item], 
        parameters: pathParameters[item]
          ? pathParameters[item].map(param => X_DATA_ATTRIBUTE in param ? addAutocompleteToParam(param, accountData) : param)
        : []
    }, {})
)
```

You could extract also to a separate fn the pathParameters mod

app/javascript/packs/OAS3Autocomplete.js
damianpm: @josemigallas @didierofrivia 
I've made some changes, it has to work in 2 scenarios:
1. When parameters is an array
Example: 
```
{
  parameters: [...]
}
```
2. When parameters comes as array, but nested in an object
Example: 
```
{
  get: {
    description: "...",
    parameters: [...]
  },
  post: {
    description: "...",
    parameters: [...]
  }
}
```

app/javascript/packs/OAS3Autocomplete.js
```diff
@@ -33,22 +33,36 @@ const addAutocompleteToParam = (param: Param, accountData: AccountData): Param =
...
-        ? pathParameters[item].map(param => X_DATA_ATTRIBUTE in param ? addAutocompleteToParam(param, accountData) : param)
-        : []
+const injectParametersField = (key, value, accountData) => {
+  const parametersArray = value.map(param => X_DATA_ATTRIBUTE in param ? addAutocompleteToParam(param, accountData) : param)
+  if (key === 'parameters') {
+    return parametersArray
+  } else {
+    return {
+      ...value,
+      parameters: parametersArray
```
didierofrivia, line 53:
```suggestion
const injectParametersField = (key, value, accountData) => {
  const parametersArray = value.map(param => X_DATA_ATTRIBUTE in param ? addAutocompleteToParam(param, accountData) : param)
  return (key === 'parameters') ? parametersArray : { ...value, parameters: parametersArray }
```

PR #2082, opened 2020-07-27
Updates Portafly specs with Apicurio

josemigallas:

Replaces existing `portafly-api.yml` with one created in Apicurio Studio. It is the same content although some Summary and Description fields have been changed for a better, smoother development. This way we have the most consistency and maintain the same format.

It includes a new endpoint for `/applications` as well, since I'm working in https://issues.redhat.com/browse/THREESCALE-5479.

The idea is from now on we all work from [this project in Apicurio Studio](https://studio-ws.apicur.io/sharing/569f83da-2210-4716-80ef-9adb9736f19a) and not the fire directly. Any changes should be done in Studio, then downloaded as YAML:
Screen Shot 2020-07-27 at 17 19 49

Screen Shot 2020-07-27 at 17 20 09

I recommend besides checking the code changes in this PR, go to Apicurio Studio and see everything makes sense there.

### DIFF
--- a/portafly/portafly_api.yml
+++ b/portafly/portafly_api.yml
@@ -9,43 +9,11 @@ info:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0
 paths:
-  /developer/accounts:
-    summary: A list of developer accounts
-    description: A list of developer accounts.
-    get:
-      responses:
-        '200':
-          $ref: '#/components/responses/developerAccountList'
-      operationId: developerAccountList
-      summary: Get developer account list
-    parameters:
-      - name: page
-        description: Page in the paginated list. Defaults to 1.
-        schema:
-          type: integer
-        in: query
-        required: false
-      - name: per_page
-        description: Number of results per page. Default and max is 500.
-        schema:
-          type: integer
-        in: query
-        required: false
-      - name: state
-        description: Account state.
-        schema:
-          type: string
-        in: query
   /users:
-    summary: A list of user
-    description: A list of users
+    summary: Users
+    description: ""
     get:
-      responses:
-        '200':
-          $ref: '#/components/responses/userList'
-      operationId: userList
-      summary: Get account users list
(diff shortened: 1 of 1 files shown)


portafly/portafly_api.yml
```diff
@@ -9,45 +9,44 @@ info:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0
 paths:
-  /developer/accounts:
-    summary: A list of developer accounts
-    description: A list of developer accounts.
+  /users:
+    summary: Users
+    description: ""
```
damianpm, line 14: I guess the idea is to not have a `description` for all possible operations, but a description for every single one, like defined in line 18
Then we can remove this, instead of keeping an empty string

portafly/portafly_api.yml
```diff
@@ -9,45 +9,44 @@ info:
...
-        schema:
-          type: string
-        in: query
-  /users:
-    summary: A list of user
-    description: A list of users
-    get:
       responses:
-        '200':
+        "200":
```
damianpm, line 28: Single quotes ?

portafly/portafly_api.yml
```diff
@@ -56,47 +55,127 @@ paths:
...
+          $ref: '#/components/schemas/Plan'
+        created_on:
           format: date-time
           description: When it was created
           type: string
-        updated_at:
+        traffic_on:
           format: date-time
-          description: When it was last updated
+          description: ??
```
damianpm, line 95: What happened here, typo?

josemigallas: It happened I didn't know what to put there.

didierofrivia (changes requested):
I'm halting this one because of the introduction of `/users` path that is not defined just yet. We might adopt KC and pull the data from there, we might well define what we want to get from KC first.

josemigallas:
OK we'll do this when it makes sense.

PR #2083, opened 2020-07-29
Uri shortener for long host label endpoints

guicassolato:
It will automatically shorten host labels of system-generated proxy endpoints, that exceed the 63-char limit imposed by RFC-1035.

Closes [THREESCALE-5089](https://issues.redhat.com/browse/THREESCALE-5089)

### DIFF
--- a/app/models/proxy.rb
+++ b/app/models/proxy.rb
@@ -216,11 +216,13 @@ class Proxy < ApplicationRecord
     def generate(name)
       template = config.fetch(name.try(:to_sym)) { return }
 
-      format template, {
+      uri = format template, {
         system_name: service.parameterized_system_name, account_id: service.account_id,
         tenant_name: provider_subdomain,
         env: proxy.proxy_env, port: proxy.proxy_port
       }
+
+      UriShortener.call(uri).to_s
     end
   end
 
--- a/app/services/service_creator.rb
+++ b/app/services/service_creator.rb
@@ -22,10 +22,6 @@ class ServiceCreator
   def call(params = {})
     call!(params)
   rescue ActiveRecord::RecordInvalid
-    return if !proxy || proxy.errors[:endpoint].none? { |message| message =~ /is too long/ }
-
-    @service.errors.add(:system_name, :invalid_for_proxy_endpoints)
-
     false
   end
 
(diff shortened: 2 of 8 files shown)


hallelujah:
I wonder, instead of using `system_name` and `tenant_name` can't we just generate a UUID?

guicassolato:
> I wonder, instead of using `system_name` and `tenant_name` can't we just generate a UUID?

I think the problem is that this will become the public URL of the service. Customers will probably prefer keep getting something easier to remember and yet still be more or less in control of the host name (up to a point).

hallelujah:
I really do not know, it is a copy paste anyway, used in scripts not something you type in your browser url bar.

guicassolato:
@hallelujah, the 63-char limit per hostname label is currently being easily reached in Openshift when service discovery is used because we generate the system name. It not (directly) chosen by the customer. It's `"#{openshift_namespace}-#{service_name}"`. And then system name of the service is used in the template to generate the proxy endpoints, where we append yet `"-#{tenant_org_name}"-apicast-staging` or `"-#{tenant_org_name}"-apicast-production`.

hallelujah:
I was thinking of using a UUID or something instead of the template when it is generated by ServiceDiscovery
Because shortening is prone to name collision. 

2 services with long enough name can generate a similar system name that will be shortened, leading to 2 generated same URLs pointing to 2 services (one accepted by OpenShift, the other not_
We can then add annotations to the route so it fixes the issue once and for all

guicassolato:
> I was thinking of using a UUID or something instead of the template when it is generated by ServiceDiscovery
> Because shortening is prone to name collision.
> 
> 2 services with long enough name can generate a similar system name that will be shortened, leading to 2 generated same URLs pointing to 2 services (one accepted by OpenShift, the other not_
> We can then add annotations to the route so it fixes the issue once and for all

@hallelujah, that's why I used a SHA-1 based on the original hostname label. The chance of name collision here is virtually zero (or as low as SHA-1 algorithm has to generate two hashes with the same 7-char prefix for two different sources). In other words, given that `"#{openshift_namespace}-#{service_name}"` has to be unique, we have pretty much the same risk of generating duplicate hostnames as, e.g., git has of generating two short commit hashes 😉.

PR #2088, opened 2020-07-30
Completely remove dev domain

hallelujah:
It is replaced by domain substitution

See https://github.com/3scale/porta/pull/1904 and https://github.com/3scale/porta/blob/a3a115c1c24d73f90b17a8f630cb37911dc6ce42/lib/three_scale/domain_substitution.rb#L14-L96

Relates to https://github.com/3scale/platform/issues/258
Closes https://issues.redhat.com/browse/THREESCALE-5265


test/unit/presenters/oauth_flow_presenter_test.rb
```diff
@@ -8,8 +8,11 @@ class OAuthFlowPresenterTest < ActiveSupport::TestCase
     @authentication_provider = FactoryBot.build_stubbed(:authentication_provider, kind: 'auth0')
     @provider = @authentication_provider.account
 
-    @request = stubs(:request)
-    @request.stubs(scheme: 'http', domain: 'example.com', query_parameters: {'plan_id' => 42})
+    env = {
+      'HTTP_HOST' => 'example.com',
+      'QUERY_STRING' => 'plan_id=42',
```
Martouta, line 13:
```suggestion
      'QUERY_STRING' => 'plan_id=42'
```

Martouta:
This can be done in another PR but all these `.lvh.me` in the installation guide should be removed:
https://github.com/3scale/porta/blob/master/INSTALL.md#setup-database

app/lib/three_scale/domain.rb
```diff
@@ -1,19 +1,20 @@
...
+        "#{scheme}://#{host}:#{port}"
+      else
+        "#{scheme}://#{host}"
+      end
     end
 
     def self.callback_endpoint(request, account, host = request.host)
       endpoint = if account.master?
-                   current_endpoint(request, account.domain) + '/master/devportal'.freeze
+                   current_endpoint(request, account.external_domain) + '/master/devportal'.freeze
```
Martouta, line 17: I know that what I am gonna say is not related to your PR but... does this work? 😕 
First of all, `master` does not have a developer portal so if this is correct, the name is definitely confusing. Second, and more importantly, when I check for this route I find only this:
image

Martouta:
Have you tested this PR in Preview yet @hallelujah ? u should before merging this.

app/lib/three_scale/domain.rb
hallelujah: Master domain == self_domain 
It will work for all providers

hallelujah:
Yes tested in preview, sidekiq, some mailers, and sso

app/lib/three_scale/domain.rb
Martouta: I understood that part @hallelujah . The part that concerns me is why it is `'/master/devportal'` and not `'/master/devportal/auth'`. And "what is this 'devportal' thing of master?".

This is why I said that it is not related to your PR. If it was working before your PR, it will keep working now, and if it wasn't working before your PR, then for sure it does not work now yet xD

But if you say that it works, I trust you.

hallelujah: Because it is just a method, but does not return the whole path ... used somewhere else

PR #2089, opened 2020-07-30
Fix proxy config affecting change on create

guicassolato:
The exclamation sign, close to “Integration” in the sidebar menu of the Admin Porta and also in the Dashboard, is showing up for API products just created, right after finishing the provider wizard. This should not happen and neither it was when ProxyConfigAffectingChanges were first introduced. I guess some change later on caused the creation of the proxy’s affecting change history to be postponed to _after_ the creation of the first proxy config itself – likely #1658.

This PR ensure the proxy’s affecting change history exists since creation of the proxy object.

### DIFF
--- a/app/models/proxy.rb
+++ b/app/models/proxy.rb
@@ -536,11 +536,9 @@ class Proxy < ApplicationRecord
     super
   end
 
-  def find_or_create_proxy_config_affecting_change
+  def affecting_change_history
     proxy_config_affecting_change || create_proxy_config_affecting_change
   end
-  alias affecting_change_history find_or_create_proxy_config_affecting_change
-  private :find_or_create_proxy_config_affecting_change
 
   def pending_affecting_changes?
     return unless apicast_configuration_driven?
--- a/app/models/proxy.rb
+++ b/app/models/proxy.rb
@@ -27,6 +27,7 @@ class Proxy < ApplicationRecord
 
   has_one :proxy_config_affecting_change, dependent: :delete
   private :proxy_config_affecting_change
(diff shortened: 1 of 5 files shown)


test/unit/tasks/proxy_test.rb
```diff
@@ -7,14 +7,14 @@ class ProxyTest < ActiveSupport::TestCase
...
         @providers = FactoryBot.create_list(:provider_account, 3)
-        proxies.each(&:affecting_change_history) # so change history exists for the proxies associated with all these accounts
 
         @proxy_with_legit_change = proxies.last
         @legit_change_date = 2.minutes.from_now.freeze
         proxy_with_legit_change.affecting_change_history.update_column(:updated_at, legit_change_date) # so change history of this proxy has been updated
 
         @providers << FactoryBot.create(:provider_account) # change history for the proxy associated with this account does not exist
         @proxy_without_change_history = proxies.last
+        ProxyConfigAffectingChange.where(proxy: proxy_without_change_history).delete_all
```
hallelujah, line 17: Do you delete all because it is created on proxy creation?

app/models/proxy.rb
```diff
@@ -651,6 +650,7 @@ def create_default_proxy_rule
   end
 
   def create_proxy_config_affecting_change(*)
+    return unless persisted?
```
hallelujah, line 653: This is called `after_commit on: :create` so it should be persisted already

hallelujah, line 653:
```suggestion
    return if new_record?
```

PR #2092, opened 2020-07-31
Deliver invitations async and fix notification message

josemigallas:

Invitation delivery is notified as "successfully sent" when it really isn't.
1. Change the message to "it will be sent"
2. Update the colum `sent_at` only after the email has been sent by the Mailer

Screen Shot 2020-08-13 at 12 10 15

[THREESCALE-4360: Using "Invite New User" button, no verification that email is sent](https://issues.redhat.com/browse/THREESCALE-4360)

**Verification steps** 

1. Go to `p/admin/account/invitations` and create a new invitation.
2. The notification should say it will be sent
3. Refresh and check the column `Sent` shows the correct value (perhaps blanks at first, email should be sent not inmediately)

### DIFF
--- a/app/controllers/buyers/invitations_controller.rb
+++ b/app/controllers/buyers/invitations_controller.rb
@@ -12,7 +12,7 @@ class Buyers::InvitationsController < Buyers::BaseController
   create! do |success, failure|
     success.html do
       redirect_to admin_buyers_account_invitations_path(@account),
-                  notice: 'Invitation was successfully sent.'
+                  notice: 'Invitation will be sent soon.'
     end
   end
 
@@ -26,7 +26,7 @@ class Buyers::InvitationsController < Buyers::BaseController
 
     respond_to do |format|
       format.html do
-        flash[:success] = "Invitation was successfully resent"
+        flash[:success] = 'Invitation will be resent soon.'
         redirect_to(admin_buyers_account_invitations_path(@account))
       end
       format.xml  { head :ok }
--- a/app/controllers/provider/admin/account/invitations_controller.rb
+++ b/app/controllers/provider/admin/account/invitations_controller.rb
@@ -10,7 +10,7 @@ class Provider::Admin::Account::InvitationsController < Provider::Admin::Account
   create! do |success, failure|
     success.html do
       redirect_to provider_admin_account_invitations_path,
-                  notice: 'Invitation was successfully sent.'
+                  notice: 'Invitation will be sent soon.'
     end
   end
 
@@ -24,7 +24,7 @@ class Provider::Admin::Account::InvitationsController < Provider::Admin::Account
 
     respond_to do |format|
       format.html do
-        flash[:success] = "Invitation was successfully resent"
+        flash[:success] = 'Invitation will be resent soon.'
         redirect_to provider_admin_account_invitations_path
       end
       format.xml  { head :ok }
--- a/app/models/invitation.rb
+++ b/app/models/invitation.rb
@@ -39,6 +39,7 @@ class Invitation < ApplicationRecord
 
   def resend
     notify_invitee unless accepted?
+    update!(sent_at: nil)
   end
 
   private
--- a/app/models/invitation.rb
+++ b/app/models/invitation.rb
@@ -38,9 +38,7 @@ class Invitation < ApplicationRecord
   end
 
   def resend
-    return if accepted?
-    notify_invitee
-    update!(sent_at: nil)
+    notify_invitee unless accepted?
   end
 
   private
--- a/app/models/invitation.rb
+++ b/app/models/invitation.rb
@@ -58,14 +58,7 @@ class Invitation < ApplicationRecord
   end
 
   def notify_invitee
-    # TODO: maybe subclass to Invitation and ProviderInvitation
-    if account.provider?
-      ProviderInvitationMailer.invitation(self).deliver_now
-    else
-      InvitationMailer.invitation(self).deliver_now
-    end
-
-    update_column(:sent_at, Time.zone.now)
+    SendUserInvitationWorker.perform_later(id)
   end
 
 end
--- a/app/models/invitation.rb
+++ b/app/models/invitation.rb
@@ -30,6 +30,7 @@ class Invitation < ApplicationRecord
   end
 
   def accepted?
+    return true
     accepted_at.present?
   end
 
@@ -38,7 +39,8 @@ class Invitation < ApplicationRecord
   end
 
   def resend
-    notify_invitee unless accepted?
+    return if accepted?
+    notify_invitee
     update!(sent_at: nil)
   end
 
@@ -59,7 +61,7 @@ class Invitation < ApplicationRecord
   end
 
   def notify_invitee
-    SendUserInvitationWorker.perform_later(id)
+    SendUserInvitationWorker.perform_later(self)
   end
 
 end
--- a/app/models/invitation.rb
+++ b/app/models/invitation.rb
@@ -30,7 +30,6 @@ class Invitation < ApplicationRecord
   end
 
   def accepted?
-    return true
     accepted_at.present?
   end
 
--- a/app/workers/send_user_invitation_worker.rb
+++ b/app/workers/send_user_invitation_worker.rb
@@ -3,7 +3,7 @@
 class SendUserInvitationWorker < ApplicationJob
   queue_as :default
 
-  RETRY_ERRORS = [
+  ERRORS = [
     OpenSSL::SSL::SSLError,
     Net::SMTPAuthenticationError,
     Net::SMTPFatalError,
@@ -14,21 +14,15 @@ class SendUserInvitationWorker < ApplicationJob
     SocketError,
   ].freeze
 
-  DISCARD_ERRORS = [
-    ActiveRecord::RecordNotFound
-  ].freeze
-
   def perform(invitation_id)
     invitation = Invitation.find(invitation_id)
 
     mailer = invitation.account.provider? ? ProviderInvitationMailer : InvitationMailer
     mailer.invitation(invitation).deliver_now!
 
-    invitation.update!(sent_at: Time.zone.now)
-  rescue *RETRY_ERRORS => error
-    logger.error(error.message)
-  rescue *DISCARD_ERRORS => error
-    logger.error(error.message)
-    false
+    invitation.update(sent_at: Time.zone.now)
+  rescue *ERRORS => e
+    logger.error(e.message)
+    retry_job
   end
 end
--- a/app/workers/send_user_invitation_worker.rb
+++ b/app/workers/send_user_invitation_worker.rb
@@ -11,12 +11,10 @@ class SendUserInvitationWorker < ApplicationJob
     Net::SMTPSyntaxError,
     Net::SMTPUnknownError,
     Net::SMTPUnsupportedCommand,
-    SocketError,
+    SocketError
   ].freeze
 
-  def perform(invitation_id)
-    invitation = Invitation.find(invitation_id)
-
+  def perform(invitation)
     mailer = invitation.account.provider? ? ProviderInvitationMailer : InvitationMailer
     mailer.invitation(invitation).deliver_now!
 
--- a/app/workers/send_user_invitation_worker.rb
+++ b/app/workers/send_user_invitation_worker.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+class SendUserInvitationWorker < ApplicationJob
+  queue_as :default
+  include Net
+
+  def perform(invitation_id)
+    invitation = Invitation.find(invitation_id)
+
+    mailer = invitation.account.provider? ? ProviderInvitationMailer : InvitationMailer
+    mailer.invitation(invitation).deliver_now!
+
+    invitation.update!(sent_at: Time.zone.now)
+  rescue OpenSSL::SSL::SSLError, Net::SMTPError, SocketError, ActiveRecord::RecordNotFound => err
+    logger.error(err.message)
+  end
+end
--- a/app/workers/send_user_invitation_worker.rb
+++ b/app/workers/send_user_invitation_worker.rb
@@ -14,19 +14,17 @@ class SendUserInvitationWorker < ApplicationJob
     SocketError
   ].freeze
 
-  DISCARD_ERRORS = [
-    ActiveJob::DeserializationError
-  ].freeze
+  rescue_from ActiveJob::DeserializationError do |e|
+    logger.error("SendUserInvitationWorker#perform raised #{e.class} with message #{e.message}")
+  end
 
   def perform(invitation)
     mailer = invitation.account.provider? ? ProviderInvitationMailer : InvitationMailer
     mailer.invitation(invitation).deliver_now!
 
     invitation.update(sent_at: Time.zone.now)
-  rescue *DISCARD_ERRORS => ex
-    Rails.logger.info "SphinxIndexationWorker#perform raised" #{ex.class} with message #{ex.message}"
   rescue *RETRY_ERRORS => e
-    logger.error(e.message)
+    logger.error("SendUserInvitationWorker#perform raised #{e.class} with message #{e.message}")
     retry_job
   end
 end
--- a/app/workers/send_user_invitation_worker.rb
+++ b/app/workers/send_user_invitation_worker.rb
@@ -2,7 +2,21 @@
 
 class SendUserInvitationWorker < ApplicationJob
   queue_as :default
-  include Net
+
+  RETRY_ERRORS = [
+    OpenSSL::SSL::SSLError,
+    Net::SMTPAuthenticationError,
+    Net::SMTPFatalError,
+    Net::SMTPServerBusy,
+    Net::SMTPSyntaxError,
+    Net::SMTPUnknownError,
+    Net::SMTPUnsupportedCommand,
+    SocketError,
+  ].freeze
+
+  DISCARD_ERRORS = [
+    ActiveRecord::RecordNotFound
+  ].freeze
 
   def perform(invitation_id)
     invitation = Invitation.find(invitation_id)
@@ -11,7 +25,10 @@ class SendUserInvitationWorker < ApplicationJob
     mailer.invitation(invitation).deliver_now!
 
     invitation.update!(sent_at: Time.zone.now)
-  rescue OpenSSL::SSL::SSLError, Net::SMTPError, SocketError, ActiveRecord::RecordNotFound => err
-    logger.error(err.message)
+  rescue *RETRY_ERRORS => error
+    logger.error(error.message)
+  rescue *DISCARD_ERRORS => error
+    logger.error(error.message)
+    false
   end
 end
--- a/app/workers/send_user_invitation_worker.rb
+++ b/app/workers/send_user_invitation_worker.rb
@@ -3,7 +3,7 @@
 class SendUserInvitationWorker < ApplicationJob
   queue_as :default
 
-  ERRORS = [
+  RETRY_ERRORS = [
     OpenSSL::SSL::SSLError,
     Net::SMTPAuthenticationError,
     Net::SMTPFatalError,
@@ -14,12 +14,18 @@ class SendUserInvitationWorker < ApplicationJob
     SocketError
   ].freeze
 
+  DISCARD_ERRORS = [
+    ActiveJob::DeserializationError
+  ].freeze
+
   def perform(invitation)
     mailer = invitation.account.provider? ? ProviderInvitationMailer : InvitationMailer
     mailer.invitation(invitation).deliver_now!
 
     invitation.update(sent_at: Time.zone.now)
-  rescue *ERRORS => e
+  rescue *DISCARD_ERRORS => ex
+    Rails.logger.info "SphinxIndexationWorker#perform raised" #{ex.class} with message #{ex.message}"
+  rescue *RETRY_ERRORS => e
     logger.error(e.message)
     retry_job
   end
--- a/test/unit/invitation_test.rb
+++ b/test/unit/invitation_test.rb
@@ -44,7 +44,7 @@ class InvitationTest < ActiveSupport::TestCase
     assert_not_nil invitation.token
   end
 
-  test 'creates a worker when created' do
+  test 'enqueues a worker when created' do
     SendUserInvitationWorker.expects(:perform_later)
     @provider.invitations.create!(:email => 'buddy@example.net')
   end
--- a/test/unit/invitation_test.rb
+++ b/test/unit/invitation_test.rb
@@ -44,39 +44,20 @@ class InvitationTest < ActiveSupport::TestCase
     assert_not_nil invitation.token
   end
 
-  test 'sends invitation email when created' do
-    ProviderInvitationMailer.any_instance.expects(:invitation)
+  test 'creates a worker when created' do
+    SendUserInvitationWorker.expects(:perform_later)
     @provider.invitations.create!(:email => 'buddy@example.net')
   end
 
-  test 'sets sent_at field when created' do
-    # This is needed, database will not save fractions of seconds
-    # so something like:
-    # Time.now.to_f #=> 1512012308.293877
-    # will be saved in DB as 1512012308
-    Timecop.freeze(Time.zone.parse('2017-11-23 03:25:08 UTC +00:00')) do
-      invitation = @provider.invitations.create!(:email => 'buddy@example.net')
-      assert_equal Time.zone.now, invitation.sent_at
-    end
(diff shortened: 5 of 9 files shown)


app/models/invitation.rb
```diff
@@ -60,12 +60,18 @@ def sent?
...
     else
-      InvitationMailer.invitation(self).deliver_now
+      InvitationMailer.register_observer(InvitationDeliveryObserver)
+      InvitationMailer.invitation(self).deliver_later
     end
+  end
 
-    update_column(:sent_at, Time.zone.now)
+  class InvitationDeliveryObserver
+    def self.delivered_email()
```
Martouta, line 72:
```suggestion
    def self.delivered_email
```

app/models/invitation.rb
```diff
@@ -60,12 +60,18 @@ def sent?
...
+      update_column(:sent_at, Time.zone.now)
```
Martouta, line 73: Does this suggestion work? If so, this is better.
```suggestion
      update!(sent_at: Time.zone.now)
```

app/models/invitation.rb
```diff
@@ -60,12 +60,18 @@ def sent?
...
     if account.provider?
-      ProviderInvitationMailer.invitation(self).deliver_now
+      ProviderInvitationMailer.register_observer(InvitationDeliveryObserver)
+      ProviderInvitationMailer.invitation(self).deliver_later
     else
-      InvitationMailer.invitation(self).deliver_now
+      InvitationMailer.register_observer(InvitationDeliveryObserver)
+      InvitationMailer.invitation(self).deliver_later
     end
+  end
```
Martouta, line 69: We should DRY this. Suggestion below 👇 
We should also ensure that all this has a test.

```ruby
def notify_invitee
  mailer_klass = account.provider? ? ProviderInvitationMailer : InvitationMailer
  mailer_klass.register_observer(InvitationDeliveryObserver)
  mailer_klass.invitation(self).deliver_later
end
```

app/controllers/buyers/invitations_controller.rb
```diff
@@ -26,7 +26,7 @@ def resend
 
     respond_to do |format|
       format.html do
-        flash[:success] = "Invitation was successfully resent"
+        flash[:success] = "Invitation will be resent soon."
```
Martouta, line 29:
```suggestion
        flash[:success] = 'Invitation will be resent soon.'
```

app/controllers/provider/admin/account/invitations_controller.rb
```diff
@@ -24,7 +24,7 @@ def resend
 
     respond_to do |format|
       format.html do
-        flash[:success] = "Invitation was successfully resent"
+        flash[:success] = "Invitation will be resent soon."
```
Martouta, line 27:
```suggestion
        flash[:success] = 'Invitation will be resent soon.'
```

app/models/invitation.rb
josemigallas: Sounds good. I think the observer is going to be "permanent" tho. It's supposed to be added to Mailer::Base once at initialization time.. but this would mean the PR will be much bigger than we expected, how strange 😉

app/models/invitation.rb
```diff
@@ -60,12 +60,18 @@ def sent?
   def notify_invitee
     # TODO: maybe subclass to Invitation and ProviderInvitation
     if account.provider?
-      ProviderInvitationMailer.invitation(self).deliver_now
+      ProviderInvitationMailer.register_observer(InvitationDeliveryObserver)
```
hallelujah, line 63: You should not use an observer, it is deprecated and we keep it for legacy. Work on removing it is a tech-debt task

app/models/invitation.rb
hallelujah, line 73: Put it in the mailer job

app/models/invitation.rb
```diff
@@ -60,12 +60,18 @@ def sent?
...
+      ProviderInvitationMailer.invitation(self).deliver_later
     else
-      InvitationMailer.invitation(self).deliver_now
+      InvitationMailer.register_observer(InvitationDeliveryObserver)
+      InvitationMailer.invitation(self).deliver_later
```
hallelujah, line 67: It is a pity we are still on 5.0 because we could use another Job then ActionMailer::DeliveryJob to do the deilvery in background

Instead of using deliver_later, I would create (temporarily until we upgrade) another mailer job that will do the `deliver_now` and set the `sent_at` attribute

app/jobs/send_user_invitation_job.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class SendUserInvitationJob < ActiveJob::Base
+  queue_as :user_invites
```
Martouta, line 4: A new queue just for this no plz 😭 for me either `low` or `default`

app/jobs/send_user_invitation_job.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class SendUserInvitationJob < ActiveJob::Base
```
Martouta, line 3:
```suggestion
class SendUserInvitationJob < ApplicationJob
```

app/jobs/send_user_invitation_job.rb
```diff
@@ -0,0 +1,14 @@
...
+  queue_as :user_invites
+
+  def perform(invitation_id)
+    invitation = Invitation.find(invitation_id)
+
+    mailer_klass = account.provider? ? ProviderInvitationMailer : InvitationMailer
+    mailer_klass.invitation(invitation).deliver_now
+
+    invitation.update_column(sent_at: Time.zone.now)
+  end
```
Martouta, line 13: I have some comments about this:
1. What if the invitation does not exist anymore? Then you should rescue that exception and ignore it.
2. Why not enqueuing the whole object instead of the ID? then it will use GlobalID
3. This is the most important one... why not doing all this job inside the mailer directly? I do not see the benefit of adding an extra job to be a middle man

app/jobs/send_user_invitation_job.rb
josemigallas: OK I don't know 🤷‍♂️ .`default` then?

app/jobs/send_user_invitation_job.rb
josemigallas: Should I move the job under `workers` folder too?

app/jobs/send_user_invitation_job.rb
josemigallas: I think the idea of using a job is to update the column async, after the email is sent, no?

app/jobs/send_user_invitation_job.rb
Martouta: yup : 👍

app/jobs/send_user_invitation_job.rb
Martouta: I do not understand the reason why not the Observer, since [it is documented even for rails 6](https://apidock.com/rails/v6.0.0/ActionMailer/Base) and does not seem deprecated at all.
I think that @hallelujah confused these observers, with [these observers](https://github.com/rails/rails-observers) that were removed from Rails already (but can work using that gem).

But if I am wrong about this, then IMO, you should just add the `invitation.update_column(sent_at: Time.zone.now)` inside the mailer itself, or another option is doing as u did here but doing `deliver_now!`with the `!`to raise an error if it fails.

hallelujah: Right, sorry! I was confused, thanks @Martouta !

test/workers/send_user_invitation_worker_test.rb
```diff
@@ -0,0 +1,67 @@
...
+
+    buyer_account = FactoryBot.create(:buyer_account)
+    invitation = FactoryBot.create(:invitation, account: buyer_account)
+
+    assert_nil invitation.sent_at
+    SendUserInvitationWorker.new.perform(invitation.id)
+    assert invitation.reload.sent_at
+  end
+
+  def test_perform_ssl_error
```
josemigallas, line 28:
```suggestion
  def test_handles_ssl_error
```

thomasmaas:
@josemigallas would be nice to see some screen shots of the different states (or animated gif) so we can be sure we're creating optimal ux within possibilities. For example, I can imagine the sent_at column should say 'not sent yet' or 'sending' instead of being empty…

josemigallas:
@thomasmaas From UX perspective there are 3 cases:
- New invitation not yet sent -> Right now there's simply a `-` indicating there's no date. We can change it to _Sending_, and _Failed_ in case of error.
- Invitation sent first time -> The date
- Invitation resent -> Currently the last value of `sent_at` but should we be more descriptive _Resending... / Failed_?

app/workers/send_user_invitation_worker.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class SendUserInvitationWorker < ApplicationJob
+  queue_as :default
+  include Net
```
guicassolato, line 5: This should not be needed.

app/workers/send_user_invitation_worker.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  def perform(invitation_id)
+    invitation = Invitation.find(invitation_id)
+
+    mailer = invitation.account.provider? ? ProviderInvitationMailer : InvitationMailer
+    mailer.invitation(invitation).deliver_now!
+
+    invitation.update!(sent_at: Time.zone.now)
+  rescue OpenSSL::SSL::SSLError, Net::SMTPError, SocketError, ActiveRecord::RecordNotFound => err
+    logger.error(err.message)
```
guicassolato, line 15: We could use here https://github.com/3scale/porta/blob/master/app/lib/system/error_reporting.rb#L7

test/unit/invitation_test.rb
```diff
@@ -44,41 +44,22 @@ def setup
...
-    # This is needed, database will not save fractions of seconds
-    # so something like:
-    # Time.now.to_f #=> 1512012308.293877
-    # will be saved in DB as 1512012308
-    Timecop.freeze(Time.zone.parse('2017-11-23 03:25:08 UTC +00:00')) do
-      invitation = @provider.invitations.create!(:email => 'buddy@example.net')
-      assert_equal Time.zone.now, invitation.sent_at
-    end
+  test 'sent_at field is nil when created' do
+    invitation = @provider.invitations.create!(:email => 'buddy@example.net')
```
guicassolato, line 53:
```suggestion
    invitation = @provider.invitations.create!(email: 'buddy@example.net')
```

test/workers/send_user_invitation_worker_test.rb
```diff
@@ -0,0 +1,82 @@
...
+
+class SendUserInvitationWorkerTest < ActiveJob::TestCase
+  def test_provider_mailer_perform
+    Timecop.freeze(Time.zone.parse('2017-11-23 03:25:08 UTC +00:00')) do
+      ProviderInvitationMailer.any_instance.expects(:invitation).once
+
+      provider_account = FactoryBot.create(:provider_account)
+      invitation = FactoryBot.create(:invitation, account: provider_account)
+
+      assert_nil invitation.sent_at
```
guicassolato, line 13:
```suggestion
      refute invitation.sent_at
```

test/workers/send_user_invitation_worker_test.rb
```diff
@@ -0,0 +1,82 @@
...
+      SendUserInvitationWorker.new.perform(invitation.id)
+      assert invitation.reload.sent_at, Time.zone.now
```
guicassolato, line 15:
```suggestion
      assert_equal invitation.reload.sent_at, Time.zone.now
```

test/workers/send_user_invitation_worker_test.rb
```diff
@@ -0,0 +1,82 @@
...
+
+    invitation = FactoryBot.create(:invitation)
+
+    assert_nil invitation.sent_at
+    SendUserInvitationWorker.new.perform(invitation.id)
+    assert_nil invitation.reload.sent_at
+  end
+
+  def test_handles_not_found_error
+    ProviderInvitationMailer.any_instance.expects(:invitation).raises(ActiveRecord::RecordNotFound)
```
guicassolato, line 63: I think these tests could be just one. You can loop through the exceptions classes and repeat the same block. Just make sure to include another argument to the assertion where you pass the name of the exception class that failed, so we have useful message in case the test starts failing.

test/workers/send_user_invitation_worker_test.rb
```diff
@@ -0,0 +1,82 @@
...
+  end
+
+  # regression test for: https://github.com/3scale/system/pull/3316
+  def test_send_invitation_with_helper_tag
+    provider = FactoryBot.create(:simple_provider)
+    buyer = FactoryBot.create(:simple_buyer, provider_account: provider)
+    FactoryBot.create(:cms_email_template, system_name: 'invitation', provider: provider, published: '{% debug:help %}', rails_view_path: 'emails/invitation')
+    invitation = FactoryBot.create(:invitation)
+
+    ProviderInvitationMailer.any_instance.expects(:invitation)
```
guicassolato, line 79: It doesn't mean the email will be rendered properly though. Not sure if it belongs to this test anyway.

app/workers/send_user_invitation_worker.rb
Martouta: That notifies to bugsnag, and does not retry.

What we should have is:

- ActiveRecord::RecordNotFound => Do not notify to bugsnag, and do not retry
- OpenSSL::SSL::SSLError, Net::SMTPError, SocketError => Do not notify to bugsnag because it is not a bug in our side, and retry the job.

test/workers/send_user_invitation_worker_test.rb
Martouta: assert equal for times/dates is a bit asking for problems... even inside the timecop. I would suggest:
```suggestion
      assert_in_delta Time.zone.now, invitation.reload.sent_at, 1.second
```

app/workers/send_user_invitation_worker.rb
guicassolato: Agreed.

app/workers/send_user_invitation_worker.rb
josemigallas: But it is 😳

test/workers/send_user_invitation_worker_test.rb
josemigallas: That isn't what the test is about, I think. It's asserting the mail is sent https://github.com/3scale/porta/pull/2092/files#diff-6667f7edb63c733c01adf25ebaef955aL154

app/workers/send_user_invitation_worker.rb
guicassolato: Why? What happens when you remove this line? I bet `test/workers/send_user_invitation_worker_test.rb` passes entirely 😉

test/workers/send_user_invitation_worker_test.rb
guicassolato: The test ensures that `ProviderInvitationMailer#invitation` is called, but keep in mind that stubbing that method with an expectation will make it return, in this case, `nil`, rather than actually _executing_ that method. Therefore, that CMS template created a couple lines before in the test won't ever get to the point of being rendered, thus making the test case pointless.

In order to test the rendering of the CMS template, with liquid tags being rendered and all, you'll need to get deeper than that, i.e. actually invoking `ProviderInvitationMailer#invitation` (not a stub), running it where it calls `ActionMailer::Base#mail` at https://github.com/3scale/porta/blob/512e40e5528ed217f84d373fc98e05a4c164e9a4/app/mailers/provider_invitation_mailer.rb#L13, and getting the email message rendered. Anyway, what I've just described does not belong to this test file because it's not an attribution of the worker; it's either an integration test or part of mailer's test.

On the other hand, if all you wanted to do was testing that the worker invokes `ProviderInvitationMailer#invitation`, then rename the test and remove the CMS template because it's not needed.

app/workers/send_user_invitation_worker.rb
josemigallas: It throws something like _SendUserInvitationWorker::Net::SMTPError is not defined_.

test/workers/send_user_invitation_worker_test.rb
josemigallas: But it's supposed to be exactly `nil`, not _false_. Do you still prefer `refute`?

test/workers/send_user_invitation_worker_test.rb
josemigallas: OK I didn't know it was being stubbed.

app/workers/send_user_invitation_worker.rb
josemigallas: In order to retry, the exception has to be risen again, right? And if rescued, the job will be discarded.

app/workers/send_user_invitation_worker.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  def perform(invitation_id)
+    invitation = Invitation.find(invitation_id)
+
+    mailer = invitation.account.provider? ? ProviderInvitationMailer : InvitationMailer
+    mailer.invitation(invitation).deliver_now!
+
+    invitation.update!(sent_at: Time.zone.now)
+  rescue *RETRY_ERRORS => error
+    logger.error(error.message)
```
Martouta, line 29: you should add [retry_job](https://api.rubyonrails.org/v4.2.0/classes/ActiveJob/Enqueuing.html#method-i-retry_job).

app/workers/send_user_invitation_worker.rb
```diff
@@ -0,0 +1,34 @@
...
+  rescue *DISCARD_ERRORS => error
+    logger.error(error.message)
+    false
```
Martouta, line 32: Rescuing here is ok for me, but better doing `rescue_from` IMO.
Example:
https://github.com/3scale/porta/blob/765e5f013bc30bb0b152294fce81e737660de6ad/app/workers/delete_plain_object_worker.rb#L19-L22

app/workers/send_user_invitation_worker.rb
```diff
@@ -0,0 +1,34 @@
...
+    Net::SMTPUnsupportedCommand,
+    SocketError,
+  ].freeze
+
+  DISCARD_ERRORS = [
+    ActiveRecord::RecordNotFound
+  ].freeze
+
+  def perform(invitation_id)
+    invitation = Invitation.find(invitation_id)
```
Martouta, line 22: This is ok, but IMO it is better enqueuing the `invitation` object directly instead of the ID, so internally it does it by [GlobalID](https://edgeguides.rubyonrails.org/active_job_basics.html#globalid).
And in the rescue error, instead of `ActiveRecord::RecordNotFound`, it is `ActiveJob::DeserializationError` is the object is not in the DB anymore.

app/workers/send_user_invitation_worker.rb
```diff
@@ -0,0 +1,34 @@
...
+
+    mailer = invitation.account.provider? ? ProviderInvitationMailer : InvitationMailer
+    mailer.invitation(invitation).deliver_now!
+
+    invitation.update!(sent_at: Time.zone.now)
```
Martouta, line 27: The `update!` _should_ never fail, but if it does (for example, if the DB is down for a while), then it will retry the whole job, and it will send the email again.

test/unit/invitation_test.rb
```diff
@@ -44,41 +44,22 @@ def setup
     assert_not_nil invitation.token
   end
 
-  test 'sends invitation email when created' do
-    ProviderInvitationMailer.any_instance.expects(:invitation)
+  test 'creates a worker when created' do
```
Martouta, line 47: minor suggestion 😄 
```suggestion
  test 'enqueues a worker when created' do
```

app/workers/send_user_invitation_worker.rb
josemigallas: I'll add `ActiveRecord::ActiveRecordError` to the _discard errors_.

Martouta: Then it is better just doing `invitation.update(sent_at: Time.zone.now)`. As: if it works, good, if it doesn't... bad luck 🤷‍♀️ 
Most of the time that should be ok 🤞 

The more reliable solutions are a bit more complicated, like using sidekiq batches and doing `on_success`... but IMO, for this, it is not worth it.

app/workers/send_user_invitation_worker.rb
josemigallas: I've been trying to use `rescue_from` but can't use the splat operator.. I'll stick with `rescue` for now

app/workers/send_user_invitation_worker.rb
josemigallas: I don't have a strong opinion about it 🤷‍♂️

app/models/invitation.rb
```diff
@@ -39,6 +39,7 @@ def accept!
 
   def resend
     notify_invitee unless accepted?
+    update!(sent_at: nil)
   end
```
Martouta, line 43: I propose this order:
```suggestion
  def resend
    return if accepted?
    update!(sent_at: nil)
    notify_invitee
  end
```

Mainly because `notify_invitee` updates the `sent_at` attribute in the background at the end of it, and we can not promise that the update of this method will be executed before the other one 😉

josemigallas: Yep...

app/workers/send_user_invitation_worker.rb
Martouta: It is the same but with less code because GlobalID would do the (de)serialization for us...

Doing `SendUserInvitationWorker.perform_later(self)` inside the invitation, and here just doing `def perform(invitation)` does not mean saving the entire object in Redis and then fetching it all here... it internally saves the object class and the ID only, and to deserialize, it does the `invitation = Invitation.find(invitation_id)` internally.

Please read this:

- [ActiveJob GlobalID](https://edgeguides.rubyonrails.org/active_job_basics.html#globalid)
- [Usage of GlobalID](https://github.com/rails/globalid#usage) in general. So you can see a bit of what the suggested code would do internally.

app/workers/send_user_invitation_worker.rb
```diff
@@ -0,0 +1,26 @@
...
+    Net::SMTPAuthenticationError,
+    Net::SMTPFatalError,
+    Net::SMTPServerBusy,
+    Net::SMTPSyntaxError,
+    Net::SMTPUnknownError,
+    Net::SMTPUnsupportedCommand,
+    SocketError
+  ].freeze
+
+  def perform(invitation)
```
Martouta, line 17: catch the `ActiveJob::DeserializationError`

https://github.com/3scale/porta/blob/ff0b72315fcd58397aeac04aa254ad6c66376f52/app/workers/sphinx_indexation_worker.rb#L11-L13

app/workers/send_user_invitation_worker.rb
```diff
@@ -14,12 +14,18 @@ class SendUserInvitationWorker < ApplicationJob
...
+  ].freeze
+
   def perform(invitation)
     mailer = invitation.account.provider? ? ProviderInvitationMailer : InvitationMailer
     mailer.invitation(invitation).deliver_now!
 
     invitation.update(sent_at: Time.zone.now)
-  rescue *ERRORS => e
+  rescue *DISCARD_ERRORS => ex
+    Rails.logger.info "SphinxIndexationWorker#perform raised" #{ex.class} with message #{ex.message}"
```
Martouta, line 27:
```suggestion
    logger.error "SendUserInvitationWorker#perform raised #{ex.class} with message #{ex.message}"
```

thomasmaas:
Needs some commit cleanup, no?

thomasmaas:
I'm not what the new behavior is. The Jira issue looks for a way for the user to be sure it was actually sent. I'm not getting that from the screenshot. Could you show with video/multiple screenshots @josemigallas

Martouta:
@thomasmaas because before we were updating the `sent_at` attribute whether the invitation was sent for real or not... now only if it is actually sent. Although I agree that the screenshot should be updated because now when it is not sent, we do not see a `-` anymore but a more clear message.

josemigallas:
> I'm not what the new behavior is. The Jira issue looks for a way for the user to be sure it was actually sent. I'm not getting that from the screenshot. Could you show with video/multiple screenshots @josemigallas

Unfortunately there are no "real time" updates, the user needs to refresh the page in order to check whether the email has been sent or not.

The _way for the user to be sure_ is that `sent_at` will be updated **only** after the email has been sent, otherwise it will show `not yet sent`.

We could show an indicative that there was an error but that would involve DB changes and I would recommend to do in a follow-up PR.

thomasmaas:
Ah, didn't see the 'not sent yet' in the table, now i get it.

PR #2093, opened 2020-07-31
Running on different port

hallelujah:
Normally, we should forbid other routes on this port but it is messing up the file
Changes:

- prometheus port driven by PROMETHEUS_EXPORTER_PORT, defaults to 9394
- prometheus path /metrics
- no more basic auth

### Tenant modes

The promtheus exporter port for unicorn has default value depending on `TENANT_MODE` environment

*  `multitenant` => 9394
* `master` => 9394
* `provider` => 9395
* `developer` => 9396

### Notes

Closes https://issues.redhat.com/browse/THREESCALE-5741

Needs https://github.com/3scale/deploy/pull/679
Closes https://github.com/3scale/platform/issues/256

## Notes

As discussed https://github.com/3scale/3scale-operator/issues/453

web server port could be 9394
sidekiq port could be 9359

So when you launch sidekiq, just do `PROMETHEUS_EXPORTER_PORT=9359 rake sidekiq`


thomasmaas:
Does this affect customers who want to use prometheus?

hallelujah:
@thomasmaas Yes, we change the endpoint to a more standard one.
If there are any documentation, I think we should ask @3scale/documentation

raelga:
Looks good to me, but I'm not an expert in Ruby code.

Agree with @thomasmaas on the documentation update and establishing a deprecation path if required, as some users may be using the current behaviour.

hallelujah:
No customer is using it as it is not documented at all.

config/routes.rb
```diff
@@ -13,15 +13,11 @@ def call(env)
 
 Rails.application.routes.draw do
 
-  constraints MasterDomainConstraint do
+  constraints PortConstraint.new(ENV.fetch('PROMETHEUS_EXPORTER_PORT', 9090).to_i) do
```
guicassolato, line 16: Won't this conflict with https://github.com/3scale/porta/blob/f46f350ed0e2bb5589f6aec577f5fb79e9f91d50/config/initializers/sidekiq.rb#L36-L38?

hallelujah: No because sidekiq and unicorn do not run in the same host.
We could change `SIDEKIQ_PROMETHEUS_EXPORTER_PORT` if you think it is necessary.

Anyway we want also to normalize all prometheus exporter.
Only one per process using default port 9090

This is not achievable for now as sidekiq-prometheus-exporter gem overrides the content of previous rack application.
So we cannot chain exporters https://github.com/Strech/sidekiq-prometheus-exporter/blob/15879ca30830a94ed254a443448859c84fb4e88c/lib/sidekiq/prometheus/exporter.rb#L47-L51

Follow https://issues.redhat.com/browse/THREESCALE-2468 for more information

fbolton:
> No customer is using it as it is not documented at all.

@hallelujah So, presumably there are no doc requirements for this issue?

andrewdavidmackenzie:
@fbolton I think that once we (now) add Prometheus metrics from System (unclear to me how many there are already....) then this *should* be documented for customers, anything special regarding ports used as part of the setup, and something on the metrics exposed in a similar fashion to that done in customer docs for apicast/backend etc... 

@hallelujah can we merge this and move on? Would go into 2.10 I assume (NOT cherry-picked to 2.9.1) and so that would be your release target for docs Fintan.

fbolton:
@andrewdavidmackenzie Ok, I just created https://issues.redhat.com/browse/THREESCALE-5825 to track the related doc requirement.

hallelujah:
@fbolton Yes definitively we need to document it.
Once we have all the metrics defined and the ports decided, we will deliver a document to the documentation team as well as a walkthrough.

hallelujah:
@andrewdavidmackenzie Putting it back at WIP again, there are some issues popping up

Martouta (approved):

config/routes.rb
```diff
@@ -11,17 +11,31 @@ def call(env)
   end
 end
 
+PROMETHEUS_EXPORTER_DEFAULT_PORTS = {
```
Martouta, line 14: https://github.com/3scale/3scale-operator/pull/458#issuecomment-686561158 👍

lib/prometheus_exporter_port.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  TENANT_MODE_CHECK = ->(mode) do
+    PROMETHEUS_EXPORTER_DEFAULT_PORTS.key?(mode.to_s) ? mode.to_s : raise(InvalidTenantModeError)
+  end
+
+  def self.call
+    mode = TENANT_MODE_CHECK.call(tenant_mode)
+    default_port = PROMETHEUS_EXPORTER_DEFAULT_PORTS[mode]
+
+    ENV.fetch('PROMETHEUS_EXPORTER_PORT', default_port).to_i
```
guicassolato, line 23: Let's remember this approach in case http://github.com/3scale/3scale-operators ask us if empty value is OK or the env var must be omitted to activate default value.

guicassolato (approved):

lib/prometheus_exporter_port.rb
hallelujah: it should be omitted otherwise it would be port `0` ...

PR #2097, opened 2020-08-03
[UX specs] Update Edit product page specs

cntlsn:

Add inline help UX specs for system name input field

### DIFF
--- a/portafly/doc/ux_specs/product/edit_product.md
+++ b/portafly/doc/ux_specs/product/edit_product.md
@@ -14,6 +14,10 @@
       * Name already exists on form submit
       * Name already exists on _onblur_
   * System name > text input [disabled]
+    * Inline help:
+      * Label include a `pf-icon-help` icon
+      * Icon toggles a popover ([PF specs](https://www.patternfly.org/v4/documentation/react/components/popover))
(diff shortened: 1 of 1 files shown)


portafly/doc/ux_specs/product/edit_product.md
```diff
@@ -14,6 +14,10 @@
       * Name already exists on form submit
       * Name already exists on _onblur_
   * System name > text input [disabled]
+    * Inline help:
```
thomasmaas, line 17: And the idea is to do this on each form where there is a system name?

cntlsn: @thomasmaas yes, it can probably just be the same message with an altered subject "Product", "Backend", "Application plan".

cntlsn: @thomasmaas  I was thinking about offering a solution to that, meaning telling users that if they really need to change the system name they should then delete and re-create the Product / Backend / Application plan. Basically just to avoid users contacting support only for that. But that should actually include a series of cautionary measures, so maybe we can skip for now.

thomasmaas: Seems more like something for docs than for in-ui

PR #2099, opened 2020-08-04
Provider::Admin::Account::Users#index => all users with td[1] display_name link

Martouta:

### DIFF
--- a/app/decorators/paginating_decorator.rb
+++ b/app/decorators/paginating_decorator.rb
@@ -1,0 +1,6 @@
+# frozen_string_literal: true
+
+class PaginatingDecorator < Draper::CollectionDecorator
+  # support for will_paginate
+  delegate :current_page, :total_entries, :total_pages, :per_page, :offset, :limit_value, :model_name, :total_count
+end
--- a/app/controllers/provider/admin/account/users_controller.rb
+++ b/app/controllers/provider/admin/account/users_controller.rb
@@ -31,7 +31,7 @@ class Provider::Admin::Account::UsersController < Provider::Admin::Account::Base
   end
 
   def collection
-    @users ||= end_of_association_chain.but_impersonation_admin.paginate(:page => params[:page])
+    @users ||= end_of_association_chain.but_impersonation_admin.paginate(page: params[:page]).decorate
   end
 
   def update_resource(user, attributes)
--- a/app/decorators/application_decorator.rb
+++ b/app/decorators/application_decorator.rb
@@ -9,6 +9,10 @@ class ApplicationDecorator < Draper::Decorator
     raise NoMethodError, __method__
   end
 
+  def self.collection_decorator_class
+    PaginatingDecorator
+  end
+
   private
 
   API_KEYS = %w[id name system_name].freeze
--- a/app/views/provider/admin/account/users/index.html.slim
+++ b/app/views/provider/admin/account/users/index.html.slim
@@ -23,11 +23,11 @@ table class="data" id="users"
         td
           - if can? :manage, user
             - if current_user == user
-              = link_to user.decorate.display_name,
+              = link_to user.display_name,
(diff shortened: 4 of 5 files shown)


app/decorators/paginating_decorator.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+class PaginatingDecorator < Draper::CollectionDecorator
+  # support for will_paginate
+  delegate :current_page, :total_entries, :total_pages, :per_page, :offset
```
Martouta, line 5: https://github.com/drapergem/draper/issues/429#issuecomment-12572503

guicassolato: Can [this problem](https://github.com/drapergem/draper/issues/429#issuecomment-30467042) also affect us? I see we use `page_entries_info` at https://github.com/3scale/porta/blob/ad6a20c26bd4e53e68b46a4a9edafac6d801f87f/app/views/forums/public/_detail.html.erb#L22

Martouta: Not sure but I can add it, just in case 👍

Gemfile.base
```diff
@@ -69,6 +69,7 @@ gem 'audited'
 gem 'acts_as_list', '~> 0.9.17'
 gem 'braintree', '~> 2.89'
 gem 'cancancan', '~> 2.3.0'
+gem 'draper-cancancan'
```
Martouta, line 72: Because otherwise, any `can?` with the object decorated does not work as we wish because the decorated object is not the same as the model object.

app/decorators/paginating_decorator.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+class PaginatingDecorator < Draper::CollectionDecorator
+  # support for will_paginate
+  delegate :current_page, :total_entries, :total_pages, :per_page, :offset, :limit_value, :model_name, :total_count
```
Martouta, line 5: https://github.com/3scale/porta/pull/2099#discussion_r464959448

guicassolato: Or we can add the method ourselves without the gem. It's only one method that needs to be patched (`CanCan::Ability#can?`) after all, and whose implementation can be improved BTW.

Gemfile.base
guicassolato: Or we can add the method ourselves without the gem. It's only one method that needs to be patched (CanCan::Ability#can?) after all, and whose implementation can be improved BTW.

Martouta: Fair enough 👍 💪

app/models/ability.rb
```diff
@@ -29,6 +29,11 @@ def self.define(&block)
     @@rules << block
   end
 
+  def can?(action, subject, *extra_args)
+    model = subject.try(:model) || subject
```
guicassolato, line 33: I think it actually needs to be in a loop until it's no longer a decorated object, no?

Martouta: @guicassolato I do not think so, because 'model' should really mean 'model' 🤔 look at this:
image
image
image
image

guicassolato: https://github.com/GoodMeasuresLLC/draper-cancancan/pull/1

Martouta: It's ironic because apparently it is currently working in that scenario (it is a collection in this case). The question is 'why' xD
I will check it to ensure that it is stable, and I may add a test.

Martouta: @guicassolato you are right about this bug.

The reason why 'it is apparently working although it is a collection' is that it is checking the `can?` for each entry of the collection. 

The solution tho is not the loop of `subject.model` 🤔 that's not working 🤔 
I am on it 👍

Martouta: Fixed and extracted to its own PR https://github.com/3scale/porta/pull/2103

PR #2100, opened 2020-08-04
Fix Invitation send message with Observer

josemigallas:
### Alternative for #2092

Invitation delivery is notified as "successfully sent" when it really isn't.
1. Change the message to "it will be sent"
2. Update the column `sent_at` only after the email has been sent by the Mailer

[THREESCALE-4360: Using "Invite New User" button, no verification that email is sent](https://issues.redhat.com/browse/THREESCALE-4360)

**Verification steps** 

1. Go to `p/admin/account/invitations` and create a new invitation.
2. The notification should say it _will_ be sent
3. Refresh and check the column `Sent` shows the correct value (perhaps blanks at first, email should be sent not inmediately)
4. Hitting _Resend_ also should update the _Sent at_ column values.

### DIFF
--- a/app/helpers/invitations_helper.rb
+++ b/app/helpers/invitations_helper.rb
@@ -1,6 +1,6 @@
 module InvitationsHelper
   def invitation_sent_date(invitation)
-    (invitation.sent_at.presence || invitation.created_at).to_s(:long)
+    invitation.sent_at || '-'
   end
 
   def invitation_status(invitation)
--- a/app/lib/three_scale/invitation_email_delivery_observer.rb
+++ b/app/lib/three_scale/invitation_email_delivery_observer.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+class ThreeScale::InvitationEmailDeliveryObserver
+  def self.delivered_email(message)
+    binding.pry
+    user = User.find_by(email: message.from)
+
+    invitations = user ? user.account.invitations : Invitation
+    invitations.find_by!(email: message.to)
+               .update!(sent_at: Time.zone.now)
+  end
+end
--- a/config/initializers/email_delivery_observer.rb
+++ b/config/initializers/email_delivery_observer.rb
@@ -1,0 +1,4 @@
+# frozen_string_literal: true
+
+ProviderInvitationMailer.register_observer ThreeScale::InvitationEmailDeliveryObserver
+InvitationMailer.register_observer ThreeScale::InvitationEmailDeliveryObserver
--- a/test/unit/three_scale/invitation_email_delivery_observer_test.rb
+++ b/test/unit/three_scale/invitation_email_delivery_observer_test.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class ThreeScale::InvitationEmailDeliveryObserverTest < ActiveSupport::TestCase
+  # include ActionMailer::TestHelper
+
+  def setup
+    # InvitationMailer.register_observer ThreeScale::InvitationEmailDeliveryObserver
+  end
+
+  def test_update_date_after_sent_only
+    # assert_no_emails do
+      invitation = FactoryBot.create(:invitation)
+      mailer = nil
+
+      date = Date.parse('2019-09-01')
+      Timecop.freeze(date) do
+      #   perform_enqueued_jobs do
+      #     mailer     = InvitationMailer.invitation(@invitation)
+          assert_nil invitation.sent_at
+      #   end
+      end
+    # end
+
+    # assert_emails 0 do
+    #   invitation.expects(:update_column).once
+
+    #   Invitation.expects(find_by).once.return(invitation)
+
+    #   assert_nil invitation.sent_at
+    # end
+
+    # assert_emails 1 do
+    #   @mailer.deliver_now
+    # end
+  end
+
+end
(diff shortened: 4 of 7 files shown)


app/helpers/invitations_helper.rb
```diff
@@ -1,6 +1,6 @@
 module InvitationsHelper
   def invitation_sent_date(invitation)
-    (invitation.sent_at.presence || invitation.created_at).to_s(:long)
+    invitation.sent_at || '-'
```
Martouta, line 3:
```suggestion
    invitation.sent_at&..to_s(:long)
```

app/lib/three_scale/invitation_email_delivery_observer.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class ThreeScale::InvitationEmailDeliveryObserver
+  def self.delivered_email(message)
+    binding.pry
```
Martouta, line 5: 👀 😝 
image

app/lib/three_scale/invitation_email_delivery_observer.rb
```diff
@@ -0,0 +1,9 @@
...
+    Invitation.find_by(email: message.to)
```
Martouta, line 6: Can an invitation to the same email be done for different accounts? If so, you should search by the account as well.

app/lib/three_scale/invitation_email_delivery_observer.rb
```diff
@@ -0,0 +1,9 @@
...
+              .update_column(:sent_at, Time.zone.now)
```
Martouta, line 7: If this works, this is better because it does not skip validations:
```suggestion
              .update!(sent_at: Time.zone.now)
```

app/helpers/invitations_helper.rb
Martouta: Or:
```suggestion
    invitation.sent_at&.to_s(:long) || '-'
```

app/lib/three_scale/invitation_email_delivery_observer.rb
Martouta, line 6: btw, use `find` instead of `find_by` here, because if it does not exist, we have a bug that we need to fix.

test/unit/three_scale/invitation_email_delivery_observer_test.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class ThreeScale::InvitationEmailDeliveryObserverTest < ActiveSupport::TestCase
+  include ActionMailer::TestHelper
+
+  def setup
+    InvitationMailer.register_observer ThreeScale::InvitationEmailDeliveryObserver
```
Martouta, line 9: isn't this done in the initializer?

app/lib/three_scale/invitation_email_delivery_observer.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class ThreeScale::InvitationEmailDeliveryObserver
```
Martouta, line 3: I think that @hallelujah used to tell to Cadu to not put things called from the initializers in `app/lib` and instead put them in `lib`. But @hallelujah should be the person saying where this class should go 😄

config/initializers/email_delivery_observer.rb
```diff
@@ -0,0 +1,4 @@
+# frozen_string_literal: true
+
+ProviderInvitationMailer.register_observer ThreeScale::InvitationEmailDeliveryObserver
+InvitationMailer.register_observer ThreeScale::InvitationEmailDeliveryObserver
```
Martouta, line 4: DRY suggestion 👇 
```suggestion
[ProviderInvitationMailer, InvitationMailer].each { |mailer_klass| mailer_klass.register_observer(ThreeScale::InvitationEmailDeliveryObserver) }
```

test/unit/three_scale/invitation_email_delivery_observer_test.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+
+  def test_update_date_after_sent_only
+    @mailer = nil
+
+    assert_emails 0 do
+      invitation = FactoryBot.create(:invitation)
+      invitation.expects(:update_column).once
+
+      Invitation.expects(find_by).once.return(invitation)
```
Martouta, line 19: imo, it is better without mocking this... it does not matter how it does it internally (calling to which methods), what matters is the final result, and this is what it should test... so imo you should use `Timecop` and then test the final date of the attribute
Example:
https://github.com/3scale/porta/blob/ad6a20c26bd4e53e68b46a4a9edafac6d801f87f/test/events/zync_event_test.rb#L64-L70

```ruby
Timecop.freeze(date) do
  ### here whatever u want to execute being the current date/time "date"
end
```

test/unit/three_scale/invitation_email_delivery_observer_test.rb
```diff
@@ -0,0 +1,30 @@
...
+
+class ThreeScale::InvitationEmailDeliveryObserverTest < ActiveSupport::TestCase
+  include ActionMailer::TestHelper
+
+  def setup
+    InvitationMailer.register_observer ThreeScale::InvitationEmailDeliveryObserver
+  end
+
+  def test_update_date_after_sent_only
+    @mailer = nil
```
Martouta, line 13: why instance variables if these are apparently used only inside this test method?

app/lib/three_scale/invitation_email_delivery_observer.rb
josemigallas: You mean 2 accounts inviting the same email? Makes sense.

Martouta: yup 😄

PR #2101, opened 2020-08-04
Fixes wizard's test request

guicassolato:
Tests requests in the Admin Portal onboarding wizard are currently broken for 2 reasons:

1. It does not take into account the API backend usage routing paths and keeps sending the request to `<sandbox_endpoint>/<api_test_path>`.
2. Even when the test request works, if the response contains characters that are not ideally escaped by `CGI.escape`, or if it's too long, the next screen of the wizard breaks with a 502 or 414 error message (respectively).

This PR addresses these two issues. It ensures a proper APIAP-compatible URL is built for the test request, including the chosen routing path and API test path (1st issue). It also encodes and truncates the test API response before carrying it along to the next step of the wizard, thus preventing from a long or poorly encoded response body breaks the redirect (2nd issue).

Screenshot 2020-08-06 at 12 38 13

**References about the limit of characters to carry on in the query string of an HTTP request (1024*10 bytes), as imposed by Unicorn and Thin:**
- Unicorn: https://yhbt.net/unicorn.git/tree/ext/unicorn_http/global_variables.h?h=v5.4.1#n68
- Thin: https://github.com/macournoyer/thin/blob/f83ab677f008afcfabcdfa8930225f6cea88f461/ext/thin_parser/thin.c#L71

Web browsers and server may support more than that: https://stackoverflow.com/questions/812925/what-is-the-maximum-possible-length-of-a-query-string. However, the request will break once it hits the Rails server anyway.

----

Closes [THREESCALE-5353](https://issues.redhat.com/browse/THREESCALE-5353)

### DIFF
--- a/app/forms/onboarding/request_form.rb
+++ b/app/forms/onboarding/request_form.rb
@@ -39,7 +39,9 @@ class Onboarding::RequestForm < Reform::Form
 
   def uri
     return unless model.api_backend
-    URI.join(api_base_url, path || SLASH).to_s
+    base_url = api_base_url
+    test_path = File.join(base_url.try(:path).to_s, path.to_s)
+    URI.join(base_url, test_path).to_s
   end
 
   def proxy_auth_params
--- a/app/controllers/provider/admin/onboarding/wizard/request_controller.rb
+++ b/app/controllers/provider/admin/onboarding/wizard/request_controller.rb
@@ -31,7 +31,7 @@ class  Provider::Admin::Onboarding::Wizard::RequestController < Provider::Admin:
                     uri: @request.uri, real_api: real_api)
 
     if success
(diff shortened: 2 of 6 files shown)


app/forms/onboarding/request_form.rb
```diff
@@ -39,7 +39,9 @@ def proxy_base_url
 
   def uri
     return unless model.api_backend
-    URI.join(api_base_url, path || SLASH).to_s
+    base_url_path = api_base_url.respond_to?(:path) ? api_base_url.path : nil
```
Martouta, line 42: Isn't it exactly the same as this? 👇 
```suggestion
    base_url_path = api_base_url.try(:path)
```

PR #2104, opened 2020-08-05
Move '.decorate' Views->Controller for 'admin_buyers_user*' paths

Martouta:

### DIFF
--- a/app/controllers/buyers/users_controller.rb
+++ b/app/controllers/buyers/users_controller.rb
@@ -2,29 +2,45 @@ class Buyers::UsersController < Buyers::BaseController
   activate_menu :audience, :accounts, :listing
 
   inherit_resources
-  actions :index, :show, :edit, :update, :destroy
+  actions :index, :show, :edit
   defaults :route_prefix => 'admin_buyers' #FIXME: inherited_resource makes us repeat this
   belongs_to :account, :collection_name => :buyer_accounts
 
   def update
     # TODO: I think this controller is used only on provider side
-    resource.validate_fields! if current_user.account.buyer?
-    resource.attributes = params[:user]
-    super
+    resource.validate_fields! if current_account.buyer?
+    resource.attributes = params[:user] # TODO: we should filter the params
+
+    if update_resource(resource, resource_params)
+      flash[:notice] = 'User was successfully updated.'
+      redirect_to action: :show
+    else
+      render :edit
+    end
+  end
+
+  def destroy
+    if resource.destroy
+      flash[:notice] = 'User was successfully deleted.'
+      redirect_to action: :index
+    else
+      flash[:error] = 'User could not be deleted.'
+      redirect_back_or_show_detail
+    end
   end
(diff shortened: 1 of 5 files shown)


app/controllers/buyers/users_controller.rb
```diff
@@ -2,29 +2,45 @@ class Buyers::UsersController < Buyers::BaseController
...
 
   def update
     # TODO: I think this controller is used only on provider side
-    resource.validate_fields! if current_user.account.buyer?
-    resource.attributes = params[:user]
-    super
+    resource.validate_fields! if current_account.buyer?
+    resource.attributes = params[:user] # TODO: we should filter the params
+
+    if update_resource(resource, resource_params)
```
guicassolato, line 14: Can we do a normal `resource.update_attributes` and start getting rid of `InheritResources`?

app/controllers/buyers/users_controller.rb
```diff
@@ -2,29 +2,45 @@ class Buyers::UsersController < Buyers::BaseController
...
   defaults :route_prefix => 'admin_buyers' #FIXME: inherited_resource makes us repeat this
   belongs_to :account, :collection_name => :buyer_accounts
 
   def update
     # TODO: I think this controller is used only on provider side
-    resource.validate_fields! if current_user.account.buyer?
-    resource.attributes = params[:user]
-    super
+    resource.validate_fields! if current_account.buyer?
+    resource.attributes = params[:user] # TODO: we should filter the params
```
guicassolato, line 12: Why do we need to set the attributes here? Is this because of extra fields somehow? It does not look safe anyway.

app/controllers/buyers/users_controller.rb
Martouta: @guicassolato Unfortunately, that wouldn't work for updating the `role`, even with keeping the previous `resource.attributes = params[:user]`

app/controllers/buyers/users_controller.rb
Martouta: > Why do we need to set the attributes here?

Absolutely no idea 🤷‍♀️ but in case of doubt, I prefer not touching and keeping exactly the same behavior.

> t does not look safe anyway.

It's not safe but it is not changing anything of that from before this PR and not introduced just now.
Looking at which params must be filtered can be a task on its own.

Martouta:
@guicassolato TL;DR : You are requesting for potentially breaking changes. That code nowadays is bad but that's what we had before this PR...
Fair enough for me: I change it as u say to have it more 'clean', but in another PR (because those are very potentially breaking changes). Because then, if it breaks, we can revert only that PR and without reverting all the changes done in this one.

guicassolato:
> Fair enough for me: I change it as u say to have it more 'clean', but in another PR (because those are very potentially breaking changes). Because then, if it breaks, we can revert only that PR and without reverting all the changes done in this one.

As you wish, @Martouta.

PR #2108, opened 2020-08-05
Decorate current_user and current_account before reaching the views.

Martouta:
Closes [THREESCALE-5812](https://issues.redhat.com/browse/THREESCALE-5812)

### Notes for the reviewer(s)
Review carefully, please. This can be risky 😄

### DIFF
--- a/app/lib/decorator_additions.rb
+++ b/app/lib/decorator_additions.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+module DecoratorAdditions
+
+  # This smells of :reek:NilCheck
+  def current_account
+    @_decorated_current_account ||= super&.decorate
+  end
+
+  # This smells of :reek:NilCheck
+  def current_user
+    @_decorated_current_user ||= super&.decorate
+  end
+
+  def current_ability
+    @current_ability ||= ::Ability.new(controller.send(:current_user))
+  end
+end
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 # Ability definitions are now split across multiple files, for better maintainability.
 #
 # To create a new set of permission rules, create a file in config/abilities,
@@ -10,13 +12,16 @@
 class Ability
   include CanCan::Ability
 
+  attr_reader :user
+
   def initialize(user)
     @user = user
     load_rules!
   end
 
+  # This smells of :reek:NilCheck
   def reload!
-    @user.try!(:reload)
+    user&.reload
     @rules = []
     @rules_index = nil
     load_rules!
@@ -42,7 +47,7 @@ class Ability
 
   def load_rules!
     @@rules.each do |rule|
-      self.instance_exec(@user, &rule)
+      instance_exec(user, &rule)
     end
   end
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -10,6 +10,8 @@ class ApplicationController < ActionController::Base
   include ThreeScale::OnPremises
   include ProxyConfigAffectingChanges::ControllerExtension
 
+  _helpers.module_eval { prepend DecoratorAdditions }
+
   before_action :set_newrelic_custom_params
 
   protect_from_forgery with: :reset_session # See ActionController::RequestForgeryProtection for details
--- a/app/controllers/provider/admin/user/base_controller.rb
+++ b/app/controllers/provider/admin/user/base_controller.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 class Provider::Admin::User::BaseController < Provider::Admin::BaseController
   activate_menu :account, :personal
 
@@ -8,6 +10,6 @@ class Provider::Admin::User::BaseController < Provider::Admin::BaseController
   protected
 
   def authorize_resource!
-    authorize! [:read, :update], current_user
+    authorize! %i[read update], current_user
   end
 end
--- a/app/views/provider/admin/accounts/suspended.html.slim
+++ b/app/views/provider/admin/accounts/suspended.html.slim
@@ -1,4 +1,4 @@
-h1 Hello #{current_user.decorate.display_name},
+h1 Hello #{current_user.display_name},
 p
   | Your 3scale trial has expired and the account is de-activated.
     To continue using the service, you’ll need to move onto a paid plan.
--- a/app/views/provider/admin/onboarding/wizard/info/intro.html.slim
+++ b/app/views/provider/admin/onboarding/wizard/info/intro.html.slim
@@ -1,5 +1,5 @@
 - content_for :title do
-  = "Hello #{current_user.decorate.informal_name},"
+  = "Hello #{current_user.informal_name},"
 
 p The best way to start discovering 3scale is through the APIcast staging gateway. You'll be up and running in minutes – ready to try all the different features of the platform.
 
--- a/app/views/shared/_service_access.slim
+++ b/app/views/shared/_service_access.slim
@@ -3,5 +3,5 @@ p
   => current_account.org_name
   ' account. Please
   - admin = current_account.admin_user
-  => mail_to admin.email, "contact #{current_account.decorate.admin_user_display_name}"
+  => mail_to admin.email, "contact #{current_account.admin_user_display_name}"
   | to request access.
--- a/app/views/shared/provider/_header.html.slim
+++ b/app/views/shared/provider/_header.html.slim
@@ -49,7 +49,7 @@ header.pf-c-page__header id="user_widget" class='Header' class=('Header--master'
               ' Signed in to
             => current_account.name
             ' as
-            = current_user.decorate.display_name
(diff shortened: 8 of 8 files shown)


app/lib/authenticated_system.rb
```diff
@@ -233,9 +232,9 @@ def send_remember_cookie!
       :expires => @current_user.remember_token_expires_at }
   end
 
-  # Account of current user, or nil if no one is logged in.
+  # This smells of :reek:NilCheck
   def current_account
-    @current_account || (current_user && current_user.account)
+    @current_account ||= current_user&.account&.decorate
```
guicassolato, line 237: We have a proper setter for the `@current_account` instance variable: https://github.com/3scale/porta/blob/1f077b3237382948e18b88a212978915903c2940/app/lib/authenticated_system.rb#L242-L244

Should we make this a setter as well?

app/models/ability.rb
```diff
@@ -11,12 +13,15 @@ class Ability
   include CanCan::Ability
 
   def initialize(user)
-    @user = user
+    @user = user.present? ? user.decorate.object : nil
```
guicassolato, line 16: This is weird. Why do we need to decorate this inside the ability definition?

Martouta: Because when we do `authorize! :whetever, Object`, or `can? :whetever, Object` it internally does for us something like this:
`Ability.new(current_user).can?(:whetever, Object)`
And the problem with that is that `current_user` is decorated and our permissions are for User model, not a UserDecorator object... so it will say that this user does not have permission although it does.

Martouta: If you want to see an example of what I am saying, do this:

1. Create `test/integration/provider/admin/user/personal_details_controller_test.rb` with this code:
```
# frozen_string_literal: true

require 'test_helper'

class Provider::Admin::User::PersonalDetailsControllerTest < ActionDispatch::IntegrationTest
  def setup
    provider = FactoryBot.create(:provider_account)
    @user = FactoryBot.create(:active_user, account: provider)
    login! provider, user: user
  end

  attr_reader :user

  NO_NOTIFICATIONS_REGEX = /You don't have access to any notifications/

  test 'No notification preferences available' do
    user.admin_sections = []
    user.save!(validate: false)

    get provider_admin_user_notification_preferences_path

    page = Nokogiri::HTML::Document.parse(response.body)
    assert_match NO_NOTIFICATIONS_REGEX, page.text
  end

  test 'Some notification preferences available' do
    user.admin_sections = ['partners']
    user.save!(validate: false)

    get provider_admin_user_notification_preferences_path

    page = Nokogiri::HTML::Document.parse(response.body)
    assert_not_match NO_NOTIFICATIONS_REGEX, page.text
  end
end
```

2. Run it, it will fail without this change.
In this case it is for the authorize in `app/controllers/provider/admin/user/base_controller.rb`.
But keep in mind that this is ONE individual scenario, but we also have plenty of `can? ...` inside views.

guicassolato: I see. It's actually to "undecorate" it. It doesn't look right anyway. `Ability` expects an `User` but what we pass may not be an `User` yet.

Martouta: > Why do we need to decorate this inside the ability definition?

Btw, we are not decorating here, but _undecorating_ instead. If it is a user model or its decorated object, it still saves the model in `@user` and not the decorated version.

app/lib/authenticated_system.rb
Martouta: I don't understand your proposal 😕

Right now we have the following situation:

- In the UI, we log in by current_user, never current_account. If we have the user, we have the account and we do not need to set it separately IMO. In fact, if we set it separately, we need to ensure that there can never be a mismatch.
- In the API, we log in by current_user (by access token) and by current_account (by provider key). The `current_account` setter is for the latter.
- This method is the reader, not the setter. And it should be 'always' the decorated version (well, we just need it that way right before reaching the views or representers)

app/models/ability.rb
Martouta: > I see. It's actually to "undecorate" it. It doesn't look right anyway. Ability expects a User but what we pass may not be a User yet.

But what do you propose? 😕

Martouta: The alternative version that I could think of is checking where we do `Ability.new` and THERE sending the current_user without being decorated... WDYT?

app/lib/authenticated_system.rb
guicassolato: No proposal actually. I'm just pointing out that this is changing the behaviour.

To agree with you, I see no reason for `current_user` to change along the request (or for the `account` of the current user to change), but current implementation allows that (weirdly or not, who knows?). Your change now makes it to memoize the account, turning this method from a simple getter into a setter. Therefore, a possible new behaviour.

In the end, there's no problem. If the tests are green and you say it's OK doing it, I'm OK with it too. It just seems important to me to highlight when a refactoring change can end up modifying the behaviour too.

app/lib/authenticated_system.rb
```diff
@@ -1,4 +1,8 @@
 module AuthenticatedSystem
+  def current_ability
```
Martouta, line 2: https://github.com/3scale/porta/pull/2108#discussion_r471570977

app/lib/authenticated_system.rb
```diff
@@ -1,4 +1,8 @@
...
+    @current_ability ||= Ability.new(current_user.present? ? current_user.decorate.object : nil)
```
Martouta, line 3: https://github.com/3scale/porta/pull/2108#discussion_r471570977

macejmic: i would go with this idea: https://github.com/3scale/porta/pull/2108#discussion_r471581718

this `current_user.decorate.object` does not look very smart.

guicassolato: `current_user.decorate` will decorate the model again if it's not a `Decorator` instance yet, right? Wdyt about:

```ruby
Ability.new(current_user.try(:model) || nil)
```

Martouta: it is `current_user.try(:object)`, but imo, it should be then: `Ability.new(current_user.try(:object) || current_user)`

Martouta: @guicassolato @macejmic changed 😎

app/lib/authenticated_system.rb
```diff
@@ -1,4 +1,8 @@
 module AuthenticatedSystem
+  def current_ability
+    @current_ability ||= Ability.new(current_user.try(:object) || current_user)
```
Martouta, line 3: https://github.com/3scale/porta/pull/2108#discussion_r471570977
https://github.com/3scale/porta/pull/2108#discussion_r472022736

app/lib/authenticated_system.rb
guicassolato: From https://github.com/drapergem/draper/blob/v3.0.1/README.md

> When writing decorator methods you'll usually need to access the wrapped model. While you may choose to use delegation (covered below) for convenience, you can always use the object (or its alias model)

😉

Martouta: @guicassolato I'm sorry but I do not understand how that quote helps us in this case.

Martouta: just to be clear, I've done `Ability.new(current_user.try(:object) || current_user)` because if it does not respond to `object`, that means that `current_user` is either the User model or `nil`, and both should do in our case.

app/lib/authenticated_system.rb
hallelujah: Maybe it is better to override current_account in a helper and use:

```ruby
module DecoratorHelper
  def current_account
     @_decorated_current_account ||= controller.current_account&.decorate
  end

  def current_user
     @_decorated_current_user ||= controller.current_user&.decorate
  end
```

app/helpers/decorator_helper.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module DecoratorHelper
+
+  # This smells of :reek:NilCheck
+  def current_account
+    @_decorated_current_account ||= controller.current_account&.decorate
```
Martouta, line 7: As @hallelujah mentioned, this could be:
```suggestion
    @_decorated_current_account ||= super&.decorate
```

Then I can make the other method be protected again 😄

app/helpers/decorator_helper.rb
```diff
@@ -0,0 +1,14 @@
...
+  end
+
+  # This smells of :reek:NilCheck
+  def current_user
+    @_decorated_current_user ||= controller.current_user&.decorate
```
Martouta, line 12: Same as https://github.com/3scale/porta/pull/2108/files#r472182050

guicassolato (approved):
Wow. Much cleaner.

app/lib/decorator_additions.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module DecoratorAdditions
```
hallelujah, line 3: That could go to concerns but OK :)

PR #2112, opened 2020-08-06
Get rid of InheritedResources for Buyers::UsersController

Martouta:
This request by Gui: https://github.com/3scale/porta/pull/2104#issuecomment-669966196

### DIFF
--- a/app/controllers/buyers/users_controller.rb
+++ b/app/controllers/buyers/users_controller.rb
@@ -1,17 +1,26 @@
+# frozen_string_literal: true
+
 class Buyers::UsersController < Buyers::BaseController
   activate_menu :audience, :accounts, :listing
 
-  inherit_resources
-  actions :index, :show, :edit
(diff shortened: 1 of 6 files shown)


app/controllers/buyers/users_controller.rb
```diff
@@ -1,17 +1,26 @@
...
   activate_menu :audience, :accounts, :listing
 
-  inherit_resources
-  actions :index, :show, :edit
-  defaults :route_prefix => 'admin_buyers' #FIXME: inherited_resource makes us repeat this
-  belongs_to :account, :collection_name => :buyer_accounts
+  before_action :find_account, except: :index
+  before_action :find_user, except: :index
+
+  actions :show, :edit
```
guicassolato, line 9: I think this is still inherited_resources, right?

Martouta: it could be 🤔 I will change it for `def show; end` and `def edit; end` then 😄

PR #2116, opened 2020-08-07
[CRUD] ApiDocs::BackendApi

Martouta:
https://issues.redhat.com/browse/THREESCALE-3160

### DIFF
--- a/app/controllers/admin/api/api_docs_services_controller.rb
+++ b/app/controllers/admin/api/api_docs_services_controller.rb
@@ -32,8 +32,7 @@ class Admin::Api::ApiDocsServicesController < Admin::Api::BaseController
   ##~ op.parameters.add @parameter_access_token
   #
   def index
-    @api_docs_services = api_docs_services.all
-    respond_with(@api_docs_services)
+    respond_with(api_docs)
   end
 
   # swagger
@@ -60,7 +59,7 @@ class Admin::Api::ApiDocsServicesController < Admin::Api::BaseController
   ##~ op.parameters.add :name => "skip_swagger_validations", :description => "Set to 'true' to skip validation of the Swagger specification, or 'false' to validate the spec. The default value is 'false'", :dataType => "boolean", :paramType => "query"
   #
   def create
-    @api_docs_service = current_account.api_docs_services.create(api_docs_params(:system_name), without_protection: true)
+    @api_docs_service = api_docs.create(api_docs_params(:system_name), without_protection: true)
     respond_with(@api_docs_service)
   end
 
@@ -136,12 +135,12 @@ class Admin::Api::ApiDocsServicesController < Admin::Api::BaseController
     params.require(:api_docs_service).permit(*permit_params)
   end
 
-  def api_docs_services
-    current_account.api_docs_services.accessible.permitted_for(current_user)
+  def api_docs
+    current_account.all_api_docs.service_accessible.permitted_for(current_user)
   end
 
   def find_api_docs_service
-    @api_docs_service = api_docs_services.find(params[:id])
+    @api_docs_service = api_docs.find(params[:id])
   end
 
   def new_service_id_permitted
--- a/app/controllers/admin/api_docs/account_api_docs_controller.rb
+++ b/app/controllers/admin/api_docs/account_api_docs_controller.rb
@@ -12,6 +12,10 @@ class Admin::ApiDocs::AccountApiDocsController < Admin::ApiDocs::BaseController
 
   before_action :redirect_to_service_scope, if: :service, only: %i[edit preview]
 
+  def index
+    @api_docs_services = api_docs_collection.page(params[:page])
+  end
+
   private
 
   def redirect_to_service_scope
@@ -26,11 +30,22 @@ class Admin::ApiDocs::AccountApiDocsController < Admin::ApiDocs::BaseController
     end
   end
 
-  def service
-    api_docs_service.service
-  end
-
   def current_scope
     current_account
+  end
+
+  def service
+    api_docs_service_owner = api_docs_service.owner
+    return if api_docs_service_owner.class.name != 'Service'
+
+    api_docs_service_owner
+  end
+
+  def find_api_docs
+    @api_docs_service = api_docs_collection.where(id: params[:id]).or(api_docs_collection.where(system_name: params[:id])).first!
+  end
+
+  def api_docs_collection
+    current_scope.all_api_docs.permitted_for(current_user)
   end
 end
--- a/app/controllers/admin/api_docs/base_controller.rb
+++ b/app/controllers/admin/api_docs/base_controller.rb
@@ -11,11 +11,13 @@ class Admin::ApiDocs::BaseController < FrontendController
   end
 
   def new
-    @api_docs_service = api_docs_services.new
+    @api_docs_service = current_scope.api_docs.new
   end
 
   def create
-    @api_docs_service = api_docs_services.new(api_docs_params(:system_name), without_protection: true)
+    @api_docs_service = api_docs_collection.new
+    @api_docs_service.assign_attributes(api_docs_params(:system_name), without_protection: true)
+
     if @api_docs_service.save
       redirect_to(preview_admin_api_docs_service_path(@api_docs_service), notice: 'ActiveDocs Spec was successfully saved.')
     else
@@ -84,13 +86,25 @@ class Admin::ApiDocs::BaseController < FrontendController
 
   attr_reader :api_docs_service
 
+  def api_docs_collection
+    raise NoMethodError, "#{__method__} not implemented in #{self.class}"
+  end
+
   def current_scope
     raise NoMethodError, "#{__method__} not implemented in #{self.class}"
   end
 
   def api_docs_params(*extra_params)
     permit_params = %i[name body description published skip_swagger_validations service_id] + extra_params
-    params.require(:api_docs_service).permit(*permit_params)
+    permitted_params = params.require(:api_docs_service).permit(*permit_params)
+
+    service_id = permitted_params[:service_id]
+    if service_id.present?
+      permitted_params[:owner_id] = service_id
+      permitted_params[:owner_type] = 'Service'
+    end
+
+    permitted_params
   end
 
   def api_docs_services
--- a/app/controllers/admin/api_docs/service_api_docs_controller.rb
+++ b/app/controllers/admin/api_docs/service_api_docs_controller.rb
@@ -6,9 +6,21 @@ class Admin::ApiDocs::ServiceApiDocsController < Admin::ApiDocs::BaseController
   activate_menu :serviceadmin, :ActiveDocs
   sublayout 'api/service'
 
+  def index
+    @api_docs_services = api_docs_collection.page(params[:page])
+  end
+
   private
+
+  def find_api_docs
+    @api_docs_service = current_scope.account.all_api_docs.find_by_id_or_system_name!(params[:id])
+  end
 
   def current_scope
     @service
+  end
+
+  def api_docs_collection
+    current_scope.api_docs.service_accessible.permitted_for(current_user)
   end
 end
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -56,7 +56,7 @@ class Account < ApplicationRecord
   self.background_deletion = [
     :users,
     :mail_dispatch_rules,
-    [:api_docs_services, { class_name: 'ApiDocs::Service' }],
+    [:all_api_docs, { class_name: 'ApiDocs::Service' }],
     :services,
     :contracts,
     :account_plans,
@@ -160,8 +160,11 @@ class Account < ApplicationRecord
   has_many :hidden_messages, -> { latest_first.received.hidden }, as: :receiver, class_name: 'MessageRecipient'
   has_many :received_messages, -> { latest_first.received.visible }, as: :receiver, class_name: 'MessageRecipient'
 
-  has_many :api_docs_services, class_name: 'ApiDocs::Service', dependent: :destroy
+  has_many :api_docs, as: :owner, class_name: 'ApiDocs::Service', dependent: :destroy, inverse_of: :owner
+  has_many :all_api_docs, class_name: 'ApiDocs::Service', dependent: :destroy
   has_many :log_entries, foreign_key: 'provider_id'
+
+  alias api_docs_services api_docs
 
   has_many :events, class_name: EventStore::Event, foreign_key: :provider_id, inverse_of: :account
   has_many :access_tokens, through: :users
--- a/app/models/api_docs/service.rb
+++ b/app/models/api_docs/service.rb
@@ -1,4 +1,5 @@
 # encoding: utf-8
+
 class ApiDocs::Service < ApplicationRecord
   CURRENT_SWAGGER_VERSION = 2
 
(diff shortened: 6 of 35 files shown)


app/models/api_docs/backend_api.rb
```diff
@@ -0,0 +1,89 @@
+# frozen_string_literal: true
+
+class ApiDocs::BackendApi < ApplicationRecord
```
guicassolato, line 3: Why not using the same model by adding a polymorphic owner, like we did for `Metric` and `ProxyRule`?

app/views/admin/api_docs/base/index.html.slim
```diff
@@ -9,7 +9,7 @@
 - if @api_docs_services.any?
   = render partial: 'active_docs_table', locals: {service: @service, api_docs_services: @api_docs_services}
 - else
-  - if current_account.api_docs_services.without_service.any? && can?(:manage, :portal)
+  - if current_account.api_docs_services.without_owner.any? && can?(:manage, :portal)
```
guicassolato, line 12: I think this is contradictory. `api_docs_services` is now defined as with `service_id` present, while `api_docs_services` looks for the opposite.

guicassolato: Also, let's keep in mind that sometimes we want all ApiDocs owned directly by the `Account`, i.e. NOT owned by a `Service`, nor a `BackendApi`; and sometimes we want ALL ApiDocs, i.e. the ones owned by the `Account` directly, by the `Service`'s and `BackendApi`'s of the account.

Perhaps `Account` could have `has_many :api_docs` and `has_many :all_api_docs` maybe? (Or `:api_docs_owned_by_self` and `:api_docs` 🤔

guicassolato (commented):
That's the direction we should go, IMHO.

Just a few more things so we don't forget (and that I can think of now 🙂):
1. This change certainly will require a lot of testing (both in OpenShift and preview) and doc update (/cc @3scale/documentation @3scale/qe)
2. A next iteration IMO would be concentrating everything related to what owns the `ApiDoc::Service` in the new polymorphic `owner` association – i.e. get rid of `service_id` and the convention used to infer it's owned by the account directly. For this, we will need a data migration to back fill `owner_id` and `owner_type` in all records.
3. Need to run the migration for other DBMSs of course

mkudlej:
@martouta @macejmic What is this change about? There are no description or associated Jira, please add them. Thank you!

app/controllers/admin/api/api_docs_backend_apis_controller.rb
```diff
@@ -0,0 +1,123 @@
+# frozen_string_literal: true
+
+class Admin::Api::ApiDocsBackendApisController < Admin::Api::BaseController
```
guicassolato, line 3: Maybe put this with the other `BackendApis`-namespaced controllers, inheriting from `app/controllers/admin/api/backend_apis/base_controller.rb`

app/controllers/admin/api/api_docs_services_controller.rb
```diff
@@ -133,15 +132,23 @@ def destroy
 
   def api_docs_params(*extra_params)
     permit_params = %i[name body description published skip_swagger_validations service_id] + extra_params
-    params.require(:api_docs_service).permit(*permit_params)
+    permitted_params = params.require(:api_docs_service).permit(*permit_params)
+
+    service_id = permitted_params[:service_id]
+    if service_id.present?
+      permitted_params[:owner_id] = service_id
```
guicassolato, line 139: I think this part could be handled in the model, e.g.
https://github.com/3scale/porta/blob/1792c848d20564505f89ee214ba560a5c84fb362/app/models/metric.rb#L10
https://github.com/3scale/porta/blob/1792c848d20564505f89ee214ba560a5c84fb362/app/models/metric.rb#L244-L252

app/controllers/admin/api_docs/account_api_docs_controller.rb
```diff
@@ -27,10 +31,18 @@ def redirect_to_service_scope
   end
 
   def service
-    api_docs_service.service
+    api_docs_service.owner
```
guicassolato, line 34: I guess this is true only if the `ApiDocs::Service` is owned by a `Service`, right? Maybe keep
```suggestion
    api_docs_service.service
```
...and solve in the model? e.g.
```ruby
class ApiDocs::Service < ApplicationRecord
  def owner_by_service?
    owner_type == 'Service'
  end

  def service
    owner if owner_by_service?
  end
end
```

app/controllers/admin/api_docs/base_controller.rb
```diff
@@ -84,13 +86,25 @@ def destroy
...
   end
 
   def api_docs_params(*extra_params)
     permit_params = %i[name body description published skip_swagger_validations service_id] + extra_params
-    params.require(:api_docs_service).permit(*permit_params)
+    permitted_params = params.require(:api_docs_service).permit(*permit_params)
+
+    service_id = permitted_params[:service_id]
+    if service_id.present?
+      permitted_params[:owner_id] = service_id
```
guicassolato, line 103: Same as https://github.com/3scale/porta/pull/2116/files?file-filters%5B%5D=.rb#r489444922

app/controllers/admin/api_docs/service_api_docs_controller.rb
```diff
@@ -6,9 +6,21 @@ class Admin::ApiDocs::ServiceApiDocsController < Admin::ApiDocs::BaseController
   activate_menu :serviceadmin, :ActiveDocs
   sublayout 'api/service'
 
+  def index
+    @api_docs_services = current_scope.all_api_docs.page(params[:page])
```
guicassolato, line 10:
```suggestion
    @api_docs_services = current_scope.api_docs.page(params[:page])
```

Will `Service` respond to `all_api_docs` too?

app/models/account.rb
```diff
@@ -160,9 +160,12 @@ def schedule_backend_sync_worker
   has_many :hidden_messages, -> { latest_first.received.hidden }, as: :receiver, class_name: 'MessageRecipient'
   has_many :received_messages, -> { latest_first.received.visible }, as: :receiver, class_name: 'MessageRecipient'
 
-  has_many :api_docs_services, class_name: 'ApiDocs::Service', dependent: :destroy
+  has_many :api_docs, as: :owner, class_name: 'ApiDocs::Service', dependent: :destroy, inverse_of: :owner
```
guicassolato, line 163: ❤️ Loved this renaming from `api_docs_services` to `api_docs`. Thanks!

app/models/api_docs/service.rb
```diff
@@ -1,32 +1,38 @@
...
   validates :name, :system_name, :base_path, :swagger_version, length: { maximum: 255 }
+  validates :owner_type, length: { maximum: 255, allow_blank: true }
   validates :description, length: { maximum: 65535 }
   validates :body, length: { maximum: 4294967295, allow_blank: true }
-  validate :service_belongs_to_account, if: -> { service_id.present? && service_id_changed? }
+  validate :service_belongs_to_account, if: -> { service_id.present? || owner_type == 'Service' }
 
-  before_validation(on: :create) { self.account ||= service&.account }
+  before_validation(on: :create) { self.account ||= owner_account }
+  before_validation(on: :save) { self.owner ||= account }
```
guicassolato, line 29: This will mandate a migration to pre-fill `owner_type` and `owner_id` of all service-owned api_docs before we have any process running this version of the code. Perhaps we can make this a bit smoother... Well, maybe not needed. Related to https://github.com/3scale/porta/pull/2116/files?file-filters%5B%5D=.rb#r489444922.

app/models/api_docs/service.rb
```diff
@@ -38,6 +44,34 @@ class ApiDocs::Service < ApplicationRecord
 
   validates :base_path, non_localhost: true
 
+  def owner_account
+    return unless owner
+
+    owner_type == 'Account' ? owner : owner.account
+  end
```
guicassolato, line 51: It would be so nice if `Account#account` returned `self`:

```ruby
  delegate :account, to: :owner, prefix: true, allow_nil: true
```

...but it does NOT 😞.

app/models/api_docs/service.rb
```diff
@@ -38,6 +44,34 @@ class ApiDocs::Service < ApplicationRecord
...
+
+  # [Deprecated][TODO] remove with service_id
+  def service=(service_to_assign)
+    self.owner = self.owner_type = self.owner_id = nil
+    super(service_to_assign)
+  end
```
guicassolato, line 57: This won't be needed if we do https://github.com/3scale/porta/pull/2116/files?file-filters%5B%5D=.rb#r489444922

guicassolato: ... and also https://github.com/3scale/porta/pull/2116/files?file-filters%5B%5D=.rb#r489448808

app/models/api_docs/service.rb
```diff
@@ -38,6 +44,34 @@ class ApiDocs::Service < ApplicationRecord
...
+
+  # [Deprecated][TODO] remove with service_id
+  def service_id=(service_id_to_assign)
+    self.owner = self.owner_type = self.owner_id = nil
+    super(service_id_to_assign)
+  end
```
guicassolato, line 63: Same as https://github.com/3scale/porta/pull/2116/files?file-filters%5B%5D=.rb#r489461266

app/models/api_docs/service.rb
```diff
@@ -103,14 +141,15 @@ def specification
   end
 
   def api_product_production_public_base_url
-    return unless service&.proxy
-    service.proxy.endpoint
+    return if !owner.respond_to?(:proxy) || owner.proxy.blank?
```
guicassolato, line 144:
```suggestion
    return unless owner.try(:proxy)
```

app/models/api_docs/service.rb
```diff
@@ -103,14 +141,15 @@ def specification
...
+
+    owner.proxy.endpoint
   end
 
   private
 
   def service_belongs_to_account
-    return true if account.services.accessible.where(id: service_id).exists?
+    return true if account.services.accessible.where(id: service_id.presence || owner_id).exists?
```
guicassolato, line 152: _Danger!_ If the `ApiDocs::Service` belongs to a `BackendApi`, this will cause a service trying to be found by the ID of the backend api.

app/models/api_docs/service.rb
```diff
@@ -1,32 +1,38 @@
...
+
 class ApiDocs::Service < ApplicationRecord
   CURRENT_SWAGGER_VERSION = 2
 
   include SystemName
   extend System::Database::Scopes::IdOrSystemName
 
   belongs_to :account, required: true
-  belongs_to :service, class_name: '::Service', inverse_of: :api_docs_services
+  belongs_to :service, class_name: '::Service', inverse_of: :deprecated_api_docs, required: false
```
guicassolato, line 10: If we go with https://github.com/3scale/porta/pull/2116/files?file-filters%5B%5D=.rb#r489448808, maybe then we can remove this, and also https://github.com/3scale/porta/blob/e8a21d46d8cd8e54f3c63f2a492d869e9e107861/app/models/service.rb#L23 and https://github.com/3scale/porta/blob/e8a21d46d8cd8e54f3c63f2a492d869e9e107861/app/models/service.rb#L61.

app/models/service.rb
```diff
@@ -19,7 +19,8 @@ class Service < ApplicationRecord
   self.background_deletion = [
     :service_plans,
     :application_plans,
-    [:api_docs_services, class_name: 'ApiDocs::Service'],
+    [:api_docs, class_name: 'ApiDocs::Service'],
+    [:deprecated_api_docs, class_name: 'ApiDocs::Service'],
```
guicassolato, line 23: See https://github.com/3scale/porta/pull/2116/files?file-filters%5B%5D=.rb#r489467271

app/models/service.rb
```diff
@@ -56,7 +57,13 @@ class Service < ApplicationRecord
   with_options(dependent: :destroy, inverse_of: :service) do |service|
     service.has_many :service_plans, as: :issuer, &DefaultPlanProxy
     service.has_many :application_plans, as: :issuer, &DefaultPlanProxy
-    service.has_many :api_docs_services, class_name: 'ApiDocs::Service'
+  end
+  has_many :deprecated_api_docs, class_name: 'ApiDocs::Service', dependent: :destroy, inverse_of: :service
```
guicassolato, line 61: See https://github.com/3scale/porta/pull/2116/files?file-filters%5B%5D=.rb#r489467271

app/views/admin/api_docs/base/_form.html.slim
```diff
@@ -9,8 +9,7 @@
     = form.input :system_name, input_html: { :disabled => true }
   = form.input :published, label: 'Publish?'
   = form.input :description, input_html: {rows: 3}
-  - unless @api_docs_service.new_record? && @api_docs_service.service_id.present?
-    = form.input :service_id, as: :select, collection: current_user.accessible_services
+  = form.input :service_id, as: :select, collection: current_user.accessible_services
```
guicassolato, line 12: I think the conditional here was to ensure that, when you click on "New ActiveDoc" from the index of ActiveDocs of a service (therefore within the service section), we don't give the user the option to select a different service. I think this makes sense and we need to keep hiding the select box in this case.

features/services/active_docs.feature
```diff
@@ -18,7 +18,5 @@ Feature: ActiveDocs pages
     Then the api doc spec is saved with this service linked
 
   Scenario: Create with failures and retry
-     And the service selector is not in the form
```
guicassolato, line 21: See https://github.com/3scale/porta/pull/2116/files?file-filters%5B%5D=.feature&file-filters%5B%5D=.rb&file-filters%5B%5D=.slim#r489472184

lib/developer_portal/lib/liquid/drops/provider.rb
```diff
@@ -215,7 +215,16 @@ def logo_url
         </ul>
       )
       def api_specs
-        Drops::Collection.for_drop(Drops::ApiSpec).new(@model.api_docs_services.accessible)
+        service_api_specs = @model.all_api_docs.accessible(::Service)
+        account_api_specs = @model.all_api_docs.accessible(::Account)
+        backend_api_api_specs = @model.all_api_docs.accessible(::BackendApi)
+        accessible_api_specs = ApiDocs::Service.where(id: service_api_specs | account_api_specs | backend_api_api_specs)
```
guicassolato, line 221: I think the logic for this should remain inside the scope. Better if we can trust `ApiDocs::Service.accessible` returns api_docs whose owners are not deleted/marked for deletion.

test/factories/api_docs/service.rb
```diff
@@ -12,5 +13,13 @@
         paths: {}
       }.to_json
     end
+
+    before(:validation) do |api_docs_service|
+      account = FactoryBot.create(:simple_provider)
+
+      api_docs_service.account = account
+      api_docs_service.owner = account
```
guicassolato, line 21: I assume this won't overwrite if you pass on `FactoryBot.build(:api_docs_service, owner: object)`, right?

app/controllers/admin/api/api_docs_backend_apis_controller.rb
macejmic: the thing is that it authorize a different action, according the existing: `app/controllers/admin/api/api_docs_services_controller.rb` and it also does not load the backend API object so it would inherit from that controller but wouldn't use it for anything.

test/factories/api_docs/service.rb
macejmic: actually, it does.

config/routes.rb
```diff
@@ -605,6 +605,7 @@ def call(env)
       end
 
       resources :active_docs, :controller => 'api_docs_services', except: %i[new edit]
+      resources :backend_api_active_docs, controller: 'api_docs_backend_apis', only: [:index, :update, :create, :destroy]
```
guicassolato, line 608: I was wondering... can we have this as an actual resource of backend_apis, i.e. `/admin/api/backend_apis/{backend_api_id}/api_docs`?

app/controllers/admin/api/api_docs_backend_apis_controller.rb
```diff
@@ -0,0 +1,128 @@
...
+  def api_docs
+    current_account.all_api_docs.owned_by_backend_api
+  end
+
+  def api_docs_params(*extra_params)
+    permit_params = %i[name body description published skip_swagger_validations backend_api_id] + extra_params
+    permitted_params = params.require(:api_docs_backend_api).permit(*permit_params)
+
+    backend_api_id = permitted_params.delete(:backend_api_id)
+    if backend_api_id.present?
```
guicassolato, line 117: I think the `backend_api_id` should not be optional here. Related to https://github.com/3scale/porta/pull/2116/files?file-filters%5B%5D=.rb#r504510783

Martouta:
I opened this PR by request of @macejmic ages ago... closing now

PR #2117, opened 2020-08-07
[UX specs] Backend / Overview page

deernano:

Create ux spec for Backend > Overview page

Jira : https://issues.redhat.com/browse/APPDUX-341
Marvel : https://marvelapp.com/prototype/12844cg4/screen/71559225

**Verification steps** 

We will not having delete modal at the moment.


portafly/doc/ux_specs/backend/create_backend.md
```diff
@@ -0,0 +1,44 @@
+# Create Backend
+
+[Back to index](../index.md)
+
+* [Mockups](https://marvelapp.com/prototype/12844cg4)
+* [Mockups JIRA ticket](https://issues.redhat.com/browse/APPDUX-343)
+* [Parent JIRA ticket](https://issues.redhat.com/browse/THREESCALE-5575)
+
+##### Title
+* Create a new backend
```
dfennessy, line 10: According to the IBM style guide, `backend` should either be written as `back end` (noun) or `back-end` (adjective), but if there is a more specific term like "server" for example, then it should be replaced.

cntlsn: @dfennessy This is a good point, thank you! We're actually discussing about removing the term altogether. I would leave it as it is for now just for consistency with the current term. It will be edited as soon as we make a decision.

portafly/doc/ux_specs/backend/create_backend.md
```diff
@@ -0,0 +1,44 @@
+# Create Backend
```
cntlsn, line 1: @deernano The PR is about the Backend Overview page, so this file shouldn't be part of it. To avoid conflicts it would be better to delete this file from the PR.

When creating a new branch for writing the UX specs for a new page, it would be better to always create the new branch from `master`. In this way, you will only submit the new changes and not all the previous ones too :)

portafly/doc/ux_specs/backend/edit_backend.md
```diff
@@ -0,0 +1,34 @@
+# Edit Backend
```
cntlsn, line 1: @deernano same as above. it would be better do delete this file from the PR.

portafly/doc/ux_specs/backend/overview.md
```diff
@@ -0,0 +1,17 @@
...
+
+* [Mockups](https://marvelapp.com/prototype/12844cg4/screen/71559225)
+* [Mockups JIRA ticket](https://issues.redhat.com/browse/APPDUX-341)
+* [Parent JIRA ticket](https://issues.redhat.com/browse/THREESCALE-5573)
+
+##### Page header
+* Edit button
+  * Secondary button
+* Delete button
+  * Customized button (A secondary button to resemble a destructive button -- temporary solution until PF Secondary destructive button is released)
```
cntlsn, line 13: @deernano we will not have a Delete button for now, so let's comment that out.
```suggestion
<!-- * Delete button
  * Customized button (A secondary button to resemble a destructive button -- temporary solution until PF Secondary destructive button is released) -->
```

portafly/doc/ux_specs/backend/create_backend.md
deernano: I c, I ll delete it

PR #2118, opened 2020-08-10
[CVE-2020-8165] Patch: Unintended unmarshalling in ActiveSupport

guicassolato:
Monkey patches Rails prior to v5.2.4.3 with the [recommended changes](https://groups.google.com/group/rubyonrails-security/attach/1c0cea804792a/5-2-cache-storage.patch?part=0.1) of [CVE-2020-8165](https://groups.google.com/g/rubyonrails-security/c/bv6fW4S0Y1c).

Closes [THREESCALE-5318](https://issues.redhat.com/browse/THREESCALE-5318)

### DIFF
--- a/config/initializers/cve-2020-8165.rb
+++ b/config/initializers/cve-2020-8165.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+# [CVE-2020-8165] https://groups.google.com/g/rubyonrails-security/c/bv6fW4S0Y1c
+# Can be removed after upgrade to Rails v5.2.4.3+
+
+ActiveSupport::Cache::MemCacheStore.prepend(Module.new do
+  protected
+
+  # Reimplements from `ActiveSupport::Cache::Strategy::LocalCache#read_entry` instead of `ActiveSupport::Cache::MemCacheStore::LocalCacheWithRaw#read_entry`
+  def read_entry(key, options)
+    if cache = local_cache
+      cache.fetch_entry(key) { super }
+    else
+      super
+    end
+  end
+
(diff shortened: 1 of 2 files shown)


config/initializers/cve-2020-8165.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+# [CVE-2020-8165] https://groups.google.com/g/rubyonrails-security/c/bv6fW4S0Y1c
+# Can be removed after upgrade to Rails v5.2.4.3+
+
+ActiveSupport::Cache::MemCacheStore.prepend(Module.new do
```
hallelujah, line 6: Should we also patch Redis Cache?

guicassolato: I thought about that, but in the end I don't think we need to do it. `RedisCacheStore` was introduced in 5.2, so it has no meaning in 5.1, just like in 5.0. So as long as we make sure to go straight to 5.2.4.3 when we migrate from 5.1 to 5.2, we're good.

config/initializers/cve-2020-8165.rb
```diff
@@ -0,0 +1,23 @@
...
+  protected
+
+  # Reimplements from `ActiveSupport::Cache::Strategy::LocalCache#read_entry` instead of `ActiveSupport::Cache::MemCacheStore::LocalCacheWithRaw#read_entry`
+  def read_entry(key, options)
+    if cache = local_cache
+      cache.fetch_entry(key) { super }
+    else
+      super
+    end
+  end
```
hallelujah, line 16: Could this have a test?

guicassolato: Done

PR #2122, opened 2020-08-12
Strong params for stats data controllers

guicassolato:
Adds Rails strong params validation to stats/data usage and usage_response_code actions.
It affects essentially analytics requests for services (API products), API backends and applications, both UI and API.
Makes it consistent with the 3scale api docs where the `since` parameter is a required param.

_Missing:_ support for the use case of the `range` param (undocumented):
https://github.com/3scale/porta/blob/ff0b72315fcd58397aeac04aa254ad6c66376f52/app/lib/stats/views/usage.rb#L148

---

Closes [THREESCALE-5798](https://issues.redhat.com/browse/THREESCALE-5798).

### DIFF
--- a/test/integration/stats/data/requests_to_api_test.rb
+++ b/test/integration/stats/data/requests_to_api_test.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 require 'test_helper'
 
 class Stats::Data::RequestsToApiTest < ActionDispatch::IntegrationTest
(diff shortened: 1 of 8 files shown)


test/integration/stats/data/requests_to_api_test.rb
```diff
@@ -19,7 +21,7 @@ def setup
   test 'usage with access token' do
     member = FactoryBot.create(:member, account: @provider_account, admin_sections: ['monitoring'])
     token  = FactoryBot.create(:access_token, owner: member, scopes: ['stats'])
-    params = { period: 'day', metric_name: 'hits', access_token: token.value }
+    params = { period: 'day', since: Time.utc(2020, 8, 1).to_date, metric_name: 'hits', access_token: token.value }
```
guicassolato, line 24: We used to have tests that suggest an expected 200 response for API requests missing the `since` param, even though this is documented as a required param 🤔: https://github.com/3scale/porta/blob/ff0b72315fcd58397aeac04aa254ad6c66376f52/app/controllers/stats/data/base_controller.rb#L31

Martouta: don't worry, it was probably unintentional at the time OR it just changed over time 🤷‍♀️ the point is that we need it 👍

guicassolato: I know why already. I'll fix it.

guicassolato: It's because of the escape condition at https://github.com/3scale/porta/blob/ff0b72315fcd58397aeac04aa254ad6c66376f52/app/lib/stats/views/usage.rb#L135

PR #2123, opened 2020-08-12
Access yabeda metrics in unicorn

hallelujah:
Have another endpoint for yabeda-sidekiq /yabeda-metrics

Also use prometheus mmap client to share metrics between processes

Use environment variable `prometheus_multiproc_dir` 

Requires https://github.com/3scale/deploy/pull/679

### DIFF
--- a/lib/prometheus_exporter_port.rb
+++ b/lib/prometheus_exporter_port.rb
@@ -4,9 +4,9 @@ class PrometheusExporterPort
 
   PROMETHEUS_EXPORTER_DEFAULT_PORTS = {
     'multitenant' => 9394,
-    'master' => 9394,
-    'provider' => 9395,
-    'developer' => 9396
+    'master' => 9395,
+    'provider' => 9396,
+    'developer' => 9394
   }.freeze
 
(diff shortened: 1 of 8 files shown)


Martouta (approved):

Martouta (approved):

lib/prometheus_exporter_port.rb
```diff
@@ -24,6 +27,6 @@ def self.call
   end
 
   def self.tenant_mode
-    ENV.fetch('TENANT_MODE', 'multitenant')
+    ENV['TENANT_MODE'].to_s.empty? ? 'multitenant' : ENV['TENANT_MODE']
```
Martouta, line 30:
```suggestion
    ENV['TENANT_MODE'].presence || 'multitenant'
```

guicassolato (approved):
Tested in Openshift. Nice work 🥇

lib/prometheus_exporter_port.rb
hallelujah: We can't, it is used before activesupport is loaded

guicassolato (approved):

PR #2124, opened 2020-08-13
Fixes trashed provider messages restore

guicassolato:
Closes [THREESCALE-1503](https://issues.redhat.com/browse/THREESCALE-1503).


guicassolato:
> Makes me wonder why it was like it was.

The index used to show `hidden_messages`, which returns a collection of `MessageRecipient`. 5 years ago it changed to list `Message`s with `Account#trashed_messages`: https://github.com/3scale/system/commit/9b096387ee70026c08bf0780912e591771a9ca3d

PR #2130, opened 2020-08-18
Not displaying suspend button when tenant is deleted

damianpm:

When you first delete tenant and after want you try to suspend it 404 page is displayed and tenant cannot be suspended.
Note: Login as master

https://issues.redhat.com/browse/THREESCALE-5738

**Verification steps** 

1. Login as master
2. Create new tenant
3. Go to /buyers/accounts/{tenant_id}/edit and delete tenant
4. Go to /buyers/accounts/{tenant_id} and try to suspend tenant

**Before:**
before

**After:**
after

### DIFF
--- a/app/models/account/buyer_methods.rb
+++ b/app/models/account/buyer_methods.rb
@@ -161,8 +161,4 @@ module Account::BuyerMethods
   def not_paying_monthly!
     settings.update_attribute(:monthly_charging_enabled, false)
   end
-
-  def account_suspended_or_scheduled_for_deletion?(account)
-    account.suspended? || account.scheduled_for_deletion?
-  end
 end
--- a/app/models/account/buyer_methods.rb
+++ b/app/models/account/buyer_methods.rb
@@ -161,4 +161,8 @@ module Account::BuyerMethods
   def not_paying_monthly!
     settings.update_attribute(:monthly_charging_enabled, false)
   end
+
+  def account_suspended_or_scheduled_for_deletion?(account)
+    account.suspended? || account.scheduled_for_deletion?
+  end
 end
--- a/app/models/account/states.rb
+++ b/app/models/account/states.rb
@@ -79,7 +79,7 @@ module Account::States
       end
 
       event :suspend do
-        transition all => :suspended, if: :can_be_suspended?
+        transition all => :suspended, if: :ready_to_be_suspended?
       end
 
       event :resume do
@@ -138,11 +138,15 @@ module Account::States
       scheduled_for_deletion? || (buyer? && provider_account.try(:should_be_deleted?))
     end
 
-    def can_be_suspended?
+    def ready_to_be_suspended?
       tenant? || marked_for_suspension
     end
 
     attr_accessor :marked_for_suspension
+
+    def suspended_or_scheduled_for_deletion?
+      suspended? || scheduled_for_deletion?
+    end
   end
 
   module ClassMethods
--- a/app/views/buyers/accounts/show.html.slim
+++ b/app/views/buyers/accounts/show.html.slim
@@ -62,13 +62,13 @@ h1 data-hook="account-show"
                                  reject_admin_buyers_account_path(@account),
                                  method: :post,
                                  class: 'reject action'
-            - if can? :suspend, @account
+            - if can?(:suspend, @account) && @account.can_suspend?
               = action_link_to 'Suspend',
-                               suspend_admin_buyers_account_path(@account),
-                               method: :post,
-                               data: { confirm: 'Are you sure?', disable_with: 'suspending…' },
-                               class: 'action suspend'
-            - if can? :resume, @account
+                                suspend_admin_buyers_account_path(@account),
+                                method: :post,
+                                data: { confirm: 'Are you sure?', disable_with: 'suspending…' },
+                                class: 'action suspend'
+            - if can?(:resume, @account) && @account.can_resume?
               = action_link_to 'Resume',
                                resume_admin_buyers_account_path(@account),
                                method: :post,
--- a/app/views/buyers/accounts/show.html.slim
+++ b/app/views/buyers/accounts/show.html.slim
@@ -62,7 +62,7 @@ h1 data-hook="account-show"
                                  reject_admin_buyers_account_path(@account),
                                  method: :post,
                                  class: 'reject action'
-            - if !@account.suspended? && !@account.scheduled_for_deletion?
+            - unless @account.account_suspended_or_scheduled_for_deletion?(@account)
               = action_link_to 'Suspend',
                                 suspend_admin_buyers_account_path(@account),
                                 method: :post,
--- a/app/views/buyers/accounts/show.html.slim
+++ b/app/views/buyers/accounts/show.html.slim
@@ -62,7 +62,7 @@ h1 data-hook="account-show"
                                  reject_admin_buyers_account_path(@account),
                                  method: :post,
                                  class: 'reject action'
-            - if can?(:suspend, @account) && @account.can_suspend?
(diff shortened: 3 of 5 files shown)


app/views/buyers/accounts/show.html.slim
```diff
@@ -63,11 +63,12 @@ h1 data-hook="account-show"
                                  method: :post,
                                  class: 'reject action'
             - if can? :suspend, @account
-              = action_link_to 'Suspend',
-                               suspend_admin_buyers_account_path(@account),
-                               method: :post,
-                               data: { confirm: 'Are you sure?', disable_with: 'suspending…' },
-                               class: 'action suspend'
+              - unless can? :resume, @account
```
Martouta, line 66: This is if the current_user has permission to ever `resume` the `@account`.
This is incorrect for 2 reasons:
1. What we want is, that if the `@account` has is `suspended` or scheduled_for_deletion`, do not show the Suspend link. Nothing to do with the current_user having permissions to change the state, but related to the state of this `@account` instead.
2. `resume` and `suspend` are 2 different actions (and they end in different states).

app/views/buyers/accounts/show.html.slim
```diff
@@ -63,11 +63,12 @@ h1 data-hook="account-show"
                                  method: :post,
                                  class: 'reject action'
             - if can? :suspend, @account
-              = action_link_to 'Suspend',
-                               suspend_admin_buyers_account_path(@account),
-                               method: :post,
-                               data: { confirm: 'Are you sure?', disable_with: 'suspending…' },
-                               class: 'action suspend'
+              - unless @account.suspended? || @account.scheduled_for_deletion?
```
Martouta, line 66: This is correct functionally but a bit hard to read. IMO it is easier:
```ruby
  - if !@account.suspended? && !@account.scheduled_for_deletion?
```

Yet, this still has one problem: If we ever change when it is possible to change to 'suspended' in the states machine (model), then that means that we have to remember to change it here. In fact, even now, there are more reasons why we do not allow to `suspend` and account. So, to do it dynamically and always according to what we write in the model, I suggest this:
```suggestion
              - if @account.can_suspend?
```

Martouta: Just to be clear, `@account.can_suspend?` means "If this object `@account` can be changed to state `suspended`"

Martouta: oh, and even better if we mix it with the condition above instead of inside it.
like: `- if can? :suspend, @account && @account.can_suspend?`

thomasmaas:
Does this also fix the scenario "don't show suspend button if suspended"?

thomasmaas (approved):

guicassolato (commented):
Should we have a test for this?

thomasmaas:
Yeah, probably we should. You were thinking of a cucumber @guicassolato ?

josemigallas (changes requested):
Preventively blocking merge since it's missing a test.

test/integration/buyers/accounts_controller_test.rb
```diff
@@ -316,6 +316,13 @@ def setup
       assert_xpath( './/div[@id="applications_widget"]//table[@class="list"]//tr', 2)
       refute_xpath( './/div[@id="applications_widget"]//table[@class="list"]//tr', /plan/i )
     end
+
+    test 'suspend button is not displayed when account is deleted or marked for deletion' do
+      ThreeScale.config.stubs(onpremises: false)
+      delete admin_buyers_account_path(@provider)
+      assert_select %(td a.button-to.action.suspend), false
```
guicassolato, line 323: I think this test could cover at least two other cases:
1. account NOT marked for deletion and NOT suspended
2. account NOT marked for deletion but already suspended

app/views/buyers/accounts/show.html.slim
```diff
@@ -62,13 +62,13 @@ h1 data-hook="account-show"
                                  reject_admin_buyers_account_path(@account),
                                  method: :post,
                                  class: 'reject action'
-            - if can? :suspend, @account
+            - if !@account.suspended? && !@account.scheduled_for_deletion?
```
guicassolato, line 65: I wonder is this business rule should be a method of the decorator (if it's only useful in the UI) or the model (since it seems to be useful [elsewhere](https://github.com/3scale/porta/blob/402c4f2af6d4ff8a03f5f946758e23cb6b65109e/app/workers/process_notification_event_worker.rb#L40) too).

test/integration/buyers/accounts_controller_test.rb
damianpm: Is there any method I can use to suspend or delete the account?
Something like `@provider.suspend`, `@provider.resume` ?
Or it's done by post request?

app/models/account/buyer_methods.rb
```diff
@@ -161,4 +161,8 @@ def paying_monthly!
   def not_paying_monthly!
     settings.update_attribute(:monthly_charging_enabled, false)
   end
+
+  def account_suspended_or_scheduled_for_deletion?(account)
```
damianpm, line 165: @guicassolato is this what you were thinking?

test/integration/buyers/accounts_controller_test.rb
damianpm: I updated the test, please take a look

app/models/account/buyer_methods.rb
guicassolato: Almost. I meant https://github.com/3scale/porta/pull/2130/commits/cfb4720fd2f6e3ff8d866d9776d6e9277f5ce540.

1. Since the module (`Account::BuyerMethods`, but also `Account::States`) is included in `Account`, the account is the context. No need to pass it as argument 😉
2. `suspended_or_scheduled_for_deletion?` has strictly to do with the state of the account, so I think better being part of `Account::States` than `Account::BuyerMethods`)

guicassolato (approved):

app/models/account/states.rb
```diff
@@ -138,11 +138,15 @@ def will_be_deleted?
       scheduled_for_deletion? || (buyer? && provider_account.try(:should_be_deleted?))
     end
 
-    def can_be_suspended?
+    def ready_to_be_suspended?
```
didierofrivia, line 141: Usually is better to use active instead of passive voice. Here `ready_to_suspend?` would be a better method name

app/views/buyers/accounts/show.html.slim
```diff
@@ -62,13 +62,13 @@ h1 data-hook="account-show"
                                  reject_admin_buyers_account_path(@account),
                                  method: :post,
                                  class: 'reject action'
-            - if can? :suspend, @account
+            - unless @account.suspended_or_scheduled_for_deletion?
```
Martouta, line 65: @damianpm @didierofrivia @guicassolato Why not `- if @account.can_suspend?` as I explained in https://github.com/3scale/porta/pull/2130#discussion_r472221448

image

image

guicassolato: Because it's not the same. `#can_suspend?` may be idiomatic but will return based on the existing constraints of the event machine, which now reduces to `#ready_to_be_suspended?` (wisely [pointed](https://github.com/3scale/porta/pull/2130#discussion_r478465369) by @didierofrivia that should be named `#ready_to_suspend?` instead, BTW).

If I understood correctly, we didn't want to add any constraint to the event machine, but just decide based on the current state.

Martouta: I disagree but I will explain properly (in-depth) on Tuesday. So today we can focus on learning as intended :wink:

PR #2135, opened 2020-08-20
Bumps pry-byebug to v3.9.0

guicassolato:
Fixes https://github.com/pry/pry/issues/2121


thomasmaas (approved):
It’s ok Gui, has been ages since I used it.

PR #2140, opened 2020-08-26
[Readme] Replacing domain link with 3scale.localhost

srijitamukherjee:

> The install.md file still refers to example.com, which is no longer in use. 
Hence replacing this with the new domain link.

> Closes [THREESCALE-5801](https://issues.redhat.com/browse/THREESCALE-5801)

**Verification steps** 
> Run Porta locally and use the new links to access portals.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,6 +1,6 @@
 # INSTALL
 
-Follow these instructions in order to set up a development environment, build and deploy this project on your machine.
+Follow these instructions to set up a development environment, build and deploy this project on your machine.
 
 ## Clone the repo, including submodules
 
@@ -28,11 +28,11 @@ Download and build all the images and start a shell session inside the container
 make bash
 ```
 
-All the source and dependencies for this project will be in place, making it possible to run porta and the tests from inside the container. See [Run Porta](#run-porta)
+All the sources and dependencies for this project will be in place, making it possible to run porta and the tests from inside the container. See [Run Porta](#run-porta)
 
 ### Running the application
 
-It's also possible to run the application by using only Docker. Firstly, setup the database by runnning `dev-setup` from your terminal:
+It's also possible to run the application by using only Docker. Firstly, set up the database by running `dev-setup` from your terminal:
 
 ```
 MASTER_PASSWORD=<master_password> USER_PASSWORD=<user_password> make dev-setup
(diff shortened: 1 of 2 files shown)


guicassolato (approved):

Martouta (approved):

Martouta:
@srijitamukherjee Please sign the commits (GPG). When you push again, no need to run the tests for Oracle and Postgres.

Martouta (approved):

INSTALL.md
```diff
@@ -302,23 +302,24 @@ yarn install
...
+You may need to set up the database from scratch again, in that case, use `db:reset` to drop it first too:
 
 ```bash
-bundle exec rake db:reset # This will drop and setup the database
+bundle exec rake db:reset # This will drop and set up the database
 ```
 
-**NOTE:** This will seed the application and creates the Master, Provider & Developer accounts which are accessible through: `http://master-account.example.com.lvh.me:3000`, `http://provider-admin.example.com.lvh.me:3000`, `http://provider.example.com.lvh.me:3000` respectively. Please take note of the credentials generated at this moment also so that you can log into each of these portals.
+**NOTE:** This will seed the application and creates the Master, Provider & Developer accounts that are accessible through: `http://master-account.stg-saas.localhost:3000/`, `http://provider-admin.stg-saas.localhost:3000/`, `http://provider.stg-saas.localhost:3000/` respectively. 
+Note down the credentials generated at this moment, to log in to each of the portals above.
```
hallelujah, line 318: In fact it should not be `stg-saas.localhost`, but `any-domain.localhost`

Martouta: Right @hallelujah , but `stg-saas.localhost` is the one in  the **example** config `config/examples/domain_substitution.yml`, as you can see:
https://github.com/3scale/porta/blob/a14d790a70c28f974f9d2a2af45575be56e64b55/config/examples/domain_substitution.yml#L1-L19

This means that it is what the people following this guide will have at this point in `config/domain_substitution.yml` because in a previous step we told them to do `cp config/examples/* config/`. Look at https://github.com/3scale/porta/blob/master/INSTALL.md#config-files.

Conclusion: IMO, the simplest approach in this line would be mentioning `stg-saas.localhost` as @srijitamukherjee did, for anyone who just wants to quickly set the Porta/System up without knowing any internals of it.

hallelujah: replace `stg-saas` by something else something like `3scale.localhost` ;)

Martouta: I like more `3scale.localhost` too 😃

Martouta (changes requested):
Please change `stg-saas.localhost` for `3sacale.localhost` both in the README and in `config/examples/domain_substitution.yml`

Martouta (approved):

PR #2146, opened 2020-08-31
Update branding_state to custom _brand on runtime

srijitamukherjee:
Signed-off-by: Srijita Mukherjee <srijita92mukherjee@gmail.com>

>While authenticating using GitHub Client-Id and Client-Secret on Developer Portal, the branding_state parameter will change to custom_brand.
Previously, the "Test authentication flow" URL – included 3scale's Client ID for GitHub app.
Now, it redirects to the external Application link provided in the GitHub Developer settings. 

>Partial fix for https://issues.redhat.com/browse/THREESCALE-4144

**Verification steps** 

>Add Client-Id and Client-Secret in Developer Portal -> SSO Integration -> GitHub.
> 
This PR is part of https://github.com/3scale/porta/pull/2050. 
Keeping 2050 for resolving the branding visibility fix, which, as of now, is under discussion.

### DIFF
--- a/app/controllers/provider/admin/authentication_providers_controller.rb
+++ b/app/controllers/provider/admin/authentication_providers_controller.rb
@@ -98,7 +98,8 @@ class Provider::Admin::AuthenticationProvidersController < FrontendController
                      username_key trust_email].freeze
 
   def update_params
-    params.require(:authentication_provider).permit(UPDATE_PARAMS)
+    defaults = { branding_state_event: 'custom_brand' }
+    params.require(:authentication_provider).permit(UPDATE_PARAMS).reverse_merge(defaults)
   end
 
   def create_params
--- a/app/controllers/provider/admin/authentication_providers_controller.rb
+++ b/app/controllers/provider/admin/authentication_providers_controller.rb
@@ -97,10 +97,11 @@ class Provider::Admin::AuthenticationProvidersController < FrontendController
                      token_url authorize_url user_info_url identifier_key
                      username_key trust_email].freeze
 
-  def update_params
-    defaults = { branding_state_event: 'custom_brand' }
-    params.require(:authentication_provider).permit(UPDATE_PARAMS).reverse_merge(defaults)
-  end
+UPDATE_PARAM_DEFAULTS = { branding_state_event: 'custom_brand' }.freeze
+
+def update_params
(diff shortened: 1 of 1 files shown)


app/controllers/provider/admin/authentication_providers_controller.rb
```diff
@@ -98,7 +98,8 @@ def authorize_authentication_provider(action = action_name)
                      username_key trust_email].freeze
 
   def update_params
-    params.require(:authentication_provider).permit(UPDATE_PARAMS)
+    defaults = { branding_state_event: 'custom_brand' }
```
guicassolato, line 101: Can we make it a constant of the class instead? E.g.

```ruby
class Provider::Admin::AuthenticationProvidersController < FrontendController
  (...)

  UPDATE_PARAM_DEFAULTS = { branding_state_event: 'custom_brand' }.freeze

  def update_params
    params.require(:authentication_provider).permit(UPDATE_PARAMS).reverse_merge(UPDATE_PARAM_DEFAULTS)
  end

  (...)
end
```

guicassolato: _Question:_ does it have to be a default on create as well? If so, should the default be better defined this at the level of the model maybe?

app/controllers/provider/admin/authentication_providers_controller.rb
```diff
@@ -97,9 +97,11 @@ def authorize_authentication_provider(action = action_name)
                      token_url authorize_url user_info_url identifier_key
                      username_key trust_email].freeze
 
-  def update_params
-    params.require(:authentication_provider).permit(UPDATE_PARAMS)
-  end
+UPDATE_PARAM_DEFAULTS = { branding_state_event: 'custom_brand' }.freeze
```
Martouta, line 100: I see several problems with this.

- This shouldn't happen only for `update`, but also for `create`.

- This is business logic, so it shouldn't be in the controller.

- This controller is for authentication providers in general, but this `branding_state_event` only exists for `Github`:
image
https://github.com/3scale/porta/blob/2346d59b412e27506980bc2a78b09437d988b515/app/models/authentication_provider/github.rb#L19-L21
So doing an `update` with this controller for another authentication provider will fail.

hallelujah:

PR #2149, opened 2020-09-01
Apply member permissions to Admin API's ActiveDoc Specs endpoints

guicassolato:
- index only includes accessible api docs specs
- show, update and delete will throw `404 Not Found` if api docs spec is not accessible
- create and update will return `422 Unprocessable Entity` if `service_id` param is not of an accessible service
- account-level api docs specs remain accessible

Close [THREESCALE-5884](https://issues.redhat.com/browse/THREESCALE-5884)

### DIFF
--- a/app/controllers/admin/api/api_docs_services_controller.rb
+++ b/app/controllers/admin/api/api_docs_services_controller.rb
@@ -134,9 +134,7 @@ class Admin::Api::ApiDocsServicesController < Admin::Api::BaseController
   end
 
   def api_docs_services
-    collection = current_account.api_docs_services.accessible
-    return collection unless current_user
-    collection.permitted_for(current_user)
+    current_account.api_docs_services.accessible.permitted_for(current_user)
   end
 
   def find_api_docs_service
--- a/app/controllers/admin/api/api_docs_services_controller.rb
+++ b/app/controllers/admin/api/api_docs_services_controller.rb
@@ -1,6 +1,7 @@
 class Admin::Api::ApiDocsServicesController < Admin::Api::BaseController
   before_action :deny_on_premises_for_master
   before_action :find_api_docs_service, only: %i[show update destroy]
+  before_action :find_service, only: %i[create update]
 
   wrap_parameters ::ApiDocs::Service, name: :api_docs_service, include: ::ApiDocs::Service.attribute_names
 
@@ -28,7 +29,7 @@ class Admin::Api::ApiDocsServicesController < Admin::Api::BaseController
   ##~ op.parameters.add @parameter_access_token
   #
   def index
-    @api_docs_services = current_account.api_docs_services.all
+    @api_docs_services = api_docs_services.all
     respond_with(@api_docs_services)
   end
 
@@ -132,8 +133,20 @@ class Admin::Api::ApiDocsServicesController < Admin::Api::BaseController
     params.require(:api_docs_service).permit(*permit_params)
   end
 
-  def find_api_docs_service
-    @api_docs_service = current_account.api_docs_services.find(params[:id])
+  def api_docs_services
+    collection = current_account.api_docs_services.accessible
+    return collection unless current_user
+    collection.permitted_for(current_user)
   end
 
+  def find_api_docs_service
+    @api_docs_service = api_docs_services.find(params[:id])
+  end
+
+  def find_service
+    service_id = api_docs_params[:service_id]
+    service_id.blank? || current_user.blank? || current_user.accessible_services.find(service_id)
+  rescue ActiveRecord::RecordNotFound
+    render_error('Service not found', status: :unprocessable_entity)
+  end
 end
--- a/app/models/api_docs/service.rb
+++ b/app/models/api_docs/service.rb
@@ -27,6 +27,7 @@ class ApiDocs::Service < ApplicationRecord
   scope :published, -> { where(published: true) }
   scope :accessible, -> { joining { service.outer }.where.has { (service_id == nil) | (service.state != ::Service::DELETE_STATE) } }
   scope :without_service, -> { where(service_id: nil) }
+  scope :permitted_for, ->(user) { where.has { (service_id == nil) | service_id.in(user.accessible_services.select(:id)) } }
 
   before_save :set_default_values
   before_save :prepare_base_path_notify
--- a/app/models/api_docs/service.rb
+++ b/app/models/api_docs/service.rb
@@ -27,7 +27,7 @@ class ApiDocs::Service < ApplicationRecord
   scope :published, -> { where(published: true) }
   scope :accessible, -> { joining { service.outer }.where.has { (service_id == nil) | (service.state != ::Service::DELETE_STATE) } }
   scope :without_service, -> { where(service_id: nil) }
-  scope :permitted_for, ->(user) { where.has { (service_id == nil) | service_id.in(user.accessible_services.select(:id)) } }
+  scope :permitted_for, ->(user = nil) { user ? where.has { (service_id == nil) | service_id.in(user.accessible_services.select(:id)) } : self }
 
   before_save :set_default_values
   before_save :prepare_base_path_notify
--- a/test/integration/admin/api/api_docs_services_controller_test.rb
+++ b/test/integration/admin/api/api_docs_services_controller_test.rb
@@ -90,7 +90,7 @@ class Admin::Api::ApiDocsServicesControllerTest < ActionDispatch::IntegrationTes
     def test_update_unexistent_service
       put admin_api_active_doc_path(api_docs_service, format: :json), update_params(service_id: Service.last.id + 1)
       assert_response :unprocessable_entity
-      assert_contains JSON.parse(response.body).dig('errors', 'service'), 'not found'
+      assert_equal 'Service not found', JSON.parse(response.body)['error']
     end
 
     test 'show' do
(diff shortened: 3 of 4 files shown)


app/controllers/admin/api/api_docs_services_controller.rb
```diff
@@ -132,8 +133,20 @@ def api_docs_params(*extra_params)
...
+  end
+
   def find_api_docs_service
-    @api_docs_service = current_account.api_docs_services.find(params[:id])
+    @api_docs_service = api_docs_services.find(params[:id])
   end
 
+  def find_service
+    service_id = api_docs_params[:service_id]
+    service_id.blank? || current_user.blank? || current_user.accessible_services.find(service_id)
```
Martouta, line 148: Question 🤔  : should these endpoints work with `provider_key` ? If so, it logs in with `current_account` and without a `current_user`

Martouta: Ok, I see, the code keeps searching by `current_account` in the methods, and here it only searches the `service_id` if it is authenticated by access token 👍 
That means that indeed it should work for `provider_key` 👍

app/models/api_docs/service.rb
```diff
@@ -27,6 +27,7 @@ class ApiDocs::Service < ApplicationRecord
   scope :published, -> { where(published: true) }
   scope :accessible, -> { joining { service.outer }.where.has { (service_id == nil) | (service.state != ::Service::DELETE_STATE) } }
   scope :without_service, -> { where(service_id: nil) }
+  scope :permitted_for, ->(user) { where.has { (service_id == nil) | service_id.in(user.accessible_services.select(:id)) } }
```
Martouta, line 30: The code is good but this scope probably needs a unit test as well 🙈 Plus, let's run the tests in the 3 DBs, just in case 🙏

test/integration/admin/api/api_docs_services_controller_test.rb
```diff
@@ -90,7 +90,7 @@ def test_update_invalid_params
     def test_update_unexistent_service
       put admin_api_active_doc_path(api_docs_service, format: :json), update_params(service_id: Service.last.id + 1)
       assert_response :unprocessable_entity
-      assert_contains JSON.parse(response.body).dig('errors', 'service'), 'not found'
```
Martouta, line 93: I hope that nobody has been relying on this error message format so far 👀

app/controllers/admin/api/api_docs_services_controller.rb
guicassolato: It should continue to work with `provider_key`. And, as you can see, the method does NOT return the found service. It is a "find_service" method that behaves more like an "authorize_service" method. The reason why I didn't want to call it `authorize_service` is because we usually use `authorize_*` when relying on CanCan for the implementation of authorization (i.e. `authorize! ...`)

app/controllers/admin/api/api_docs_services_controller.rb
```diff
@@ -132,8 +133,20 @@ def api_docs_params(*extra_params)
...
 
+  def api_docs_services
+    collection = current_account.api_docs_services.accessible
+    return collection unless current_user
+    collection.permitted_for(current_user)
+  end
+
   def find_api_docs_service
-    @api_docs_service = current_account.api_docs_services.find(params[:id])
+    @api_docs_service = api_docs_services.find(params[:id])
```
Martouta, line 143: Perhaps we should also add an `attr_reader :api_docs_service` and then in all the other methods (such as `update`) we could use `api_docs_service` instead of `@api_docs_service`.
Not a big deal either way 👍

app/controllers/admin/api/api_docs_services_controller.rb
```diff
@@ -28,7 +29,7 @@ class Admin::Api::ApiDocsServicesController < Admin::Api::BaseController
   ##~ op.parameters.add @parameter_access_token
   #
   def index
-    @api_docs_services = current_account.api_docs_services.all
+    @api_docs_services = api_docs_services.all
     respond_with(@api_docs_services)
   end
```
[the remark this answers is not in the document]
guicassolato: Yeah. IDK why it sets an instance variable 😕

test/integration/admin/api/api_docs_services_controller_test.rb
guicassolato: Yep. It is a breaking change.

test/integration/admin/api/api_docs_services_controller_test.rb
```diff
@@ -103,6 +103,83 @@ def test_update_unexistent_service
...
+        assert_contains api_docs_services_ids, account_level_api_docs_service.id
+      end
+
+      protected
+
+      def path_params
+        { access_token: access_token.value, format: :json }
+      end
+
+      def api_doc_params(**extra_params)
```
Martouta, line 169: If this class inherits from `Admin::Api::ApiDocsServicesControllerTest`, you can reuse this method.

We should also remove the `login!` in the parent class because using `login!` for API tests is not a very good idea 😄

app/controllers/admin/api/api_docs_services_controller.rb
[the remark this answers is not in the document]
Martouta: no need at all 😂 and the `.all` is also overkill 🔫

Martouta (dismissed):
It is good 👍 The only thing that I would request out of safety is:
1. Adding a unit test for `permitted_for` and run CircleCI for the 3 DBs. https://github.com/3scale/porta/pull/2149/files#r481141428
2. Inform Product/Support about the change in the response error message. Just in case someone complains, they probably need to know. https://github.com/3scale/porta/pull/2149/files#r481142243

app/controllers/admin/api/api_docs_services_controller.rb
Martouta: Makes sense 👍

app/models/api_docs/service.rb
guicassolato: Done

Martouta (approved):
Awesome!! Thanks! 👏

test/integration/admin/api/api_docs_services_controller_test.rb
guicassolato: /cc @3scale/product @3scale/qe @3scale/support 

When one tries to update an existing ActiveDoc modifying the {{service_id}} attribute to a value that is NOT a valid ID of an API of the provider, instead of the previous error response in the typical format of ActiveRecord validation errors:

```json
{ "errors" : { "service" : ["not found"] } }
```

the PR modifies the response to the standard format:
```json
{ "error" : "Service not found" }
```

app/controllers/admin/api/api_docs_services_controller.rb
```diff
@@ -132,8 +133,20 @@ def api_docs_params(*extra_params)
     params.require(:api_docs_service).permit(*permit_params)
   end
 
+  def api_docs_services
+    collection = current_account.api_docs_services.accessible
+    return collection unless current_user
+    collection.permitted_for(current_user)
```
hallelujah, line 139: I'd do it inside the scope `permitted_for`, default will be `nil`
```ruby
scope :permitted_for, ->(user) { user ? where.has { (service_id == nil) | service_id.in(user.accessible_services.select(:id)) : self } }
```

app/controllers/admin/api/api_docs_services_controller.rb
hallelujah, line 143: Why not change this to `params.require(:id)`

app/controllers/admin/api/api_docs_services_controller.rb
```diff
@@ -132,8 +133,20 @@ def api_docs_params(*extra_params)
...
   def find_api_docs_service
-    @api_docs_service = current_account.api_docs_services.find(params[:id])
+    @api_docs_service = api_docs_services.find(params[:id])
   end
 
+  def find_service
+    service_id = api_docs_params[:service_id]
+    service_id.blank? || current_user.blank? || current_user.accessible_services.find(service_id)
+  rescue ActiveRecord::RecordNotFound
+    render_error('Service not found', status: :unprocessable_entity)
```
hallelujah, line 150: `not_found` or `unprocessable_entity`? I'd prefer the first one as the error clearly mentioned it.
You try to find an object that belongs to a resource that does not exist.

Applying unprocessable_entity would be for example, trying to query an object that exists but in a state that cannot be processed.

guicassolato: Actually, we have 3 options here:
- not found
- unprocessable entity
- forbidden

In reality, what happens here is that the service is either _forbidden_ or _not found_. The reason why I didn't go with _not found_ is because the resource targeted is the api doc spec (not the service) and the api doc spec may exist. It would be weird to fetch an existing api doc spec, then try to modify the parameter `service_id` of the spec to a value that is not acceptable and receive a 404 as response, but a 200 when the value if acceptable for the parameter. The target resource is not the one missing and it reveals too much to the user about the dependencies between objects.

Finally, I didn't go with _forbidden_ because it's considered a good practice not to distinguish between "forbidden" and "not found" when a resource exists but it's forbidden to the requestor. Otherwise, we can open a breach for exploring existing object ids in the system.

app/controllers/admin/api/api_docs_services_controller.rb
guicassolato: This param is in the URL. It will always be present or the system will respond with a 404.

Martouta (approved):

PR #2151, opened 2020-09-01
Fix OracleEnhancedAdapter log issue

guicassolato:
Fix for https://app.circleci.com/pipelines/github/3scale/porta/13712/workflows/b9c2e0ba-0fda-4974-9a8e-a0e08c5efa53/jobs/184671

We'd better consider updating activerecord-oracle_enhanced-adapter soon anyway.

### DIFF
--- a/config/initializers/oracle.rb
+++ b/config/initializers/oracle.rb
@@ -23,6 +23,7 @@ if System::Database.oracle?
   ActiveRecord::ConnectionAdapters::OracleEnhancedAdapter.class_eval do
     remove_const(:IDENTIFIER_MAX_LENGTH)
     const_set(:IDENTIFIER_MAX_LENGTH, 128)
+    remove_method(:log)
 
     prepend(Module.new do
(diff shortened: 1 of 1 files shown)


Martouta (approved):
Alt Text

config/initializers/oracle.rb
```diff
@@ -23,6 +23,7 @@ def column(name, type, options = {})
   ActiveRecord::ConnectionAdapters::OracleEnhancedAdapter.class_eval do
     remove_const(:IDENTIFIER_MAX_LENGTH)
     const_set(:IDENTIFIER_MAX_LENGTH, 128)
+    remove_method(:log)
```
hallelujah, line 26: If you prepend no need to remove the method no?
But that's fine

hallelujah (approved):

config/initializers/oracle.rb
guicassolato: We need to remove it because it calls super and we cannot pass through the old implementation.

PR #2152, opened 2020-09-03
Fix member users can manage service resources without permission

guicassolato:
**What this PR do**
- [x] Ensure members cannot navigate to any service page if user does not have permission over the service (service-level access control)
- [x] Ensure members cannot navigate to ApiDocs::Service pages when the user does not have the “Integration & Application Plans” permission (permission-level access control)
- [x] Ensure members cannot use ApiDocs::Service endpoints of the Admin API unless they have the “Integration & Application Plans” permission (permission-level access control)
- [x] Filter account index of ApiDocs::Service (Audience > Developer Portal > ActiveDocs) to consider services the user has access to
- [x] Ensure only permitted services are listed in the select box to choose the service of an ApiDocs::Service
- [x] Ensure user cannot toggle visibility, update, nor delete ApiDocs::Services whose service the user does not have access to
- [x] Rename `Service.permitted_for_user` -> `Service.permitted_for`
- [x] Refactoring of `Service.permitted_for` to only include services of the provider the user belongs to
- [x] Fix crash of Dashboard when the user does not have permission over any service

**What this PR does NOT do**
- Permission-level access control other than the ones mentioned above

Closes [THREESCALE-952](https://issues.redhat.com/browse/THREESCALE-952) and a small part of [THREESCALE-5867](https://issues.redhat.com/browse/THREESCALE-5867)

### DIFF
--- a/app/controllers/admin/api/api_docs_services_controller.rb
+++ b/app/controllers/admin/api/api_docs_services_controller.rb
@@ -1,7 +1,10 @@
+# frozen_string_literal: true
+
 class Admin::Api::ApiDocsServicesController < Admin::Api::BaseController
   before_action :deny_on_premises_for_master
+  before_action :authorize_api_docs
   before_action :find_api_docs_service, only: %i[show update destroy]
-  before_action :find_service, only: %i[create update]
+  before_action :new_service_id_permitted, only: %i[create update]
 
   wrap_parameters ::ApiDocs::Service, name: :api_docs_service, include: ::ApiDocs::Service.attribute_names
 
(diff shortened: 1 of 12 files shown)


app/controllers/admin/api/api_docs_services_controller.rb
```diff
@@ -141,10 +144,14 @@ def find_api_docs_service
...
-  def find_service
+  def new_service_id_permitted
     service_id = api_docs_params[:service_id]
     service_id.blank? || current_user.blank? || current_user.accessible_services.find(service_id)
   rescue ActiveRecord::RecordNotFound
     render_error('Service not found', status: :unprocessable_entity)
   end
+
+  def authorize_api_docs
+    authorize! :manage, :plans if current_user
```
hallelujah, line 155: This is because it is also accessible with provider_key! :+1:

hallelujah (approved):

PR #2153, opened 2020-09-03
Require 'plans' permission for member users to manage API backends

guicassolato:
Permission-level control over the API backend section endpoints of the Admin Portal and Admin API.

API backends are not scoped by service (API product) and it is not possible to do member access control at the API backend level the way we do for services at the service-level. Therefore, here we are simply enforcing that only member users with the “Integration & Application Plans” permission can access the API backend section.

Creating and deleting API backends remain admin users-exclusive.

| Resource                                | Action                                         | Required permission |
|-----------------------------------------|------------------------------------------------|-----------------|
| ☑︎ backend api                           | new backend api                                | \<admin-only\>  |
|                                         | create backend api                             | \<admin-only\>  |
|                                         | show backend api (overview)                    | plans           |
|                                         | edit backend api                               | plans           |
|                                         | update backend api                             | plans           |
|                                         | delete backend api                             | \<admin-only\>  |
| ☐ backend api stats                     | backend api stats                              | monitoring      |
| ☑︎ backend api metrics                   | index backend api metrics                      | plans           |
|                                         | new backend api top-level metric               | plans           |
|                                         | create backend api top-level metric            | plans           |
|                                         | new backend api method metric                  | plans           |
|                                         | create backend api method metric               | plans           |
|                                         | edit backend api metric                        | plans           |
|                                         | update backend api metric                      | plans           |
|                                         | delete backend api metric                      | plans           |
| ☑︎ backend api mapping rules             | index backend api mapping rules                | plans           |
|                                         | new backend api mapping rule                   | plans           |
|                                         | create backend api mapping rule                | plans           |
|                                         | edit backend api mapping rule                  | plans           |
|                                         | update backend api mapping rule                | plans           |
|                                         | delete backend api mapping rule                | plans           |

Closes [THREESCALE-5915](https://issues.redhat.com/browse/THREESCALE-5915).

### DIFF
--- a/app/controllers/admin/api/backend_apis/base_controller.rb
+++ b/app/controllers/admin/api/backend_apis/base_controller.rb
@@ -17,6 +17,7 @@ class Admin::Api::BackendApis::BaseController < Admin::Api::BaseController
   end
 
   def authorize
-    authorize! :manage, BackendApi
+    return unless current_user # provider_key access
+    authorize! :edit, BackendApi
   end
 end
--- a/app/controllers/provider/admin/backend_apis/base_controller.rb
+++ b/app/controllers/provider/admin/backend_apis/base_controller.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 class Provider::Admin::BackendApis::BaseController < Provider::Admin::BaseController
+  before_action :authorize
   before_action :find_backend_api
 
   activate_menu :backend_api
@@ -10,5 +11,9 @@ class Provider::Admin::BackendApis::BaseController < Provider::Admin::BaseContro
 
   def find_backend_api
     @backend_api = current_account.backend_apis.accessible.find(params[:backend_api_id])
+  end
+
+  def authorize
+    authorize! :edit, BackendApi
   end
 end
--- a/app/controllers/provider/admin/backend_apis_controller.rb
+++ b/app/controllers/provider/admin/backend_apis_controller.rb
@@ -1,19 +1,16 @@
 # frozen_string_literal: true
 
 class Provider::Admin::BackendApisController < Provider::Admin::BaseController
-  before_action :find_backend_api, except: %i[new create]
-  before_action :authorize
+  load_and_authorize_resource :backend_api, through: :current_user, through_association: :accessible_backend_apis
 
   activate_menu :backend_api, :overview
   layout 'provider'
 
   def new
     activate_menu :dashboard
-    @backend_api = collection.build params[:backend_api]
   end
 
   def create
-    @backend_api = current_account.backend_apis.build(create_params)
     if @backend_api.save
       redirect_to provider_admin_backend_api_path(@backend_api), notice: 'Backend created'
     else
@@ -50,23 +47,13 @@ class Provider::Admin::BackendApisController < Provider::Admin::BaseController
   DEFAULT_PARAMS = %i[name description private_endpoint].freeze
   private_constant :DEFAULT_PARAMS
 
-  def find_backend_api
-    @backend_api = current_account.backend_apis.accessible.find(params[:id])
+  def backend_api_params(*extra_params)
+    params.require(:backend_api).permit(DEFAULT_PARAMS | extra_params)
   end
 
   def create_params
-    params.require(:backend_api).permit(DEFAULT_PARAMS | %i[system_name])
+    backend_api_params(:system_name)
   end
 
-  def update_params
-    params.require(:backend_api).permit(DEFAULT_PARAMS)
-  end
-
-  def authorize
-    authorize! :manage, BackendApi
-  end
-
-  def collection
-    current_account.backend_apis
-  end
+  alias update_params backend_api_params
 end
--- a/app/models/account/provider_methods.rb
+++ b/app/models/account/provider_methods.rb
@@ -50,6 +50,7 @@ module Account::ProviderMethods
     end
 
     has_many :backend_apis, inverse_of: :account, dependent: :destroy
+    has_many :accessible_backend_apis, -> { accessible }, class_name: 'BackendApi'
 
     has_attached_file :proxy_configs, storage: :s3, path: '.sandbox_proxy/sandbox_proxy_:id.lua'
     do_not_validate_attachment_file_type :proxy_configs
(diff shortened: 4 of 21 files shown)


app/controllers/provider/admin/backend_apis/base_controller.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 class Provider::Admin::BackendApis::BaseController < Provider::Admin::BaseController
+  before_action :authorize_plans!
```
hallelujah, line 4: Does it have an API counterpart?

app/controllers/provider/admin/backend_apis/base_controller.rb
```diff
@@ -11,4 +12,8 @@ class Provider::Admin::BackendApis::BaseController < Provider::Admin::BaseContro
   def find_backend_api
     @backend_api = current_account.backend_apis.accessible.find(params[:backend_api_id])
   end
+
+  def authorize_plans!
+    authorize! :manage, :plans
```
hallelujah, line 17: Why not `authorhize! :manage, BackendAPI` like https://github.com/3scale/porta/blob/455e9aae0f9a27aa8b97e0bdbf72b26451901e2a/app/controllers/admin/api/backend_apis_controller.rb#L132

guicassolato: I will be. Please don't review this just yet.

test/integration/admin/api/services/backend_usages_controller_test.rb
```diff
@@ -144,15 +144,15 @@ def setup
...
+      post admin_api_service_backend_usages_path(collection_params), params: { path: 'foo/bar', backend_api_id: backend_api.id }
       assert_response :not_found
     end
   end
 
   class WithMemberAccessToken < self
     def setup
       super
-      @member = FactoryBot.create(:member, account: tenant, admin_sections: %w[partners])
+      @member = FactoryBot.create(:member, account: tenant, admin_sections: %w[partners plans]) # FIXME: it should not require 'partners' permission
```
guicassolato, line 155: This is because we don't have a `plans` scope for the Admin API
https://github.com/3scale/porta/blob/f6f7a12ad20fb847b82397004eba3cfec8e39882/app/models/access_token.rb#L21

thus the `account_management` scope is mapped to the `partners` permission
https://github.com/3scale/porta/blob/f6f7a12ad20fb847b82397004eba3cfec8e39882/app/models/access_token.rb#L28

As a consequence, a user that has in the UI only the "Integration & Application Plans" permission (i.e. `plans`), because the corresponding endpoints in the Admin API are all scoped under the "Account Management" scope, requires his/her privileges to be raised to include as well the "Accounts -- Applications" (i.e. `partners`) permission, both in the UI and API, only so the can perform "Integration & Application Plans"-related actions in the API.

app/controllers/provider/admin/backend_apis_controller.rb
```diff
@@ -1,19 +1,16 @@
 # frozen_string_literal: true
 
 class Provider::Admin::BackendApisController < Provider::Admin::BaseController
-  before_action :find_backend_api, except: %i[new create]
-  before_action :authorize
+  load_and_authorize_resource :backend_api, through: :current_user, through_association: :accessible_backend_apis
```
hallelujah, line 4:
```suggestion
  load_and_authorize_resource :backend_api, through: :current_user, through_association: :accessible
```

And no need to create an association for this

https://github.com/CanCanCommunity/cancancan/blob/2.3.0/lib/cancan/controller_resource_loader.rb#L67

app/models/account/provider_methods.rb
```diff
@@ -50,6 +50,7 @@ def build_kind(kind:, **attributes)
     end
 
     has_many :backend_apis, inverse_of: :account, dependent: :destroy
+    has_many :accessible_backend_apis, -> { accessible }, class_name: 'BackendApi'
```
hallelujah, line 53: See https://github.com/3scale/porta/pull/2153/files#r484969685

app/controllers/provider/admin/backend_apis_controller.rb
guicassolato: I'm afraid it doesn't work like this. What you suggested will invoke `accessible` on `current_user`, resulting in
```
NoMethodError: undefined method `accessible' for #<User:xxx>
```

app/controllers/admin/api/backend_apis/base_controller.rb
```diff
@@ -17,6 +17,7 @@ def backend_api
   end
 
   def authorize
-    authorize! :manage, BackendApi
+    return unless current_user # provider_key access
+    authorize! :edit, BackendApi
```
Martouta, line 21: Why not `authorize! action_name.to_sym, BackendApi` as you did for the other controller? 🤔

config/abilities/provider_admin.rb
```diff
@@ -79,10 +79,9 @@
     service.account_id == user.account_id && can?(:manage, :multiple_services) and not service.default_or_last?
   end
 
-  if account.provider_can_use?(:api_as_product)
-    can :manage, BackendApi
-    can :manage, BackendApiConfig
-  end
+  can %i[index show edit update create destroy], BackendApi
```
Martouta, line 82: It think that it should be like this:
```suggestion
  can %i[index show edit update new create destroy], BackendApi
```

Because `Provider::Admin::BackendApisController` has the `new` action.

Martouta: My bad, I see that see handle that scenario in `config/initializers/cancan_hacks.rb` checking `show`.

Martouta (approved):

app/controllers/admin/api/backend_apis/base_controller.rb
guicassolato: This is for metrics and mapping rules, not the api backend that owns those. Whoever can _edit_ an api backend (but not  necessarily _create_ and _delete_ those) can also do everything related to the metrics and mapping rules of the backend. Think of it as handling the composing parts of a backend is equivalent to editing the backend ;)

app/controllers/provider/admin/backend_apis/base_controller.rb
guicassolato: Done

thomasmaas (approved):

Martouta (approved):
great job! thanks! 💪 🥇 👏

PR #2155, opened 2020-09-07
Automatically redirect users to SSO if it is enforced

HemaHG:

Automatically redirect users to SSO if it is enforced

  closes https://issues.redhat.com/browse/THREESCALE-2795

output
1. **When SSO is configured but not enforced**
sso_not_enforced

2. **when only one SSO integration is configured and enforced**
only_one_sso

3. **when more than one SSO integration is configured but not enforced**
two

4.**when more than one SSO integration is configured and enforced**
more_than_one_published

### DIFF
--- a/app/controllers/provider/sessions_controller.rb
+++ b/app/controllers/provider/sessions_controller.rb
@@ -7,8 +7,8 @@ class Provider::SessionsController < FrontendController
   before_action :ensure_provider_domain
   before_action :find_provider
   before_action :instantiate_sessions_presenter, only: [:new, :create]
-  before_action :redirect_if_logged_in, :only => [:new]
-  before_action :redirect_to_sso_params, :only => [:new]
+  before_action :redirect_if_logged_in, only: %i[new]
+  before_action :redirect_to_enforced_sso, only: %i[new]
 
   def new
     @session = Session.new
@@ -65,9 +65,7 @@ class Provider::SessionsController < FrontendController
   end
 
   def redirect_if_logged_in
-    if logged_in? && current_account.provider?
-      redirect_to provider_admin_dashboard_path
-    end
+    redirect_to provider_admin_dashboard_path if logged_in?
   end
 
   def authenticate_user
@@ -103,7 +101,8 @@ class Provider::SessionsController < FrontendController
     @presenter = Provider::SessionsPresenter.new(domain_account)
   end
 
-  def redirect_to_sso_params
-    redirect_to authorization_provider_bounce_path(published_authentication_providers.first.system_name) if domain_account.settings.enforce_sso? && published_authentication_providers.count == 1
+  def redirect_to_enforced_sso
+    return if !domain_account.settings.enforce_sso? || published_authentication_providers.count != 1
+    redirect_to authorization_provider_bounce_path(published_authentication_providers.first.system_name)
   end
 end
--- a/app/controllers/provider/sessions_controller.rb
+++ b/app/controllers/provider/sessions_controller.rb
@@ -7,9 +7,10 @@ class Provider::SessionsController < FrontendController
   before_action :ensure_provider_domain
   before_action :find_provider
   before_action :instantiate_sessions_presenter, only: [:new, :create]
+  before_action :redirect_if_logged_in, :only => [:new]
+  before_action :redirect_to_sso_params, :only => [:new]
 
   def new
-    redirect_to provider_admin_dashboard_url if logged_in?
     @session = Session.new
     @authentication_providers = published_authentication_providers
   end
@@ -100,5 +101,9 @@ class Provider::SessionsController < FrontendController
 
   def instantiate_sessions_presenter
     @presenter = Provider::SessionsPresenter.new(domain_account)
+  end
+
+  def redirect_to_sso_params
+    redirect_to authorization_provider_bounce_path(published_authentication_providers.first.system_name) if domain_account.settings.enforce_sso? && published_authentication_providers.count == 1
   end
 end
--- a/test/integration/provider/sessions_controller_test.rb
+++ b/test/integration/provider/sessions_controller_test.rb
@@ -39,19 +39,25 @@ class Provider::SessionsControllerTest < ActionDispatch::IntegrationTest
   end
 
   test "redirect users to SSO when there's only one authentication provider and enforce_sso is on" do
-  @provider.settings.update_column(:enforce_sso, true)
-  get new_provider_sessions_path
-  request = ActionDispatch::TestRequest.create
-  assert_redirected_to authorization_provider_bounce_path(authentication_provider.system_name)
+    @provider.settings.update_column(:enforce_sso, true)
+    get new_provider_sessions_path
+    assert_redirected_to authorization_provider_bounce_path(authentication_provider.system_name)
+  end
+
+  test "redirect users to SSO when there's more than one authentication provider but only one is published " do
+    @provider.settings.update_column(:enforce_sso, true)
+    FactoryBot.create(:self_authentication_provider, account: @provider)
+    get new_provider_sessions_path
+    assert_redirected_to authorization_provider_bounce_path(authentication_provider.system_name)
   end
 
   test "does not redirect when there's more than one authentication provider" do
-  @provider.settings.update_column(:enforce_sso, true)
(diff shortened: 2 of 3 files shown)


hallelujah (commented):
Could you add a test for this @HemaHG ?

app/controllers/provider/sessions_controller.rb
```diff
@@ -10,8 +10,14 @@ class Provider::SessionsController < FrontendController
 
   def new
     redirect_to provider_admin_dashboard_url if logged_in?
-    @session = Session.new
-    @authentication_providers = published_authentication_providers
+    if domain_account.settings.enforce_sso? && domain_account.sso_authorizations.size == 1
+      domain_account.self_authentication_providers.published.map do |auth|
```
Martouta, line 14: Why `map`? 😕 Shouldn't we redirect to only 1 authentication provider?

Martouta: And it should be the same as the `sso_authorization` above, I think.

Martouta: I mean: 
```suggestion
  auth = domain_account.sso_authorizations.authentication_provider
```

app/controllers/provider/sessions_controller.rb
```diff
@@ -10,8 +10,14 @@ class Provider::SessionsController < FrontendController
 
   def new
     redirect_to provider_admin_dashboard_url if logged_in?
-    @session = Session.new
-    @authentication_providers = published_authentication_providers
+    if domain_account.settings.enforce_sso? && domain_account.sso_authorizations.size == 1
```
Martouta, line 13: If you do this: https://github.com/3scale/porta/pull/2155/files#r484300055
Then this line probably should check that the auth provider of the sso_authorization is published before doing the count/size == 1

app/controllers/provider/sessions_controller.rb
```diff
@@ -10,6 +10,7 @@ class Provider::SessionsController < FrontendController
 
   def new
     redirect_to provider_admin_dashboard_url if logged_in?
+    redirect_to authorization_provider_bounce_path(published_authentication_providers.first.system_name) if domain_account.settings.enforce_sso? && published_authentication_providers.count == 1
```
hallelujah, line 13: Those could be 2 `before_action`

hallelujah (approved):

app/controllers/provider/sessions_controller.rb
```diff
@@ -7,9 +7,10 @@ class Provider::SessionsController < FrontendController
   before_action :ensure_provider_domain
   before_action :find_provider
   before_action :instantiate_sessions_presenter, only: [:new, :create]
+  before_action :redirect_if_logged_in, :only => [:new]
```
Martouta, line 10:
```suggestion
  before_action :redirect_if_logged_in, only: %i[new]
```

I see that the method `redirect_if_logged_in` already exists in this file and contains:
```ruby
def redirect_if_logged_in
  if logged_in? && current_account.provider?
    redirect_to provider_admin_dashboard_path
  end
end
```
I wonder if the condition `current_account.provider?` is necessary... is this controller ever used for buyers? I think not, but if so, what do we have to do when the logged in account is a buyer?

app/controllers/provider/sessions_controller.rb
```diff
@@ -101,4 +102,8 @@ def session_return_to
   def instantiate_sessions_presenter
     @presenter = Provider::SessionsPresenter.new(domain_account)
   end
+
+  def redirect_to_sso_params
```
Martouta, line 106: This name is neither correct nor understandable IMO...
Maybe we can go for something like `redirect_to_enforced_sso`

app/controllers/provider/sessions_controller.rb
```diff
@@ -7,9 +7,10 @@ class Provider::SessionsController < FrontendController
   before_action :ensure_provider_domain
   before_action :find_provider
   before_action :instantiate_sessions_presenter, only: [:new, :create]
+  before_action :redirect_if_logged_in, :only => [:new]
+  before_action :redirect_to_sso_params, :only => [:new]
```
Martouta, line 11: Follow the format suggested above 😄

test/integration/provider/sessions_controller_test.rb
```diff
@@ -37,4 +37,21 @@ class Provider::SessionsControllerTest < ActionDispatch::IntegrationTest
...
+  end
+
+  test "does not redirect when there's more than one authentication provider" do
+  @provider.settings.update_column(:enforce_sso, true)
+  FactoryBot.create(:self_authentication_provider, account: @provider, published: true)
+  get new_provider_sessions_path
+  assert_response :success
+
+  # password login is disabled, there's no passoword login form
+  refute_match 'Email or Username', response.body
```
Martouta, line 55: This is very error-prone. `refute_match` is something to be avoided as much as possible, because it might not match for the wrong reasons, for example by just changing in the UI `Email or Username` for `Username or Email`.
If there is another way to check this, it is better 😄 Otherwise, have a test for when it must match and when doesn't, so if there is a change like the example I gave you, then at least we will detect the failure sooner.

app/controllers/provider/sessions_controller.rb
```diff
@@ -101,4 +102,8 @@ def session_return_to
   def instantiate_sessions_presenter
     @presenter = Provider::SessionsPresenter.new(domain_account)
   end
+
+  def redirect_to_sso_params
+    redirect_to authorization_provider_bounce_path(published_authentication_providers.first.system_name) if domain_account.settings.enforce_sso? && published_authentication_providers.count == 1
```
Martouta, line 107: Suggestion to do it in 2 lines just for readability 😄 this is very subjective tho and you might disagree 👍 
```suggestion
    return if !domain_account.settings.enforce_sso? || published_authentication_providers.count != 1
    redirect_to authorization_provider_bounce_path(published_authentication_providers.first.system_name)
```

test/integration/provider/sessions_controller_test.rb
```diff
@@ -37,4 +37,21 @@ class Provider::SessionsControllerTest < ActionDispatch::IntegrationTest
 
     assert_redirected_to "http://example.net/?provider_id=#{account.id}&user_id=#{account.admin_user.id}"
   end
+
+  test "redirect users to SSO when there's only one authentication provider and enforce_sso is on" do
```
Martouta, line 41: This is good 👍 but it lacks one scenario for testing: when there is more than one authentication provider but for non-published ones 😄

test/integration/provider/sessions_controller_test.rb
HemaHG: test is already there for match case in test/integration/sso_enforce_flow_test.rb
    # password signup form is visible
    get new_provider_sessions_path
    assert_response :success
    assert_match 'Email or Username', response.body

Martouta (approved):
Great job! 👏 🥇

thomasmaas:
@HemaHG could you post a  short video (or 4 short videos) that shows the user experience when signing in of the following scenarios:

- only one SSO integration is configured but not enforced
- only one SSO integration is configured and enforced
- more than one SSO integration is configured but not enforced
- more than one SSO integration is configured and enforced

Thanks!

app/controllers/provider/sessions_controller.rb
```diff
@@ -101,4 +100,9 @@ def session_return_to
   def instantiate_sessions_presenter
     @presenter = Provider::SessionsPresenter.new(domain_account)
   end
+
+  def redirect_to_enforced_sso
+    return if !domain_account.settings.enforce_sso? || published_authentication_providers.count != 1
```
hallelujah, line 105:
```suggestion
    return unless domain_account.settings.enforce_sso? && published_authentication_providers.count == 1
```

hallelujah (approved):
Just pinpointing in my comment. LGTM!

app/controllers/provider/sessions_controller.rb
Martouta: Then it can also become:
```suggestion
    return unless domain_account.settings.enforce_sso? && published_authentication_providers.one?
```

hallelujah:
Merging @HemaHG Thank you! Make another PR for the improvements

PR #2156, opened 2020-09-07
Improve performance in API Services List

Martouta:
1st part of [THREESCALE-4528](https://issues.redhat.com/browse/THREESCALE-4528)

#### N + 1 for Proxy and Account

image

image

##### Query by account_id and state

image

##### Query by account_id and state and permitted_for(user)

image

#### Paginate
Not sure if it is a good idea because there are people already using this API and this is a breaking change, but to some extend it is necessary because no matter how much we optimize, there is always a limit of what we are able to load and respond at once.

### Verification steps

1. Create many services, for different accounts, and keeping in mind that a few services must have state 'deleted'.
Still not keeping in mind the **permitted_for**.

`bundle exec rails console` and run:
```ruby
tenant_1st = Account.tenants.first!

# 1. create a new tenant
def create_tenant(signup_params)
  begin
    Signup::ProviderAccountManager.new(Account.master).create(signup_params)  
  rescue ThreeScale::Core::APIClient::ConnectionError  , ThreeScale::Core::APIClient::JSONError
  end  
end

signup_params = Signup::SignupParams.new(user_attributes: {email: 'testtenant@example.com', username: 'testtenant', role: :admin, password: '123456'}, account_attributes: {name: 'testtenant'})

create_tenant(signup_params)


# 2. create a bunch of services for those tenants

def create_service(tenant, name)
  begin
    tenant.services.create!(name: name)
  rescue ThreeScale::Core::APIClient::ConnectionError, ThreeScale::Core::APIClient::JSONError
  end
end

def mark_service_as_deleted(service)
  begin
    service.mark_as_deleted
  rescue ThreeScale::Core::APIClient::ConnectionError, ThreeScale::Core::APIClient::JSONError
  end
end

1000.times do |i|
  tenant = Account.tenants.last(2).to_a.sample
  create_service(tenant, "myy service #{i}")
  service = Service.last!
  mark_service_as_deleted(service) if service.id % 5 == 0
end

# create a member user for the 1st tenant
tenant_1st.users.create!(username: 'member', password: '123456', role: :member, email: 'member@example.com')
user = tenant_1st.users.where(role: :member).first!

# allow only some services for this user
user.admin_sections = [:services, :partners]
user.member_permission_service_ids = tenant_1st.services.pluck(:id).shuffle.first(tenant_1st.services.count / 3)
user.save!

# Create an access token to access the API
admin_token_value, member_token_value = [tenant_1st.admin_user, user].map { |u| AccessToken.create!(owner: u, permission: 'ro', scopes: ['account_management'], name: 'token').value }
```

2. Run the rails server. `UNICORN_WORKERS=8 bundle exec rails server -b 0.0.0.0`

3. Check the logs in another tab. `tail -f log/development.log`

4. In another tab, do the request: `curl -v  -X GET "http://provider-admin.3scale.localhost:3000/admin/api/services.json?access_token=889bfa550a2f2a3a07d112a4bdbbc9b2c686145e434ef48be7aaecc234b8fbc3"` . This is a local access token, no secrets exposed 😉

### DIFF
--- a/test/integration/admin/api/services_controller_test.rb
+++ b/test/integration/admin/api/services_controller_test.rb
@@ -72,7 +72,7 @@ class Admin::Api::ServicesControllerTest < ActionDispatch::IntegrationTest
       get admin_api_services_path(access_token: access_token_value, format: :json, per_page: 3, page: 2)
       assert_response :success
       response_service_ids = JSON.parse(response.body)['services'].map { |response_service| response_service.dig('service', 'id') }
-      assert_equal provider.accessible_services.order(:id).offset(3).limit(3).select(:id).map(&:id), response_service_ids
+      assert_equal provider.accessible_services.order(:id).offset(3).limit(3).pluck(:id), response_service_ids
     end
 
     test 'delete with api_key' do
--- a/test/integration/admin/api/services_controller_test.rb
+++ b/test/integration/admin/api/services_controller_test.rb
@@ -67,6 +67,14 @@ class Admin::Api::ServicesControllerTest < ActionDispatch::IntegrationTest
 
     attr_reader :provider, :service
 
+    test 'index can be paginated' do
+      FactoryBot.create_list(:simple_service, 5, account: provider)
+      get admin_api_services_path(access_token: access_token_value, format: :json, per_page: 3, page: 2)
+      assert_response :success
+      response_service_ids = JSON.parse(response.body)['services'].map { |response_service| response_service.dig('service', 'id') }
+      assert_equal provider.accessible_services.order(:id).offset(3).limit(3).select(:id).map(&:id), response_service_ids
+    end
+
     test 'delete with api_key' do
       assert_change(of: -> { service.reload.deleted? }, from: false, to: true) do
         delete admin_api_service_path(service, provider_key: provider.api_key)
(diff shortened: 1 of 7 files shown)


test/integration/admin/api/services_controller_test.rb
```diff
@@ -67,6 +67,14 @@ class TenantHostTest < ActionDispatch::IntegrationTest
 
     attr_reader :provider, :service
 
+    test 'index can be paginated' do
+      FactoryBot.create_list(:simple_service, 5, account: provider)
+      get admin_api_services_path(access_token: access_token_value, format: :json, per_page: 3, page: 2)
+      assert_response :success
+      response_service_ids = JSON.parse(response.body)['services'].map { |response_service| response_service.dig('service', 'id') }
+      assert_equal provider.accessible_services.order(:id).offset(3).limit(3).select(:id).map(&:id), response_service_ids
```
guicassolato, line 75:
```suggestion
      assert_equal provider.accessible_services.order(:id).offset(3).limit(3).pluck(:id), response_service_ids
```

guicassolato (approved):

test/integration/admin/api/services_controller_test.rb
Martouta: Unfortunately, this fails for Oracle 😓 
image

I will revert this commit in another PR.

PR #2160, opened 2020-09-10
[backend_client] Allow setting port from :host configuration

hallelujah:
So we can use a different port than 80 or 443 in backend

Fixes:

```
  Failed to open TCP connection to http://backend-listener-internal:80:80
(getaddrinfo: Name or service not known
```

* Update 3scale_client to 2.11 to set backend port

### DIFF
--- a/Gemfile.base
+++ b/Gemfile.base
@@ -80,7 +80,7 @@ gem 'json', '~> 1.8.1'
 
 gem 'mysql2', '~> 0.5.3'
 
-gem '3scale_client', '~> 2.6.1', require: false
+gem '3scale_client', '~> 2.11  ', require: false
 gem 'analytics-ruby', require: false
 
 group :development, :test do
--- a/Gemfile.base
+++ b/Gemfile.base
@@ -80,7 +80,7 @@ gem 'json', '~> 1.8.1'
 
 gem 'mysql2', '~> 0.5.3'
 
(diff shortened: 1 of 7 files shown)


app/lib/backend_client.rb
```diff
@@ -1,9 +1,30 @@
...
+
+    protected
+
+    def uri
+      URI(config[:url].presence || "#{scheme}://#{config[:host]}")
+    end
+
+    def scheme
+      config[:secure] ? 'https' : 'http'
+    end
```
hallelujah, line 28: Not memoizing because they are class methoids, and it is not thread-safe ... Though the content would not likely be changed BUT the object reference would ...

Gemfile.base
```diff
@@ -80,7 +80,7 @@ gem 'json', '~> 1.8.1'
 
 gem 'mysql2', '~> 0.5.3'
 
-gem '3scale_client', '~> 2.6.1', require: false
+gem '3scale_client', '~> 2.11  ', require: false
```
slopezz, line 83:
```suggestion
gem '3scale_client', '~> 2.11', require: false
```

slopezz (dismissed):
Without knowing too much ruby, a part from the indentation, the rest looks nice.

I will let @3scale/system-ruby team to validate it

Martouta (dismissed):

Martouta (approved):

PR #2178, opened 2020-09-21
Removing bubbles

damianpm:

Remove outdated inline help popovers ("bubbles")

https://issues.redhat.com/browse/THREESCALE-6040

**Verification steps** 

Check bubbles are removed from the following pages:


app/controllers/api/integrations_controller.rb
```diff
@@ -266,10 +260,6 @@ def deploying_hosted_proxy_key
     "#{current_account.id}/deploying_hosted"
   end
 
-  def update_onboarding_mapping_bubble
-    onboarding.bubble_update('mapping') if proxy_rules_added_for_last_method_metric?
-  end
-
   def proxy_rules_added_for_last_method_metric?
```
Martouta, line 263: Now this method is not used at all, is it? If that's the case, we should remove it 😄

app/controllers/api/services_controller.rb
```diff
@@ -57,8 +56,6 @@ def create
   def update
     if integration_settings_updater_service.call(service_attributes: service_params, proxy_attributes: proxy_params)
       flash[:notice] =  t('flash.services.update.notice', resource_type: product_or_service_type)
-      onboarding.bubble_update('api') if service_name_changed?
```
Martouta, line 60: For the method `service_name_changed?`, same as https://github.com/3scale/porta/pull/2178/files#r492712701

app/controllers/api/services_controller.rb
```diff
@@ -57,8 +56,6 @@ def create
...
-      onboarding.bubble_update('deployment') if integration_method_changed? && !integration_method_self_managed?
```
Martouta, line 61: For the method `integration_method_self_managed?`, same as https://github.com/3scale/porta/pull/2178/files#r492712701

app/views/api/integrations/apicast/shared/_proxy_endpoints.html.slim
```diff
@@ -9,11 +9,9 @@
                           placeholder: "https://api.#{parameterized_org_name_of_the_current_account}.com" },
             hint: apicast_endpoint_input_hint(service, environment: 'staging')
 
-  = help_bubble( 'proxy_endpoint_bubble', t('api.integrations.proxy.proxy_endpoint_help_html'))
```
Martouta, line 12: We still use the locale `proxy_endpoint_help_html`? if not, which seems so, then we should remove it:
https://github.com/3scale/porta/blob/7ddf357807fcca2400c674bdf7c7ad0a1aa364d7/config/locales/en.yml#L514-L517

Same for the rest of locales related to these bubbles.

didierofrivia:
@Martouta I think @damianpm is also forgetting about the :onboarding state machine.. but maybe this shouldn't be part of this PR, it will also need to do a migration that is disruptive.

didierofrivia:
@damianpm I'd also double check, there's still many views and helpers injecting onboarding-related class names and partials.

damianpm:
> @Martouta I think @damianpm is also forgetting about the :onboarding state machine.. but maybe this shouldn't be part of this PR, it will also need to do a migration that is disruptive.

In fact, I was to ask about the migration, I have no idea how to do it.
Do you think you can take care of that part @Martouta , when you have the time?

app/models/onboarding.rb
```diff
@@ -42,15 +24,7 @@ def self.null
   end
 
   def active?
```
josemigallas, line 26: Suggestion: remove method `active?` and use  `persisted?` instead.

Martouta:
@damianpm create a jira for the migration plz, because it is a blocking migration so i do not think that we will do it soon.

app/views/api/integrations/apicast/shared/_proxy_endpoints.html.slim
damianpm: You are right, removed!

josemigallas (approved):
LGTM

damianpm:
@Martouta do you think we can limit the scope of this PR, just removing the bubbles from the UI and the tests affected, and let for a subsequent PR the removal of the "onboarding" state machine in ruby?

Martouta:
yes @damianpm 👍 I will review in some minutes keeping that in mind 👍

Martouta (approved):
Ok to carry on in a following PR 💪 Good job! 🥇

PR #2179, opened 2020-09-23
[TopicIndex] get first 8M chars of sphinx_post_bodies

macejmic:
closes https://issues.redhat.com/browse/THREESCALE-5939

### DIFF
--- a/app/indices/topic_index.rb
+++ b/app/indices/topic_index.rb
@@ -23,7 +23,7 @@ module TopicIndex
   end
 
   def sphinx_post_bodies
-    posts.pluck(:body).join(' ')
+    posts.pluck(:body).join(' ').to_s[0...80000]
   end
 
   module ForPost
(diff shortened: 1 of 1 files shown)


app/indices/topic_index.rb
```diff
@@ -23,7 +23,7 @@ module TopicIndex
   end
 
   def sphinx_post_bodies
-    posts.pluck(:body).join(' ')
+    posts.pluck(:body).join(' ').to_s[0...80000]
```
hallelujah, line 26:
```suggestion
    posts.pluck(:body).join(' ').to_s[0...8_000_000]
```

hallelujah (approved):

hallelujah:
Note: finally Thinking Sphinx is making the query length customizable

https://github.com/pat/thinking-sphinx/commit/900d2e139338a5e9497a98b7788d1d6677f61192
It is only on develop branch though and we should investigate more later when we can upgrade and if any issue happens

PR #2180, opened 2020-09-23
THREESCALE-5952 Multiple Memcache servers

hallelujah:
Closes THREESCALE-5952

Making the plural in MEMCACHE_SERVERS actually effective ...

Basically you can add multiple servers for Dalli following https://github.com/petergoldstein/dalli#configuration

```
ActiveSupport::Cache.lookup_store :dalli_store, ['localhost:332211', 'localhost:112233'], compress: true                                                                            
# => #<ActiveSupport::Cache::DalliStore:0x00000000093b2208
 @data=#<Dalli::Client:0x00000000093b20f0 @options={:compress=>true}, @ring=nil, @servers=["localhost:332211", "localhost:112233"]>,
 @options={:compress=>true}>
```

## Proof

With:

```
MEMCACHE_SERVERS="localhost:11211,localhost:11311" rails c
Rails.cache.dalli
```

### Before
image

### After
image

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -134,7 +134,11 @@ module System
     # We don't want Rack::Cache to be used
     config.action_dispatch.rack_cache = false
 
-    config.cache_store = config_for(:cache_store)
+    args = config_for(:cache_store)
+    store_type = args.shift
+    options = args.extract_options!
+    servers = args.flat_map { |arg| arg.split(',') }
+    config.cache_store = [store_type, servers, options]
 
     # Configure the default encoding used in templates for Ruby 1.9.
     config.encoding = "utf-8"


hallelujah:
Closing it is stale. Not much attention

hallelujah:
Reopening this PR

guicassolato (approved):
I actually think Dalli's docs about configuration options a bit confusing. One might think that `servers` can be passed as an option (along with the various other actual options documented), but that's not the case.

You did it right! Servers have to be supplied as an array of strings, and _then_ options can be added if that's the case.

E.g., for 3 memcached servers, with options `namespace`, `username` and `password`, by setting in Porta's `config/cache_store.yaml` config file:

```yaml
# config/cache_store.yaml
production:
  - :dalli_store
  - server-a:11211
  - server-b:11211
  - server-c:11211
  - namespace: my_prefix
    username: john
    password: secret
```

Actually, the above would work even without this PR, but with the change, it's also covered for when the servers are set as a comma-separated list of servers, such as when using the `MEMCACHE_SERVERS` env var on OpenShift: https://github.com/3scale/porta/blob/24c9bd2e6242ab1c165b9494e5ee1379ddc75e6f/openshift/system/config/cache_store.yml#L3

PR #2185, opened 2020-09-30
Add Product and Backends index

josemigallas:

Products/Backends lists are being taken out from the Dashboard into their own pages. This is only the Rails part of it.

[APPDUX-545: Implement Ruby code for Product / Backend Apis](https://issues.redhat.com/browse/APPDUX-545)

### DIFF
--- a/app/controllers/api/services_controller.rb
+++ b/app/controllers/api/services_controller.rb
@@ -2,6 +2,7 @@
 
 class Api::ServicesController < Api::BaseController
   include ServiceDiscovery::ControllerMethods
+  include ThreeScale::Search::Helpers
 
   activate_menu :serviceadmin, :overview
 
@@ -14,6 +15,12 @@ class Api::ServicesController < Api::BaseController
 
   with_options only: %i[edit update settings usage_rules] do |actions|
     actions.sublayout 'api/service'
+  end
+
+  def index
+    activate_menu :dashboard
+    @services = current_user.accessible_services
+                            .paginate(pagination_params)
   end
 
   def show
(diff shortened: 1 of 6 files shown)


config/routes.rb
```diff
@@ -764,7 +764,7 @@ def call(env)
           resources :pricing_rules, :only => [:edit, :update, :destroy]
         end
 
-        resources :services, except: :index do
+        resources :services do
```
Martouta, line 767: This `services#index` also needs a test 😄

app/controllers/api/services_controller.rb
```diff
@@ -2,6 +2,7 @@
 
 class Api::ServicesController < Api::BaseController
   include ServiceDiscovery::ControllerMethods
+  include ThreeScale::Search::Helpers
```
Martouta, line 5: What for?

app/controllers/provider/admin/backend_apis_controller.rb
```diff
@@ -1,12 +1,20 @@
 # frozen_string_literal: true
 
 class Provider::Admin::BackendApisController < Provider::Admin::BaseController
-  before_action :find_backend_api, except: %i[new create]
+  include ThreeScale::Search::Helpers
```
Martouta, line 4: Same as https://github.com/3scale/porta/pull/2185/files#r498084981

config/routes.rb
josemigallas: Yes I'm not sure where to add it 🤷‍♂️

app/controllers/api/services_controller.rb
josemigallas: It's needed for pagination

PR #2187, opened 2020-10-02
Removing bubbles code from onboarding

damianpm:

This is a follow-up of https://github.com/3scale/porta/pull/2178
Removing remnant code from bubbles in onboarding model
The Onboarding model can't be totally removed because is needed for the wizard

https://issues.redhat.com/browse/THREESCALE-6040

### DIFF
--- a/app/controllers/provider/admin_controller.rb
+++ b/app/controllers/provider/admin_controller.rb
@@ -26,7 +26,7 @@ class Provider::AdminController < FrontendController
   end
 
   def onboarding_active?
-    onboarding.active?
+    onboarding.persisted?
   end
 
   def has_permission?
--- a/app/controllers/provider/admin_controller.rb
+++ b/app/controllers/provider/admin_controller.rb
@@ -6,7 +6,7 @@ class Provider::AdminController < FrontendController
 
     if start_onboarding_wizard
       redirect_to provider_admin_onboarding_wizard_root_path
-    elsif has_permission? && onboarding_active?
+    elsif has_permission? && onboarding.persisted?
       redirect_to provider_admin_dashboard_path
     else
       redirect_back_or_default provider_admin_dashboard_path
@@ -23,10 +23,6 @@ class Provider::AdminController < FrontendController
     false
   ensure
     analytics.track('Wizard Step', step: 'started') if started
-  end
-
-  def onboarding_active?
-    onboarding.persisted?
   end
 
   def has_permission?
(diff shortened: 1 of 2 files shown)


test/unit/models_test.rb
```diff
@@ -14,7 +14,7 @@ class ModelsTest < ActiveSupport::TestCase
       'Proxy' => %w[endpoint sandbox_endpoint], 'OIDCConfiguration' => %w[oidc_configurable_type], 'Invitation' => %w[token],
       'Settings' => Switches::SWITCHES.map { |switch| "#{switch}_switch" },
       'Invoice' => %w[pdf_file_name pdf_content_type state friendly_id fiscal_code vat_code currency creation_type], 'DeletedObject' => %w[owner_type object_type],
-      'Onboarding' => %w[wizard_state bubble_api_state bubble_metric_state bubble_deployment_state bubble_mapping_state bubble_limit_state],
+      'Onboarding' => %w[wizard_state],
```
damianpm, line 17: @Martouta [this test](https://github.com/3scale/porta/blob/master/test/unit/models_test.rb) is failing because of thie change.
Is it something we have to remove _after_ the migrations?

Martouta: It is normal. You should not change this test. What this is saying is "ignore these attributes, we are intentionally not validating the size of their strings in our code because they are not used"

Martouta (changes requested):
Things to change:
1. https://github.com/3scale/porta/pull/2187#discussion_r499734142
2. https://github.com/3scale/porta/pull/2178/files#r493248788

app/controllers/provider/admin_controller.rb
```diff
@@ -26,7 +26,7 @@ def start_onboarding_wizard
   end
 
   def onboarding_active?
```
Martouta, line 28: I think that we should remove this method too. Since the concept of 'active?' doesn't exist anymore. We should just use onboarding.persisted? instead

app/controllers/provider/admin_controller.rb
```diff
@@ -6,7 +6,7 @@ def show
 
     if start_onboarding_wizard
       redirect_to provider_admin_onboarding_wizard_root_path
-    elsif has_permission? && onboarding_active?
+    elsif has_permission? && onboarding.persisted?
```
damianpm, line 9: @Martouta  `onboarding_active?` deleted 😄

Martouta (approved):

PR #2190, opened 2020-10-06
🗑 Remove non-APIAP related code

josemigallas:

[From JIRA]
We kept a lot of Ruby code, views, etc, while we still had to support (non-APIAP) products (old services) in Saas. Once we rollout all Saas customers to APIAP, we can remove those pieces of code. This includes the rolling update api_as_product itself and related code.

[THREESCALE-3759: Remove (deprecated) non-APIAP code](https://issues.redhat.com/browse/THREESCALE-3759)

**Verification steps** 

- [ ] Provider
- [ ] Master


Martouta (dismissed):

Martouta:
So far so good 👍 Still very WIP according to the goal of the title 😄

test/models/backend_api_test.rb
```diff
@@ -21,7 +21,6 @@ def test_default_api_backend
     @account.stubs(:provider_can_use?).with(:apicast_v1).returns(true)
     @account.stubs(:provider_can_use?).with(:apicast_v2).returns(true)
     @account.expects(:provider_can_use?).with(:proxy_private_base_path).at_least_once.returns(false)
-    @account.expects(:provider_can_use?).with(:api_as_product).at_least_once.returns(false)
```
Martouta, line 24: I see that this used to be the scenario of non-APIAP, my question here is why do we keep this test? do we want APIAP (our current scenario) to behave the exact same way? if so, is this test repeated then?

app/controllers/api/integrations_controller.rb
```diff
@@ -4,7 +4,7 @@ class Api::IntegrationsController < Api::BaseController
   before_action :find_service
   before_action :find_proxy
   before_action :authorize
-  before_action :hide_for_apiap, only: :edit
+  before_action :hide, only: :edit
```
Martouta, line 7: I see that this 'hide' does `raise ActiveRecord::RecordNotFound`.
IMO, what we should do instead is removing this `hide` method, also the `edit`, and remove the action from `routes.rb`

app/models/backend_api.rb
```diff
@@ -38,7 +38,7 @@ class BackendApi < ApplicationRecord
 
   alias_attribute :api_backend, :private_endpoint
 
-  before_validation :set_private_endpoint, :set_port_private_endpoint
```
Martouta, line 41: Can we get rid of the `private_endpoint` attr then? Not saying that it has to be in this same PR 😄

app/views/provider/admin/onboarding/wizard/info/explain.html.slim
```diff
@@ -25,5 +25,5 @@ ol.explain.fa-ul
       strong> Your API
       ' returns the response to the customer  through the gateway.
 
-- next_path = apiap? ? new_provider_admin_onboarding_wizard_backend_api_path : new_provider_admin_onboarding_wizard_api_path
+- next_path = new_provider_admin_onboarding_wizard_backend_api_path
 = link_to "Got it! Lets add my API", next_path, class: "button button--next"
```
Martouta, line 29:
```suggestion
= link_to "Got it! Lets add my API", new_provider_admin_onboarding_wizard_backend_api_path, class: "button button--next"
```

app/views/provider/admin/dashboards/widgets/_service_navigations.html.slim
```diff
@@ -66,7 +66,7 @@ nav.DashboardNavigation id=widget.id
                                           icon_name: 'wrench'
 
           - else
-            = dashboard_navigation_link "Integrate this #{apiap? ? 'Product' : 'API'}",
+            = dashboard_navigation_link "Integrate this Product",
```
Martouta, line 69:
```suggestion
            = dashboard_navigation_link 'Integrate this Product',
```

app/models/account.rb
```diff
@@ -563,7 +563,7 @@ def provider_id_for_audits
   end
 
   def independent_mapping_rules_enabled?
-    provider_can_use?(:api_as_product) || provider_can_use?(:independent_mapping_rules)
+    true
   end
```
Martouta, line 567: Then do we even need this method at all?

app/controllers/api/integrations_controller.rb
```diff
@@ -264,11 +256,7 @@ def toggle_land_path
...
 
-  def render_edit_or_show(opts = {})
-    render (apiap? ? :show : :edit), opts
-  end
-
-  def redirect_to_edit_or_show
-    redirect_to (apiap? ? :show : edit_path)
+  def render_show(opts = {})
+    render :show, opts
   end
```
Martouta, line 261: We should get rid of this `render_show` method too 😄

app/controllers/api/integrations_controller.rb
```diff
@@ -251,7 +243,7 @@ def proxy_params
       basic_fields << :jwt_claim_with_client_id_type
     end
 
-    basic_fields << { backend_api_configs_attributes: %i[_destroy id path] } if provider_can_use?(:api_as_product)
+    basic_fields << { backend_api_configs_attributes: %i[_destroy id path] }
```
Martouta, line 246: we can have it in the initialization of `basic_fields` instead, right?

features/old/api/settings.feature
```diff
@@ -10,16 +10,14 @@ Feature: API Settings
 
   Scenario: Referrer filtering on backend v2
     Given provider "foo.3scale.localhost" uses backend v2 in his default service
-      And I have rolling updates "api_as_product" disabled
-      And I log in as provider "foo.3scale.localhost"
+      And I log in as provider "foo.example.com"
```
Martouta, line 13: 👀 
```suggestion
      And I log in as provider "foo.3scale.localhost"
```

app/models/account.rb
josemigallas: That triggers a LOT of scary changes but yes 😱

app/models/backend_api.rb
josemigallas: I'll add a TODO just in case

features/old/api/settings.feature
josemigallas: goot catch

test/models/backend_api_test.rb
josemigallas: I see.. well the tests is for BackendAPI so it should only apply to an APIAP scenario, right? I think this line doesn't do anything since the test passes anyway.

app/controllers/api/integrations_controller.rb
josemigallas: This looks complex, could we leave it for a future PR?

Martouta: Sure 👍

app/controllers/api/proxy_rules_controller.rb
```diff
@@ -44,7 +44,7 @@ def destroy
   private
 
   def authorize!
-    current_account.independent_mapping_rules_enabled? || raise(CanCan::AccessDenied)
+    true || raise(CanCan::AccessDenied)
   end
```
Martouta, line 48: `true or ...` is always `true`, so let's remove this method too 😄

app/controllers/api/integrations_controller.rb
Martouta: But add a 'TODO' comment then plz 🙈

test/functional/api/services_controller_test.rb
```diff
@@ -0,0 +1,106 @@
...
+
+    member.member_permissions << permission
+
+    @request.host = @provider.domain
+
+    login_as member
+  end
+
+  def test_load_and_authorize_resource
+    get :show, id: FactoryBot.create(:service).id
```
Martouta, line 19: these functional tests are deprecated... plz use integration tests instead

josemigallas: I don't know how this ended up here..

app/controllers/api/proxy_rules_controller.rb
josemigallas: I noticed that raising an error `before_action` would mean this controller is no longer useful, am I wrong?

Martouta: yes, but the thing is that now it will never raise, for the lazy evaluation.
`true or raise exception` => `never raise exception`

josemigallas: Of course, you're right

test/integration/api/services_controller_test.rb
```diff
@@ -257,9 +246,7 @@ def setup
       assert_equal 0, Service.last.backend_api_configs.count
     end
 
-    test 'should create the default Backend if API as Product is disabled' do
-      Account.any_instance.stubs(:provider_can_use?).with(:api_as_product).returns(false).at_least_once
-
+    test 'should create the default Backend' do
```
Martouta, line 249: Remove this test, no? 😕

app/views/api/services/settings.html.slim
```diff
@@ -1,21 +1,6 @@
...
-            p.inline-hints= @service.service_plans.first.class.human_attribute_name 'default_plan_hint'
-          - else
-            p.inline-hints Please create a service plan first.
-
 = semantic_form_for @service, :url => admin_service_path(@service) do |form|
-  = render :partial => 'api/services/forms/settings', :locals => { :form => form }
+  = render :partial => 'api/services/forms/integration_settings', :locals => {:form => form}
   = form.buttons do
-    = form.commit_button
+    = form.commit_button "Update Product", button_html: { name: 'update_settings' }
```
Martouta, line 6:
```suggestion
    = form.commit_button 'Update Product', button_html: { name: 'update_settings' }
```

app/views/api/services/settings.html.slim
```diff
@@ -1,21 +1,6 @@
...
-          = select_tag :id, options_from_collection_for_select(@service.service_plans.not_custom, "id", "name", lambda { |p| p.master? }),
-            :include_blank => true, :id => "default_plan", disabled: !service_plans_exists
-          - if service_plans_exists
-            p.inline-hints= @service.service_plans.first.class.human_attribute_name 'default_plan_hint'
-          - else
-            p.inline-hints Please create a service plan first.
-
 = semantic_form_for @service, :url => admin_service_path(@service) do |form|
-  = render :partial => 'api/services/forms/settings', :locals => { :form => form }
+  = render :partial => 'api/services/forms/integration_settings', :locals => {:form => form}
```
Martouta, line 4:
```suggestion
  = render partial: 'api/services/forms/integration_settings', locals: {form: form}
```

app/models/proxy.rb
```diff
@@ -240,30 +236,14 @@ def default_production_endpoint
...
   def deployment_strategy
-    strategy = case deployment_option
-               when 'self_managed' then SelfManagedAPIcast
-               when 'hosted' then HostedAPIcast
-               end
-
-    strategy.try!(:new, self)
-  end
-
-  def deployment_strategy_apiap
```
guicassolato, line 263: This method was introduced because it was needed in a view https://github.com/3scale/porta/commit/f78c868fb83249cba4556ea0cfb6b6e15bed0818. We shouldn't now promote it entirely to be the new behaviour of `Proxy#deployment_strategy`. Instead, I'd ditch `#deployment_strategy_apiap`, keep `#deployment_strategy` the way it is right now (with the case condition) and in the view revert it to how it was used before:

```ruby
proxy.default_staging_endpoint unless proxy.self_managed?
```

app/views/api/integrations/apicast/shared/_proxy_endpoints.html.slim
```diff
@@ -4,14 +4,14 @@
 = f.inputs "API gateway", id: 'proxy-endpoints', data: {apicast_custom_urls: apicast_custom_urls?}, class: "apicast-only-settings #{'hidden' if is_service_mesh}", disabled: is_service_mesh do
   = f.input :sandbox_endpoint, label: "Staging Public Base URL",
             input_html: { class: 'proxy-endpoint',
-                          data: { default: (proxy.default_staging_endpoint_apiap ) },
+                          data: { default: (proxy.default_staging_endpoint unless proxy.self_managed?) },
```
guicassolato, line 7: I was thinking... This will cause the default to be null when the service is self-managed, what currently doesn't happen with all providers on APIAP.

Another option could be renaming `Proxy#deployment_strategy_apiap` to `Proxy#hosted_service_deployment_strategy` and here use `proxy.hosted_service_deployment_strategy.default_staging_endpoint` (and proxy.hosted_service_deployment_strategy.default_production_endpoint below)

test/unit/tasks/backend_api_test.rb
```diff
@@ -2,30 +2,6 @@
 
 module Tasks
   class BackendApiTest < ActiveSupport::TestCase
-    class DestroyOrphans < ActiveSupport::TestCase
```
Martouta, line 5: I would feel safer extracting the changes/removals in `lib/tasks/backend_api.rake` and `test/unit/tasks/backend_api_test.rb` to its own PR 🤔

josemigallas: https://github.com/3scale/porta/pull/2247

app/controllers/api/integrations_controller.rb
```diff
@@ -32,15 +33,7 @@ def update
...
-      return redirect_to admin_service_integration_path(@service) if apiap?
-
-      if @proxy.send_api_test_request!
-        api_backend = @proxy.api_backend
-        done_step(:api_sandbox_traffic) if api_backend.present? && ApiClassificationService.test(api_backend).real_api?
-        return redirect_to edit_path
-      end
-
-      render :edit
+      redirect_to admin_service_integration_path(@service)
```
Martouta, line 36: Since `@service`and `@proxy` are mandatory and found at the beginning of the request... they should be used by `attr_reader :service, :proxy` inside this controller instead of getting it by the instance variable

app/models/proxy.rb
```diff
@@ -197,9 +197,9 @@ def default_staging_endpoint; end
 
     def default_production_endpoint; end
 
-    def default_staging_endpoint_apiap; end
+    def hosted_default_staging_endpoint; end
```
guicassolato, line 200: No need to keep this.

IMO, we could now do `hosted_service_deployment_strategy.default_staging_endpoint`.

app/models/proxy.rb
```diff
@@ -197,9 +197,9 @@ def default_staging_endpoint; end
...
 
-    def default_production_endpoint_apiap; end
+    def hosted_default_production_endpoint; end
```
guicassolato, line 202: https://github.com/3scale/porta/pull/2190/files#r519862581

app/models/proxy.rb
```diff
@@ -241,11 +241,11 @@ def default_production_endpoint
       generate(production_endpoint)
     end
 
-    def default_staging_endpoint_apiap
+    def hosted_default_staging_endpoint
```
guicassolato, line 244: https://github.com/3scale/porta/pull/2190/files#r519862581

app/models/proxy.rb
```diff
@@ -241,11 +241,11 @@ def default_production_endpoint
...
       default_staging_endpoint
     end
 
-    def default_production_endpoint_apiap
+    def hosted_default_production_endpoint
```
guicassolato, line 248: https://github.com/3scale/porta/pull/2190/files#r519862581

app/models/proxy.rb
```diff
@@ -476,8 +476,8 @@ def force_apicast_version
   delegate :default_production_endpoint, :default_staging_endpoint,
            to: :deployment_strategy, allow_nil: true
 
-  delegate :default_production_endpoint_apiap, :default_staging_endpoint_apiap,
-           to: :deployment_strategy_apiap, allow_nil: true
+  delegate :hosted_default_production_endpoint, :hosted_default_staging_endpoint,
+           to: :hosted_service_deployment_strategy, allow_nil: true
```
guicassolato, line 480: https://github.com/3scale/porta/pull/2190/files#r519862581

app/views/api/integrations/apicast/shared/_proxy_endpoints.html.slim
```diff
@@ -4,14 +4,14 @@
 = f.inputs "API gateway", id: 'proxy-endpoints', data: {apicast_custom_urls: apicast_custom_urls?}, class: "apicast-only-settings #{'hidden' if is_service_mesh}", disabled: is_service_mesh do
   = f.input :sandbox_endpoint, label: "Staging Public Base URL",
             input_html: { class: 'proxy-endpoint',
-                          data: { default: (proxy.default_staging_endpoint_apiap ) },
+                          data: { default: (proxy.default_hosted_staging_endpoint ) },
```
guicassolato, line 7:
```suggestion
                          data: { default: (proxy.hosted_service_deployment_strategy.default_staging_endpoint ) },
```

app/views/api/integrations/apicast/shared/_proxy_endpoints.html.slim
```diff
@@ -4,14 +4,14 @@
...
                           readonly: apicast_urls_readonly?,
                           placeholder: "https://api.#{parameterized_org_name_of_the_current_account}.com" },
             hint: apicast_endpoint_input_hint(service, environment: 'staging')
 
   = f.input :endpoint, label: "Production Public Base URL",
             input_html: { class: 'proxy-endpoint',
-                          data: { default: (proxy.default_production_endpoint_apiap ) },
+                          data: { default: (proxy.default_hosted_production_endpoint ) },
```
guicassolato, line 14:
```suggestion
                          data: { default: (proxy.proxy.hosted_service_deployment_strategy.default_production_endpoint) },
```

app/models/proxy.rb
guicassolato: Alternatively, you can do:

```ruby
delegate :default_staging_endpoint, :default_production_endpoint,
  to: :hosted_service_deployment_strategy,
  prefix: 'hosted'
```

and in the view call `proxy.hosted_default_staging_endpoint` (instead of `proxy.hosted_service_deployment_strategy.default_staging_endpoint`).

app/controllers/api/integrations_controller.rb
```diff
@@ -204,18 +199,18 @@ def message_bus?(proxy)
     proxy.oidc? && ZyncWorker.config.message_bus
   end
 
-  def hide_for_apiap
-    raise ActiveRecord::RecordNotFound if apiap?
+  def hide
```
didierofrivia, line 202: I think this is not needed anymore, you might want to remove the method entirely,

didierofrivia: Ok, I didn't see the #TODO and comments

app/javascript/src/NewService/components/NewServiceForm.jsx
```diff
@@ -29,22 +28,19 @@ const NewServiceForm = (props: Props) => {
...
-    : <ServiceDiscoveryForm formActionPath={providerAdminServiceDiscoveryServicesPath} apiap={apiap} setLoadingProjects={setLoadingProjects} />
-
-  const title = apiap ? 'New Product' : 'New API'
+    ? <ServiceManualForm template={template} formActionPath={adminServicesPath} backendApis={backendApis} />
+    : <ServiceDiscoveryForm formActionPath={providerAdminServiceDiscoveryServicesPath} setLoadingProjects={setLoadingProjects} />
 
   return (
     <React.Fragment>
-      <h1>{title}</h1>
+      <h1>New Product</h1>
```
didierofrivia, line 36: Right, we are not using I18N here, it was just for "Product" or "Service", no?

app/lib/apicast/curl_command_builder.rb
```diff
@@ -111,7 +111,7 @@ def service
       end
     end
 
-    def initialize(proxy, environment: :staging, build_from_config: -> { apiap? }, use_api_test_path: -> { !apiap? })
+    def initialize(proxy, environment: :staging, build_from_config: -> { true }, use_api_test_path: -> { false })
```
didierofrivia, line 114:
```suggestion
    def initialize(proxy, environment: :staging, build_from_config: true, use_api_test_path: false)
```

features/api/services/settings.feature
```diff
@@ -23,81 +21,19 @@ Feature: Integration Settings
...
-    And I should see field "Production Public Base URL" enabled
-
-  @javascript
-  Scenario: Production and Staging URL are grayed out when 3scale-managed APIcast
-    Given I log in as provider "foo.3scale.localhost"
-    And I go to the settings page for service "API" of provider "foo.3scale.localhost"
-
-    When I click on the label "APIcast"
-    Then I should see field "Staging Public Base URL" disabled
-    And I should see field "Production Public Base URL" disabled
```
didierofrivia, line 103: I'm not up-to-date with the definition, but all these tests are removed because now they belong to another section? Because Proxy integrations are agnostic to APIAP or non-APIP, no?

app/lib/apicast/curl_command_builder.rb
didierofrivia: https://github.com/3scale/porta/pull/2254 forget it

app/javascript/src/NewService/components/NewServiceForm.jsx
josemigallas: Correct

features/api/services/settings.feature
josemigallas: AFAIK integration settings are not used in APIAP 🤔

app/models/account.rb
```diff
@@ -562,10 +562,6 @@ def provider_id_for_audits
     end
   end
 
-  def independent_mapping_rules_enabled?
-    provider_can_use?(:api_as_product) || provider_can_use?(:independent_mapping_rules)
```
Martouta, line 566: I guess that we can also get rid of these 2 lines:
https://github.com/3scale/porta/blob/99901577f26e32c57a086f7c3c2b0e0d9240ca96/config/examples/rolling_updates.yml#L19-L20

app/models/backend_api.rb
```diff
@@ -107,11 +108,6 @@ def schedule_deletion
...
   end
 
-  def set_private_endpoint
-    return if account.provider_can_use?(:api_as_product)
-    self.private_endpoint ||= default_api_backend
-  end
-
   def set_port_private_endpoint
     Proxy::PortGenerator.new(self).call(:private_endpoint)
   end
```
Martouta, line 113: We do not need this anymore either because now the field 'private_endpoint' is empty ('blank') and the 1st thing that this line does is just returning in this scenario 😄 

https://github.com/3scale/porta/blob/99901577f26e32c57a086f7c3c2b0e0d9240ca96/app/models/proxy.rb#L689-L696

app/services/proxy_test_service.rb
```diff
@@ -103,7 +103,6 @@ def api_test_path
   end
 
   def backend_api_config_path
-    return unless proxy.provider_can_use?(:api_as_product)
     proxy.backend_api_configs.first&.path
```
Martouta, line 106: We should at least ensure that it is always the same one by the order...
```suggestion
    proxy.backend_api_configs.order(:id).first&.path
```

app/presenters/api/integrations_show_presenter.rb
```diff
@@ -53,29 +53,12 @@ def environments_have_same_config?
...
-    provider_can_use? :api_as_product
-  end
-
   def apicast_config_ready?
-    if apiap?
-      proxy.service.backend_apis.any?
-    else
-      any_sandbox_configs?
-    end
+    proxy.service.backend_apis.any?
```
Martouta, line 57: I will accept this for now because it is not your fault and the PR would get too big if I start complaining about these things 😂

app/services/service_creator.rb
```diff
@@ -37,12 +37,8 @@ def save!(params)
   end
 
   def save_default_backend_api(params)
-    if provider_can_use?(:api_as_product)
-      return true if %i[path private_endpoint].none? { |key| params.key?(key) }
-      backend_api_proxy.update!(params)
-    else
-      backend_api_proxy.update!(private_endpoint: BackendApi.default_api_backend)
-    end
+    return true if %i[path private_endpoint].none? { |key| params.key?(key) }
```
Martouta, line 40: If this comment is correct: https://github.com/3scale/porta/pull/2190/files#diff-437d3143cf330769c5e83c1d9385f74182d8ef63f03d4024af7ee6b399a6af0dR41
then it means that this field is not used anywhere, then:
```suggestion
    return true params.key?(:path)
```

app/lib/logic/rolling_updates.rb
```diff
@@ -299,16 +299,6 @@ def enabled?
         end
       end
 
-      class ApiAsProduct < Base
```
Martouta, line 302: we can also remove this RU, i think:
https://github.com/3scale/porta/blob/183bab3a90bbdf747297a9ec4a6183f0ca199ffe/app/lib/logic/rolling_updates.rb#L132-L136

app/controllers/api/services_controller.rb
```diff
@@ -55,28 +54,28 @@ def create
...
 
   def update
     if integration_settings_updater_service.call(service_attributes: service_params, proxy_attributes: proxy_params)
-      flash[:notice] =  t('flash.services.update.notice', resource_type: product_or_service_type)
+      flash[:notice] =  t('flash.services.update.notice', resource_type: 'Product')
       redirect_back_or_to :action => :settings
     else
-      flash.now[:error] = t('flash.services.update.error', resource_type: product_or_service_type)
-      render action: params[:update_settings].present? ? settings_page : :edit # edit page is only page with free form fields. other forms are less probable to have errors
+      flash.now[:error] = t('flash.services.update.error', resource_type: 'Product')
```
Martouta, line 71: if it is always the same resource type, why we do not simply add the name directly in the string of the locale instead of passing the variable 'Product' from everywhere?

app/models/account.rb
josemigallas: #2274

app/models/backend_api.rb
josemigallas: Does it depends directly on this PR or can it be extracted into a parallel one?

Martouta: many things of this PR can be extracted to other PRs 😄 like this 👍🏼 😉

app/controllers/api/services_controller.rb
josemigallas: Yeah, good catch 💪

app/lib/logic/rolling_updates.rb
josemigallas: #2274

app/services/proxy_test_service.rb
Martouta: unless there is only 1

Martouta: my bad 😄

josemigallas: #2280

app/models/backend_api.rb
guicassolato: I don't see why we'd remove the `set_port_private_endpoint` callback. I agree we no longer need the `set_private_endpoint` one, but, still, this does NOT mean `BackendApi`'s `private_endpoint` will be empty/`nil`. It can still be set manually, like in https://github.com/3scale/porta/blob/2d3d1f33ffe144b8e4f0006fdbf22e988341fae3/app/models/account/provider_methods.rb#L346

Perhaps there was a confusion between `BackendApi`'s `private_endpoint` and `Proxy`'s `private_endpoint`?

PR #2191, opened 2020-10-06
[WIP] Removing Signup requires a description of intended use checkbox

HemaHG:

Removing **Signup requires a description of intended use** checkbox

[THREESCALE-885](https://issues.redhat.com/browse/THREESCALE-885)


thomasmaas:
Do we want to remove all mentions in code of `intentions_required` and remove field from db ?

PR #2194, opened 2020-10-07
New Dashboard widgets for 2.10 ER1

josemigallas:
⚠️ This is not to be merged yet, we're using this branch as base.

We're removing the tabs in favor of two smaller widgets showing the top 5 last updated products and backends. Also removing the list of services from the Context Switcher and replacing it for a link to a Products and a Backend APIs index pages.

[Address UI scalability issues in API Management Service](https://issues.redhat.com/browse/APPDUX-462)

**Verification steps** 

* Provider dashboard shows the 2 new widgets
* Products/Backend index pages can be visited:
  * Searchbar works
  * Pagination works


app/decorators/application_decorator.rb
```diff
@@ -15,10 +15,11 @@ def self.collection_decorator_class
 
   private
 
-  API_KEYS = %w[id name system_name].freeze
+  API_KEYS = %w[id name system_name links apps_count backends_count unread_alerts_count products_count].freeze
```
Martouta, line 18: We spoke about this 😄 Improvement in process ⚠️  👍

app/decorators/backend_api_decorator.rb
```diff
@@ -18,4 +18,18 @@ def as_json(options = {})
...
+      { name: 'Edit', path: h.edit_provider_admin_backend_api_path(object) },
+      { name: 'Overview', path: h.provider_admin_backend_api_path(object) },
+      { name: 'Analytics', path: h.provider_admin_backend_api_stats_usage_path(object) },
+      { name: 'Methods & Metrics', path: h.provider_admin_backend_api_metrics_path(object) },
+      { name: 'Mapping Rules', path: h.provider_admin_backend_api_mapping_rules_path(object) },
+    ]
+  end
+
+  def products_count
+    object.services.size
```
Martouta, line 33: I don't know how this is done internally, but if it does the following, it is scary as hell 😱
1. backend_api left outer join backend_api_config left outer join services
2. count distinct services.id where backend_api.id is this one

instead, we should do this:
1. backend_api left outer join backend_api_config
2. count distinct backend_api_config.service_id where backend_api.id is this one

app/decorators/service_decorator.rb
```diff
@@ -56,4 +56,30 @@ def as_json(options = {})
...
+      { name: 'Integration', path: h.path_to_service(object) },
+    ]
+  end
+
+  def apps_count
+    cinstances.size
+  end
+
+  def backends_count
+    backend_apis.size
```
Martouta, line 76: inverse situation to https://github.com/3scale/porta/pull/2194/files#r501110645

app/controllers/provider/admin/dashboards_controller.rb
```diff
@@ -14,6 +14,16 @@ def show
     @services           = current_user.accessible_services
     @messages_presenter = current_presenter
     @unread_messages_presenter = unread_messages_presenter
+
+    @widget_products = @services.order(updated_at: :desc)
+                                .decorate
+                                .take(5)
```
Martouta, line 20: if we use `paginate` instead as with the other controllers, how many would there be by default? is it worth it? in that case, it would be easier to DRY this up.

app/controllers/api/services_controller.rb
```diff
@@ -16,6 +17,15 @@ class Api::ServicesController < Api::BaseController
     actions.sublayout 'api/service'
   end
 
+  def index
+    activate_menu :dashboard
+    @services = current_user.accessible_services
+                            .order(updated_at: :desc)
+                            .paginate(pagination_params)
+                            .decorate
+                            .to_json(only: %i[name updated_at id], methods: %i[links apps_count backends_count unread_alerts_count])
```
Martouta, line 26: we should definitely DRY this 😄

app/controllers/provider/admin/backend_apis_controller.rb
```diff
@@ -1,11 +1,22 @@
...
   activate_menu :backend_api, :overview
   layout 'provider'
 
+  def index
+    activate_menu :dashboard
+    @backend_apis = current_account.backend_apis
+                                   .order(updated_at: :desc)
+                                   .paginate(pagination_params)
+                                   .decorate
+                                   .to_json(only: %i[name updated_at id private_url], methods: %i[links products_count])
```
Martouta, line 17: let's DRY this 😄

app/controllers/provider/admin/dashboards_controller.rb
Martouta: https://github.com/3scale/porta/pull/2194/files#r501113662

app/controllers/provider/admin/dashboards_controller.rb
```diff
@@ -14,6 +14,16 @@ def show
...
+                                .to_json(only: %i[name updated_at id], methods: %i[links])
+    @widget_backends = current_account.backend_apis.includes(:services)
+                                                    .order(updated_at: :desc)
+                                                    .decorate
+                                                    .take(5)
+                                                    .to_json(only: %i[name updated_at id], methods: %i[links])
```
Martouta, line 26: https://github.com/3scale/porta/pull/2194/files#r501113662

app/controllers/provider/admin/backend_apis_controller.rb
Martouta, line 17: https://github.com/3scale/porta/pull/2194/files#r501113662

app/controllers/provider/admin/dashboards_controller.rb
```diff
@@ -14,6 +14,16 @@ def show
     @services           = current_user.accessible_services
     @messages_presenter = current_presenter
     @unread_messages_presenter = unread_messages_presenter
+
+    @widget_products = @services.order(updated_at: :desc)
+                                .decorate
+                                .take(5)
+                                .to_json(only: %i[name updated_at id], methods: %i[links])
```
Martouta, line 21: https://github.com/3scale/porta/pull/2194/files#r501113662

app/decorators/service_decorator.rb
```diff
@@ -56,4 +56,30 @@ def as_json(options = {})
...
+      { name: 'Overview', path: h.admin_service_path(object) },
+      { name: 'Analytics', path: h.admin_service_stats_usage_path(object) },
+      { name: 'Applications', path: h.admin_service_applications_path(object) },
+      { name: 'ActiveDocs', path: h.admin_service_api_docs_path(object) },
+      { name: 'Integration', path: h.path_to_service(object) },
+    ]
+  end
+
+  def apps_count
+    cinstances.size
```
Martouta, line 72: maybe a `sum` of all the `contracts_count` field in the DB for all the `application_plans` of this service.
that would be more efficient.

app/decorators/backend_api_decorator.rb
josemigallas: What about calling `service.backend_api_configs` like it's done in `Account > Backends`?
https://github.com/3scale/porta/blob/ed7ad0d57ab2c8de7d4241afd5008f89de31fc59/app/controllers/api/backend_usages_controller.rb#L16

app/decorators/service_decorator.rb
josemigallas: Would `service.backend_api_configs` be valid here, like it's in `Account > Backends`?
https://github.com/3scale/porta/blob/ed7ad0d57ab2c8de7d4241afd5008f89de31fc59/app/controllers/api/backend_usages_controller.rb#L16

app/controllers/provider/admin/dashboards_controller.rb
josemigallas: There won't be pagination in the dashboard, only the a list of 5 items.

app/controllers/api/services_controller.rb
josemigallas: > TLDR: similar code but no repetition.

Well, I think DRY is not always the best solution: 2 equal lines of code don't necessarily have to be repeated.
In this case and IMO the scopes are different enough (different models, different controllers) so that DRY is not applicable.
Also, today we order by the same column (updated_at), tomorrow we don't know, maybe hits.
As for `paginate`, I wouldn't obscure it behind some decorator but rather leave it explicit for a quicker reading (plus, we'd need to pass the params).
Finally, keep in mind that each to_json passes its own fields and methods.

app/decorators/backend_api_decorator.rb
guicassolato: I see what you're suggesting and indeed it would be an improvement:

```ruby
object = BackendApi.find(2)

object.services.count
# SELECT COUNT(*) FROM `services` INNER JOIN `backend_api_configs` ON `services`.`id` = `backend_api_configs`.`service_id` WHERE `backend_api_configs`.`backend_api_id` = 2

object.backend_api_configs.select(:service_id).distinct.count
# SELECT DISTINCT COUNT(DISTINCT `backend_api_configs`.`service_id`) FROM `backend_api_configs` WHERE `backend_api_configs`.`backend_api_id` = 2
```

However, I wonder if it's worth the effort. The proper way of optimising this would be by caching this count, but we don't have the time to do this now. Perhaps we can move on with this the way it is and consider another round of improvements later, with caching and all?

app/controllers/api/services_controller.rb
```diff
@@ -17,11 +17,11 @@ class Api::ServicesController < Api::BaseController
 
   def index
     activate_menu :dashboard
-    @services = current_user.accessible_services
-                            .order(updated_at: :desc)
-                            .paginate(pagination_params)
-                            .decorate
-                            .to_json(only: %i[name updated_at id system_name], methods: %i[links apps_count backends_count unread_alerts_count])
+    @raw_services = current_user.accessible_services
```
Martouta, line 20: If you are not going to use this variable outside this method, better make it a local variable of the method rather than an instance variable of the class that can be used all over the controller, the views and the helpers.

```suggestion
    raw_services = current_user.accessible_services
```

Martouta: I see that you use it for the counter... then better `@total_services_count = ...`
The less 'global variables' with less 'global info', the better.

app/controllers/api/services_controller.rb
```diff
@@ -14,6 +15,15 @@ class Api::ServicesController < Api::BaseController
     actions.sublayout 'api/service'
   end
 
+  def index
```
Martouta, line 18: I see many tips (in the logs) to improve the performance.

image

app/javascript/src/BackendApis/components/IndexPage.js
```diff
@@ -0,0 +1,163 @@
...
+        </Level>
+        <p className="api-table-description">
+          Explore and manage all your internal APIs.
+        </p>
+        <Divider/>
+        <Toolbar id="top-toolbar" className="pf-c-toolbar">
+          <div className="pf-c-toolbar__content">
+            <ToolbarItem>
+              <InputGroup className="api-table-search">
+                <TextInput placeholder="Find a Backend" name="findBackend" id="findProduct" type="search" aria-label="Find a product" />
```
Martouta, line 134:
```suggestion
                <TextInput placeholder="Find a Backend" name="findBackend" id="findProduct" type="search" aria-label="Find a backend" />
```

🙈

thomasmaas:
Looking good! 

The only thing we didn't really think about is that the current product shows up as the title of the context switcher, which is a bit weird as individual products/backends are no longer part of it.
Screenshot 2020-10-14 at 14 20 15

app/controllers/api/services_controller.rb
Martouta: For a following PR 👍

app/controllers/api/services_controller.rb
Martouta: For a following PR 👍

app/decorators/application_decorator.rb
```diff
@@ -12,4 +12,8 @@ def api_selector_api_link
   def self.collection_decorator_class
     PaginatingDecorator
   end
+
+  def updated_at
+    object.updated_at.to_s :long if object.updated_at?
```
Martouta, line 17: Where is all this called/used from? 🤔

app/decorators/backend_api_decorator.rb
Martouta: Yep, makes sense all of it! 💪  👍

app/decorators/service_decorator.rb
Martouta: similarly to https://github.com/3scale/porta/pull/2194/files#r504635406, next round for improvements 👍

app/models/backend_api.rb
```diff
@@ -1,9 +1,17 @@
...
   include SystemName
   include ProxyConfigAffectingChanges::ModelExtension
 
+  self.allowed_search_scopes = %i[query]
+
+  scope :by_query, ->(query) do
+    options = {ids_only: true, per_page: 1_000_000, star: true, ignore_scopes: true, with: { }}
+    where(id: search(ThinkingSphinx::Query.escape(query), options))
+  end
+
```
Martouta, line 14: As I mentioned in the description of https://github.com/3scale/porta/pull/2204, we'll later have to polish all this and add tests 👍

test/integration/api/services_controller_test.rb
```diff
@@ -12,6 +12,11 @@ def setup
 
   attr_reader :provider, :service
 
+  test 'index' do
+    get admin_services_path
+    assert_response :success
+  end
+
```
Martouta, line 19: later we will have to improve this test

test/integration/provider/admin/backend_apis_controller_test.rb
```diff
@@ -13,6 +13,11 @@ class Provider::Admin::BackendApisControllerTest < ActionDispatch::IntegrationTe
 
   attr_reader :provider
 
+  test '#index' do
+    get provider_admin_backend_apis_path
+    assert_response :success
+  end
```
Martouta, line 19: later we will have to improve this test

Martouta (dismissed):
Great work! 👏 It can be polished but let's consider it for another round knowing that this feature is for today 😉

thomasmaas (dismissed):
approved based on click testing in preview. search wasn't working yet due to indexing.

cntlsn:
looking good, great work!
a couple of things I noticed while clicking around in preview:
* In the two widgets in the Dashboard, clicking on a Product or Backend name wouldn't lead you to the relative Product or Backend page
* In both the Product Dashboard widget and the Product page, when clicking on the 'Edit' item of the kebab menu the page reloads instead of leading you to the Edit Product page.

damianpm (changes requested):
There are two issues to fix:
1.- Icons are not equal in Dashboard
icons
2.- There is an error in `isPlainButtonAction`, this may be the cause you can't click on API names:
Screenshot 2020-10-14 at 15 05 51

app/decorators/application_decorator.rb
josemigallas: This parses `updated_at` to a human friendly format, depending on the user's time zone when calling `to_json`.

josemigallas:
> There are two issues to fix:
> 1.- Icons are not equal in Dashboard
> icons
> 2.- There is an error in `isPlainButtonAction`, this may be the cause you can't click on API names:
> Screenshot 2020-10-14 at 15 05 51

1. Good catch, fixed 👍 
2. I added a comment there, it's supposed to be valid according to PF documentation.. but it might has been removed recently. It shouldn't be a problem, we can take a look at it afterwards.

thomasmaas:
@josemigallas I'd rather fix the broken links before merging, so we can deploy master to saas if we have to.

app/javascript/src/Dashboard/components/APIDataListItem.jsx
```diff
@@ -0,0 +1,90 @@
...
+              {updatedAt}
+            </DataListCell>
+          ]}
+        />
+        <DataListAction
+          aria-labelledby="multi-actions-item1 multi-actions-action1"
+          id="actions-menu"
+          aria-label="Actions"
+          // TODO: throwing a warning, does this prop really exist?
+          isPlainButtonAction
```
damianpm, line 68: I think we can simply remove this prop. According to PF docs is not required and I don't see any difference in the rendered output
pf-docs
```suggestion
```

josemigallas: OK sounds good to me!

damianpm (dismissed):
LGTM, please run tests when ready

Martouta (dismissed):

thomasmaas (approved):
All mentioned issues looking good now in preview

Martouta (approved):

features/step_definitions/applications/applications_steps.rb
```diff
@@ -168,7 +168,7 @@
 
 Given(/^I'm on that application page$/) do
   click_on 'Dashboard'
-  click_on 'Overview'
+  click_on 'default' # this is supposed to be the name of the service
```
Martouta, line 171:
```suggestion
  click_on @application.service.name
```

Looks more dynamic 🤔 

No need to be today 👍

features/old/stats/provider_side.feature
```diff
@@ -15,17 +15,20 @@ Feature: Provider stats
 
   Scenario: Stats access
     When I log in as provider "foo.example.com"
+    And I follow "default"
```
Martouta, line 18: Ugly but let's polish later

Martouta (approved):
Great job! 👏

Martouta (approved):

Martouta (dismissed):

Martouta (approved):

features/old/services/switch.feature
```diff
@@ -11,18 +12,19 @@ Feature: Services switch
     Given current domain is the admin domain of provider "foo.example.com"
     When I log in as provider "foo.example.com"
 
+  # FIXME!!
```
Martouta, line 15: This should be the `@wip` instead of the next one 🙈 My bad

Martouta (approved):

damianpm (approved):
🥳

PR #2200, opened 2020-10-12
Removing "intentions required" settings parameter

HemaHG:

Removing **Signup requires a description of intended use** checkbox

[THREESCALE-885](https://issues.redhat.com/browse/THREESCALE-885)


HemaHG:
cannot remove the `intentions_required `from the db, because is also used to check the Description while creating the Application

hallelujah:
@HemaHG you can remove the condition for the `Description` no?

hallelujah:

PR #2201, opened 2020-10-13
API Endpoint Provider ProxyConfig List

Martouta:
Last part of [THREESCALE-4528](https://issues.redhat.com/browse/THREESCALE-4528)
Depends on the other PRs of the issue.

APIcast needs to fetch all proxy configs of a provider and doing so by fetching first all its services (just for its IDs) and then doing an API request to proxy configs for each service is just a too slow approach.

### DIFF
--- a/app/controllers/admin/api/account/proxy_configs_controller.rb
+++ b/app/controllers/admin/api/account/proxy_configs_controller.rb
@@ -21,7 +21,7 @@ class Admin::Api::Account::ProxyConfigsController < Admin::Api::BaseController
 
   def proxy_configs
     accessible_services_ids = System::Database.oracle? ? accessible_services.pluck(:id) : accessible_services.select(:id)
-    ProxyConfig.joins(:proxy).where("proxies.service_id in (?)", accessible_services_ids).by_environment(environment)
+    ProxyConfig.joins(:proxy).where(proxies: { service_id: accessible_services_ids }).by_environment(environment)
   end
 
   def accessible_services
--- a/app/controllers/admin/api/account/proxy_configs_controller.rb
+++ b/app/controllers/admin/api/account/proxy_configs_controller.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+class Admin::Api::Account::ProxyConfigsController < Admin::Api::BaseController
+  include ApiAuthentication::BySsoToken
+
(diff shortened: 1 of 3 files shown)


app/controllers/admin/api/account/proxy_configs_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+  represents :json, collection: ::ProxyConfigsRepresenter
+
+  clear_respond_to
+  respond_to :json
+
+  rescue_from(ProxyConfig::InvalidEnvironmentError) do
+    render_error 'invalid environment', status: :unprocessable_entity
+  end
+
+  def index
```
Martouta, line 16: This endpoint is done for APIcast... should we also document it for APIdocs in order for customers to use it and have support for it ? @3scale/system-ruby

Martouta:
```ruby
tenants = Account.tenants.last(2).to_a

def make_pcs(service, number)
  number.times { service.proxy.proxy_configs.create!(environment: ProxyConfig::ENVIRONMENTS.sample, content: ({ proxy: { hosts: ['example.com']}}.to_json)) }
end

tenants.each do |t|
  t.services.each do |s|
    ServiceCreator.new(service: s).call(private_endpoint: BackendApi.default_api_backend)
    make_pcs(s, 30)
  end
end
```

app/controllers/admin/api/account/proxy_configs_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+    respond_with proxy_configs.order(:id).paginate(pagination_params)
+  end
+
+  private
+
+  def proxy_configs
+    accessible_services_ids = System::Database.oracle? ? accessible_services.pluck(:id) : accessible_services.select(:id)
+    ProxyConfig.joins(:proxy).where("proxies.service_id in (?)", accessible_services_ids).by_environment(environment)
```
macejmic, line 24: no need to change but it could be also done as: `where(proxies: { id: accessible_services_ids })`

macejmic (dismissed):

macejmic (approved):

app/controllers/admin/api/account/proxy_configs_controller.rb
Martouta: https://github.com/3scale/porta/pull/2205

PR #2202, opened 2020-10-13
[Bugfix][Test][Oracle] Test IDs with 'select' + 'map' instead of 'pluck'

Martouta:
This reverts commit 56f99c34ede3a7d63bd59981bbc448dfdaf8db81.

https://github.com/3scale/porta/pull/2156#discussion_r503959285

image


guicassolato (approved):
Pity.

_Side note:_ Maybe we should investigate how much problem this difference between the oracle adapter and the others may be causing all over the system.

PR #2221, opened 2020-10-20
Adding ApiDocs proxy to OAS v3

damianpm:

**ActiveDocs** is now using **swagger-ui v3** with **OAS 3**. But we are missing the ApiDocs proxy implementation (proxy is implemented only for v2), it results in a CORS issue when trying to make http requests:

cors

https://issues.redhat.com/browse/THREESCALE-5736

**Verification steps** 

- Go to ActiveDocs
- Create a valid OAS3 spec or use the default Echo Api spec
- Swagger-ui should be able to make the requests.
- Check also the Documentation page in the Developer Portal

Blocked by https://github.com/3scale/platform/issues/434


damianpm:
Closing as we are not going to use cors proxy anymore.
Instead we are using Apicast CORS Request Handling Policy

PR #2227, opened 2020-10-26
Fix ContextSelector title

josemigallas:
[THREESCALE-6239: Update ContextSwitcher when in Products and Backneds pages](https://issues.redhat.com/browse/THREESCALE-6239)

* Backends
Screenshot 2020-10-26 at 16 43 24
Screenshot 2020-10-26 at 16 43 45

* Products
Screenshot 2020-10-26 at 16 43 08
Screenshot 2020-10-26 at 16 43 16

**Verification steps** 

* Products and Product pages should be "Products"
* Backends and Backend pages should be "Backends"


thomasmaas (commented):
- Products and Backends fix: yes.
- Product and Backend fix: not without a solution for the product/backend name

Maybe split in 2 prs?

josemigallas:
> * Products and Backends fix: yes.
> * Product and Backend fix: not without a solution for the product/backend name
> 
> Maybe split in 2 prs?

@christiemolloy is working in parallel on this

cntlsn (approved):
looks good to me from the screenshots!
(couldn't test locally)

spec/javascripts/Navigation/ActiveMenuTitle.spec.jsx
```diff
@@ -11,29 +11,28 @@ it('should return the proper title depending on the current menu', () => {
...
   const wrapper = getWrapper('serviceadmin', { name: 'Test' }, false)
-  expect(wrapper.text()).toEqual('Api: Test')
-  expect(wrapper.find('i').first().prop('class')).toEqual('fa fa-puzzle-piece')
+  expect(wrapper.text()).toEqual('Products')
+  expect(wrapper.find('i').first().prop('class')).toEqual('fa fa-cubes')
 })
 
-it('should return a default title', () => {
-  expect(getWrapper().text()).toEqual('Choose an API')
+it('should not return a default title', () => {
```
didierofrivia, line 36: This means our default case will be returning an empty line?

josemigallas: The default case is the ContextSelector "current section" would be empty, but the list would be still there and. However, this case should never happen.

didierofrivia: Safe to remove then, no?

josemigallas: Yes indeed! Although it's there basically to make Flow (and I think Eslint) happy. Making the return type optional makes it unnecessarily verbose and what would be the result anyway? Not rendering the component? I think it's even more safe to at least have a failsafe and render it empty. WDYT?

didierofrivia: It's good enough I'd say. Maybe if needed, next iteration (if there's such a thing) we could refactor the entire switch case to a fn that takes a hash with the options.

didierofrivia (approved):

PR #2228, opened 2020-10-26
Move Settings out of Masthead into Context Switcher

josemigallas:

[THREESCALE-6241: Move "Settings" inside Context Switcher](https://issues.redhat.com/browse/THREESCALE-6241)

Screenshot 2020-10-28 at 10 54 01

**Verification steps** 

1. Account Settings can be accessed from Context Switcher
2. Item gets highlighted inside ContextSwitcher when in Account Settings

### DIFF
--- a/app/helpers/menu_helper.rb
+++ b/app/helpers/menu_helper.rb
@@ -96,6 +96,12 @@ module MenuHelper
     end
   end
 
+  def settings_link
+    can?(:manage, current_account)
+      ? provider_admin_account_path
+      : edit_provider_admin_user_personal_details_path
+  end
+
   def current_api
     (@backend_api || @service || Service.none).decorate
   end
--- a/app/helpers/menu_helper.rb
+++ b/app/helpers/menu_helper.rb
@@ -97,9 +97,7 @@ module MenuHelper
   end
 
   def settings_link
-    can?(:manage, current_account)
-      ? provider_admin_account_path
-      : edit_provider_admin_user_personal_details_path
+    can?(:manage, current_account) ? provider_admin_account_path : edit_provider_admin_user_personal_details_path
   end
 
   def current_api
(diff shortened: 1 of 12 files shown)


app/views/shared/_api_selector.html.slim
```diff
@@ -1,6 +1,9 @@
 = javascript_pack_tag 'api_selector'
 
+- settings_link = can?(:manage, current_account) ? provider_admin_account_path : edit_provider_admin_user_personal_details_path
```
damianpm, line 3: Is there any reason why `settings_link` is not defined in [menu_helper.rb](https://github.com/3scale/porta/blob/master/app/helpers/menu_helper.rb), as `current_api`, for example?

josemigallas: Ah, you're right it should be there

app/helpers/menu_helper.rb
```diff
@@ -96,6 +96,12 @@ def audience_link
     end
   end
 
+  def settings_link
+    can?(:manage, current_account)
```
damianpm, line 100: It seems you can't have a multiline ternary in ruby...
Screenshot 2020-10-28 at 15 39 38

josemigallas: D'oh 🤣

damianpm (approved):

cntlsn:
looks good to me from screenshot/description!
@josemigallas just to double-check, whatever sub-page of Account settings the user will be, the context switcher will still display "Account settings", correct? (Which is basically how it currently works already)

cntlsn:
looks good to me from screenshot/description!
@josemigallas just to double-check, whatever sub-page of Account settings the user will be, the context switcher will still display "Account settings", correct? (Which is basically how it currently works already)

josemigallas:
> looks good to me from screenshot/description!
> @josemigallas just to double-check, whatever sub-page of Account settings the user will be, the context switcher will still display "Account settings", correct? (Which is basically how it currently works already)

Yes it will and also it will be highlighted in the dropdown. I just tested it again just in case.

PR #2230, opened 2020-10-27
[WIP - Won't merge or take any decision until we speak about it ;)] Optimize all endpoints/requests for proxy configs

Martouta:
[PRs that relies on](https://github.com/3scale/porta/pull/2230#issuecomment-731033611).
Closes [THREESCALE-6237](https://issues.redhat.com/browse/THREESCALE-6237)

**Careful**: The way that searches for the latest now is inside its environment, not inside all of them AND then check the environment. We must be careful that this does not break anything.

[Example of the improvement in execution](https://issues.redhat.com/browse/THREESCALE-6237?focusedCommentId=15510093&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15510093).

Now it does not do joins and the selects are only of IDs (except to check if the service is 'deleted' and to fetch the entire rows of the selected proxy configs) so it uses indexes to access that data 😄 💃🏼 
Additionally, the `latest_versions` is used to group only the selected _proxy_ids_ and not all of them.

### DIFF
--- a/app/controllers/admin/api/account/proxy_configs_controller.rb
+++ b/app/controllers/admin/api/account/proxy_configs_controller.rb
@@ -29,7 +29,7 @@ class Admin::Api::Account::ProxyConfigsController < Admin::Api::BaseController
   ##~ op.parameters.add :name => "version", :description => "Filter by version", :dataType => "string", :required => false, :paramType => "query"
   #
   def index
-    respond_with proxy_configs.order(:id).paginate(pagination_params)
+    respond_with proxy_configs.order(id: :asc).paginate(pagination_params)
   end
 
   private
--- a/app/controllers/admin/api/account/proxy_configs_controller.rb
+++ b/app/controllers/admin/api/account/proxy_configs_controller.rb
@@ -43,7 +43,7 @@ class Admin::Api::Account::ProxyConfigsController < Admin::Api::BaseController
   end
 
   def proxy_ids
-    ProxiesForProviderOwnerAndWatcherService.call(owner: current_account, watcher: current_user || current_account).select(:id)
+    ProxiesForProviderOwnerAndWatcherQuery.call(owner: current_account, watcher: current_user || current_account).select(:id)
   end
 
   def environment
--- a/app/controllers/admin/api/account/proxy_configs_controller.rb
+++ b/app/controllers/admin/api/account/proxy_configs_controller.rb
@@ -35,20 +35,15 @@ class Admin::Api::Account::ProxyConfigsController < Admin::Api::BaseController
   private
 
   def proxy_configs
-    @proxy_configs ||= fetch_proxy_configs
+    @proxy_configs ||= if version == 'latest'
+                         ProxyConfig.latest_versions(environment: environment, proxy_ids: proxy_ids)
+                       else
+                         ProxyConfig.by_version(version).by_environment(environment).by_proxy_ids(proxy_ids)
+    end.by_host(host)
   end
 
-  def fetch_proxy_configs
-    proxy_configs = ProxyConfig.by_host(host).where(proxy_id: ProxiesForProviderOwnerAndWatcherService.call(owner: current_account, watcher: watcher).select(:id))
-    if version == 'latest'
-      proxy_configs.latest_versions(environment: environment)
-    else
-      proxy_configs.by_version(version).by_environment(environment)
-    end
-  end
-
-  def watcher
-    (current_user || current_account)
+  def proxy_ids
+    ProxiesForProviderOwnerAndWatcherService.call(owner: current_account, watcher: current_user || current_account).select(:id)
   end
 
   def environment
--- a/app/controllers/admin/api/account/proxy_configs_controller.rb
+++ b/app/controllers/admin/api/account/proxy_configs_controller.rb
@@ -35,15 +35,20 @@ class Admin::Api::Account::ProxyConfigsController < Admin::Api::BaseController
   private
 
   def proxy_configs
-    @proxy_configs ||= ProxyConfig.joins(:proxy)
-      .where(proxies: { service_id: accessible_services.pluck(:id) })
-      .by_environment(environment)
-      .by_host(host)
-      .by_version(version)
+    @proxy_configs ||= fetch_proxy_configs
   end
 
-  def accessible_services
-    (current_user || current_account).accessible_services
+  def fetch_proxy_configs
+    proxy_configs = ProxyConfig.by_host(host).where(proxy_id: ProxiesForProviderOwnerAndWatcherService.call(owner: current_account, watcher: watcher).select(:id))
+    if version == 'latest'
+      proxy_configs.latest_versions(environment: environment)
+    else
+      proxy_configs.by_version(version).by_environment(environment)
+    end
+  end
+
+  def watcher
+    (current_user || current_account)
   end
 
   def environment
--- a/app/controllers/admin/api/account/proxy_configs_controller.rb
+++ b/app/controllers/admin/api/account/proxy_configs_controller.rb
@@ -27,6 +27,8 @@ class Admin::Api::Account::ProxyConfigsController < Admin::Api::BaseController
   ##~ op.parameters.add @parameter_environment
   ##~ op.parameters.add :name => "host",    :description => "Filter by host",    :dataType => "string", :required => false, :paramType => "query"
   ##~ op.parameters.add :name => "version", :description => "Filter by version", :dataType => "string", :required => false, :paramType => "query"
+  ##~ op.parameters.add @parameter_page
+  ##~ op.parameters.add @parameter_per_page
   #
   def index
     respond_with proxy_configs.order(id: :asc).paginate(pagination_params)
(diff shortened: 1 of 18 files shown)


app/services/fetch_proxy_configs_service.rb
```diff
@@ -0,0 +1,66 @@
+# frozen_string_literal: true
+
+module FetchProxyConfigsService
+  module_function
+
+  def call(environment:, scope: ProxyConfig.all, host: nil, version: nil)
+    if version.to_s == 'latest'
+      ProxyConfig.where.has { id.in(my {latest_proxy_config_ids(scope, environment, host)}) }.by_host(host)
```
Martouta, line 8: the order of the 'by_host' needs a unit test + manually checking performance

Martouta:
⚠️ Depends on:
https://github.com/3scale/porta/pull/2256 Test System::Domain.apicast_info [Closed in favor of the next PR]
https://github.com/3scale/porta/pull/2261 Remove unused old endpoint /master/api/domain/:id [Merged]
https://github.com/3scale/porta/pull/2255 [Doc][API] GET Account/ProxyConfigs: Find -> Filter [Merged]
https://github.com/3scale/porta/pull/2266 Clean proxy configs curl command builder and model scope for services [NEEDS REVIEWS]

Martouta:
TODO: paginate?
TODO: inform that it slightly changes behaviour and might be dangerous
TODO: the order of the 'by_host' needs a unit test + manually checking performance

app/controllers/admin/api/services/proxy/configs_controller.rb
```diff
@@ -30,11 +30,11 @@ class Admin::Api::Services::Proxy::ConfigsController < Admin::Api::Services::Bas
   ##~ op.parameters.add @parameter_environment
   #
   def index
-    respond_with(service_proxy_configs)
+    respond_with service_proxy_configs.order(:id).paginate(pagination_params)
```
macejmic, line 33: is `order(:id)` without specifying the direction the same for all DBs?

app/models/proxy_config.rb
```diff
@@ -35,21 +35,20 @@ class InvalidEnvironmentError < StandardError; end
...
-  scope :by_version, ->(version) do
-    next unless version
-    next current_versions if version == 'latest'
-
-    where(version: version)
+  scope :by_version,     ->(version) { where(version: version) if version }
+  scope :by_proxy_ids,   ->(proxy_ids) { where.has { proxy_id.in(proxy_ids) } if proxy_ids }
+
+  def self.latest_versions(environment:, proxy_ids: nil)
+    where.has do
```
macejmic, line 42: did you run `explain` on this SQL, and if so, can you share the result?

app/services/proxies_for_owner_and_watcher_service.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+class ProxiesForOwnerAndWatcherService
```
macejmic, line 3: this sounds like a perfect candidate for a query object.. these kinds of classes should be separated from the business/service logic.. good example in here: https://mkdev.me/en/posts/how-to-use-query-objects-to-refactor-rails-sql-queries

app/models/proxy_config.rb
Martouta: I cannot share it publicly. It is in a comment in the Jira. Linked in the description of the PR.

app/controllers/admin/api/services/proxy/configs_controller.rb
Martouta: It should be by default `asc` everywhere 🤔 But I will add it just in case and for clarify 👍🏼

app/services/proxies_for_owner_and_watcher_service.rb
Martouta: Makes sense! 👍🏼 Thanks for catching it! 👏🏼

Martouta: I was doubting about this, and still am, because most of it is in Ruby 🤔  I will do the commit and see the difference here 👍🏼

app/queries/proxies_for_owner_and_watcher_query.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+class ProxiesForOwnerAndWatcherQuery
```
Martouta, line 3: Not really convinced about this being a query object.

https://github.com/3scale/porta/pull/2230#discussion_r528623780

macejmic: it's more Query object than a Service for me. i would keep it this way.

macejmic (approved):

app/controllers/admin/api/account/proxy_configs_controller.rb
```diff
@@ -27,23 +27,25 @@ class Admin::Api::Account::ProxyConfigsController < Admin::Api::BaseController
...
 
   private
 
   def proxy_configs
-    @proxy_configs ||= ProxyConfig.joins(:proxy)
-      .where(proxies: { service_id: accessible_services.pluck(:id) })
-      .by_environment(environment)
-      .by_host(host)
-      .by_version(version)
+    @proxy_configs ||= if version == 'latest'
```
guicassolato, line 40: Should this logic be in the controller? It looked cleaner before IMO.

app/models/proxy_config.rb
```diff
@@ -37,21 +37,20 @@ class InvalidEnvironmentError < StandardError; end
...
+      id.in(
+        ProxyConfig
+          .selecting { MAX(id).as('id') } # There is only 1 proxy_config with the maximum version for the same proxy_id and environment
```
guicassolato, line 47: What if `proxy_ids` is `nil`? We get the max version across different services?

app/models/proxy_config.rb
```diff
@@ -37,21 +37,20 @@ class InvalidEnvironmentError < StandardError; end
...
+          .by_environment(environment)
+          .by_proxy_ids(proxy_ids)
+          .grouping { proxy_id }
```
guicassolato, line 50: https://github.com/3scale/porta/pull/2230/files#r531027458

Never mind.

app/controllers/admin/api/services/proxy/configs_controller.rb
```diff
@@ -106,12 +108,14 @@ def proxy_config
   end
 
   def service_proxy_configs
-    service.proxy.proxy_configs.by_environment(environment)
+    ProxyConfig
+      .by_environment(environment)
+      .where(proxy_id: ProxiesForServiceOwnerAndWatcherQuery.call(owner: service, watcher: current_account).select(:id))
```
guicassolato, line 113: I think it would be more idiomatic with a simple `service.proxy_configs`. Defining a class to delegate the the filtering of services that are permitted to the current user breaks the way we doing in other controllers.

app/controllers/admin/api/account/proxy_configs_controller.rb
```diff
@@ -27,23 +27,25 @@ class Admin::Api::Account::ProxyConfigsController < Admin::Api::BaseController
...
+                         ProxyConfig.latest_versions(environment: environment, proxy_ids: proxy_ids)
```
guicassolato, line 41: Same as https://github.com/3scale/porta/pull/2230/files#r531033847. It should be something simpler and more direct here, like `current_account.proxy_configs`, IMO.

app/queries/proxies_for_owner_and_watcher_query.rb
```diff
@@ -0,0 +1,33 @@
...
+  def initialize(owner:, watcher: nil)
```
guicassolato, line 4: The words "owner" and "watcher" are a bit counter-intuitive considering how we name other things in general in Porta. It's like introducing new concepts for things we already know by different names. "owner" looks more like "scope" to me, while "watcher" is a "user".

app/models/proxy_config.rb
Martouta: That's not possible.
https://github.com/3scale/porta/blob/2d3d1f33ffe144b8e4f0006fdbf22e988341fae3/db/postgres_schema.rb#L1071
https://github.com/3scale/porta/blob/2d3d1f33ffe144b8e4f0006fdbf22e988341fae3/db/schema.rb#L1071
https://github.com/3scale/porta/blob/2d3d1f33ffe144b8e4f0006fdbf22e988341fae3/db/oracle_schema.rb#L1122

app/queries/proxies_for_owner_and_watcher_query.rb
Martouta: changing names is ok for me, i can agree that it can be better named. However, I disagree with the names that you suggested:
- 'user' sounds like an instance of a User, and here it means 'whoever is logged in, whether is a **user** or an **account**'
- 'scope' is definitely not even similar to 'owner'. 'owner' is like we call a similar concept in attributes of AccessToken, Metric, DeletedObject, ProxyRule, etc.

Martouta:
Closing this PR to follow a different approach in other PRs without losing sight or commit history.

PR #2234, opened 2020-10-29
Clean up dashboard tabs and old widgets

josemigallas:

When we changed the tabs with the new widgets we left behind some code that is no longer necessary.

**Verification steps** 

Dashboard should work as expected (charts, etc).


app/views/provider/admin/dashboards/_ajax_load.html.slim
```diff
@@ -1,11 +1,6 @@
...
   document.addEventListener('DOMContentLoaded', function () {
     let widgetPath = #{json widget.path}
-    let serviceId = #{json widget.params[:service_id]}
 
-    if (serviceId === null) {
-      dashboardWidget.loadAudienceWidget(widgetPath)
-    } else {
-      dashboardWidget.loadServiceWidget(serviceId, widgetPath)
-    }
+    dashboardWidget.loadAudienceWidget(widgetPath)
```
didierofrivia, line 5: Not super necessary, but we could extract this from the partial and execute it in the pack js file.

didierofrivia (approved):
An improvement would be extracting the js code from the partial. Otherwise, this is good go to.

PR #2235, opened 2020-10-30
includes field Redirect to Proxy Rule edit form

josemigallas:

Allow editting Proxy pro feature adds a redirect url field to the mapping rule object which a user can edit in the UI.

Screenshot 2020-10-30 at 11 19 12

[THREESCALE-6210: Redirect url field is not rendered when editing a mapping rule](https://issues.redhat.com/browse/THREESCALE-6210)

### DIFF
--- a/app/views/api/proxy_rules/edit.html.slim
+++ b/app/views/api/proxy_rules/edit.html.slim
@@ -1,4 +1,7 @@
 = content_for :sublayout_title, 'Edit Mapping Rule'
 
 = semantic_form_for @proxy_rule, url: admin_service_proxy_rule_path(@service, @proxy_rule) do |form|
-  = render 'shared/proxy_rules/form', form: form, proxy_rule: @proxy_rule, metrics: @service.metrics
+  = render 'shared/proxy_rules/form', form: form,
+                                      proxy_rule: @proxy_rule,
+                                      metrics: @service.metrics,
+                                      include_redirect: @service.using_proxy_pro?
--- a/app/views/api/proxy_rules/new.html.slim
+++ b/app/views/api/proxy_rules/new.html.slim
@@ -1,5 +1,7 @@
 = content_for :sublayout_title, 'New Mapping Rule'
 
 = semantic_form_for @proxy_rule, url: admin_service_proxy_rules_path(@service) do |form|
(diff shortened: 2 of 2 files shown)


guicassolato:
I think this must be seen with care.

1. No problem _a priori_ with enabling the field for self-managed services, considering we never know what kinda proxy they might be running. For standard APIcast configuration-driven, I'm afraid the redirect URL on proxy rules is not an option and IMHO we should be going in the other direction, i.e., removing the field altogether rather than adding it where's it's missing. It's a legacy feature anyway.
2. With APIAP, this can become a bit of a mess if not addressed as well at
    https://github.com/3scale/porta/blob/c5908fd404feedce47615001703c556936c73edd/app/views/provider/admin/backend_apis/mapping_rules/edit.html.slim#L4
    At the same time, who knows how the self-managed proxy would behave with an APIAP freshly generated config, right? I can only assume the provider is using a custom version of APIcast (or perhaps the Istio integration?), extended to support again the redirect URL on proxy rules. In other words, is having the redirect URL in the config even useful if the rest of the config is not supported by the version of the proxy they are running?

josemigallas:
@guicassolato according to the JIRA this is already possible from the API endpoint. If I'm not mistaken we're adding anything new here.

guicassolato:
> @guicassolato
> According to the JIRA this is already possible from the API endpoint. If I'm not mistaken we're adding anything new here.

I understand that, @josemigallas. That's why I mentioned _"going in the other direction"_ as an option. We're adding more of this legacy field where we believe it's missing, rather than removing it from the API. I just wanted to make sure that's what we really want.

josemigallas:
> > @guicassolato
> > According to the JIRA this is already possible from the API endpoint. If I'm not mistaken we're adding anything new here.
> 
> I understand that, @josemigallas. That's why I mentioned _"going in the other direction"_ as an option. We're adding more of this legacy field where we believe it's missing, rather than removing it from the API. I just wanted to make sure that's what we really want.

Gotcha. I don't think this is right place to discuss it then, information will be lost. The PR is meant as a quick fix for a rather urgent support ticket and if you think not fixing it is a possibility here, let's move the discussion up into the JIRA to give it more visibility.

thomasmaas:
Left a comment in the issue. Turns out this is the last customer using proxy pro. Not sure on best way forward.

thomasmaas (approved):

Martouta (approved):

PR #2229, opened 2020-10-26
Add current API title above left nav

christiemolloy:
**What this PR does / why we need it**:

Apart of the UX-UI-Improvements in the 2.10 ER2 release. It moves the current API name from the Context Selector to the top of the navigation in the left side.

**Which issue(s) this PR fixes** 

https://issues.redhat.com/browse/APPDUX-598

**Verification steps** 

1. Click on a Product
2. Inside the Product page check to see that the product name is on top of the left vertical navigation.

**Special notes for your reviewer**:

Screen Shot 2020-10-26 at 1 26 41 PM

PR #2236, opened 2020-10-30
Add current API title above left nav

josemigallas:
Needed signed commits, Ref #2229


thomasmaas (approved):
Would be nice in a future iteration to see if we could use the backend and product icons as a semi transparent background image to give a bit more weight to the header. But I guess that would be non-patternfly compliant.

Really like it that long titles continue on the next line.

thomasmaas (approved):

PR #2239, opened 2020-11-04
[Only some Unit Tests left] THREESCALE-1608 without member users accessing to buyers without cinstances

Martouta:
IMO this is what it should do.

I also took the chance to do a mini-refactor 😄 

If everyone agrees, I will add unit tests to cover all cases.

### DIFF
--- a/app/models/account/buyer_methods.rb
+++ b/app/models/account/buyer_methods.rb
@@ -27,6 +27,18 @@ module Account::BuyerMethods
 
     has_many :contracts, foreign_key: :user_account_id, dependent: :destroy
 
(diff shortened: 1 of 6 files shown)


guicassolato (approved):
That's excellent, @Martouta, and also refactoring some old code. I'm approving even though I know it still misses a couple tests.

Martouta:
Merging because it is not pointing to master and it is already an improvement.
I haven't invested more time to add the tests because we might end up not merging this feature at all.
Either way, we should cherry-pick one commit that is just 100% refactoring and contains the proper testing, so I will 😉

PR #2243, opened 2020-11-06
Fix summary of provider users' permission groups

guicassolato:
Before
Screenshot 2020-11-06 at 11 46 01

After
Screenshot 2020-11-06 at 11 46 19

### DIFF
--- a/app/helpers/users_helper.rb
+++ b/app/helpers/users_helper.rb
@@ -1,5 +1,6 @@
-module UsersHelper
+# frozen_string_literal: true
 
+module UsersHelper
   #
   # Use this to wrap view elements that the user can't access.
   # !! Note: this is an *interface*, not *security* feature !!
@@ -202,5 +203,11 @@ module UsersHelper
 
   def show_provider_sso_status_for_user?
     current_account.provider_can_use?(:provider_sso) && current_account.self_authentication_providers.published.any?
+  end
+
+  def permission_groups_summary(user = current_user)
+    member_permission_ids = user.member_permission_ids
+    return '-' if member_permission_ids.blank?
+    I18n.t('admin_sections.permission_groups_summary').values_at(*member_permission_ids).join(', ')
   end
 end
--- a/app/views/provider/admin/account/users/_form.html.erb
+++ b/app/views/provider/admin/account/users/_form.html.erb
@@ -11,12 +11,12 @@
   <%= content_tag :div, id: 'user-permissions-form',
                         data: { state: @user.as_json(root: false, only: [:role], methods: [:admin_sections, :member_permission_service_ids]),
                                 services: @services.as_json(root: false, only: [:id, :name, :description]),
-                                features: AdminSection.permissions_for_account(current_account) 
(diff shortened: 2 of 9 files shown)


app/helpers/users_helper.rb
```diff
@@ -203,4 +204,10 @@ def strong_password_definition
   def show_provider_sso_status_for_user?
     current_account.provider_can_use?(:provider_sso) && current_account.self_authentication_providers.published.any?
   end
+
+  def permission_groups_summary(user = current_user)
```
guicassolato, line 208: This could be in https://github.com/3scale/porta/blob/master/app/decorators/user_decorator.rb alternatively. I preferred having it here in the end because I don't see us including this string in the API response anyway.

app/views/provider/admin/account/users/_form.html.erb
```diff
@@ -11,30 +11,31 @@
...
-                       id="user_member_permission_ids_<%= role %>"
+                       id="user_member_permission_ids_<%= permission %>"
                        name="user[member_permission_ids][]"
                        type="checkbox"
-                       value="<%= role %>"
-                      <%= "checked" if @user.admin_sections.include?(role) %>  >
-                <%= sanitize AdminSection.labels(role).html_safe %>
+                       value="<%= permission %>"
+                      <%= "checked" if @user.admin_sections.include?(permission) %>  >
+                <%= sanitize t("admin_sections.#{permission}").html_safe %>
```
guicassolato, line 36: This part is all replaced in JS (except when disabled in the browser). Updating so it doesn't break in the fallback.

app/views/provider/admin/account/users/index.html.slim
```diff
@@ -38,12 +38,12 @@ table class="data" id="users"
             = user.sso_authorizations.exists? ? 'yes' : 'not yet'
         td = user.role
 
-        - if can?(:manage, :groups)
+        - if can?(:manage, :permissions)
```
guicassolato, line 41: I think this is completely useless and we could probably remove it. I don't see a case where one can have access to this page and NOT be able to manage member permissions.

I kept the conditional for the time being only because it was there, except that this has nothing to do with "groups", which is a provider switch.

config/abilities/master_admin.rb
```diff
@@ -64,6 +64,8 @@
     can :manage, Invitation
     can :manage, Invoice, provider_account_id: user.account_id unless user.account.master_on_premises?
 
+    can :manage, :permissions
```
guicassolato, line 67: Added to be able to detach "member permissions" from the provider's "groups" switch. See https://github.com/3scale/porta/pull/2243/files#r518718702.

Martouta (dismissed):
AWESOME! THANKS! 👏

josemigallas:
All good except _API_ is printed in lower case
Screenshot 2020-11-06 at 17 43 32

josemigallas (approved):

Martouta (dismissed):

josemigallas (approved):

Martouta (approved):

PR #2245, opened 2020-11-09
[Refactor] Contract.permitted_for(user)

Martouta:
Extracted from another PR 😄

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -211,19 +211,11 @@ class User < ApplicationRecord
   end
 
   def accessible_cinstances
-    cinstances = account.provided_cinstances
+    account.provided_cinstances.permitted_for(self)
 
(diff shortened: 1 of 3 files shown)


app/models/user.rb
```diff
@@ -211,19 +211,11 @@ def accessible_service_tokens
   end
 
   def accessible_cinstances
-    cinstances = account.provided_cinstances
+    account.provided_cinstances.permitted_for(self)
 
     # TODO: this has no clear migration path
     # once we would enable the :service_permissions feature for everyone
     # members that have no access to service would stop seeing applications
```
josemigallas, line 218: Is the comment still relevant?

Martouta: Not sure but I think so.

josemigallas (approved):

PR #2259, opened 2020-11-16
Removing access to products and backend pages by entering url endpoints in master tenant

HemaHG:

Removing access to products and backend pages by entering url endpoints in master tenant

https://issues.redhat.com/browse/THREESCALE-6223


HemaHG:
the PR depends on the issue https://issues.redhat.com/browse/THREESCALE-6222, which is moved back to Analysis

PR #2256, opened 2020-11-16
[WIP - Oracle....] Test System::Domain.apicast_info

Martouta:
I need it for testing the following PR 😉 
Extracted to its own PR because the other one is a bit too big now.

PR #2261, opened 2020-11-17
Remove unused old endpoint /master/api/domain/:id

Martouta:
It used to be called from https://github.com/3scale/wildcard-router-service until 2.5 or 2.6 but that repo is not used anymore and now Zync is used instead for that purpose and it does not use that endpoint but a completely different approach.

Removing this simplifies a lot the optimization of proxy config endpoints. Besides, this piece of code had a bug for oracle and postgres: https://github.com/3scale/porta/pull/2256#discussion_r525084068


josemigallas (approved):

macejmic (approved):

guicassolato (approved):
Really hard to tell if some one is using this. Ideally, we would first deprecate the endpoint, then remove it. However, I think it's safe to remove it given:
1. the urgency of this to move on with other stuff;
2. the fact the endpoint is restricted to master domain and master user access tokens;
3. that we know this endpoint was [introduced in 2018](https://github.com/3scale/system/pull/8903) so the wildcard-router could have info about domains;
4. we couldn't find any other clear dependency on this endpoint (neither in porta, nor in https://github.com/3scale/APIcast).

Martouta:
5. the endpoint is not documented anywhere 😄 (that on its own isn't big but added to all of the previous reasons then it is a thing)

Martouta:
Merging.
@thomasmaas @didierofrivia if you disagree, we can revert 👍🏼

didierofrivia:
It was pretty much safe to remove, if we start receiving issues from support, we could always revert. Good work!

PR #2265, opened 2020-11-18
Fix rubocop warnings

josemigallas:
🚨

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -623,13 +623,13 @@ GEM
       rspec (~> 3.0)
     rspec_junit_formatter (0.4.1)
       rspec-core (>= 2, < 4, != 2.12.0)
-    rubocop (0.93.1)
+    rubocop (0.92.0)
       parallel (~> 1.10)
       parser (>= 2.7.1.5)
       rainbow (>= 2.2.2, < 4.0)
-      regexp_parser (>= 1.8)
+      regexp_parser (>= 1.7)
       rexml
-      rubocop-ast (>= 0.6.0)
+      rubocop-ast (>= 0.5.0)
       ruby-progressbar (~> 1.7)
       unicode-display_width (>= 1.4.0, < 2.0)
     rubocop-ast (1.1.1)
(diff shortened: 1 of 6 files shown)


app/lib/system/deploy.rb
```diff
@@ -71,8 +70,8 @@ def self.call(_)
 
     def self.load_info!(deploy_info = parse_deploy_info)
       self.info = Info.new(deploy_info)
-    rescue StandardError => error
-      self.info = InvalidInfo.new(error)
+    rescue StandardError => e
```
Martouta, line 73: why on earth would rubocop say that `e` is clearer than `error` ? 😕

Martouta (approved):
okidoki 😄

app/lib/system/deploy.rb
josemigallas: It's not about clarity but naming convention. We could change it to `error` or disable it altogether.

Gemfile.lock
```diff
@@ -630,13 +630,13 @@ GEM
       rspec (~> 3.0)
     rspec_junit_formatter (0.4.1)
       rspec-core (>= 2, < 4, != 2.12.0)
-    rubocop (0.93.1)
+    rubocop (0.92.0)
```
Martouta, line 633: Why?
image

josemigallas: So it's aligned with https://github.com/3scale/porta/pull/2265/files#diff-e97904d8a0443f46ad2c8ceb5fbe5355e52f6be540378ab8ea1e98870395d7deR15

josemigallas: I'm trying to figure out why Circleci throws an error locally but passes here

guicassolato: Can we try going v1.x? E.g.

https://rubygems.org/gems/rubocop/versions/1.9.1
https://github.com/codeclimate/codeclimate-rubocop/branches/all?query=channel%2Frubocop-1-9

hallelujah:

PR #2266, opened 2020-11-20
Clean proxy configs curl command builder and model scope for services

Martouta:
Just a previous small cleanup for #2230
What this does:
1. Remove unused scope `for_services`
2. Makes the method `max_version` private because it is meant to be used only inside the `update_version` method in the same model.
3. Refactor `Apicast::CurlCommandBuilder::Builder#proxy_from_config` to don't fetch an entire array from the DB only to take the first record if there is any. Plus, now the request is ordered.

### DIFF
--- a/app/lib/apicast/curl_command_builder.rb
+++ b/app/lib/apicast/curl_command_builder.rb
@@ -134,9 +134,8 @@ module Apicast
     end
 
     def proxy_from_config
-      proxy_configs = proxy.proxy_configs.by_environment(environment.to_s).current_versions.to_a
-      return if proxy_configs.empty?
-      ProxyFromConfig.new(proxy_configs.first.send(:parsed_content))
+      proxy_config = proxy.proxy_configs.by_environment(environment.to_s).current_versions.order(:id).first
+      ProxyFromConfig.new(proxy_config.parsed_content) if proxy_config
     end
   end
 end
--- a/app/models/proxy_config.rb
+++ b/app/models/proxy_config.rb
@@ -35,9 +35,7 @@ class ProxyConfig < ApplicationRecord
   scope :newest_first,   -> { order(version: :desc) }
   scope :by_environment, ->(env) { where(environment: VALID_ENVIRONMENTS[env]) }
   scope :by_host,        ->(host) { where.has { hosts =~ "%|#{host}|%" } if host }
-  scope :for_services,   ->(services) do
-    joins(:proxy).merge(::Proxy.where(service_id: services))
-  end
+
   scope :current_versions, -> do
     table = BabySqueel[:proxy_configs].alias(:versions)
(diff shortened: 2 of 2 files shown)


app/lib/apicast/curl_command_builder.rb
```diff
@@ -134,9 +134,8 @@ def builder_class
     end
 
     def proxy_from_config
-      proxy_configs = proxy.proxy_configs.by_environment(environment.to_s).current_versions.to_a
-      return if proxy_configs.empty?
-      ProxyFromConfig.new(proxy_configs.first.send(:parsed_content))
+      proxy_config = proxy.proxy_configs.by_environment(environment.to_s).current_versions.order(:id).first
```
guicassolato, line 137: This doesn't improve much I think. The reason why it looks like it can be potentially a long array is because `ProxyConfig.current_versions` is generic enough to be used alone, but combined with the other scopes, effectively, it's already returning only one record here.

guicassolato: Maybe `.limit(1)` instead of `.first` could really help, because that would prevent from an unnecessary longer table scan.

guicassolato: Never mind. AR will take care of it and virtually it will be same.

guicassolato (approved):

app/lib/apicast/curl_command_builder.rb
guicassolato: The `order(:id)` could be removed, BTW. It won't change anything, though it won't do any harm either.

Martouta: > This doesn't improve much I think. The reason why it looks like it can be potentially a long array is because ProxyConfig.current_versions is generic enough to be used alone, but combined with the other scopes, effectively, it's already returning only one record here.

I know. This is what https://github.com/3scale/porta/pull/2230 is for 😉 But it is a huge PR and I wanted to extract small cleanups in here first 😄 

> Maybe .limit(1) instead of .first could really help, because that would prevent from an unnecessary longer table scan.
> Never mind. AR will take care of it and virtually it will be same.

You replied yourself 😄 💪🏼 

> The order(:id) could be removed, BTW. It won't change anything, though it won't do any harm either.

I do not think so. When I was doing https://github.com/3scale/porta/pull/2256#discussion_r525084068 I could see that there can be more than 1 result and that the order in Oracle and Postgres matters 😄

Martouta: Btw, it is not only 1 record here, it is one record per version, proxy_id and version, and here we are not filtering by proxy_id but basically all of them

Martouta: my bad, it filters by proxy 😉

PR #2267, opened 2020-11-20
Can't promote product config with Istio through UI

josemigallas:

Istio was never made APIAP-compatible so now promoting a product's integration config may throw an error. This depends on whether the Product had any associated Backend prior changing to Istio.

This change removes the validation `api_backend_exists?` in ProxyConfig when the proxy belongs to an Isito product.

[THREESCALE-6348: Unable to create product with Istio through UI](https://issues.redhat.com/browse/THREESCALE-6348)

**Verification steps** 

> Create a product
> In Integration > Backend remove all backends
> In Integration > Settings set deployment to Istio
> In Integration > Configuration it should be possible to Update Config

### DIFF
--- a/app/models/proxy_config.rb
+++ b/app/models/proxy_config.rb
@@ -26,7 +26,7 @@ class ProxyConfig < ApplicationRecord
   validates :content, :version, :environment, presence: true
   validates :environment, inclusion: { in: ENVIRONMENTS }
   validate :service_token_exists
-  validate :api_backend_exists, on: :create, if: -> { !service_mesh_integration? }
+  validate :api_backend_exists, on: :create, unless: :service_mesh_integration?
   validates :content, length: { maximum: MAX_CONTENT_LENGTH }
 
   before_create :denormalize_hosts
--- a/app/models/proxy_config.rb
+++ b/app/models/proxy_config.rb
@@ -24,7 +24,7 @@ class ProxyConfig < ApplicationRecord
   validates :content, :version, :environment, presence: true
   validates :environment, inclusion: { in: ENVIRONMENTS }
   validate :service_token_exists
-  validate :api_backend_exists, on: :create
+  validate :api_backend_exists, on: :create, if: -> { !service_mesh_integration? }
   validates :content, length: { maximum: MAX_CONTENT_LENGTH }
 
   after_create :update_version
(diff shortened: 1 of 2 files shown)


app/models/proxy_config.rb
```diff
@@ -111,11 +113,14 @@ def clone_to(environment:)
...
     errors.add :service_token, :missing
   end
 
   def api_backend_exists
-    return if proxy&.api_backend_present?
+    # FIXME: we should remove the nil check
+    return if proxy&.api_backend_present? || proxy&.service_mesh_integration?
+
     errors.add :api_backend, :missing
   end
```
Martouta, line 125: I am not sure if we should keep a validation called 'api_backend_exists' that adds an error if the backend API does not exist OR it has service mesh integration. It would make more sense to me to either change the name of the method to do as it says it does or change the condition to validate. Something like this:

```ruby
validate :api_backend_exists, on: :create, if: -> { !service_mesh_integration? }

private

delegate :service_mesh_integration?, to: :proxy, allow_nil: true

def api_backend_exists
    # FIXME: we should remove the nil check
    return if proxy&.api_backend_present?

    errors.add :api_backend, :missing
  end
```

test/unit/proxy_config_test.rb
```diff
@@ -113,6 +113,17 @@ def test_create_without_service_token
     assert proxy_config.valid?
   end
 
+  test 'missing api backend when service_mesh' do
+    service = FactoryBot.create(:simple_service)
+    proxy = FactoryBot.create(:proxy, service: service)
```
Martouta, line 118:
```suggestion
    proxy = service.proxy
```

The factory already creates it

test/unit/proxy_config_test.rb
```diff
@@ -113,6 +113,17 @@ def test_create_without_service_token
...
+    proxy.stubs(:service_mesh_integration?).returns(true)
```
Martouta, line 119: Better change it for real in the DB so we do not rely on being the same instance in memory.
```suggestion
    proxy.update!(deployment_option: Service::DeploymentOption.service_mesh)
```

test/unit/proxy_config_test.rb
```diff
@@ -113,6 +113,17 @@ def test_create_without_service_token
...
+    proxy_config = FactoryBot.build(:proxy_config, environment: 'sandbox', proxy: proxy)
```
Martouta, line 120: I guess that the specific environment in this case does not matter.
```suggestion
    proxy_config = FactoryBot.build(:proxy_config, proxy: proxy)
```

app/models/proxy_config.rb
josemigallas: Cool. Any thoughts about removing the nil check?

test/unit/proxy_config_test.rb
josemigallas: It says `deployment_option` is not an attribute so I'm gonna use the stub for now if that's OK

app/models/proxy_config.rb
```diff
@@ -24,14 +26,14 @@ class InvalidEnvironmentError < StandardError; end
   validates :content, :version, :environment, presence: true
   validates :environment, inclusion: { in: ENVIRONMENTS }
   validate :service_token_exists
-  validate :api_backend_exists, on: :create
+  validate :api_backend_exists, on: :create, if: -> { !service_mesh_integration? }
```
guicassolato, line 29:
```suggestion
  validate :api_backend_exists, on: :create, unless: :service_mesh_integration?
```

app/models/proxy_config.rb
Martouta: No thoughts 😄 it is just weird 🙈 for the previous check of proxy not being optional

guicassolato (approved):

PR #2268, opened 2020-11-23
[Openshift] core.yml - backend setting

macejmic:
related to: https://github.com/3scale/platform/pull/322

```
[Core] Using http://backend-listener:3000/internal/ as URL
```

Porta is taking the default value for CORE URL as it's being read from the wrong ENV variable.

### DIFF
--- a/openshift/system/config/core.yml
+++ b/openshift/system/config/core.yml
@@ -1,5 +1,5 @@
 base: &default
-  url: <%= ENV.fetch('THREESCALE_CORE_INTERNAL_API', 'http://backend-listener:3000/internal/') %>
+  url: <%= ENV.fetch('BACKEND_ROUTE', 'http://backend-listener:3000/internal/') %>
   fake_server: false
   username: <%= ENV.fetch('CONFIG_INTERNAL_API_USER', nil) %>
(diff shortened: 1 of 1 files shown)


openshift/system/config/core.yml
```diff
@@ -1,5 +1,5 @@
 base: &default
-  url: <%= ENV.fetch('THREESCALE_CORE_INTERNAL_API', 'http://backend-listener:3000/internal/') %>
+  url: <%= ENV.fetch('BACKEND_ROUTE', 'http://backend-listener:3000/internal/') %>
```
Martouta, line 2: Why do we need to define this twice?

https://github.com/3scale/porta/blob/7fabcc4ef1b25e498f76511f2a0b42c04170bdf2/openshift/system/config/backend.yml#L2-L3

macejmic: because this config is being used in here: https://github.com/3scale/porta/blob/master/config/initializers/core.rb#L6

Martouta: That doesn't tell me why we need it twice 🙈 if it is the same thing, we should define it only once and then from the code point to the same place, no? 😕 🤔

Martouta (approved):
Ok, 🐬 will speak with Ops 😄

PR #2270, opened 2020-11-23
Add missing integration tests for ProxyConfig endpoints

Martouta:
Extracted from #2230 . I need them to ensure that the refactoring won't break anything.

### DIFF
--- a/test/integration/admin/api/account/proxy_configs_controller_test.rb
+++ b/test/integration/admin/api/account/proxy_configs_controller_test.rb
@@ -117,7 +117,7 @@ class Admin::Api::Account::ProxyConfigsControllerTest < ActionDispatch::Integrat
     assert_same_elements expected_proxy_config_ids_latest_version, response_proxy_config_ids
   end
 
-  test '#index with host param: it searches first for latest version of each proxy/service and then filters that result by host (the order of this matters)' do
+  test '#index for latest with host param: it searches first for latest version of each proxy/service and then filters that result by host (the order of this matters)' do
     service1 = FactoryBot.create(:simple_service, :with_default_backend_api, account: provider)
     service2 = FactoryBot.create(:simple_service, :with_default_backend_api, account: provider)
 
@@ -147,6 +147,18 @@ class Admin::Api::Account::ProxyConfigsControllerTest < ActionDispatch::Integrat
     ), params: {host: 'v2.example.com', version: 'latest'}
 
     assert_equal [proxy_config_service2_version2.id], response_proxy_config_ids
+  end
+
+  test '#index for latest with host param: if the same host is twice, it only returns the latest' do
+    service = FactoryBot.create(:simple_service, :with_default_backend_api, account: provider)
+    _proxy_config_version_old, proxy_config_version_new = FactoryBot.create_list(:proxy_config, 2, proxy: service.proxy, environment: ProxyConfig::ENVIRONMENTS.first, content: content_hosts('foo.example.com'))
+
+    get admin_api_account_proxy_configs_path(
+      environment: ProxyConfig::ENVIRONMENTS.first,
+      access_token: access_token_value(user: provider.admin_user)
+    ), params: {host: 'foo.example.com', version: 'latest'}
+
+    assert_equal [proxy_config_version_new.id], response_proxy_config_ids
   end
 
   private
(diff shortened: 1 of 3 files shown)


test/integration/admin/api/account/proxy_configs_controller_test.rb
```diff
@@ -111,6 +111,24 @@ def setup
...
+    end
+
+    FactoryBot.create(:proxy_config, proxy: services.last.proxy, environment: ProxyConfig::ENVIRONMENTS.first, content: ( { proxy: { hosts: ['3scale.lvh.me'] } }.to_json ) )
+
+    get admin_api_account_proxy_configs_path(
+      environment: ProxyConfig::ENVIRONMENTS.first,
+      access_token: access_token_value(user: provider.admin_user)
+    ), params: {host: 'example.org', version: 'latest'}
+
+    expected_proxy_config_ids = services.first.proxy.proxy_configs.by_host('example.org').select(:id).map(&:id)
```
guicassolato, line 128: `services.first` means hat it's neither of the two services created at the beginning of the test, right? Should we maybe add a second request with a different host and assert a different service is returned?

test/integration/admin/api/account/proxy_configs_controller_test.rb
```diff
@@ -111,6 +111,24 @@ def setup
     assert_same_elements expected_proxy_config_ids, response_proxy_config_ids
   end
 
+  test '#index of latest verisons filtered by host is only found among those' do
+    services = FactoryBot.create_list(:simple_service, 2, :with_default_backend_api, account: provider)
+    services.each do |service|
+      hosts = %w[3scale.localhost example.com]
+      FactoryBot.create(:proxy_config, proxy: service.proxy, environment: ProxyConfig::ENVIRONMENTS.first, content: ( { proxy: { hosts: hosts } }.to_json ) )
```
josemigallas, line 118: Don't use parenthesis around a method call (says rubocop):
```suggestion
      FactoryBot.create(:proxy_config, proxy: service.proxy, environment: ProxyConfig::ENVIRONMENTS.first, content: { proxy: { hosts: hosts } }.to_json )
```

test/integration/admin/api/account/proxy_configs_controller_test.rb
```diff
@@ -111,6 +111,24 @@ def setup
     assert_same_elements expected_proxy_config_ids, response_proxy_config_ids
   end
 
+  test '#index of latest verisons filtered by host is only found among those' do
+    services = FactoryBot.create_list(:simple_service, 2, :with_default_backend_api, account: provider)
+    services.each do |service|
+      hosts = %w[3scale.localhost example.com]
```
josemigallas, line 117: We can move this out of the loop:
```suggestion
      hosts = %w[3scale.localhost example.com]
```

josemigallas (changes requested):
Could we use more meaningful host names? Something that helps readability of each test: `staging.example`, `production.example`...

test/integration/admin/api/account/proxy_configs_controller_test.rb
Martouta: not what it means 🙈 I will write it clearer following Josemi's advice 😉

test/integration/admin/api/account/proxy_configs_controller_test.rb
```diff
@@ -140,8 +117,44 @@ def setup
     assert_same_elements expected_proxy_config_ids_latest_version, response_proxy_config_ids
   end
 
+  test '#index with host param: it searches first for latest version of each proxy/service and then filters that result by host (the order of this matters)' do
+    service1 = FactoryBot.create(:simple_service, :with_default_backend_api, account: provider)
+    service2 = FactoryBot.create(:simple_service, :with_default_backend_api, account: provider)
```
guicassolato, line 122:
```suggestion
    service1, service2 = FactoryBot.create_list(:simple_service, 2, :with_default_backend_api, account: provider)
```

guicassolato (commented):
It would be nice to have a case where a new proxy config it created but with the same host as before, then searching by host showing that the outdated one is not returned (even if it has the same host).

test/integration/admin/api/account/proxy_configs_controller_test.rb
Martouta: I did it this way bcz it seemed to me by my chat with @josemigallas that it would be easier to understand 🤔

guicassolato (approved):

PR #2271, opened 2020-11-24
Update SaaS developer portal Github login

HemaHG:
Overcome from the authentication via URL query parameters deprecation 

https://issues.redhat.com/browse/THREESCALE-4441

### DIFF
--- a/app/lib/three_scale/oauth2/github_client.rb
+++ b/app/lib/three_scale/oauth2/github_client.rb
@@ -49,7 +49,6 @@ module ThreeScale
       end
 
       def emails
-        access_token.options[:mode] = :query
         @emails ||= access_token.get('user/emails', :headers => { 'Accept' => 'application/vnd.github.v3' }).parsed
       end
     end
--- a/test/unit/three_scale/oauth2/github_client_test.rb
+++ b/test/unit/three_scale/oauth2/github_client_test.rb
@@ -29,4 +29,13 @@ class ThreeScale::OAuth2::GithubClientTest < ActiveSupport::TestCase
   test '#org_name' do
     assert_equal 'org', @oauth2.org_name
   end
+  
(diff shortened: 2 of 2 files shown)


hallelujah (commented):
Can you make a test for this change @HemaHG ?

hallelujah:

app/lib/three_scale/oauth2/github_client.rb
```diff
@@ -49,7 +49,7 @@ def primary_email
       end
 
       def emails
-        access_token.options[:mode] = :query
+        access_token.options[:mode] = :header
```
guicassolato, line 52:
```suggestion
```

You can remove this line. "header" mode is the default anyway.

raelga:
FYI:
```
Deprecation timeline
Brownouts

During a brownout, authentication using query parameters will temporarily fail. The goal is to trigger alerts (assuming there are any) on our customers' services to help them find unmigrated authentication calls.

The brownouts are scheduled for:

    March 17, 2021
        From 07:00 UTC to 10:00 UTC
        From 16:00 UTC to 19:00 UTC

    April 14, 2021
        From 07:00 UTC to 10:00 UTC
        From 16:00 UTC to 19:00 UTC

Removal date

All authentication using query parameters will return a status code of 401 like all other auth failures starting on:

    May 5, 2021 at 16:00 UTC
```
https://developer.github.com/changes/2020-02-10-deprecating-auth-through-query-param/

HemaHG (commented):
where the write the test for this particular change

guicassolato:
> where to write the test for this particular change

@HemaHG, how about in https://github.com/3scale/porta/blob/master/test/unit/three_scale/oauth2/github_client_test.rb?

guicassolato (approved):

hallelujah (approved):

PR #530, opened 2019-01-22
[Mapping Rules] Sorting mapping rules

didierofrivia:
Depends on #658

**What this PR does / why we need it**:

Feature: Sort mapping rules + last called
feb-25-2019 15-04-32

**Which issue(s) this PR fixes** 

fixes https://issues.jboss.org/browse/THREESCALE-1787

**Verification steps** 
1. Visit APIcast configuration page
2. Play with the order of the mapping rules
3. Profit!

**Special notes for your reviewer**:

* Done the easiest way
* This is meant to be removed in the near future favoring an entire page craft in React using APIcast json schema config.

TODO:
* [x] Fixing some bugs
* [x] Cuke
* [x] Rails part

PR #2274, opened 2020-11-25
Remove rolling update independent_mapping_rules

josemigallas:

Since APIAP is now the default, there's no need for this rolling update

[THREESCALE-6413: Remove Rolling Update for Independent Mapping Rules](https://issues.redhat.com/browse/THREESCALE-6413)


Martouta:
Remove this:
https://github.com/3scale/porta/blob/2d3d1f33ffe144b8e4f0006fdbf22e988341fae3/app/lib/logic/rolling_updates.rb#L132-L136
https://github.com/3scale/porta/blob/2d3d1f33ffe144b8e4f0006fdbf22e988341fae3/config/examples/rolling_updates.yml#L20

features/old/proxy.feature
```diff
@@ -81,18 +73,17 @@ Feature: Proxy integration
     And I go to the integration page for service "one"
     Then I should see "A valid APIcast Policies endpoint must be provided"
 
-  @javascript
-  Scenario: Sorting mapping rules
+  @wip
```
Martouta, line 76: 👀

josemigallas: Yep.. it should pass, but it won't. I followed the same steps in the UI but the result is different and I haven't been able to figure out why.

test/integration/api/proxy_rules_controller_test.rb
```diff
@@ -5,138 +5,81 @@
 class Api::ProxyRulesControllerTest < ActionDispatch::IntegrationTest
   def setup
     @provider = FactoryBot.create(:provider_account)
+    @service = FactoryBot.create(:service, account: @provider)
     login_provider @provider
```
Martouta, line 9: minor suggestion, since we are here: 🙈  
```suggestion
    login! @provider
```

The whole file is good tho 👍🏼 😉

features/step_definitions/proxy_steps.rb
```diff
@@ -64,30 +64,22 @@ def stub_deploy_calls!
...
+  click_on 'Create Mapping Rule'
 end
 
 Given(/^I save the proxy config$/) do
   click_on 'proxy-button-save-and-deploy'
 end
 
 Then(/^the mapping rules should be in the following order:$/) do |table|
-  data = @provider.default_service.proxy.proxy_rules.includes(:metric).ordered
+  data = @provider.default_service.proxy.proxy_rules.includes(:metric).order_by(:delta)
```
Martouta, line 82: This is wrong 🙈 `the mapping rules should be in the following order` implies by `position`. Using this:
https://github.com/3scale/porta/blob/9ade227d1914f2bf580b65d259e15e176693c5b8/app/models/proxy_rule.rb#L7

app/views/api/integrations/apicast/shared/_mapping_rules.html.slim
```diff
@@ -1,40 +1,4 @@
...
-            = render 'api/integrations/apicast/shared/mapping_rule', form: f, rule: rule, metrics: metrics, index: (i + 1)
-        tr#new-proxy-rule-template.ui-sortable-item style="display:none;"
-          = render 'api/integrations/apicast/shared/mapping_rule', form: f, rule: @proxy.proxy_rules.new(delta: 1), disabled: true, metrics: metrics
-      tfoot
-        tr
-          th colspan=(@service.using_proxy_pro? ? '7' : '6')
-            = link_to "Add Mapping Rule", '#add-proxy-rule', class: 'action add', id: 'add-proxy-rule'
-
-- content_for :javascripts do
-  = javascript_include_tag 'mapping_rules'
```
Martouta, line 40: Why removing this? it was not inside the `else` 😕

Martouta:
Remember to remove this method here:
https://github.com/3scale/porta/blob/9ade227d1914f2bf580b65d259e15e176693c5b8/app/models/account.rb#L565-L567

app/views/api/integrations/apicast/shared/_mapping_rules.html.slim
Martouta: Wait, you may be right here, @didierofrivia should review this since he did https://github.com/3scale/porta/pull/530

In that case, it means:
1. The sorting by dragging was working only for the page of non-APIAP. Introduced in 2.5 in [THREESCALE-1787](https://issues.redhat.com/browse/THREESCALE-1787). But we have the new page of APIAP since 2.7 and it may not work by drag and drop.
2. Mystery solved about the 🥒  `Scenario: Sorting mapping rules`... it shouldn't work since it is the APIAP page, so the solution should be removing the cucumber.
3. We should consider if we want to implement the drag and drop for APIAP. If so, out of the scope of this PR, of course.

app/assets/javascripts/mapping_rules.js
```diff
@@ -1,34 +0,0 @@
-$(document).ready(function () {
```
Martouta, line 1: ok, removing this or not depends on https://github.com/3scale/porta/pull/2274/files#r537618015

features/old/proxy.feature
Martouta: check https://github.com/3scale/porta/pull/2274/files#r537618015

> I followed the same steps in the UI but the result is different and I haven't been able to figure out why.

I followed the steps in the UI and I couldn't drag the mapping rules on the APIAP page. But it could be just me for whatever reason.

Martouta (changes requested):
Check the comments I've left

Martouta:
If you get this PR to work on its own, it could point to `master` directly and wouldn't need to be pointing to the rest of changed of removing the non-APIAP code.

app/views/api/integrations/apicast/shared/_mapping_rules.html.slim
josemigallas: This `content_for :javascript` used to be inside `else`. It was wrongly indented in this commit https://github.com/3scale/porta/commit/56ad25225cd115e36b21dfef2d5653a0ff79ea6a#diff-32cff577d8401c8cfb608cfe94d5643ba1f66497a9412a683635e1aad801c249L39-R40.

features/old/proxy.feature
josemigallas: Yeah, there is no drag feature anymore. They're re-ordered by `position` via editing.

features/step_definitions/proxy_steps.rb
```diff
@@ -73,36 +73,30 @@ def stub_deploy_calls!
...
+MAPPING_RULE_ATTR = %w[http_method pattern position metric].freeze
+
 Then(/^the mapping rules should be in the following order:$/) do |table|
   data = @provider.default_service.proxy.proxy_rules.includes(:metric).ordered
-  MAPPING_RULE_ATTR = %w[http_method pattern delta metric].freeze
   data.each_with_index do |mapping_rule, index|
     MAPPING_RULE_ATTR.each do |attr|
       actual_value = mapping_rule.public_send(attr)
       actual_value = actual_value.name if attr == 'metric'
+      # actual_value = 'delta' if attr == 'position'
```
Martouta, line 99: 👀 this probably can already be removed

Martouta (approved):
Perfect 👏🏼 💪🏼

Martouta (dismissed):

Martouta (approved):

macejmic (approved):

PR #2276, opened 2020-11-25
[Bugfix] NoMethodError in stats/data/applications#usage. 'id' for nil:NilClass

Martouta:
Closes [THREESCALE-6354](https://issues.redhat.com/browse/THREESCALE-6354)

### DIFF
--- a/app/lib/stats/views/usage.rb
+++ b/app/lib/stats/views/usage.rb
@@ -23,24 +23,22 @@ module Stats
                 :values      => data
               }
 
-        unless @cinstance.nil?
-          result[:application] = {
+        if @cinstance
+          plan = @cinstance.plan
+          account = @cinstance.user_account
+
+          result_application = {
             :id    => @cinstance.id,
             :name  => @cinstance.name,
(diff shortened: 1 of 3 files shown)


app/lib/stats/views/usage.rb
```diff
@@ -158,6 +136,26 @@ def extract_range_and_granularity(options)
...
+        result_application = {
+          id:          @cinstance.id,
+          name:        @cinstance.name,
+          state:       @cinstance.state,
+          description: @cinstance.description,
+          service: { id: @cinstance.service_id }
+        }
+
+        result_application[:plan]    = {id: plan.id,    name: plan.name}    if plan
+        result_application[:account] = {id: account.id, name: account.name} if account
```
guicassolato, line 154: Couldn't this be masking a bigger problem? Should we really have applications whose plan was deleted or missing a buyer account?

Besides, could the omitting of these attributes when the rest of info of the application is available cause side-effects in the UI or API?

Martouta:
No time for this

PR #2278, opened 2020-11-26
Remove private_endpoint from Proxy

josemigallas:

Since APIAP is now the default, there's no need for this rolling update

TBD

### DIFF
--- a/app/models/backend_api.rb
+++ b/app/models/backend_api.rb
@@ -12,6 +12,8 @@ class BackendApi < ApplicationRecord
   DELETED_STATE = :deleted
   ECHO_API_HOST = 'echo-api.3scale.net'
 
+  after_create :create_default_metrics
+
   before_destroy :validate_destroyed_by_association_or_not_used_by_services
 
   has_many :proxy_rules, as: :owner, dependent: :destroy, inverse_of: :owner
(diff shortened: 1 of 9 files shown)


app/models/proxy.rb
```diff
@@ -685,27 +685,4 @@ def self.extract_domain(url)
...
-      return if attribute_value.blank?
-
-      begin
-        uri = URI.parse(attribute_value)
-        value = URI::Generic.new(uri.scheme, uri.userinfo, uri.host, uri.port, uri.registry, uri.path, uri.opaque, uri.query, uri.fragment).to_s
-        @model[attribute] = value
-      rescue URI::InvalidURIError
-        @model.errors.add(attribute, 'Invalid domain')
-      end
-    end
```
Martouta, line 705: we still use this... inside this same class

app/models/proxy.rb
```diff
@@ -685,27 +685,4 @@ def self.extract_domain(url)
...
-  end
-
-  def generate_port(proxy_attribute)
-    PortGenerator.new(self).call(proxy_attribute)
-  end
```
Martouta, line 710: we still use this... inside this same class

Martouta (changes requested):
we cannot remove the port generator because we use it for the proxy model

app/models/backend_api.rb
```diff
@@ -38,9 +40,6 @@ class BackendApi < ApplicationRecord
 
   alias_attribute :api_backend, :private_endpoint
```
Martouta, line 41: but now the private_endpoint is removed... so, what about this?

Martouta (dismissed):

Martouta:
U will have to fix the tests tho 👍🏼

Martouta (changes requested):
TODO:
1. Fix the tests of CircleCI.
2. Extract to other PRs the unrelated stuff to what the title/description of the PR says

hallelujah:

PR #2280, opened 2020-11-26
proxy.backend_api_configs.first ordered by ID asc

josemigallas:
In case the result of 'first' is random

### DIFF
--- a/app/services/proxy_test_service.rb
+++ b/app/services/proxy_test_service.rb
@@ -104,7 +104,7 @@ class ProxyTestService
 
   def backend_api_config_path
     return unless proxy.provider_can_use?(:api_as_product)
-    proxy.backend_api_configs.first&.path
+    proxy.backend_api_configs.order(:id).first&.path
   end
 
   def api_test_host


Martouta (changes requested):
My bad, we do not need this.

#### Technical 'explanation' of why.

a `proxy` belongs to a `service`: 
https://github.com/3scale/porta/blob/2d3d1f33ffe144b8e4f0006fdbf22e988341fae3/app/models/proxy.rb#L20

`proxy.backend_api_configs` means exactly the same as `proxy.service.backend_api_configs`:

https://github.com/3scale/porta/blob/2d3d1f33ffe144b8e4f0006fdbf22e988341fae3/app/models/proxy.rb#L115

a `service` has many `backend_api_configs`:

https://github.com/3scale/porta/blob/2d3d1f33ffe144b8e4f0006fdbf22e988341fae3/app/models/service.rb#L12

https://github.com/3scale/porta/blob/2d3d1f33ffe144b8e4f0006fdbf22e988341fae3/app/lib/backend_api_logic/service_extension.rb#L8

This so far means that doing `proxy.backend_api_configs.first&.path` would mean this in the DB:
```sql
SELECT  *
FROM backend_api_configs
WHERE service_id = (SELECT service_id FROM proxies WHERE id = X LIMIT 1)
LIMIT 1
```
where 'X' here is any given proxy id number in the query instead.
This on its own would be a problem because some DBs/RDBs do not have a default order, so having many and selecting picking the first one when the first one for every time that the query is executed is a different record is of course problematic... it would mean that every time we execute this method, we would see a different result randomly.

Whoever, the reason I see we do not need this `order(:id)` is:
https://github.com/3scale/porta/blob/2d3d1f33ffe144b8e4f0006fdbf22e988341fae3/app/models/backend_api_config.rb#L7

what this line does is that by default all queries to `backend_api_configs` are sorted by id asc, so it adds this to the query (which is what we need):
```sql
ORDER BY id ASC
```

Martouta:
https://github.com/3scale/porta/pull/2280#pullrequestreview-539693156

PR #2288, opened 2020-12-03
Remove integration edit route

josemigallas:
Closes [THREESCALE-6460: Remove integrations/edit route and pages](https://issues.redhat.com/browse/THREESCALE-6460)


test/unit/presenters/api/integrations_show_presenter_test.rb
```diff
@@ -50,8 +50,6 @@ def test_environments_have_same_config?
   end
 
   test 'state_modifier' do
-    @proxy.account.stubs(:provider_can_use?).with(:api_as_product).returns(true)
```
Martouta, line 53: I recommend not doing this change here to avoid conflicts with https://github.com/3scale/porta/pull/2291, where you actually remove this entire test.
But it's up to you 👍🏼

cdn/font-awesome/4.3.0/css/font-awesome.css
```diff
@@ -913,7 +913,6 @@
 .fa-tablet:before {
   content: "\f10a";
 }
-.fa-mobile-phone:before,
```
Martouta, line 916: Then also remove it from here: https://github.com/3scale/porta/blob/master/cdn/font-awesome/4.3.0/css/font-awesome.min.css

app/helpers/services_helper.rb
```diff
@@ -27,7 +27,7 @@ def show_mappings?
   end
 
   def path_to_service(service)
```
Martouta, line 29: plz remove this method and use directly `admin_service_integration_path(service)` instead

Martouta: Not necessarily in this same PR 👍🏼

app/controllers/api/integrations_controller.rb
```diff
@@ -66,15 +65,15 @@ def update_production
 
     done_step(:apicast_gateway_deployed, final_step=true) if ApiClassificationService.test(@proxy.api_backend).real_api?
 
-    redirect_to action: :edit, anchor: 'proxy'
+    redirect_to action: :show, anchor: 'proxy'
```
Martouta, line 68: The problem of redirecting here to a legit action is that then it seems like this is also a reachable legit path, and it isn't 😄 so I propose removing it directly since it is not used anymore (for APIAP) 😉 Proposal pointing to this PR in https://github.com/3scale/porta/pull/2293

Alternatively, if you do not like my proposal, I would prefer if you add here a 'TODO' comment for it and instead of redirecting to a legit action, just do `render head: :success, nothing: true` or even `render json: {success: true}` is an acceptable middle-step.

test/unit/signup/account_manager_test.rb
```diff
@@ -127,7 +127,6 @@ class ProviderAccountManagerTest < Signup::AccountManagerTest
 
       test 'first service has a private endpoint in order to be functional for non-APIAP accounts' do
         Account.any_instance.stubs(:provider_can_use?).returns(false)
-        Account.any_instance.stubs(:provider_can_use?).with(:api_as_product).returns(false)
```
Martouta, line 130: The changes in this file don't belong to this PR. And reading the title of the tests, these changes don't match what the title says either 😞

Martouta (changes requested):
I've finally got to review everything in this PR except for the locales so far.

So, except for the locales that I do not know yet, everything else is good except for this:
1. https://github.com/3scale/porta/pull/2288#discussion_r537436352
2. https://github.com/3scale/porta/pull/2288#discussion_r537437896 (you do not need to do it in this same PR but then add a TODO comment)
3. https://github.com/3scale/porta/pull/2288#discussion_r537506820 (This is an important one). Check https://github.com/3scale/porta/pull/2293 pointing to this PR.
4. https://github.com/3scale/porta/pull/2288#discussion_r537509991 Also an important one. Check https://github.com/3scale/porta/pull/2294 pointing to the `master` branch.

 Btw, this is optional but recommended:
https://github.com/3scale/porta/pull/2288#discussion_r537434804

test/unit/signup/account_manager_test.rb
Martouta: Proposal done in https://github.com/3scale/porta/pull/2294 so it is in a logic unit (and I added tests for related untested stuff before actually changing the code)

cdn/font-awesome/4.3.0/css/font-awesome.css
josemigallas: We shouldn't even touch this files, my bad!

config/routes.rb
```diff
@@ -12,6 +14,7 @@ def call(env)
   end
 end
 
+# rubocop:disable Metrics/BlockLength
```
Martouta, line 17: In this case, this comment is good, because this is just how `routes.rb` is like, and there isn't much to do about it.
👍🏼

config/routes.rb
```diff
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
```
Martouta, line 1: Plz also check the code in the whole file to ensure that we won't break anything here 😄

cdn/font-awesome/4.3.0/css/font-awesome.css
Martouta: @josemigallas plz remember to remove this change then

features/step_definitions/proxy_steps.rb
```diff
@@ -78,6 +74,7 @@ def stub_deploy_calls!
     MAPPING_RULE_ATTR.each do |attr|
       actual_value = mapping_rule.public_send(attr)
       actual_value = actual_value.name if attr == 'metric'
+      # actual_value = 'delta' if attr == 'position'
```
Martouta, line 77: We can already remove this comment, can't we?

config/routes.rb
josemigallas: Oh, how would this break anything?

Martouta: this makes all string (well, not really, only literal ones) frozen, so if we make somewhere in this file something like `variable = 'hello'; variable << ' world'`, even if it is in a more subtle way, it will raise `FrozenError`now.

josemigallas: It looks safe, I don't see any string variable declarations or modifications.

Martouta (approved):
Great job 👏🏼

hallelujah:
@josemigallas @Martouta up for merging this?

PR #2299, opened 2020-12-11
New active docs tweak

josemigallas:
2 minor changes to the "new active doc" page.
Current:
Screenshot 2020-12-11 at 18 49 05

New:
Screenshot 2020-12-11 at 18 46 50

### DIFF
--- a/app/views/admin/api_docs/base/_form.html.slim
+++ b/app/views/admin/api_docs/base/_form.html.slim
@@ -1,6 +1,8 @@
 - content_for :javascripts do
   = javascript_include_tag 'provider/codemirror'
 
+- commit_button_label = action_name == 'edit' ? 'Update Spec' : 'Create Spec'
+
 = form.inputs do
   = form.input :name
   - if @api_docs_service.new_record?
@@ -15,7 +17,7 @@
   = form.semantic_errors :base_path
   = form.input :skip_swagger_validations, as: :boolean
 = form.buttons do
-  = form.commit_button
+  = form.commit_button commit_button_label
 
 css:
   #api_docs_service_body_input .CodeMirror {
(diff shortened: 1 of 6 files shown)


app/views/admin/api_docs/base/_form.html.slim
```diff
@@ -15,7 +15,7 @@
   = form.semantic_errors :base_path
   = form.input :skip_swagger_validations, as: :boolean
 = form.buttons do
-  = form.commit_button
+  = form.commit_button 'Create Spec'
```
damianpm, line 18: If not from here, where is coming from the "Create Service" text in the button?

josemigallas: It's inferred from the model, the form is for `ApiDocs::Service`: https://github.com/3scale/porta/blob/9f413e8361f2752c47c44b4b9beaab27f5c0669a/app/views/admin/api_docs/base/new.html.erb#L3

macejmic (approved):

damianpm (approved):

damianpm (changes requested):
Test failing

app/controllers/admin/api_docs/base_controller.rb
```diff
@@ -12,6 +12,7 @@ def index
 
   def new
     @api_docs_service = api_docs_services.new
+    @commit_button_label = 'Create Spec'
```
Martouta, line 15: Do we need an instance variable in the controller for this? This either goes to the view directly or at least a helper method that returns 'Create Spec' or 'Update Spec' depending on the `action_name` for the view to use

Martouta (changes requested):
Unnecessary instance variable in the controller for a string in the view 🙈

app/views/admin/api_docs/base/_form.html.slim
```diff
@@ -1,6 +1,8 @@
 - content_for :javascripts do
   = javascript_include_tag 'provider/codemirror'
 
+- commit_button_label = action_name == 'edit' ? 'Update Spec' : 'Create Spec'
```
Martouta, line 4: Safer:
```suggestion
- commit_button_label = @api_docs_service.new_record? ? 'Create Spec' : 'Update Spec'
```

Martouta (dismissed):

damianpm (dismissed):

Martouta (approved):

PR #2305, opened 2020-12-15
[DOC] Generate Liquid Drops MarkDown & HTML

Martouta:
I have to make changes in the `drops` docs in https://github.com/3scale/porta/pull/2306 but clearly, this has not being generated for ages and all the previous changes do not belong to the other PR... so I am generating the docs from the current `master` branch here by just doing `bundle exec rake doc:liquid:generate && HTML=true bundle exec rake doc:liquid:generate` there. And not adding changes in other files because they are problematic (see [THREESCALE-6506](https://issues.redhat.com/browse/THREESCALE-6506) for more)
Blocking https://github.com/3scale/porta/pull/2306

### DIFF
--- a/doc/liquid/tags.html
+++ b/doc/liquid/tags.html
@@ -1,198 +1,57 @@
-
-          <ul id="table-of-contents">
-            <a name="#%7Btoc.html_id%7D"></a>
-            <li>
-<a href="#tag-active_docs">Tag 'active_docs'</a>
-          <ul id="toc-h2-tag-active_docs" class="">
-            
-          </ul>
-        </li>
-<li>
-<a href="#tag-braintree_customer_form">Tag 'braintree_customer_form'</a>
-          <ul id="toc-h2-tag-braintree_customer_form" class="">
-            
-          </ul>
-        </li>
-<li>
-<a href="#tag-csrf">Tag 'csrf'</a>
-          <ul id="toc-h2-tag-csrf" class="">
-            
-          </ul>
-        </li>
-<li>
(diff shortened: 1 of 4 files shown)


doc/liquid/tags.html
```diff
@@ -269,101 +122,27 @@ <h1>
...
-     <td>Yes</td>
-     <td>Sign Up directly to plans of your choice by adding one
-         or more hidden fields with a name <code>plan_ids[]</code>.
-         If a parameter of such name is found in the current URL,
-         the input field is added automagically.
-     </td>
-   </tr>
- </table>
-
+<!-- raw HTML omitted -->
```
guicassolato, line 144: I think this should not be omitted. Have you checked how it looks in the UI?

Martouta: wow, u r right! all of the sudden it is escaping this:
https://github.com/3scale/porta/blob/df1b415ad931e42ef5d9f51b88223076a74b6319/lib/developer_portal/lib/liquid/tags/form.rb#L14-L80

I didn't check the UI because I was expecting only 'string changes' introduced by ourselves in the descriptions... MY BAD! 🤦‍♀️ Gotta fix it! Good catch! 💪🏼 👏🏼

Martouta: Explained the mystery in [THREESCALE-6506](https://issues.redhat.com/browse/THREESCALE-6506) because the problem is only in filters.html and tags.html which I do not need to change. I do not have time for more. drops.md and drops.html are fine, which is what I need.

josemigallas (approved):

PR #2306, opened 2020-12-15
Remove PaymentGateway integration for 'Adyen'

Martouta:
Depends on https://github.com/3scale/porta/pull/2305 
⚠️ Blocks https://github.com/3scale/porta/pull/2307

It has been deprecated since 2.6 (see [THREESCALE-2769](https://issues.redhat.com/browse/THREESCALE-2769)) and according to Product, we do not have currently anybody in SaaS using it so we can remove it.
It needs to be removed to be able to update the gem of ActiveMerchant. Otherwise, it would require also editing the 3scale gem [active_merchant-adyen12](https://github.com/3scale/active_merchant-adyen12), but removing Adyen, we can also 'lock' that repo and write a deprecated/unmaintained message in the README.

I've edited manually the `license.xml` because the license_finder with format XML did not work for me locally even after trying to follow what the README of https://github.com/3scale/license_finder_xml_reporter says. (see [THREESCALE-6510](https://issues.redhat.com/browse/THREESCALE-6510))


macejmic (approved):

guicassolato (approved):
LGTM. Hopefully we'll be able to do the same for Ogone and Authorize.Net

test/unit/payment_gateway_setting_test.rb
```diff
@@ -47,16 +47,6 @@ def setup
     assert @gateway_setting.configured?
   end
 
-  test '#configured? failed if any gateway_settings missing for adyen12' do
```
josemigallas, line 50: Would it make sense to keep this test but treating adyen as an unknown gateway_type? To ensure nobody uses it or it isn't installed by error or something...

Martouta: To ensure that nobody uses it, I will run a rake task before merging this.
So by the time that I merge this, it will be safe.

Martouta: https://github.com/3scale/porta/pull/2300
The goal of that is to have 2 rake tasks:
1. Save in a file info of providers actively using a payment gateway setting. Meaning specifically: id. admin_domain and state of non-scheduled_for_deletion providers that have a payment gateway setting of the type of the param that is configured (usable).
2. Remove all payment gateway setting records in the DB of a specific type. Using this one without having the first one would be crazy, but if the first one is empty or product gives up the 'ok' (highly likely), then we can run this one and forget in the code that Adyen ever existed.

josemigallas (changes requested):
The only thing missing:
https://github.com/3scale/porta/blob/5f433ca009b5734515ae3316221680266a10305d/.rubocop_todo.yml#L429

didierofrivia (approved):
LGTM

didierofrivia:
Maybe worth to trigger all builds?

josemigallas (approved):

PR #2308, opened 2020-12-16
[WIP] Pay with Stripe with 3DS2

Martouta:
Some things to polish yet, but it works 😄

### DIFF
--- a/app/lib/payment_gateways/stripe_crypt.rb
+++ b/app/lib/payment_gateways/stripe_crypt.rb
@@ -15,34 +15,39 @@ module PaymentGateways
     end
 
     def update!(params)
-      stripe_params = params.require(:stripe)
-      customer = create_stripe_customer(stripe_params)
-      update_account(customer, stripe_params)
-    rescue Stripe::CardError => error
-      report_error(error.message)
-    end
+      payment_method_id = params.require(:stripe).require(:payment_method_id)
+      payment_method = Stripe::PaymentMethod.retrieve(payment_method_id, api_key)
+      card = payment_method.card
 
-    def create_stripe_customer(stripe_params)
-      token   = stripe_params.require(:token)
-      api_key = payment_gateway_options.fetch(:login)
-      Stripe::Customer.create({ card: token,
-                                description: account.org_name,
-                                email: user.email,
-                                metadata: {
-                                  '3scale_account_reference' => buyer_reference
-                                }
-                              }, api_key)
-    end
-
-    def update_account(customer, stripe_params)
-      account.credit_card_expires_on_month = stripe_params.require(:expires_on_month)
-      account.credit_card_expires_on_year = stripe_params.require(:expires_on_year)
-      account.credit_card_partial_number = stripe_params.require(:partial_number)[-4..-1]
-      account.credit_card_auth_code = customer.id
+      account.credit_card_expires_on_month = card.exp_month
+      account.credit_card_expires_on_year = card.exp_year
+      account.credit_card_partial_number = card.last4
+      account.credit_card_auth_code = payment_method.customer
+      account.payment_method_id = payment_method_id
       account.save
     end
 
+    def create_stripe_customer
+      Stripe::Customer.create({
+        description: account.org_name,
+        email: user.email,
+        metadata: {  '3scale_account_reference' => buyer_reference }
+      }, api_key)
+    end
+
+    def create_stripe_setup_intent(customer = create_stripe_customer)
+      Stripe::SetupIntent.create({
+        payment_method_types: ['card'],
+        usage: 'off_session',
+        customer: customer.id
+      }, api_key)
+    end
+
     private
+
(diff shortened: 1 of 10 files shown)


config/environments/development.rb
```diff
@@ -75,7 +75,7 @@
   # routes, locales, etc. This feature depends on the listen gem.
   # config.file_watcher = ActiveSupport::EventedFileUpdateChecker
 
-  config.three_scale.payments.enabled = false
+  config.three_scale.payments.enabled = true
```
kevprice83, line 78: How will we handle this for SaaS? Does this impact all customers' stripe edit form transparently? Will they need to apply some change manually?

app/lib/payment_gateways/stripe_crypt.rb
```diff
@@ -15,35 +15,40 @@ def initialize(user)
...
     end
 
-    def update_account(customer, stripe_params)
-      account.credit_card_expires_on_month = stripe_params.require(:expires_on_month)
-      account.credit_card_expires_on_year = stripe_params.require(:expires_on_year)
-      account.credit_card_partial_number = stripe_params.require(:partial_number)[-4..-1]
-      account.credit_card_auth_code = customer.id
-      account.save
+    def create_stripe_setup_intent(customer = create_stripe_customer)
+      Stripe::SetupIntent.create({
```
kevprice83, line 39: Are we using the stripe gem here now instead of activemerchant?? I see the activemerchant gem doesn't support the setupintents API and only the [paymentintents API](https://github.com/activemerchant/active_merchant/blob/abc50bf93aff4dac8368119fd81280fc7f1e07d4/lib/active_merchant/billing/gateways/stripe_payment_intents.rb)

kevprice83 (commented):
So far this looks good, I see that it is WIP so apologies if the parts I am highlighting you are working on implementing but just wanted to call out a couple of things just in case :) 

Shouldn't we be retrieving the [Mandate object](https://stripe.com/docs/api/mandates/retrieve?lang=ruby) as well during this flow (assuming we are using the stripe gem that is)? 

See the explanation on the [Merchant Initiated transaction flow](https://stripe.com/en-gb-es/guides/strong-customer-authentication#merchant-initiated-transactions-including-variable-subscriptions) as when we request an exemption, during the step when the customer stores a new card, we should ask the customer to _agree_ to us charging the card in a future payment.

> To use merchant-initiated transactions, you need to authenticate the card either when it’s being saved or on the first payment. Finally, you need to get an agreement from the customer (also referred to as a “mandate”), in order to charge their card at a later point.

So I think if we use the `Mandate` [object](https://stripe.com/docs/api/mandates?lang=ruby) to get this customer acceptance we can store that for later transactions and charge the customer without breaking the regulations.

config/environments/development.rb
Martouta: We do not need to change this for SaaS. In SaaS, it is already true. And for development, we do not need it anymore since https://github.com/3scale/porta/pull/2309
So you won't even see this file changing in this PR when it is ready 😄

app/lib/payment_gateways/stripe_crypt.rb
Martouta: No, the activemerchant gem supports setup intents, and in fact, we are using them. Check  this:

https://github.com/activemerchant/active_merchant/blob/abc50bf93aff4dac8368119fd81280fc7f1e07d4/lib/active_merchant/billing/gateways/stripe_payment_intents.rb#L7

In this implementation, we are using Stripe gem to create the customer/setupIntent, and the activemerchant gem to 'charge' the customer. This is only because it is the way how we were doing it before and the less we touch in a hurry, the less risk of breaking things and having to spend Christmas fixing them (or worse).
But FYI nowadays we might be able to implement all of this with activemerchant only without the stripe gem.

config/environments/development.rb
kevprice83: Apologies, I see now this is _development.rb_ 🤦

app/lib/payment_gateways/stripe_crypt.rb
```diff
@@ -15,35 +15,40 @@ def initialize(user)
...
-    rescue Stripe::CardError => error
-      report_error(error.message)
+      payment_method_id = params.require(:stripe).require(:payment_method_id)
+      payment_method = Stripe::PaymentMethod.retrieve(payment_method_id, api_key)
+      card = payment_method.card
+
+      account.credit_card_expires_on_month = card.exp_month
+      account.credit_card_expires_on_year = card.exp_year
+      account.credit_card_partial_number = card.last4
+      account.credit_card_auth_code = payment_method.customer
```
guicassolato, line 25: Should we be creating a `PaymentDetail` here instead of using these (deprecated) helper methods?

Martouta:
1st part done [THREESCALE-6515](https://issues.redhat.com/browse/THREESCALE-6515)

PR #2315, opened 2020-12-21
Gem activemerchant version '1.107.4' before Ruby 2.4 deprecation (with monkeypatch)

Martouta:
Fixes [THREESCALE-6534](https://issues.redhat.com/browse/THREESCALE-6534) 🤞🏼 

I've also tested in Preview to ensure that it works for Ruby 2.4 just as we have it in Production. However, testing this in Preview is not as trivial as just following the PDF guide in the Jira Epic for a couple of reasons:
1. [This configuration for preview](https://github.com/3scale/porta/blob/091832b55730a9145c2cf1f6190663145f99f790/config/environments/preview.rb#L32) and [this line of logic in the code](https://github.com/3scale/porta/blob/091832b55730a9145c2cf1f6190663145f99f790/app/models/payment_transaction.rb#L27) together reject payments in preview, which is good for security, but for testing what I have personally done is deploying to preview with a branch changing that line of code for this: `if System::Application.config.three_scale.payments.enabled || %w[2445583175001 2445583388807].include?(account.provider_account.id.to_s)`, being those 2 ids my tenants so I can test with them and I also keep the safety of not making real payments for everybody else.
2. [THREESCALE-6541](https://issues.redhat.com/browse/THREESCALE-6541)

### DIFF
--- a/config/initializers/active_merchant.rb
+++ b/config/initializers/active_merchant.rb
@@ -9,3 +9,21 @@ if Rails.application.config.three_scale.active_merchant_logging
   ActiveMerchant::Billing::Gateway.wiredump_device = Rails.root.join('log/activemerchant.log').open('a')
   ActiveMerchant::Billing::Gateway.wiredump_device.sync = true
 end
+
+ActiveMerchant::Billing::StripeGateway.prepend(Module.new do
+  def headers(options = {})
+    key = options[:key] || @api_key
+    idempotency_key = options[:idempotency_key]
+
+    headers = {
+      'Authorization' => 'Basic ' + Base64.encode64(key.to_s + ':').strip.delete("\n"),
+      'User-Agent' => "Stripe/v1 ActiveMerchantBindings/#{ActiveMerchant::VERSION}",
+      'Stripe-Version' => api_version(options),
+      'X-Stripe-Client-User-Agent' => stripe_client_user_agent(options),
+      'X-Stripe-Client-User-Metadata' => {ip: options[:ip]}.to_json
+    }
+    headers['Idempotency-Key'] = idempotency_key if idempotency_key
+    headers['Stripe-Account'] = options[:stripe_account] if options[:stripe_account]
+    headers
+  end
+end)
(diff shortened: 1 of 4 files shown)


config/initializers/active_merchant.rb
```diff
@@ -9,3 +9,21 @@
...
   ActiveMerchant::Billing::Gateway.wiredump_device.sync = true
 end
+
+ActiveMerchant::Billing::StripeGateway.prepend(Module.new do
+  def headers(options = {})
+    key = options[:key] || @api_key
+    idempotency_key = options[:idempotency_key]
+
+    headers = {
+      'Authorization' => 'Basic ' + Base64.encode64(key.to_s + ':').strip.delete("\n"),
```
Martouta, line 19: All this is exactly the same as [the original code of this version of the gem](https://github.com/activemerchant/active_merchant/blob/b2f5e89eb383429d47e446f248d7bfe4f95ac3d0/lib/active_merchant/billing/gateways/stripe.rb#L611-L625) except that I am also deleting all `"\n"`.

Martouta: Failure: without the monkey patch
Success: with the monkey patch

image

guicassolato, line 19:
```suggestion
      'Authorization' => 'Basic ' + Base64.encode64(key.to_s + ':').strip.chomp,
```

Martouta: It is not the same. I need it to remove it from the middle of the string XD

guicassolato (approved):

config/initializers/active_merchant.rb
kevprice83: Hmm so that failure is saying the API key provided was invalid. Have we fixed two problems here?

Martouta: @kevprice83 What do you mean? that this bug was also already in `master`?

secretsurf: you did fix two bugs, @Martouta. https://issues.redhat.com/browse/THREESCALE-6533 🥇

PR #2325, opened 2021-01-07
Bumps braintree gem

guicassolato:
It bumps https://github.com/braintree/braintree_ruby to [v2.104.1](https://github.com/braintree/braintree_ruby/tree/2.104.1), which (among other things) extends risk data support (adding `fraud_service_provider` and other fields) and introduces some 3D-Secure-related enhancements.

Closes [THREESCALE-6558](https://issues.redhat.com/browse/THREESCALE-6558) and possibly [THREESCALE-6550](https://issues.redhat.com/browse/THREESCALE-6550).

----

**Tested in preview (without 3D-Secure):**

cc-in-3scale
customer-in-braintree
invoice-in-3scale
transaction-in-braintree

### DIFF
--- a/Gemfile.base
+++ b/Gemfile.base
@@ -66,7 +66,7 @@ gem 'stripe', '~> 5.28.0' # we need the stripe gem because activemerchant can no
 gem 'audited'
 
 gem 'acts_as_list', '~> 0.9.17'
-gem 'braintree', '~> 2.89'
+gem 'braintree', '~> 2.93'
 gem 'cancancan', '~> 2.3.0'
 gem 'formtastic', '~> 1.2.4'
 gem 'rmagick', '~> 2.15.3', require: false
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -191,7 +191,7 @@ GEM
       debug_inspector (>= 0.0.1)
     bootsnap (1.4.5)
       msgpack (~> 1.0)
-    braintree (2.91.0)
+    braintree (2.104.1)
       builder (>= 2.0.0)
     browser (2.6.1)
     bugsnag (6.11.1)
@@ -837,7 +837,7 @@ DEPENDENCIES
   baby_squeel (~> 1.3.1)
   bcrypt (~> 3.1.7)
   bootsnap (~> 1.4)
-  braintree (~> 2.89)
+  braintree (~> 2.93)
   browser
   bugsnag (~> 6.11)
   bullet (~> 5.6)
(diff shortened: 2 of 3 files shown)


Martouta (approved):
Let's deploy to preview first to ensure that it works with Ruby 2.4

HemaHG (approved):

kevprice83:
I have one question about this fix. Given that it fails [in this method](https://github.com/3scale/porta/blob/master/app/models/payment_transaction.rb#L26-L71) when executing `purchase` does that mean if 3scale cannot parse the response object correctly in the future it will fail again? Is there a way for us to always be able to extract the true status of the transaction so any future changes on the Payment Gateway Provider's side doesn't entirely break the integration? 

In theory it looks like we should be doing that [here](https://github.com/3scale/porta/blob/master/app/models/payment_transaction.rb#L22-L23) but somehow it is still logged as `failed` even when it was a success in this case. Is this the default value? cc @guicassolato

guicassolato:
> I have one question about this fix. Given that it fails [in this method](https://github.com/3scale/porta/blob/master/app/models/payment_transaction.rb#L26-L71) when executing `purchase` does that mean if 3scale cannot parse the response object correctly in the future it will fail again? Is there a way for us to always be able to extract the true status of the transaction so any future changes on the Payment Gateway Provider's side doesn't entirely break the integration?
> 
> In theory it looks like we should be doing that [here](https://github.com/3scale/porta/blob/master/app/models/payment_transaction.rb#L22-L23) but somehow it is still logged as `failed` even when it was a success in this case. Is this the default value? cc @guicassolato

@kevprice83, it's not really 3scale that parses the response. It's the active_merchant gem that does that. The problem was caused by a recent upgrade of the active_merchant gem without a corresponding one of the braintree gem. Ruby gems usually handle these dependencies between them through spec file, but in this case both, active_merchant and braintree, chose not to do it. (The former because braintree is only one among dozens of PSPs it supports so they don't want to force useless dependencies on users; the former because braintree can indeed be used regardless of active_merchant.)

All 3scale does in terms of "parsing" is using a couple helper methods to fetch the minimum info we need (check whether success or failed, get auth code and response message, basically. These never changed.

As for those other [two lines](https://github.com/3scale/porta/blob/master/app/models/payment_transaction.rb#L22-L23) you pointed, they have nothing to do with the integration with PSPs _per se_. They are Rails scope methods to select data from our database, once the records have been already obtained, parsed and stored.

In the problem we faced, active_merchant failed to parse the response message due to incompatible versions between this gem and the braintree gem, it felt back to building a "result=false" response to the caller (i.e. 3scale), 3scale acknowledged this response as a "failed" payment transaction attempt and it stored as such in the db.

kevprice83:
Okay that makes sense @guicassolato thanks for the explanation. So as long as the gems are kept "in sync" i.e. both at compatible versions with each other then we should never see such failures.

PR #2329, opened 2021-01-11
Use the same cim_gateway object inside DeveloperPortal::Admin::Account::AuthorizeNetController#hosted_success

Martouta:
Solve the possible problems in https://github.com/3scale/porta/pull/2324#discussion_r554892061 . File number 8.

### DIFF
--- a/lib/developer_portal/app/controllers/developer_portal/admin/account/authorize_net_controller.rb
+++ b/lib/developer_portal/app/controllers/developer_portal/admin/account/authorize_net_controller.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module DeveloperPortal::Admin::Account
   class AuthorizeNetController < PaymentDetailsBaseController
     def show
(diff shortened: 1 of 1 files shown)


lib/developer_portal/app/controllers/developer_portal/admin/account/authorize_net_controller.rb
```diff
@@ -15,19 +15,19 @@ def hosted_success
       auth_response = cim_gateway.get_customer_profile(
```
josemigallas, line 15: If `auth_response` comes from `cim_gateway`, you might as well instantiate it inside `update_user_and_perform_action!` instead of passing it as argument.

Martouta: In fact, it can all be in the same method 😄

josemigallas (approved):

PR #2330, opened 2021-01-11
[Hotfix] NameError (uninitialized constant ThreeScale::SpamProtection::SessionStore)

Martouta:
image

image

### DIFF
--- a/app/lib/three_scale/spam_protection/protector.rb
+++ b/app/lib/three_scale/spam_protection/protector.rb
@@ -1,21 +1,4 @@
 module ThreeScale::SpamProtection
-
-  class SessionStore
-    attr_reader :session
-
-    def initialize(session)
-      @session = session
-    end
-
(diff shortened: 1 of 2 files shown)


guicassolato (approved):
Nothing against this change, quite the contrary actually as the `ThreeScale::SpamProtection::SessionStore` is clearly not a private class of `ThreeScale::SpamProtection`, being required elsewhere as well. Still, I assume `NameError` exceptions only happens in dev env and I'm afraid the simple extraction of the file may not be enough not to fix it.

Martouta:
I think that indeed only happens in development but it seems that extracting the file fixes it because if you do from the rails console `ThreeScale::SpamProtection::SessionStore`, it recognizes it after this change but not before it.

Martouta:
Confirmed:
- Only in development
- This fixes it

PR #2334, opened 2021-01-12
Refactor stripe into react

Martouta:
Handed over to @damianpm 💪🏼 

Follow-up of https://github.com/3scale/porta/pull/2324
Part of https://issues.redhat.com/browse/THREESCALE-6531

It refactors Stripe client-side into a React component

TODOS:

- [x] Add tests
- [x] Add form validation

PR #2337, opened 2021-01-18
[Bugfix][Stripe - Save CC] When a human clicks on 'Save', confirm to Stripe; When Stripe confirms, Submit form to the 3scale Porta server

Martouta:
Closes [THREESCALE-6573](https://issues.redhat.com/browse/THREESCALE-6573)

It is not the cleanest solution, but the cleaning process must be inside #2334, and this way, we allow customers to save credit cards in the meantime.

### DIFF
--- a/app/views/payment_gateways/_stripe.html.erb
+++ b/app/views/payment_gateways/_stripe.html.erb
@@ -96,38 +96,33 @@
 
       document.getElementById('stripe_payment_method_id').value = setupIntent.payment_method;
 
-      stripeForm.submit();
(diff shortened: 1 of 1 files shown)


app/views/payment_gateways/_stripe.html.erb
```diff
@@ -123,6 +123,7 @@
         }
       }
     }).then(handleIntentResult);
+    return false;
```
damianpm, line 126: Why do we need to return false?
I think you can delete this line

guicassolato (approved):

PR #2338, opened 2021-01-18
[WIP] [Bugfix] ActionView::Template::Error in developer_portal/admin/accounts#show (undefined method `client_secret')

josemigallas:
Closes [THREESCALE-6590](https://issues.redhat.com/browse/THREESCALE-6590)

The problem is when loading the `stripe_form` liquid tag in a request that is not `/admin/account/stripe`. In the case of this tenant from `/account/show`. And we kinda allow that since `stripe_form` is documented liquid for tenants to use it in their Developer Portal.

But to be honest, I am not happy with this solution, because, although it works, it is taking responsibility that IMO belongs to the controller, and by allowing the use of certain liquid outside their intended controllers, we will end up moving a lot of the responsibility of the controllers to the inside of this `render` of each liquid.

### TODO
I've spoken with Hery and the solution will be from the JS making an AJAX call to our server to retrieve the setup intent client secret and needed data as JSON instead of just doing this in rendering the view.
You can use it or cache it for later.
Because adding a drop just for creating setup intent is a no-no. And it should always live in a controller.
Customers could build their own checkout page anywhere as they will have the setup intent created already, closer to an API drive dev portal.

### DIFF
--- a/app/views/payment_gateways/_stripe.html.erb
+++ b/app/views/payment_gateways/_stripe.html.erb
@@ -1,6 +1,6 @@
 <div id="stripe-form-wrapper"
-  data-stripe-publishable-key="<%= @stripe_publishable_key %>"
-  data-setup-intent-secret="<%= @intent.client_secret %>"
+  data-stripe-publishable-key="<%= stripe_publishable_key %>"
+  data-setup-intent-secret="<%= intent.client_secret %>"
   data-billing-address="<%= stripe_billing_address_json %>"
   data-success-url="<%= hosted_success_admin_account_stripe_path %>"
   data-credit-card-stored="<%= current_account.credit_card_stored? %>">
--- a/lib/developer_portal/app/controllers/developer_portal/admin/account/stripe_controller.rb
+++ b/lib/developer_portal/app/controllers/developer_portal/admin/account/stripe_controller.rb
@@ -4,8 +4,6 @@ module DeveloperPortal::Admin::Account
   class StripeController < PaymentDetailsBaseController
 
     def show
-      @stripe_publishable_key = site_account.payment_gateway_options[:publishable_key]
-      @intent = stripe_crypt.create_stripe_setup_intent
       render template: 'accounts/payment_gateways/show'
     end
 
--- a/lib/developer_portal/lib/liquid/tags/stripe_form.rb
+++ b/lib/developer_portal/lib/liquid/tags/stripe_form.rb
@@ -5,7 +5,10 @@ module Liquid
     class StripeForm < Liquid::Tags::PaymentGatewayBaseForm
       desc 'Renders the stripe form'
       def render(context)
-        render_erb context, 'payment_gateways/stripe', text: @text
+        controller = context.registers[:controller]
+        stripe_publishable_key = controller.send(:site_account).payment_gateway_options[:publishable_key]
+        intent = PaymentGateways::StripeCrypt.new(controller.send(:current_user)).create_stripe_setup_intent
+        render_erb context, 'payment_gateways/stripe', text: @text, intent: intent, stripe_publishable_key: stripe_publishable_key
       end
     end
   end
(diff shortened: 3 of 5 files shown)


app/views/payment_gateways/_stripe.html.erb
```diff
@@ -10,119 +10,123 @@
...
-    </fieldset>
-
-    <fieldset>
-      <div class="form-group">
-        <div class="col-md-10 operations">
-          <%= form.submit 'Save details', id: 'stripe-submit',
-                                          class: 'btn btn-primary pull-right',
-                                          disabled: true %>
+  <% if @intent.nil? %>
+    <p>Stripe is unavailable at the moment. Please try again later.</p>
```
Martouta, line 14: We should figure out in which scenario this is possible, but I doubt that the reason is Stripe being unavailable... in that case, an error would be raised and we are not catching it at the moment so it would just crash (5XX error)

Martouta: > We should figure out in which scenario this is possible

I mean figuring out the root cause. This does not seem like a Frontend problem.

josemigallas: Yes, it's not a real error message.. but I thing this one might be a random error. Related to timing maybe, weird user behavior, maybe mixed with a slow connection... I'll try a couple more scenarios in case I find out something else.

features/step_definitions/credit_card_steps.rb
```diff
@@ -112,3 +112,7 @@
     find(:css, '#braintree_nonce', visible: :hidden).set('some_braintree_nonce')
   end
 end
+
+When "Stripe is going to fail" do
+  PaymentGateways::StripeCrypt.any_instance.stubs(:create_stripe_setup_intent).returns(nil)
```
guicassolato, line 117: Ultimately, this method reduces to https://github.com/stripe/stripe-ruby/blob/9787913b355049af915cf075b29666e55c5ab7f2/lib/stripe/stripe_response.rb#L74. I see it very hard that it would return `nil`. This requires more investigation.

josemigallas: Well I think it's the only explanation for `@intent` to be `nil` 🤔

hallelujah: I think @josemigallas is right it could come from https://github.com/stripe/stripe-ruby/blob/9787913b355049af915cf075b29666e55c5ab7f2/lib/stripe/stripe_client.rb#L558

Either stripe did not respond or the response was ununderstandable

guicassolato: The line you pointed goes in a `raise`, @hallelujah. Where are rescuing from it?

guicassolato: 1. https://github.com/3scale/porta/blob/ca33031586eda06b77823e2c03b9b56d7ee3937f/lib/developer_portal/app/controllers/developer_portal/admin/account/stripe_controller.rb#L8

2. https://github.com/3scale/porta/blob/ca33031586eda06b77823e2c03b9b56d7ee3937f/app/lib/payment_gateways/stripe_crypt.rb#L34

3. https://github.com/stripe/stripe-ruby/blob/9787913b355049af915cf075b29666e55c5ab7f2/lib/stripe/resources/setup_intent.rb#L6

4. https://github.com/stripe/stripe-ruby/blob/9787913b355049af915cf075b29666e55c5ab7f2/lib/stripe/api_operations/create.rb#L7

    4.1. https://github.com/stripe/stripe-ruby/blob/9787913b355049af915cf075b29666e55c5ab7f2/lib/stripe/api_operations/request.rb#L25

    4.2. https://github.com/stripe/stripe-ruby/blob/9787913b355049af915cf075b29666e55c5ab7f2/lib/stripe/stripe_client.rb#L241

    4.3. https://github.com/stripe/stripe-ruby/blob/9787913b355049af915cf075b29666e55c5ab7f2/lib/stripe/stripe_response.rb#L74

5. https://github.com/stripe/stripe-ruby/blob/9787913b355049af915cf075b29666e55c5ab7f2/lib/stripe/api_operations/create.rb#L8

    5.1. https://github.com/stripe/stripe-ruby/blob/9787913b355049af915cf075b29666e55c5ab7f2/lib/stripe/util.rb#L74

Martouta:
Closing in favor of Hery's proposal. I will do it in another PR 😉

PR #2339, opened 2021-01-19
Stripe async payment intents

guicassolato:
### TODOs
- [x] [bugfix] New Stripe customer whenever `<dev-portal>/admin/account/stripe` is loaded -- extracted to #2342 
- [x] New `payment_intents` table
- [x] Create/confirm payment intent on purchase with stripe
- [x] New API endpoint `POST /api/payment_callbacks/stripe_callbacks`
- [x] Refactor `PaymentTransaction#purchase_with_stripe` and related methods (move them to another class)
- [x] Write tests for all the above :P
- [x] Add `endpoint_secret` as another payment gateway setting field of the provider

In separate PRs:
- [ ] Show in the UI (provider billing settings page) some instructions to configure the Stripe webhook
- [ ] Act on payment intent return statuses `requires_action` and `requires_payment_method` -- i.e. build a checkout page in the Dev Portal for buyers to pay for failed-to-charge invoices online

### How to try this locally
- Run porta and migrations
- Create a Stripe test merchant account (in case you don't have one already)
- Create a RW access token with the Billing API scope
- Download and run the [Stripe CLI](https://stripe.com/docs/cli) pointing it to `<provider-domain>/api/payment_callbacks/stripe_callbacks?access_token=<access-token>`
- Add credit card to developer accounts
- Create and charge invoices (You may need to hit charge more than once depending on the credit card used.)
- In the Rails console, check for the `params.dig('next_action', 'use_stripe_sdk', 'stripe_js')` of a payment transaction with message "Requires action". You need to paste that link into the browser and authenticate the transaction.
- Check in the [Stripe Dashboard](https://dashboard.stripe.com) for customers, setup intents, payment methods and payment intents created

Stripe test credit card numbers for 3D Secure available [here](https://stripe.com/docs/payments/3d-secure#three-ds-cards).

----

Closes [THREESCALE-6531](https://issues.redhat.com/browse/THREESCALE-6531)

Contains commits of #2342

### DIFF
--- a/app/controllers/finance/api/payment_callbacks/stripe_callbacks_controller.rb
+++ b/app/controllers/finance/api/payment_callbacks/stripe_callbacks_controller.rb
@@ -1,0 +1,43 @@
+# frozen_string_literal: true
+
+class Finance::Api::PaymentCallbacks::StripeCallbacksController < Finance::Api::BaseController
+  before_action :ensure_stripe_payment_gateway
+
+  class StripeCallbackError < StandardError; end
+  class InvalidStripeEvent < StripeCallbackError; end
+
+  # Undocumented endpoint used for update callbacks of async-authorized payment transactions (mostly due to SCA regulations)
+  def create
+    sig_header = request.headers['Stripe-Signature']
+    endpoint_secret = 'whsec_6LXKKvZM3Sbf4gv5dHoZ7WTXB08vp14N' # TODO: fetch from provider's payment gateway settings
+
+    stripe_event = Stripe::Webhook.construct_event(request.raw_post, sig_header, endpoint_secret)
+    payment_intent_data = case stripe_event.type # Also checked by PaymentIntent#update_from_stripe_event, but here it can save us some processing and ensure an immediate response at the level of the controller in case of unsupported event types
+                          when 'payment_intent.succeeded'
+                            stripe_event.data.object
+                          else
+                            raise InvalidStripeEvent
+                          end
+
+    payment_intent = PaymentIntent.by_invoice(current_account.buyer_invoices).find_by!(payment_intent_id: payment_intent_data['id'])
+
+    unless payment_intent.update_from_stripe_event(stripe_event)
+      exception = StripeCallbackError.new('Cannot update Stripe payment intent')
+      System::ErrorReporting.report_error(exception, event: stripe_event, payment_intent: payment_intent)
+    end
+
+    head :ok
+  rescue JSON::ParserError, Stripe::SignatureVerificationError
+    render_error('Signature verification failed', status: :bad_request)
+  rescue InvalidStripeEvent
+    render_error(:not_found, status: :not_found)
+  end
+
+  protected
+
+  def ensure_stripe_payment_gateway
+    return if current_account.payment_gateway_type == :stripe
+    render_error(:not_found, status: :not_found)
+    false
+  end
+end
--- a/app/controllers/finance/api/payment_callbacks/stripe_callbacks_controller.rb
+++ b/app/controllers/finance/api/payment_callbacks/stripe_callbacks_controller.rb
@@ -5,11 +5,14 @@ class Finance::Api::PaymentCallbacks::StripeCallbacksController < Finance::Api::
 
   class StripeCallbackError < StandardError; end
   class InvalidStripeEvent < StripeCallbackError; end
+  class MissingStripeEndpointSecret < StripeCallbackError; end
 
   # Undocumented endpoint used for update callbacks of async-authorized payment transactions (mostly due to SCA regulations)
   def create
     sig_header = request.headers['Stripe-Signature']
-    endpoint_secret = 'whsec_6LXKKvZM3Sbf4gv5dHoZ7WTXB08vp14N' # TODO: fetch from provider's payment gateway settings
+    endpoint_secret = payment_gateway_options[:endpoint_secret].presence
+
+    raise MissingStripeEndpointSecret unless endpoint_secret
 
     stripe_event = Stripe::Webhook.construct_event(request.raw_post, sig_header, endpoint_secret)
     payment_intent_data = case stripe_event.type # Also checked by PaymentIntent#update_from_stripe_event, but here it can save us some processing and ensure an immediate response at the level of the controller in case of unsupported event types
@@ -23,7 +26,7 @@ class Finance::Api::PaymentCallbacks::StripeCallbacksController < Finance::Api::
 
     unless payment_intent.update_from_stripe_event(stripe_event)
       exception = StripeCallbackError.new('Cannot update Stripe payment intent')
-      System::ErrorReporting.report_error(exception, event: stripe_event, payment_intent: payment_intent)
+      report_error(exception, event: stripe_event, payment_intent: payment_intent)
     end
 
     head :ok
@@ -31,13 +34,20 @@ class Finance::Api::PaymentCallbacks::StripeCallbacksController < Finance::Api::
     render_error('Signature verification failed', status: :bad_request)
   rescue InvalidStripeEvent
     render_error(:not_found, status: :not_found)
+  rescue MissingStripeEndpointSecret => exception
+    report_error(exception)
+    render_error('Configuration is missing', status: :unprocessable_entity)
   end
 
   protected
 
   def ensure_stripe_payment_gateway
-    return if current_account.payment_gateway_type == :stripe
+    return if payment_gateway_type == :stripe
     render_error(:not_found, status: :not_found)
     false
   end
+
+  delegate :payment_gateway_type, :payment_gateway_options, to: :current_account
+
+  delegate :report_error, to: System::ErrorReporting
 end
--- a/app/models/invoice.rb
+++ b/app/models/invoice.rb
@@ -37,6 +37,7 @@ class Invoice < ApplicationRecord
   has_many :line_items, -> { oldest_first }, dependent: :destroy, inverse_of: :invoice
 
   has_many :payment_transactions, -> { oldest_first }, dependent: :nullify, inverse_of: :invoice
+  has_many :payment_intents, dependent: :destroy, inverse_of: :invoice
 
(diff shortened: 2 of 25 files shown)


app/models/payment_transaction.rb
```diff
@@ -152,10 +152,67 @@ def get_profile_response(credit_card_auth_code, gateway)
...
+    response = gateway.confirm_intent(payment_intent.payment_intent_id, payment_method_id, options.merge(off_session: false))
+
+    ensure_stripe_payment_intent(response) do |payment_intent_data|
+      payment_intent_status = payment_intent_data['status']
+      payment_intent.update!(state: payment_intent_status)
+
+      next if payment_intent_status == 'succeeded'
+
+      # Because Stripe won't wrap the response into an 'error' thus making ActiveMerchant to think it's a success.
+      # See https://github.com/activemerchant/active_merchant/blob/b2f5e89eb383429d47e446f248d7bfe4f95ac3d0/lib/active_merchant/billing/gateways/stripe.rb#L690
```
hallelujah, line 201: :fire: :fire: :fire: :fire: 

Is there an issue opened about that in the upstream? Would be good to investigate, we might not be the only one to stumble upon this issue or maybe we do not use it right...

app/models/account/credit_card.rb
```diff
@@ -50,6 +50,8 @@ def credit_card_stored_attribute
     case provider_account.try(:payment_gateway_type)
     when :authorize_net
       :credit_card_authorize_net_payment_profile_token
+    when :stripe
```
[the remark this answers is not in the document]
Martouta: yes, we did. This is just bcz before we stored all the credit card data together, but now we store first the customer, later everything else, so now we cannot assume that the whole credit card is stored just bcz the customer is 😄

db/migrate/20210119101158_create_payment_intents.rb
```diff
@@ -0,0 +1,16 @@
+class CreatePaymentIntents < ActiveRecord::Migration[5.0]
+  disable_ddl_transaction!
+
+  def change
+    create_table :payment_intents do |t|
```
Martouta, line 5: We should consider if we also want to add the column `tenant_id` and their corresponding triggers or no need 😄

db/migrate/20210119101158_create_payment_intents.rb
```diff
@@ -0,0 +1,16 @@
...
+      t.references :invoice, null: false
+      t.string :payment_intent_id
```
Martouta, line 7: shouldn't we enforce uniqueness? that would also help for the indexation. Like:
```ruby
t.string :payment_intent_id, unique: true, null: false
```
and below when you add the index:
```ruby
add_index :payment_intents, :payment_intent_id, index_options.merge({unique: true})
```

test/integration/finance/api/payment_callbacks/stripe_callbacks_controller_test.rb
```diff
@@ -0,0 +1,97 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class Finance::Api::PaymentCallbacks::StripeCallbacksControllerTest < ActionDispatch::IntegrationTest
+  class Create < self
```
Martouta, line 6: 🙈 
```suggestion
  class CreateTest < self
```

app/services/finance/stripe_charge_service.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  def with_payment_intent_data_from(response)
+    payment_intent_data = extract_payment_intent_data_from(response)
+    yield(payment_intent_data) if payment_intent_data.present?
+    response
+  end
+
+  def extract_payment_intent_data_from(response)
+    response_params = response.params
+    return response_params if response.success?
```
guicassolato, line 72: Gotta fix this. It won't work when the gateway is the old `ActiveMerchant::Billing::StripeGateway`, i.e. when there's no payment intents and it still uses the old Charge API.

app/controllers/finance/api/payment_callbacks/stripe_callbacks_controller.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+class Finance::Api::PaymentCallbacks::StripeCallbacksController < Finance::Api::BaseController
+  before_action :ensure_stripe_payment_gateway
+
+  class StripeCallbackError < StandardError; end
+  class InvalidStripeEvent < StripeCallbackError; end
+
+  # Undocumented endpoint used for update callbacks of async-authorized payment transactions (mostly due to SCA regulations)
```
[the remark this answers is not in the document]
guicassolato: I think this is acceptable. Plus, until we finally remove it, `provider_key` may work for this endpoint as well, as it does not depend on the user for anything.

app/models/payment_transaction.rb
```diff
@@ -152,10 +152,67 @@ def get_profile_response(credit_card_auth_code, gateway)
     gateway.cim_gateway.get_customer_profile(:customer_profile_id => credit_card_auth_code)
   end
 
+  # TODO: Move all methods below to another class
```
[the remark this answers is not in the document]
guicassolato: Done.

app/controllers/finance/api/payment_callbacks/stripe_callbacks_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+
+    head :ok
+  rescue JSON::ParserError, Stripe::SignatureVerificationError
+    render_error('Signature verification failed', status: :bad_request)
+  rescue InvalidStripeEvent
+    render_error(:not_found, status: :not_found)
+  rescue MissingStripeEndpointSecret => exception
+    report_error(exception)
+    render_error('Configuration is missing', status: :unprocessable_entity)
+  end
```
hallelujah, line 40: I would use `rescue_from` here

app/models/invoice.rb
```diff
@@ -37,6 +37,7 @@ class InvalidInvoiceStateException < RuntimeError; end
   has_many :line_items, -> { oldest_first }, dependent: :destroy, inverse_of: :invoice
 
   has_many :payment_transactions, -> { oldest_first }, dependent: :nullify, inverse_of: :invoice
+  has_many :payment_intents, dependent: :destroy, inverse_of: :invoice
```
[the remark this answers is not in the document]
hallelujah: I mean can't it be `dependent :delete_all` ?

app/models/payment_intent.rb
```diff
@@ -0,0 +1,40 @@
...
+    SUCCEEDED_STATES.include?(state)
+  end
+
+  def update_from_stripe_event(event)
+    return unless event.type == 'payment_intent.succeeded'
+    return if succeeded? || invoice.paid?
+
+    payment_intent_data = event.data.object
+
+    transaction do
```
hallelujah, line 26: I was thinking, in case of the callback triggered multiple times by Stripe in a short amount of time, there is no lock mechanism, meaning we could have erratic random results:

1. Everything goes fine, invoice is marked as paid, transaction is successful, amount is charged only once
2. Invoice  marked as paid, amount is charged once, transaction failed (double attempt to pay)
3. Invoice marked as failed/unpaid, amount is charged  once, transaction failed
4. etc ...

There is no guarantee that this code will run only once at the same time. Or if it runs multiple times then it is idempotent

hallelujah (approved):

hallelujah:
I approved but take into consideration my comment especially https://github.com/3scale/porta/pull/2339#discussion_r565072105

Martouta (approved):

Martouta (approved):

hallelujah (approved):

app/models/payment_intent.rb
```diff
@@ -0,0 +1,40 @@
...
+class PaymentIntent < ApplicationRecord
+  SUCCEEDED_STATES = %w[succeeded].freeze
+
+  belongs_to :invoice, inverse_of: :payment_intents
+
+  validates :invoice, :payment_intent_id, :state, presence: true
+  validates :payment_intent_id, :state, length: { maximum: 255 }
+
+  scope :latest, -> (count = 1) { reorder(created_at: :desc, id: :desc).limit(count) }
+  scope :latest_pending, -> (count = 1) { where.not(state: SUCCEEDED_STATES).latest(count) }
```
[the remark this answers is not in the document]
guicassolato:
```ruby
[1] pry(main)> PaymentIntent.latest_pending.to_sql
=> "SELECT  `payment_intents`.* FROM `payment_intents` WHERE (`payment_intents`.`state` != 'succeeded') ORDER BY `payment_intents`.`created_at` DESC, `payment_intents`.`id` DESC LIMIT 1"
```

```sql
mysql> EXPLAIN SELECT  `payment_intents`.* FROM `payment_intents` WHERE (`payment_intents`.`state` != 'succeeded') ORDER BY `payment_intents`.`created_at` DESC, `payment_intents`.`id` DESC LIMIT 1;
+----+-------------+-----------------+------------+------+--------------------------------+------+---------+------+------+----------+-----------------------------+
| id | select_type | table           | partitions | type | possible_keys                  | key  | key_len | ref  | rows | filtered | Extra                       |
+----+-------------+-----------------+------------+------+--------------------------------+------+---------+------+------+----------+-----------------------------+
|  1 | SIMPLE      | payment_intents | NULL       | ALL  | index_payment_intents_on_state | NULL | NULL    | NULL |    2 |   100.00 | Using where; Using filesort |
+----+-------------+-----------------+------------+------+--------------------------------+------+---------+------+------+----------+-----------------------------+
1 row in set, 1 warning (0.04 sec)
```

app/models/payment_intent.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  scope :by_invoice, ->(invoice) { where(invoice: invoice) }
+
+  def succeeded?
+    SUCCEEDED_STATES.include?(state)
+  end
+
+  def update_from_stripe_event(event)
+    return unless event.type == 'payment_intent.succeeded'
```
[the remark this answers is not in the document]
guicassolato: Stripe gem doesn't have it, but I'll move it to a constant of ours. Thanks.

PR #2342, opened 2021-01-22
[Bugfix] find or create Stripe customer whenever <dev-portal>/admin/account/stripe is loaded

Martouta:
Creates the customer in Stripe only once (or when the customer is deleted) and not each time that the page is loaded.
1st commit extracted from #2339

### DIFF
--- a/app/lib/payment_gateways/stripe_crypt.rb
+++ b/app/lib/payment_gateways/stripe_crypt.rb
@@ -29,7 +29,7 @@ module PaymentGateways
       setup_intent_params = {
         payment_method_types: ['card'],
         usage: 'off_session',
-        customer: customer.id
+        customer: find_or_create_customer.id
       }
       Stripe::SetupIntent.create(setup_intent_params, api_key)
     end
@@ -38,13 +38,26 @@ module PaymentGateways
 
     delegate :payment_detail, to: :account
 
-    def customer
+    def customer_id
+      payment_detail.credit_card_auth_code
+    end
+
+    def find_or_create_customer
+      customer = Stripe::Customer.retrieve(customer_id, api_key) if customer_id
+      customer.try(:id) ? customer : create_customer
+    end
+
+    def create_customer
       customer_params = {
         description: account.org_name,
         email: user.email,
         metadata: {  '3scale_account_reference' => buyer_reference }
       }
-      Stripe::Customer.create(customer_params, api_key)
+
+      Stripe::Customer.create(customer_params, api_key).tap do |customer|
+        payment_detail.credit_card_auth_code = customer.id
+        payment_detail.save
+      end
     end
 
     def api_key
--- a/app/lib/payment_gateways/stripe_crypt.rb
+++ b/app/lib/payment_gateways/stripe_crypt.rb
@@ -27,9 +27,13 @@ module PaymentGateways
       setup_intent_params = {
         payment_method_types: ['card'],
         usage: 'off_session',
-        customer: find_or_create_customer.id
+        customer: customer.id
       }
       Stripe::SetupIntent.create(setup_intent_params, api_key)
+    end
+
+    def customer
+      @customer ||= find_or_create_customer
     end
 
     private
@@ -38,20 +42,20 @@ module PaymentGateways
 
     def find_or_create_customer
       customer_id = payment_detail.credit_card_auth_code
-      customer_id.present? ? Stripe::Customer.retrieve(customer_id, api_key) : create_customer
+      return create_customer if customer_id.blank?
+
+      customer = Stripe::Customer.retrieve(customer_id, api_key)
+      customer.deleted? ? create_customer : customer
     end
 
     def create_customer
       customer_params = {
         description: account.org_name,
         email: user.email,
-        metadata: {  '3scale_account_reference' => buyer_reference }
+        metadata: { '3scale_account_reference' => buyer_reference }
       }
 
-      Stripe::Customer.create(customer_params, api_key).tap do |customer|
-        payment_detail.credit_card_auth_code = customer.id
-        payment_detail.save
-      end
+      Stripe::Customer.create(customer_params, api_key).tap { |stripe_customer| payment_detail.update(credit_card_auth_code: stripe_customer.id) }
     end
 
     def api_key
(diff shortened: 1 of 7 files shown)


guicassolato (approved):
Thank you!

guicassolato (dismissed):

app/lib/payment_gateways/stripe_crypt.rb
```diff
@@ -38,13 +38,26 @@ def create_stripe_setup_intent
...
     delegate :payment_detail, to: :account
 
-    def customer
+    def customer_id
+      payment_detail.credit_card_auth_code
+    end
+
+    def find_or_create_customer
+      customer = Stripe::Customer.retrieve(customer_id, api_key) if customer_id
+      customer.try(:id) ? customer : create_customer
```
josemigallas, line 47: Should we contempt the recently deleted case?
```suggestion
      customer.try(:id) && !customer.try(:deleted) ? customer : create_customer
```

Martouta: IMO already the `.try(:id)` can be overkill and could be simplified to:
```ruby
customer_id.present? ? Stripe::Customer.retrieve(customer_id, api_key) : create_customer
```
Because `Stripe::Customer.retrieve(customer_id, api_key)` either works and returns the customer with ID and everything, or just raises an error, but it is impossible to return any object without any ID.


That being said, the `deleted` check makes perfect sense to me 😉 thanks for the catch 👍🏼

app/lib/payment_gateways/stripe_crypt.rb
```diff
@@ -38,13 +36,9 @@ def create_stripe_setup_intent
...
 
-    def customer_id
-      payment_detail.credit_card_auth_code
-    end
-
     def find_or_create_customer
-      customer = Stripe::Customer.retrieve(customer_id, api_key) if customer_id
-      customer.try(:id) ? customer : create_customer
+      customer_id = payment_detail.credit_card_auth_code
+      customer_id.present? ? Stripe::Customer.retrieve(customer_id, api_key) : create_customer
```
guicassolato: And what happens when Stripe does not find the customer? According to https://github.com/stripe/stripe-ruby/blob/9787913b355049af915cf075b29666e55c5ab7f2/lib/stripe/api_resource.rb#L103, it will _build_ one. The idea here before was to ensure the customer exists in Stripe and therefore `customer.try(:id)`.

app/lib/payment_gateways/stripe_crypt.rb
```diff
@@ -27,31 +27,35 @@ def create_stripe_setup_intent
...
 
     delegate :payment_detail, to: :account
 
     def find_or_create_customer
       customer_id = payment_detail.credit_card_auth_code
-      customer_id.present? ? Stripe::Customer.retrieve(customer_id, api_key) : create_customer
+      return create_customer if customer_id.blank?
+
+      customer = Stripe::Customer.retrieve(customer_id, api_key)
+      customer.deleted? ? create_customer : customer
```
guicassolato, line 48: Ok. I see you've more or less changed it back. But is `customer.deleted?` equivalent to `customer.id == nil`? One thing is having been deleted, another perhaps it is never having been created and therefore not retrieved in the first place.

josemigallas: >When requesting the ID of a customer that has been deleted, a subset of the customer’s information will be returned, including a deleted property, which will be true.

I think both should be checked. We don't know what "subset" means here but it could even include an ID. Maybe the proper way is "listing" customers (filtered by email?) and check whether it exists there.

hallelujah: I think that means that the customer was deleted in Stripe, which is safe to check. Because now we do not know if the customer was deleted or not in Stripe for the current Account.

Martouta: If the customer has never been created, it won't return nil, it will raise an error

Martouta: I will ensure what happens and cover that case

app/lib/payment_gateways/stripe_crypt.rb
Martouta: Let's move this conversation to https://github.com/3scale/porta/pull/2342#discussion_r563567816

app/lib/payment_gateways/stripe_crypt.rb
Martouta: image

This is what I meant... If the customer exists but was once deleted vs when the customer has never existed

Martouta: @guicassolato The `new` builds the empty object (having its `id` present though), but the `refresh` does the request to the internet, and that's where the error is raised in that scenario
https://github.com/stripe/stripe-ruby/blob/9787913b355049af915cf075b29666e55c5ab7f2/lib/stripe/api_resource.rb#L101-L106

Martouta: Basically the options are:
- Returning `nil` can NEVER happen.
- Return an object with `customer.id == nil` can NEVER happen
- Returning an object with the entire customer info with `deleted?` responding to `true` is when the customer has been created in the past in Stripe but has been deleted there.
- Returning an object with the entire customer info with `deleted?` responding to `false` is when the customer was once created in Stripe and has never been deleted there.
- Raising `Stripe::InvalidRequestError` happens when the customer has never been created in Stripe for this 'tenant'.

Martouta: Another possible error raised is not having connectivity to Stripe btw 😉

Martouta: `Stripe::InvalidRequestError` is not something that I ever expect to happen, because if it has never been created, then how on earth that ID ended in our DB as a valid customer ID?
Anyway, if you want, I can rescue that error and report to Bugsnag just in case.

guicassolato: Ok. Good enough then. Thanks.

guicassolato (approved):

PR #2345, opened 2021-01-26
Default plan selector - change to PF typeahead select component

srambach:

Changes the default plan selector in Application Plans to be a React/Patternfly component

select

[APPDUX-764: [3scale][2.11][LO-prio] Improve select default Application plan](https://issues.redhat.com/browse/APPDUX-764)

**Verification steps** 

Select Product > Applications > Application Plans > Default Plan card
(url: apiconfig/services/:id/application_plans)
Marvel mockup: https://marvelapp.com/prototype/6g85cf6/screen/76056565


cntlsn:
@srambach @josemigallas I'm a bit confused about what UI component this PR is actually changing, the linked mockup doesn't match with the Jira issue and the verification steps.
Is this about the 'Default Application plan' select or the 'Change Application plan' select?

josemigallas:
> @srambach @josemigallas I'm a bit confused about what UI component this PR is actually changing, the linked mockup doesn't match with the Jira issue and the verification steps.
> Is this about the 'Default Application plan' select or the 'Change Application plan' select? Or both?

You're right, we mixed up both Jiras. It should be OK now!

cntlsn (approved):
It might need a round of microcopy edits, let's wait for @porueesq to review it.
Component looks good!

app/controllers/api/plans_base_controller.rb
```diff
@@ -112,6 +112,18 @@ def plans_index_path
...
   protected
 
+  def new_masterize_plan(issuer, assoc)
+    if params[:id].blank? || issuer.send(assoc) == @plan
+      issuer.send("#{assoc}=", nil)
+    elsif @plan.nil?
+      raise ActiveRecord::RecordNotFound, "Plan doesn't exist"
+    else
+      issuer.send("#{assoc}=", @plan)
+    end
```
guicassolato, line 122: I see some DRY opportunity here when comparing to `#generic_masterize_plan`

app/controllers/api/plans_base_controller.rb
```diff
@@ -112,6 +112,18 @@ def plans_index_path
 
   protected
 
+  def new_masterize_plan(issuer, assoc)
```
guicassolato, line 115: Are we intending to use this method in another controller the inherits from `Api::PlansBaseController`?

app/controllers/api/plans_base_controller.rb
josemigallas: `generic_masterize_plan` will be removed in a following PR (right after this one is merged), after all of the "old" selects have been replaced with this one. At which point there will be only one `#masterize_plan` method, that's why I don't think refactoring is worth it BUT any suggestions welcome.

app/controllers/api/plans_base_controller.rb
josemigallas: Not that I am aware of.

porueesq (approved):
LGTM :+1:

app/views/master/providers/plans/_widget.html.slim
```diff
@@ -20,7 +20,7 @@ div.dashboard_card.round#provider-change-plan
   br
   = form_tag edit_master_provider_plan_path(provider), :method => :get, :remote => true, :class => 'formtastic colorbox' do
     fieldset.inputs
-      legend Change Plan
+      legend Change Plan PEPE
```
didierofrivia, line 23: Plan PEPE?

app/javascript/src/utilities/patternfly-utils.jsx
```diff
@@ -0,0 +1,43 @@
...
+export interface Record {
+  id: number | string,
+  name: string
+}
+
+type Props = Record & {
+  disabled?: boolean | void
+}
+
+export class SelectOptionObject implements ISelectOptionObject {
```
didierofrivia, line 19: Would be nice to have a factory instead of a class

app/javascript/src/Plans/components/DefaultPlanSelectCard.jsx
```diff
@@ -0,0 +1,85 @@
...
+        }
+      })
+      .catch(err => {
+        console.error(err)
+        // $FlowFixMe
+        $.flash.error('An error ocurred. Please try again later.')
+      })
+      .finally(() => setIsLoading(false))
+
+    setIsLoading(true)
```
didierofrivia, line 59: I haven't seen any tests asserting the different responses and how the UI would behave

app/javascript/src/Plans/components/DefaultPlanSelectCard.jsx
```diff
@@ -0,0 +1,85 @@
...
+const NO_DEFAULT_PLAN: ApplicationPlan = { id: -1, name: '(No default plan)' }
+
+const DefaultPlanSelectCard = ({ product, initialDefaultPlan, path }: Props) => {
+  const [defaultPlan, setDefaultPlan] = useState<ApplicationPlan>(initialDefaultPlan ?? NO_DEFAULT_PLAN)
+
+  const [isLoading, setIsLoading] = useState(false)
+
+  const onSelectPlan = (plan: ApplicationPlan) => {
+    const body = plan.id >= 0 ? new URLSearchParams({ id: plan.id.toString() }) : undefined
+    const url = path.replace(':id', String(product.id))
```
didierofrivia, line 34: A bit dangerous to treat it like a string. Maybe we could make path of type URL?

app/javascript/src/Common/components/Spinner.scss
```diff
@@ -0,0 +1,211 @@
+// The Spinner is still not available in the version of @patternfly/react-core we use, therefor we need all the following styles
```
didierofrivia, line 1: This means they haven't written tests for the Spinner yet, right?

app/controllers/api/plans_base_controller.rb
```diff
@@ -28,8 +28,9 @@ def authorize_action
     authorize! :create, :plans
   end
 
-  def resource(id = params[:id])
-    collection.readonly(false).find_by_id(id)
+  def resource(id = params[:id].presence)
```
didierofrivia, line 31: wouldn't it raise with the `.presence` ?

app/javascript/packs/default_plan_selector.js
```diff
@@ -0,0 +1,26 @@
+// @flow
+
+import { DefaultPlanSelectWrapper } from 'Plans'
+import { safeFromJsonString } from 'utilities/json-utils'
+
```
damianpm, line 5: Not important, but I'd omit new ine

app/javascript/src/Common/components/Spinner.jsx
```diff
@@ -0,0 +1,24 @@
...
+
+import './Spinner.scss'
+
+type Props = {
+  size?: 'sm' | 'md' | 'lg' | 'xl',
+  className?: string
+}
+
+const Spinner = ({ size = 'md', className }: Props) => {
+  const classParameters = `pf-c-spinner pf-m-${size.toString()} ${className || ''}`
```
damianpm, line 13: Is `toString()` needed if size is already a string?
Why not using default value for `className`:
```suggestion
const Spinner = ({ size = 'md', className = '' }: Props) => {
  const classParameters = `pf-c-spinner pf-m-${size} ${className}`
```

app/javascript/src/Common/components/Spinner.jsx
```diff
@@ -0,0 +1,24 @@
...
+  return (
+    <span className={classParameters} role="progressbar" aria-valuetext="Loading projects">
+      <span className="pf-c-spinner__clipper"></span>
+      <span className="pf-c-spinner__lead-ball"></span>
+      <span className="pf-c-spinner__tail-ball"></span>
+    </span>
+  )
+}
+
+export { Spinner }
```
damianpm, line 24: Add new line at end of files, or our linter will complain

app/javascript/src/Plans/components/DefaultPlanSelect.jsx
```diff
@@ -0,0 +1,58 @@
...
+
+type Props = {
+  plan: ApplicationPlan,
+  plans: ApplicationPlan[],
+  onSelectPlan: (ApplicationPlan) => void,
+  isDisabled?: boolean
+}
+
+const DefaultPlanSelect = ({ plan, plans, onSelectPlan, isDisabled = false }: Props) => {
+  // $FlowFixMe Flow is wrong here
```
damianpm, line 20: Can you add a note explaining why? It would be good when working with this in the future

app/javascript/src/Plans/components/DefaultPlanSelect.jsx
```diff
@@ -0,0 +1,58 @@
...
+import { SelectOptionObject, toSelectOption } from 'utilities/patternfly-utils'
+
+import './DefaultPlanSelect.scss'
+
+import type { ApplicationPlan } from 'Types'
+
+type Props = {
+  plan: ApplicationPlan,
+  plans: ApplicationPlan[],
+  onSelectPlan: (ApplicationPlan) => void,
```
damianpm, line 15: Other arrow functions with a single argument don't have the `()`, like line 28

app/javascript/src/Plans/components/DefaultPlanSelect.jsx
```diff
@@ -0,0 +1,58 @@
...
+
+    const newPlan = plans.find(p => p.id.toString() === option.id)
+    if (newPlan) {
+      onSelectPlan(newPlan)
+    }
+  }
+
+  const onClear = () => {
+    setSelection(null)
+  }
```
damianpm, line 36:
```suggestion
  const onClear = () => setSelection(null)
```

app/javascript/src/Plans/components/DefaultPlanSelectCard.jsx
```diff
@@ -0,0 +1,85 @@
...
+  Card,
+  CardBody
+} from '@patternfly/react-core'
+import { DefaultPlanSelect } from 'Plans'
+import { Spinner } from 'Common'
+
+import type { Product, ApplicationPlan } from 'Types'
+
+import './DefaultPlanSelectCard.scss'
+
```
damianpm, line 18: I'd rather omit empty lines, and group all imports as a block

app/javascript/src/Plans/components/DefaultPlanSelectCard.jsx
```diff
@@ -0,0 +1,85 @@
...
+        <Form onSubmit={e => e.preventDefault()}>
+          <FormGroup
+            label="Default plan"
+            fieldId="application_plan_id"
+            helperText="Default application plan (if any) is selected automatically upon service subscription."
+          >
+            {isLoading && <Spinner size='sm' className='pf-u-ml-md' />}
+            <DefaultPlanSelect
+              plan={defaultPlan}
+              plans={[NO_DEFAULT_PLAN, ...product.appPlans].filter(p => p.id !== defaultPlan.id)} // Don't show the current default plan
```
damianpm, line 74: I'd extract this into a const, instead of filtering inline

app/views/api/plans/_default_plan.html.slim
```diff
@@ -1,3 +1,4 @@
+/ DEPRECATED: remove in favor of javascript_pack_tag 'default_plan_selector'. app/views/api/application_plans/_listing.html.slim
```
damianpm, line 1: WDYM with deprecated? The file should be removed?

josemigallas: Yes, all plan selectors will be replaced by this one and this file removed (in another PR or PRs).

damianpm:
@josemigallas have you tried this locally?
Maybe we can create a feature branch, merge there, deploy to preview and test it, before merging to master.

app/views/master/providers/plans/_widget.html.slim
josemigallas: Oops

app/javascript/src/Plans/components/DefaultPlanSelectCard.jsx
josemigallas: In this case path is "guaranteed" to be a string but in any case, URL can't be used for paths but for a full URL, which we don't know.

app/javascript/src/Plans/components/DefaultPlanSelect.jsx
josemigallas: It's hard to explain, it's some kind of bug and it complains about a prop being `number` instead of `string`, but the type is defined as `number | string`. Not sure how to explain that 😟

app/javascript/src/Common/components/Spinner.jsx
josemigallas: But there is one 😬

app/javascript/src/Plans/components/DefaultPlanSelectCard.jsx
josemigallas: Everything for an approval 🙏 🤣

app/controllers/api/plans_base_controller.rb
josemigallas: @guicassolato

app/javascript/src/Plans/components/DefaultPlanSelect.jsx
josemigallas: OK I came up with something in the end

app/controllers/api/plans_base_controller.rb
hallelujah: No basically `#presence` works with `nil`

It does something like:

```ruby
def presence
  present? ? self : nil
end

### e.g

"".presence # => nil
nil.presence # => nil
1.presence # => 1
Object.new.presence # => #<Object:0x00007fcc731ffd78>
```

app/controllers/api/plans_base_controller.rb
```diff
@@ -28,8 +28,9 @@ def authorize_action
...
+    return unless id
+    collection.readonly(false).find(id)
   end
```
hallelujah, line 34: You wanted to do probably this:

```suggestion
  def resource(id = params[:id])
    return unless id.present?
    collection.readonly(false).find(id)
  end
```

Because in your version if you call `resource("")` it will still execute `collection.readonly(false).find("")`

app/controllers/api/plans_base_controller.rb
```diff
@@ -112,16 +113,14 @@ def plans_index_path
...
 
-  def generic_masterize_plan(issuer, assoc)
-    masterize_plan do
-      if @plan.nil? || issuer.send(assoc) == @plan
-        issuer.send("#{assoc}=", nil)
-      else
-        issuer.send("#{assoc}=", @plan)
-      end
+  def assign_plan!(issuer, assoc)
+    plan = (!@plan || issuer.send(assoc) == @plan) ? nil : @plan
```
hallelujah, line 117: Basically if it is the same plan passed then you would unset it? That looks strange to me.

josemigallas: @guicassolato

app/controllers/api/plans_base_controller.rb
josemigallas: @guicassolato

app/controllers/api/plans_base_controller.rb
guicassolato: I haven's changed the logic of it.

didierofrivia:
The JS part looks good... however, there are still some cukes and units failing.

damianpm (approved):
JS part LGTM, let's wait for a Ruby review

app/helpers/api/plans_helper.rb
```diff
@@ -0,0 +1,24 @@
...
+      'current-plan': current_application_plan_data(service),
+      'path': masterize_admin_service_application_plans_path(':id')
+    }
+  end
+
+  def application_plans_data(plans)
+    plans.not_custom
+         .alphabetically
+         .to_json(root: false, only: %i[id name])
+  end
```
hallelujah, line 18: Probably we should put this code into the controller, assigning a variable.

But it is OK for now

hallelujah (approved):
Approved. New PR welcome to improve it!

Thank you for the job 🚀

app/javascript/src/utilities/alert.js
```diff
@@ -0,0 +1,5 @@
+// @flow
+
+// HACK: our flash handler is injected in the global $ in app/assets/javascripts/flash.js
+export const notice = window.$.flash.notice
+export const error = window.$.flash.error
```
hallelujah, line 5: <3

app/javascript/src/Plans/components/DefaultPlanSelectCard.jsx
```diff
@@ -10,6 +10,7 @@ import {
 } from '@patternfly/react-core'
 import { DefaultPlanSelect } from 'Plans'
 import { Spinner } from 'Common'
+import * as alert from 'utilities/alert'
```
hallelujah, line 13: Nice one! I like this, at least we can change it to `import * as alert from 'jquery'` later

hallelujah (approved):

app/javascript/src/Plans/components/DefaultPlanSelectCard.jsx
josemigallas: My idea is to eventually remove "flash" from the UI and start using Patternfly's `Alert`. This way we'll only have to change the implementation of `utilities/alert`.

hallelujah: Yes that is the way to go, please remove assets pipeline from the code, it should be handled by webpack

PR #2350, opened 2021-02-01
Add Stripe payment form

damianpm:

For any failed invoice payment (using Stripe) in the developer portal, we need to add a checkout form to enable developers retry payment. 

- [x] Improve getting the client secret for the invoice (@guicassolato)
- [x] Review UX (@cntlsn)
- [x] Add success page (redirecting to show page)
- [x] Update Docs to explain how to update show liquid template (@porueesq) (doing [here](https://issues.redhat.com/browse/THREESCALE-6657))

https://issues.redhat.com/browse/THREESCALE-6623

**Verification steps** 

To test this PR you will need a [Stripe](https://stripe.com) test merchant account with the corresponding _Secret Key_ and  _Publishable Key_. Check the [docs](https://stripe.com/docs/keys) if needed.

You will also need a Stripe _Webhook Signing Secret_. In case your environment is reachable through the Internet and with a valid TLS certificate, you can obtain the signing secret by creating a Webhook in the Stripe Dashboard (details in the steps below). Otherwise, you need to download and install the [Stripe CLI](https://stripe.com/docs/stripe-cli).
 
**1. In Admin portal:**
- Go to **Account Settings > Personal > Tokens**
- Create a **Read & Write** token with the **Billing API** scope
- If your are running 3scale in an environment publicly reachable and with a verifiable TLS certificate:
    - Go the **Stripe Dashboard**
    - Go to **Developer > Webhooks**
    - Click **Add endpoint**
    - Fill in with the following endpoint URL: `https://<provider-admin-domain>/api/payment_callbacks/stripe_callbacks?access_token=<value-of-access-token>`
    - In **Events to send**, add "payment_intent.succeeded"
    - Click **Save**
    - Now **Click to reveal** the **Signing secret** of the webhook just created and take note of it
- If your are NOT running 3scale in an environment publicly reachable or the TLS certificate cannot be verified:
    - Launch the Stripe CLI listening to `http(s)://<provider-admin-domain>/api/payment_callbacks/stripe_callbacks?access_token=<value-of-access-token>`
    - Take note of the "webhook signing secret" that the Stripe CLI prints in the terminal
- Now back in the **3scale Admin Portal**, Go to **Audience > Billing > Charging & Gateway**
- Select **Charging enabled** and click **Save**
- In **Credit card gateway > Gateway**, select **Stripe**
- Add your Secret Key, Publishable Key and Webhook Signing Secret
- Click "Save"

**2. In Developer Portal:**
- Login with a developer account
- Go to **Settings > Credit Card Details**
- Add a test credit card
- Use `4000002760003184`, as this card requires authentication on all transactions, for `MM/YY` use any future date and any `CVC`. See [docs](https://stripe.com/docs/testing#regulatory-cards)

**3. In Admin portal:**
- Go to **Audience > Accounts > Listing**
- Select the same account whose credit card info was added in the Dev Portal before
- At the top, follow the breadcrumb link to **invoices**
- Click **Create invoice**
- Click on the generated invoice to edit it
- Click the **+ Add button**, to add a line item
- Put any _Name_, _Quantity_ and _Description_
- Click on **Create Line Item**
- Click on _Issue invoice_
- Click on _Charge_

Payment will fail. That is expected, as the selected test credit card required authentication.

**4. In Admin Portal:**

- If you are working with a **newly created instance** of the Developer Portal, skip this step.
- If you want to update an **existing instance of the Developer Portal**, do the following:
  - Go to **Audience > Developer Portal > Content**
  - Edit **Root > Invoices > Show** template
  - Make this change:
Screenshot 2021-02-04 at 16 36 53

**5. In Developer Portal:**
- Go to **Settings > Invoices**
- The invoice you just created should appear as **pending**
- Click on **Show**
- Click on **Pay Invoice**
- Add a new card number, date and CVC (you can use "4111 1111 1111 1111")
- Click **Pay**

**Screenshots:**

> Screenshot 2021-02-04 at 16 39 31

> invoice02


lib/developer_portal/app/views/developer_portal/invoices/show.html.liquid
```diff
@@ -18,7 +18,9 @@ table#line-items {
         {{ invoice.period_begin | date: '%B, %Y' }} Invoice
       </div>
       <div class="panel-body panel-footer">
-
+        {% if invoice.payment_transactions.size > 0 %}
```
guicassolato, line 21: This should probably change to something that checks that state of the invoice and if a client secret can later be obtained from an existing pending payment intent, which actually would require an new method in https://github.com/3scale/porta/blob/485125888a3208188da19062c289ffabe55ebcca/lib/developer_portal/lib/liquid/drops/invoice.rb#L157

E.g.:

```ruby
module Liquid
  module Drops
    class Invoice < Drops::Model
      # ...
      def pay_now?
        @invoice.pending? && @invoice.payment_intents.pending.any?
      end
      # ...
    end
  end
end
```

```suggestion
        {% if invoice.pay_now? %}
```

lib/developer_portal/app/controllers/developer_portal/admin/account/invoices_controller.rb
```diff
@@ -31,4 +37,20 @@ def authorize_finance
...
+  def payment_intent
+    @payment_intent ||= @invoice.payment_intents.latest_pending.first
+  end
+
+  def api_key
+    payment_gateway_options[:login]
+  end
+
+  def fetch_stripe_client_secret
+    return unless payment_intent
```
guicassolato, line 52: Maybe here we should raise a 404 if the payment intent is missing.

Related to https://github.com/3scale/porta/pull/2350/files?file-filters%5B%5D=.liquid&file-filters%5B%5D=.rb#r568718935.

cntlsn:
Hey @damianpm this looks fine to me! But I would have a couple of questions:
* Is there a screen between the two screenshots posted in the PR description which includes a form to enter the new credit card details? Or are the credit card details in the second screenshot actually enabled input fields?
* What would happen after clicking the "Pay" button? Will the user leave the current window or will the payment be processed in that same window (I'm thinking of spinner, successful message..)? 
* Would placing the 'Pay Invoice' button on the same row of the 'Cance' button generate any kind of technical issues?
* What kind of notifications would an API consumer get when the payment has failed? An email? Does the email include instructions on how to get to the pending invoice, or maybe a deep link?

On a side note:
* Do the steps you describe under the "1. In Admin portal:" paragraph the real steps an API Provider needs to follow today in the Admin portal in order to add Stripe as a paying method?

app/javascript/packs/invoice_payment.js
```diff
@@ -0,0 +1,86 @@
...
+    })
+  }
+
+  const orderComplete = (paymentIntentId) => {
+    setLoading(false)
+    resultMessageLink.setAttribute(
+      'href',
+      'https://dashboard.stripe.com/test/payments/' + paymentIntentId
+    )
+    resultMessage.classList.remove('hidden')
```
damianpm, line 58: TODO: Remove this

lib/developer_portal/lib/liquid/drops/invoice.rb
```diff
@@ -172,6 +172,14 @@ def pdf_url
         @invoice.pdf.expiring_url
       end
 
+      desc "Returns true if the invoice is pending to be paid."
```
guicassolato, line 175:
```suggestion
      desc "Returns true if the online payment page is available for the invoice."
```

damianpm:
Adding @porueesq, for help with a microcopy review

thomasmaas:
I think we have a rake task that overwrites liquid file only if created_at and updated_at date are equal (file hasn't been changed by customer). We might want to consider using it for this.

damianpm:
@cntlsn 
I will record a video of both scenarios, when successful and when failed, and show it to you tomorrow (so you don't have to do all the setup) (@porueesq may also find it useful)
About the other questions:

>     * Would placing the 'Pay Invoice' button on the same row of the 'Cance' button generate any kind of technical issues?
It's possible to do it, it will look more or less like this:
Screenshot 2021-02-03 at 18 02 38

@cntlsn Please confirm if I should make the change

>     * What kind of notifications would an API consumer get when the payment has failed? An email? Does the email include instructions on how to get to the pending invoice, or maybe a deep link?
AFAIK there is no email, but @guicassolato can answer you better
You can see failed payments in the invoices tab
>     * Do the steps you describe under the "1. In Admin portal:" paragraph the real steps an API Provider needs to follow today in the Admin portal in order to add Stripe as a paying method?
Yes, this is going to be the way to do it, and should be documented

cntlsn:
Thank you @damianpm for your answer, looking forward to seeing the videos!

About the button placement, I would prefer this latter option. If possible using a styling more similar to the "Pay" button displayed on the second screen, so that it would stand out more as a CTA.

damianpm:
> Thank you @damianpm for your answer, looking forward to seeing the videos!
> 
> About the button placement, I would prefer this latter option. If possible using a styling more similar to the "Pay" button displayed on the second screen, so that it would stand out more as a CTA.

@cntlsn About the button, I'm trying to match the same style of the previous page (invoices page):
Screenshot 2021-02-03 at 18 22 28
WDYT?

cntlsn:
@damianpm let's try to make it consistent with other CTA already existing in other areas of the UI

damianpm:
@cntlsn 
Using btn-success
Provider_Name

cntlsn (approved):
Looks good to me!

Considering the urgency of this feature I wouldn't do any other UI work on this for now, but I have created a couple of jira issues for lo-prio UX enhancements:

- https://issues.redhat.com/browse/THREESCALE-6676
- https://issues.redhat.com/browse/THREESCALE-6677

lib/developer_portal/app/views/developer_portal/invoices/show.html.liquid
```diff
@@ -11,11 +11,16 @@ table#line-items {
...
-        </a>
-        {{ invoice.period_begin | date: '%B, %Y' }} Invoice
+        <div class="clearfix">
+          <a href="{{ urls.invoices }}">
+            <i class="fa fa-chevron-left"></i>
+            Cancel
+          </a>
+          {{ invoice.period_begin | date: '%B, %Y' }} Invoice
+          {% if invoice.pay_now? %}
+            <a href="{{invoice.url}}/payment" class="pull-right btn btn-success pay-invoice-btn">Pay Invoice</a>
```
porueesq, line 21:
```suggestion
            <a href="{{invoice.url}}/payment" class="pull-right btn btn-success pay-invoice-btn">Pay invoice</a>
```

lib/developer_portal/app/controllers/developer_portal/admin/account/invoices_controller.rb
```diff
@@ -1,34 +1,88 @@
...
+  end
+
+  def payment_succeeded
+    @payment_intent = @invoice.payment_intents.find_by!(reference: stripe_payment_intent_params[:id])
+
+    service = Finance::StripePaymentIntentUpdateService.new(site_account, stripe_payment_intent)
+    if service.call
+      flash[:notice] = 'Payment transaction updated'
+    else
+      flash[:error] = 'Failed to update payment transaction'
```
porueesq, line 44: @damianpm Is this message shown to the user?

lib/developer_portal/app/views/developer_portal/admin/account/invoices/payment.html.erb
```diff
@@ -0,0 +1,94 @@
...
+              </dl>
+            </div>
+          </div>
+          <div class="row stripe-form" data-publishable-key="<%= @stripe_publishable_key %>"
+            data-client-secret="<%= @client_secret %>" data-line-item="<%= @invoice.line_items.to_json %>">
+            <form id="payment-form">
+              <div id="card-element"></div>
+              <button id="submit-payment">
+                <div class="spinner hidden" id="spinner"></div>
+                <span id="button-text">Pay</span>
```
porueesq, line 83:
```suggestion
                <span id="button-text">Pay invoice</span>
```

lib/developer_portal/app/controllers/developer_portal/admin/account/invoices_controller.rb
damianpm: Yes, it appears for a few seconds, and then it vanishes

porueesq (changes requested):
@damianpm A question and a couple of small changes in microcopy.

lib/developer_portal/app/controllers/developer_portal/admin/account/invoices_controller.rb
porueesq: In this case, what can the user do to solve this error? 

If the user has alternatives to solve the error that triggers this message, I think the message should persist with a suggested solution. For example: _Failed to update payment transaction. Review your credit card details_

What do you think @damianpm ?

app/javascript/packs/invoice_payment.js
```diff
@@ -0,0 +1,87 @@
...
+    '::placeholder': {
+      color: '#32325d'
+    }
+  },
+  invalid: {
+    fontFamily: 'Arial, sans-serif',
+    color: '#fa755a',
+    iconColor: '#fa755a'
+  }
+}
```
hallelujah, line 19: I'd prefer css file because it won't be customized by customers otherwise

app/javascript/packs/invoice_payment.js
```diff
@@ -0,0 +1,87 @@
...
+  const callbackForm = document.querySelector('#payment-callback-form')
+  const payButton = document.querySelector('#submit-payment')
+  const cardError = document.querySelector('#card-error')
+  const errorMsg = document.querySelector('#card-error')
+  const spinner = document.querySelector('#spinner')
+  const buttonText = document.querySelector('#button-text')
+
+  const stripe = await loadStripe(stripePublishableKey)
+  const elements = stripe.elements()
+  const card = elements.create('card', { style })
```
hallelujah, line 35: Can we pass to stripe elements classes instead I think yes
https://stripe.com/docs/js/elements_object/create_element?type=card#elements_create-options

app/models/payment_intent.rb
```diff
@@ -11,8 +11,9 @@ class PaymentIntent < ApplicationRecord
   validates :reference, :state, length: { maximum: 255 }
   validates :reference, uniqueness: true
 
-  scope :latest, -> (count = 1) { reorder(created_at: :desc, id: :desc).limit(count) }
-  scope :latest_pending, -> (count = 1) { where.not(state: SUCCEEDED_STATES).latest(count) }
+  scope :latest, ->(count = 1) { reorder(created_at: :desc, id: :desc).limit(count) }
+  scope :pending, ->() { where.not(state: SUCCEEDED_STATES) }
+  scope :latest_pending, ->(count = 1) { pending.latest(count) }
```
hallelujah, line 16: I do not think we really need this scope because we can just do `latest(count).pending`

app/services/finance/stripe_payment_intent_update_service.rb
```diff
@@ -1,13 +1,13 @@
 # frozen_string_literal: true
 
 class Finance::StripePaymentIntentUpdateService
-  def initialize(provider_account, stripe_event)
-    @stripe_event = stripe_event
-    @payment_intent_data = stripe_event.data.object
-    @payment_intent = PaymentIntent.by_invoice(provider_account.buyer_invoices).find_by!(reference: payment_intent_data['id'])
+  def initialize(provider_account, source_object)
+    @source_object = source_object.to_hash
+    @payment_intent_data = @source_object.dig(:data, :object) || @source_object
```
hallelujah, line 6: Why is in some cases `@payment_intent_data == @source_object`? Not really obvious from here

lib/developer_portal/app/controllers/developer_portal/admin/account/invoices_controller.rb
```diff
@@ -1,34 +1,88 @@
...
     assign_drops invoices: invoices, pagination: pagination
   end
 
   def show
-    invoice = current_account.invoices.visible_for_buyer.find(params[:id])
     assign_drops invoice: Liquid::Drops::Invoice.wrap(invoice)
   end
 
+  def payment
+    @payment_intent = @invoice.payment_intents.latest_pending.first!
```
hallelujah, line 32: :warning: Objects leaked in the liquid page should be wrapped into a `Liquid::Drop` otherwise man can access to any ORM object in the database.
This is an issue because a provider can just access master and change whatever there ...

lib/developer_portal/app/controllers/developer_portal/admin/account/invoices_controller.rb
```diff
@@ -1,34 +1,88 @@
...
   def find_provider
     @provider = site_account
   end
+
+  def accessible_invoices
+    current_account.invoices.visible_for_buyer
+  end
+
+  def find_invoice
+    @invoice = accessible_invoices.find(params[:id])
```
hallelujah, line 70: Same security issue without `Liquid::Drop`

hallelujah (changes requested):
Please do not assign pure ORM objects in the controller, they need to be wrapped into drops
This is a security issue

lib/developer_portal/app/controllers/developer_portal/admin/account/invoices_controller.rb
hallelujah: Second thoughts, payment is an ERB template so it should be fine.

lib/developer_portal/config/routes.rb
```diff
@@ -65,7 +65,12 @@
 
       resource :personal_details, :only => [:show, :update]
       resource :password, :only => [:new, :show, :create, :update]
-      resources :invoices, :only => [:index, :show]
+      resources :invoices, :only => [:index, :show] do
+        member do
+          get :payment
+          post :payment_succeeded
```
hallelujah, line 71: It might not succeed so the naming is misleading

app/services/finance/stripe_payment_intent_update_service.rb
guicassolato: That is because of https://github.com/3scale/porta/pull/2350/commits/4110d34e3da592fcfe8a0cae83744e2be3caf0cc. The Stripe webhook sends a `Stripe::Event`, while our own callback after the checkout page sends a `Hash` of just the part of the payment intent (without being wrapped within a Stripe event).

lib/developer_portal/app/controllers/developer_portal/admin/account/invoices_controller.rb
guicassolato: Right... and easy fix.

lib/developer_portal/config/routes.rb
guicassolato: I had doubts myself and we can totally rename it just in case the behaviour changes in the future. Right now though the JS part won't call this action unless it was a success.

lib/developer_portal/app/controllers/developer_portal/admin/account/invoices_controller.rb
hallelujah: @porueesq good remark! But flash messages should be short (technically hard to put many characters there). We should find another way of doing this, probably validations or something else.

hallelujah (approved):
LGTM 🚀

lib/developer_portal/app/controllers/developer_portal/admin/account/invoices_controller.rb
porueesq: Yes @hallelujah , if the user needs to perform an action to correct the issue, we need another alternative to show the message.

app/javascript/packs/invoice_payment.js
damianpm: @hallelujah 
In fact, we can totally omit to pass the style or any classes in the options object. It will use default stripe styles (injected from stripe script). 
If we want to allow customization, we need to discuss how to do it. We are not using a liquid template, but an `.erb`, so it's not going to be so easy to add an external stylesheet. I think customization of css should be a follow-up task.
WDYT?

damianpm: BTW, the styles we have in `app/javascript/src/PaymentGateways/stripe/styles/stripe.scss` are for the placement of the form in our page. The ones passed to `elements.create` affect only the _card element_ inside the form. So, no confict.

lib/developer_portal/app/controllers/developer_portal/admin/account/invoices_controller.rb
porueesq: @damianpm Hence, the flash message is for the Admin Portal, and the screenshots are for the Developer Portal, right?

If these are correct, then no additional changes are necessary :+1:

porueesq (approved):
LGTM! :clap:

damianpm:
> LGTM 🚀
> 
> I am approving because I consider the CSS issues as second order problem we can address in another PR

@hallelujah but you didn't approve ;)

lib/developer_portal/app/controllers/developer_portal/admin/account/invoices_controller.rb
```diff
@@ -1,25 +1,51 @@
...
 
   def show
-    invoice = current_account.invoices.visible_for_buyer.find(params[:id])
-    assign_drops invoice: Liquid::Drops::Invoice.wrap(invoice)
+    assign_drops invoice: Liquid::Drops::Invoice.wrap(find_invoice)
+  end
+
+  def payment
+    @invoice = find_invoice
+    payment_intent = @invoice.payment_intents.pending.latest.first!
```
hallelujah, line 32: Can't we create the payment intent here if we do not find any?

guicassolato: We can. In another PR though, I'd say.

hallelujah: I think you do not want to create payment intents every time you display the payment page …

OK here should be the flow and I see the struggle because it is more complicated:

- On click on the “Pay” button
- send an AJAX request to create or retrieve a payment intent for this invoice
- get the response with the payment intent secret (https://stripe.com/docs/api/errors#errors-payment_intent-client_secret) and the publishable key and return them in the response
- use those keys in your stripe JS configuration
- submit Stripe Elements
- receive Stripe response and update the fields in the form
- submit the form to `InvoicesController#payment_succeeded`

guicassolato: (I took the liberty to number your bullets above.)

For step 2, we need to be 100% certain not to end up confirming the payment intent then, just create it (in case it cannot be retrieved) without auto confirmation.

hallelujah: Once it is released we cannot really do another PR because it involves changing customer front end again

hallelujah (approved):
Up to you to merge or not considering the comments left.

Remember though it will be difficult to do another PR on top of that once the `pay_now?` is used by customers

lib/developer_portal/app/controllers/developer_portal/admin/account/invoices_controller.rb
guicassolato: I disagree. The customer can remove the `pay_now?` check _if_ they want to, and actually there are good reasons not to as well.

hallelujah: OK then :)

guicassolato (approved):
I shouldn't be approving myself as some of the commits are mine. Just a way to express my opinion that we should move with the PR and merge it.

I don't think the issue with releasing the `pay_now?` liquid method now is such a big of a deal. Customers may prefer not to default to enabling manual payment on invoices that never failed and this method will continue to be useful then. On the other hand, if we understand that customers should not be able to control this, then the method can change to `def pay_now?; true; end` and it's done.

The rest about creating or fetching payment intents with Ajax instead of as now, immediately as part of the request to the `payment` action, can be done transparently to the users while this page is an erb instead of a liquid.

PR #2355, opened 2021-02-03
Fixes/Cleanup in Vertical Nav

josemigallas:
* clean up
* move dataset generation into helper
* fixes console warning
* fixes Flow types

### DIFF
--- a/app/helpers/menu_helper.rb
+++ b/app/helpers/menu_helper.rb
@@ -101,6 +101,6 @@ module MenuHelper
   end
 
   def current_api
-    (@backend_api || @service || Service.none).decorate
+    @backend_api || @service
   end
 end
--- a/app/helpers/vertical_nav_helper.rb
+++ b/app/helpers/vertical_nav_helper.rb
@@ -1,6 +1,15 @@
 # frozen_string_literal: true
 
 module VerticalNavHelper
+  def vertical_nav_data
+    {
+      'current-api': current_api.to_json,
+      sections: vertical_nav_sections.to_json,
+      'active-section': active_submenu,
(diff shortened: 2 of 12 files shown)


app/helpers/menu_helper.rb
```diff
@@ -101,6 +101,6 @@ def settings_link
   end
 
   def current_api
-    (@backend_api || @service || Service.none).decorate
```
thomasmaas, line 104: isn't `Service.none` there for a reason?

josemigallas: it's supposed to return `nil` when there is no service. `Service.None` returns an empty array and mess things up. AFAIR `current_api` is used nowhere else.

thomasmaas: But given it's a helper method it can be called from any view so I think it should be robust enough to deal with a no service situation. But probably input needed from ruby engineer.

didierofrivia: it might have been because `nil` raised from where `current_api` was called... I agree that an empty array is no way an indicator of the current api but I'd double check the usage to avoid panic.

didierofrivia (approved):
Before merging check the usage of the helper method `current_api`

app/helpers/menu_helper.rb
josemigallas: @thomasmaas a no service situation would be @nil, since this method is supposed to return 1 instance of BackendApi. For a method that returns a collection, `Service.none` would be appropriate.

I double checked it and it's only used by `vertical_nav`, so we're safe.

didierofrivia (approved):

PR #2356, opened 2021-02-04
[asdf] For those using asdf

hallelujah:
In order to have the same local development environment, some people use https://github.com/asdf-vm/asdf

This PR is to help them to know which version of Ruby and NodeJS we use

Ruby 2.6.6
Node v14.15.4


.tool-versions.sample
```diff
@@ -0,0 +1,2 @@
+ruby 2.6.6
```
guicassolato, line 1: Should we start with v2.5 as it's the one we have the tests running on or better push dev to 2.6 already and thus force the upgrade once things start breaking in CI?

hallelujah: I'd try to leave old versions and push as much to the edge ;)
As long as it is working it is a good way also to push for 2.6 in production

guicassolato (approved):

PR #2358, opened 2021-02-09
Braintree transaction source recurring

guicassolato:
- Adds `transaction_source: 'recurring'` to purchase with braintree
- New service `Finance::ChargingService` (to remove charign logic from `PaymentTransaction`


hallelujah (approved):
Thank you!

PR #2359, opened 2021-02-10
Adding missing title for line items in Invoices

damianpm:

When creating an invoice manually, the Line Items section is missing a title

None, but follow-up of https://github.com/3scale/porta/pull/2350

**Verification steps** 

- Go to `/buyers/accounts/[id]/invoices`:
   link
- Create an Invoice
- Add line items

**Before**:
Screenshot 2021-02-10 at 14 07 31

**After**:
Screenshot 2021-02-10 at 14 07 53

### DIFF
--- a/app/lib/pdf/finance/invoice_generator.rb
+++ b/app/lib/pdf/finance/invoice_generator.rb
@@ -102,7 +102,7 @@ module Pdf
       end
 
       def print_line_items
-        @pdf.text "Line items"
+        subtitle('<b>Line items</b>')
         opts = { width: TABLE_FULL_WIDTH, headers: InvoiceReportData::LINE_ITEMS_HEADING }
         @pdf.table(@data.line_items, @style.table_style.merge(opts))
         move_down
--- a/app/lib/pdf/finance/invoice_generator.rb
+++ b/app/lib/pdf/finance/invoice_generator.rb
@@ -102,6 +102,7 @@ module Pdf
       end
 
       def print_line_items
+        @pdf.text "Line items"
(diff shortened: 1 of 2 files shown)


thomasmaas:
And this title does exist for automatically generated invoices @damianpm ?

cntlsn:
Thank you @damianpm, a quick UI comment: I find that using the same style for the card header and the table header is a bit confusing. Would it be possible for the table header to use the same style we are using in other pages?

app/lib/pdf/finance/invoice_generator.rb
```diff
@@ -102,6 +102,7 @@ def print_details
       end
 
       def print_line_items
+        @pdf.text "Line items"
```
thomasmaas, line 105: use `subtitle` like above?

damianpm: No need:
Screenshot 2021-02-10 at 16 45 09

thomasmaas: Visually not but if you would parse the pdf with a screenreader it would be better as it adds semantics. And if one we decide to style sub headers a certain way too.

cntlsn (dismissed):

app/lib/pdf/finance/invoice_generator.rb
```diff
@@ -102,6 +102,7 @@ def print_details
       end
 
       def print_line_items
+        subtitle('<b>Line items</b>')
```
damianpm, line 105: @thomasmaas you are right, using `subtitle`

thomasmaas (approved):
Sorry for being a pain in the ass. Just want us to be as good as we can be and not cut corners.

cntlsn (approved):

PR #2362, opened 2021-02-15
Upgrade Flow to lastest

josemigallas:
## Description
Upgrades Flow and Flow-typed to the latest version.

The new version of Flow includes new "ignore" comments among other things and allows to ignore specific erorrs (no more unsafe $FlowFixMe everywhere).

## Verification
Run `yarn lint`. No errors should pop up.

PR #2364, opened 2021-02-17
[2.11] New Application page remake

josemigallas:

* Reimplements New Application page in React
* Adds a "Create Application" button to all application indexes:
  - /apiconfig/services/:id/applications
  - /buyers/accounts/:id/applications
  - /buyers/applications

[THREESCALE-6878: Improve Select in 'Create new Application' form](https://issues.redhat.com/browse/THREESCALE-6878)

[THREESCALE-6878: Add 'Create new Application' flow to Audience > Applications index](https://issues.redhat.com/browse/THREESCALE-6876)

[THREESCALE-6879: Add 'Create new Application' flow to Product > Applications index](https://issues.redhat.com/browse/THREESCALE-6879)


app/assets/stylesheets/provider/_buttons.scss
```diff
@@ -164,3 +164,19 @@ form.formtastic .button-bar, .button-bar {
   padding: line-height-times(1/4) line-height-times(1/3) !important;
   height: line-height-times(5/4);
 }
+
+/****** HACK ******/
+// This fixes missing PF4
```
hallelujah, line 169: Add a TODO? so this can be grepped later?

app/views/api/applications/new.html.slim
```diff
@@ -0,0 +1,5 @@
+h2 New Application
+
+div id="new-application-form" data=(new_application_form_metadata(provider: @provider, service: @service))
```
hallelujah, line 3: Why don't you just use  `current_account` instead of `@provider`
No need of defining `@provider` in the controller

It is defined as a helper in https://github.com/3scale/porta/blob/06215fac9733963435a266c60ead0bde51b3dff8/app/lib/authenticated_system.rb#L12

app/controllers/buyers/applications_controller.rb
```diff
@@ -6,8 +8,9 @@ class Buyers::ApplicationsController < FrontendController
   helper DisplayViewPortion::Helper
 
   before_action :authorize_partners
-  before_action :find_buyer, :only => [:new, :create]
-  before_action :authorize_multiple_applications, :only => [ :new, :create ]
+  before_action :find_buyer, :only => [:create]
```
hallelujah, line 11: Why not `:new`?

josemigallas: Because now buyer is not always available and it will return 404.

test/unit/helpers/buyers/applications_helper_test.rb
```diff
@@ -1,66 +1,24 @@
...
-    hash[application_plan.id] = service.id
-    assert_equal relation_plans_services(provider), hash.to_json
-  end
-
-  test "relation_plans_services should return a empty hash" do
-    assert_equal relation_plans_services(Account.providers.new), {}.to_json
+    assert new_application_form_metadata(provider: provider)
+    assert new_application_form_metadata(provider: provider, buyer: buyer)
+    assert new_application_form_metadata(provider: provider, service: service)
+    assert new_application_form_metadata(provider: provider, buyer: buyer, service: service)
```
hallelujah, line 21: Could you compare returned values and make the test better understandable?
`assert` checks for truthy values not really useful here

test/unit/helpers/buyers/applications_helper_test.rb
```diff
@@ -1,66 +1,24 @@
...
-
-    assert_equal service_plan_contracted_for_service(buyer), hash.to_json
-  end
-
-  test "service_plan_contracted_for_service should return a empty hash" do
-    assert_equal service_plan_contracted_for_service(Account.providers.new), {}.to_json
+  test 'new_application_form_metadata requires a provider' do
+    assert_raise ArgumentError do
+      new_application_form_metadata
+    end
```
hallelujah, line 10: Well this is not needed :) It is how Ruby works with keywords argument, you do not need to test Ruby

hallelujah (changes requested):

test/unit/helpers/buyers/applications_helper_test.rb
josemigallas: To be honest I don't even think these tests make any sense. It is impossible to compare with all possible return values.. IMO this is covered already with integration tests and cucumbers, since the React won't work unless the metadata is correct.

If you're OK with it I would remove it.. I just updated it because somebody added it long ago but like I said, I don't think it makes sense anymore.

app/assets/stylesheets/provider/_buttons.scss
josemigallas: This is an alternative to TODO we've been using for a while since @didierofrivia introduced it. If you grep `* HACK` you'll see there are some already.

app/javascript/packs/PF4Styles/application.js
```diff
@@ -0,0 +1,8 @@
+import 'patternflyStyles/modal'
```
damianpm, line 1: I know we've been importing selectively, but do you think we may import the whole PF css file? This is growing too much

app/javascript/packs/new_application.js
```diff
@@ -0,0 +1,41 @@
...
+
+  const { createServicePlanPath, createApplicationPath, createApplicationPlanPath, serviceSubscriptionsPath } = dataset
+  const product = safeFromJsonString(dataset.product)
+  const products = safeFromJsonString(dataset.products)
+  const servicePlansAllowed = safeFromJsonString(dataset.servicePlansAllowed)
+  const buyer = safeFromJsonString(dataset.buyer)
+  const buyers = safeFromJsonString(dataset.buyers)
+  const definedFields = safeFromJsonString(dataset.definedFields)
+  const validationErrors = safeFromJsonString(dataset.errors) || {}
+  console.log(definedFields) // TODO need definedFields on all contexts
```
damianpm, line 25: We have added `no-console` rule

app/javascript/src/Common/components/UserDefinedField.jsx
```diff
@@ -0,0 +1,65 @@
...
+      label={label}
+      isRequired={required}
+      fieldId={id}
+      name={name}
+      item={item}
+      items={choices.map(str => ({
+        id: str,
+        name: str
+      }))}
+      placeholderText=""
```
damianpm, line 39: Is it required to pass `placeholderText` when empty string?
I think is not
```suggestion
```

app/javascript/src/NewApplication/components/NewApplicationForm.jsx
```diff
@@ -0,0 +1,184 @@
...
+  const definedFieldsInitialState = definedFields ? definedFields.reduce((state, field) => {
+    state[field.id] = ''
+    return state
+  }, {}) : {}
+  const [definedFieldsState, setDefinedFieldsState] = useState<{[string]: string}>(definedFieldsInitialState)
+  const handleOnDefinedFieldChange = (id) => (value) => {
+    setDefinedFieldsState(state => ({ ...state, [id]: value }))
+  }
+
+  const [loading, setLoading] = useState<boolean>(false)
```
damianpm, line 67: I'd move this to line 57

app/javascript/src/NewApplication/index.js
```diff
@@ -0,0 +1,10 @@
+// @flow
+
+export * from './components/BuyerSelect'
+export * from './components/ServicePlanSelect'
+export * from './components/ProductSelect'
+export * from './components/ApplicationPlanSelect'
+// $FlowFixMe $FlowError[cannot-resolve-module] failing because file name contains module's name NewApplication
```
damianpm, line 7: Will https://github.com/3scale/porta/pull/2362 fix all this $FlowFixMe issues?
If true, we should first merge that, and rebase here

app/assets/stylesheets/provider/_buttons.scss
didierofrivia: It started because we will modifying vendor code, mainly patternfly and webpacker

app/javascript/packs/PF4Styles/application.js
josemigallas: Adding all PF css in the main layout would be ideal but right now it messes with lots of elements. I think even if it's growing a lot we have fine control over our styles.

app/assets/stylesheets/provider/_buttons.scss
hallelujah: OK but please share back to the community by:

1. Opening an issues in the upstream
2. Opening a PR
3. Point the link in this PR so there are cross-references

app/views/buyers/applications/new.html.slim
```diff
@@ -0,0 +1,8 @@
+- if @account
+  = render '/buyers/accounts/menu'
```
hallelujah, line 2: Hmmm this is strange :) why `@account` is the condition? Not explicit enough, especially when you do not use it anymore in the rest of the file

app/views/buyers/applications/new.html.slim
```diff
@@ -0,0 +1,8 @@
...
+
+h2 New Application
+
+div id="new-application-form" data=(new_application_form_metadata(provider: current_account, buyer: @buyer, cinstance: @cinstance))
+  = javascript_pack_tag 'new_application'
+  = javascript_pack_tag 'PF4Styles/application'
```
hallelujah, line 8: Those JS are dependent on the wrapping div??

features/old/applications/providers/create_application.feature
```diff
@@ -1,4 +1,4 @@
-@ignore-backend
+@javascript @ignore-backend
```
hallelujah, line 1: The rule of thumb. When touching old cucumber, we migrate to the new way

features/step_definitions/custom_web_steps.rb
```diff
@@ -106,8 +105,16 @@ def matches_path?(url, path)
...
 
 Then /^the "([^"]*)" select should not contain "([^"]*)" option$/ do |label, text|
-  select = find_field(label)
-  assert select.all('option', :text => text).empty?, %(The "#{label}" select should not contain "#{text}" option, but it does)
+  if page.has_css?('.pf-c-form__label', text: label)
+    select = find('.pf-c-form__label', text: label).sibling('.pf-c-select')
+    select.find('.pf-c-select__toggle-button').click
+    selector = '.pf-c-select__menu-item'
+  else
+    # DEPRECATED: remove when all selects have been replaced for PF4
```
hallelujah, line 113:
```suggestion
    # DEPRECATED: remove when all selects have been replaced for PF4
    ThreeScale::Deprecation.warn "[cucumber] Detected a select not using PF4 css"
```

features/step_definitions/web_steps.rb
```diff
@@ -39,12 +39,24 @@ def with_scope(locator)
 
 When /^(?:|I )fill in "([^"]*)" with "([^"]*)"(?: within "([^"]*)")?$/ do |field, value, selector|
   with_scope(selector) do
-    fill_in(field, :with => value, visible: true)
+    if page.has_css?('.pf-c-form__label', text: field)
+      input = find('.pf-c-form__label', text: field).sibling('input')
+      input.set value
+    else
+      # DEPRECATED: remove when all forms implement PF4
```
hallelujah, line 46:
```suggestion
      # DEPRECATED: remove when all forms implement PF4
      ThreeScale::Deprecation.warn "[cucumber] Detected a form not using PF4 css"
```

app/assets/stylesheets/provider/_buttons.scss
josemigallas: The way I see it, there are already tags all over the codebase: TODO, FIXME, HACK... They exists as a reminder that something needs to be done, eventually. There is no point in creating JIRAs or issues that will be closed a month after because nobody took them.

IMHO what we should do is to regularly (once a month maybe) take a look at the existing tags and address 1 or 2. This is a globally accepted convention so it should be easy for the community to identify and fix them.

features/step_definitions/web_steps.rb
```diff
@@ -39,12 +39,24 @@ def with_scope(locator)
...
   end
 end
 
 When /^I fill in "(.+?)" with:$/ do |field, text|
-  fill_in(field, :with => text, visible: true)
+  if page.has_css?('.pf-c-form__label', text: field)
+    input = find('.pf-c-form__label', text: field).sibling('input')
+    input.set value
+  else
+    # DEPRECATED: remove when all forms implement PF4
```
hallelujah, line 57:
```suggestion
    # DEPRECATED: remove when all forms implement PF4
    ThreeScale::Deprecation.warn "[cucumber] Detected a form not using PF4 css"
```

features/step_definitions/web_steps.rb
```diff
@@ -74,11 +86,20 @@ def with_scope(locator)
...
-    else # this is selenium, needs slightly different treatment
-      select.find(:xpath, XPath::HTML.option(value)).click
+    if page.has_css?('.pf-c-form__label', text: field)
+      select = find('.pf-c-form__label', text: field).sibling('.pf-c-select')
+      within select do
+        find('.pf-c-select__toggle-button').click unless select['class'].include?('pf-m-expanded')
+        click_on(value, exact: true)
+      end
+    else
+      # DEPRECATED: remove when all selects have been replaced for PF4
```
hallelujah, line 96:
```suggestion
      # DEPRECATED: remove when all selects have been replaced for PF4
      ThreeScale::Deprecation.warn "[cucumber] Detected a form not using PF4 css"
```

hallelujah (changes requested):

app/helpers/buyers/applications_helper.rb
```diff
@@ -1,49 +1,103 @@
...
+      'create-service-plan-path': new_admin_service_service_plan_path(':id'),
+      'service-subscriptions-path': admin_buyers_account_service_contracts_path(':id'),
+      'service-plans-allowed': provider.settings.service_plans.allowed?.to_json,
+      product: service && product_data(service).to_json,
+      products: !service && data_products(provider).to_json,
+      buyer: buyer && buyer_data(buyer).to_json,
+      buyers: !buyer && data_buyers(provider).to_json,
+      'defined-fields': data_defined_fields(provider).to_json,
+      errors: cinstance&.errors.to_json
+    }.compact
```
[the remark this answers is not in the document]
josemigallas: Yes, it gives a bit of a JS headache otherwise.

features/old/applications/providers/edit_application.feature
```diff
@@ -0,0 +1,60 @@
+@ignore-backend
```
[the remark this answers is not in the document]
josemigallas: Fair enough

app/helpers/buyers/applications_helper.rb
```diff
@@ -1,49 +1,103 @@
...
+      'create-application-path': buyer ? admin_buyers_account_applications_path(buyer) : admin_buyers_applications_path,
+      'create-application-plan-path': new_admin_service_application_plan_path(':id'),
+      'create-service-plan-path': new_admin_service_service_plan_path(':id'),
+      'service-subscriptions-path': admin_buyers_account_service_contracts_path(':id'),
+      'service-plans-allowed': provider.settings.service_plans.allowed?.to_json,
+      product: service && product_data(service).to_json,
+      products: !service && data_products(provider).to_json,
+      buyer: buyer && buyer_data(buyer).to_json,
+      buyers: !buyer && data_buyers(provider).to_json,
+      'defined-fields': data_defined_fields(provider).to_json,
```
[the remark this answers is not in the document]
josemigallas: Yes it's either this or to call `as_json` on every attribute and then `to_json` on the hash.

app/controllers/buyers/applications_controller.rb
```diff
@@ -78,6 +85,7 @@ def create
       redirect_to(admin_service_application_path(@cinstance.service, @cinstance))
     else
       @cinstance.extend(AccountForNewPlan)
+      # FIXME: It always redirects to /buyers/applications but it should redirect to the page it came from.
```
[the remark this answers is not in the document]
josemigallas: There are 3 index pages for applications, depending on the context. I think rails should be clever enough to know where to redirect to, or maybe we should have 3 different controllers. However I don't think we should "workaround" it by using session storage.

features/old/applications/providers/create_application.feature
josemigallas: WDYM the new way?

app/assets/stylesheets/provider/_buttons.scss
hallelujah: I mean open an issue in PF4 not here :)

app/controllers/buyers/applications_controller.rb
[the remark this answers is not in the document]
hallelujah: This is not about rails this is about your flow and a problem with any web application

app/helpers/buyers/applications_helper.rb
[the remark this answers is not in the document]
hallelujah: You can do as_json on the general Hash, it will recursively call `as_json` to all the values

```rub
# activesupport-5.0.7.2/lib/active_support/core_ext/object/json.rb:147:
# Owner: Hash
# Visibility: public
# Signature: as_json(options=?)
# Number of lines: 16

def as_json(options = nil) #:nodoc:
  # create a subset of the hash by applying :only or :except
  subset = if options
    if attrs = options[:only]
      slice(*Array(attrs))
    elsif attrs = options[:except]
      except(*Array(attrs))
    else
      self
    end
  else
    self
  end

  Hash[subset.map { |k, v| [k.to_s, options ? v.as_json(options.dup) : v.as_json] }]
end
```

test/unit/helpers/buyers/applications_helper_test.rb
hallelujah: I did not see any integration test that makes use of this.
I saw:

- React Test but not calling the ruby method
- Cucumber test that tested one part of it not all possible values

If you think they do not make sense because there is too much data in them then  you could just stub `buyer_data`, `product_data` etc ... to make them smaller.

```ruby
buyer.stub(buyer_data: {foo: :bar})
product.stub(product_data: {greeting: :hello}
# etc...
```

app/helpers/buyers/applications_helper.rb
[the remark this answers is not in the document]
josemigallas: Nevermind, it has to be done this way in order to have the correct attribute names (kebab-case to camelCase transformation).

test/unit/helpers/buyers/applications_helper_test.rb
josemigallas: Cucumber test won't pass if the helper doesn't work correctly, because the React component won't render. I think it's enough with the cukes, it doesn't make sense to assert the return of a helper that's only supposed to arrange data in a certain way for the views.

app/decorators/service_decorator.rb
```diff
@@ -47,6 +47,18 @@ def api_selector_api_link
 
   alias link api_selector_api_link
 
+  def new_application_data
+    {
+      id: id.to_s,
+      name: name,
+      systemName: system_name,
+      updatedAt: updated_at,
+      appPlans: object.plans.stock.select(:id, :name).as_json(root: false),
```
hallelujah, line 56: I think `object` is not necessary but it works too :)
You generally use `#object` when there is conflicting methods inside the decorator (like you defined method `#plans`

app/decorators/buyer_decorator.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+class BuyerDecorator < AccountDecorator
```
hallelujah, line 3: :heart:  I love that

app/helpers/buyers/applications_helper.rb
```diff
@@ -1,49 +1,22 @@
...
+      'create-service-plan-path': new_admin_service_service_plan_path(':id'),
+      'service-subscriptions-path': admin_buyers_account_service_contracts_path(':id'),
+      'service-plans-allowed': provider.settings.service_plans.allowed?.to_json,
+      product: service && ServiceDecorator.new(service).new_application_data.to_json,
+      products: !service && provider.application_products_data.to_json,
+      buyer: buyer && BuyerDecorator.new(buyer).new_application_data.to_json,
+      buyers: !buyer && provider.application_buyers_data.to_json,
+      'defined-fields': provider.application_defined_fields_data.to_json,
+      errors: cinstance&.errors.to_json
+    }.compact
```
hallelujah, line 19: This can be put in the ProviderDecorator, to use the route helpers use https://github.com/3scale/porta/blob/master/app/lib/system/url_helpers.rb

```ruby
class ProviderDecorator < AccountDecorator
   include System::UrlHelpers.system_url_helpers
end
```

app/views/api/applications/index.html.slim
```diff
@@ -1,2 +1,4 @@
-- content_for :sublayout_title, 'Applications'
-= render '/buyers/applications/listing'
+h1 Applications
+a.pf-c-button.pf-m-primary.pf-c-button__in-title href=new_admin_service_application_path(@service) Create Application
```
hallelujah, line 2: Well though that is OK, you can use 
```suggestion
= link_to 'Create Application', new_admin_service_application_path(@service), class: %w[pf-c-button pf-m-primary pf-c-button__in-title]
```

app/views/buyers/applications/index.html.slim
```diff
@@ -1,17 +1,30 @@
...
+    = @plan.name
+  - elsif @service
+    ' on service
+    = @service.name
+  - elsif @account
+    ' for
+    = @account.name
+
+- if !@account
+  a.pf-c-button.pf-m-primary.pf-c-button__in-title href=new_admin_buyers_application_path Create Application
```
hallelujah, line 17: Same can use `= link_to ...`

features/buyers/applications/new.feature
```diff
@@ -0,0 +1,167 @@
...
+Feature: Create application
+  In order to control the way my buyers are using my API
+  As a provider
+  I want to create their applications
+
+  Background:
+    Given a provider "foo.3scale.localhost"
+    And provider "foo.3scale.localhost" uses backend v2 in his default service
+    And provider "foo.3scale.localhost" has multiple applications enabled
+    And provider "foo.3scale.localhost" has "service_plans" switch allowed
```
hallelujah, line 11: Basically knowing the domain `foo.3scale.localhost` is not really helpful in any case. 
Following comment applies to the whole file not only to `foo.3scale.localhost`


The reader needs to know what and not the implementation details. We try to remove useless information from cucumber and focus on readable business logic https://cucumber.io/docs/bdd/better-gherkin/

Cucumber should be readable and get to the point of the feature tested, hiding all small steps is the task of the developer not the reader.

A good example is from this file: https://github.com/3scale/system/pull/6977/files#diff-866bfaaad5caa1bcbeda34bcd98bbf6e80457aa4d26c47dedbe0f09811a56ade

To this file https://github.com/3scale/system/pull/6977/files#diff-1900a5761627b919a04a7650ffbb6c57ce4aa0051941bdbb3054dedd8e928acd

But I understand it is much work for already existing cucumbers. You can do them on another PR. For new cucumbers however please do the new correct way.

test/unit/helpers/buyers/applications_helper_test.rb
hallelujah: I am not a big fan of helpers arranging data ... Too much logic for a helper.
The goal of a helper is building something for the view without logic linked to the data.
Once you have to check a state of the data in the helper then it is not anymore a helper.

Thus putting it in a decorator or another pattern. And you can test this pattern object with stubs (to make your test easier)

app/javascript/packs/new_application.js
```diff
@@ -0,0 +1,41 @@
...
+
+  const { createServicePlanPath, createApplicationPath, createApplicationPlanPath, serviceSubscriptionsPath } = dataset
+  const product = safeFromJsonString(dataset.product)
+  const products = safeFromJsonString(dataset.products)
+  const servicePlansAllowed = safeFromJsonString(dataset.servicePlansAllowed)
+  const buyer = safeFromJsonString(dataset.buyer)
+  const buyers = safeFromJsonString(dataset.buyers)
+  const definedFields = safeFromJsonString(dataset.definedFields)
+  const validationErrors = safeFromJsonString(dataset.errors) || {}
+  const error: string | void = validationErrors.hasOwnProperty('base') ? validationErrors.base[0] : undefined
```
damianpm, line 25: Is not enough `error:  ?string` ?

damianpm (commented):
Please rebase master

app/javascript/packs/new_application.js
josemigallas: Didn't work well the first time but I'll try that again.

features/buyers/applications/new.feature
josemigallas: OK in that case, I am going to restore it and leave inside `old`. That way at least we know the Feature needs to be properly refactored.

test/unit/helpers/buyers/applications_helper_test.rb
josemigallas: OK I moved it to the decorator and improved the tests a little bit

app/javascript/packs/new_application.js
josemigallas: OK I thought it didn't work but the problem is we don't want a maybe type here, we want either a string or undefined because optional props `error?: string` don't accept it. I've found the whole `?type` thing is prone to error and in the end it's easier to just be explicit.

josemigallas:
> Please rebase master

Conflicts resolved

app/views/buyers/applications/new.html.slim
josemigallas: If there's an account, print the account's menu.

app/views/buyers/applications/new.html.slim
josemigallas: First one loads the React component, whereas the second adds Patternfly styles.

app/controllers/buyers/applications_controller.rb
```diff
@@ -6,8 +8,9 @@ class Buyers::ApplicationsController < FrontendController
   helper DisplayViewPortion::Helper
 
   before_action :authorize_partners
-  before_action :find_buyer, :only => [:new, :create]
-  before_action :authorize_multiple_applications, :only => [ :new, :create ]
+  before_action :find_buyer, :only => :create
```
guicassolato, line 11:
```suggestion
  before_action :find_buyer, only: :create
```

app/controllers/buyers/applications_controller.rb
```diff
@@ -6,8 +8,9 @@ class Buyers::ApplicationsController < FrontendController
...
+  before_action :find_plans, :only => %i[new create]
```
guicassolato, line 12:
```suggestion
  before_action :find_plans, only: %i[new create]
```

app/controllers/buyers/applications_controller.rb
```diff
@@ -6,8 +8,9 @@ class Buyers::ApplicationsController < FrontendController
...
+  before_action :authorize_multiple_applications, :only => [:create]
```
guicassolato, line 13:
```suggestion
  before_action :authorize_multiple_applications, only: [:create]
```

app/controllers/buyers/applications_controller.rb
```diff
@@ -52,25 +55,28 @@ def index
...
-    @plans = accessible_plans.stock
+    @products = accessible_services
 
     if params[:account_id]
-      @account = current_account.buyers.find params[:account_id]
-      activate_menu :buyers, :accounts
+      # We're under Account context
+      find_buyer
+      @cinstance = @account.bought_cinstances.build
+      extend_cinstance_for_new_plan
```
guicassolato, line 64: Damn. This is hacky. Can we use a decorator instead?

app/controllers/buyers/applications_controller.rb
```diff
@@ -52,25 +55,28 @@ def index
...
+      activate_menu :buyers, :accounts, :listing
+    else
+      # We're under Applications context
+      activate_menu :audience, :applications, :listing
     end
```
guicassolato, line 69: I was thinking about making this more readable. WDYT?

```ruby
  def create
    @products = accessible_services

    return activate_menu :audience, :applications, :listing if applications_context?

    find_buyer
    @cinstance = @account.bought_cinstances.build
    extend_cinstance_for_new_plan

    activate_menu :buyers, :accounts, :listing
  end

  protected

  def buyers_context?
    params[:account_id]
  end

  def applications_context?
    !buyers_context?
  end
```

app/controllers/buyers/applications_controller.rb
```diff
@@ -177,7 +183,11 @@ def find_cinstance
   end
 
   def find_buyer
-    @buyer = current_account.buyers.find(params[:account_id])
+    @account = current_account.buyers.find(params[:account_id])
```
guicassolato, line 186: Would it be acceptable if we did
```suggestion
    account_id = params[:account_id]
    return unless account_id
    @account = current_account.buyers.find(account_id)
```
and then extended the `before_action` back to covering more actions (create, new, index)?

app/controllers/buyers/applications_controller.rb
josemigallas: I don't think we need both context mehods though

app/controllers/buyers/applications_controller.rb
josemigallas: I guess it would. I don't like hiding things though!

hallelujah (approved):

app/helpers/patternfly_helper.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+module PatternflyHelper
+  def pf_button_in_title(label, href)
+    link_to label, href, class: %w[pf-c-button pf-m-primary pf-c-button__in-title]
```
damianpm, line 5: We have some pf button helpers [here](https://github.com/3scale/porta/blob/master/app/helpers/buttons_helper.rb#L147)
Maybe you can add it in same place

josemigallas: Yes, I saw them and my idea was to move them inside this new helper. Otherwise it's a bit of a pain to maintain.. if we're to move everything to PF, better start cleaning up the older helpers.

hallelujah (approved):
Go for it, we can make improvements based on this :muscle:

PR #2365, opened 2021-02-17
Replace context selector classes with Patternfly equivalents

srambach:

This PR replaces custom styling classes with the Patternfly equivalents for the context selector in the header. It also updates the styling to match current Patternfly.

https://issues.redhat.com/browse/THREESCALE-6693

**Verification steps** 

Use the content selector in the masthead to verify it still operates. Inspect the elements to see that the context selector Patternfly classes are now included instead of custom "PopNavigation" classes.

image

### DIFF
--- a/app/javascript/src/patternflyStyles/pf4Base.css
+++ b/app/javascript/src/patternflyStyles/pf4Base.css
@@ -7,6 +7,8 @@
 
 @import '~@patternfly/patternfly/components/Page/page.css';
 @import '~@patternfly/patternfly/components/Nav/nav.css';
+/* @import '~@patternfly/patternfly/components/ContextSelector/context-selector.css'; */
+@import '~@patternfly/patternfly/layouts/Level/level.css';
 @import '~@patternfly/patternfly/patternfly-fonts.css';
 
 /****** HACK ******/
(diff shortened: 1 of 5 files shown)


app/javascript/src/patternflyStyles/pf4Base.css
```diff
@@ -7,6 +7,8 @@
 
 @import '~@patternfly/patternfly/components/Page/page.css';
 @import '~@patternfly/patternfly/components/Nav/nav.css';
+/* @import '~@patternfly/patternfly/components/ContextSelector/context-selector.css'; */
```
damianpm, line 10: Do we need this commented line?

damianpm (changes requested):
Can you please update the snapshots?
You can do it running:

`yarn jest -- -u `

app/javascript/src/patternflyStyles/pf4Base.css
srambach: @damianpm I think that might actually be a better place for it so I put it back and removed the other include. Good catch. 
On the tests, jest still fails tests and doesn't create new snapshots. The purpose of this PR was an improvement for QE, so maybe tests need to be rewritten to use the PF classes? That's not my area, so I'm not sure.

damianpm: OK, I'll merge to a different branch, fix the tests, and then we can merge to master

damianpm (approved):
Merging to `context-selector-qe-classes`, then fixing tests

damianpm:
@srambach check https://github.com/3scale/porta/pull/2373

srambach:
@damianpm Thanks!

PR #2367, opened 2021-02-18
[2.11] New Select with Modal component

josemigallas:

Implements a Select form group with an embedded modal featuring:
* PF4 styles
* Typeahead
* Inside modal:
  * Searches by name, searches on "Enter" keydown
  * Button is disabled until an item is selected
  * Fully paginated
  * When clicking the X, search results are reset (unlike in the gif)

select

modal

It's the base for all 2.11 tickets.


cntlsn (commented):
Thank you @josemigallas , I have a couple of questions:
* is a limit set to only display 5 items per page?
* what's the deal with the clearing action on the input field? clicking on the 'X' icon only clears the input field but doesn't actually bring back all the items of the table until the user submits an empty search? if that's the case, then I find it quite unintuitive. if we remove the 'X' icon and its clearing action, would the user still need to enter an empty search, or manually deleting the term in the input field would do the job?

josemigallas:
> Thank you @josemigallas , I have a couple of questions:
> 
> * is a limit set to only display 5 items per page?

Yes, that's the number in the spec. It is not fixed though, 5 is only the default.

> * what's the deal with the clearing action on the input field? clicking on the 'X' icon only clears the input field but doesn't actually bring back all the items of the table until the user submits an empty search? if that's the case, then I find it quite unintuitive. if we remove the 'X' icon and its clearing action, would the user still need to enter an empty search, or manually deleting the term in the input field would do the job?

This has been fixed and now the X (clear button) also resets the search 🙂 .

app/javascript/src/Common/components/SelectWithModal.jsx
```diff
@@ -0,0 +1,140 @@
...
+  ]
+
+  // $FlowFixMe $FlowIssue It should not complain since Record.id has union "number | string"
+  const options = getItems(items).map(toSelectOption)
+
+  const handleOnFilter = (e) => {
+    const term = new RegExp(e.target.value, 'i')
+
+    const filteredRecords: T[] = term !== '' ? items.filter(b => term.test(b.name)) : items
+    let filteredItems = filteredRecords.length === 0
```
didierofrivia, line 89: what's the reason of using `let` here?

cntlsn:
Great, looks good!

> > * is a limit set to only display 5 items per page?
> 
> Yes, that's the number in the spec. It is not fixed though, 5 is only the default.

Ok! but since we're not displaying an interaction to change the number of items per page, we can consider the default value a "fixed" value from a user standpoint, could you confirm?

app/javascript/src/Common/components/SelectWithModal.jsx
josemigallas: None, I must have left it from an older version.

josemigallas:
> Ok! but since we're not displaying an interaction to change the number of items per page, we can consider the default value a "fixed" value from a user standpoint, could you confirm?

Yes, it is fixed.

didierofrivia (approved):
LGTM

guicassolato (approved):
Nice work!

hallelujah (approved):

PR #2371, opened 2021-02-19
Adds eslint warning for console

josemigallas:
We should have eslint warn us when we forget to remove our console logs. I would rather have it throw an error but then it would make it impossible to compile assets (since it needs a green lint check).


damianpm (approved):

hallelujah (approved):

hallelujah:
Thanks!

PR #2379, opened 2021-02-24
[circleci] Autorun tests on master

hallelujah:
So that when a merge is done in master, the workflows are run automatically

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -131,6 +131,11 @@ only-master-filter: &only-master-filter
     branches:
       only: master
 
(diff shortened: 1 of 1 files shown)


hallelujah:
Closing, it does not work.

https://circleci.com/docs/2.0/configuration-reference/#requires

> A list of jobs that must succeed for the job to start. Note: When jobs in the current workflow that are listed as dependencies are not executed (due to a filter function for example), their requirement as a dependency for other jobs will be ignored by the requires option. **However, if all dependencies of a job are filtered, then that job will not be executed either.**

PR #2385, opened 2021-03-02
[fix] Unicorn should log to STDOUT by default

hallelujah:
Just revert back this commit ee26e0f3c for the logs
Rails server logs are by default to stdout, so should we do it in development too

This causes binding.pry to malfunction, it stales forever

### DIFF
--- a/config/unicorn/development.rb
+++ b/config/unicorn/development.rb
@@ -26,7 +26,7 @@ listen prometheus_port
 
 pid app_dir.join('tmp/pids/unicorn.pid').to_s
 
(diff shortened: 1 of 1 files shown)


guicassolato (approved):
Attention to `RAILS_LOG_TO_STDOUT=true` now meaning it is OFF.

(To be fair, `RAILS_LOG_TO_STDOUT=0` meant it was ON before 🙂)

guicassolato (dismissed):

guicassolato (approved):

PR #2388, opened 2021-03-04
Fix item count in Products and BackendApis pagination

josemigallas:

It fixes a bug in pagination showing always the count of all items regardless of search.

[THREESCALE-6385: Products/Backends pagination shows all pages even when filtering](https://issues.redhat.com/browse/THREESCALE-6385)

**Verification steps** 

> Go to Products/Backends index pages
> Search using Sphinx and check the count of items is reduced.

Sphinx is probably not going to work properly but having a result of '0' is actually good for verifying this PR.

### DIFF
--- a/app/controllers/api/services_controller.rb
+++ b/app/controllers/api/services_controller.rb
@@ -19,12 +19,12 @@ class Api::ServicesController < Api::BaseController
   def index
     activate_menu :products
     search = ThreeScale::Search.new(params[:search] || params)
-    @raw_services = current_user.accessible_services
-    @services = @raw_services.order(updated_at: :desc)
-                             .paginate(pagination_params)
-                             .scope_search(search)
-                             .decorate
-                             .to_json(only: %i[name updated_at id system_name], methods: %i[links apps_count backends_count unread_alerts_count])
+    @services = current_user.accessible_services
+                            .order(updated_at: :desc)
+                            .scope_search(search)
+    @page_services = @services.paginate(pagination_params)
+                              .decorate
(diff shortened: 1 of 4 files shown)


app/controllers/api/services_controller.rb
```diff
@@ -19,12 +19,12 @@ class Api::ServicesController < Api::BaseController
   def index
     activate_menu :products
     search = ThreeScale::Search.new(params[:search] || params)
-    @raw_services = current_user.accessible_services
```
thomasmaas, line 22: we are not using `@raw_services` in the view?

app/controllers/provider/admin/backend_apis_controller.rb
```diff
@@ -12,12 +12,12 @@ class Provider::Admin::BackendApisController < Provider::Admin::BaseController
   def index
     activate_menu :backend_apis
     search = ThreeScale::Search.new(params[:search] || params)
-    @raw_backend_apis = current_account.backend_apis
```
thomasmaas, line 15: same question as above

app/views/api/services/index.html.slim
```diff
@@ -1,4 +1,4 @@
 = javascript_pack_tag 'products_index'
 = javascript_pack_tag 'PF4Styles/dashboard'
 
-div#products data-products=(@services) data-products-count=(@raw_services.size)
```
thomasmaas, line 4: and here's my answer

thomasmaas (approved):

PR #2389, opened 2021-03-04
SCSS lint: remove SelectorFormat rule

damianpm:

scss-lint `SelectorFormat` rule is not useful anymore, since it's not compatible with Patternfly at all.
Instead of disabling the rule everywhere, we should remove it.

None ;)


didierofrivia (dismissed):

josemigallas (approved):
Makes sense to me!

PR #2394, opened 2021-03-15
Bugfix: get billing adress for braintree

damianpm:

- It fixes an issue where the billing address was not pre-filled when going to the edit form.
- When Braintree fraud detection is enabled, billing address validation is failing because the region is in an unexpected format: it should be a two or three chars code, see [docs](https://braintree.github.io/braintree-web/current/ThreeDSecure.html#~billingAddress).

https://issues.redhat.com/browse/THREESCALE-6495

**Verification steps** 

- To enable Billing and adding Braintree to the Admin Portal, follow the instructions in https://github.com/3scale/porta/pull/2322

- In Dev Portal:
Go to **Settings** > **Credit Card Details**
Click on **Edit Credit Card Details and Billing Address**

brain

### DIFF
--- a/app/helpers/payment_details_helper.rb
+++ b/app/helpers/payment_details_helper.rb
@@ -68,4 +68,8 @@ module PaymentDetailsHelper
     }.to_json
   end
 
+  def get_country_code(billing_address_data)
+    merchant_countries.select {|c| c.include? billing_address_data.country}.first[1]
+  end 
+
 end
--- a/app/helpers/payment_details_helper.rb
+++ b/app/helpers/payment_details_helper.rb
@@ -69,7 +69,8 @@ module PaymentDetailsHelper
   end
 
   def get_country_code(billing_address_data)
-    merchant_countries.select {|c| c.include? billing_address_data.country}.first[1]
+    _label, code = merchant_countries.find {|label, code| label.downcase == billing_address_data.country.to_s.downcase }
+    code
   end 
 
 end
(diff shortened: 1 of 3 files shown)


app/helpers/payment_details_helper.rb
```diff
@@ -68,4 +68,8 @@ def stripe_billing_address_json
     }.to_json
   end
 
+  def get_country_code(billing_address_data)
+    merchant_countries.select {|c| c.include? billing_address_data.country}.first[1]
```
hallelujah, line 72: This can raise `NilError` when the country is not included in the list (for example the `billing_address_data.country == nil`)

thomasmaas (dismissed):
from ux perspective it ok for me.

app/helpers/payment_details_helper.rb
thomasmaas: @josemigallas  any suggestion to improve?

hallelujah: Why @josemigallas ? :smile: 

```suggestion
    label, _code = merchant_countries.find {|label, code| label == billing_address_data.country.to_s.downcase }
    code
```

Still this is suboptimal...

josemigallas: Then maybe use safe navigation? Not sure if it's syntactically correct:
```ruby
merchant_countries.select({|c| c.include? billing_address_data.country})&.first[1]
```

Or just split in two steps:
```ruby
countries = merchant_countries.select {|c| c.include? billing_address_data.country}
countries.first[1] if countries.present?
```

thomasmaas (approved):

hallelujah (approved):

PR #2400, opened 2021-03-23
[payment-gateway] Add three_ds_enabled settings

hallelujah:
This is the base for allowing customers to opt-in for 3DS secure

* Added `:boolean` option to `PaymentGateway`
* Added `three_ds_enabled` option to Braintree

## TODO:

- [x] Add and fix tests
- [x] Adapt the JS part to take care of this new option
- [x] [Adapt the Ruby part to send the correct `verify_card` to the Payment Gateway](https://github.com/3scale/porta/pull/2403) Doing separately
- [x] set `transaction_source: 'unscheduled'`

As commented by Braintree support;

> **Question: About metered recurring transactions**
> We do metered recurring payment, so we set transaction_source: 'recurring' in later transaction sales, what will be the result:
> If the payment method is subjected to SCA andwe use verify_card: true to create the payment_method
> we use verify_card: false to create the payment_method

> **Braintree Support answer:**
> The result will still ultimately depend on the card-issuing bank's decision. However, if there is a transaction or verification with 3DS information associated, subsequent recurring/unscheduled transactions are much more likely to be approved. If no transaction or verification has 3DS information, subsequent transaction attempts will more likely be 2099 declined for lack of 3DS.
>
> Please note, if the amount is not fixed, it may be more appropriate to pass `transaction_source: unscheduled`.

## Screenshots

**Before**

image

------------------------

**After**
image

### DIFF
--- a/app/models/payment_gateway.rb
+++ b/app/models/payment_gateway.rb
@@ -4,16 +4,15 @@ class PaymentGateway
   def initialize(type, options = {})
     @type = type
     @deprecated = options.delete(:deprecated).presence
-    @boolean_fields = options.delete(:boolean) || []
+    @boolean_field_keys = options.delete(:boolean) || []
     @fields = options
   end
 
   def non_boolean_fields
-    fields - boolean_fields
+    fields.except(*boolean_field_keys)
   end
 
-  attr_reader :type, :deprecated, :fields, :boolean_fields
-
+  attr_reader :type, :deprecated, :fields, :boolean_field_keys
 
   # So far hardcoded list of gateways, later maybe this will be loaded from a config file or db.
   GATEWAYS = [
--- a/app/models/payment_gateway.rb
+++ b/app/models/payment_gateway.rb
@@ -4,15 +4,21 @@ class PaymentGateway
   def initialize(type, options = {})
     @type = type
     @deprecated = options.delete(:deprecated).presence
+    @boolean_fields = options.delete(:boolean) || []
     @fields = options
   end
 
-  attr_reader :type, :deprecated, :fields
+  def non_boolean_fields
+    fields - boolean_fields
+  end
+
+  attr_reader :type, :deprecated, :fields, :boolean_fields
+
 
   # So far hardcoded list of gateways, later maybe this will be loaded from a config file or db.
   GATEWAYS = [
     PaymentGateway.new(:authorize_net, deprecated: true, login: 'LoginID', password: 'Transaction Key'),
-    PaymentGateway.new(:braintree_blue, public_key: 'Public Key', merchant_id: 'Merchant ID', private_key: 'Private Key'),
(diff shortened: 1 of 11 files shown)


app/models/payment_gateway.rb
```diff
@@ -4,17 +4,23 @@ class PaymentGateway
...
+
 
   # So far hardcoded list of gateways, later maybe this will be loaded from a config file or db.
   GATEWAYS = [
-    PaymentGateway.new(:authorize_net, deprecated: true, login: 'LoginID', password: 'Transaction Key'),
-    PaymentGateway.new(:braintree_blue, public_key: 'Public Key', merchant_id: 'Merchant ID', private_key: 'Private Key'),
-    PaymentGateway.new(:ogone, deprecated: true, login: 'PSPID', password: 'Password', user: 'User Id', signature: "SHA-IN Pass phrase", signature_out: "SHA-OUT Pass phrase"),
-    PaymentGateway.new(:stripe, login: 'Secret Key', publishable_key: 'Publishable Key', endpoint_secret: 'Webhook Signing Secret')
+    PaymentGateway.new(:authorize_net, deprecated: true, login: 'LoginID', password: 'Transaction Key', three_ds_enabled: '3D Secure enabled', boolean: %i[three_ds_enabled]),
+    PaymentGateway.new(:braintree_blue, public_key: 'Public Key', merchant_id: 'Merchant ID', private_key: 'Private Key', three_ds_enabled: '3D Secure enabled', boolean: %i[three_ds_enabled]),
```
guicassolato, line 21: Wow. Boolean fields. Nice!

app/views/finance/provider/settings/show.html.slim
```diff
@@ -79,11 +79,15 @@
                   li class="string"
                     = fields.label name, label
```
thomasmaas, line 80: Does this result in a working label (one you can click to toggle the box?)

I'd prefer we stick to this:

Screenshot 2021-03-23 at 12 56 18
Screenshot 2021-03-23 at 12 56 02

thomasmaas: inline hint text suggestion:

> Make sure your Braintree account has 3D Secure enabled before you enable it here as otherwise payments initiated from the developer portal will fail.

hallelujah: This is not formtastic so ... could you make a commit for that? Or add a suggestion?

hallelujah: @thomasmaas this option is available for all payment gateways so the hint should be more generic or  somebody should tell me what to add in case it is not a valid hint for the payment gateway. Meaning it needs research and definition :)

thomasmaas: But we agreed to only show it when Braintree is selected, right? I guess you mean at the data level this option is not tight to 1 payment gateway. Working on a re-phrase.

hallelujah: It is only for Braintree for now. What I meant isn't that something we want for all? Like preparing something (defining) instead of going back and forth everytime
Can you make the commit @thomasmaas ? Would be nice to

PR #2408, opened 2021-04-06
Updates installation guide for Arch linux and Postgres database

jsmadis:

Updates installation guide for Arch linux users and updates the installation guide when Postgres is used.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -247,7 +247,7 @@ sudo systemctl restart memcached
 Postgres, MySQL or Oracle has to be running for the application to work. The easiest way to do it is in a [Docker](https://www.docker.com/) container by simply running:
 
 ```shell
-docker run -d -p 5433:5432 -e POSTGRES_USER=postgres -e POSTGRES_DB=3scale_system_development --name postgres10 circleci/postgres:10.5-alpine
+docker run -d -p 5432:5432 -e POSTGRES_USER=postgres -e POSTGRES_DB=3scale_system_development --name postgres10 circleci/postgres:10.5-alpine
 ```
 
 Alternatively, you can run Postgres directly on your machine by following [this article](https://developer.fedoraproject.org/tech/database/postgresql/about.html).
@@ -320,6 +320,13 @@ PKG_CONFIG_PATH=/usr/lib/imagemagick6/pkgconfig bundle install
 ```
 
 ## Setup Database
+
+If you are using a different database than MySQL you need to export the env variable `DATABASE_URL`.
+
+For Postgres it would be:
+```bash
+export DATABASE_URL=postgresql://postgres:@localhost/3scale_system_development
+```
 
 Finally, initialize the database with some seed data by running:
 
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -309,6 +309,16 @@ Run [Yarn](https://www.yarnpkg.com/) to install all the required dependencies:
 yarn install
 ```
 
+## Manual setup for Arch linux
+
+The setup is the same as for Fedora, but the naming of packages might differ.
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -309,8 +309,25 @@ Run [Yarn](https://www.yarnpkg.com/) to install all the required dependencies:
 yarn install
 ```
 
+## Manual setup for Arch linux
+
+The setup is the same as for Fedora, with
```
guicassolato, line 314: It looks the sentence is missing a part here.

guicassolato (approved):

INSTALL.md
```diff
@@ -309,8 +309,25 @@ Run [Yarn](https://www.yarnpkg.com/) to install all the required dependencies:
...
 ```
 
+## Manual setup for Arch linux
+
+The setup is the same as for Fedora, but the naming of packages might differ.
+Since Porta needs ImageMagick6, you need to install [libmagick6](https://aur.archlinux.org/packages/libmagick6/) from AUR.
+
+When running `bundle install`, you need to specify `PKG_CONFIG_PATH`.
+```shell
+PKG_CONFIG_PATH=/usr/lib/imagemagick6/pkgconfig bundle install
```
hallelujah, line 319: Can use linuxbrew (homebrew) and be done with different distribution of Linux?
I am worried of the number of platform we are going to officially support for development

This could be put in Troubleshooting in wiki, and if it is up to date then good if not then "best customer support"

hallelujah (changes requested):

INSTALL.md
```diff
@@ -309,8 +309,25 @@ Run [Yarn](https://www.yarnpkg.com/) to install all the required dependencies:
...
+```
+
 ## Setup Database
 
+If you are using a different database than MySQL you need to export the env variable `DATABASE_URL`.
```
hallelujah, line 324: Well not really, basically this is configuration driven, you pick the configuration you need and that's it

The `config/examples/*.yml` files are, as named, _examples_
They are default values that can run the application by default and also indicators of what could be done/modified

I would not state this as a step to run postgresql because you can just modify the `config/database.yml` file at your need

INSTALL.md
jsmadis: I am not sure if it can be done with different distributions of Linux. But this is a solution when you need to have more versions of ImageMagick on your system.

I agree that this might be more suitable for Troubleshooting in the wiki. Do we have something like that somewhere?

INSTALL.md
jsmadis: So we should say that in order to run porta with a different DB they need to change `config/database.yml`.
Because the DB is chosen by the `DATABASE_URL` value.

The problem for me was that in the steps for fedora there is postgres as the database and there was no step that pointed that I should change something in order to run the porta with postgres.

hallelujah: Well it is opinionated, there is no "1 way to do it". You pick the one you want.
If you wish you could describe it more generically for the 3 databases but ... it is how Rails does it so that is why there is not really need of mentioning that.

I really mean that if somebody wants to develop on this project, they MUST know Rails config/database.yml

I'd prefer putting in INSTALL.md what we really support and step to install on few OS, and create a Troubleshooting wiki page

INSTALL.md
hallelujah: Wiki was opened few months ago, go ahead if you want to create a page

INSTALL.md
jsmadis: I agree that INSTALL.md should contain only what we really support. I will move the specific stuff to the wiki page.
However, the installation guide for Fedora shows steps to deploy Postgres and never mentions that you need to modify `/config/database.yml`.
I would propose changing Postgres to MySQL in the INSTALL.md and creating a wiki page for the setup of different DBs.

PR #2409, opened 2021-04-07
Updating formtastic gem to 3

HemaHG:

Upgrading Formtastic gem to 3 as it is supported by the current rails version

closes [THREESCALE-5661](https://issues.redhat.com/browse/THREESCALE-5661)


HemaHG:
I will open a new one

PR #2410, opened 2021-04-07
⚠️ Legal terms settings features not being run

josemigallas:
These test hasn't being running for a while! The file defines "Scenario Outline" which is a template, but there are no values (examples) for that template so the test cases are not detected.


thomasmaas (approved):
These are legal term for dev portal. My question is, why were they disabled?

josemigallas:
> These are tests for legal term of dev portal. My question is, why were they disabled?

They were disabled by accident, nobody noticed until now.

PR #2411, opened 2021-04-08
master to 2.11 candidate

thomasmaas:


gsaslis:
Don't see anything wrong here with wanting to bring 3scale-2.11-candidate up-to-date with master. 

This will unblock corresponding jira issues (whose content is included on master) from being handed over to QE, as QE will now have a way to verify those using the candidate release stream.

PR #2415, opened 2021-04-15
[2.11] Implement inline errors to the Add Backend form

josemigallas:

Renders the following inline errors whenever the form is sent with invalid data:
- Backend not found: this should never happen but the submit button can be disabled manually and the form submitted (see image).
Screenshot 2021-04-15 at 18 28 51

- Path already taken: this will be returned by the server. In this case, the selected backend will still be selected despite the page refresh.

Demo:
add-backend-usage

[THREESCALE-6883: Implements inline validations and flash errors upon submition](https://issues.redhat.com/browse/THREESCALE-6883)

> Subtask of [3scale][2.11][HI-prio] Improve Select in 'Add Backend to Product' form
[THREESCALE-6882: Embed New Backend form into a modal](https://issues.redhat.com/browse/THREESCALE-6882)

### DIFF
--- a/app/controllers/api/backend_usages_controller.rb
+++ b/app/controllers/api/backend_usages_controller.rb
@@ -30,6 +30,7 @@ class Api::BackendUsagesController < Api::BaseController
       redirect_to admin_service_backend_usages_path(@service)
     else
       flash[:error] = "Couldn't add Backend to Product"
+      @inline_errors = @backend_api_config.errors.as_json
       render 'new'
     end
   end
@@ -74,6 +75,10 @@ class Api::BackendUsagesController < Api::BaseController
   end
 
   def ensure_same_account_backend_api
-    current_account.backend_apis.find(backend_api_config_params[:backend_api_id])
+    return if current_account.backend_apis.find_by(id: backend_api_config_params[:backend_api_id])
+
+    flash[:error] = "Couldn't add Backend to Product"
+    @inline_errors = { backend_api_id: ['Not a valid backend'] }
+    render 'new'
   end
 end
--- a/app/javascript/packs/add_backend_usage.js
+++ b/app/javascript/packs/add_backend_usage.js
@@ -3,6 +3,8 @@
 import { AddBackendFormWrapper } from 'BackendApis'
 import { safeFromJsonString } from 'utilities/json-utils'
 
+import type { Backend } from 'Types'
+
 const containerId = 'add-backend-form'
 
 document.addEventListener('DOMContentLoaded', () => {
@@ -12,10 +14,17 @@ document.addEventListener('DOMContentLoaded', () => {
     return
   }
 
-  const { backends, url, backendsPath } = container.dataset
+  const { dataset } = container
+  const { url, backendsPath, backendApiId } = dataset
+
+  const backends = safeFromJsonString<Backend[]>(dataset.backends) || []
+  const backend = backends.find(b => String(b.id) === backendApiId) || null
+  const inlineErrors = safeFromJsonString(dataset.inlineErrors)
 
   AddBackendFormWrapper({
-    backends: safeFromJsonString(backends) || [],
+    backends,
+    backend,
+    inlineErrors,
     url,
     backendsPath
   }, containerId)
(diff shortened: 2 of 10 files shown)


thomasmaas (dismissed):

app/assets/javascripts/application.js
```diff
@@ -236,8 +236,10 @@
       location.href = "?view=" + view + "&s=" + param + "&for_category=" + category;
     });
 
+    // DEPRECATED: since the introduction of PF4 and React, colorbox is being removed. Also jquery-ujs has been replaced with rails-ujs.
+    // Added #colorbox selector to target only non-React forms
     // show errors from ajax in formtastic
-    $('form').live('ajax:error', function(event, xhr, status, error) {
+    $('#colorbox form').live('ajax:error', function(event, xhr, status, error) {
```
damianpm, line 242: colorbox is also used in the Developer Portal, have you checked it doesn't affect old dev portals?
(I don't remember if this fie is loaded in dev portal)

app/javascript/src/BackendApis/components/NewBackendModal.jsx
```diff
@@ -0,0 +1,77 @@
+// @flow
+
+import * as React from 'react'
+
```
damianpm, line 4: For consistency with `NewBackendForm.jsx`, you coud import `useState` and `useEffect`

app/javascript/src/BackendApis/components/NewBackendModal.jsx
```diff
@@ -0,0 +1,77 @@
...
+      const backend = JSON.parse(xhr.responseText)
+      onCreateBackend(backend)
+    } else if (status === 'error') {
+      const errors = JSON.parse(xhr.responseText)
+      setErrors(errors)
+    }
+  }
+
+  React.useEffect(() => {
+    $('form#new_backend_api_config')
```
damianpm, line 43: Why we need jQuery here, can we add listeners in other way?
If complicated and out of scope, we could add a todo

josemigallas: This has to do with the way Rails makes async calls. I think we haven't done this with React before but the library is injected into jQuery so I'm afraid this is how it's done.. ugly yes. We can't even import $ from node_modules because that's a different version and it won't include jquery-ujs.

In Rails 5 there's a new module that replaces jquery-ujs, however porta uses it extensively and migrating to rails-ujs (the new module) would be a massive task. There are already some todos pointing to this. But again, I think it's not realistic to think we're going to "fix" this anytime soon. At least we'd need to migrate everything to React and stop using colorbox and the rest of old libraries.

app/assets/javascripts/application.js
josemigallas: I'll take a more thorough look. What I'm trying to do is to prevent jquery-ujs from injecting callbacks in our new components. Maybe it's better to do it the other way around: add an excluding selector to filter our PF forms `$('form:not(.pf-c-form)')`

app/javascript/src/BackendApis/components/NewBackendModal.jsx
josemigallas: Actually I just saw there's a `$FlowFixMe` in the next line. These are to be treated as TODO.

app/lib/three_scale/oauth2/github_client.rb
```diff
@@ -49,7 +49,6 @@ def primary_email
       end
 
       def emails
-        access_token.options[:mode] = :query
```
hallelujah, line 52: This is not related to this PR I think, probably need to rebase on master

hallelujah (dismissed):

app/lib/three_scale/oauth2/github_client.rb
```diff
@@ -49,7 +49,6 @@ def primary_email
       end
 
       def emails
```
josemigallas, line 51:
```suggestion
      def emails
        access_token.options[:mode] = :query
```

Gemfile.prod
```diff
@@ -1,8 +1 @@
 eval_gemfile 'Gemfile.base'
-
```
damianpm, line 2: I think you have to rebase, see https://github.com/3scale/porta/pull/2423

damianpm (changes requested):

damianpm (approved):

hallelujah (approved):

PR #2416, opened 2021-04-16
🚨  removes Form module in order to fix Flow name-mapper, fix some war…

josemigallas:
* Remove module `Form` since it was messing up with other imports, plus it's only used by Settings and it's going to be replaced with PF4 soon.
* Use the proper ignore codes (FlowFixMe, FlowIgnore, FlowIssue) in each case.
* Fix some Flow warnings (run `yarn flow --include-warnings` to see if there are any).


app/javascript/src/Plans/index.js
```diff
@@ -3,9 +3,9 @@
 export * from './components/DefaultPlanSelect'
 export * from './components/DefaultPlanSelectCard'
 export * from './components/DefaultPlanSelectWrapper'
-// $FlowIgnore[cannot-resolve-module] name-mapper is replacing 'Plans'
+// $FlowFixMe[cannot-resolve-module] name-mapper is replacing 'Plans', to fix this rename module or fix name_mapper expression in flowconfig
```
damianpm, line 6: Do you know why `Plans` is replaced?

josemigallas: Yes, our `name-mapper` setup replaces any term within the regexp (see .flowconfig) with the absolute path of porta. That's how we are able to use import paths like this. Example:
```jsx
import * from 'Plans/foo'
// is replaced during compilation with
import * from '/user/path/to/porta/src/javascript/src/Plans/foo'
```

Therefore anything that matches the same regexp will be replaced, for instance `ApplicationPlansTableCard`:
```jsx
export * from './components/ApplicationPlansTableCard'
// is replaced during compilation with
import * from './components/Application/user/path/to/porta/src/javascript/src/Plans/TableCard'
```

josemigallas: This is why I removed the module app/javascript/src/Form in this PR, because any component named `SomethingForm.jsx` was being replaced and it's a PITA.

damianpm (approved):

PR #2420, opened 2021-04-20
[gem] Update mimemagic because of license

hallelujah:
The gem version has been yanked because of some licensing issues

https://github.com/mimemagicrb/mimemagic/issues/98

### DIFF
--- a/Gemfile.base
+++ b/Gemfile.base
@@ -91,7 +91,8 @@ group :development, :test do
   end
 end
 
-gem 'nokogiri', '>= 1.6.8'
+gem 'mimemagic', '~> 0.3.10'
+gem 'nokogiri', '~> 1.10.10'
 gem 'dalli', '~> 2.7'
 gem 'secure_headers', '~> 6.3.0'
 gem 'faraday', '~> 0.15.3'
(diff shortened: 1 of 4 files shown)


Gemfile.base
```diff
@@ -91,7 +91,8 @@ group :development, :test do
   end
 end
 
-gem 'nokogiri', '>= 1.6.8'
+gem 'mimemagic', '~> 0.3.10'
+gem 'nokogiri', '~> 1.10.10'
```
gsaslis, line 95: I see we're not only updating the version of nokogiri (looking at the mimemagic linked PR some kind of version update does seem to be required though I was not able to spot exact versions recommended there), but we're also changing the policy here when choosing which nokogiri versions we can accept updates for: 
`>=` vs. `~>` 

is that intentional, or perhaps some copy-paste error? 🤔

hallelujah: This is correct, nokogiri 1.11+ drops support of ruby 2.4 which is used in some deployments (not disclosing which ones for security reasons 😖 ...)

gsaslis: Do you think this is worth also documenting in the commit message?

As long as we have it documented at least in this comment, I'm good with that.

guicassolato (approved):

Gemfile.base
hallelujah: It is documented in another PR on the main branch https://github.com/3scale/porta/pull/2391

PR #2421, opened 2021-04-20
[2.11] New Mapping Rule page

josemigallas:

Re implement New Mapping Rule page using React and PF4.

Product:
Screenshot 2021-04-20 at 17 57 06

Backend API:
Screenshot 2021-04-20 at 17 57 12

THREESCALE-6874 Improve UI for creating a Product Mapping rule > [THREESCALE-6956: Re-implement New Mapping Rule page with React and PF4](https://issues.redhat.com/browse/THREESCALE-6956)


app/models/service.rb
```diff
@@ -519,6 +519,10 @@ def plugin_deployment?
     DeploymentOption.plugins.include?(deployment_option)
   end
 
+  def top_level_metrics
```
guicassolato, line 522: Maybe this could move closer to https://github.com/3scale/porta/blob/16e746c7665d179b07f2a5b9415bbb4c6adc92c3/app/models/service.rb#L319

guicassolato:
Do we really want to separate _Methods_ and _Metrics_ in two radio options and then two different drop-drown selectors? Why if, from the perspective of defining a mapping rule, those are exactly the same thing?

cntlsn:
Hi @guicassolato, I'm not sure I understand what you mean with
> Why if, from the perspective of defining a mapping rule, those are exactly the same thing?

The rationale for the separation of Methods and Metrics into two groups is the following:
* Avoid a long list
* Help the user understand what kind of item they are selecting

Here is the click-though mockup https://marvelapp.com/prototype/13h9054e/screen/76056757
What do you think?

guicassolato:
> What do you think?

@cntlsn, personally, I think these two fields should remain being one.

This is how I see this:

In the end, what we are asking in the form is "What should it be incremented/decremented when path pattern X of the API is requested?" And the answer should be straightforward, e.g. "The indicator/metric Y". So, the mapping is "X → Y". The conceptual difference of if Y is a "metric" or a "method metric" has no meaning here.

Having instead two separate fields adds to code maintenance and, IMHO, it can also backfire in its purpose of educating the user about metric and methods metric being two different things, because when defining a mapping rule is precisely where those concepts mean exactly the same thing.

Even if one argues that incrementing a metric is not the same thing as incrementing a method metric, because incrementing a method metrics also increments the Hits metric while incrementing another top level metric does not, I'd still say that it doesn't matter here. This conceptual difference is something that is intrinsic to the definitions of what metrics and method metrics are, rather than a property inherited from the mapping rule functionality. From the mapping rule perspective, it should be straightforward: "What metric should it be incremented?", and if that metric happens to be a special type of metric (a method metric), it doesn't matter, IMO.

cntlsn (approved):
As discussed offline, we will move ahead with the current implementation and will consider iterate according to @guicassolato 's feedback in the future, after collecting more opinions on the topic.

app/javascript/src/MappingRules/components/NewMappingRule.jsx
```diff
@@ -0,0 +1,88 @@
...
+import * as React from 'react'
+
+import {
+  ActionGroup,
+  Button,
+  Form,
+  PageSection,
+  PageSectionVariants
+} from '@patternfly/react-core'
+import { CSRFToken } from 'utilities/utils'
```
damianpm, line 12: I think you have to update this import
```suggestion
import { CSRFToken } from 'utilities/CSRFToken'
```

josemigallas: Good catch! I'll rebase with master and push again

app/views/api/services/show.html.slim
```diff
@@ -116,8 +116,6 @@ section class="service-widget"
           =< link_to 'Settings', settings_admin_service_path(service), :title => "Edit service settings"
 
         ul.service-settings[data-hint="basics"]
-          - if service.deployment_option.present?
-            - deployment_option_phrased = t(service.proxy.deployment_option, scope: "deployment_options.phrased")
```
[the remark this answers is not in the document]
josemigallas: Sorry this changes didn't belong here, moving conversation to #2442.

guicassolato (approved):

damianpm (approved):

hallelujah (approved):

PR #2424, opened 2021-04-22
[gem] Remove sidekiq-pro

hallelujah:
Context: we have it used on one of our production deployment.
We mainly rely on 2 features for it:

- Reliability
- Batches

First, after years used by customers on their own deployments without sidekiq-pro, no issues reported to us so far about reliability
We think it is time to say goodbye on this awesome gem.

Second, we successfully make use of an alternative to Pro batches sidekiq-batch

All in one, the necessity of sidekiq-pro dimmed so the decision is to take it down for good

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,4 +2,3 @@
 
 eval_gemfile 'Gemfile.base'
 gem 'pg', '~> 0.21.0'
-gem 'sidekiq-batch', '~> 0.1.6'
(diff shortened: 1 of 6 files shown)


guicassolato (approved):

guicassolato (approved):

hallelujah:
Tested on preview with billing jobs too (for batches)
It works!

PR #2425, opened 2021-04-28
[2.11] New table "Products using backend"

josemigallas:

Adds pagination and a search bar to the list of used products in Backend overview page.
* The table is paginated and filtered in client side.
* In order to filter, either click the magnifier button or hit enter.

Screenshot 2021-04-28 at 13 38 20

Screenshot 2021-04-28 at 13 38 25

OLD:
Screenshot 2021-04-28 at 14 06 54

[THREESCALE-6877: [3scale][2.11][HI-prio] Improve list of Products in Backend overview page](https://issues.redhat.com/browse/THREESCALE-6877)


features/provider/admin/backend_api.feature
```diff
@@ -16,6 +14,6 @@ Feature: Backend API overview page
       | Product 2 |
 
   Scenario: Only accessible products are visible
-    Given service "Product 2" becomes unnaccessible
+    Given service "Product 2" becomes inaccessible
```
josemigallas, line 17: 😅

cntlsn (changes requested):
Thank you @josemigallas, this looks good to me. However, I noticed a couple of things that could be improved:
* the search bar should include a placeholder
* the products in the list would better be sorted alphabetically in this case

WDYT?

app/decorators/backend_api_decorator.rb
```diff
@@ -17,6 +17,12 @@ def add_backend_usage_backends_data
     }
   end
 
+  def products_used_table_data
+    ServiceDecorator.decorate_collection(services.accessible.sort_by(&:name))
```
hallelujah, line 21: I liked the `order` version though. Why the change?

hallelujah: Another remark, customers with backend api shared across many services might have hard time to load this ...
Do you know about any UI/UX solution @cntlsn @josemigallas @3scale/system-ux  ?

cntlsn: @hallelujah can you please expand on why it could take longer to load the list? one of the main reasons behind this UI change is actually to try to speed up page loading.

hallelujah: Well we have issues already with customers with more than 500 services on-prem ...
I imagine that this will generate a huge JSON and also a huge data attribute.

josemigallas: @hallelujah I couldn't make `order` work.

As per performance, the JSON is not an issue, the bottleneck would be the DB I think. But no I have not tested it.

hallelujah: I'd say the problem will be it might be slow :) Do not think only if it is DB or JS
Probably the implementation is not solving at all the problem as quoted and stressed below:

> @hallelujah can you please expand on why it could take longer to load the list? **one of the main reasons behind this UI change is actually to try to speed up page loading.**

JS/ DB ruby are only tools but if you load all the services there is no miracle ... it will be slow :(

I'd propose an enhancement loading the services partially (20 services at a time for example, paginating) and the search would be done in AJAX. _This is a proposition of implementation not a specification_

josemigallas: Yes, loading items per-page is basically what we are looking for in the new design. But it makes the implementation much harder and IMO moving to react and cleaning up old logic will be an improvement. Actually I think we can't really assure anything until we start stress testing.

app/decorators/service_decorator.rb
```diff
@@ -45,7 +45,15 @@ def api_selector_api_link
     end
   end
 
-  alias link api_selector_api_link
+  def used_by_backend_table_data
```
guicassolato, line 48: The name of this method almost sounds like a comment, a comment that would be slightly inaccurate if taken literally about what the methods is used by 🙂

features/step_definitions/backend_api_steps.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+Given /^a backend api "([^"]*)"( that is unnaccessible)?$/ do |name, unnaccessible|
```
guicassolato, line 3: (Apart from the typo) "accessible"/"inaccessible" was a very unfortunate choice or ours in the code to distinguish if a record is active or been deleted. I wouldn't expose this mistake in the cucumber step.

```suggestion
Given /^a backend api "([^"]*)"( that is deleted)?$/ do |name, deleted|
```

app/decorators/service_decorator.rb
josemigallas: What do you suggest instead?

app/decorators/backend_api_decorator.rb
```diff
@@ -17,6 +17,12 @@ def add_backend_usage_backends_data
     }
   end
 
+  def products_used_table_data
```
guicassolato, line 20:
```suggestion
  def products_table_data
```

app/decorators/service_decorator.rb
guicassolato: IDK, but a service (API product) is not used _by_ a backend. Regardless, the response seems to be nothing related to backend. It's name, system_name and URL of the service, as JSON.

guicassolato: Maybe just 

```suggestion
  def table_data
```

josemigallas: Fair enough

spec/javascripts/BackendApis/components/ProductsUsedList.spec.jsx
```diff
@@ -0,0 +1,67 @@
...
+})
+
+it('should be able to navigate to a product overview page', () => {
+  const products = mockProducts(1)
+  const wrapper = mountWrapper({ products })
+  expect(wrapper.find('tbody tr').first().find('Button').prop('href')).toEqual(products[0].href)
+})
+
+// FIXME: input not receiving change event
+it.skip('should be filterable by name', () => {
```
damianpm, line 46: Can you create a Jira to fix this? or we will forget

josemigallas: Sure, done!

damianpm (approved):
LGTM, js side

guicassolato (approved):
LGTM

app/decorators/backend_api_decorator.rb
hallelujah: Did you do a test locally creating the 1k services and accounts with FactoryBot?
By the way what failed with the order query? Can I help you with that?
It is better to do it from the Database as it will be faster

PR #2427, opened 2021-04-29
Fixed THREESCALE-6428 - Add APIAP Routing policy just before APICast policy

srijitamukherjee:
Signed-off-by: Srijita Mukherjee <srmukher@redhat.com>

Currently, when using APIAP some policies are broken because Routing policy is
always first.
hence placing the Routing policy just above the Apicast policy.

> Closes THREESCALE-6428 

**Verification steps** 
To be documented.

### DIFF
--- a/app/lib/backend_api_logic/routing_policy.rb
+++ b/app/lib/backend_api_logic/routing_policy.rb
@@ -10,17 +10,12 @@ module BackendApiLogic
       other_routing_rules = other_routing_policies.flat_map { |policy| policy['configuration']['rules'] }
 
       routing_policy = Builder.new(service).to_h
-      
+
       return other_policies if routing_policy['configuration']['rules'].concat(other_routing_rules).empty?
 
-      index_apicast = other_policies.index { |policy| policy['name'] == 'apicast'}
-      result = []
-      if index_apicast > 0
-        result.push(*other_policies[0..index_apicast - 1])
-      end
-      result << routing_policy
-      result.push(*other_policies[index_apicast..-1])
-      return result
+      apicast_policy_index = other_policies.index { |policy| policy['name'] == 'apicast'}
+      other_policies.insert(apicast_policy_index, routing_policy).compact
+
     end
 
     def with_subpaths?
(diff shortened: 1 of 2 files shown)


app/lib/backend_api_logic/routing_policy.rb
```diff
@@ -10,10 +10,17 @@ def policy_chain
...
 
-      [routing_policy, *other_policies].compact
+      index_apicast = other_policies.index { |policy| policy['name'] == 'apicast'}
+      result = []
+      if index_apicast > 0
+        result.push(*other_policies[0..index_apicast - 1])
+      end
+      result << routing_policy
+      result.push(*other_policies[index_apicast..-1])
+      return result
```
guicassolato, line 23: You should try https://ruby-doc.org/core-2.5.0/Array.html#method-i-insert here.

app/lib/backend_api_logic/routing_policy.rb
```diff
@@ -10,10 +10,17 @@ def policy_chain
       other_routing_rules = other_routing_policies.flat_map { |policy| policy['configuration']['rules'] }
 
       routing_policy = Builder.new(service).to_h
-
+
```
guicassolato, line 13: Try setting up the auto-removal of empty white-spaces in your IDE to avoid this.

app/lib/backend_api_logic/routing_policy.rb
guicassolato: Just one raw idea here:

```ruby
apicast_policy_index = other_policies.index { |policy| policy['name'] == 'apicast'}
other_policies.insert(apicast_policy_index || 0, routing_policy).compact
```

srijitamukherjee: @guicassolato This looks good, I will update it. Just a question on `apicast_policy_index || 0` , I assume this to handle the case when there is no apicast policy, but AFAIU 3scale Apicast policy is default and will always be there and cannot be deleted. Is that the case  or not ?

guicassolato: You are right. apicast policy should always be there and there's no need for the `|| 0` fallback.

guicassolato (approved):

jsmadis (approved):

PR #2431, opened 2021-05-03
[task] A task to fix services that do not have proxy

hallelujah:
Proxy is mandatory for a service to function.
The application is failing if a service does not have proxy

This is also useful when using the obfuscated database from production.
We truncate `proxies` table so we need to launch this task everytime we restore the database

### DIFF
--- a/app/workers/create_default_proxy_worker.rb
+++ b/app/workers/create_default_proxy_worker.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+class CreateDefaultProxyWorker < ApplicationJob
+
+  rescue_from(ActiveJob::DeserializationError, ActiveRecord::RecordNotFound) do |exception|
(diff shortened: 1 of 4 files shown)


guicassolato (approved):

app/workers/create_default_proxy_worker.rb
```diff
@@ -0,0 +1,23 @@
...
+  # :reek:UtilityFunction
+  def perform(service)
+    service.create_default_proxy
+  end
+
+  class BatchEnqueueWorker < ApplicationJob
+
+    # :reek:UtilityFunction
+    def perform(*)
+      Service.includes(:proxy).where(proxies: {service_id: nil}).find_each do |service|
```
hallelujah, line 18: This could be a scope but it is not used anywhere else so not needed for now

```ruby
      # in Service class
      scope(:without_proxy), -> { includes(:proxy).where(proxies: {service_id: nil}) }
```

josemigallas (approved):

PR #2434, opened 2021-05-04
[2.11] New Metrics index page for Products

josemigallas:

Re-implements Metrics index page with React and Patterrnfly styles.

Mockup:
Screenshot 2021-05-04 at 18 00 20

[THREESCALE-6867: [3scale][2.11][HI-prio] Improve list of Methods and Metrics in Product context](https://issues.redhat.com/browse/THREESCALE-6867)


thomasmaas (approved):

app/javascript/src/Metrics/components/ProductIndexPage.jsx
```diff
@@ -0,0 +1,67 @@
...
+  const currentTabKey = (new URL(window.location.href)).searchParams.get('tab')
+  return (
+    <IndexPage
+      metrics={metrics}
+      metricsCount={metricsCount}
+      infoCard={currentTabKey === 'metrics' ? (
+        <>
+        Metrics track API usage. Metrics have these features:
+        <List>
+          <ListItem><i><b>Hits</b></i> is the 3scale native metric to which all methods report. This metric tracks the number of calls made to your API.</ListItem>
```
porueesq, line 38:
```suggestion
          <ListItem><b>Hits</b> is the 3scale native metric to which all methods report. This metric tracks the number of calls made to your API.</ListItem>
```

Let's just stick to one font type, in this case bold, to indicate that Hits is a proper name in the context of 3scale.These font variations can make text more difficult to read—and each new variation requires some orientation by the user.

app/javascript/src/Metrics/components/ProductIndexPage.jsx
```diff
@@ -0,0 +1,67 @@
...
+          <ListItem>To track usage that does not increase the hit count, add top-level metrics.</ListItem>
+          <ListItem>To have specific calls to your API tracked by specific metrics, you must map a metric to one or more URL patterns listed in <a href={mappingRulesPath}>Mapping rules</a>.</ListItem>
+        </List>
+        </>
+      ) : (
+        <>
+          Methods track API usage.  Methods have these features:
+          <List>
+            <ListItem>Method calls trigger the <i>Hits</i> metric.</ListItem>
```
porueesq, line 47:
```suggestion
            <ListItem>Method calls trigger the <b>Hits</b> metric.</ListItem>
```


Let's just stick to one font type, in this case bold, to indicate that Hits is a proper name in the context of 3scale.

porueesq (changes requested):
A couple of changes to improve accessibility. Still researching about this, probably will need to get back to this content in the future.

spec/javascripts/Metrics/components/MetricsTable.spec.jsx
```diff
@@ -0,0 +1,59 @@
...
+  mapped: true,
+  updatedAt: Date.now().toString()
+}))
+const metricsCount = metrics.length * 2
+
+const defaultProps = {
+  activeTabKey: 'methods',
+  metrics,
+  metricsCount,
+  createButton: <button>Add metric</button>
```
porueesq, line 25:
```suggestion
  createButton: <button>Add a metric</button>
```

porueesq (changes requested):
A small change.

app/javascript/src/Metrics/components/ProductIndexPage.jsx
josemigallas: Regarding [1] keep in mind that although technically this is JS, once "compiled" it will be actual HTML.

porueesq (approved):

app/javascript/src/Metrics/components/IndexPage.scss
```diff
@@ -0,0 +1,36 @@
...
+@import "~@patternfly/patternfly/components/Card/card.scss";
+@import '~@patternfly/patternfly/components/Table/table.css';
+@import '~@patternfly/patternfly/components/Pagination/pagination.css';
+@import '~@patternfly/patternfly/components/Toolbar/toolbar.css';
+@import '~@patternfly/patternfly/components/OptionsMenu/options-menu.css';
+@import "~@patternfly/patternfly/components/List/list.scss";
+
+main.Main {
+  // Since the whole content is rendered by React, we need to eliminate the padding adding by .Main
+  padding: 0;
```
josemigallas, line 12:
```suggestion
  padding: unset;
```

app/indices/metric_index.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+ThinkingSphinx::Index.define(:metric, with: :real_time) do
+  indexes :friendly_name, as: :friendly_name
```
hallelujah, line 4: I am thinking ...

We did not change before https://github.com/3scale/porta/blob/40771d47b8364c4721251740d1b4d933209e3624/config/thinking_sphinx.yml#L19-L21 and https://github.com/3scale/porta/blob/40771d47b8364c4721251740d1b4d933209e3624/config/docker/thinking_sphinx.yml#L18-L20 because we index in "real time" but we probably would like to do it for consistency.

app/presenters/api/metrics_index_presenter.rb
```diff
@@ -0,0 +1,62 @@
...
+      'create-metric-path': create_path,
+      'mapping-rules-path': admin_service_proxy_rules_path(service),
+      'metrics': page_metrics.to_json,
+      'metrics-count': collection.total_entries
+    }
+  end
+
+  def raw_collection
+    @raw_collection ||= metrics? ? top_level_metrics : method_metrics
+  end
```
hallelujah, line 29: I would make this private/protected method

app/presenters/api/metrics_index_presenter.rb
```diff
@@ -0,0 +1,62 @@
...
+  def metric_table_data(metric)
+    {
+      id: metric.id,
+      name: metric.friendly_name,
+      systemName: metric.system_name,
+      updatedAt: metric.updated_at,
+      path: edit_admin_service_metric_path(metric.owner, metric),
+      unit: metric.unit,
+      description: metric.description,
+      mapped: metric.owner.proxy.proxy_rules.map(&:metric).include?(metric)
```
hallelujah, line 59: Rubocop does not complain? This is assuming too much of existence of each method call.

Theorically:
- metric.owner could be nil, how to handle that?
- owner.proxy could be nil, how to handle that?

Also I would probably avoid N+1, something along (of course cleaned up and tested):

```ruby
metric.owner.proxy_rules.joins(:metric).where(metrics: {id: metric.id}).exist?
```

hallelujah (changes requested):
Looks great 🚀 , some changes needed I think ;)

app/indices/metric_index.rb
josemigallas: OK why aren't there more models included anyway? 🤔

app/presenters/api/metrics_index_presenter.rb
josemigallas: Yeah, I don't know where that came from but it's ugly as hell. I think maybe it would be as simple as check if the metric has mapping rules: `metric.proxy_rules.any?`.

app/indices/metric_index.rb
hallelujah: I think we can safely remove it because we are on thinking-sphinx v3

https://freelancing-gods.com/thinking-sphinx/v3/advanced_config.html#indexed-models

hallelujah (approved):
Good for me :rocket: 

There is a missing test for metrics_index but it is small enough that we do not have to do one ;)

damianpm (approved):
In general LGTM, the only thing that I don't like so much is the fact that it does a lot of url manipulation imperatively.
We have route handing in rails, and now we have some small pieces in js...
But, that's how Porta works

PR #2438, opened 2021-05-05
Refactor js utils module

josemigallas:
This is a proposal for improving the structure of our utilities. Maybe we should also follow a naming convention.

### DIFF
--- a/spec/javascripts/ActiveDocs/OAS3Autocomplete.spec.jsx
+++ b/spec/javascripts/ActiveDocs/OAS3Autocomplete.spec.jsx
@@ -2,7 +2,7 @@
 
 // $FlowIgnore[cannot-resolve-module] OAS3Autocomplete module not found by Flow since not included in name-mapper.
 import { autocompleteOAS3 } from 'OAS3Autocomplete'
-import * as utils from 'utilities/utils'
+import * as utils from 'utilities/fetchData'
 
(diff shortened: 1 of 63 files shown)


didierofrivia (approved):
🍶

spec/javascripts/ActiveDocs/OAS3Autocomplete.spec.jsx
```diff
@@ -2,7 +2,7 @@
 
 // $FlowIgnore[cannot-resolve-module] OAS3Autocomplete module not found by Flow since not included in name-mapper.
 import { autocompleteOAS3 } from 'OAS3Autocomplete'
-import * as utils from 'utilities/utils'
+import * as utils from 'utilities/fetchData'
```
damianpm, line 5: why `import * as utils from 'utilities/fetchData'` here? 
```suggestion
import * as utils from 'utilities'
```

damianpm (changes requested):

spec/javascripts/ActiveDocs/OAS3Autocomplete.spec.jsx
josemigallas: It needs to be only the fetch module and I wanted the fewest line changes possible.

damianpm (approved):

PR #2439, opened 2021-05-05
[THREESCALE-6468] Adds proxy config affecting changes to the gateway_configuration

jsmadis:

When `ClientID Token Claim` or `ClientID Token Claim Type` were changed it didn't show the promote to the new staging version. `Gateway_configuration` table wasn't tracked by the `proxy_config_affecting_changes` and this PR is fixing that.
 
https://issues.redhat.com/browse/THREESCALE-6468

**Verification steps** 

### DIFF
--- a/test/unit/proxy_config_affecting_changes_test.rb
+++ b/test/unit/proxy_config_affecting_changes_test.rb
@@ -66,7 +66,7 @@ class ProxyConfigAffectingChangesTest < ActiveSupport::TestCase
       define_proxy_config_affecting_attributes :settings
       self.table_name = 'gateway_configurations'
 
-      store :settings, accessors: [:jwt_claim_with_client_id, :jwt_claim_with_client_id_type], coder: JSON
+      store :settings, accessors: %i[jwt_claim_with_client_id jwt_claim_with_client_id_type], coder: JSON
     end
 
     test 'tracks proxy config affecting changes on attribute write' do
--- a/test/unit/proxy_config_affecting_changes_test.rb
+++ b/test/unit/proxy_config_affecting_changes_test.rb
@@ -60,6 +60,38 @@ class ProxyConfigAffectingChangesTest < ActiveSupport::TestCase
     end
   end
 
+  class StoredModelTest < ActiveSupport::TestCase
+    class StoredModel < ActiveRecord::Base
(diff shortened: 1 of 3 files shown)


test/unit/proxy_config_affecting_changes_test.rb
```diff
@@ -60,6 +60,38 @@ def self.column_names
     end
   end
 
+  class StoredModelTest < ActiveSupport::TestCase
+    class StoredModel < ActiveRecord::Base
+      include ProxyConfigAffectingChanges::ModelExtension
+      define_proxy_config_affecting_attributes :settings
+      self.table_name = 'gateway_configurations'
+
+      store :settings, accessors: [:jwt_claim_with_client_id, :jwt_claim_with_client_id_type], coder: JSON
```
guicassolato, line 69: Not important, but codeclimate will probably suggest:

```suggestion
      store :settings, accessors: %i[jwt_claim_with_client_id jwt_claim_with_client_id_type], coder: JSON
```

test/unit/proxy_config_affecting_changes_test.rb
```diff
@@ -60,6 +60,38 @@ def self.column_names
...
+      define_proxy_config_affecting_attributes :settings
+      self.table_name = 'gateway_configurations'
+
+      store :settings, accessors: %i[jwt_claim_with_client_id jwt_claim_with_client_id_type], coder: JSON
+    end
+
+    test 'tracks proxy config affecting changes on attribute write' do
+      within_thread do
+        tracker = ProxyConfigAffectingChanges::Tracker.new
+        Thread.current[ProxyConfigAffectingChanges::TRACKER_NAME] = tracker
```
guicassolato, line 75: I see some usage for this in other tests as well (and indeed it is already being used elsewhere), so WDYT about moving a (protected) method?

```suggestion
        tracker = build_tracker
```

```ruby
  protected

  def build_tracker
    Thread.current[ProxyConfigAffectingChanges::TRACKER_NAME] = ProxyConfigAffectingChanges::Tracker.new
  end
```

jsmadis: Since the other tests are using it as well I moved it to the static method instead

guicassolato (approved):

PR #2442, opened 2021-05-10
Removes hidden element from Service overview

josemigallas:
This is part of the cleanup of APIAP. I noticed it causes a null pointer exception in preview due to proxy being `nil`, even when this element is going to be hidden because of the `is-hidden` class.

The element was previously rendered here:
Screenshot 2021-05-10 at 17 26 54

[THREESCALE-7061: Remove permanently hidden elements from service overview](https://issues.redhat.com/browse/THREESCALE-7061)


hallelujah (approved):
<3 Nice cleanup thanks!

PR #2444, opened 2021-05-11
Clean up APIAP traces in Services Controller

josemigallas:
Clean up old APIAP related methods from `ServicesController`.

[THREESCALE-7065: Clean up service controller unused methods](https://issues.redhat.com/browse/THREESCALE-7065)


hallelujah (approved):

app/controllers/api/services_controller.rb
```diff
@@ -54,28 +54,28 @@ def create
     creator = ServiceCreator.new(service: @service)
 
     if can_create? && creator.call(create_params)
-      flash[:notice] = t('flash.services.create.notice', resource_type: product_or_service_type)
+      flash[:notice] = t('flash.services.create.notice', resource_type: 'Product')
```
guicassolato, line 57: If it's always "Product" now, do we still need `resource_type`?

app/views/api/services/forms/_integration_settings.html.slim
```diff
@@ -1,150 +1,29 @@
...
-  $(document).ready(function(){
-    // warning when changing auth mode
-    var auth_mode_changed = false;
-    var current_auth_mode = "#{@service.proxy.authentication_method}";
+= javascript_pack_tag 'settings'
 
-    $("input[name='service[backend_version]']").change( function(){
-      auth_mode_changed = true;
-    });
+- @service.deployment_option ||= 'hosted'
```
guicassolato, line 3: Not sure if I like changing the state of `@service` in a view. Can we use a helper, or a decorator that overrides `Service#deployment_option` with something like `super || 'hosted'`?

app/views/api/services/settings.html.slim
```diff
@@ -1,22 +1,6 @@
...
-            p.inline-hints= @service.service_plans.first.class.human_attribute_name 'default_plan_hint'
-          - else
-            p.inline-hints Please create a service plan first.
-
 = semantic_form_for @service, :url => admin_service_path(@service) do |form|
-  = render :partial => 'api/services/forms/settings', :locals => { :form => form }
+  = render :partial => 'api/services/forms/integration_settings', :locals => {:form => form}
   = form.buttons do
-    = form.commit_button 'Update Product'
+    = form.commit_button "Update Product", button_html: { name: 'update_settings', class: 'pf-c-button pf-m-primary' }
```
guicassolato, line 6: Nothing wrong with the single quotes here 😉

```suggestion
    = form.commit_button 'Update Product', button_html: { name: 'update_settings', class: 'pf-c-button pf-m-primary' }
```

guicassolato (commented):
I've left a few comments. Most of them are not related to the clean-up itself but things that could have been done differently in the past and now just happen to be touched again with clean-up. We may or may not want to fix them now.

app/controllers/api/services_controller.rb
josemigallas: Good catch

app/views/api/services/forms/_integration_settings.html.slim
josemigallas: OK I'm not sure why this is needed but I moved it the closest where it is used, in `integrations_helper`.

guicassolato (approved):
I haven't tried myself – neither locally, nor on OpenShift – but it LGTM.

josemigallas:
> I haven't tried myself – neither locally, nor on OpenShift – but it LGTM.

It works locally and on preview. Merging.

PR #2448, opened 2021-05-12
[rake] Update the value of APIcast master access token

hallelujah:
This task allow restoring the APIcast master access token

### DIFF
--- a/lib/tasks/access_tokens.rake
+++ b/lib/tasks/access_tokens.rake
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+namespace :access_tokens do
+  namespace :update do
+    desc 'Update APIcast master access token'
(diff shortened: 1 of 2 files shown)


lib/tasks/restore/access_token.rake
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+namespace :restore do
```
guicassolato, line 3: Are we planning on having/moving other tasks under `lib/tasks/restore`?

hallelujah: Well I could just create without namespace but I have no idea

hallelujah: it could be `access_tokens:master:update` wdyt?

hallelujah:
Closing in favor of #2462

PR #2450, opened 2021-05-14
Fedora instructions update and small fix-ups

akostadinov:

Updating Fedora instructions. Avoid prometheus errors after clean checkout.
And other small fixups.

no issue

**Verification steps**
 
- Follow Fedora 34 installation instructions and look at console log.
- IntelliJ able to run full minitest suite
- IntelliJ files are ignored by Git

**Note**

Commits should be independently reviewable.


INSTALL.md
```diff
@@ -230,9 +230,11 @@ ruby -v
 
 #### Node version
 
-The project supports **Version: 8.X.X**.
+The project supports **Version: 10.X.X**.
 
-Install Node version 8 and ensure that the node is properly configured in the `PATH` environment variable.
+```bash
+dnf mosule install nodejs:10
```
jsmadis, line 236: Shouldn't it be a module?

```suggestion
dnf module install nodejs:10
```

.env.example
```diff
@@ -0,0 +1,10 @@
+MASTER_PASSWORD=master_password_changeme
+USER_PASSWORD=admin_password_changeme
+APICAST_REGISTRY_URL=https://apicast-staging.proda.3sca.net/policies
+PROMETHEUS_EXPORTER_PORT=9398
```
hallelujah, line 4: Sadly you cannot really use that if you launch sidekiq and the webserver at the same time on the same host

Sidekiq will launch its own webserver with this environment variable. And the application webserver will listen to the same port ... so there will be a conflict

Not sure how to solve this with `.env` file without having 2 different env names

.env.example
```diff
@@ -0,0 +1,10 @@
+MASTER_PASSWORD=master_password_changeme
```
hallelujah, line 1: I'd say use a really easy to type password e.g. `secret` 
For somebody that copy/paste this example file, no need to edit thus more convenient

.env.example
```diff
@@ -0,0 +1,10 @@
...
+USER_PASSWORD=admin_password_changeme
+APICAST_REGISTRY_URL=https://apicast-staging.proda.3sca.net/policies
+PROMETHEUS_EXPORTER_PORT=9398
+ADMIN_ACCESS_TOKEN=secret
+APICAST_ACCESS_TOKEN=apicastsecret_changeme
```
hallelujah, line 6: Well the same here and following envs, use simple `token` string ... try to make it easy to setup.

hallelujah (changes requested):

.env.example
akostadinov: Do you think this can be made clear?

.env.example
hallelujah: Well security on your local machine? I don't understand. How is it secure? Or what will be the issues?
Remember, developers do not like to configure ... Just copy/paste and good to go: "sane defaults"

.env.example
hallelujah: How do you practically use `--index` without introducing another environment variable? This looks more complicated to me to setup and understand than having an alias or letting people manage their own environment variables.

Btw, some of us use aliases:

```shell
alias porta_sidekiq="PROMETHEUS_PORT=9593 bundle exec sidekiq"
alias porta_server="PROMETHEUS_PORT=9394 bundle exec rails s"
```

Also sidekiq 6 removes the `--index` option so I do not really know if we need to commit with that.

I really do not know what dotenv is adding to the project. It was setup at the beginning but I think it is abandoned for reason.

Also it raises the entry barrier by a lot when you need to configure some environments in order to boot the app and make it secure... Especially when this project is already complicated to setup.

Locally, it is fine if the secret is called `secret` as long as bootstrapping the application is not bothersome. I think it is a convention that all applications/softwares have in common. Think about your factory configured router/tv with credentials `admin/admin`. Nobody really expect `some_user/change_me_i_am_not_secure`

Yes for production use you should use something else but not a file either :)

akostadinov:
@hallelujah , I removed `.env.example` altogether and added some info about environment variables to `INSTALL.md`. Please let me know if you can suggest further improvements.

.env.apicast
```diff
@@ -1,3 +0,0 @@
-THREESCALE_PORTAL_ENDPOINT=http://master-account-admin.example.com:3000
```
hallelujah, line 1: Good catch on this file, it was used in a really old Dockerfile but we forgot to remove this
For the history:

- The former Dockerfile and Makefile had the roles to build a local integration environment through docker-compose
- We removed it because it was not maintained and was hard to maintain

My opinion on local integration. I prefer a shared project that will spin up the whole 3scale or part of it in another project. And each service could point either to the container image or a local git version.

Something like in another project:

```
# Edit some configuration file to point to your local git directories or container
# Then
make 3scale
```

And it will spin all the components respectively for integration
A bit like https://github.com/guicassolato/porta-dev-tools

hallelujah (dismissed):
Looks good to me.

I am not against keeping the .env.example but clearly simplify it.

.env.apicast
akostadinov: > I am not against keeping the .env.example but clearly simplify it.

Removed because you said that you want to remove `.env` support some day. So I thought not to promote its usage.

hallelujah: I mean dotenv and dotenv-rails ... ;)
FYI, you can still use `.env` with autoenv which is a shell integration

.gitignore
```diff
@@ -14,7 +14,6 @@
 .deploy_info
 .ruby-gemset
 .ruby-version
-.tool-versions
```
hallelujah, line 17: I wonder why removing this. Do you want to support it officially?

tmp/metrics/.gitignore
```diff
@@ -0,0 +1,2 @@
+*
+!.gitignore
```
hallelujah, line 2: Isn't `tmp/*` just ignored?

hallelujah (changes requested):

tmp/metrics/.gitignore
akostadinov: The problem is that if `/tmp/metrics` does not exist, then a lot of errors are raised in the log. This is to make sure that the directory exists after repo is cloned and user does not need to create it manually.

.gitignore
akostadinov: it is a redundant entry, we need it only once

.env.apicast
akostadinov: IMO whoever wants to use `autoenv` or something else, will know how to do. I tried to document env variables so that anybody can set them as they like. Makes sense? Or do you want the file back?

tmp/metrics/.gitignore
hallelujah: You can just use `.gitkeep`, but wondering, can't you create the directory before running the metrics. it is only for IntelliJ right? Also if so they should fix that in their plugin.

akostadinov:
Splitting this one in pieces.

PR #2452, opened 2021-05-17
Remove "create application" button in multitenant portal

josemigallas:
Since the New Application form doesn't have server-side pagination, the page freezes due to the large amount of Accounts in Multitenant. We're hiding the buttons for now while we work on a more scalable solution.


hallelujah (approved):

hallelujah:
Closing, a proper fix is in the work

PR #2453, opened 2021-05-17
Fix Typeahead select filter

josemigallas:

The typeahead filter in our Select component does not work as it should. In order to fix it we need to define our own `onFilter` callback.

Before:
Screenshot 2021-05-17 at 13 20 53

Now:
Screenshot 2021-05-17 at 13 20 23

### DIFF
--- a/app/javascript/src/Common/components/Select.jsx
+++ b/app/javascript/src/Common/components/Select.jsx
@@ -45,6 +45,19 @@ const Select = <T: Record>({
     onSelect(selected || null)
   }
 
+  const handleFilter = (e: SyntheticInputEvent<HTMLInputElement>) => {
+    const { value } = e.currentTarget
+    let filteredItems = items
+
+    if (value !== '') {
+      const term = new RegExp(value, 'i')
+      filteredItems = items.filter(i => term.test(i.name))
+    }
+
+    // $FlowIssue[prop-missing] className and disabled are optional
+    return filteredItems.map(toSelectOption)
+  }
+
   return (
     <FormGroup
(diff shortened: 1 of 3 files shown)


app/views/api/applications/index.html.slim
```diff
@@ -1,4 +1,4 @@
 h1 Applications
-= pf_button_in_title 'Create Application', new_admin_service_application_path(@service)
+= pf_button_in_title 'Create Application', new_admin_service_application_path(@service) unless site_account.master?
```
thomasmaas, line 2: Not sure this should be part of this pr as it has nothing to do with the typeahead filter? Also, the problem of with the new application page where a drop down of accounts causes page time outs could occur for any provider as well if they have a large amount of developer accounts, no?

app/views/buyers/applications/index.html.slim
```diff
@@ -13,7 +13,7 @@ h1
     ' for
     = @account.name
 
-= pf_button_in_title 'Create Application', create_application_link_href(@account)
+= pf_button_in_title 'Create Application', create_application_link_href(@account) unless site_account.master?
```
thomasmaas, line 16: same as above: out of scope.

thomasmaas (changes requested):

app/views/api/applications/index.html.slim
josemigallas: Thanks for spotting this

thomasmaas (approved):

damianpm (approved):

PR #2454, opened 2021-05-18
[cve] Update bundler to 2.2.18

hallelujah:
Closes CVE-2020-36327 https://issues.redhat.com/browse/THREESCALE-7085

Is required by https://github.com/3scale/deploy/pull/706

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -75,7 +75,7 @@ attach-to-workspace: &attach-to-workspace
     at: .
 
 system-builder-ruby25: &system-builder-ruby25
-  image: quay.io/3scale/system-builder:ruby25
+  image: quay.io/3scale/system-builder:ruby25-bundler-2.2.18
   environment:
     BUNDLE_FROZEN: true
     BUNDLE_PATH: 'vendor/bundle'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1005,4 +1005,4 @@ DEPENDENCIES
   zip-zip
 
 BUNDLED WITH
-   1.17.3
+   2.2.18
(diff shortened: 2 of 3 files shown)


HemaHG (dismissed):

gsaslis (commented):
I see some more places where bundler is installed: 

- https://github.com/3scale/porta/blob/2a05a51a6ef35967f3c438ceff3a5d548f5bcd6a/openshift/system/Dockerfile#L49
- https://github.com/3scale/porta/blob/2a05a51a6ef35967f3c438ceff3a5d548f5bcd6a/bin/update#L18
- https://github.com/3scale/porta/blob/2a05a51a6ef35967f3c438ceff3a5d548f5bcd6a/bin/setup#L20

Should the bundler version be pinned there too perhaps?

gsaslis (dismissed):
I think we should at least fix the Dockerfile

gsaslis (commented):
@hallelujah i am not sure how the `porta/bin/setup` and `porta/bin/update` files are used. If you think those don't need pinning the bundler version, to ensure 2.2.17 is always used, then this is ok to merge by me.

hallelujah:
There is no version in those files and they are generated by rails

gsaslis:
@hallelujah it looks like there is something weird going on with the versions of bundler... ? looks like the fix went into 2.2.10 and then rolled back in 2.2.11 ?

https://github.com/rubysec/ruby-advisory-db/pull/468#issuecomment-840029917

hallelujah:
So what should we do? Roll back to 2.2.10 but why did they revert the change? Maybe because it is a breaking change?

gsaslis:
@hallelujah i am trying to figure that out myself. not sure if you've found some answer to that question already. Otherwise, it seems like updating to 2.2.10 would be a good solution. Then we can update to some later version that includes the fix again.

gsaslis (approved):
lgtm !

PR #2455, opened 2021-05-18
[cve] Update bundler to 2.2.18 on 2.10-stable

hallelujah:
Closes CVE-2020-36327 https://issues.redhat.com/browse/THREESCALE-7085

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM quay.io/3scale/system-builder:ruby25-oracle
+FROM quay.io/3scale/system-builder:ruby25-bundler-2.2.18
 
 ARG CUSTOM_DB=mysql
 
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM quay.io/3scale/system-builder:ruby25-oracle
+FROM quay.io/3scale/system-builder:ruby25-bundler-2.2.18
```
thomasmaas, line 1: we don't need something oracle specific here @hallelujah ?

thomasmaas (approved):

Dockerfile
hallelujah: Good spotting it! But we do not need because the branch was merged into ruby-25 already.

PR #2456, opened 2021-05-18
Display "Show all" button only when items > 20

josemigallas:
A select with less than 21 items won't show the sticky button:
Screenshot 2021-05-18 at 11 32 08

But one with more than 20 will:
Screenshot 2021-05-18 at 11 38 07


cntlsn (approved):

spec/javascripts/Common/components/SelectWithModal.spec.jsx
```diff
@@ -50,20 +50,10 @@ it('should render itself', () => {
   expect(wrapper.exists()).toBe(true)
 })
 
-it('should display all items, a title and a sticky button', () => {
+it('should display all items and a title, but no sticky footer', () => {
```
damianpm, line 53: Add: with <= 20

damianpm (approved):

PR #2459, opened 2021-05-19
Update new Application form to load data async

josemigallas:

When the new New Application page loads all buyers in our Multitenant account it freezes due to the large amount. This can potentially happen for any other provider!

The solution is to download only a limited set (the 20 most recently updated accounts) that will be displayed in the select and when the user opens the modal and navigates to an empty page, a request will be sent to the server to populate it.

Pages where this is used:
- Audience > Applications > New application
  - Buyers and Products select
- Product page > Applications > New application
  - Buyers select
- Account page > Applications > New application
  - Products select

Screenshot 2021-05-31 at 11 01 31

Screenshot 2021-05-31 at 11 01 34

Screenshot 2021-05-31 at 10 07 26

**Fixing Service plan logic**:

There is a buyer (Account 0) and 3 products (API, API 2 and API3). The buyer is subscribed to API, API 2 has no service plans and API 3 has 2 plans. Verify the 'Service plan' select behaves appropriately in the following situations:

For API, it is not possible to change the service plan since a contract between buyer and service exists:

Screenshot 2021-09-23 at 16 50 38
-
For API 2, it is no possible to create the application because it has no service plans and it is required to create an application:

Screenshot 2021-09-23 at 16 50 56
-
For API 3, when there is no default service plan, the first one will be selected (this is to make it consistent with the backend code although it doesn't make sense)

Screenshot 2021-09-23 at 16 51 25
-
For API 3 again but this time there's a default plan, it will be selected by default

Screenshot 2021-09-23 at 16 51 44

[THREESCALE-7030: Address scalability in 'Create new Application' form](https://issues.redhat.com/browse/THREESCALE-7030)

**Verification for N+1 queries / underlying performance**:

Audience new app (http://provider-admin.lvh.me:3000/p/admin/applications/new)
- [ ] buyers and products on initial render
- [ ] buyers modal (pagination / search)
- [ ] products modal (pagination / search)

Product new app (http://provider-admin.lvh.me:3000/apiconfig/services/4/applications/new)
- [ ] current product data
- [ ] buyers on initial render
- [ ] buyers modal (pagination / search)

Buyer new app (http://provider-admin.lvh.me:3000/buyers/accounts/3/applications/new)
- [ ] current buyer data
- [ ] products on initial render
- [ ] products modal (pagination / search)


app/decorators/account_decorator.rb
```diff
@@ -3,9 +3,30 @@
...
+  self.include_root_in_json = false
+
+  def new_application_data
+    {
+      id: id.to_s,
+      name: name,
+      description: "Admin: #{admin_user_display_name}",
+      createdAt: created_at.to_s(:long),
+      contractedProducts: contracts,
+      createApplicationPath: h.admin_buyers_account_applications_url(object),
```
hallelujah, line 15: I would use `path` instead of `url` it is generally accepted convention that website use relative path to their domain instead of including domain
```suggestion
      createApplicationPath: h.admin_buyers_account_applications_path(object),
```

features/provider/admin/applications/create_application.feature
```diff
@@ -0,0 +1,93 @@
...
+    Then I log in as provider "foo.3scale.localhost"
+
+  Scenario: Create a an application
+    When I create an application "My App" from the audience context
+    Then I should be on the provider side "My App" application page
+    And should see "Application was successfully created"
+    And buyer "bob" should have 1 cinstance
+
+  Scenario: Create an application in single application mode
+    Given provider "foo.3scale.localhost" has multiple applications disabled
```
hallelujah, line 22: This has nothing to do with the new cucumber. If you still use that syntax please let those cucumber remain in the `old` directory

hallelujah (changes requested):

app/controllers/api/services_controller.rb
```diff
@@ -18,13 +18,13 @@ class Api::ServicesController < Api::BaseController
...
-                            .scope_search(search)
-    @page_services = @services.paginate(pagination_params)
-                              .decorate
-                              .to_json(only: %i[name updated_at id system_name], methods: %i[links apps_count backends_count unread_alerts_count])
+    @presenter ||= Api::ServicesIndexPresenter.new(user: current_user, search_params: params[:search] || params, pagination_params: pagination_params)
+    services = @presenter.paginated_services
+
+    respond_to do |format|
+      format.html
+      format.json { render json: { items: services.map(&:new_application_data), count: services.total_entries } }
```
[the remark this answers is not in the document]
josemigallas: > So you make the same request but with JSON from your view?

Initial render via HTML template, then the modal needs to send a request from JS, so only JSON is needed. Both are the same `index` request, that's why I didn't create a new route for this.

> Do we always use every object in this controller for both index display and search?
 
I don't think so, only services.

features/provider/admin/applications/create_application.feature
josemigallas: OK in the end I'll move them to old, I think it's more important to merge this.

app/javascript/src/NewApplication/components/NewApplicationForm.jsx
```diff
@@ -123,20 +139,24 @@ const NewApplicationForm = ({
...
 
-        {buyers ? (
+        {mostRecentlyCreatedBuyers ? (
           <BuyerSelect
             buyer={buyer}
-            buyers={buyers}
+            mostRecentlyCreatedBuyers={mostRecentlyCreatedBuyers}
+            buyersCount={buyersCount}
             onSelectBuyer={setBuyer}
+            buyersPath="/buyers/accounts.json" // TODO: pass url in props
```
damianpm, line 148: Same question, for this PR?

app/javascript/src/NewApplication/data/Products.js
```diff
@@ -0,0 +1,27 @@
...
+  })
+
+  if (query !== '') {
+    searchParams.append('search[query]', query)
+    searchParams.append('utf8', '✓')
+  }
+
+  const url = `${path}?${searchParams.toString()}`
+
+  return getJSON(url).then(data => data.json())
```
damianpm, line 24: Maybe some error handling here or in the method definition

spec/javascripts/BackendApis/components/BackendSelect.spec.jsx
```diff
@@ -34,14 +34,19 @@ it('should render itself', () => {
...
-
-  expect(wrapper.find('TableModal').prop('cells')).toMatchObject([
-    { title: 'Name', propName: 'name' },
-    { title: 'Private Base URL', propName: 'privateEndpoint' },
-    { title: 'Last updated', propName: 'updatedAt' }
-  ])
+describe('when there are more than 20 backends', () => {
+  const backends = new Array(21).fill({}).map((_, i) => ({ id: i, name: `API ${i}`, privateEndpoint: `api/${i}`, systemName: `api_${i}` }))
+
+  it('should have a table with Name, Private Base URL and Last updated', () => {
```
damianpm, line 40: with Name, Admin and Signup date ? That's what I see rendered

app/javascript/src/NewApplication/components/NewApplicationForm.jsx
josemigallas: It should be done as part of this PR.

app/javascript/src/Common/components/SelectWithPaginatedModal.jsx
```diff
@@ -0,0 +1,153 @@
...
+          cells={cells}
+          isOpen={modalOpen}
+          isLoading={isLoading}
+          selectedItem={item}
+          pageItems={pageDictionary[page]}
+          itemsCount={count}
+          onSelect={handleOnModalSelect}
+          onClose={() => {
+            setModalOpen(false)
+            // TODO: cancel ongoing requests
```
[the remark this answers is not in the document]
josemigallas: It's not critical but IMO it should be done in this PR.

app/javascript/src/NewApplication/data/Products.js
josemigallas: There was at first but then I thought it wasn't useful. If errors are handled at this level, then something like `undefined` or `[]` has to be returned and then the component won't know there's been an error.
By handling it at the component level we can update the UI with a meaningful error message, retry the request, etc.

thomasmaas:
Works well. 2 visual issues (possibly PatternFly)

1. height of modal changes during fetching which feels clumsy
2. width of numbered input field is fixed and only fits one digit (Safari)

app/decorators/account_decorator.rb
```diff
@@ -3,9 +3,30 @@
...
   private
 
   def admin_user
     @admin_user ||= (super || User.new).decorate
   end
+
+  def contracts
+    bought_service_contracts.map do |contract|
+      hash = contract.service.as_json(only: %i[id name], root: false)
+      hash['withPlan'] = contract.plan.as_json(only: %i[id name], root: false)
```
hallelujah, line 29: Are you sure you want the single value `hash['withPlan'` and not the entire `hash` in the Array?

josemigallas: It's actually mapping the hash `id` and `name` 🤔 withPlan is lost and still the logic doesn't break. Maybe I added some redundancy? Could you check the following logic makes sense?
https://github.com/3scale/porta/blob/e24aca233b76b5f727b065ac4095329247c2f27e/app/javascript/src/Logic/BuyerLogic.js#L12-L15

cntlsn (dismissed):

app/javascript/src/NewApplication/components/BuyerSelect.jsx
```diff
@@ -2,45 +2,54 @@
...
+    fieldId="account_id"
+    id="account_id"
+    name="account_id"
+    item={buyer}
+    items={mostRecentlyCreatedBuyers}
+    itemsCount={buyersCount}
+    cells={cells}
+    onSelect={onSelectBuyer}
+    fetchItems={(params) => fetchPaginatedBuyers(buyersPath, params)}
+    onAbortFetch={() => console.log('abort')}
```
damianpm, line 46: There is no `onAbortFetch` prop in https://github.com/3scale/porta/blob/2.11/THREESCALE-7030_new_app_typeahead/app/javascript/src/Common/components/SelectWithModal.jsx#L32

And `console.log` will make lint fail

app/javascript/src/NewApplication/components/ProductSelect.jsx
```diff
@@ -2,42 +2,54 @@
...
+    fieldId="account_id"
+    id="product"
+    name=""
+    item={product}
+    items={mostRecentlyUpdatedProducts}
+    itemsCount={productsCount}
+    cells={cells}
+    onSelect={onSelectProduct}
+    fetchItems={(params) => fetchPaginatedProducts(productsPath, params)}
+    onAbortFetch={() => console.log('abort')}
```
damianpm, line 46: Same from prev comment

spec/javascripts/NewApplication/components/NewApplicationForm.spec.jsx
```diff
@@ -159,88 +538,167 @@ describe('when in Audience context', () => {
...
+          selectProduct(wrapper, selectedProduct)
 
-    act(() => wrapper.find('UserDefinedField').at(1).props().onChange('Active'))
-    wrapper.update()
-    expect(isButtonDisabled()).toBe(false)
+          expect(wrapper.find('Select[name="cinstance[service_plan_id]"]').prop('item')).toBe(servicePlans[0])
+          expect(wrapper.find('input[type="hidden"][name="cinstance[service_plan_id]"]').exists()).toBe(true)
+        })
+      })
+    })
```
damianpm, line 702: Do we need such a great test file? It's testing a lot...

damianpm (approved):
JS side, LGTM

hallelujah (approved):
Thank you!
:shipit:

hallelujah (approved):

PR #2462, opened 2021-05-26
[feature] K8s management tasks

hallelujah:
In K8s context, pods are deployed and we spin off some management tasks like:

- Create a default proxy for services that do not have one
- Restore the Master APIcast access token to the one set in pod environment

We need to make those tasks reboot proof thus introducing activejob unique jobs

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -17,6 +17,7 @@ gem 'dotenv-rails', '~> 2.7'
 gem 'rails', '~> 5.0.7'
 
 # Needed for XML serialization of ActiveRecord::Base
+gem "activejob-uniqueness", "~> 0.2.0"
 gem 'activemodel-serializers-xml'
 
 gem 'protected_attributes_continued', '~> 1.3.0'
@@ -244,6 +245,7 @@ group :test do
 
   # performance tests
   gem 'ruby-prof'
+  gem 'with_env'
 end
 
 group :development, :test do
--- a/app/workers/restore_apicast_master_token_worker.rb
+++ b/app/workers/restore_apicast_master_token_worker.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+# This will only perform in the context of k8s or OpenShift
+# It sets the value of the access token from an environment variable
+class RestoreApicastMasterTokenWorker < ApplicationJob
+  unique :until_executed
+
+
+  def perform(*)
+    if ENV['APICAST_ACCESS_TOKEN'].blank?
+      Rails.logger.info 'Cannot execute this job without an access token'
+      return
+    end
+
+    token_name = ENV.fetch('APICAST_TOKEN_NAME', 'APIcast mapping-service')
+    token = ENV.fetch('APICAST_ACCESS_TOKEN')
+    master = Account.master
+    access_token = master.access_tokens.find_by!(name: token_name)
+    # Need to do that because `:value` is a readonly attribute
+    AccessToken.where(id: access_token.id).limit(1).update_all(value: token) # rubocop:disable Rails/SkipsModelValidations
+  end
+end
--- a/config/initializers/active_job_uniqueness.rb
+++ b/config/initializers/active_job_uniqueness.rb
@@ -1,40 +1,44 @@
 # frozen_string_literal: true
 
-ActiveJob::Uniqueness.configure do |config|
-  # Global default expiration for lock keys. Each job can define its own ttl via :lock_ttl option.
-  # Stategy :until_and_while_executing also accept :on_runtime_ttl option.
-  #
-  # config.lock_ttl = 1.day
+Rails.application.configure do
+  config.after_initialize do
+    ActiveJob::Uniqueness.configure do |config|
+      # Global default expiration for lock keys. Each job can define its own ttl via :lock_ttl option.
+      # Stategy :until_and_while_executing also accept :on_runtime_ttl option.
+      #
+      # config.lock_ttl = 1.day
 
-  # Prefix for lock keys. Can not be set per job.
-  #
-  # config.lock_prefix = 'activejob_uniqueness'
+      # Prefix for lock keys. Can not be set per job.
+      #
+      # config.lock_prefix = 'activejob_uniqueness'
 
-  # Default action on lock conflict. Can be set per job.
-  # Stategy :until_and_while_executing also accept :on_runtime_conflict option.
-  # Allowed values are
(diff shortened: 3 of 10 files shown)


Gemfile
```diff
@@ -17,6 +17,7 @@ gem 'dotenv-rails', '~> 2.7'
 gem 'rails', '~> 5.0.7'
 
 # Needed for XML serialization of ActiveRecord::Base
+gem "activejob-uniqueness", "~> 0.2.0"
```
guicassolato, line 20: How about?

```suggestion
gem 'activejob-uniqueness', require: 'active_job/uniqueness/sidekiq_patch'
```

config/initializers/active_job_uniqueness.rb
```diff
@@ -0,0 +1,40 @@
...
+  # config.on_conflict = :raise
+
+  # Digest method for lock keys generating. Expected to have `hexdigest` class method.
+  #
+  # config.digest_method = OpenSSL::Digest::MD5
+
+  # Array of redis servers for Redlock quorum.
+  # Read more at https://github.com/leandromoreira/redlock-rb#redis-client-configuration
+  #
+  # config.redlock_servers = [ENV.fetch('REDIS_URL', 'redis://localhost:6379')]
```
guicassolato, line 29: In Porta we also provide support for Redis Sentinel and Redis namespace. See, e.g., https://github.com/3scale/porta/blob/aa9897594ad654f9c54ea1ababcbc4e0b8fed165/config/application.rb#L267

Is it okay to rely only on `REDIS_URL` here?

app/workers/restore_apicast_master_token_worker.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+# This will only perform in the context of k8s or OpenShift
+# It sets the value of the access token from an environment variable
+class RestoreApicastMasterTokenWorker < ApplicationJob
+  unique :until_executed,
+         on_conflict: ->(job) { Rails.logger.info "Another job is already scheduled for: #{job.class} #{job.arguments}" }
```
guicassolato, line 7: It looks like we'll be reusing this proc in the future. Worth defining it (as a class helper method) in `ApplicationJob`, or in a mixin?

I know it can get ugly, so feel free to ignore if you prefer starting as-is for now.

test/workers/restore_apicast_master_access_token_worker_test.rb
```diff
@@ -0,0 +1,22 @@
...
+  include ActiveJob::TestHelper
+  include WithEnv
+
+  def test_update_apicast_master_token
+    FactoryBot.create_list(:access_token, 2)
+    apicast_token_name = 'APIcast master token'
+    master_token = FactoryBot.create(:access_token, name: apicast_token_name, owner: master_account.first_admin!)
+    random_token = SecureRandom.hex
+
+    with_env 'APICAST_ACCESS_TOKEN' => random_token, 'APICAST_TOKEN_NAME' => apicast_token_name do
```
guicassolato, line 15: Just for the record, a few other places where we could leverage this gem:

```
$ rg "ENV[\[\.].+[^\|]=" features spec test
test/integration/proxy_example_test.rb
723:      ENV['THREESCALE_CONFIG_FILE'] = folder.join(package.spec_file).to_s

test/unit/db/seeds_test.rb
31:    ENV_VARIABLES.each { |env_name, _env_value| ENV[env_name] = nil }
160:    ENV_VARIABLES.each { |env_name, env_value| ENV[env_name] = env_value }

test/unit/tasks/multitenant/tenants_test.rb
17:          ENV['FILE'] = 'corrupted_accounts'
31:          ENV['FILE'] = 'corrupted_accounts'
```

Gemfile
hallelujah: I thought I would not use that for now. I do not think we need to unlock the jobs via the sidekiq-web and adding too much burden on queue. See https://github.com/veeqo/activejob-uniqueness/issues/16

- ttl is only 1 day, so lock will be released after 1 day. Let see first if we have issues with that
- if there is an issue I prefer we do it with a console command for now

config/initializers/active_job_uniqueness.rb
hallelujah: Good catch thanks.

app/workers/restore_apicast_master_token_worker.rb
hallelujah: I think it is a fair point.

config/initializers/active_job_uniqueness.rb
```diff
@@ -0,0 +1,40 @@
...
+  # config.lock_prefix = 'activejob_uniqueness'
+
+  # Default action on lock conflict. Can be set per job.
+  # Stategy :until_and_while_executing also accept :on_runtime_conflict option.
+  # Allowed values are
+  #   :raise - raises ActiveJob::Uniqueness::JobNotUnique
+  #   :log - instruments ActiveSupport::Notifications and logs event to the ActiveJob::Logger
+  #   proc - custom Proc. For example, ->(job) { job.logger.info('Oops') }
+  #
+  config.on_conflict = ->(job) { job.logger.info "Another job is already scheduled for: #{job.class} #{job.arguments}" }
```
guicassolato, line 20: Much better than what I had suggested. Thanks!

config/initializers/active_job_uniqueness.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  # Digest method for lock keys generating. Expected to have `hexdigest` class method.
+  #
+  # config.digest_method = OpenSSL::Digest::MD5
+
+  # Array of redis servers for Redlock quorum.
+  # Read more at https://github.com/leandromoreira/redlock-rb#redis-client-configuration
+  #
+  config.redlock_servers = [System.redis]
```
guicassolato, line 29: I think we patch the redis config inside of an `after_initialize` block 😬

app/workers/restore_apicast_master_token_worker.rb
hallelujah: Yes why not :)

Also I decided to use this gem as it is ActiveJob compatible, the sidekiq-lock one is not maintained since 3 years (probably because it is stable enough but who knows it is really scary)

config/initializers/active_job_uniqueness.rb
hallelujah: Ah 💩 !

guicassolato (dismissed):

PR #2465, opened 2021-05-27
configurable user session TTL

akostadinov:

Add support to configure user session time to live [THREESCALE-693](https://issues.redhat.com/browse/THREESCALE-693)

fixes THREESCALE-693

**Verification steps** 

1. set `user_sesion_ttl` in `settings.yml` to something small, like 1 minute `PT1M`
2. run porta 
3. open some porta provider console
4. login
5. check that you can navigate between pages
6. wait without activity for more than 1 minute
7. try to navigate to another page
8. you should be asked to login again

I couldn't figure out what other tests I can add. Existing tests already vefiry default TTL of 2 weeks

Still under investigation what needs to be added to OpenShift operator. In any case it should work as before OOB so operator changes can be applied any time after this is merged.

Another question is where should we document this?

### DIFF
--- a/app/models/user_session.rb
+++ b/app/models/user_session.rb
@@ -1,7 +1,5 @@
 class UserSession < ApplicationRecord
 
-  TTL = 2.weeks.freeze
-
   belongs_to :user
   belongs_to :sso_authorization, required: false
   has_one :account, through: :user
@@ -60,4 +58,18 @@ class UserSession < ApplicationRecord
     def set_unique_key
       self.key = SecureRandom.urlsafe_base64(32)
     end
+
+  class << self
+    def ttl_value(duration)
+
+      return 2.weeks if duration.blank?
+
+      ActiveSupport::Duration.seconds(Integer(duration))
+    rescue ArgumentError
+      raise ApplicationConfigurationError, "user session TTL should be specified as a number of seconds"
+    end
+  end
+
+  TTL = ttl_value(Rails.configuration.three_scale.user_session_ttl).freeze
+
 end
--- a/config/docker/settings.yml
+++ b/config/docker/settings.yml
@@ -14,6 +14,7 @@ base: &default
   events_shared_secret: <%= ENV['EVENTS_SHARED_SECRET'] %>
   recaptcha_public_key: <%= ENV.fetch('RECAPTCHA_PUBLIC_KEY', 'YOUR_RECAPTCHA_PUBLIC_KEY') %>
   recaptcha_private_key: <%= ENV['RECAPTCHA_PRIVATE_KEY'] %>
+  user_session_ttl: <%= ENV.fetch('USER_SESSION_TTL', 2.weeks) %>
   impersonation_admin:
     username: <%= ENV.fetch('IMPERSONATION_ADMIN_USERNAME', '3scaleadmin') %>
     domain: <%= ENV.fetch('IMPERSONATION_ADMIN_DOMAIN', '3scale.net') %>
--- a/config/examples/settings.yml
+++ b/config/examples/settings.yml
@@ -14,6 +14,7 @@ base: &default
   events_shared_secret: <%= ENV['EVENTS_SHARED_SECRET'] %>
   recaptcha_public_key: <%= ENV.fetch('RECAPTCHA_PUBLIC_KEY', 'YOUR_RECAPTCHA_PUBLIC_KEY') %>
   recaptcha_private_key: <%= ENV['RECAPTCHA_PRIVATE_KEY'] %>
+  user_session_ttl: <%= ENV.fetch('USER_SESSION_TTL', 2.weeks) %>
   impersonation_admin:
     username: <%= ENV.fetch('IMPERSONATION_ADMIN_USERNAME', '3scaleadmin') %>
     domain: <%= ENV.fetch('IMPERSONATION_ADMIN_DOMAIN', '3scale.net') %>
--- a/openshift/system/config/settings.yml
+++ b/openshift/system/config/settings.yml
@@ -14,6 +14,7 @@ base: &default
   onpremises: true
   recaptcha_public_key: <%= ENV.fetch('RECAPTCHA_PUBLIC_KEY', 'YOUR_RECAPTCHA_PUBLIC_KEY') %>
   recaptcha_private_key: <%= ENV['RECAPTCHA_PRIVATE_KEY'] %>
+  user_session_ttl: <%= ENV.fetch('USER_SESSION_TTL', 2.weeks) %>
 
   # System Emails
   noreply_email: <%= ENV.fetch('NOREPLY_EMAIL', "no-reply@#{superdomain}") %>
--- a/app/errors/application_configuration_error.rb
+++ b/app/errors/application_configuration_error.rb
@@ -1,0 +1,4 @@
+# frozen_string_literal: true
+
+class ApplicationConfigurationError <  StandardError
+end
--- a/app/lib/authentication/by_cookie_token.rb
+++ b/app/lib/authentication/by_cookie_token.rb
@@ -25,7 +25,7 @@ module Authentication
 
       # These create and unset the fields required for remembering users between browser closes
       def remember_me
-        remember_me_for 2.weeks
+        remember_me_for UserSession::TTL
       end
 
       def remember_me_for(time)
--- a/test/unit/user_session_test.rb
+++ b/test/unit/user_session_test.rb
@@ -72,4 +72,15 @@ class UserSessionTest < ActiveSupport::TestCase
     assert_includes stale, session2
     refute_includes stale, session3
   end
+
+  test 'parse session ttl settings value' do
+    assert_kind_of ActiveSupport::Duration, UserSession::TTL
+    assert UserSession::TTL.frozen?
+
+    assert_equal 2.weeks, UserSession.send(:ttl_value, nil)
+    assert_equal 2.weeks, UserSession.send(:ttl_value, "")
+    assert_equal 1.week, UserSession.send(:ttl_value, 1.week.seconds.to_i)
+    assert_equal 3.weeks, UserSession.send(:ttl_value, 3.weeks.seconds.to_s)
+    assert_raise(ApplicationConfigurationError) { UserSession.send(:ttl_value, "123p") }
+  end
 end


app/models/user_session.rb
```diff
@@ -1,6 +1,15 @@
 class UserSession < ApplicationRecord
 
-  TTL = 2.weeks.freeze
+  TTL = case Rails.configuration.three_scale.user_sesion_ttl
+        when nil, ""
+          2.weeks
+        when Numeric, /^\d*$/
```
guicassolato, line 6:
```suggestion
        when Numeric, /^\d+$/
```

akostadinov: Good catch, thanks!

app/models/user_session.rb
```diff
@@ -1,6 +1,15 @@
...
+  TTL = case Rails.configuration.three_scale.user_sesion_ttl
+        when nil, ""
+          2.weeks
+        when Numeric, /^\d+$/
+          # session duration specified in seconds
+          ActiveSupport::Duration.parse("PT#{Rails.configuration.three_scale.user_sesion_ttl}S")
+        else
+          # treat as ISO 8601 duration
+          ActiveSupport::Duration.parse(Rails.configuration.three_scale.user_sesion_ttl)
+        end.freeze
```
guicassolato, line 12: With a a bit of DRY maybe:

```suggestion
  user_sesion_ttl = Rails.configuration.three_scale.user_sesion_ttl.presence
  TTL = begin
          case user_sesion_ttl
          when Numeric, /^\d+$/ # session duration specified in seconds
            ActiveSupport::Duration.parse("PT#{user_sesion_ttl}S")
          else
            ActiveSupport::Duration.parse(user_sesion_ttl)
          end
        rescue ActiveSupport::Duration::ISO8601Parser::ParsingError, TypeError
          # do we care about logging?
          2.weeks
        end.freeze
```

config/examples/settings.yml
```diff
@@ -14,6 +14,8 @@ base: &default
   events_shared_secret: <%= ENV['EVENTS_SHARED_SECRET'] %>
   recaptcha_public_key: <%= ENV.fetch('RECAPTCHA_PUBLIC_KEY', 'YOUR_RECAPTCHA_PUBLIC_KEY') %>
   recaptcha_private_key: <%= ENV['RECAPTCHA_PRIVATE_KEY'] %>
+  # configure user session timeout in seconds or ISO 8601 duration
+  user_sesion_ttl: <%= ENV.fetch('USER_SESSION_TTL', "P14D") %>
```
hallelujah, line 18:
```suggestion
  user_sesion_ttl: <%= ENV.fetch('USER_SESSION_TTL',  14.days.seconds ) %>
```

openshift/system/config/settings.yml
```diff
@@ -14,6 +14,8 @@ base: &default
   onpremises: true
   recaptcha_public_key: <%= ENV.fetch('RECAPTCHA_PUBLIC_KEY', 'YOUR_RECAPTCHA_PUBLIC_KEY') %>
   recaptcha_private_key: <%= ENV['RECAPTCHA_PRIVATE_KEY'] %>
+  # configure user session timeout in seconds or ISO 8601 duration
+  user_sesion_ttl: <%= ENV.fetch('USER_SESSION_TTL', "P14D") %>
```
hallelujah, line 18:
```suggestion
  user_sesion_ttl: <%= ENV.fetch('USER_SESSION_TTL',  14.days.seconds ) %>
```

app/models/user_session.rb
```diff
@@ -1,6 +1,15 @@
 class UserSession < ApplicationRecord
 
-  TTL = 2.weeks.freeze
+  TTL = case Rails.configuration.three_scale.user_sesion_ttl
```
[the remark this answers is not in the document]
akostadinov: I also don't find the constant choice very flexible. This is the least disruptive way to implement the feature and it can change whenever we decide to allow configurable sessions per tenant and/or portal. Will keep constant if you don't mind.

will change to only seconds

config/docker/settings.yml
```diff
@@ -14,6 +14,8 @@ base: &default
   events_shared_secret: <%= ENV['EVENTS_SHARED_SECRET'] %>
   recaptcha_public_key: <%= ENV.fetch('RECAPTCHA_PUBLIC_KEY', 'YOUR_RECAPTCHA_PUBLIC_KEY') %>
   recaptcha_private_key: <%= ENV['RECAPTCHA_PRIVATE_KEY'] %>
+  # configure user session timeout in seconds or ISO 8601 duration
+  user_sesion_ttl: <%= ENV.fetch('USER_SESSION_TTL', "P14D") %>
```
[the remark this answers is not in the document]
akostadinov: Makes perfect sense, thank you.

app/models/user_session.rb
akostadinov: I usually prefer to have things fail instead of silently ignoring user specified configuration. I'm also going to remove ISO8601 parsing per other comments. Do you have a strong preference to apply default configuration in case of a bad user-supplied value?

guicassolato: This is not a critical feature, therefore, in case of bad user input, I'd be happy enough with falling back to a safe default plus a log message. However, this is no strong preference of mine, no. You should follow your heart 🙂

akostadinov: because there are security concerns about session ttl, I'd rather fail. Please check what kind of exception I'm returning, I was not sure I chose the correct one.

app/models/user_session.rb
```diff
@@ -1,6 +1,13 @@
...
```
akostadinov, line 3: @hallelujah , left it as constant for least destructive way for the change and leave refactoring for the time we decide to make configurable on a finer grain. Let me know if you are ok with that.

guicassolato (dismissed):
Good job. Let's not forget the rest of this change has to be done in https://github.com/3scale/3scale-operator.

app/models/user_session.rb
hallelujah: I really prefer not to bother with those if cases

Those are settings, we are not supposed to guard against all things a user could do. Keep it simple

hallelujah (changes requested):
I think having a constant that just refers to the config is enough, Do not guard against the user input as it is done at boot time

app/models/user_session.rb
akostadinov: @hallelujah , what change exactly do you suggest? To remove the `else` block or remove all conditions? If it is the latter, it will make IMO more complicated usage as all callers will need to do these checks instead of having them here. Or otherwise we'll have to change this to a method or something. Please let me know.

hallelujah: yes just remove the check, this is not needed really to my opinion

```suggestion
  TTL = Rails.configuration.three_scale.user_sesion_ttl
```

guicassolato: The 3scale operator usually prefers implementation where we prepare for env vars to be set with empty strings.

hallelujah: This is different. While we have to do that more often, better have it in one place for all those settings. And not have sparse code that takes care of configuration values in multiple files. In fact I am really thinking of getting rid of such constants defined dynamically and have them gathered in one single `AppSettings` instance

Fair enough, I think this is OK for now but I'd suggest to make another PR to that, below a proposed approach:


```ruby
# lib/app_settings.rb
class AppSettings
   include Singleton

  attr_reader :config

  def initialize
     # Or just include ActiveSupport::Configurable ... everything works
    @config = ActiveSupport::Configurable::Configuration.new
    super
  end

  def self.configure(key, &block)
    instance.configure(key, &block)
  end

  def configure(key, &processor)
    config[key] = processor.call
  end

  def get(key)
    config.fetch key
  end

  def self.get(key)
    instance.get(key)
  end
end
```

Then you can use it in an initializer

```ruby
# config/initializers/configuration.rb
require 'app_settings'

AppSettings.configure('user_session.ttl') do
 ENV['USER_SESSION_TTL'].presence? || 2.weeks
end
```

That way we could probably get rid of the logic in the `config/*.yml` files and have a single way to require those configuration:

```ruby
 def remember_me   
   remember_me_for AppSettings.get('user_session.ttl')
end
```

guicassolato: I like this idea, though I'd think of a path to make `AppSettings` a proper replacement for `Rails.settings.three_scale` in the future, rather than yet another settings system.

hallelujah: Yes exactly what I was thinking. Juggling with different files inside of the config/application.rb might not be the proper way to do that.

I propose would be:

1. YAML config files should be free from logic, it is hard to override them with https://github.com/3scale/3scale-amp-openshift-templates/blob/baed04acb897f5d3d25ab5fddb2fcc3778c22a0d/amp/amp.yml#L2176-L2177
2. cleanup this mess https://github.com/3scale/porta/blob/61a168335181a09d1671da684a49f23078fe933d/config/application.rb#L143-L210
3. Try to put code logic where it belongs to, meaning in a class or initializer
  * not int the config/application.rb
4. Ensure the values are well formatted

hallelujah:
Looks good, can you fix codeclimate errors?

app/models/user_session.rb
```diff
@@ -56,6 +54,19 @@ def user_agent=(value)
     self[:user_agent] = value
   end
 
+  private_class_method def self.user_session_ttl_parse(duration)
```
hallelujah, line 57: By the way you should just put this method one block below and you do not have to use `private_class_method`

akostadinov: I moved it. With `private_class_method` I wanted to make clear it is an internal utility method not for reuse. And I have to leave it, otherwise robocop gives a warning

```
private (on line 57) does not make singleton methods private. Use private_class_method or private inside a class << self block instead.
```

Moving to a utility class IMO better be done once your suggestion above for a new configuration class is implemented but let me know if you want to see it elsewhere.

Please review again. Hope no other validator will find an issue with it now.

hallelujah: I was thinking of that yes, sorry I did not make myself clear enough

private

class << self
  def user_session_ttl_parse(duration)
    # ...
  end
end

hallelujah (approved):

app/models/user_session.rb
```diff
@@ -60,4 +58,18 @@ def user_agent=(value)
     def set_unique_key
       self.key = SecureRandom.urlsafe_base64(32)
     end
+
+  private_class_method def self.user_session_ttl_parse(duration)
+    case duration
+    when NilClass, ""
```
hallelujah, line 64: you can even do 
```suggestion
    when duration.blank?
```

app/models/user_session.rb
```diff
@@ -60,4 +58,18 @@ def user_agent=(value)
...
+      2.weeks
+    when Numeric, /^\d+$/
+      ActiveSupport::Duration.seconds(duration.to_i)
+    else
+      raise ArgumentError, "user session TTL should be specified as a number of seconds"
+    end.freeze
```
hallelujah, line 70: Hmmm do you want to have the object returned by the method frozen? I would not expect that.

Another thing instead of the `case .. when`

```ruby
def user_session_ttl_parse(duration)
  return 2.weeks if duration.blank?

  ActiveSupport::Duration.seconds(Integer(duration))
rescue ArgumentError
  raise ApplicationConfigurationError, "user session TTL should be specified as a number of seconds"
end
```
Creating a tagged class to know exactly where it belongs to

hallelujah (changes requested):
I think the method should not freeze the result

hallelujah (approved):

akostadinov:
Moved freezing, added a test for it, renamed and refactored utility method, moved it inside a `class << self`

I'm concerned about moving method to `class <<self` block because it is public now and other public class methods are already defined as `def self.mymethod` within the same file so this is kind of inconsistent. Or did you mean to add `private` keyword inside the block? Sorry I'm a little confused about the intention of some of the requested changes.

I also didn't define `DEFAULT_TTL` constant to avoid extending public API with something that may not be intended to be used from outside the class. Not sure whether defining that was important to you. I can define it if you want.

hallelujah (approved):

hallelujah:
About the public method it is fine, I think it is clear now that it is a utility method but that is fine to leave it there for now. Once we do the refactoring of all the settings, we can extract it.
Also testing a private method is not really testable are they?

PR #2470, opened 2021-06-01
Refactor applications controllers

josemigallas:
**JIRA issue**
[THREESCALE-7275: Refactor Applications controllers so they better represent the 3 context](https://issues.redhat.com/browse/THREESCALE-7275)

We recently implemented 2 new pages for creating an application:

- Audience new application (new)
- Service new application (new)
- Account new application (old)

Currently there are 2 different controllers managing applications:

- `Buyers::ApplicationsController`
- `Api::ApplicationsController`

This PR unifies all common logic and delegates Applications management to a new `Provider::Admin::ApplicationsController` hosted at `p/admin/applications`.
The old controllers will handle their own `:index`, `:new` and `:create` actions, according to the following routes:

```rb
# /p/provider/applications
namespace :provider, :path => 'p' do
  namespace :admin do
    resources :applications do
      member do
        put :accept
        delete :reject
        post :suspend
        post :resume
        put :change_user_key
        put :change_plan
        get :edit_redirect_url
      end
    end
  end
end

# /apiconfig/services/:service_id/applications
resources :applications, only: %i[index new create]

# buyers/accounts/applications
resources :applications, only: %i[index new create]
```


app/controllers/api/applications_controller.rb
```diff
@@ -50,6 +50,8 @@ def show
 
   def edit; end
 
+  # This will be used with
+  # /apiconfig/services/4/applications/new
```
guicassolato, line 54: It's okay while WIP, but don't forget to remove before ready for review. This kind of comment can easily go outdated. Checking the Rails routes is usually a better way to relate controller actions and paths.

app/controllers/buyers/accounts/applications_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+                                                  .order(params[:sort] => params[:direction])
+                                                  .preload(:service, user_account: [:admin_user], plan: [:pricing_rules])
+                                                  .paginate(pagination_params)
+                                                  .decorate
+
+    display_view_portion!(:service) if current_account.multiservice?
+  end
+
+  # This will be used with
+  # /buyers/accounts/3/applications/new
```
guicassolato, line 48: 2 comments here:
1. "buyers" and "accounts" literally mean the same thing in this context, so `/buyers/accounts/{id}` feels a bit weird. Could we have either just `buyers/{id}` or (better) `accounts/{id}`?
2. We have to decide if the Admin Portal controllers will be all under `app/controllers/provider/admin` or `app/controllers/apiconfig`, one of these, one of these and other, or pretty much anywhere under `app/controllers`

app/controllers/buyers/base_controller.rb
```diff
@@ -4,19 +4,60 @@
...
   defaults :route_prefix => 'admin_buyers'
 
   activate_menu :buyers
 
+  layout 'provider'
+
+  attr_reader :buyer
+
+  # TODO: this should be done by buy! method
+  def create
```
guicassolato, line 20: This doesn't seem to belong to this controller.

app/controllers/api/applications_controller.rb
```diff
@@ -1,74 +1,36 @@
 # frozen_string_literal: true
 
-class Api::ApplicationsController < Api::BaseController
```
guicassolato, line 3: It looks like the only thing we had in `Api::BaseController` was the "bubbles" helper. I guess we don't need that for any of the different contexts for the Applications page, right?

cc/ @thomasmaas

app/controllers/api/applications_controller.rb
```diff
@@ -1,74 +1,36 @@
...
+  def find_service
+    @service = accessible_services.find params[:service_id]
   end
 
-  def find_cinstance
-    @cinstance = accessible_not_bought_cinstances
-                   .includes(plan: %i[service original plan_metrics pricing_rules])
-                   .where(service: @service)
-                   .find(params[:id])
+  def accessible_plans
```
guicassolato, line 27: Where is this method being used?

guicassolato: I see it's used in `BaseApplicationsController`

app/controllers/buyers/applications_controller.rb
```diff
@@ -1,236 +1,27 @@
...
-
-    @cinstances = accessible_not_bought_cinstances
-      .scope_search(@search).order_by(params[:sort], params[:direction])
-      .preload(:service, user_account: [:admin_user], plan: [:pricing_rules])
-      .paginate(pagination_params)
-      .decorate
+    super
 
-    display_view_portion!(:service) if current_account.multiservice?
+    @search.account_id = @account.id
```
guicassolato, line 12: Shouldn't the `@search` be modified _before_ applying it to the scope?

app/controllers/api/applications_controller.rb
```diff
@@ -1,74 +1,36 @@
...
     @utilization = @cinstance.backend_object.utilization(@cinstance.service.metrics)
   end
 
-  def edit; end
+  def index
+    super
 
-  def new
-    @provider = current_account
+    @search.service_id = @service.id
```
guicassolato, line 18: Same question as https://github.com/3scale/porta/pull/2470/files?file-filters%5B%5D=.rb#r649792533

app/controllers/api/applications_controller.rb
```diff
@@ -1,74 +1,36 @@
...
-                   .where(service: @service)
-                   .find(params[:id])
+  def accessible_plans
+    super.where(issuer: @service)
   end
 
-  def authorize_partners
-    authorize! :manage, :partners
+  def accessible_not_bought_cinstances
+    find_service
```
guicassolato, line 32: I see this also called in a `before_action`, so I assume the repetition here has to do either with the order of the callback or with enabling it for a different action, which makes me wonder if the `before_action`s really belong to the parent class or would be better repeated in each controller.

Calling this method again here also causes an extra SQL query (because the implementation of the method in the base controller does not memoise). This could could be saved.

app/controllers/buyers/applications_controller.rb
```diff
@@ -1,236 +1,27 @@
...
-  end
-
-  module AccountForNewPlan
-    # dummy methods for formstastic
-    def service_plan_id; end
-  end
-
-  def extend_cinstance_for_new_plan
-    @cinstance.extend(AccountForNewPlan)
+    find_buyer
```
guicassolato, line 23: Exactly the same as https://github.com/3scale/porta/pull/2470/files?file-filters%5B%5D=.rb#r649798379

spec/javascripts/Logic/BuyerLogic.spec.js
```diff
@@ -0,0 +1,45 @@
...
+    const logic = new BuyerLogic({ ...buyer, contractedProducts: [] })
+
+    expect(logic.getContractedServicePlan(product)).toEqual(null)
+  })
+
+  it('should return null when product is not contracted ', () => {
+    const logic = new BuyerLogic({ ...buyer, contractedProducts: [{ id: 999, name: 'The Other product', withPlan: plan }] })
+
+    expect(logic.getContractedServicePlan(product)).toEqual(null)
+  })
```
josemigallas, line 39:
```suggestion

  })
```

app/controllers/api/applications_controller.rb
```diff
@@ -1,74 +1,46 @@
...
-                      .scope_search(@search).order_by(params[:sort], params[:direction])
-                      .preload(:service, user_account: [:admin_user], plan: [:pricing_rules])
-                      .paginate(pagination_params)
-                      .decorate
-  end
+  sublayout 'api/service'
 
   def show
-    @utilization = @cinstance.backend_object.utilization(@cinstance.service.metrics)
+    NoMethodError.new 'create: this action should be handled by ApplicationsController'
```
guicassolato, line 19: Let's remove the route instead, so it returns a 404.

app/controllers/api/applications_controller.rb
```diff
@@ -1,74 +1,46 @@
...
   end
 
-  def edit; end
+  def new; end
 
-  def new
-    @provider = current_account
+  def create
+    NoMethodError.new 'create: this action should be handled by ApplicationsController'
```
guicassolato, line 25: Same as https://github.com/3scale/porta/pull/2470/files?file-filters%5B%5D=.rb#r650989815.

app/controllers/buyers/applications_controller.rb
```diff
@@ -1,236 +1,31 @@
...
-
-  def destroy
-    if @cinstance.destroy
-      flash[:notice] = 'The application was successfully deleted.'
-      redirect_to admin_buyers_applications_path
-    else
-      flash[:notice] = 'Not possible to delete application'
-      redirect_to :back
-    end
+    NoMethodError.new 'create: this action should be handled by ApplicationsController'
```
guicassolato, line 22: Same as https://github.com/3scale/porta/pull/2470/files?file-filters%5B%5D=.rb#r650989815

app/controllers/api/applications_controller.rb
```diff
@@ -1,74 +1,51 @@
...
+  sublayout 'api/service'
 
-    @search.service_id = @service.id
+  def create
+    super
 
-    if params[:application_plan_id]
-      @plan = @service.application_plans.find params[:application_plan_id]
-      @search.plan_id = @plan.id
+    if @cinstance.save
```
guicassolato, line 24: This is a bit weird. It means the `#create` in `ApplicationsControllerMethods`does not really "create" the application, but just initializes it.

What about renaming that to something that communicates better that the application is being initialized/set or validated? `#create` could either invoke the method (similarly to `super` here), or it could be triggered with a `before_action ..., on: :create`

app/lib/applications_controller_methods.rb
```diff
@@ -0,0 +1,114 @@
...
+    @service_plan = @service.service_plans.find(service_plan_id)
+  end
+
+  def plan_id
+    @plan_id ||= params.require(:cinstance).permit(:plan_id).tap { |plan_params| plan_params.require(:plan_id) }[:plan_id]
+  end
+
+  def accessible_services
+    @accessible_services ||= current_user.accessible_services.includes(:application_plans)
+  end
```
guicassolato, line 88:
```suggestion
  end
  helper_method :accessible_services
```

And remove from the controllers.

app/controllers/api/applications_controller.rb
```diff
@@ -1,74 +1,51 @@
...
-    @utilization = @cinstance.backend_object.utilization(@cinstance.service.metrics)
   end
 
-  def edit; end
+  protected
 
-  def new
-    @provider = current_account
+  def define_search_scope
+    super({ service: @service.id })
```
guicassolato, line 36:
```suggestion
  def define_search_scope(opts = {}
    super(opts.reverse_merge(service: @service.id))
```

app/controllers/buyers/applications_controller.rb
```diff
@@ -1,236 +1,43 @@
...
-
-  module AccountForNewPlan
-    # dummy methods for formstastic
-    def service_plan_id; end
-  end
-
-  def extend_cinstance_for_new_plan
-    @cinstance.extend(AccountForNewPlan)
+  def define_search_scope
+    super({ account: @account.id })
```
guicassolato, line 40:
```suggestion
  def define_search_scope(opts = {})
    super opts.reverse_merge(account: @account.id)
```

app/lib/applications_controller_methods.rb
```diff
@@ -0,0 +1,114 @@
+# frozen_string_literal: true
+
+module ApplicationsControllerMethods
+  # Common methods should be index, new, create
+
```
guicassolato, line 5:
```suggestion
```

test/functional/provider/admin/applications_controller_test.rb
```diff
@@ -0,0 +1,74 @@
+# frozen_string_literal: true
```
guicassolato, line 1: We shouldn't be adding any new tests in `test/functional`. Please favour `test/integration` instead.

app/controllers/api/applications_controller.rb
```diff
@@ -1,74 +1,51 @@
...
-    @search.service_id = @service.id
+  def create
+    super
 
-    if params[:application_plan_id]
-      @plan = @service.application_plans.find params[:application_plan_id]
-      @search.plan_id = @plan.id
+    if @cinstance.save
+      flash[:notice] = 'Application was successfully created.'
+      redirect_to provider_admin_application_path(@cinstance)
```
guicassolato, line 26: I think it should redirect to the index page in the same scope (API product, in this case).

app/controllers/buyers/applications_controller.rb
```diff
@@ -1,236 +1,43 @@
...
-                   end
-
-    @cinstance = current_account.provider_builds_application_for(@account, application_plan, params[:cinstance], service_plan)
-    @cinstance.validate_human_edition!
+    super
 
     if @cinstance.save
       flash[:notice] = 'Application was successfully created.'
-      redirect_to(admin_service_application_path(@cinstance.service, @cinstance))
+      redirect_to provider_admin_application_path(@cinstance)
```
guicassolato, line 30: I think it should redirect to the index page in the same scope (buyer, in this case).

app/controllers/provider/admin/applications_controller.rb
```diff
@@ -0,0 +1,112 @@
...
+
+  helper_method :accessible_services
+
+  activate_menu :audience, :applications, :listing
+
+  layout 'provider'
+
+  def index
+    super
+  end
```
hallelujah, line 24: No need of this. It is redundant

guicassolato: I bet code climate complained about it.

test/integration/provider/admin/applications/applications_controller_test.rb
```diff
@@ -0,0 +1,188 @@
...
+      @service = master_account.default_service
+      FactoryBot.create(:cinstance, service: service)
+    end
+
+    test 'index retrieves all master\'s provided cinstances except those whose buyer is master' do
+      get provider_admin_applications_path
+
+      assert_response :ok
+      expected_cinstances_ids = master_account.provided_cinstances.not_bought_by(master_account).pluck(:id)
+      assert_same_elements expected_cinstances_ids, assigns(:cinstances).map(&:id)
```
hallelujah, line 20: `assigns` are deprecated, you probably want to do a functional test but it is also deprecated

So there are few approaches:

- test that you get the needed data in the view
- use service object approach, and test that the service object is called correctly (expect) in the controller

Both can be combined ;)

test/integration/provider/admin/applications/applications_controller_test.rb
```diff
@@ -0,0 +1,188 @@
...
+      end
+
+      attr_reader :service
+
+      test 'index shows the services column when the provider is multiservice' do
+        provider.services.create!(name: '2nd-service')
+        assert provider.reload.multiservice?
+        get provider_admin_applications_path
+        page = Nokogiri::HTML::Document.parse(response.body)
+        assert page.xpath("//tr").text.match /Service/
```
hallelujah, line 59: Maybe better to reference which table:

`page.xpath('//table#id/tr/td[text() = 'Service']')`

hallelujah (commented):
That is some :rocket: PR :) :clap: 

Some things to change but :tophat: for the work

hallelujah (changes requested):
Some of my comments are valid for other same type of cases. I was not going to comment on all them ;)

app/controllers/api/applications_controller.rb
josemigallas: Sounds good to me 👍

app/lib/applications_controller_methods.rb
```diff
@@ -0,0 +1,114 @@
...
+  def index
```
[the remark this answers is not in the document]
josemigallas: Yes, that makes sense to me. Same idea as @guicassolato regarding `#create`.

app/views/buyers/applications/index.html.slim
```diff
@@ -1,21 +1,16 @@
...
-  - if @plan
-    ' on plan
-    = @plan.name
-  - elsif @service
-    ' on service
-    = @service.name
-  - elsif @account
-    ' for
-    = @account.name
+h2 = "Applications for #{@account.name}"
```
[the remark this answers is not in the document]
josemigallas: Everything has been separated into the 3 contexts. Now instead of checking if @plan, if @service etc it renders either `provider/admin/applications/index`, `buyers/applications/index`, etc. Same for `/new`.

app/controllers/provider/admin/applications_controller.rb
josemigallas: If I define the method like this, codeclimate complaints about it.
If I remove the redundant method, then codeclimate complaints about it.

See conflicting warnings in the following image:
Screenshot 2021-06-16 at 17 46 11 (1)

I personally would like to make it explicit the controller implements `index`, even if it's redundant.

josemigallas: The other option is rename `index` and `create` in the module and invoke them via filters, no longer needing to call `super`.

app/controllers/api/applications_controller.rb
josemigallas: Index, not show page?

app/controllers/provider/admin/applications_controller.rb
```diff
@@ -0,0 +1,112 @@
...
+
+  def show
+    @utilization = @cinstance.backend_object.utilization(@cinstance.service.metrics)
+  end
+
+  def new
+    super
+  end
```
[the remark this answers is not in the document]
josemigallas: Let's reach an agreement in this thread https://github.com/3scale/porta/pull/2470#discussion_r660539060

app/controllers/provider/admin/applications_controller.rb
```diff
@@ -0,0 +1,112 @@
...
+    if @cinstance.save
+      flash[:notice] = 'Application was successfully created.'
+      redirect_to provider_admin_application_path(@cinstance)
+    else
+      @cinstance.extend(AccountForNewPlan)
+      render action: :new
+    end
+  end
+
+  def edit; end
```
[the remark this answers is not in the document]
josemigallas: I'd prefer making them explicit, but let's discuss in the other thread https://github.com/3scale/porta/pull/2470#discussion_r660539060

app/controllers/provider/admin/referrer_filters_controller.rb
```diff
@@ -42,6 +42,6 @@ def find_cinstance
   end
 
   def return_url
-    admin_service_application_path(@cinstance.service, @cinstance)
+    provider_admin_application_path(@cinstance)
```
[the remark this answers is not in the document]
josemigallas: There's only 1 show page and I moved it under `p/admin/applications` since it's the "main" context so to speak. We could add a show page for each context but IMO since a particular application does not depend on context, I would leave it like this.

app/lib/applications_controller_methods.rb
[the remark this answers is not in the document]
josemigallas: Refactored to `find_state` and `find_applications`.

app/lib/applications_controller_methods.rb
```diff
@@ -0,0 +1,114 @@
...
+
+  def define_search_scope(opts = {})
+    @search = ThreeScale::Search.new(params[:search] || params)
+    @search.account = params['account_id'] if params.key?('account_id')
+    @search.plan_id = params['application_plan_id'] if params.key?('application_plan_id')
+    @search.merge!(opts)
+  end
+
+  def change_state(action, message, *rest)
+    @cinstance.send("#{action}!", *rest)
```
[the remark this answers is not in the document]
josemigallas: I have no idea TBH

app/controllers/provider/admin/applications_controller.rb
hallelujah: You can waive the issue in codeclimate if that does not make any sense ;)

app/lib/applications_controller_methods.rb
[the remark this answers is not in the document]
hallelujah: I think that is fine

guicassolato (approved):
Tested on OpenShift. LGTM.
Nice work!

hallelujah (approved):
LGTM 

Note, next time do not add empty actions they are really not necessary. If codeclimate complains you can waive it on their site or disable the check in the config file or inline

PR #2471, opened 2021-06-02
[gem] Use 3scale/activejob-uniqueness

hallelujah:
For Ruby 2.4 support, as some older deployments still uses this EOL Ruby


josemigallas (commented):
LGTM

PR #2472, opened 2021-06-04
Sanitize service name

damianpm:

Sanitize user input before rendering

https://issues.redhat.com/browse/THREESCALE-6785

### DIFF
--- a/app/views/sites/emails/edit.html.erb
+++ b/app/views/sites/emails/edit.html.erb
@@ -14,7 +14,7 @@
 
       <% @services.each do |service| %>
         <%= form.semantic_fields_for service do |sf| %>
-          <%= sf.input :support_email, :label => "Service #{service.name}", :input_html => { :value => service.support_email, :id => "service_#{service.id}_support_email", :name => "service_#{service.id}_support_email" } %>
+          <%= sf.input :support_email, :label => "Service #{ sanitize service.name }", :input_html => { :value => service.support_email, :id => "service_#{service.id}_support_email", :name => "service_#{service.id}_support_email" } %>
        <% end
         end
       end %>
--- a/test/functional/sites/emails_controller_test.rb
+++ b/test/functional/sites/emails_controller_test.rb
@@ -1,0 +1,23 @@
+require 'test_helper'
+
+class Sites::EmailsControllerTest < ActionController::TestCase
+
+  def setup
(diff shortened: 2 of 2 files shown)


app/views/sites/emails/edit.html.erb
```diff
@@ -14,7 +14,7 @@
 
       <% @services.each do |service| %>
         <%= form.semantic_fields_for service do |sf| %>
-          <%= sf.input :support_email, :label => "Service #{service.name}", :input_html => { :value => service.support_email, :id => "service_#{service.id}_support_email", :name => "service_#{service.id}_support_email" } %>
+          <%= sf.input :support_email, :label => "Service #{ sanitize service.name }", :input_html => { :value => service.support_email, :id => "service_#{service.id}_support_email", :name => "service_#{service.id}_support_email" } %>
```
guicassolato, line 17: Planning to add a test?

damianpm: To be honest, I don't know what should I test here... that it renders text?

damianpm: I think we can leave the tests for the form builder refactor @hallelujah suggested

hallelujah (changes requested):
Agree with @guicassolato 

You can  look at https://github.com/3scale/porta/blob/02489fb39247ce827433f53ddd39c1aac880c5c2/test/integration/user-management-api/users_test.rb#L397

test/functional/sites/emails_controller_test.rb
```diff
@@ -0,0 +1,23 @@
...
+    @provider  = FactoryBot.create(:provider_account)
+    @service = @provider.default_service
+    request.host = @provider.admin_domain
+    login_as(@provider.admins.first)
+  end
+
+  test 'show Service name sanitized' do
+    SERVICE_NAME_XSS = "API<script>alert('XSS')</script>"
+    SERVICE_NAME_XSS_SANITIZED = "APIalert('XSS')"
+    @service.name = SERVICE_NAME_XSS
```
hallelujah, line 15: Move those constants outside of method body. Or create them as simple variable:
```suggestion
    service_name_xss = "API<script>alert('XSS')</script>"
        service_name_xss_sanitized = "APIalert('XSS')"
    @service.name =     service_name_xss
```

hallelujah (approved):

PR #2475, opened 2021-06-04
Fix nightly build

hallelujah:
Removed the workflow as the build will happen in quay.io directly

After discussion with @eguzki  We decide to do latest and name it nightly (for backward compatibility)

:warning: `nightly` tag is deprecated and should not be used anymore

@3scale/support @3scale/documentation @3scale/qe  If any documentation mentions nightly build please correct it

We are for now going to keep `nightly` tag for backward compatibility but it will reference the `latest` build
The `nightly` tag might disappear without notice after 3 months


akostadinov (dismissed):
Looks good. Dockerfile expects to run in the root context based on the `ADD` statements.

eguzki:
LGTM

.github/workflows/nightly-build.yml
```diff
@@ -1,9 +1,10 @@
-name: Nightly build
+name: Build and push to quay.io the latest code
```
akostadinov, line 1: Provided that it is no longer a nightly but per-push, shall we also rename the file `nightly-build.yml`?

hallelujah: Well I will keep it like that because it still push nightly tag :)
I changed it in zync

akostadinov: Although I recognize that many projects also call nightly builds, builds that are performed more often that once a day. So this is just a question to consider. It's fine with me to keep as is.

.github/workflows/nightly-build.yml
```diff
@@ -23,6 +24,9 @@ jobs:
         name: Build and push
         uses: docker/build-push-action@v2
         with:
-          context: openshift/system
+          context: .
+          file: openshift/system/Dockerfile
           push: true
-          tags: quay.io/3scale/porta:nightly
+          tags: |
+            quay.io/3scale/porta:latest
```
akostadinov, line 31: Again just to consider, could this confuse anybody trying to pull latest stable build but get latest CI build?

hallelujah: What do you mean? `latest` is the build of the default branch

.github/workflows/quay-io-build-latest.yml
```diff
@@ -23,6 +24,9 @@ jobs:
         name: Build and push
         uses: docker/build-push-action@v2
         with:
-          context: openshift/system
+          context: .
+          file: openshift/system/Dockerfile
           push: true
-          tags: quay.io/3scale/porta:nightly
+          tags: |
+            quay.io/3scale/porta:latest
```
akostadinov, line 31: Again just to consider, could this confuse anybody trying to pull latest stable build but get latest CI build instead?

.github/workflows/quay-io-build-latest.yml
```diff
@@ -23,6 +24,9 @@ jobs:
...
+            quay.io/3scale/porta:nightly
```
akostadinov, line 32: Again just to consider, could this confuse anybody trying to pull latest stable build but get latest CI build instead?

.github/workflows/quay-io-build-latest.yml
akostadinov, line 31: Again just to consider, could this confuse anybody trying to pull latest stable build but get latest CI build instead?

.github/workflows/nightly-build.yml
akostadinov: In priciple - yes. As a user though, I usually expect when I do `podman pull fedora` to bring me latest stable fedora release, not a nightly.

gsaslis (commented):
I'm curious about why we chose github actions vs. circleci for this job. 

Is there some policy with regards to which jobs will land in each of the two CI systems?

hallelujah:
@gsaslis The choice is just by me... But I had some reasons:

- I wanted to use our credits of GH actions
- I do not want to add a 15' job for each merge in CircleCI. Especially to reduce the cost and also to free some parallel job there . Those can be used on other PR work
- It was straight-forward to set it up.

PR #2476, opened 2021-06-04
Latest container build should use quay.io registry

hallelujah:
Otherwise we will be limited by docker rate limits

```
Could not pull base image: API error (500): toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
```

### DIFF
--- a/openshift/system/Dockerfile
+++ b/openshift/system/Dockerfile
@@ -1,4 +1,4 @@
-FROM centos/ruby-25-centos7
+FROM quay.io/centos7/ruby-25-centos7
 
 USER root
 


openshift/system/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM centos/ruby-25-centos7
+FROM quay.io/centos/ruby-25-centos7
```
gsaslis, line 1:
```suggestion
FROM quay.io/centos7/ruby-25-centos7
```

gsaslis (changes requested):
i think there's a `7` missing there

openshift/system/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM centos/ruby-25-centos7
+FROM quay.io/centos7/ruby-25-centos7
```
gsaslis, line 1:
```suggestion
FROM registry.access.redhat.com/ubi7/ruby-27
```

gsaslis (changes requested):
Better to switch to the official ruby 2.7 image as it will also get security updates and comes from a trusty source.

openshift/system/Dockerfile
hallelujah: Then we will need to modify a bit the Dockerfile but yes

akostadinov: Another option would be to start with clean UBI image and install `rh-ruby25` in build or is the package gone already?

akostadinov: But not sure how UBI on RHEL7 works now. I think the following repos need to be enabled.

- rhel-7-server-rpms
- rhel-7-server-optional-rpms
- rhel-7-server-extras-rpms
- rhel-server-rhscl-7-rpms

Or maybe only the last one?

hallelujah: We can't use UBI images or we should use 3rd party packages and then there is no point on using the UBI image anymore compared to centos

gsaslis (approved):
Let's Get This Merged !

PR #2477, opened 2021-06-07
[circleci] Try Mysql 8 latest in circleci

hallelujah:
Closes https://issues.redhat.com/browse/THREESCALE-6825

- [The caching_sha2_password is incompatible with RHEL 7](https://access.redhat.com/documentation/en-us/red_hat_software_collections/3/html/3.2_release_notes/chap-migration#sect-Migration-MySQL-Changes). We need to set `--default-authentication-plugin=mysql_native_password` on the server

> The MySQL 8.0 server provided by the rh-mysql80 Software Collection is configured to use mysql_native_password as the default authentication plug-in because client tools and libraries in Red Hat Enterprise Linux 7 are incompatible with the caching_sha2_password method, which is used by default in the upstream MySQL 8.0 version.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -89,7 +89,7 @@ system-builder-ruby25: &system-builder-ruby25
     RAILS_ENV: test
 
 mysql-container: &mysql-container
-  image: circleci/mysql:5.7-ram
+  image: circleci/mysql:8-ram
   environment:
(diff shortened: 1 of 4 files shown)


akostadinov (approved):

hallelujah:
Low prio

PR #2478, opened 2021-06-08
Fixed the title mismatch on the product page

srijitamukherjee:
Signed-off-by: Srijita Mukherjee <srmukher@redhat.com>

> This PR updates the page title from Services to Products for the Products page.

> https://issues.redhat.com/browse/THREESCALE-7123

**Verification steps** 

> On visiting the Products page, the html title shows Products instead of Services.

> IMO, there are 2 ways to solve this. 

> 1. Just update the services to products in the page title. This is what is done in this PR.
> 2. To convert all service controllers to product, which might cause more regressions.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -40,7 +40,7 @@ module ApplicationHelper
                     scope: [:controller, :action, name],
                     default: action.humanize)
 
-    "#{controller_name} - #{action_name}"
+    "#{controller_name == "Services" ? "Products" : controller_name} - #{action_name}"
   end
 
   def link_to_unless_current_styled(text, url, options = {})


app/helpers/application_helper.rb
```diff
@@ -40,7 +40,7 @@ def default_title
                     scope: [:controller, :action, name],
                     default: action.humanize)
 
-    "#{controller_name} - #{action_name}"
+    "#{controller_name == "Services" ? "Products" : controller_name} - #{action_name}"
```
hallelujah, line 43: Can you create a test for that?
I think this will not work because of the capitalized words

hallelujah (changes requested):

PR #2480, opened 2021-06-09
Fields definition API

jsmadis:

This PR adds fields definition to the admin API.

https://issues.redhat.com/browse/THREESCALE-4082

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -658,7 +658,7 @@ without fake Core server your after commit callbacks will crash and you might ge
 
       resources :features, :controller => 'account_features', :except => [:new, :edit]
 
-      resources :fields_definitions, :controller => 'fields_definitions', :except => [:new, :edit]
+      resources :fields_definitions, controller: 'fields_definitions', except: %i[new edit]
 
       resource :nginx, :only => [:show], :defaults => { :format => 'zip' } do
         collection do
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -658,6 +658,8 @@ without fake Core server your after commit callbacks will crash and you might ge
 
       resources :features, :controller => 'account_features', :except => [:new, :edit]
 
+      resources :fields_definitions, :controller => 'fields_definitions', :except => [:new, :edit]
+
       resource :nginx, :only => [:show], :defaults => { :format => 'zip' } do
         collection do
           get :spec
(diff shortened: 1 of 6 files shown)


jsmadis:
FYI this PR exposes an old bug in the fields definition (https://issues.redhat.com/browse/THREESCALE-7189). The bug will be fixed in the separate PR

config/routes.rb
```diff
@@ -646,6 +646,8 @@ def call(env)
 
       resources :features, :controller => 'account_features', :except => [:new, :edit]
 
+      resources :fields_definitions, :controller => 'fields_definitions', :except => [:new, :edit]
```
josemigallas, line 649:
```suggestion
      resources :fields_definitions, controller: 'fields_definitions', except: %i[new edit]
```

config/webpacker.yml
```diff
@@ -33,6 +33,7 @@ development:
   <<: *default
   compile: true
   webpack_compile_output: true
+  check_yarn_integrity: false
```
josemigallas, line 36: 🚨 
```suggestion
```

spec/acceptance/api/fields_definition_spec.rb
```diff
@@ -0,0 +1,40 @@
+require 'rails_helper'
```
josemigallas, line 1:
```suggestion
#frozen_string_literal: true

require 'rails_helper'
```

spec/acceptance/api/fields_definition_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+
+resource 'FieldsDefinition' do
+  let(:resource) { FactoryBot.create(:fields_definition,
+                                     account: provider,
+                                     target: 'Account',
+                                     name: 'stuff',
+                                     choices: %w[Orange Apple Banana]) }
```
josemigallas, line 8: Unimportant, but I think this way is more consistent with the ~rest~ majority of the codebase:
```suggestion
  let(:resource) { FactoryBot.create(:fields_definition, account: provider,
                                                         target: 'Account',
                                                         name: 'stuff',
                                                         choices: %w[Orange Apple Banana]) }
```

josemigallas (changes requested):
The requested change is because of the check integrity settings in the webpack config

josemigallas (commented):
Looking at the spec file, should we also add cases for `show` and `destroy`?

hallelujah (approved):
:rocket:

jsmadis:
I will add some tests, @josemigallas  is right :)

jsmadis:
Rebased with the actual master because it contains the fix for the failing cucumber tests

josemigallas (approved):

PR #2481, opened 2021-06-10
Fix New Application Form 🐛 

josemigallas:

The logic around contracted service plans is broken (thank you Flow).

The logic has been extracted to a different file and tested.

- [x] Service plan dropdown is always enabled, but it should be fixed when the selected product is subscribed to a service plan.

**Verification steps** 

* Product with no service plan: can't fill form, need to create a plan first.
* Product with no subscription: need to select a plan in Service Plan dropdown
* Product with subscription: dropdown is disabled and the plan selected, user can't change it

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -17,4 +17,4 @@ experimental.const_params=true
 max_header_tokens=1
 module.name_mapper='.*\(.s?css\)' -> 'empty/object'
 module.system=haste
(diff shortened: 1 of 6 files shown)


thomasmaas (approved):

thomasmaas:
Question that might be out of scope: when service plans are 'hidden' in the settings, do we also hide the select?

damianpm (approved):

josemigallas:
> Question that might be out of scope: when service plans are 'hidden' in the settings, do we also hide the select?

@hallelujah @guicassolato @cntlsn TBH we might have talked about this already but I can't remember. The logic surrounding service plans and subscriptions is so confusing... according to the previous implementation, I don't think the select was hidden, only disabled, but I might be wrong.

cntlsn:
@thomasmaas @josemigallas please find and discuss the UX requirements over here: https://issues.redhat.com/browse/THREESCALE-6878

PR #2484, opened 2021-06-14
master to 2.11 candidate

thomasmaas:


hallelujah (changes requested):
I am not against merge PR but I think this is more political than anything with code changes.
Basically we want master and the candidate heads, at some point in time, to have the same sha1.
I wonder why can't we make a fast forward on the candidate to master?

Because the PR is too huge to be reviewed. I have my threshold on how much my brain can review a PR.

I would propose to just  fast-forward candidate's HEAD to the default branch sha1 pointed at that time. Maybe tagging for the sake of history if we want.

What do you think @3scale/system @3scale/qe @3scale/product 

I really do not see the point of having this PR done, I really mean it that nobody is going to pay many attention to it in term of code diff.

josemigallas:
@hallelujah 
Even if this is a big PR, supposedly we have already reviewed it in smaller batches when we merged them into master. However, I don't think it is really that huge, most of the changes seems to be coming from a lock file and we could even set  up a session to do a group-reviewing, with those of us who have contributed the most (you and me).

mkudlej:
Has anybody asked why this PR is so huge?
According our new workflow:
1) if any PR is done, it can be merged into Master branch
2) if any Jira(or group of Jiras in case of big change) is developed, it can be merged into Candidate branch
3) if any Jira(or group of Jiras in case of big change) is tested, it can be merged into stable branch

So I am curious why commits related to Jiras were not merged in these three phases?
@secretsurf @thomasmaas Does the Porta team use the new workflow for taking care of PRs/Jiras?

thomasmaas:
@mkudlej agreed, so far we've been merging into candidate on a biweekly basis but it would be better if individual contributors merge into candidate from master (or forward) when they merge a pr to master.

thomasmaas:
merged and pushed through console which is what i've been doing so far anyway. I don't see a need to look at commits again. prs should be reviewed before they merged into master, this is just administrative.

PR #2485, opened 2021-06-14
letter_opener made easy

akostadinov:

Makes easy to enable `letter_opener` by a simple environment variable at runtime.

```
$ LETTER_OPENER=1 bundle exec sidekiq ...
```

No issue.

**Verification steps** 

- export LETTER_OPENER=1
- bundle exec sidekiq ...
- in Porta web UI invire a user
- a browser tab should open with email content present

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -57,8 +57,15 @@ Rails.application.configure do
   config.action_dispatch.best_standards_support = :builtin
 
   config.action_mailer.perform_deliveries = true
-  # See emails in browser
-  config.action_mailer.delivery_method = defined?(LetterOpener) ? :letter_opener : :test
+
+  config.action_mailer.delivery_method =
+    if defined?(LetterOpener)
+      :letter_opener
+    elsif config.action_mailer.smtp_settings[:address]
+      :smtp
+    else
+      :test
+    end
 
   # Debug mode disables concatenation and preprocessing of assets.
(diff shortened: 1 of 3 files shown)


config/environments/development.rb
```diff
@@ -57,8 +57,14 @@
   config.action_dispatch.best_standards_support = :builtin
 
   config.action_mailer.perform_deliveries = true
-  # See emails in browser
-  config.action_mailer.delivery_method = defined?(LetterOpener) ? :letter_opener : :test
+
+  if defined?(LetterOpener)
+    config.action_mailer.delivery_method = :letter_opener
+  elsif config.action_mailer.smtp_settings[:address]
+    config.action_mailer.delivery_method = :smtp
```
akostadinov, line 64: This is to allow using actual SMTP delivery when SMTP configuration is present.

hallelujah:
Nice PR, can you fix Code Climate? also and trigger the `javascript_tests` and `mysql_build` checks?

hallelujah (changes requested):
Remark, in your comment you mentionned sidekiq but some emails are sent in the request so technically `rails server` is also impacted

akostadinov:
hmm, I was thinking it's gonna complain if I use `var = case ...` but it got the other way around. I should have checked locally.

hallelujah (approved):

PR #2490, opened 2021-06-16
THREESCALE-7134 async_deploy methods from Proxy

akostadinov:

THREESCALE-7134

### DIFF
--- a/app/models/proxy.rb
+++ b/app/models/proxy.rb
@@ -327,18 +327,6 @@ class Proxy < ApplicationRecord
     end
   end
 
-  def save_and_async_deploy(attrs, user)
-    saved = update_attributes(attrs)
-
-    analytics.track('Sandbox Proxy updated', analytics_attributes.merge(success: saved))
-
-    saved && async_deploy(user)
-  end
-
-  def async_deploy(user)
-    ApicastV1DeploymentService.async_deploy(user, self)
-  end
-
   def hosts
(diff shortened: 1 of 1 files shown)


app/models/proxy.rb
```diff
@@ -327,18 +327,6 @@ def track_apicast_version_change
     end
   end
 
-  def save_and_async_deploy(attrs, user)
```
hallelujah, line 330: Isn't this used somewhere else?

hallelujah (changes requested):
I think you are removing a method but not where it is used

app/models/proxy.rb
akostadinov:
```
🐚 rg save_and_async_deploy
```
shows nothing

srijitamukherjee: It was used here: https://github.com/3scale/porta/pull/2487/files#diff-8aa693f2978224422216c485c5a77372dd282ca5599f558c185744473e8a10d7L130 
which is already removed. 

Thus this PR also looks good.

srijitamukherjee (approved):

app/models/proxy.rb
akostadinov: I think it is more of a problem to remove usages and leave implementation because it may not show up in tests. While keeping calls but remove implementation is more likely to break :)

hallelujah (approved):

PR #2492, opened 2021-06-16
THREESCALE-7134 Removing Apicast V1 class

HemaHG:

> [THREESCALE-7134](https://issues.redhat.com/browse/THREESCALE-7134) - Removing Apicast V1 class


test/unit/service_test.rb
```diff
@@ -569,8 +569,7 @@ class AsynchronousDeletionOfService < ActiveSupport::TestCase
     service = Service.new(account: account = Account.new)
     service.proxy = Proxy.new(service: service)
 
-    account.stubs(:provider_can_use?).with(:apicast_v1).returns(true) # irrelevant for this test
-    account.stubs(:provider_can_use?).with(:apicast_v2).returns(true) # irrelevant for this test
+    account.stubs(:provider_can_use?).with(:apicast_v2).returns(true)
```
jsmadis, line 572: We should keep the comment in the end of the line 
```suggestion
    account.stubs(:provider_can_use?).with(:apicast_v2).returns(true) # irrelevant for this test
```

HemaHG:
The failing testing is fixed in the PR https://github.com/3scale/porta/pull/2495

test/integration/user-management-api/services/proxies_test.rb
```diff
@@ -63,9 +63,7 @@ def test_update_for_service_mesh
     params = provider_key_params.merge(proxy: { credentials_location: 'headers' })
     @service.update_column :deployment_option, 'service_mesh_istio'
 
-    assert_difference @service.proxy.proxy_configs.production.method(:count), +1 do
-      put(admin_api_service_proxy_path(params))
-    end
+    put(admin_api_service_proxy_path(params))
```
akostadinov, line 66: Could you explain this change? Seems like the assertion is gone, but I can't understand what was original intention and why it can't work anymore.

test/unit/apicast/sandbox_proxy_test.rb
```diff
@@ -8,6 +8,6 @@ def test_proxy_host
     service.proxy.apicast_configuration_driven = false
     service.proxy.sandbox_endpoint = nil
 
-    assert_equal "#{service.system_name}-#{service.account_id}.staging.apicast.io", proxy.proxy_host
+    assert_equal "#{service.system_name}-#{service.account_id}.staging.apicast.dev", proxy.proxy_host
```
akostadinov, line 11: Also I can't understand why this change is needed.

test/unit/proxy_test.rb
```diff
@@ -192,17 +192,10 @@ def test_deployable
...
 
   def test_apicast_configuration_driven
-    @proxy.provider.stubs(:provider_can_use?).with(:apicast_v1).returns(true)
     @proxy.provider.stubs(:provider_can_use?).with(:apicast_v2).returns(true)
 
     @proxy.apicast_configuration_driven = true
     assert @proxy.apicast_configuration_driven
-
-    @proxy.apicast_configuration_driven = false
-    refute @proxy.apicast_configuration_driven
```
akostadinov, line 202: Do we remove the whole field? If not, why this test should be removed?

test/unit/proxy_test.rb
```diff
@@ -454,7 +446,7 @@ def test_apicast_configuration_driven
   end
 
   test 'sandbox_endpoint set on creation' do
-    assert_match %r{^https://}, @proxy.sandbox_endpoint
+    assert_match %r{^http://}, @proxy.sandbox_endpoint
```
akostadinov, line 449: Why has this changed?

akostadinov (changes requested):
@hallelujah requested for my removal of a rolling update, to first deprecate the rolling update and only later remove the class. I think this request will apply here as well. See this for details:

https://github.com/3scale/porta/pull/2496/files#diff-11abcaa744f5e2710aee0d37a5cce03dda18ef2d0a91312f6f5bb5843cae50f8R68

test/unit/apicast/sandbox_proxy_test.rb
HemaHG: with the removal of apicastv1 class ".io" endpoint doesn't exist

test/unit/proxy_test.rb
HemaHG: apicast_configuration_driven method is only checking if the provider can use apicast_v2 so I don't think this test is need

app/models/proxy.rb
```diff
@@ -406,7 +406,7 @@ def set_correct_endpoints
   end
 
   def apicast_configuration_driven
-    if provider && provider.provider_can_use?(:apicast_v2) && !provider.provider_can_use?(:apicast_v1)
+    if provider && provider.provider_can_use?(:apicast_v2)
```
hallelujah, line 409: Do you think we should keep this condition?

test/unit/proxy_test.rb
hallelujah: @HemaHG So do you think we should keep this test then? and Why?

hallelujah (changes requested):

test/integration/api/integrations_controller_test.rb
```diff
@@ -158,7 +158,6 @@ def test_update_proxy_rule_position
     rolling_updates_on
     Account.any_instance.stubs(:provider_can_use?).with(:proxy_pro).returns(true)
 
-    Proxy.any_instance.expects(:save_and_deploy).never
```
akostadinov, line 161: Was this method removed? Perhaps leave this clause until it is removed.

app/lib/logic/rolling_updates.rb
```diff
@@ -250,16 +250,6 @@ def missing_config
         end
       end
 
-      class ApicastV1 < Base
```
akostadinov, line 253: As I said earlier, @hallelujah suggested requested to deprecate rolling updates before removing them. See 
https://github.com/3scale/porta/pull/2496/files#diff-11abcaa744f5e2710aee0d37a5cce03dda18ef2d0a91312f6f5bb5843cae50f8R281

test/unit/apicast/sandbox_proxy_test.rb
akostadinov: Would you tell me how that works? I don't understand why it stopped existing.

akostadinov (dismissed):
Looks good to me.

app/lib/logic/rolling_updates.rb
HemaHG: As @hery said this class is no more used, removing the class itself

akostadinov (approved):

hallelujah:
This will be done at the very end right?

akostadinov:
Why on the very end? I though nothing checks this rolling update anymore.

hallelujah:
If so then lets remove it. Good job everybody!

hallelujah (approved):
🚀 :shipit:

PR #2493, opened 2021-06-16
Fix provider host passed in the Zync update tenant request

guicassolato:
Otherwise it will break when domain substitution is enabled.


hallelujah (approved):
Thank you very much! :rocket:

PR #2494, opened 2021-06-17
🐛 Hide Service Plan select when they're switched off

josemigallas:

Right now, Service plan select is always visible because  of `servicePlansAllwed` not being calculated correctly.

[THREESCALE-7120: Create application page has Service plan field despite service plan option is turned off](https://issues.redhat.com/browse/THREESCALE-7120)

**Verification steps** 

See JIRA

### DIFF
--- a/app/decorators/provider_decorator.rb
+++ b/app/decorators/provider_decorator.rb
@@ -2,15 +2,19 @@
 
 class ProviderDecorator < ApplicationDecorator
   include System::UrlHelpers.system_url_helpers
+  include PlansHelper
+
+  delegate :can?, :settings, to: :h
 
   self.include_root_in_json = false
 
   def new_application_form_data(buyer: nil, service: nil, cinstance: nil)
+    # TODO: Reduce data by not including service_plans when service_plans_management_visible? is false
     data = {
       'create-application-plan-path': new_admin_service_application_plan_path(':id'),
       'create-service-plan-path': new_admin_service_service_plan_path(':id'),
       'service-subscriptions-path': admin_buyers_account_service_contracts_path(':id'),
-      'service-plans-allowed': settings.service_plans.allowed?.to_json,
+      'service-plans-allowed': service_plans_management_visible?.to_json,
       'defined-fields': application_defined_fields_data.to_json
     }
     data[:errors] = cinstance.errors.to_json if cinstance
--- a/features/old/applications/providers/create_application.feature
+++ b/features/old/applications/providers/create_application.feature
@@ -83,7 +83,6 @@
 
    Scenario: Create a new application without having a service subscription
      Given a service "second" of provider "foo.3scale.localhost"
-       And provider has "service_plans_ui_visible" hidden
        And a published service plan "non_default" of service "second" of provider "foo.3scale.localhost"
        And a published application plan "second_app_plan" of service "second" of provider "foo.3scale.localhost"
        And buyer "bob" is not subscribed to the default service of provider "foo.3scale.localhost"
@@ -168,7 +167,6 @@
       And I press "Update Application"
       And should see "can't be blank"
 
-
    Scenario: Edit an application
      Given buyer "bob" has application "UltraWidget" with description "Slightly less awesome widget"
      And I go to the buyer account page for "bob"
@@ -180,7 +178,6 @@
      Then I should see "Application was successfully updated"
      And I should be on the provider side "Not So Ultra Widget" application page
      And I should see "Not So Ultra Widget" in a header
-
 
    # TODO: also in single app mode
    Scenario: Edit an application with extra fields
--- a/test/decorators/provider_decorator_test.rb
+++ b/test/decorators/provider_decorator_test.rb
@@ -1,0 +1,41 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class ProviderDecoratorTest < Draper::TestCase
+
+  def setup
+    provider = FactoryBot.create(:provider_account)
+    @decorator = ProviderDecorator.new provider
+    @decorator.stubs(:service_plans_management_visible?).returns(true)
+  end
+
+  test 'new_application_form_data in Dashboard context' do
+    @decorator.expects(:audience_context_new_application_form_data).returns({}).once
+
+    assert @decorator.new_application_form_data
+  end
+
+  test 'new_application_form_data in Account context' do
+    buyer = FactoryBot.create(:buyer_account)
+    @decorator.expects(:buyer_context_new_application_form_data).with(buyer).returns({}).once
+
+    assert @decorator.new_application_form_data(buyer: buyer)
+  end
+
+  test 'new_application_form_data in Product context' do
+    service = FactoryBot.create(:service)
+    @decorator.expects(:product_context_new_application_form_data).with(service).returns({}).once
+
+    assert @decorator.new_application_form_data(service: service)
+  end
+
+  test 'new_application_form_data inline errors' do
+    cinstance = FactoryBot.create(:cinstance)
+    errors = ['Name is required']
+    cinstance.stubs(:errors).returns(errors)
+    @decorator.expects(:audience_context_new_application_form_data).returns({}).once
+
+    assert_equal errors.to_json, @decorator.new_application_form_data(cinstance: cinstance)[:errors]
+  end
+end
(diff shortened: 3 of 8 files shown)


app/lib/logic/contracting.rb
```diff
@@ -45,6 +45,7 @@ def provider_builds_application_for(buyer, application_plan, application_attrs =
         application.validate_human_edition!
 
         application.plan_id = application_plan.id
+        application_attrs.delete 'plan_id' # TODO: is there a better way to do this?
         application.attributes = application_attrs
```
akostadinov, line 49: if this is a hash, you can do `application.attributes = application_attrs.except('plan_id')`

app/decorators/provider_decorator.rb
```diff
@@ -69,4 +69,10 @@ def application_defined_fields_data
                         FieldsDefinitionDecorator.new(field).new_application_data(self)
                       end
   end
+
+  protected
+
+  def service_plans_visible
+    h.can?(:manage, :service_contracts) && settings.service_plans.allowed? && settings.service_plans_ui_visible?
```
hallelujah, line 76: Maybe refactor that and this one https://github.com/3scale/porta/blob/master/app/helpers/vertical_nav_helper.rb#L94
and maybe other places into a helper (though it does not belong to a helper per se)

It will make it easier to refactor later

features/old/applications/providers/create_application.feature
```diff
@@ -134,6 +133,58 @@
...
+    And I select "API" from "Product"
+    But the application plans select should not contain custom plans of provider "foo.3scale.localhost"
+    And I select "Basic" from "Application plan"
+    And I fill in "Name" with "UltimateWidget"
+    And I fill in "Description" with "Awesome ultimate super widget"
+    And I press "Create Application"
+    Then I should be on the provider side "UltimateWidget" application page
+    And should see "Application was successfully created"
+    And buyer "bob" should have 1 cinstance
+
```
hallelujah, line 187: I'd say just do not write any more cucumber like this. Remove that and redo it in another PR

features/old/applications/providers/create_application.feature
```diff
@@ -83,7 +83,6 @@
 
    Scenario: Create a new application without having a service subscription
      Given a service "second" of provider "foo.3scale.localhost"
-       And provider has "service_plans_ui_visible" hidden
```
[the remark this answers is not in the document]
josemigallas: I don't know why it was here in the first place, the scenario doesn't test service_plans visibility, only that there is no subscription. Moreover, See at line 96 the use is supposed to pick a service plan:
```
And I select "non_default" from "Service plan"
```
But if _provider has "service_plans_ui_visible" hidden_ the select will be hidden. I think this was another false-positive case.

features/old/applications/providers/create_application.feature
josemigallas: It's a terrible test but it's still a test and the bug must have one. I'd rather leave it here and refactor the whole file later.

I can offer adding an integration test instead of this cucumber, but that would take more time. I'd really keep these scenarios **TEMPORARILY** and redo the whole file in another PR.

app/lib/logic/contracting.rb
```diff
@@ -44,8 +44,7 @@ def provider_builds_application_for(buyer, application_plan, application_attrs =
         application.validate_contract_hierarchy!
         application.validate_human_edition!
 
-        application.plan_id = application_plan.id
-        application.attributes = application_attrs
+        application.attributes = application_attrs.merge('plan_id' => application.plan_id)
```
guicassolato, line 47:
```suggestion
        # `#merge` works here because `application_attrs` is a `ActionController::Parameters`; with a normal `Hash` this will break due to old mass-assignement protection
        application.attributes = application_attrs.merge('plan_id' => application.plan_id)
```

guicassolato: Another option:

```ruby
  attrs = application_attrs.dup
  application.plan_id = attrs.delete(:plan_id)
  application.attributes = attrs
```

test/unit/helpers/buyers/applications_helper_test.rb
```diff
@@ -15,6 +15,9 @@ def setup
 
     @fields = { name: 'Field' }
     ProviderDecorator.any_instance.stubs(:application_defined_fields_data).returns([@fields])
+
+    # TODO: Improve component by removing service_plans data when service_plans_visible is false
+    ProviderDecorator.any_instance.stubs(:service_plans_visible).returns(true)
```
guicassolato, line 20: `any_instance` is a bit dangerous, but I guess we're already drinking from this poison. Ideally, we would control what the whole decoration returns, replacing it with an instance on which these methods could be stubbed.

app/lib/logic/contracting.rb
guicassolato: In the future, let's try to move this to Rails strong parameters and avoid the headache.

guicassolato (dismissed):
Some minor improvements could be useful, but given the urgency I'd say it's okay for now.

test/unit/helpers/buyers/applications_helper_test.rb
josemigallas: Agreed 100%. I'm changing this helper class in https://github.com/3scale/porta/pull/2459 so I'll improve the tests there.

app/lib/logic/contracting.rb
```diff
@@ -45,7 +45,7 @@ def provider_builds_application_for(buyer, application_plan, application_attrs =
         application.validate_human_edition!
 
         application.plan_id = application_plan.id
-        application.attributes = application_attrs
+        application.attributes = application_attrs.except('plan_id')
```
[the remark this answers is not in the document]
josemigallas: I reverted this in the end to the original implementation since it's no longer throwing an error. Don't ask me why.

test/unit/helpers/buyers/applications_helper_test.rb
josemigallas: In the end I moved the tests to ProviderDecoratorTests and make them more concise. If they need to be improved we can still do it in the aforementioned PR.

features/old/applications/providers/create_application.feature
hallelujah: The issue is that we never redo them Remember your last PRs ... :)

test/decorators/provider_decorator_test.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def setup
+    provider = FactoryBot.create(:provider_account)
+    @decorator = ProviderDecorator.new provider
+    @decorator.stubs(:service_plans_management_visible?).returns(true)
+  end
+
+  test 'new_application_form_data in Dashboard context' do
+    data = {}
+    @decorator.stubs(:audience_context_new_application_form_data).returns(data).once
```
hallelujah, line 15: Instead of `stubs ... once` maybe `expects ... once`
Stubbing ≠ mocking

josemigallas: > Also assignment of `data` is not needed here
It is in order to avoid repetition

hallelujah (approved):

test/decorators/provider_decorator_test.rb
hallelujah: Yes well you repeat it in other tests always :) but here is not necessary, it is used only once. See below tests they are the same

test/decorators/provider_decorator_test.rb
```diff
@@ -0,0 +1,44 @@
...
+    @decorator.expects(:product_context_new_application_form_data).with(service).returns(data).once
+
+    assert @decorator.new_application_form_data(service: service)
+  end
+
+  test 'new_application_form_data inline errors' do
+    cinstance = FactoryBot.create(:cinstance)
+    errors = ['Name is required']
+    cinstance.stubs(:errors).returns(errors)
+    @decorator.stubs(:audience_context_new_application_form_data).returns({}).once
```
hallelujah, line 40: stubs once does not really do anything. Stub is always and should not expect any other modifier and does not expect/raise anything. `expects` raises if the expectations are not met

https://github.com/freerange/mocha#quick-start

test/decorators/provider_decorator_test.rb
josemigallas: Ah, you're right. There used to be repetition here but I ended up simplifying it.

test/decorators/provider_decorator_test.rb
josemigallas: Oh I missed this one, it's fixed now

hallelujah (approved):

PR #2495, opened 2021-06-17
THREESCALE-7134  Removing v1_compatible

HemaHG:

>PART of   [THREESCALE-7134](https://issues.redhat.com/browse/THREESCALE-7134) Removing v1_compatible

### DIFF
--- a/app/forms/onboarding/request_form.rb
+++ b/app/forms/onboarding/request_form.rb
@@ -49,7 +49,7 @@ class Onboarding::RequestForm < Reform::Form
   end
 
   def test_api!
-    return unless ProxyDeploymentService.call(model, v1_compatible: true)
+    return unless ProxyDeploymentService.call(model)
 
     proxy_test_service = ProxyTestService.new(model)
 
--- a/app/models/proxy.rb
+++ b/app/models/proxy.rb
@@ -353,7 +353,7 @@ class Proxy < ApplicationRecord
 
     return false unless saved
 
-    success = ProxyDeploymentService.call(self, v1_compatible: true)
+    success = ProxyDeploymentService.call(self)
     analytics.track('Sandbox Proxy Deploy', success: success)
     success
   end
--- a/app/services/proxy_deployment_service.rb
+++ b/app/services/proxy_deployment_service.rb
@@ -3,7 +3,6 @@
 class ProxyDeploymentService
   delegate :apicast_configuration_driven,
            :deployable?,
-           :service_mesh_integration?,
            :provider,
            :proxy_configs, to: :@proxy
 
@@ -15,11 +14,9 @@ class ProxyDeploymentService
     new(*args).call
   end
 
-  # TODO: remove v1_compatible when deploy_v1 is finally removed
-  def initialize(proxy, environment: :staging, v1_compatible: false)
+  def initialize(proxy, environment: :staging)
     @proxy = proxy
     @environment = environment
-    @v1_compatible = v1_compatible
   end
 
   def call
@@ -38,12 +35,10 @@ class ProxyDeploymentService
   def deploy
     return true unless deployable?
 
-    if service_mesh_integration?
-      deploy_v2 && deploy_production_v2
-    elsif apicast_configuration_driven?
+    if apicast_configuration_driven?
       deploy_v2
-    elsif @v1_compatible
-      deploy_v1
+    else
+      deploy_v2 && deploy_production_v2
     end
   end
 
@@ -53,11 +48,6 @@ class ProxyDeploymentService
     elsif @proxy.ready_to_deploy?
       provider.deploy_production_apicast
     end
-  end
-
-  # Deprecated
-  def deploy_v1
-    ApicastV1DeploymentService.new(provider).deploy(@proxy)
   end
 
   def deploy_v2
--- a/test/integration/user-management-api/services/proxies_test.rb
+++ b/test/integration/user-management-api/services/proxies_test.rb
@@ -43,6 +43,7 @@ class Admin::Api::Services::ProxiesTest < ActionDispatch::IntegrationTest
   def test_update
     params = provider_key_params.merge(proxy: { endpoint: 'https://alaska.wild' })
 
+    ProxyDeploymentService.any_instance.expects(:deploy_v2).times(2)
     Proxy.update_all(apicast_configuration_driven: false)
 
     assert_no_change of: ProxyConfig.method(:count) do
@@ -51,7 +52,6 @@ class Admin::Api::Services::ProxiesTest < ActionDispatch::IntegrationTest
     end
 
     Proxy.update_all(apicast_configuration_driven: true)
-    ProxyDeploymentService.any_instance.expects(:deploy_v2).once
 
     put(admin_api_service_proxy_path(params))
     assert_response :success
@@ -59,6 +59,7 @@ class Admin::Api::Services::ProxiesTest < ActionDispatch::IntegrationTest
 
   def test_update_for_service_mesh
     rolling_updates_on
+    Proxy.update_all(apicast_configuration_driven: false)
     params = provider_key_params.merge(proxy: { credentials_location: 'headers' })
     @service.update_column :deployment_option, 'service_mesh_istio'
 
--- a/test/unit/proxy_test.rb
+++ b/test/unit/proxy_test.rb
@@ -483,7 +483,6 @@ class ProxyTest < ActiveSupport::TestCase
                               api_backend: 'http://example.com',
                               api_test_path: '/path',
                               apicast_configuration_driven: false)
-    # TODO: remove this when ApicastV1DeploymentService class is removed
     ::ApicastV1DeploymentService.any_instance.expects(:deploy).with(proxy).returns(true).never
 
     assert proxy.save_and_deploy
--- a/test/unit/proxy_test.rb
+++ b/test/unit/proxy_test.rb
@@ -483,15 +483,8 @@ class ProxyTest < ActiveSupport::TestCase
                               api_backend: 'http://example.com',
                               api_test_path: '/path',
                               apicast_configuration_driven: false)
-    ::ApicastV1DeploymentService.any_instance.expects(:deploy).with(proxy).returns(true)
-
-    analytics.expects(:track).with('Sandbox Proxy Deploy', success: true)
-    analytics.expects(:track).with('Sandbox Proxy updated',
-                                   api_backend: 'http://example.com:80',
-                                   api_test_path: '/path',
-                                   success: true)
-    analytics.expects(:track).with('APIcast Hosted Version Change', {enabled: false, service_id: proxy.service_id, deployment_option: 'hosted'})
-    Logic::RollingUpdates.stubs(skipped?: true)
+    # TODO: remove this when ApicastV1DeploymentService class is removed
+    ::ApicastV1DeploymentService.any_instance.expects(:deploy).with(proxy).returns(true).never
 
     assert proxy.save_and_deploy
     assert proxy.persisted?
--- a/test/unit/services/proxy_deployment_service_test.rb
+++ b/test/unit/services/proxy_deployment_service_test.rb
@@ -8,7 +8,6 @@ class ProxyDeploymentServiceTest < ActiveSupport::TestCase
   end
 
   test 'deploy when apicast_configuration_driven' do
-    @proxy.expects(:service_mesh_integration?).returns(false)
     @proxy.expects(:apicast_configuration_driven).returns(true)
 
     service = ProxyDeploymentService.new(@proxy)
@@ -18,7 +17,7 @@ class ProxyDeploymentServiceTest < ActiveSupport::TestCase
   end
 
   test 'deploy with service mesh integration' do
-    @proxy.expects(:service_mesh_integration?).returns(true)
+    @proxy.expects(:apicast_configuration_driven).returns(false)
 
     service = ProxyDeploymentService.new(@proxy)
     service.expects(:deploy_v2).returns(true)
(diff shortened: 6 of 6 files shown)


app/services/proxy_deployment_service.rb
```diff
@@ -40,10 +38,8 @@ def deploy
 
     if service_mesh_integration?
       deploy_v2 && deploy_production_v2
-    elsif apicast_configuration_driven?
+    else
       deploy_v2
-    elsif @v1_compatible
-      deploy_v1
     end
```
akostadinov, line 43: I wonder if we shall raise some error in case no condition matches.

hallelujah: I think the business logic should strip the `v1_compatible`
Why would you just remove the option without removing the logic, and where it is called.

app/models/proxy.rb
```diff
@@ -355,7 +355,7 @@ def save_and_deploy(attrs = {})
 
     return false unless saved
 
-    success = ProxyDeploymentService.call(self, v1_compatible: true)
+    success = ProxyDeploymentService.call(self)
```
hallelujah, line 358: Shouldn't we remove this? What do you think? Can you explain why you keep it?

app/forms/onboarding/request_form.rb
```diff
@@ -49,7 +49,7 @@ def proxy_auth_params
   end
 
   def test_api!
-    return unless ProxyDeploymentService.call(model, v1_compatible: true)
+    return unless ProxyDeploymentService.call(model)
```
hallelujah, line 52: Can you explain why you keep it?

app/models/proxy.rb
akostadinov: Why keep the call or why keep `#save_and_deploy`?

josemigallas: I think the Service is still useful, isn't it? To encapsulate the deployment logic, so removing the argument should be the right call.

app/services/proxy_deployment_service.rb
josemigallas: If I recall correctly, we encapsulated all deployment logic within this Service so there should not be any other usage outside of it. Removing the argument `v1_compatible` and the V1Service will delete all related logic.

akostadinov: Still this is changing original logic from `if elsif elsif` to `if else`. What if we deploy neither service mesh, nor apicast v2? I heard about other integrations.

test/unit/forms/onboarding/request_form_test.rb
```diff
@@ -61,8 +61,6 @@ class Onboarding::RequestFormTest < ActiveSupport::TestCase
     proxy.save!
     proxy.update! sandbox_endpoint: 'http://staging.apicast.io'
 
-    ProxyDeploymentService.any_instance.expects(:deploy).returns(true)
```
akostadinov, line 64: This is related to the question above about `if else` statement.

test/unit/services/proxy_deployment_service_test.rb
```diff
@@ -9,7 +9,6 @@ def setup
 
   test 'deploy when apicast_configuration_driven' do
     @proxy.expects(:service_mesh_integration?).returns(false)
-    @proxy.expects(:apicast_configuration_driven).returns(true)
```
akostadinov, line 12: This one also seem related to the question above about `if else` statement.

test/integration/provider/admin/onboarding/wizard/request_controller_test.rb
```diff
@@ -31,11 +31,6 @@ class Provider::Admin::Onboarding::Wizard::RequestControllerTest < ActionDispatc
 
     put provider_admin_onboarding_wizard_request_path(request: { path: 'some invalid !!! path' })
     assert_response :success
-
-    ProxyDeploymentService.any_instance.expects(call: false)
-    put provider_admin_onboarding_wizard_request_path(request: { path: '/path' })
-    assert_response :success
-    assert_match 'please try again', response.body
```
akostadinov, line 38: This doesn't look as a V1 specific test. Could you explain why test failed?

test/integration/user-management-api/services/proxies_test.rb
```diff
@@ -43,13 +43,6 @@ def test_crud_provider_key
   def test_update
     params = provider_key_params.merge(proxy: { endpoint: 'https://alaska.wild' })
 
-    Proxy.update_all(apicast_configuration_driven: false)
-
-    assert_no_change of: ProxyConfig.method(:count) do
-      put(admin_api_service_proxy_path(params))
-      assert_response :success
-    end
```
akostadinov, line 51: This doesn't look as a V1 specific test but might be service mesh or other integration. Could you explain why test failed?

test/unit/services/proxy_deployment_service_test.rb
```diff
@@ -47,26 +46,6 @@ def setup
...
   end
 
-  test 'deploy neither service mesh or apicast configuration driven' do
-    @proxy.expects(:service_mesh_integration?).returns(false)
-    @proxy.expects(:apicast_configuration_driven).returns(false)
-
-    service = ProxyDeploymentService.new(@proxy)
-
-    refute service.call
-  end
```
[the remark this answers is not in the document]
HemaHG: this condition is related to `v1_compatiblie ` so if we are removing the logic related to `v1_compatiblie`, I think this test is not needed anymore

app/services/proxy_deployment_service.rb
HemaHG: I have reverted the change, not sure whether to raise some error if we deploy neither service mesh, nor apicast v2?
 please give your opinion @hallelujah

test/integration/provider/admin/onboarding/wizard/request_controller_test.rb
HemaHG: I have changed the test_api method in https://github.com/3scale/porta/blob/2542542cc655c9aa3e79c734b1391949394aebc4/app/forms/onboarding/request_form.rb. which is called in the update method of request_controller so the test was failing

app/forms/onboarding/request_form.rb
```diff
@@ -49,8 +49,6 @@ def proxy_auth_params
   end
 
   def test_api!
-    return unless ProxyDeploymentService.call(model, v1_compatible: true)
```
josemigallas, line 52: This is virtually `return unless false` which always returns, therefore I think the correct implementation would be:
```ruby
def test_api!
  return
end
```

josemigallas, line 52: I don't think you should be removing this line, just the argument.

app/services/proxy_deployment_service.rb
josemigallas: I agree with @akostadinov we should add an `else` (with a test).

test/integration/provider/admin/onboarding/wizard/request_controller_test.rb
josemigallas: Agree with @akostadinov , I don't think we have to remove this.

test/integration/user-management-api/services/proxies_test.rb
josemigallas: Agree with @akostadinov , I don't think we have to remove this.

test/unit/proxy_test.rb
```diff
@@ -483,16 +483,6 @@ def test_apicast_configuration_driven
...
-    ::ApicastV1DeploymentService.any_instance.expects(:deploy).with(proxy).returns(true)
-
-    analytics.expects(:track).with('Sandbox Proxy Deploy', success: true)
-    analytics.expects(:track).with('Sandbox Proxy updated',
-                                   api_backend: 'http://example.com:80',
-                                   api_test_path: '/path',
-                                   success: true)
-    analytics.expects(:track).with('APIcast Hosted Version Change', {enabled: false, service_id: proxy.service_id, deployment_option: 'hosted'})
-    Logic::RollingUpdates.stubs(skipped?: true)
-
```
josemigallas, line 495: This should not be removed, it's not related to the `v1_compatible` option. Instead, you should probably extract a new test:
`save_and_deploy` tests the method save_and_deploy
`save_and_deploy_v1_regression` tests that it does not throw an uncontrolled error or something like that

test/unit/services/proxy_deployment_service_test.rb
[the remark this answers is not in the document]
josemigallas: I think think this is related to `v1_compatible` and the test case is still valid: _When deploying with neither service mesh or apicast configuration driven -> it should refute_
It's what @akostadinov says, the method probably just needs a fallback `else`case.

app/services/proxy_deployment_service.rb
hallelujah: So as per this line https://github.com/3scale/porta/blob/9cfb6e7dbf79dc72ce19c93e2fbc2db75b0dec28/app/services/proxy_deployment_service.rb#L39

We should only go inside this method if it is either Service Mesh, APIcast hosted or APIcast self_managed see:

https://github.com/3scale/porta/blob/9cfb6e7dbf79dc72ce19c93e2fbc2db75b0dec28/app/models/proxy.rb#L466-L468
and follow references:

https://github.com/3scale/porta/blob/9cfb6e7dbf79dc72ce19c93e2fbc2db75b0dec28/app/models/proxy.rb#L467
https://github.com/3scale/porta/blob/9cfb6e7dbf79dc72ce19c93e2fbc2db75b0dec28/app/models/service.rb#L143

So it is safe not to add another branching condition in case of other integrations.

Also we should just remove the `v1_compatible: true` because this branch as explained earlier is not needed anymore

app/forms/onboarding/request_form.rb
hallelujah: You are right see https://github.com/3scale/porta/pull/2495/files#r660448399

test/integration/user-management-api/services/proxies_test.rb
hallelujah: Right, this can be service_mesh

test/unit/proxy_test.rb
hallelujah: Agree with @josemigallas just keep the test and remove (or change) it when ApicastV1DeploymentService is removed. We will see when the time comes

app/services/proxy_deployment_service.rb
```diff
@@ -42,8 +40,6 @@ def deploy
       deploy_v2 && deploy_production_v2
     elsif apicast_configuration_driven?
```
hallelujah, line 41: I expect this to change to

```ruby
   if apicast_configuration_driven?
      deploy_v2
   else # this is service mesh
      deploy_v2 && deploy_production_v2
   end
```

test/unit/services/proxy_deployment_service_test.rb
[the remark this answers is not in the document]
hallelujah: I think if we do like https://github.com/3scale/porta/pull/2495/files#r660455415 we can remove this test

app/forms/onboarding/request_form.rb
hallelujah: The reason is https://github.com/3scale/porta/pull/2495/files#r660448399

jsmadis (approved):

test/unit/proxy_test.rb
```diff
@@ -478,20 +478,13 @@ def test_apicast_configuration_driven
     end
   end
 
+ #TODO: remove the test when the ApicastV1DeploymentService class is removed
```
josemigallas, line 481:
```suggestion
```
Rather than removing the test, it should say remove [line 487](https://github.com/3scale/porta/pull/2495/files#diff-cc3f6466d2246468906e67e3164b6339cd96c6e611daa30aad5eb2cfe82e15b7R487):
```ruby
::ApicastV1DeploymentService.any_instance.expects(:deploy).with(proxy).returns(true).never
```

test/integration/user-management-api/services/proxies_test.rb
```diff
@@ -51,14 +52,14 @@ def test_update
...
     Proxy.update_all(apicast_configuration_driven: true)
-    ProxyDeploymentService.any_instance.expects(:deploy_v2).once
 
     put(admin_api_service_proxy_path(params))
     assert_response :success
   end
 
   def test_update_for_service_mesh
     rolling_updates_on
+    Proxy.update_all(apicast_configuration_driven: false)
```
josemigallas, line 62: Would this work?
```suggestion
     @proxy.expects(:apicast_configuration_driven).returns(false)
```

test/unit/services/proxy_deployment_service_test.rb
josemigallas, line 57: The expectation has changed, `deploy` will now always return even if neither service mesh or apicast config driven. So, instead of removing this test, change the assertion to:
```diff
+service.expects(:deploy_v2).once
+service.expects(:deploy_production_v2).once

- refute service.call
+ assert service.call
```

josemigallas (changes requested):

test/unit/services/proxy_deployment_service_test.rb
HemaHG: now we are checking only the apicast_configuration_driven? conditionhttps://github.com/3scale/porta/pull/2495#discussion_r660455415

test/integration/user-management-api/services/proxies_test.rb
HemaHG: it won’t work

test/unit/proxy_test.rb
```diff
@@ -483,15 +483,8 @@ def test_apicast_configuration_driven
...
-
-    analytics.expects(:track).with('Sandbox Proxy Deploy', success: true)
-    analytics.expects(:track).with('Sandbox Proxy updated',
-                                   api_backend: 'http://example.com:80',
-                                   api_test_path: '/path',
-                                   success: true)
-    analytics.expects(:track).with('APIcast Hosted Version Change', {enabled: false, service_id: proxy.service_id, deployment_option: 'hosted'})
-    Logic::RollingUpdates.stubs(skipped?: true)
+    # TODO: remove this when ApicastV1DeploymentService class is removed
+    ::ApicastV1DeploymentService.any_instance.expects(:deploy).with(proxy).returns(true).never
```
akostadinov, line 487: I don't think we can forget removing it as it will raise, so maybe the `TODO` comment not needed.

test/unit/services/proxy_deployment_service_test.rb
akostadinov: Then aren't we executing a conditional branch instead of NOT executing any (like it was in the past). Hery stated above that it should be harmless but the above assessment suggests otherwise.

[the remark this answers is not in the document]
josemigallas: OK yes we can remove this one 👍

test/unit/proxy_test.rb
HemaHG: now we are validating only apicast_configuration_driven -> true or false
```
   if apicast_configuration_driven?
      deploy_v2
   else # this is service mesh (and "future"others that do not have staging)
      deploy_v2 && deploy_production_v2
   end
```

josemigallas (approved):

hallelujah (approved):

PR #2496, opened 2021-06-17
THREESCALE-7134 removing :policies rolling update

akostadinov:

THREESCALE-7134

removing the `:policies` rolling update based on the comment in `Proxy#policy_chain`

### DIFF
--- a/app/lib/logic/rolling_updates.rb
+++ b/app/lib/logic/rolling_updates.rb
@@ -64,6 +64,9 @@ module Logic
         OPENSHIFT_PROVIDER_ID = 2
 
         def initialize(provider)
+          if deprecated?
+            Rails.logger.warn "Deprecated rolling update used: #{name} at:\n\t" + caller(4).join("\n\t")
+          end
           @provider = provider
         end
 
@@ -86,6 +89,10 @@ module Logic
           else
             raise_invalid_config
           end
+        end
+
+        def deprecated?
+          false
         end
 
         def raise_invalid_config
@@ -268,6 +275,16 @@ module Logic
       class BillableContracts < Base
         def missing_config
           false
+        end
+      end
+
+      class Policies < Base
+        def enabled?
+          true
+        end
+
+        def deprecated?
+          true
         end
       end
 
--- a/app/lib/logic/rolling_updates.rb
+++ b/app/lib/logic/rolling_updates.rb
@@ -271,17 +271,7 @@ module Logic
         end
       end
 
-      class Policies < Base
-        def missing_config
-          false
-        end
-      end
-
       class PolicyRegistry < Base
-        def enabled?
-          super && provider_can_use?(:policies)
-        end
-
         def missing_config
           false
         end
--- a/app/models/proxy.rb
+++ b/app/models/proxy.rb
@@ -151,8 +151,6 @@ class Proxy < ApplicationRecord
   end
 
   def policy_chain
-    # TODO: We need to remove this rolling update as it should be available for everyone using APIcast V2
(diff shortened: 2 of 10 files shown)


app/models/proxy.rb
```diff
@@ -151,8 +151,6 @@ def find_policy_config_by(name:, version:)
   end
 
   def policy_chain
-    # TODO: We need to remove this rolling update as it should be available for everyone using APIcast V2
-    return [] unless provider_can_use?(:policies)
     (policies_config.presence || []).each_with_object([]) do |config, chain|
       chain << config.slice('name', 'version', 'configuration') if config['enabled']
```
akostadinov, line 155: @hallelujah , these two lines are resulting in a reek error
```
Proxy#policy_chain refers to 'config' more than self (maybe move it to another class?) 
```

What do you think shall we do? I don't see an easy fix except for maybe creating a class `PoliciesConfig` and extract functionality there. But appears to be an overkill on the first sight.

Please let me know how should this be handled.

hallelujah: reek is totally right. It could be another PR, or another commit, as you wish. It should be `PolicyChainGenerator` class I think. PolicyConfig does not seem right as the PolicyConfig is per policy and the chain is a chain of policy configs.

akostadinov: I'm looking at this.

One slightly related question came though. Do we really need `policies_config.presence || []` here? It seems that we can happily go with `policies_config || []` because we don't mind if `policies_config` is blank or not. Except if we can get an empty string or another unsuitable type here.

Thanks!

hallelujah: no because it can be an empty string so presence is mandatory

app/lib/logic/rolling_updates.rb
```diff
@@ -271,17 +271,7 @@ def missing_config
...
 
-      class Policies < Base
-        def missing_config
-          false
-        end
-      end
-
       class PolicyRegistry < Base
-        def enabled?
-          super && provider_can_use?(:policies)
```
hallelujah, line 282: I would be conservative here. We decide to enable it for everybody. Just make it explicit we are enabling it, then in another PR we can just remove it, making sure that `provider_can_use.*(:policies)` and `provider_can_use.*(:policy_registry)` are all removed from the code.

Instead I would suggest:

```ruby
  def enabled?
    true
  end

  def missing_config
     true
  end
```

hallelujah (changes requested):
Lets be conservative ;)

akostadinov:
@hallelujah , I believe I made the desired changes. Please check again.

app/lib/logic/rolling_updates.rb
```diff
@@ -272,16 +279,16 @@ def missing_config
...
 
-      class PolicyRegistry < Base
-        def enabled?
-          super && provider_can_use?(:policies)
+        def deprecated?
+          true
         end
+      end
 
+      class PolicyRegistry < Base
```
hallelujah, line 291: We probably also need deprecated in this policy right?

akostadinov: It is not being removed. Should it be removed?

hallelujah: We do not remove the feature just will enforce it

akostadinov: This is another rolling update and is out of scope. Usages are not removed so should not be deprecated. I suggest we merge this one and then work on the other one separately.

hallelujah (approved):

PR #2497, opened 2021-06-17
self_managed apicast v2 no need to override method

akostadinov:
THREESCALE-7134

I'm using assumptions here but as far as I can understand the branch
`apicast_configuration_driven` means that v2 is used, v1 otherwise.

If this assumption is correct, then `generate(nil)` will return `nil` so
there is no point in overriding the method.

But there might be a third option I don't know about.


akostadinov:
So the discarded branch can be can be service mesh or another integration, not only v1 so I'm closing this one

PR #2498, opened 2021-06-22
PoliciesConfig custom attribute type

akostadinov:

This is in relation to #2496 where we hit a reek error with the `#policy_chain`
method. By refactoring `policies_config` attribute to be a custom type
we can avoid this style error and hopefully make things more organized overal.

Now this PR is not complete and is only a rough PoC. Other files need to be changed
also tests to be fixed and added.

I'm submitting it to get feedback whether this direction is desirable and also
to get feedback on validation and other approaches used.

Please check only last commit as this refactoring is on top of unmerged #2496.

### DIFF
--- a/app/controllers/admin/api/services/proxy/policies_controller.rb
+++ b/app/controllers/admin/api/services/proxy/policies_controller.rb
@@ -18,8 +18,7 @@ class Admin::Api::Services::Proxy::PoliciesController < Admin::Api::Services::Ba
   ##~ op.parameters.add @parameter_service_id_by_id_name
   #
   def show
-    policies_config = Proxy::PoliciesConfig.new(proxy.policies_config)
-    respond_with(policies_config, represent_on_error: :resource)
+    respond_with(proxy.policies_config, represent_on_error: :resource)
   end
 
   ##~ e = sapi.apis.add
@@ -41,9 +40,7 @@ class Admin::Api::Services::Proxy::PoliciesController < Admin::Api::Services::Ba
       ApicastV2DeploymentService.new(@proxy).call(environment: :sandbox)
     end
 
-    policies_config = Proxy::PoliciesConfig.new(proxy.policies_config)
-    policies_config.valid? # Needed to set errors if any
-    respond_with(policies_config, represent_on_error: :resource)
+    respond_with(proxy.policies_config, represent_on_error: :resource)
   end
 
   private
--- a/app/models/attributes/policies_config.rb
+++ b/app/models/attributes/policies_config.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+class Attributes::PoliciesConfig < ActiveRecord::Type::Text
+  def cast(value)
+    Proxy::PoliciesConfig.new(value)
+  end
+
+  def serialize(value)
+    value.to_json
+  end
+
+  def changed_in_place?(raw_old_value, new_value)
+    new_value != cast(raw_old_value)
+  end
+end
--- a/app/models/proxy.rb
+++ b/app/models/proxy.rb
@@ -129,7 +129,7 @@ class Proxy < ApplicationRecord
     deployment_option&.inquiry
   end
 
-  delegate :find_policy_config_by, :policy_chain, to: :policies_config
+  delegate :policy_chain, to: :policies_config
   delegate :self_managed?, :hosted?, to: :deployment_option
   delegate :service_token, to: :service, allow_nil: true
 
@@ -137,6 +137,9 @@ class Proxy < ApplicationRecord
     !hosted? && !self_managed?
   end
 
+  def find_policy_config_by(name:, version:)
+    policies_config.find_by(name: name, version: version)
+  end
 
   def oidc_configuration
     super || build_oidc_configuration(standard_flow_enabled: true)
@@ -498,15 +501,16 @@ class Proxy < ApplicationRecord
       @enabled = symbolized_attributes[:enabled]
     end
 
-    def matches?(name:, version:)
-      self.name == name && self.version == version
+    def matches?(other)
+      name == other.name && version == other.version
     end
+    alias =~ matches?
 
     def default?
       name == DEFAULT_POLICY["name"]
     end
 
-    def self.create_default
+    def self.default
       new(DEFAULT_POLICY)
     end
 
@@ -521,7 +525,7 @@ class Proxy < ApplicationRecord
     end
 
     def ==(other)
-      %i[name version configuration enabled].all? { |key| send(key) == other.send(key) }
+      %i[name version configuration enabled].all? { |key| public_send(key) == other.public_send(key) }
     end
     alias eql? ==
 
@@ -560,9 +564,9 @@ class Proxy < ApplicationRecord
     end
 
     def find_by(name:, version:)
-      find { |config| config.matches?(name: name, version: version) }
+      target = PolicyConfig.new({name: name, version: version, configuration: {}, enabled: false})
+      find { |config| config.matches?(target) }
     end
-    alias find_policy_config_by find_by
 
     def chain
       select(&:enabled).map(&:to_chain_value)
@@ -583,7 +587,7 @@ class Proxy < ApplicationRecord
 
     def policies_normalize(parsed_policies)
       policies = parsed_policies.map { |attrs| PolicyConfig.new(attrs) }
-      policies.tap { |list| list.push(PolicyConfig.create_default) if policies.none?(&:default?) }
+      policies.tap { |list| list.push(PolicyConfig.default) if policies.none?(&:default?) }
     end
 
     def policies_configs_are_correct
--- a/app/models/proxy.rb
+++ b/app/models/proxy.rb
@@ -14,11 +14,9 @@ class Proxy < ApplicationRecord
 
   self.background_deletion = [:proxy_rules, [:proxy_configs, { action: :delete }], [:oidc_configuration, { action: :delete, has_many: false }]]
 
-  DEFAULT_POLICY = { 'name' => 'apicast', 'humanName' => 'APIcast policy', 'description' => 'Main functionality of APIcast.',
-                     'configuration' => {}, 'version' => 'builtin', 'enabled' => true, 'removable' => false, 'id' => 'apicast-policy'  }.freeze
-
   belongs_to :service, touch: true, inverse_of: :proxy, required: true
   attr_readonly :service_id
+  attribute :policies_config, Attributes::PoliciesConfig.new
 
   has_many :proxy_rules, -> { order(position: :asc) }, dependent: :destroy, inverse_of: :proxy
   has_many :proxy_configs, dependent: :delete_all, inverse_of: :proxy
@@ -131,31 +129,7 @@ class Proxy < ApplicationRecord
     deployment_option&.inquiry
   end
 
-  def policies_config
-    parsed_config = read_and_parse_policies_config
-    return parsed_config if errors[:policies_config].present?
-
-    if parsed_config.detect { |c| c['name'] == DEFAULT_POLICY['name'] }
-      parsed_config
-    else
-      parsed_config.push(DEFAULT_POLICY)
-    end
-  end
-
-  def policies_config=(attr_policies_config)
-    super(attr_policies_config.is_a?(String) ? attr_policies_config : attr_policies_config.to_json)
-  end
-
-  def find_policy_config_by(name:, version:)
-    policies_config.find { |config| config['name'] == name && config['version'] == version }
-  end
-
-  def policy_chain
-    (policies_config.presence || []).each_with_object([]) do |config, chain|
-      chain << config.slice('name', 'version', 'configuration') if config['enabled']
-    end
-  end
-
+  delegate :find_policy_config_by, :policy_chain, to: :policies_config
   delegate :self_managed?, :hosted?, to: :deployment_option
   delegate :service_token, to: :service, allow_nil: true
 
@@ -502,7 +476,12 @@ class Proxy < ApplicationRecord
   class PolicyConfig
     include ActiveModel::Validations
 
+    DEFAULT_POLICY = { 'name' => 'apicast', 'humanName' => 'APIcast policy', 'description' => 'Main functionality of APIcast.',
+                       'configuration' => {}, 'version' => 'builtin', 'enabled' => true, 'removable' => false, 'id' => 'apicast-policy'  }.freeze
+
     attr_accessor :name, :version, :configuration, :enabled
+
+    delegate :to_json, :as_json, to: :to_h
 
     validates :name, :version, presence: true
     validate :configuration_is_object
@@ -519,6 +498,37 @@ class Proxy < ApplicationRecord
       @enabled = symbolized_attributes[:enabled]
     end
 
+    def matches?(name:, version:)
+      self.name == name && self.version == version
+    end
+
+    def default?
+      name == DEFAULT_POLICY["name"]
+    end
+
+    def self.create_default
+      new(DEFAULT_POLICY)
+    end
+
+    def to_chain_value
+      to_h.slice('name', 'version', 'configuration')
+    end
+
+    def to_h
+      %i[name version configuration enabled].each_with_object({}) do |key, obj|
+        obj[key] = public_send key
+      end.compact.as_json
+    end
+
+    def ==(other)
+      %i[name version configuration enabled].all? { |key| send(key) == other.send(key) }
+    end
+    alias eql? ==
+
+    def hash
+      to_h.hash
+    end
+
     private
 
     def configuration_is_object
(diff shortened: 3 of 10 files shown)


app/models/proxy.rb
```diff
@@ -536,44 +521,76 @@ class PoliciesConfig
...
+      end
     end
   end
 
-  def read_and_parse_policies_config
-    read_and_parse_policies_config!
-  rescue JSON::ParserError
-    []
-  end
+  class PoliciesConfigType < ActiveRecord::Type::Text
```
akostadinov, line 579: Yesterday we spoke about this. I believe we better extend `ActiveRecord` rather than `ActiveModel` because in Rails 2.5 they are the same class. But in 2.6 they are not. And here we are dealing with databases. So `ActiveRecord` appears to me as the appropriate option that should work with future Rails versions as well.

app/models/proxy.rb
```diff
@@ -529,49 +516,105 @@ def configuration_is_object
...
+
     def initialize(policies_config)
-      @policies_config = policies_config.map { |attrs| PolicyConfig.new(attrs) }
+      parsed_config = read_and_parse_policies_config(policies_config)
+
+      if parsed_config.detect { |c| c['name'] == DEFAULT_POLICY['name'] }
+        parsed_config
+      else
+        parsed_config.push(DEFAULT_POLICY)
+      end
```
josemigallas, line 535:
```suggestion
      parsed_config.push(DEFAULT_POLICY) unless parsed_config.detect { |c| c['name'] == DEFAULT_POLICY['name'] }
```

app/models/proxy.rb
```diff
@@ -529,49 +516,105 @@ def configuration_is_object
...
+    end
   end
 
   def policies_config_structure
-    parsed_config = read_and_parse_policies_config!
-    policies_object = PoliciesConfig.new(parsed_config)
-    return if policies_object.valid?
-    policies_object.errors.each do |attribute, message|
+    return if policies_config.valid?
+    policies_config.errors.each do |attribute, message|
```
josemigallas, line 615: At least my linter complaints about no adding blank lines after guards 😄 
```suggestion

    policies_config.errors.each do |attribute, message|
```

josemigallas (commented):
Couple of unimportant suggestion

app/models/attributes/policies_config.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class Attributes::PoliciesConfig < ActiveRecord::Type::Text
+  def cast(value)
+    if value.is_a?(Proxy::PoliciesConfig)
```
hallelujah, line 5: Code Climate will complain for sure :)

Anyway I do not think we need this, because `#cast` is called when retrieving values from the database.

app/models/proxy.rb
```diff
@@ -529,49 +530,71 @@ def configuration_is_object
   class PoliciesConfig
     include ActiveModel::Validations
 
-    delegate :each, to: :policies_config
+    delegate :each, :find, :select, :map, :detect, :reject, to: :policies_config
```
hallelujah, line 533:
```suggestion
    delegate :each, to: :policies_config
    include Enumerable
```

app/controllers/admin/api/services/proxy/policies_controller.rb
```diff
@@ -41,7 +40,7 @@ def update
       ApicastV2DeploymentService.new(@proxy).call(environment: :sandbox)
     end
 
-    policies_config = Proxy::PoliciesConfig.new(proxy.policies_config)
+    policies_config = proxy.policies_config
     policies_config.valid? # Needed to set errors if any
```
hallelujah, line 44: I think this should be having the states from the `update_attribute`

It would be too late (attributes already saved in database) to display any error here...

hallelujah (commented):
@akostadinov You mean Rails 5.2 and Rails 6.x right ?

app/models/attributes/policies_config.rb
akostadinov: I can't remember code climate complaining. But I can't review weaved warnings. Trying to filter them out yields nothing.

This casting was actually needed, IIRC for the `*changed?`methods.

app/controllers/admin/api/services/proxy/policies_controller.rb
akostadinov: This was how it originally worked I believe. I will try to extend `invalid config - does not update policies` test to make sure invalid entries are not saved to database.

app/models/attributes/policies_config.rb
akostadinov: I ran locally to double check:
```
$ reek app/models/attributes/policies_config.rb
Inspecting 1 file(s):
S

app/models/attributes/policies_config.rb -- 2 warnings:
  [4]:UtilityFunction: Attributes::PoliciesConfig#cast doesn't depend on instance state (maybe move it to another class?) [https://github.com/troessner/reek/blob/v6.0.4/docs/Utility-Function.md]
  [12]:UtilityFunction: Attributes::PoliciesConfig#serialize doesn't depend on instance state (maybe move it to another class?) [https://github.com/troessner/reek/blob/v6.0.4/docs/Utility-Function.md]
```

And using instance methods is how custom types are supposed to be defined as far as I can see. So the 2 errors are ignored. There is no complaint about type checking. And it was needed, that's why I added. Do you want me to investigate where it was needed?

app/models/proxy.rb
akostadinov: Updated, much nicer!

app/controllers/admin/api/services/proxy/policies_controller.rb
akostadinov: I think I missed your point originally. It can be safely removed. Thanks!

I did some testing to see that it makes no difference. Also `test/integration/user-management-api/services/proxy/policies_test.rb` already tests that error messages are rendered.

app/models/attributes/policies_config.rb
```diff
@@ -0,0 +1,19 @@
...
+    end
+  end
+
+  def serialize(value)
+    value.to_json
+  end
+
+  def changed_in_place?(raw_old_value, new_value)
+    new_value != cast(raw_old_value)
+  end
```
hallelujah, line 18: I think this is not needed if you have `deserialize` method

app/models/proxy.rb
```diff
@@ -519,6 +496,30 @@ def attributes=(attributes)
...
+      %i[name version configuration enabled].each_with_object({}) do |key, obj|
+        obj[key] = public_send key
+      end.compact.as_json
+    end
+
+    def ==(other)
+      self.class == other.class &&
+        %i[name version configuration enabled].all? { |key| send(key) == other.send(key) }
+    end
+    alias eql? ==
```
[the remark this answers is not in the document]
hallelujah: `eql?` ≠ `==`

There are 3 levels of "equality" in Ruby
- `==` soft equivalence
- `eql?` which uses `#hash`, basically the ones used to find same key in a `Hash` and should be the same class
- `equal?` the same object in memory


In which case do you need `eql?`?

app/models/attributes/policies_config.rb
hallelujah: Well it should :) https://github.com/troessner/reek/blob/master/docs/Simulated-Polymorphism.md
Maybe we allow it in our configuration :(

app/models/proxy.rb
```diff
@@ -528,50 +529,73 @@ def configuration_is_object
...
+      map(&:to_h).to_json(*args)
+    end
+
+    def find_by(name:, version:)
+      find { |config| config.matches?(name: name, version: version) }
+    end
+    alias find_policy_config_by find_by
+
+    def chain
+      select(&:enabled).map(&:to_h).map { |policy| policy.slice('name', 'version', 'configuration') }
```
hallelujah, line 563:
```suggestion
      select(&:enabled).map(&:to_chain_value)
```

app/models/proxy.rb
```diff
@@ -528,50 +529,73 @@ def configuration_is_object
...
+      policies = parsed_policies.map { |attrs| PolicyConfig.new(attrs) }
+      policies.tap { |list| list.push(PolicyConfig.create_default) if policies.none?(&:default?) }
+    end
+
     def policies_configs_are_correct
-      policies_config.each do |policy_config|
-        # TODO: 5: errors.merge!(policy_config.errors)
-        policy_config.errors.each { |attribute, message| errors.add(attribute, message) } unless policy_config.valid?
+      # TODO: 5: errors.merge!(policy_config.errors)
+      reject(&:valid?).map(&:errors).each do |policy_errors|
```
hallelujah, line 582: `select(&:invalid?)` reads better (just avoiding negative term in the second place)

app/representers/proxy_representer.rb
```diff
@@ -31,7 +31,7 @@ class ProxyRepresenter < ThreeScale::Representer
   property  :sandbox_endpoint
   property  :api_test_path
   property  :api_test_success
-  property  :policies_config
+  property  :policies_config, exec_context: :decorator
```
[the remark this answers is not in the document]
akostadinov: I thought that this is needed to override `#policies_config`. I'll try to remove this change and see if it still works.

app/models/proxy.rb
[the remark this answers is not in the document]
akostadinov: Probably copy/paste without giving it a good thought. Thanks.

app/models/attributes/policies_config.rb
akostadinov: Will try. Thanks.

app/models/proxy.rb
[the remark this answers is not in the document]
akostadinov: btw it doesn't appear to work like this
```
[11] pry(main)> pc = Proxy::PoliciesConfig.new([{ 'name' => 'schema', 'version' => '1', 'configuration' => {} }].to_json)
...
[12] pry(main)> pc.eql? Proxy::PoliciesConfig.new([{ 'name' => 'schema', 'version' => '1', 'configuration' => {} }].to_json)

=> false

[15] pry(main)> [pc.hash, Proxy::PoliciesConfig.new([{ 'name' => 'schema', 'version' => '1', 'configuration' => {} }].to_json).hash]
=> [-2449155912816168853, -2449155912816168853]
```

It seems to be a common practice to have `#eql?` aliased to `#==`. Shall I override `#eql?`?

app/models/attributes/policies_config.rb
akostadinov: `#deserialize` by default delegates to `#cast`. So it is always defined. This, as far as source comment reads, is needed in case something changed underlying object in place. Not sure anything can do that. I added it just in case.

from active_model/type/value.rb:
```

      # Determines whether the mutable value has been modified since it was
      # read. Returns +false+ by default. If your type returns an object
      # which could be mutated, you should override this method. You will need
      # to either:
      #
      # - pass +new_value+ to Value#serialize and compare it to
      #   +raw_old_value+
      #
      # or
      #
      # - pass +raw_old_value+ to Value#deserialize and compare it to
      #   +new_value+
      #
      # +raw_old_value+ The original value, before being passed to
      # +deserialize+.
      #
      # +new_value+ The current value, after type casting.
      def changed_in_place?(raw_old_value, new_value)
        false
      end
```

app/models/attributes/policies_config.rb
akostadinov: Seems like it is not needed. Maybe it failed before some fixes. As I wrote to the other comment, `#deserialize` just delegates to `#cast` so they are almost the same thing.

app/models/proxy.rb
```diff
@@ -529,49 +530,71 @@ def configuration_is_object
...
 
     validate :policies_configs_are_correct
 
     def initialize(policies_config)
-      @policies_config = policies_config.map { |attrs| PolicyConfig.new(attrs) }
+      parsed = policies_config.presence || []
+      parsed = parsed.is_a?(Array) ? parsed.as_json : Array(JSON.parse(parsed))
+
+      @policies_config = policies_normalize(parsed)
+    rescue JSON::ParserError
```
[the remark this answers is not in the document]
akostadinov: I've just tried to preserve original error messages and behavior.

wrt parsing I tried to strictly follow what we had in the past, just had to remove extra methods because of codeclimate errors, my desire was to keep as before for easier review and fewer changes. In the past you could get such invalid JSON and there are tests for invalid JSON input. We can of course change that on your request. I made existing tests pass. e.g. when invalid JSON is read from database or API call.

wrt duplicates, I can't spot where in the past we checked for duplicates. So it will be a new requirement to avoid duplicates. I think it's worth doing as a separate PR.

Please let me know how to proceed (and will write YARD once I know).

app/models/attributes/policies_config.rb
```diff
@@ -0,0 +1,19 @@
...
+    if value.is_a?(Proxy::PoliciesConfig)
+      value
+    else
+      Proxy::PoliciesConfig.new(value)
+    end
+  end
+
+  def serialize(value)
+    value.to_json
+  end
```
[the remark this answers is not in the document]
akostadinov: See above, it delegates to `#cast` and cast is fixed now.

test/integration/api/integrations_controller_test.rb
```diff
@@ -221,11 +221,16 @@ def test_update_proxy_rule_position
...
     api_backend = proxy_attributes.delete 'api_backend'
+    policies_config = proxy_attributes.delete 'policies_config'
     proxy.reload
 
     assert_equal proxy.attributes.slice(*proxy_attributes.keys), proxy_attributes
     assert_equal 'http://bye-world-api.3scale.net:80', proxy.api_backend
 
+    assert JSON.parse(policies_config).all? { |policy|
+      proxy.policies_config.detect Proxy::PolicyConfig.new(policy)
+    }
```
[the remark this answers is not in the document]
akostadinov: I just fixed the existing test because direct comparison is not possible. In the same way like it is done for `proxy.api_backend`. I updated the comparison to be simpler.

app/models/proxy.rb
```diff
@@ -519,6 +496,42 @@ def attributes=(attributes)
...
+    end
+
+    def ==(other)
+      self.class == other.class &&
+        %i[name version configuration enabled].all? { |key| send(key) == other.send(key) }
+    end
+    alias eql? ==
+
+    def hash
+      self.class.hash ^ to_h.hash
```
hallelujah, line 532: Same remark about bit shifting below

hallelujah (changes requested):

app/representers/proxy_representer.rb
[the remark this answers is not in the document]
akostadinov: I must have done some mistake when triggering tests because this is indeed needed. Not sure where on the internets I found it. But without it `ProxiesControllerTest` and `proxy_spec.rb` fail on converting the Proxy object to JSON properly. The overriding of `#policies_config` seems ignored without it.

app/representers/proxy_representer.rb
```diff
@@ -47,6 +47,10 @@ class ProxyRepresenter < ThreeScale::Representer
   property :jwt_claim_with_client_id, if: ->(*) { oidc? }
   property :jwt_claim_with_client_id_type, if: ->(*) { oidc? }
 
+  def policies_config
+    represented.policies_config.to_json
```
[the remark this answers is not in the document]
akostadinov: I tried, unfortunately it doesn't work. The representer seems to be calling `#to_s`. I'm wondering whether this behavior correct?

test/integration/user-management-api/services/proxy/policies_test.rb
```diff
@@ -12,23 +15,28 @@ def setup
...
+    assert_equal JSON.parse(response.body), {"policies_config"=>[DEFAULT_POLICY]}
   end
 
   def test_update_without_errors
-    put admin_api_service_proxy_policies_path(valid_params.merge(
-      { proxy: { policies_config: [{'name' => 'alaska', 'version' => '1', 'configuration' => {}}].to_json }}))
-    assert_match "{\"name\":\"alaska\",\"version\":\"1\",\"configuration\":{}}", response.body
+    example_policy = {'name' => 'alaska', 'version' => '1', 'configuration' => {}}
+
+    put admin_api_service_proxy_policies_path(valid_params.merge({ proxy: { policies_config: [example_policy].to_json }}))
```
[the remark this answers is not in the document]
akostadinov: It didn't work for me. This worked if you think it is better:
```ruby
put admin_api_service_proxy_policies_path(valid_params.merge({ params: { proxy: { policies_config: [example_policy].to_json }}}))
```

This is the definition of valid params:
```ruby
  def valid_params
    {
      service_id:   @service.id,
      access_token: @token.value,
      format:       :json
    }
  end
```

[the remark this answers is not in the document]
hallelujah: Ah well because it needs the service_id and format in the path ...
Fair enough, lets not waste time on this

hallelujah (approved):
:rocket: :1st_place_medal:

PR #2499, opened 2021-06-23
Remove whenever, use sidekiq-cron only

hallelujah:
We want to use sidekiq-cron in the context of containers

This PR is about:

- not running subshell while running tasks, for memory management.
  sidekiq pods get oom killed when multiple tasks spawn subshell + rails environment
- Cleaning definition of cron tasks and where they are run
- Update sidekiq-cron
- backward compatibility for smooth upgrade

## Backward compatibility

Former job arguments are handled

| Job             | Arguments |
| --------------- | --------- |
| CronJob::Worker | "CreateDefaultProxyWorker::BatchEnqueueWorker.perform_later" |
| CronJob::Worker | "DestroyAllDeletedObjectsWorker.perform_later(Service.to_s)" |
| CronJob::Worker | "DeleteProvidedAccessTokensWorker.perform_async" |
| CronJob::Worker | "Pdf::Dispatch.daily" |
| CronJob::Worker | "Cinstance.notify_about_expired_trial_periods" |
| CronJob::Worker | "LogEntry.delete_old" |
| CronJob::Worker | "Audited.audit_class.delete_old" |
| CronJob::Worker | "SegmentDeleteUsersWorker.perform_later" |
| CronJob::Worker | "FindAndDeleteScheduledAccountsWorker.perform_async" |


app/lib/three_scale/jobs.rb
```diff
@@ -0,0 +1,102 @@
...
+      [Finance::BillingStrategy, :daily_rest]
+    ]).freeze
+
+    HOUR = Task.map([
+      [Rails, :env]
+    ]).freeze # just a fake job to ensure cron works
+
+    SPHINX_INDEX_ALL = [RakeTask.new('sphinx:enqueue')].freeze
+
+    CUSTOM = [].freeze
```
akostadinov, line 99: Is this constant used anywhere?

app/lib/three_scale/jobs.rb
```diff
@@ -0,0 +1,102 @@
...
+module ThreeScale
+  module Jobs
+
+    class Task
+      def initialize(object, method, *args)
+        @object, @method, @arguments = object, method, args
+      end
+
+      def name
+        [@object, @method, @arguments].join('_').tr('():.', '_')
```
akostadinov, line 13: I'm not sure this is really needed but to distinguish between dots and the 3 strings concatenation, perhaps it can be `join("__")`.

akostadinov: jut as a quick visual clue in case anybody has any chance to be confused, perhaps it will be apparent anyway.

app/lib/three_scale/jobs.rb
```diff
@@ -0,0 +1,108 @@
...
+
+      def init_args
+        YAML.dump([@object, @method, @arguments])
+      end
+    end
+
+    class RakeTask < Task
+
+      def initialize(task_name) # rubocop:disable Lint/MissingSuper
+        @task_name = task_name
```
akostadinov, line 53: How about something like
```
Rails.application.load_tasks unless ::Rake.application
task = ::Rake.application[@task_name]
if task
  super(task, :invoke)
else
  # why should we allow missing tasks in cron at all?
  super(Rails.logger, :warn, "CRON: Rails task #{task_name} not defined")
end
```

then run would be
```
def run
  super
ensure
  @object.try :reenable
end
```

akostadinov: Not sure how it's gonna work with serialization though.

test/unit/lib/three_scale/jobs_test.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class ThreeScale::JobsTest < ActiveSupport::TestCase
+
+  def test_task_run
+    task = ThreeScale::Jobs::Task.new(String, :new, 'hello world')
+    task.run
+  end
```
akostadinov, line 10: maybe add a test case that exceptions would be raised. e.g. `Integer, :sqrt, -4`

akostadinov (dismissed):
Looks nice to me. I have some minor suggestions to consider but they are not important. So marking as approved.

app/lib/three_scale/jobs.rb
```diff
@@ -0,0 +1,108 @@
...
+
+        task.invoke
+      ensure
+        # We need to reenable the task as the instance is shared among workers
+        task&.reenable
+      end
+
+      def init_args
+        [@task_name]
+      end
```
akostadinov, line 70: I wonder if we can avoid defining this one for each sub-class. If we have in `Task`
```ruby
def initialize(...)
  @init_args = method(__method__).parameters.map do |_, name|
                 binding.local_variable_get(name)
               end
end
```
This means sub-class should call `super` in `#initialize` too.

akostadinov: Probably not, as this is going to run inside the parent class :/

akostadinov: yeah, no access to `binding`, pls ignore

app/lib/three_scale/jobs.rb
hallelujah: It is just cosmetic ... extracted from the test and also I should probably remove the trailing `_` but not worth it.

app/lib/three_scale/jobs.rb
hallelujah: I do not get you, this looks more complicated to *read* than previous implementation with separate purposes. 

A general `Task` that takes serializable objects, another `RakeTask` that takes the name of a rake task (maybe I could add arguments of the rake task but YAGNI)

The idea is to use simple basic objects, Rails.logger is really not serializable at all (because it holds an IO)

app/lib/three_scale/jobs.rb
hallelujah: Just keeping it as before, not used but keeping things as they were to make less changes possible.
I can remove it if you think necessary

app/lib/three_scale/jobs.rb
hallelujah: Looks complicated no? I would be scared if I have to read that code again in the future

test/unit/lib/three_scale/jobs_test.rb
hallelujah: The question would be: "raised by `what`?" 

-  `Integer.sqrt(-4)` raises an exception 
- or that `task.run` is aware of that exception?

There is no logic in my code that actively raises an exception. I do not think those kind of tests serve any purpose in understanding the logic, otherwise we can as well test any and all Exception like NoMethodError, RuntimeError etc .... and we will never be done testing.

I could probably test the `System::ErrorReporting` but it is just a capture the error, send it to a error service then re-raise it .... Dunno if it is good to have it here.

hallelujah (commented):
Thank you for your comments @akostadinov I replied inline ;)

app/lib/three_scale/jobs.rb
akostadinov: Better remove IMO, maybe in another PR if you want to make smaller steps.

test/unit/lib/three_scale/jobs_test.rb
akostadinov: The main benefit is to check behavior when a task raises some error, whatever it is.

My line of thinking is that tasks would raise errors from time to time. And we need to make sure that such occasions do not disturb our cron system as well they are properly reported in the log or elsewhere.

`Integer.sqrt(-4)` was just a suggested expression that would raise an error. Of course one can just use `raise "something"` :)

app/lib/three_scale/jobs.rb
akostadinov: Definitely. I see a lot of such magic in Ruby to save just a few lines of code, so I thought it is liked by people :)

Perhaps can be made a little clearer by splitting the ugly code into a method.
```ruby
def initialize(...)
  @init_args = passed_parameters(:initialize, binding)
  ...
end

def passed_parameters(for_method, method_binding)
  self.method(for_method).parameters.map do |_, name|
    method_binding.local_variable_get(name)
  end
end
```

But if you prefer the simplicity to read, it probably can't get any better than your original version.

P.S. above code is untested

app/lib/three_scale/jobs.rb
akostadinov: The point is to avoid instantiating tasks that do not exist. But this complicates things as you rightfully point out.

Perhaps at least add logging of missing tasks inside `#run` instead of just returning?

app/workers/cron_job.rb
```diff
@@ -1,30 +1,24 @@
...
-
-    def rake(task)
-      system('rake', task)
-    end
-
-    def runner(command)
-      system('rails', 'runner', command)
+  class Worker < ApplicationJob
+
+    def perform(hash)
```
guicassolato, line 6: We're changing the API. We need to be careful while deploying this, to make sure there are no jobs currently enqueued.

It might be new to Saas, but on-prem customers could still be affected.

hallelujah: I will create another class that only uses this one.
Thanks for the review

test/unit/lib/three_scale/jobs_test.rb
hallelujah: But that is the goal of Sidekiq, ... if it raises an error then it will be re-run and you can see why it failed.
I really do not see the point checking that at the level of `Task` as it really does not have any error handling.

I would probably agree to have such test if you tell me we do not have error handling in the `Task` class and we need one. But in this case it is really on purpose. By design, I do not want to handle error: they are bubbled up to Sidekiq

app/workers/cron_job.rb
hallelujah: In fact, it should be inside the documentation for the upgrade. What should be done is:

- disabling the sidekiq cron tasks
- when all sidekiq cron related tasks have been all processed make the upgrade

The issue why we can't have a smooth transition is that because of the Cron job enqueuer. So better stick to a documentation upgrade

app/lib/three_scale/jobs.rb
hallelujah: Still that reads awfully complicated ;)
People write meta-programming for some purposes. Generally they have their reason (especially in Rails).
But here I do not see any other reason than "copying what they write" which is a bad reason :)

Meta-programming is good when it facilitates method generation or object generation, using it anytime `because we can` is not really 

I agree sometimes we do some stuffs only  by pleasure, https://github.com/rails/rails/commit/22af62cf486721ee2e45bb720c42ac2f4121faf4#commitcomment-8135 

Still it should be first screen reader friendly.

Here I do not see a valid reason for that except `because I can` (which can be a valid reason too in another context)

test/unit/lib/three_scale/jobs_test.rb
akostadinov: Cool. As you see fit.

app/lib/three_scale/jobs.rb
akostadinov: The main reason is to avoid adding an override method in every sub-class. But it must be a balance between complication and verbosity so I'm alright with your judgement to drop it. It was just a suggestion.

app/workers/cron_job.rb
guicassolato: I would prefer the documentation route as well so we can avoid an extra class to maintain. Let's just make sure that works for deployments with the 3scale operator as well (not only for deployment with OpenShift templates).

cc @miguelsorianod @eguzki

guicassolato (dismissed):

app/workers/cron_job.rb
hallelujah: So last commit makes it backward compatible, when we do not need it anymore on future version we will remove that

guicassolato (approved):

PR #2500, opened 2021-06-23
removing deployment service class

srijitamukherjee:

Removing ApicastV1 Deployment service class

 part of [THREESCALE-7134](https://issues.redhat.com/browse/THREESCALE-7134)


test/unit/proxy_test.rb
```diff
@@ -478,17 +478,6 @@ def test_apicast_configuration_driven
     end
   end
 
-  test 'save_and_deploy' do
```
hallelujah, line 481: Do we have another test for apicast v2 or service mesh?

srijitamukherjee: no there is not.

test/unit/proxy_test.rb
```diff
@@ -483,8 +483,13 @@ def test_apicast_configuration_driven
                               api_backend: 'http://example.com',
                               api_test_path: '/path',
                               apicast_configuration_driven: false)
-    ::ApicastV1DeploymentService.any_instance.expects(:deploy).with(proxy).returns(true).never
+    ::ApicastV1DeploymentService.any_instance.expects(:call).returns(true)
```
hallelujah, line 486: This will fail if you removed it no?
Replace it with the equivalent of v2 if it does not exist already

HemaHG: since you are using Apicast, change `apicast_configuration_driven: false` to `true`

hallelujah (dismissed):
🚀

hallelujah (dismissed):

hallelujah (approved):

jsmadis (approved):

PR #2503, opened 2021-06-29
[gh-actions] To fast forward the candidate branch to default branch HEAD

hallelujah:
So we just have to press a button to do that

* the new candidate points by default to the current refs/heads/master
* the release version is required, it can be 2.11 or 2.12
  the workflow will create or update a branch named 3scale-{{ release }}-candidate

### DIFF
--- a/.github/workflows/fast-forward-candidate-branch.yml
+++ b/.github/workflows/fast-forward-candidate-branch.yml
@@ -3,7 +3,7 @@ on:
   workflow_dispatch:
     inputs:
       release:
-        description: 'The release version number e.g (2.11 or 2.12)'
+        description: 'The release version number (e.g, 2.11)'
         required: true
       ref:
         description: 'The SHA1 or branch name the candidate branch will point to'
--- a/.github/workflows/fast-forward-candidate-branch.yml
+++ b/.github/workflows/fast-forward-candidate-branch.yml
@@ -1,0 +1,21 @@
+name: Fast-forward candidate branch to HEAD of default branch
+on:
+  workflow_dispatch:
+    inputs:
+      release:
+        description: 'The release version number e.g (2.11 or 2.12)'
+        required: true
+      ref:
+        description: 'The SHA1 or branch name the candidate branch will point to'
+        required: true
+        default: refs/heads/master
+
+jobs:
+  fast-forward:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/fast-forward-candidate-branch.yml
```diff
@@ -0,0 +1,16 @@
...
+        description: 'The release number'
+        required: true
+        default: '2.11'
+
+jobs:
+  fast-forward:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - run: git push origin HEAD:3scale-${{ github.event.inputs.release }}-candidate
```
hallelujah, line 15: HEAD refers here to the branch against which you run the workflow.

So maybe make it configurable and pointing to the HEAD of main branch by default

.github/workflows/fast-forward-candidate-branch.yml
```diff
@@ -0,0 +1,21 @@
...
+on:
+  workflow_dispatch:
+    inputs:
+      release:
+        description: 'The release version number e.g (2.11 or 2.12)'
+        required: true
+      ref:
+        description: 'The SHA1 or branch name the candidate branch will point to'
+        required: true
+        default: refs/heads/master
```
hallelujah, line 11: I can't find a way to get the github default branch name :(

.github/workflows/fast-forward-candidate-branch.yml
```diff
@@ -0,0 +1,21 @@
+name: Fast-forward candidate branch to HEAD of default branch
+on:
+  workflow_dispatch:
+    inputs:
+      release:
+        description: 'The release version number e.g (2.11 or 2.12)'
```
guicassolato, line 6:
```suggestion
        description: 'The release version number (e.g, 2.11, 2.12)'
```

hallelujah, line 6: You need to explicitely type the correct branch number :)

.github/workflows/fast-forward-candidate-branch.yml
guicassolato: Can we just fix "master" for now?

hallelujah: I want it to be possible with other candidate too. Like if we need to have a 2.10 patch fix for example. This is not based on master.

guicassolato (approved):

PR #2511, opened 2021-07-02
[THREESCALE-7134] Removal of the proxy logs high level

jsmadis:

THREESCALE-7134 - removing proxy logs high level 


akostadinov:
Is this a ready PR or WIP? Should we have a database migration script added?

jsmadis:
@akostadinov this PR is ready. It only deletes high-level usage of the `proxy_rules`, which means that it doesn't touch anything database-related which might break something.

akostadinov (approved):
Got it now. Removes these from the API and leave removing DB and possibly other usages for later.

I am thinking we may want to deprecate any model usage, like some warning message in the logs whenever model class is initialized. This will allow us understand whether and where there are usages. See https://github.com/3scale/porta/pull/2496/files#diff-11abcaa744f5e2710aee0d37a5cce03dda18ef2d0a91312f6f5bb5843cae50f8R68 

WDYT?

jsmadis:
@akostadinov From my perspective, that is unnecessary. The usage of the model is almost completely deleted, and the UI was not accessible for a long time. Only the old accounts will have some data in the model, but the feature is already deprecated.

PR #2512, opened 2021-07-05
Disable Yarn integrity checks in Development

josemigallas:

Yarn's integrity checks detect unsatisfied or missing NPM dependencies. The problem is some of the error might be inevitable since are thrown by dependencies of a dependency and it's really annoying to get this error "out of nowhere" and block certain commands like `bundle` or the `assets:compile`.

```
error Couldn't find an integrity file                                                                                                                                                           
error Found 1 errors.
                                                                                                                                                                           
========================================
  Your Yarn packages are out of date!
  Please run `yarn install --check-files` to update.
========================================

To disable this check, please change `check_yarn_integrity`
to `false` in your webpacker config file (config/webpacker.yml).

yarn check v1.22.4
info Visit https://yarnpkg.com/en/docs/cli/check for documentation about this command.
```

**Verification steps** 

Hard to say but I'm sure everyone has seen this, specially @3scale/system-ux .

> Webpacker has officially removed yarn integrity since v5, and they recommend to disable it if causing problems.
> See [PR](https://github.com/rails/webpacker/pull/2518)
> And issues [rails/webpacker#2354](https://github.com/rails/webpacker/issues/2354), [rails/webpacker#2517](https://github.com/rails/webpacker/issues/2517)
> We are still using v4, but I think we can remove it.
> WDYT @didierofrivia @hallelujah ?

### DIFF
--- a/config/webpacker.yml
+++ b/config/webpacker.yml
@@ -33,6 +33,7 @@ development:
   <<: *default
   compile: true
   webpack_compile_output: true
+  check_yarn_integrity: false
 
   dev_server:
     host: localhost


hallelujah:
I always overcome it by removing `node_modules` and do `yarn install:safe`

@didierofrivia what do you think about this?

josemigallas:
> I always overcome it by removing `node_modules` and do `yarn install:safe`
> 
> @didierofrivia what do you think about this?

I usually am able to work around it as well but it's become super annoying lately and the only solution that works for me is disabling it.

config/webpacker.yml
```diff
@@ -33,6 +33,7 @@ development:
   <<: *default
   compile: true
   webpack_compile_output: true
+  check_yarn_integrity: false
```
damianpm, line 36: Webpacker has officially removed yarn integrity since v5, and they recommend to disable it if causing problems.
See [PR](https://github.com/rails/webpacker/pull/2518)
And issues https://github.com/rails/webpacker/issues/2354, https://github.com/rails/webpacker/issues/2517
We are still using v4, but I think we can remove it.
WDYT @didierofrivia @hallelujah ?

didierofrivia: I think so too. Let's disable it.

didierofrivia (approved):

damianpm (approved):

hallelujah (approved):

PR #2517, opened 2021-07-12
[THREESCALE-7134] Removing the nginx proxy

jsmadis:

Removing the Nginx proxy


app/services/proxy_deployment_service.rb
```diff
@@ -43,11 +43,7 @@ def deploy
   end
 
   def deploy_production
-    if apicast_configuration_driven?
-      deploy_production_v2
-    elsif @proxy.ready_to_deploy?
-      provider.deploy_production_apicast
-    end
+    deploy_production_v2 if apicast_configuration_driven?
```
jsmadis, line 46: This is the only change I am not sure about.

The `provider.deploy_production_apicast` function tests `return unless proxy_configs.present? && proxy_configs_conf.present?` which are objects of the apicast_v1. I am not sure if we should add a Rails warning to be sure or if this is enough.

@hallelujah

hallelujah: No need to add a warning IMO, we are removing them anyway

hallelujah (approved):

PR #2518, opened 2021-07-12
[THREESCALE-7134] Removal of the nginx controller

jsmadis:

Removing the Nginx controller


hallelujah:
I think the JSON format is still used https://github.com/3scale/APIcast/blob/e7f6ebbf15d1b053283f0d77ae7176e1c7bfcb44/doc/management-api.md?plain=1#L32

hallelujah (changes requested):
https://github.com/3scale/APIcast/blob/e7f6ebbf15d1b053283f0d77ae7176e1c7bfcb44/doc/management-api.md?plain=1#L32

The JSON format is still used :( so we can't remove the controller (pity)

@3scale/support can you confirm this is still supported?

kevprice83:
@hallelujah yes still supported and also required for usage with [`THREESCALE_CONFIG_FILE`](https://github.com/3scale/APIcast/blob/e7f6ebbf15d1b053283f0d77ae7176e1c7bfcb44/doc/parameters.md#threescale_config_file)

akostadinov:
Just to confirm, we support apicast v2 deployed outside container and it would download config from `THREESCALE_CONFIG_FILE`? Because for container use case, it is documented that config should be mounted as a volume.

jsmadis:
@akostadinov Yes the Apicast can be deployed anywhere and you can either download the config file or use the admin portal URL.

Closing this PR since the endpoint is  still being used.

PR #2520, opened 2021-07-12
Update swagger-ui to 3.51.1+

damianpm:

We have fixed `swagger-ui` versions to 3.25.0 in https://github.com/3scale/porta/pull/2075, because of an issue with node version in our servers.

Issue is fixed and we can finally update packages safely!

None

**Verification steps** 

```
rm -rf node_modules
yarn install
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -100,13 +100,10 @@
     "redux-api-middleware": "^3.0.1",
     "redux-thunk": "^2.3.0",
     "select2": "^4.0.6-rc.1",
-    "swagger-ui": "3.25.0",
+    "swagger-ui": "^3.51.1",
     "url-polyfill": "^1.1.3",
     "validate.js": "^0.13.1",
     "virtual-dom": "^2.1.1",
     "whatwg-fetch": "^3.0.0"
-  },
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -113,13 +113,10 @@
     "redux-api-middleware": "^3.0.1",
     "redux-thunk": "^2.3.0",
     "select2": "^4.0.6-rc.1",
-    "swagger-ui": "3.25.0",
+    "swagger-ui": "^3.51.1",
```
josemigallas, line 116: I would stick to using exact versions to prevent more unexpected errors. Not with swagger-ui specifically but with any NPM package.
@didierofrivia @damianpm opinions?

damianpm: Not sure, maybe having this unexpected errors is good, because it notifies that something has changed

damianpm: @didierofrivia, WDYT ?

damianpm: Let's merge this, and later we decide if we should block all dependencies to exact versions, or not

josemigallas (approved):

PR #2522, opened 2021-07-13
Turn Applications helpers into presenters

josemigallas:

This was really part of https://github.com/3scale/porta/pull/2459 but since it grew very big I decided to separate concerns. Right now the data for the NewApplication form is gathered by different helpers but they require a lot of arguments and they are hard(er) to test. Also, helpers are supposed to extend a model's logic but in this case we're simply preparing data to feed the frontend.

The presenter is used in `new` and in `create` so I opted for a singleton approach + helper method instead of declaring it as instance variable inside `new`.


test/unit/helpers/applications_helper_test.rb
```diff
@@ -0,0 +1,67 @@
...
+  end
+
+  test "application_defined_fields_data(provider)" do
+    skip 'TODO'
+  end
+
+  test "remaining_trial_days should return the right expiration date text" do
+    time = Time.utc(2015, 1,20, 10, 10, 10)
+    cinstance = FactoryBot.build(:cinstance, trial_period_expires_at: time)
+    expected_date = '&ndash; trial expires in <time datetime="2015-01-20T10:10:10Z" title="20 Jan 2015 10:10:10 UTC">20 days</time>'
```
hallelujah, line 61:
```suggestion
    expected_date = %Q[&ndash; trial expires in <time datetime="#{time.to_s(:iso8601)}" title="#{time.to_s}">20 days</time>]
```

test/unit/presenters/api/applications_new_presenter_test.rb
```diff
@@ -0,0 +1,45 @@
...
+    form_data = presenter.new_application_form_data
+
+    assert form_data.key? :"create-application-path"
+    assert form_data.key? :"create-application-plan-path"
+    assert form_data.key? :"service-subscriptions-path"
+    assert form_data.key? :"service-plans-allowed"
+    assert form_data.key? :"defined-fields"
+    assert form_data.key? :"most-recently-created-buyers"
+    assert form_data.key? :"buyers-count"
+    assert form_data.key? :product
```
hallelujah, line 28: I agree it is more verbose but you could do:

`assert_same_elements expected_keys, form_data.keys `

test/unit/presenters/api/applications_new_presenter_test.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    assert_not form_data.key? :buyer
+    assert_not form_data.key? :"most-recently-updated-products"
+    assert_not form_data.key? :"products-count"
+    assert_not form_data.key? :errors
```
hallelujah, line 33: 1. `refute` is preferred to `assert_not`
2. `assert_does_not_contain` is more readable ;)

```ruby
%i[buyer most-recently-updated-products products-count errors].each do |key|
  assert_does_not_contain form_data.keys, key
end
```

app/presenters/provider/admin/applications_new_presenter.rb
```diff
@@ -0,0 +1,35 @@
...
+      'products-count': raw_products.size,
+    }
+    data.merge new_application_form_base_data(provider, cinstance)
+  end
+
+  protected
+
+  def current_account
+    provider
+  end
```
guicassolato, line 33:
```suggestion
  alias current_account provider
```

app/controllers/api/applications_controller.rb
```diff
@@ -19,6 +19,8 @@ class Api::ApplicationsController < FrontendController
 
   sublayout 'api/service'
 
+  helper_method :presenter
```
guicassolato, line 22: This feels weird, especially because the presenter is only for the `new` action.

app/controllers/api/applications_controller.rb
```diff
@@ -19,6 +19,8 @@ class Api::ApplicationsController < FrontendController
...
+
   def index; end
 
   def new; end
```
guicassolato, line 26: Related to https://github.com/3scale/porta/pull/2522/files#r668873445

```suggestion
  def new
    @presenter = Api::ApplicationsNewPresenter.new(provider: current_account,
                                                   service: @service,
                                                   user: current_user,
                                                   cinstance: @cinstance)
  end
```

And you can get rid of the helper method. IMO, it's more straightforward.

app/controllers/api/applications_controller.rb
josemigallas: It's also used in `create` (see PR description).

test/unit/helpers/applications_helper_test.rb
```diff
@@ -0,0 +1,67 @@
...
+  test "most_recently_updated_products" do
+    skip 'TODO'
+  end
+
+  test "application_defined_fields_data(provider)" do
+    skip 'TODO'
+  end
+
+  test "remaining_trial_days should return the right expiration date text" do
+    time = Time.utc(2015, 1,20, 10, 10, 10)
```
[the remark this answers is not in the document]
josemigallas: I'll update the test, I think it should not assert the html element so thoroughly but that the displayed message is correct.

test/unit/presenters/api/applications_new_presenter_test.rb
josemigallas: No `refute_same_elements` then?

app/helpers/applications_helper.rb
```diff
@@ -0,0 +1,98 @@
+# frozen_string_literal: true
+
+module ApplicationsHelper
```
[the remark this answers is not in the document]
josemigallas: I'm not sure about a concern either, since this is not related to active records or models but the views.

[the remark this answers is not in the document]
hallelujah: Do you use those methods in the views or inside the presenter?

concerns are for any class (think of them as modules )

app/concerns/applications.rb
```diff
@@ -0,0 +1,65 @@
+# frozen_string_literal: true
+
+module Applications
+  extend ActiveSupport::Concern
+
```
hallelujah, line 5:
```suggestion
  included do
    include System::UrlHelpers.system_url_helpers
  end
```

app/presenters/api/applications_new_presenter.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+class Api::ApplicationsNewPresenter
+  include Applications
+  include PlansHelper
+  include System::UrlHelpers.system_url_helpers
```
hallelujah, line 6: I would prefer doing https://github.com/3scale/porta/pull/2522/files#r673858878 rather than repeating everytime when you include Applications

test/unit/concerns/applications_test.rb
```diff
@@ -0,0 +1,123 @@
...
+    include System::UrlHelpers.system_url_helpers
+
+    def setup
+      @provider = FactoryBot.create(:simple_provider)
+    end
+
+    def master_account
+      @master_account ||= Account.master || FactoryBot.create(:master_account)
+    end
+
```
hallelujah, line 18: You do not need this, `master_account` is defined in https://github.com/3scale/porta/blob/df50136a2d268c873373bef227680153354f2a83/test/test_helpers/master.rb#L5 which is included in

guicassolato (approved):

hallelujah (approved):
:rocket: :medal_military:

PR #2524, opened 2021-07-13
Disable messagebus as it was removed from the UI

hallelujah:
It is not used in the UI see https://github.com/3scale/porta/commit/9da51ed20f2fb7db4396c71459d3bf6c739bc039


akostadinov (approved):
Looks good. I assume additional integration testing would be needed before merging, unless such has been performed already.

hallelujah:
What do you mean by integration testing @akostadinov 
This is a default setting

PR #2527, opened 2021-07-15
🐛 "Create new application" button in the Developer Portal redirects to messages inbox

josemigallas:

The problem is when trying to find the service, if it's multi-service and there is no ID present, it redirects to `choose_service` page:
https://github.com/3scale/porta/blob/eb3a27beef090fe599ef52e0444635d99d17a654/lib/developer_portal/app/controllers/developer_portal/admin/applications_controller.rb#L157

Once in that page, if there's only 1 available service, it tries to redirect back to the new page with the first service's ID:
https://github.com/3scale/porta/blob/24c9bd2e6242ab1c165b9494e5ee1379ddc75e6f/lib/developer_portal/app/controllers/developer_portal/admin/applications_controller.rb#L96

BUT

it turns out the path is to `messages`, not the new application page:
```
# wrong path?
redirect_to new_provider_admin_application_path(2)
-> /admin/messages?action=new&controller=provider%2Fadmin%2Fapplications&service_id=2
```

SOLUTION:
Instead of redirecting twice, include a new clause to `service` in case there is only 1 service available. Redirect to choose_service as last resort.

[THREESCALE-7317: "Create new application" button in the Developer Portal redirects to messages inbox](https://issues.redhat.com/browse/THREESCALE-7317)

**Verification steps** 

* Enable **Multiple applications** and **Multiple services** in your dev portal: /p/admin/cms/switches
* Sign up as a new user in your dev portal (confirmation code will appear in the logs)
* Go to /admin/services, make sure there are at least 2 services, only 1 subscribed:
Screenshot 2021-07-15 at 14 00 02

* Go to /admin/applications, click on "Create application"
* The new application form should render correctly, with the Plan of the subscribed service.

### DIFF
--- a/lib/developer_portal/app/controllers/developer_portal/admin/applications_controller.rb
+++ b/lib/developer_portal/app/controllers/developer_portal/admin/applications_controller.rb
@@ -114,7 +114,7 @@ class DeveloperPortal::Admin::ApplicationsController < ::DeveloperPortal::BaseCo
 
   def service
     @service ||= if single_service?
-                   services.default
+                   services.first
                  elsif (service_id = params[:service_id])
                    services.where(id: service_id).or(services.where(system_name: service_id)).first
                  elsif current_account.services_can_create_app_on.count == 1
--- a/lib/developer_portal/app/controllers/developer_portal/admin/applications_controller.rb
+++ b/lib/developer_portal/app/controllers/developer_portal/admin/applications_controller.rb
@@ -117,8 +117,8 @@ class DeveloperPortal::Admin::ApplicationsController < ::DeveloperPortal::BaseCo
                    services.find_by(id: service_id) || services.find_by(system_name: service_id)
                  elsif single_service?
                    services.default
-                 elsif (available_services = current_account.services_can_create_app_on && available_services.count == 1)
-                   available_services.first
+                 elsif current_account.services_can_create_app_on.count == 1
+                   current_account.services_can_create_app_on.first
                  end
   end
 
--- a/lib/developer_portal/app/controllers/developer_portal/admin/applications_controller.rb
+++ b/lib/developer_portal/app/controllers/developer_portal/admin/applications_controller.rb
@@ -17,6 +17,8 @@ class DeveloperPortal::Admin::ApplicationsController < ::DeveloperPortal::BaseCo
 
   liquify prefix: 'applications'
 
+  delegate :services, :multiservice?, to: :site_account
+
   def index
     cinstances = current_account.bought_cinstances.includes(:service)
                    .order_for_dev_portal.paginate(page: params[:page])
@@ -92,9 +94,7 @@ class DeveloperPortal::Admin::ApplicationsController < ::DeveloperPortal::BaseCo
   end
 
   def choose_service
-    @services = current_account.services_can_create_app_on
-    redirect_to new_provider_admin_application_path(service_id: @services.first.id) if @services.count == 1
-    assign_drops services: Liquid::Drops::Collection.new(@services)
+    assign_drops services: Liquid::Drops::Collection.new(current_account.services_can_create_app_on)
   end
 
   def destroy
@@ -113,10 +113,12 @@ class DeveloperPortal::Admin::ApplicationsController < ::DeveloperPortal::BaseCo
   end
 
   def service
-    @service ||= if not site_account.multiservice?
-                   site_account.services.default
-                 elsif service_id = params[:service_id]
-                   site_account.services.find_by_id_or_system_name(service_id)
+    @service ||= if !multiservice?
+                   services.default
+                 elsif current_account.services_can_create_app_on.count == 1
+                   services.first
+                 elsif (service_id = params[:service_id])
+                   services.find_by(id: service_id) || services.find_by(system_name: service_id)
                  end
   end
 
--- a/lib/developer_portal/app/controllers/developer_portal/admin/applications_controller.rb
+++ b/lib/developer_portal/app/controllers/developer_portal/admin/applications_controller.rb
@@ -113,10 +113,10 @@ class DeveloperPortal::Admin::ApplicationsController < ::DeveloperPortal::BaseCo
   end
 
   def service
-    @service ||= if (service_id = params[:service_id])
-                   services.find_by(id: service_id) || services.find_by(system_name: service_id)
-                 elsif single_service?
+    @service ||= if single_service?
                    services.default
+                 elsif (service_id = params[:service_id])
+                   services.find_by(id: service_id) || services.find_by(system_name: service_id)
                  elsif current_account.services_can_create_app_on.count == 1
                    current_account.services_can_create_app_on.first
                  end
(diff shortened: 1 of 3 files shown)


lib/developer_portal/app/controllers/developer_portal/admin/applications_controller.rb
```diff
@@ -17,6 +17,8 @@ class DeveloperPortal::Admin::ApplicationsController < ::DeveloperPortal::BaseCo
 
   liquify prefix: 'applications'
 
+  delegate :services, :multiservice?, to: :site_account
```
akostadinov, line 20: For me personally this delegation is a little confusing. But feel free to ignore if others are ok with that. Maybe it's my lack of experience with this part of the code.

akostadinov, line 20: For me personally this delegation is a little confusing. But feel free to ignore if others are ok with that. Maybe it's my lack of experience with this part of the code.

akostadinov (approved):
I'm not sure about the delegation but other changes make sense to me.

lib/developer_portal/app/controllers/developer_portal/admin/applications_controller.rb
```diff
@@ -113,10 +113,12 @@ def applications
...
 
   def service
-    @service ||= if not site_account.multiservice?
-                   site_account.services.default
-                 elsif service_id = params[:service_id]
-                   site_account.services.find_by_id_or_system_name(service_id)
+    @service ||= if !multiservice?
+                   services.default
+                 elsif current_account.services_can_create_app_on.count == 1
+                   services.first
```
guicassolato, line 119: If multiple services switch is disabled and there is no default, then no service will be selected, right? Does it make sense changing to:

```suggestion
    @service ||= if site_account.service_items_in_menu? || current_account.services_can_create_app_on.count == 1
                   services.default || services.first
```

BTW, `Account#service_items_in_menu?` is a horrible method name in a model and it seems no longer to be used anywhere. We might want to enjoy the opportunity to perhaps rename it as `single_service?`.

lib/developer_portal/app/controllers/developer_portal/admin/applications_controller.rb
```diff
@@ -113,10 +113,12 @@ def applications
...
+                 elsif (service_id = params[:service_id])
```
guicassolato, line 120: I wonder if this condition shouldn't be the first one to check. 🤔

lib/developer_portal/app/controllers/developer_portal/admin/applications_controller.rb
josemigallas: Why is it confusing?

lib/developer_portal/app/controllers/developer_portal/admin/applications_controller.rb
josemigallas: I think there was a reason for this but I can't remember now so I'll move it first 😄

lib/developer_portal/app/controllers/developer_portal/admin/applications_controller.rb
josemigallas: In that case, wouldn't `services.first` be enough?

josemigallas: After giving it a though, I'd rather not change the logic of this block. I moved the `service_id` condition to first and renamed the method but otherwise the outcome should be the same. I don't know what will happen if we returned `services.first` so it will redirect to `choose_service` when service is `nil`.

lib/developer_portal/app/controllers/developer_portal/admin/applications_controller.rb
guicassolato: Maybe the reason why `service_id` went last was so the user could not force it when multiple services is not enabled. I think it's safer not to touch this. Sorry for suggesting before.

josemigallas: That must be it

lib/developer_portal/app/controllers/developer_portal/admin/applications_controller.rb
```diff
@@ -113,10 +113,12 @@ def applications
   end
 
   def service
-    @service ||= if not site_account.multiservice?
-                   site_account.services.default
-                 elsif service_id = params[:service_id]
-                   site_account.services.find_by_id_or_system_name(service_id)
+    @service ||= if (service_id = params[:service_id])
+                   services.find_by(id: service_id) || services.find_by(system_name: service_id)
```
hallelujah: Avoid doing 2 requests:

```suggestion
                   services.where(id: service_id).or(system_name: service_id).first
```

lib/developer_portal/app/controllers/developer_portal/admin/applications_controller.rb
hallelujah: We were thinking of removing default service years ago so I think we should just use `services.first`

lib/developer_portal/app/controllers/developer_portal/admin/applications_controller.rb
```diff
@@ -113,10 +113,12 @@ def applications
...
-    @service ||= if not site_account.multiservice?
-                   site_account.services.default
-                 elsif service_id = params[:service_id]
-                   site_account.services.find_by_id_or_system_name(service_id)
+    @service ||= if single_service?
+                   services.default
+                 elsif (service_id = params[:service_id])
+                   services.where(id: service_id).or(services.where(system_name: service_id)).first
+                 elsif current_account.services_can_create_app_on.count == 1
+                   current_account.services_can_create_app_on.first
```
guicassolato, line 121: `Account#services_can_create_app_on` means only services that the buyer account already has at least one `ServiceContract` with. Meaning this condition does not cover for the case where the buyer has `ServiceContract` with one service and here tries to create a new application with an application plan of another service. Is this what we want here?

Sorry. I still don't understand this condition very well.

josemigallas: All I can say is that this is the current behaviour 🤷‍♂️

guicassolato: And what happens when multi-services is enabled, the buyer lands to the new application page without a `service_id` and the buyer has not `ServiceContract`? `service` will be `nil`, right?

guicassolato: cc @thomasmaas

josemigallas: If there are no subscriptions for the account, the "create application" button will be hidden.
https://github.com/3scale/porta/blob/df50136a2d268c873373bef227680153354f2a83/lib/developer_portal/app/views/developer_portal/applications/index.html.liquid#L42-L44

If the user manually goes to `applications/new`: `service` will be `nil` and they will be redirected to the `choose_service` screen, which will be empty.

guicassolato: Fair enough. Let's move on and see how it goes. I guess "choosing the service" is the critical step that absolutely needs to happen upfront before hitting this point.

Thanks for explaining!

hallelujah (approved):

PR #2528, opened 2021-07-15
seems like unused representers

akostadinov:

I couldn't find any place these representers are being used. Trying to see if tests prove otherwise.


akostadinov:
It seems used with the `Policy` model. I was looking at `Proxy::PoliciesConfig` and `Proxy::PolicyConfig` only.

PR #2531, opened 2021-07-20
Updates oracle setup guide

jsmadis:

Updates oracle setup guide.

### DIFF
--- a/SETUP_ORACLE.md
+++ b/SETUP_ORACLE.md
@@ -1,5 +1,19 @@
 # Setup Oracle
 
+## Install Oracle dependencies using the script
+
+1. Run the script with sudo
+```shell
+$ sudo ./script/oracle/install-instantclient-packages.sh 
+```
+
+2. Add following ENV variables to your system (`~/.profile` or `~/.zshrc`)
+
+```shell
+LD_LIBRARY_PATH="/opt/oracle/instantclient/:$LD_LIBRARY_PATH"
+ORACLE_HOME=/opt/oracle/instantclient/
+```
+
 ## Install Oracle Instant Client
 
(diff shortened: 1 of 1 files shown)


hallelujah (approved):

script/oracle/install-instantclient-packages.sh
```diff
@@ -45,3 +45,7 @@ rm -rf /opt/system/vendor/oracle
 ln -sf /opt/oracle/instantclient_19_6/libsqora.so.19.1 /opt/oracle/instantclient_19_6/libsqora.so
 ln -sf /opt/oracle/instantclient_19_6 /opt/oracle/instantclient
 cp config/oracle/*.ini /etc/
+
+# Fix  ORA-12637: Packet receive failed
+# https://github.com/oracle/docker-images/issues/1352
+echo "DISABLE_OOB=ON" >> /opt/oracle/instantclient/network/admin/sqlnet.ora
```
hallelujah, line 51: Is it something customer also have or does this only happen on development? This file is shared with the oracle system image customers build. I do not want to set that by default if not needed

jsmadis: It only happened during the development.

I see, I will update the oracle setup guide with my experience then.

hallelujah (approved):

SETUP_ORACLE.md
```diff
@@ -1,5 +1,19 @@
 # Setup Oracle
 
+## Install Oracle dependencies using the script
+
+1. Run the script with sudo
+```shell
+$ sudo ./script/oracle/install-instantclient-packages.sh
```
akostadinov, line 7: @jsmadis , btw there are RPMs from oracle that can be used in Fedora or other RPM based distros

PR #2533, opened 2021-07-21
[WIP][THREESCALE-7134] Remove proxy logs low level

jsmadis:

Removes proxy logs from the database.

### DIFF
--- a/app/models/account/provider_methods.rb
+++ b/app/models/account/provider_methods.rb
@@ -35,7 +35,6 @@ module Account::ProviderMethods
     has_many :backend_api_metrics, through: :backend_apis, source: :metrics
     has_many :proxies, through: :services
     has_many :proxy_rules, through: :proxies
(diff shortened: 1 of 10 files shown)


app/models/account/provider_methods.rb
```diff
@@ -35,7 +35,6 @@ module Account::ProviderMethods
     has_many :backend_api_metrics, through: :backend_apis, source: :metrics
     has_many :proxies, through: :services
     has_many :proxy_rules, through: :proxies
-    has_many :proxy_logs, foreign_key: :provider_id
```
hallelujah, line 38: It is better first to remove the code then making the migration. Otherwise it will cause many errors on deploy

jsmadis:
Closing since @guicassolato recommended waiting for 1 on-premises release between removing the code and dropping the table.

Issue: https://issues.redhat.com/browse/THREESCALE-7410

PR #2536, opened 2021-07-22
[THREESCALE-7134] Remove code reference for the proxy_logs model

jsmadis:

Removes code that references proxy_logs model. 


akostadinov:
Don't we need migrations and schema updates with this change?

hallelujah (approved):

hallelujah:
It is funny nothing else is using it! Good job!

PR #2537, opened 2021-07-22
[THREESCALE-7134] Remove proxy conf columns code part

jsmadis:

Removes code part of the proxy conf columns.

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -2,6 +2,9 @@
 
 class Account < ApplicationRecord
   attribute :credit_card_expires_on, :date
+  self.ignored_columns = %i[proxy_configs_file_name proxy_configs_content_type proxy_configs_file_size
+                            proxy_configs_updated_at proxy_configs_conf_file_name proxy_configs_conf_content_type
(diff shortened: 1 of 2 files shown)


jsmadis:
@hallelujah I am not sure if I can remove the attached files. I haven't found any reference in the database scheme or anywhere else.

app/models/account.rb
```diff
@@ -2,6 +2,9 @@
 
 class Account < ApplicationRecord
   attribute :credit_card_expires_on, :date
+  self.ignored_columns = %i[proxy_configs_file_name proxy_configs_content_type proxy_configs_file_size
+                            proxy_configs_updated_at proxy_configs_conf_file_name proxy_configs_conf_content_type
+                            proxy_configs_conf_file_size proxy_configs_conf_updated_at]
```
hallelujah, line 7: :+1:

hallelujah (approved):

hallelujah:
Yes you can remove them, it is not used anymore anyway. They refer to old Lua files that are uploaded to S3

PR #2543, opened 2021-07-22
Adds captcha to Dev portal Login

josemigallas:

$Subject

[THREESCALE-6645: Developer portal lack of brute force protection on login](https://issues.redhat.com/browse/THREESCALE-6645)

**Verification steps** 

1. Set spam protection 'Always' in Developer Portal > Spam protection
2. Go to the signup page in the dev portal. It should be required to pass the captcha


features/developer_portal/login.feature
```diff
@@ -4,12 +4,49 @@ Feature: Login feature
...
+    And a buyer "bob" signed up to provider "foo.3scale.localhost"
+    And the current domain is "foo.3scale.localhost"
 
   @security
   Scenario: Buyer can log in with csrf protection enabled
-    Given the current domain is foo.3scale.localhost
     When I go to the login page
-     And I fill in the "bob" login data
+    And I fill in the "bob" login data
+    Then I should be logged in the Development Portal
```
akostadinov, line 15: how do we know csrf is enabled?

lib/developer_portal/app/controllers/developer_portal/login_controller.rb
```diff
@@ -26,14 +27,16 @@ def new
...
+    if sign_in_user
       self.current_user = @user
       create_user_session!
       flash[:notice] = @strategy.new_user_created? ? 'Signed up successfully' : 'Signed in successfully'
       redirect_back_or_default(@strategy.redirect_to_on_successful_login)
     elsif @strategy.redirects_to_signup?
       @strategy.on_signup(session)
       redirect_to @strategy.signup_path(params), notice: 'Successfully authenticated, please complete the signup form'
+    elsif !spam_check(@user)
+      render_creation_error('Spam protection failed.')
```
akostadinov, line 39: I can't understand how is this supposed to work. Under `sign_in_user` we have
```
 @user = @strategy.authenticate(params.merge(request: request))
```

So if that returns a user, then we don't go to spam protection. If it returns `nil` or `false`, then what is the point in passing `@user` as a parameter to `spam_check(@user)`?

I assume I'm missing details how this works. So excuse my ignorance.

features/developer_portal/login.feature
josemigallas: I'm gonna sound silly but why is CSRF relevant in the context of captcha? 😅

lib/developer_portal/app/controllers/developer_portal/login_controller.rb
josemigallas: You're not missing any details. I, on the contrary, am taking shots in the dark since I don't understand very well how the whole spam check flow works.
Ideally, `@user` is set inside `sign_in_user` so that it can be passed to `spam_check`. I am trying to replicate the same code from reset_password.

guicassolato: @akostadinov is right. If `#sign_in_user` sets a `@user`, you'll never get to `elsif !spam_check(@user)`. On the other hand, if it does not set a `@user`, then `@user` will be falsy and I don't see how `#spam_check` can decide based on that (`nil` or `false`? not cool, so no).

jsmadis: Fixed with my commit

app/lib/three_scale/spam_protection/integration/controller.rb
```diff
@@ -20,8 +20,9 @@ def spam_check(object, session_store: ThreeScale::SpamProtection::SessionStore)
         when :none
           true
         when :auto
-          if object.spam?
-            session_store.new(request.session).mark_possible_spam
+          stored_session = session_store.new(request.session)
+          if object.spam? && stored_session.marked_as_possible_spam?
+            stored_session.mark_possible_spam
```
jsmadis, line 25: This is a fix for the bug.

If the credentials were correct it would mark it as spam because we are passing a dummy (empty) model to it when the user is not logged in.

lib/developer_portal/app/controllers/developer_portal/login_controller.rb
```diff
@@ -25,8 +26,9 @@ def new
 
   def create
     logout_keeping_session!
+    return render_creation_error('Spam protection failed.') unless spam_check(buyer)
```
jsmadis, line 29: In my opinion, we must do a spam check before we are authenticating the user (so they can't DDoS the system).

The problem here is what model do we pass to the `spam_check`?

- `nil` - does make sense, but it's weird in my opinion
- dummy model - similar to `nil`, but it makes more sense in my opinion. However, the object is marked as spam with a 1.0 probability. That's why I added the check if the session is marked as spam.
- we find the user based on his username/email - we can't do this because it can be used for the DDoS.

In conclusion, I chose to go with the dummy model.

jsmadis: > Does this mean user will be asked for captcha before every single login?
Technically no, because we have 3 settings for the captcha (None, Suspicious only, and Always).
Suspicious only should decide based on our own checks if we trigger reCaptcha or not.

> Or that we just leave reCaptcha decide whether to ask user for something or not based on its own algorithms?
We are only checking if it passed the reCaptcha.

features/developer_portal/login.feature
akostadinov: Scenarion name is `Buyer can log in with csrf protection enabled`. That's why I asked about CSRF. Perhaps scenario name needs to be changed? Or CSRF part made more obvious.

features/developer_portal/login.feature
```diff
@@ -4,12 +4,34 @@ Feature: Login feature
...
+    When I go to the login page
+    Then I should see the captcha
+
+  @recaptcha
+  Scenario: Spam check fails (because of fast fill and no js)
+    Given provider "foo.3scale.localhost" has "spam protection level" set to "auto"
+    And I go to the login page
+    When timestamp spam check will return probability 1
+    And I fill in the "bob" login data
+    Then I should see the captcha
```
akostadinov, line 37: Perhaps test login with correct and incorrect captcha data by using steps `I fill in the captcha correctly` and `I fill in the captcha incorrectly`. Other than that PR looks great to me.

features/developer_portal/login.feature
josemigallas: I removed this, it didn't make sense IMO

akostadinov (approved):

lib/developer_portal/app/controllers/developer_portal/login_controller.rb
akostadinov: If there is an active attempt to brute-force a password or automate login, then the process will probably create a clean session by every attempt to login. So I'm not sure if this would effectively eliminate the captcha for bots and enable it only for real users mistyping their passwords.

I think such logic needs to be coordinated with somebody with experience with recaptcha.

PR #2544, opened 2021-07-23
instructions about environment setup

akostadinov:

Updates instructions about porta environemnt setup.


INSTALL.md
```diff
@@ -323,8 +323,18 @@ You may need to set up the database from scratch again, in that case, use `db:re
...
-Note down the credentials generated at this moment, to log in to each of the portals above.
+### Generating credentials
+
+Command above will seed the application and create the Master, Provider & Developer accounts that are accessible through: `http://master-account.3scale.localhost:3000/`, `http://provider-admin.3scale.localhost:3000/`, `http://provider.3scale.localhost:3000/` respectively. Take note of the credentials generated at this moment, to log in to each of the portals above.
+
+Alternatively you can set environment variables before running command to have predictable setup. For more you can look at `db/seed.rb`.
+
+```bash
+MASTER_PASSWORD=p
+USER_PASSWORD=ap # this is provider admin password
```
hallelujah, line 334: why not just `p` ? We used to have it as default. Simple, no need to think :)

INSTALL.md
```diff
@@ -323,8 +323,18 @@ You may need to set up the database from scratch again, in that case, use `db:re
...
+ADMIN_ACCESS_TOKEN=secret
+APICAST_ACCESS_TOKEN=apicastsecret
```
hallelujah, line 336: Same why not just `secret`. Simple, no need to think too much if it is for APIcast or for the Admin portal

akostadinov: I like to use different strings so I understand which variable affects which part of the system. Otherwise I don't know whether `MASTER_PASSWORD` or `USER_PASSWORD` is in effect for example (when they are the same string). 

I guess a personal quirk.

I updated to `p` and `secret`.

hallelujah: Well we do not really care if they are actually checked. This is done by test. What we need is a *quick* *simple* bootstrap.
Having too much variables makes things complicated and just make the brain confused

hallelujah (approved):

PR #2550, opened 2021-07-27
Removes submodules from porta codebase

jsmadis:

Removes docker-gateway submodule and makes the porta submodule free.


akostadinov (approved):
This is great. Why was it there in the first place?

hallelujah:
@akostadinov because of tests see https://github.com/3scale/porta/pull/2549

PR #2552, opened 2021-07-27
[oracle] Fix ORACLE_DIR

hallelujah:
So we do not have to recreate Oracle container everytime


akostadinov (approved):
Seems like a reasonable approach to me. Usually I'd put database data under `/var/lib`. But if oracle is already under `/opt` it makes sense.

hallelujah:
the `/opt` is inside the container .. the ORACLE_DIR is inside your home ... you do not want to use sudo to access `/var/lib`

PR #2554, opened 2021-07-28
[WIP] workaround oracle failure with PolicyConfig

akostadinov:

Just to show where am I with figuring out this issue. Not to be merged like that.

### DIFF
--- a/app/models/attributes/policies_config.rb
+++ b/app/models/attributes/policies_config.rb
@@ -2,11 +2,18 @@
 
 class Attributes::PoliciesConfig < ActiveRecord::Type::Text
   def cast(value)
+    if :oracle == System::Database.adapter.to_sym && value.is_a?(ActiveRecord::OracleEnhanced::Type::Text::Data)
+      value = value.to_s
+    end
     Proxy::PoliciesConfig.new(value)
   end
 
   def serialize(value)
-    value.to_json
+    if :oracle == System::Database.adapter.to_sym
+      ActiveRecord::OracleEnhanced::Type::Text::Data.new(value.presence&.to_json)
+    else
+      value.to_json
+    end
   end
 
   def changed_in_place?(raw_old_value, new_value)
--- a/app/models/proxy.rb
+++ b/app/models/proxy.rb
@@ -537,6 +537,7 @@ class Proxy < ApplicationRecord
     include Enumerable
 
     delegate :each, :to_json, :as_json, to: :policies_config
+    alias to_s to_json
     attr_reader :policies_config
     protected :policies_config
 
--- a/app/models/proxy.rb
+++ b/app/models/proxy.rb
@@ -16,7 +16,7 @@ class Proxy < ApplicationRecord
 
   belongs_to :service, touch: true, inverse_of: :proxy, required: true
   attr_readonly :service_id
-  attribute :policies_config, Attributes::PoliciesConfig.new
+  attribute :policies_config, :policies_config
 
   has_many :proxy_rules, -> { order(position: :asc) }, dependent: :destroy, inverse_of: :proxy
   has_many :proxy_configs, dependent: :delete_all, inverse_of: :proxy
--- a/config/initializers/active_record_types.rb
+++ b/config/initializers/active_record_types.rb
@@ -1,0 +1,3 @@
+ActiveSupport.on_load(:active_record) do
+  ActiveRecord::Type.register(:policies_config, Attributes::PoliciesConfig)
+end


app/models/attributes/policies_config.rb
```diff
@@ -2,11 +2,18 @@
...
+    if :oracle == System::Database.adapter.to_sym && value.is_a?(ActiveRecord::OracleEnhanced::Type::Text::Data)
+      value = value.to_s
+    end
     Proxy::PoliciesConfig.new(value)
   end
 
   def serialize(value)
-    value.to_json
+    if :oracle == System::Database.adapter.to_sym
+      ActiveRecord::OracleEnhanced::Type::Text::Data.new(value.presence&.to_json)
```
akostadinov, line 13: Please use this for illustrative purposes, we may probably be able to use another custom class type that is only for Oracle instead of making conditional logic by using `adapter` parameter like
```
ActiveRecord::Type.register(:boolean, ActiveRecord::OracleEnhanced::Type::Boolean, adapter: :oracleenhanced)
```

When we serialize into a `Data` object, I see in log
```
  SQL (0.7ms)  UPDATE "PROXIES" SET "PROXIES"."UPDATED_AT" = '2021-07-28 11:52:07.877513', "PROXIES"."POLICIES_CONFIG" = to_clob('[{"name":"my-policy","version":"0.5.0","configuration":{},"enabled":false},{"name":"apicast","version":"builtin","configuration":{},"enabled":true}]'), "PROXIES"."LOCK_VERSION" = 10 WHERE "PROXIES"."ID" = :a1 AND "PROXIES"."LOCK_VERSION" = :a2  [["id", 10000], ["lock_version", 9]]
```
when just `to_json`:
```
  SQL (6.6ms)  UPDATE "PROXIES" SET "PROXIES"."UPDATED_AT" = '2021-07-28 12:05:31.336869', "PROXIES"."POLICIES_CONFIG" = '[{"name":"my-policy","version":"0.5.0","configuration":{},"enabled":false},{"name":"apicast","version":"builtin","configuration":{},"enabled":true}]', "PROXIES"."LOCK_VERSION" = 11 WHERE "PROXIES"."ID" = :a1 AND "PROXIES"."LOCK_VERSION" = :a2  [["id", 10000], ["lock_version", 10]]
```

The difference is `to_clob` [1] which might be important in certain setups.

[1] https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/TO_CLOB-character.html#GUID-82E2FAD3-B0C8-4A06-A882-26211EE0524C

app/models/proxy.rb
```diff
@@ -537,6 +537,7 @@ class PoliciesConfig
     include Enumerable
 
     delegate :each, :to_json, :as_json, to: :policies_config
+    alias to_s to_json
```
akostadinov, line 540: This appears to me as a bug in enhanced adapter. Seems like `to_s` is called on the policies_config attribute instead of going through the `#serialize` method. I still have no idea why and how does that happen.

Without this alias, the database record is updated with a string like 
```
[5] pry(main)> proxy.attributes_before_type_cast
...
 "policies_config"=>
  "#<Proxy::PoliciesConfig:0x000000000bdb6590>\",\"configuration\":{},\"enabled\":false},{\"name\":\"apicast\",\"version\":\"builtin\",\"configuration\":{},\"enabled\":true}]"
...
```

app/models/attributes/policies_config.rb
hallelujah: Yes we need `to_clob` see https://github.com/3scale/porta/blob/master/config/initializers/oracle.rb#L49-L50

We need this in order to be able to update LOB column otherwise the adapter will set them to NULL see https://github.com/rsim/oracle-enhanced/issues/1588#issuecomment-343353756

akostadinov: Looked more into it. From what I see, I suspect `to_clob` does not matter at all. When writing the field I can see in the logs:
```
>  SQL (1.7ms)  UPDATE "PROXIES" SET "PROXIES"."UPDATED_AT" = '2021-07-29 09:01:50.764140', "PROXIES"."POLICIES_CONFIG" = to_clob('[{"name":"my-policy","version":"1.0.0","configuration":{},"enabled":true},{"name":"my-other-policy","version":"0.5.0","configuration":{},"enabled":false},{"name":"apicast","version":"builtin","configuration":{},"enabled":true}]'), "PROXIES"."LOCK_VERSION" = 2 WHERE "PROXIES"."ID" = :a1 AND "PROXIES"."LOCK_VERSION" = :a2  [["id", 10277], ["lock_version", 1]]
>  Writable Large Object (1.6ms)  SELECT "POLICIES_CONFIG" FROM "PROXIES" WHERE "ID" = 10277 FOR UPDATE
```

I read this as basically the field is updated after the initial `UPDATE` statement.

It happens in [#write_lobs line 727](https://github.com/rsim/oracle-enhanced/blob/release17/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb#L727) where just `value.to_s` is used. Unfortunately this is non-serialized value.

On [line 718](https://github.com/rsim/oracle-enhanced/blob/release17/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb#L718) serialization should happen but it is conditional in case value is `ActiveRecord::Type::Serialized` which it is not. I have no idea why value is not serialized always.

akostadinov: Writing a column is skipped in case attribute is `blank?` on [line 718](https://github.com/rsim/oracle-enhanced/blob/release17/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb#L718) So the `#to_clob` possibly matters for blank values only?

akostadinov: IMO it is a bug that `#write_lobs` on doesn't serialize all values or at least provide some mechanism for that.

akostadinov:
FYI I submitted a [bug report](https://github.com/rsim/oracle-enhanced/issues/2200) upstream, hopefully oracle_enhanced developers can give suggestions in case this is fixable in another way. I can also submit a PR to oracle_enhanced in case I don't see a reply soon.

So far the easy workaround would be to just alias `#to_s` to `#to_json` and see if upstream gets fixed, then pick up changes when we upgrade Rails.

I would ignore the `#to_clob` and avoid this serialization into a `Data` object, because it seems to only kick in when value is blank. But for `policies_config` we never write blank values anyway.

@hallelujah, please let me know how you want to see this issue fixed.

akostadinov:
For now going with #2557

PR #2557, opened 2021-07-30
workaround oracle enhanced adapter calling #to_s instead of proper #serialize

akostadinov:
This is a workaround for the issue is better explained #2554 and https://github.com/rsim/oracle-enhanced/issues/2200


hallelujah (approved):
Good! We can make a proper fix of the adapter later.

Thank you 🚀

PR #2559, opened 2021-08-04
Fix active_merchant_mode

hallelujah:
This is needed for staging, which will set it to 'test' instead of production (the default)

Closes https://issues.redhat.com/browse/THREESCALE-7393


jsmadis (approved):
LGTM

hallelujah:
Finally we do not need this, as per review by @guicassolato

PR #2561, opened 2021-08-05
use context for Quay password instead of global env

akostadinov:

Use circle CI context for quay credentials instead of global env.

One could run oracle tests without creating a branch in main repo when part of the org. Or so I assume.

Also perhaps there is a way to approve such runs from external contributors?

**Verification steps** 

Oracle tests should run properly, at least DB container should be created properly.

I'm not sure how to test this PR. Will try to run from my fork branch. If this doesn't work, I'm not sure.

Also I'm not aware how this will affect workflow when we want changes from an external contributor to be tested on oracle. I think at least it will not become harder than before.


akostadinov:
superseded by #2562

PR #2565, opened 2021-08-05
remove broken tests

akostadinov:

Removing broken perf tests.

<del>Rename test classes and files to avoid collisions between them.</del>


guicassolato (changes requested):
I don't see how current class names are a problem for Ruby. It must be some environment issue, to which I hardly believe that renaming is the right solution.

In case inevitably we need to do this, let's please avoid abbreviations like "Int" for "Integration". It took an extra second to realize it didn't stand for "Integer".

akostadinov:
@guicassolato , it is a problem if you locally in IntelliJ start all minitests for example it complains for same class trying to inherit from multiple superclasses because integration and unit tests class names collide. What names would you suggest me to use?

akostadinov:
@guicassolato , the issue that happens is that running tests in IntelliJ IDEA results in errors like:
```
test/integration/api_docs/tracking_controller_test.rb:5:in `<top (required)>': superclass mismatch for class TrackingControllerTest (TypeError)
````

I think it is not a good practice to have conflicting class names in a project anyway. Would you suggest a better renaming pattern that would not create such confusion? How about using `TrackingControllerIntegrationTest`?

hallelujah:
Well you are not supposed to run all those tests at the same time.
They do not serve the same purpose.

I think it is OK but I would prefer to have smaller PR for each renaming

hallelujah (approved):

PR #2568, opened 2021-08-09
ActiveDocs: Do not inject autocomplete fields in plain/text responses

damianpm:

When testing an API using swagger-ui, if the response returns just plain text (not json), it will be displayed as array of chars.

https://issues.redhat.com/browse/THREESCALE-6379

**Verification steps** 

- You need an OAS v3, with and endpoint that returns only plain text.
  - You can use this example I've created to test it:
  - https://gist.github.com/damianpm/93d5876313a75ac7807fb6c4a5dc3da5
- In Admin portal, create a new ActiveDoc

- Be sure the new ActiveDoc is NOT associated to any service ("Service" field must be empty)
  - In this way, swagger-ui will make the requests directly to the test API I've created to test. If you select a service, then Apicast will be used as proxy and you'll need to configure Apicast to avoid CORS issues.

**Before:**

**After:**

### DIFF
--- a/app/javascript/packs/OAS3Autocomplete.js
+++ b/app/javascript/packs/OAS3Autocomplete.js
@@ -59,8 +59,8 @@ const injectParametersToPath = (path: PathItemObject, commonParameters?: Array<P
   }, {})
 )
 
-const injectAutocompleteToResponseBody = (responseBody: ResponseBody, accountData: AccountData): ResponseBody => (
-  {
+const injectAutocompleteToResponseBody = (responseBody: ResponseBody, accountData: AccountData): ResponseBody => {
+  const res = (responseBody.paths && accountData) ? {
     ...responseBody,
     paths: Object.keys(responseBody.paths).reduce(
       (paths, path) => {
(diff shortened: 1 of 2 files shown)


app/javascript/packs/OAS3Autocomplete.js
```diff
@@ -69,10 +69,15 @@ const injectAutocompleteToResponseBody = (responseBody: ResponseBody, accountDat
         paths[path] = injectParametersToPath(responseBody.paths[path], commonParameters, accountData)
         return paths
       }, {})
-  }
-)
+  } : responseBody
+  return res
+}
 
 const injectServerToResponseBody = (responseBody: ResponseBody, serviceEndpoint: string): ResponseBody => {
```
josemigallas, line 76:
```suggestion
const injectServerToResponseBody = (responseBody: ResponseBody | string, serviceEndpoint: string): ResponseBody => {
```

josemigallas (changes requested):

app/javascript/packs/OAS3Autocomplete.js
```diff
@@ -88,15 +91,7 @@ const injectServerToResponseBody = (responseBody: ResponseBody | string, service
   }
 }
 
-// response.body.method is not present when fetching the spec,
-// is present when doing a request to one of the paths
-const isSpecFetched = (response: SwaggerResponse): boolean => !!response.body.method
```
damianpm, line 88: I just realized this method is totally useless, don't remember why it was included

app/javascript/packs/OAS3Autocomplete.js
```diff
@@ -59,8 +59,11 @@ const injectParametersToPath = (path: PathItemObject, commonParameters?: Array<P
   }, {})
 )
 
-const injectAutocompleteToResponseBody = (responseBody: ResponseBody, accountData: AccountData): ResponseBody => (
-  {
+const injectAutocompleteToResponseBody = (responseBody: ResponseBody | string, accountData: AccountData): ResponseBody | string => {
+  if (typeof responseBody === 'string') {
+    return responseBody
+  }
+  const res = (responseBody.paths && accountData) ? {
```
josemigallas, line 66:
```suggestion
  const res = (typeof responseBody !== 'string' && responseBody.paths && accountData) ? {
```
Alternatively add it to guard in line 63
```
if (typeof responseBody === 'string' || !responseBody.paths || !accountData) {
  return responseBody
}
```

josemigallas (approved):

PR #2569, opened 2021-08-11
Update bundle to 2.2.25

josemigallas:

Updates bundle otherwise `loofah` won't be installed.

[THREESCALE-7398: Bundle wont install dependencies because of loofah](https://issues.redhat.com/browse/THREESCALE-7398)

**Verification steps** 

```
bundle install
```

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1002,4 +1002,4 @@ DEPENDENCIES
   zip-zip
 
 BUNDLED WITH
-   2.2.18
+   2.2.25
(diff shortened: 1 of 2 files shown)


josemigallas:
> I retested in a clean environment and it did not happen anymore. Probably a hiccup in rubygems API.
> 
> If you still want to do the change then do it also:
> 
> * https://github.com/3scale/system-builder/blob/a814068f7b26de7eb11c5b5c02b2d440ee055cc2/Dockerfile#L3
> * https://github.com/3scale/porta/blob/102b955f30a13a3373da0887f475d008e3ff8bd3/openshift/system/Dockerfile#L52
> * and in the dist-git for productization

Yeah, I've noticed the error isn't happening anymore 😅 but let's do it.

hallelujah (approved):

PR #2573, opened 2021-08-12
Plans index page should show no custom plans

josemigallas:

The Plans table is showing a total number that doesn't match what is paginated. This happens because the paginated items (the ones rendered in the table) are `plans.not_custom` but the count is calculated from `plans`.

Solution: instead of filtering `not_custom` plans in the view helper, do it in the controller when `plans` is declared.
Before:
Screenshot 2021-08-12 at 09 39 31
After:
Screenshot 2021-08-17 at 17 40 54

[THREESCALE-7426: application plans index has empty pages](https://issues.redhat.com/browse/THREESCALE-7426)

**Verification steps** 

You'll need some plans and custom plans. In master, go to `Master API > Audience > Application Plans` and look at the table's pagination widget. Verify the number of items in the table matches the total count. For instance, "before" image there are less than 40 items but pagination says there are 170.

### DIFF
--- a/app/controllers/api/plans_base_controller.rb
+++ b/app/controllers/api/plans_base_controller.rb
@@ -43,7 +43,8 @@ class Api::PlansBaseController < Api::BaseController
 
   def find_plans
     search = ThreeScale::Search.new(params[:search] || params)
-    @plans = collection.order_by(params[:sort], params[:direction])
+    @plans = collection.not_custom
+                       .order_by(params[:sort], params[:direction])
                        .scope_search(search)
     @page_plans = @plans.paginate(pagination_params)
   end
--- a/app/helpers/api/plans_helper.rb
+++ b/app/helpers/api/plans_helper.rb
@@ -12,8 +12,7 @@ module Api::PlansHelper
   end
 
   def application_plans_data(plans)
-    plans.not_custom
-         .alphabetically
+    plans.alphabetically
          .to_json(root: false, only: %i[id name])
   end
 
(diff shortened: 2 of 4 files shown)


jsmadis (approved):
LGTM

app/controllers/api/plans_base_controller.rb
```diff
@@ -43,7 +43,8 @@ def find_plan
 
   def find_plans
     search = ThreeScale::Search.new(params[:search] || params)
-    @plans = collection.order_by(params[:sort], params[:direction])
+    @plans = collection.not_custom
```
hallelujah, line 46: Can collection be:

```ruby
def collection
  super.not_custom
end
```

So it applies everywhere in this controller?

app/helpers/api/plans_helper.rb
```diff
@@ -12,8 +12,7 @@ def default_application_plan_data(service, plans)
   end
 
   def application_plans_data(plans)
-    plans.not_custom
-         .alphabetically
+    plans.alphabetically
```
hallelujah, line 15: Why the change here?

josemigallas: The data helper really does not care if "plans" are custom or not, it simply arranges the data for the UI.

app/controllers/api/plans_base_controller.rb
josemigallas: OK but it's that what's supposed to be? Always `not_custom`?

hallelujah (approved):

PR #2574, opened 2021-08-12
[Bugfix] Braintree: Do not save CC that failed 3DSecure verification

damianpm:

When saving Credit Card details for Braintree, some CC that should be rejected are being stored.

The reasons why a CC may be rejected are multiple, some related to the card itself (wrong data entered by user, etc.), some to another issues (timeouts, etc). 

Braintree errors are a bit cryptic and technical (e.g: "THREEDS_LOOKUP_VALIDATION_ERROR"), and not very useful for regular users. For this reason, a generic error is shown when validation fails:

"An error occurred, please review your CC details or try later."

Later, we should improve error messages.

Additionally, fixing some alignment issues in Braintree injected form fields.

Related to 
https://issues.redhat.com/browse/THREESCALE-6860

**Verification steps** 

You can use Braintree test cards:
https://developer.paypal.com/braintree/docs/guides/3d-secure/testing-go-live#testing

### DIFF
--- a/app/javascript/src/PaymentGateways/braintree/BraintreeForm.jsx
+++ b/app/javascript/src/PaymentGateways/braintree/BraintreeForm.jsx
@@ -57,14 +57,8 @@ const BraintreeForm = ({
   }, [braintreeNonceValue])
 
   const get3DSecureError = (response) => {
-    if (
-      response.name === 'BraintreeError' ||
-      response.threeDSecureInfo.status !== 'authenticate_successful' ||
-      response.threeDSecureInfo.status !== 'authenticate_attempt_successful'
-    ) {
-      return CC_ERROR_MESSAGE
-    }
-    return null
+    const { threeDSecureInfo: { status: message } = { status: null } } = response
+    return message && message.match(/authenticate_(attempt_)?successful/) ? null : CC_ERROR_MESSAGE
   }
 
   const get3DSecureNonce = async (payload) => {
--- a/app/javascript/src/PaymentGateways/braintree/BraintreeForm.jsx
+++ b/app/javascript/src/PaymentGateways/braintree/BraintreeForm.jsx
@@ -21,6 +21,8 @@ import './BraintreeCustomerForm.scss'
 import type { Node } from 'react'
 import type { BraintreeFormProps, BillingAddressData } from 'PaymentGateways'
 
+const CC_ERROR_MESSAGE = 'An error occurred, please review your CC details or try later.'
+
 const BraintreeForm = ({
   braintreeClient,
   billingAddress,
@@ -54,12 +56,22 @@ const BraintreeForm = ({
     }
   }, [braintreeNonceValue])
 
+  const get3DSecureError = (response) => {
+    if (
+      response.name === 'BraintreeError' ||
+      response.threeDSecureInfo.status !== 'authenticate_successful' ||
+      response.threeDSecureInfo.status !== 'authenticate_attempt_successful'
+    ) {
+      return CC_ERROR_MESSAGE
+    }
+    return null
+  }
+
   const get3DSecureNonce = async (payload) => {
     const threeDSecureInstance = await create3DSecureInstance(threeDSecure, braintreeClient)
     const response = await veryfyCard(threeDSecureInstance, payload, billingAddressData)
-    const error = response.name === 'BraintreeError'
-      ? response.code === 'THREEDS_LOOKUP_VALIDATION_ERROR' ? response.details.originalError.details.originalError.error.message : response.message
-      : null
+
+    const error = get3DSecureError(response)
     const nonce = response.nonce || null
 
     return {
@@ -80,7 +92,7 @@ const BraintreeForm = ({
   }
 
   const handleCardError = (error: string) => {
-    setCardError(`Credit card errors found: ${error}. Please correct your CC data.`)
+    setCardError(error)
     clearHostedFields()
     setIsLoading(false)
   }
--- a/app/javascript/src/PaymentGateways/braintree/BraintreeCustomerForm.scss
+++ b/app/javascript/src/PaymentGateways/braintree/BraintreeCustomerForm.scss
@@ -16,3 +16,11 @@
   font-style: italic;
   margin-bottom: 2rem;
 }
+
+.list-unstyled {
+  .form-control.col-md-6 {
+    // Hack: needed because Braintree styles are colliding with ours
+    // scss-lint:disable ImportantRule
+    width: 50% !important;
+  }
+}


app/javascript/src/PaymentGateways/braintree/BraintreeForm.jsx
```diff
@@ -54,12 +56,22 @@ const BraintreeForm = ({
     }
   }, [braintreeNonceValue])
 
+  const get3DSecureError = (response) => {
+    if (
+      response.name === 'BraintreeError' ||
+      response.threeDSecureInfo.status !== 'authenticate_successful' ||
+      response.threeDSecureInfo.status !== 'authenticate_attempt_successful'
```
akostadinov, line 63: Wouldn't `response.threeDSecureInfo.status !== 'authenticate_successful' || response.threeDSecureInfo.status !== 'authenticate_attempt_successful'` always be `true`?

Even if `response.threeDSecureInfo.status` is one of the values, lets say `authenticate_successful`, then it can't at the same time be `authenticate_attempt_successful`?

app/javascript/src/PaymentGateways/braintree/BraintreeForm.jsx
```diff
@@ -21,6 +21,8 @@ import './BraintreeCustomerForm.scss'
 import type { Node } from 'react'
 import type { BraintreeFormProps, BillingAddressData } from 'PaymentGateways'
 
+const CC_ERROR_MESSAGE = 'There was an error validating your credit card details, please double check and enter valid credentials.'
```
akostadinov, line 24: Where did `or try later` go? Or did you purposefully decide not to include it?

app/javascript/src/PaymentGateways/braintree/BraintreeForm.jsx
damianpm: No, `response.threeDSecureInfo.status` can't be at the same time `authenticate_successful` and `authenticate_attempt_successful`.
The idea is that it will evaluate to `true` when any of the 3 conditions is true.

damianpm:
> Didn't we agree on the following copy for the generic error message:
> "An error occurred, please review your CC details or try later"
> 
> I think the "or try later/again" part of the message is especially important for errors such as timeouts, where basically the error is not thrown because of issues with CC details.

I copied the message from the last chat you shared (the one with 3 points), but I agree, let's use "An error occurred, please review your CC details or try later"

damianpm:
> Does this pr also prevent the erroneuos card from being saved?

Yes, because if an error is found, it prevents from sending the request to the backend

app/javascript/src/PaymentGateways/braintree/BraintreeForm.jsx
akostadinov: Exactly. 

```
> x = 'authenticate_successful'
> x != 'authenticate_successful' || x != 'authenticate_attempt_successful'
=> true
```

The above expression is always `true` regardless of `x` value. Probably your intention was to do like below?
```
      response.name === 'BraintreeError' || (
        response.threeDSecureInfo.status !== 'authenticate_successful' &&
        response.threeDSecureInfo.status !== 'authenticate_attempt_successful'
      )  
```

guicassolato: I too like better this approach of looking for a success instead of looking for an error. Shouldn't we be going all-in in this case and remove the `response.name === 'BraintreeError' ||` part altogether?

Also, why now suppressing Braintree original error message? Is this due to some concern of a possible XSS exploit?

What I had in mind:

```javascript
const get3DSecureError = (response) => {
  var status = response.threeDSecureInfo.status
  return status && status.match(/authenticate_(attempt_)?successful/) ? null : response.message || response.details.originalError.error.message || status
}
```

(To confirm if either `details.originalError` or `details.originalError.details.originalError`. Not sure which one is correct.)

damianpm:
@guicassolato @akostadinov 

I have updated the method, please take a look again.
- Gui: We supressed original Braintree error messages, after a discussion with @thomasmaas  and @cntlsn , because messages are cryptic and/or very technical, being useless for most users. We considered more convenient to have a generic, but human friendly message.
- About the response from Braintree (when errors), the problem is that it's a bit inconsistent, it doesn't return always the same fields.
  - `response.name` sometimes is defined, sometimes is not.
  - The same happens with `response.threeDSecureInfo.status`, it may be defined, or not.
  - And the same happens with `response.details.originalError.details.originalError.error.message`
  - We can have `response.name === 'BraintreeError'`, but it may not have a `response.threeDSecureInfo.status` field.
  - And we can have an error when `response.name !== 'BraintreeError'`

app/javascript/src/PaymentGateways/braintree/BraintreeForm.jsx
```diff
@@ -54,12 +56,20 @@ const BraintreeForm = ({
     }
   }, [braintreeNonceValue])
 
+  const get3DSecureError = (response) => {
+    return (
+      response.name === 'BraintreeError' ||
+        !!response.threeDSecureInfo.status.match(/authenticate_(attempt_)?successful/) === false
```
akostadinov, line 62: I don't know about js, but are there any benefits of using `!!response.threeDSecureInfo.status.match(/authenticate_(attempt_)?successful/) === false` instead of `!response.threeDSecureInfo.status.match(/authenticate_(attempt_)?successful/)`?

app/javascript/src/PaymentGateways/braintree/BraintreeForm.jsx
```diff
@@ -54,12 +56,20 @@ const BraintreeForm = ({
...
+        ? CC_ERROR_MESSAGE
+        : null
+    )
+  }
```
didierofrivia, line 66: If we consider that `response.status` could be `undefined` and thus treat it as an error, then it could be:

```suggestion
  const get3DSecureError = (response) => (
      response.name === 'BraintreeError' ||
        !response.status && !!response.threeDSecureInfo.status.match(/authenticate_(attempt_)?successful/) === false
        ? CC_ERROR_MESSAGE
        : null
    )

```

It's a pity we don't have types defined for the specific statuses, nor a list of them. Pattern matching is OK for now to identify them, but would be better to assert the specific statuses / errors in the future... If Braintree doesn't provide them, we should write them ourselves based on Braintree API documentation.

app/javascript/src/PaymentGateways/braintree/BraintreeForm.jsx
didierofrivia: @akostadinov the double `!!` will cast `true` or `false` out of every value, `null`, `undefined` or `0` would be `false` and pretty much any other `string`, `number`, `object` or `array`(even if they are empty) would be `true`

didierofrivia: @akostadinov You are right, even tho the double bang will cast every "falsey" value to `false`, the single one will turn that "truthy" one matching our assertion

damianpm: @akostadinov @didierofrivia  

The problem is not when _falsy_, but when _truthy_.
If `response.threeDSecureInfo.status.match(/authenticate_(attempt_)?successful/)` finds a match, it will return an array, then it will be flipped to `false`, thanks to JS dynamic type coercion:

Screenshot 2021-08-18 at 10 18 35

That's why we need the longest:
`!!response.threeDSecureInfo.status.match(/authenticate_(attempt_)?successful/) === false`

damianpm: In the other hand, yes, we should check if `response.threeDSecureInfo.status` is defined, I will push the change

didierofrivia: Yet another reason to have good types and understanding of their API. A week from now and we will forget why the double band and false assertion...

app/javascript/src/PaymentGateways/braintree/BraintreeForm.jsx
damianpm: I agree, but there are no types provided by Braintree, we need to craft them. That's out of scope of this quick-fix.

app/javascript/src/PaymentGateways/braintree/BraintreeForm.jsx
damianpm: We should create a new task for that, adding types and improving error messages

app/javascript/src/PaymentGateways/braintree/BraintreeForm.jsx
```diff
@@ -54,12 +56,23 @@ const BraintreeForm = ({
     }
   }, [braintreeNonceValue])
 
+  const get3DSecureError = (response) => {
+    return (
+      response.name === 'BraintreeError' ||
+        (
+          !response.threeDSecureInfo.status ||
+          !!response.threeDSecureInfo.status.match(/authenticate_(attempt_)?successful/) === false
```
akostadinov, line 64: If we don't have `status` then we consier this an error? Is this the intended logic?

as a nit-pick with the logic above, the parentheses don't seem needed as well I didn't really understand why `!response.threeDSecureInfo.status.match(/authenticate_(attempt_)?successful/)` will not work.

app/javascript/src/PaymentGateways/braintree/BraintreeForm.jsx
damianpm: @akostadinov ⬆️

app/javascript/src/PaymentGateways/braintree/BraintreeForm.jsx
damianpm: @akostadinov 

> If we don't have status then we consier this an error?

Yes, when OK the response returns `response.threeDSecureInfo.status` with either 'authenticate_successful' or 'authenticate_attempt_successful' value.

About the logic of !!, explanaition was lost with new commit, check [here](https://github.com/3scale/porta/pull/2574#discussion_r691014450)

damianpm: About the status, you can see this Braintree [doc](https://developer.paypal.com/braintree/docs/guides/3d-secure/server-side#status-codes):
Only "authenticate_attempt_successful" and "authenticate_successful" are accepted, there was also "authenticate_successful_issuer_not_participating" but is now deprecated.

akostadinov: Can we have a test or two that will cover this logic?

damianpm: It's difficult to test this part, because those fields (credit card inputs) are injected by braintree in an iframe.
I wouldn't try to do it in this PR

app/javascript/src/PaymentGateways/braintree/BraintreeForm.jsx
```diff
@@ -54,12 +56,23 @@ const BraintreeForm = ({
     }
   }, [braintreeNonceValue])
 
+  const get3DSecureError = (response) => {
+    return (
+      response.name === 'BraintreeError' ||
```
guicassolato, line 61: I still don't think we need to check this. It either has `authenticate_(attempt_)?successful` or it's an error, no?

didierofrivia: if it reduces to that, then it's not needed at all

app/javascript/src/PaymentGateways/braintree/BraintreeForm.jsx
```diff
@@ -57,14 +57,12 @@ const BraintreeForm = ({
...
-      response.name === 'BraintreeError' ||
-      response.threeDSecureInfo.status !== 'authenticate_successful' ||
-      response.threeDSecureInfo.status !== 'authenticate_attempt_successful'
-    ) {
-      return CC_ERROR_MESSAGE
-    }
-    return null
+    return (
+      !response.threeDSecureInfo ||
+      !!response.threeDSecureInfo.status.match(/authenticate_(attempt_)?successful/) === false
```
didierofrivia, line 62: So we are confident, without having types, that `response.threeDSecureInfo.status` won't be `undefined`?

app/javascript/src/PaymentGateways/braintree/BraintreeForm.jsx
```diff
@@ -57,14 +57,12 @@ const BraintreeForm = ({
...
-    if (
-      response.name === 'BraintreeError' ||
-      response.threeDSecureInfo.status !== 'authenticate_successful' ||
-      response.threeDSecureInfo.status !== 'authenticate_attempt_successful'
-    ) {
-      return CC_ERROR_MESSAGE
-    }
-    return null
+    return (
+      !response.threeDSecureInfo ||
```
damianpm, line 61: @guicassolato 
You were right, we don't need to verify `response.name === 'BraintreeError'`, but we need to verify `response.threeDSecureInfo`

- With unsupported cards, it doesn't try to do the 3Dsecure verification. The response will include `response.name`, but not `response.threeDSecureInfo` object at all.
- When the card is supported, it calls 3DSecure `verifyCard`, and response will have `threeDSecureInfo.status`

app/javascript/src/PaymentGateways/braintree/BraintreeForm.jsx
damianpm: Unfortunately, there are no types available, but checking the docs and doing a lot of testing, I think yes.
When `response.threeDSecureInfo`, it will contain `status`

didierofrivia: And what does the documentation about their API say?

damianpm: Well, you need to check several docs, but you can start with
General reference: https://braintree.github.io/braintree-web/current/ThreeDSecure.html
Status codes: https://developer.paypal.com/braintree/docs/guides/3d-secure/server-side#status-codes
Testing: https://developer.paypal.com/braintree/docs/guides/3d-secure/testing-go-live

didierofrivia: I'd rather trust you have, indeed, read it and can assure us that not checking for `status` `undefined` is possible since their docs says so... instead of pausing my tasks to read their several docs. Also, if it was up to me, I'd just check if `undefined` and report an error.... :)

akostadinov: +1 for checking `undefined` regardless of docs.

damianpm: Ok, checking if undefined, see line 62

guicassolato: One last suggestion that looks cleaner IMHO:

```javascript
const get3DSecureError = (response) => {
  const { threeDSecureInfo: { status: message } = { status: null } } = response
  return message && message.match(/authenticate_(attempt_)?successful/) ? null : CC_ERROR_MESSAGE
}
```

akostadinov: @guicassolato , this looks awesome, I'm impressed by Destructuring assignment syntax.

didierofrivia: you could go even cleaner and do:

```js
const get3DSecureError = ({ threeDSecureInfo: { status: message } = { status: null } } ) => (
  message && message.match(/authenticate_(attempt_)?successful/) ? null : CC_ERROR_MESSAGE
)
```

didierofrivia: @akostadinov JS could be elegant AF ;)

damianpm: Great suggestion!
I'll go with @guicassolato 's one, is more readable and will be easy to add types to the response in the future

guicassolato (approved):

PR #2577, opened 2021-08-25
Use our own node-sass version

hallelujah:
In order to build on ppc64le we are using our node-sass fork
See https://github.com/3scale/node-sass/pull/2

Also when upgrading to Rails 5.2.2 we will no more use node-sass but Dart sass

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,7 @@
     "moment": "^2.24.0",
     "moment-range": "^4.0.2",
     "moment-timezone": "^0.5.5",
-    "node-sass": "^4.9.4",
+    "node-sass": "git://github.com/3scale/node-sass#v4",
     "null-loader": "^3.0.0",
     "numeral": "^2.0.6",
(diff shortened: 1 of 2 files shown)


thomasmaas (approved):
Looks there's some movement in https://github.com/sass/node-sass/pull/3168 so might be able to use official build again in future?

guicassolato (approved):
I guess for ppc64 we'll also need to merge https://github.com/3scale/node-sass/pull/2, but pointing to our fork since now, as it already bumps the version to 4.14.1, and under our roof, right?

hallelujah:
@guicassolato we already use v4.14.1 We can point to this already!
https://github.com/3scale/porta/blob/06df579608ef9563d1b4ca64c9deaed3994f9476/yarn.lock#L8470

Then when the other PR is merged we will update it in another PR

hallelujah:
No need of it we managed to build on ppc64le with vanilla node-sass

PR #2582, opened 2021-08-31
[WIP] THREESCALE-6916 change to load_from_hash! method

kevprice83:
Fixes THREESCALE-6916 - See JIRA for details.

Summary of what this PR does:

- Uses the `load_from_hash!` method instead of `load_from_hash` so that jobs which have been removed can be deleted before injecting new jobs. `SPHINX_DELTA` & `SPHINX_INDEX` were both removed in tagged release 2.9 and caused issues reported in the JIRA.


kevprice83:
Removed WIP from title as I cannot see any existing tests for this module. I f tests are required here please point me in the right direction and I'll work on adding those.

guicassolato (approved):

PR #2583, opened 2021-09-01
Use gitlab fork of pry-byebug

hallelujah:
So we can use pry-remote again
To be tested though ;)


hallelujah:
Closing, I feel like it is better to use [pry_shell](https://github.com/meinac/pry-shell)

See https://www.youtube.com/watch?v=Lzs_PL_BySo

PR #2587, opened 2021-09-02
[gem] Update factory_bot_rails

hallelujah:
Upgrading gems are always good ;)

With the introduction of `FactoryBot.use_parent_strategy = true` by default,
all the related associations are now using `build` strategy  if we invoke `FactoryBot.build`

If we want to revert back to old behaviour (not advised though) we should set `FactoryBot.use_parent_strategy = false`


test/test_helpers/factory_bot.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
```
hallelujah, line 1: This file is to keep track of defaults in FacotryBot

PR #2588, opened 2021-09-03
Bugfix: create mapping rule fails because of pattern validation

damianpm:

Creating new mapping rule fails when the pattern validation fails.
The pattern is being validated by Rails, but the error is not passed to React.

This PR:

- Pass the error to React
- Shows the error message
- Update the helper text, it was misleading, suggesting the use of unsupported characters ("-")

https://issues.redhat.com/browse/THREESCALE-7510

**Verification steps** 

- Go to **Product > Integration > Mapping rules > Create Mapping rule**
- Create a Mapping rule with an invalid pattern: missing the initial "/" or including an unsupported character like "-" or "*"
- Create a mapping rule with a valid pattern

https://user-images.githubusercontent.com/13486237/132003834-e8ad4e35-cb9c-4dc2-9ee4-b33345ee88ae.mov

### DIFF
--- a/app/javascript/packs/new_mapping_rule.js
+++ b/app/javascript/packs/new_mapping_rule.js
@@ -12,13 +12,14 @@ document.addEventListener('DOMContentLoaded', () => {
     return
   }
 
-  const { url, httpMethods, topLevelMetrics, methods, isProxyProEnabled } = container.dataset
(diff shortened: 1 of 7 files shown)


akostadinov (approved):
Looks good to me. May need an additional look from somebody with better js experience.

hallelujah (approved):

PR #2590, opened 2021-09-06
moves "Add Backend" button from backends table into page header

josemigallas:

Moves "Add Backend" button to the page header.

Before:
Screenshot 2021-09-06 at 12 36 54

After:
Screenshot 2021-09-06 at 12 37 22

[THREESCALE-7519: Move "Add Backend" button from table to page header](https://issues.redhat.com/browse/THREESCALE-7519)

**Verification steps** 

Check button at `Product > Integation > Backends`

### DIFF
--- a/app/views/api/backend_usages/index.html.slim
+++ b/app/views/api/backend_usages/index.html.slim
@@ -1,6 +1,8 @@
 h1 Backends
 p.intro For a Product to work, it needs to have at least one Backend with a Private Base URL (your API). If multiple Backends are added to a Product, each Backend should have a unique Public Path.
 
+= pf_button_in_title 'Add Backend', new_admin_service_backend_usage_path(@service)
+
 
 table#backend_api_configs.data
   thead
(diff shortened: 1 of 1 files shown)


app/views/api/backend_usages/index.html.slim
```diff
@@ -1,6 +1,8 @@
 h1 Backends
 p.intro For a Product to work, it needs to have at least one Backend with a Private Base URL (your API). If multiple Backends are added to a Product, each Backend should have a unique Public Path.
 
+= pf_button_in_title 'Add Backend', new_admin_service_backend_usage_path(@service)
```
damianpm, line 4: I haven't seen this method before, but maybe you can use [this one](https://github.com/3scale/porta/blob/master/app/helpers/buttons_helper.rb#L150), we are already using it in several places

josemigallas: I created this method a while ago to encapsulate Patternfly's CTA. I think we should not mix "3scale style" with "PF style". From behind, it uses the method you're referring.

https://github.com/3scale/porta/search?q=pf_button_in_title

damianpm (approved):

PR #2591, opened 2021-09-06
Improve New Backend validations

josemigallas:

Changes `minimun_length` for `presence`.
Adds validation for `system_name`.
Updates tests.

[THREESCALE-7520: New Backend form "system name" validation not working](https://issues.redhat.com/browse/THREESCALE-7520)

**Verification steps** 

Product >Integration >Backends > Add Backend > Create new backend

### DIFF
--- a/app/javascript/src/BackendApis/components/NewBackendForm.jsx
+++ b/app/javascript/src/BackendApis/components/NewBackendForm.jsx
@@ -23,8 +23,9 @@ type Props = {
 }
 
 const VALIDATION_CONSTRAINTS = {
-  name: { length: { minimum: 1 } },
-  privateEndpoint: { url: { schemes: ['http', 'https', 'ws', 'wss'] } }
+  name: { presence: { allowEmpty: false } },
+  systemName: { format: { pattern: '^[a-z0-9-_]*$', flags: 'i' } },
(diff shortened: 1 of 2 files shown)


app/javascript/src/BackendApis/components/NewBackendForm.jsx
```diff
@@ -23,8 +23,9 @@ type Props = {
 }
 
 const VALIDATION_CONSTRAINTS = {
-  name: { length: { minimum: 1 } },
-  privateEndpoint: { url: { schemes: ['http', 'https', 'ws', 'wss'] } }
+  name: { presence: { allowEmpty: false } },
+  systemName: { format: { pattern: '^[a-z0-9-_]*$', flags: 'i' } },
```
damianpm, line 27: Are you sure this is the same pattern being validated in rails?

josemigallas: No.. but unless I'm mistaken it matches the same rules: only ASCII letters, numbers, dashes and underscores.

josemigallas: In the end I copied the same regex to avoid any possible issues.

damianpm (approved):

PR #2593, opened 2021-09-06
[THREESCALE-7475] Api docs calls won't destroy user session because of CSRF

jsmadis:

Adds CSRF token to the HTTP calls that are made through the 3scale API Docs.

https://issues.redhat.com/browse/THREESCALE-7475

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -16,7 +16,7 @@ class ApplicationController < ActionController::Base
 
   # Disable CSRF protection for non xml requests.
   skip_before_action :verify_authenticity_token, if: -> do
-    (params.key?(:provider_key) || params.key?(:access_token)) && request.format.xml?
+    (params.key?(:provider_key) || params.key?(:access_token)) && !request.format.xml?
   end
 
   before_action :set_timezone
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -16,7 +16,7 @@ class ApplicationController < ActionController::Base
 
   # Disable CSRF protection for requests to REST API.
   skip_before_action :verify_authenticity_token, if: -> do
-    (params.key?(:provider_key) || params.key?(:access_token)) && api_controller?
+    api_controller? && (params.key?(:provider_key) || params.key?(:access_token))
   end
 
   before_action :set_timezone
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -16,7 +16,7 @@ class ApplicationController < ActionController::Base
 
   # Disable CSRF protection for requests to REST API.
   skip_before_action :verify_authenticity_token, if: -> do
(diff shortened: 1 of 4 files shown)


akostadinov (dismissed):

hallelujah:
Maybe the calls should not return a `Set-Cookie` headers

That is a viable solution too.
I don't think any API request should use cookies/sessions as they have token to identify the user

jsmadis:
> I don't think any API request should use cookies/sessions as they have token to identify the user

@hallelujah, I agree, but that's the point of the other issue. We can postpone this PR, and I will try to fix our API request to ignore the cookies/sessions. Then we can see if it's a viable solution for this problem also.

hallelujah:
What is the other issue? I think (if not mistaken) we should use 
 this for all APIs https://api.rubyonrails.org/classes/ActionController/RequestForgeryProtection/ClassMethods.html#method-i-skip_forgery_protection

I wonder we do not have it though
https://github.com/3scale/porta/blob/b2fd9c02596ce109f17d692475384c4aae7259d2/app/controllers/admin/api/base_controller.rb

hallelujah:
So it is because we use a completely wrong logic

https://github.com/3scale/porta/blob/master/app/controllers/application_controller.rb#L18-L20

jsmadis:
Oh, I see. Thanks @hallelujah 

The fix is even easier then.

app/controllers/application_controller.rb
```diff
@@ -16,7 +16,7 @@ class ApplicationController < ActionController::Base
 
   # Disable CSRF protection for non xml requests.
   skip_before_action :verify_authenticity_token, if: -> do
-    (params.key?(:provider_key) || params.key?(:access_token)) && request.format.xml?
+    (params.key?(:provider_key) || params.key?(:access_token)) && !request.format.xml?
```
hallelujah, line 19:
```suggestion
    (params.key?(:provider_key) || params.key?(:access_token)) && is_api_controller?
```

Then create a method 
```
class ApplicationController < ActionController::Base
  def api_controller?
    false
  end
end

# And in other :real" API base controllers set it to true

File: app/controllers/application_controller.rb [hunk 2]
```diff
@@ -14,9 +14,9 @@ class ApplicationController < ActionController::Base
 
   protect_from_forgery with: :reset_session # See ActionController::RequestForgeryProtection for details
 
-  # Disable CSRF protection for non xml requests.
+  # Disable CSRF protection for requests to REST API.
   skip_before_action :verify_authenticity_token, if: -> do
-    (params.key?(:provider_key) || params.key?(:access_token)) && request.format.xml?
+    (params.key?(:provider_key) || params.key?(:access_token)) && api_controller?
```
akostadinov, line 19: probably worthless micro-optimization, but shouldn't we first check `api_controller?` and then the other checks?

akostadinov (approved):

hallelujah (approved):

PR #2596, opened 2021-09-07
Order all plans alphabetically by default

josemigallas:

Force all plans to be listed alphabetically by default.

Bug reported in:
- [THREESCALE-6875: [3scale][2.11][LO-prio] Improve list of Application plans in Application plans index page](https://issues.redhat.com/browse/THREESCALE-6875)
- [THREESCALE-6871: [3scale][2.11][LO-prio] Improve UI for changing Application plan for existing Application](https://issues.redhat.com/browse/THREESCALE-6871)

### DIFF
--- a/app/controllers/api/plans_base_controller.rb
+++ b/app/controllers/api/plans_base_controller.rb
@@ -44,7 +44,7 @@ class Api::PlansBaseController < Api::BaseController
   def find_plans
     search = ThreeScale::Search.new(params[:search] || params)
     @plans = collection.not_custom
-                       .order_by(params[:sort] || :name, params[:direction] || :asc)
+                       .order_by(params[:sort].presence || :name, params[:direction].presence || :asc)
                        .scope_search(search)
     @page_plans = @plans.paginate(pagination_params)
   end
--- a/app/controllers/api/plans_base_controller.rb
+++ b/app/controllers/api/plans_base_controller.rb
@@ -44,7 +44,7 @@ class Api::PlansBaseController < Api::BaseController
   def find_plans
     search = ThreeScale::Search.new(params[:search] || params)
     @plans = collection.not_custom
-                       .order_by(params[:sort], params[:direction])
+                       .order_by(params[:sort] || :name, params[:direction] || :asc)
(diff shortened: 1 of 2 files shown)


akostadinov (dismissed):

app/controllers/api/plans_base_controller.rb
```diff
@@ -44,7 +44,7 @@ def find_plan
   def find_plans
     search = ThreeScale::Search.new(params[:search] || params)
     @plans = collection.not_custom
-                       .order_by(params[:sort], params[:direction])
+                       .order_by(params[:sort] || :name, params[:direction] || :asc)
```
hallelujah, line 47:
```suggestion
                       .order_by(params[:sort].presence || :name, params[:direction].presence || :asc)
```

app/helpers/api/plans_helper.rb
```diff
@@ -19,7 +19,7 @@ def default_application_plan_data(service, plans)
   end
 
   def application_plans_data(plans)
-    plans.alphabetically
+    plans.order(name: :asc)
```
hallelujah, line 22: Why does not that works? https://github.com/3scale/porta/blob/b2fd9c02596ce109f17d692475384c4aae7259d2/app/models/plan.rb#L130

josemigallas: So the problem was I was passing 2 different collections through the same method. Once is an array and the other a ActiveRecord collection.
Moreover, `alphabetically` is redundant because plans are being ordered by the controller (since it depends on the query params)

app/helpers/api/plans_helper.rb
```diff
@@ -36,8 +31,7 @@ def application_plans_table_data(service_id:, page_plans:, plans_size:)
   end
 
   def application_plans_index_data(plans)
-    plans.alphabetically
-         .decorate
+    plans.decorate
```
josemigallas, line 34: No need to call `alphabetically` because plans are ordered in the controller.

hallelujah (approved):

akostadinov (approved):

PR #2597, opened 2021-09-07
🐛 Sort "Backends used by product" list alphabetically

josemigallas:

backends have to be sorted by name.

Screenshot 2021-09-07 at 14 42 38

Bug reported in: [THREESCALE-6866: [3scale][2.11][LO-prio] Improve list of Backends in Product overview page](https://issues.redhat.com/browse/THREESCALE-6866)

**Verification steps** 

Go to some product overview page and check the list.

### DIFF
--- a/app/decorators/service_decorator.rb
+++ b/app/decorators/service_decorator.rb
@@ -68,7 +68,7 @@ class ServiceDecorator < ApplicationDecorator
   end
 
   def backends_table_data
-    BackendApiDecorator.decorate_collection(backend_apis.sort_by(&:name))
+    BackendApiDecorator.decorate_collection(backend_apis.reorder(name: :asc))
                        .map(&:table_data)
                        .to_json
   end
--- a/app/decorators/service_decorator.rb
+++ b/app/decorators/service_decorator.rb
@@ -68,7 +68,7 @@ class ServiceDecorator < ApplicationDecorator
   end
 
   def backends_table_data
-    BackendApiDecorator.decorate_collection(backend_apis)
+    BackendApiDecorator.decorate_collection(backend_apis.sort_by(&:name))
                        .map(&:table_data)
                        .to_json
   end
--- a/features/api/overview.feature
+++ b/features/api/overview.feature
@@ -1,0 +1,16 @@
+Feature: Product > Overview page
+  Background:
+    Given a provider is logged in with a product "My Product"
+    And it uses the following backends:
+      | Name      | path |
+      | Backend 3 | /v3  |
+      | Backend 2 | /v2  |
+      | Backend 1 | /v1  |
+
+  @javascript
+  Scenario: List of backends used is ordered alphabetically
+    When I go to the overview page of product "My Product"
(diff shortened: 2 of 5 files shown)


app/decorators/service_decorator.rb
```diff
@@ -68,7 +68,7 @@ def new_application_data
   end
 
   def backends_table_data
-    BackendApiDecorator.decorate_collection(backend_apis)
+    BackendApiDecorator.decorate_collection(backend_apis.sort_by(&:name))
```
hallelujah, line 71: Is backend_apis an `Array` or an `ActiveRecord::Collection` ?

if the latter, then we should use 
```suggestion
    BackendApiDecorator.decorate_collection(backend_apis.desc(name: :asc))
```

josemigallas: @hallelujah Do you mean
```suggestion
    BackendApiDecorator.decorate_collection(backend_apis.order(name: :asc))
```

josemigallas: It is a `BackendApi::ActiveRecord_Associations_CollectionProxy` and it doesn't have that sort method.

hallelujah: I do not get why `order` does not work

```ruby
[2] pry(main)> service.backend_apis.order(name: :asc)                                                                                                                                               
  BackendApi Load (1.4ms)  SELECT `backend_apis`.* FROM `backend_apis` INNER JOIN `backend_api_configs` ON `backend_apis`.`id` = `backend_api_configs`.`backend_api_id` WHERE `backend_api_configs`.`service_id` = 2 ORDER BY `backend_api_configs`.`id` ASC, `backend_apis`.`name` ASC
=> [#<BackendApi:0x000000000cde9958
  id: 2,
  name: "API Backend",
  system_name: "api",
  description: "Backend of API",
  private_endpoint: "https://echo-api.3scale.net:443",
  account_id: 2,
  created_at: Tue, 20 Apr 2021 12:41:24 UTC +00:00,
  updated_at: Tue, 20 Apr 2021 12:41:24 UTC +00:00,
  tenant_id: nil,
  state: "published">]
```

hallelujah, line 71:
```suggestion
    BackendApiDecorator.decorate_collection(backend_apis.reorder(name: :asc))
```

Sorry that was the thing

hallelujah (changes requested):

app/decorators/service_decorator.rb
josemigallas: Done

thomasmaas (approved):
You're using `reorder` because our default ordering is different (I guess). Changing the default doesn't make sense?

josemigallas:
> You're using `reorder` because our default ordering is different (I guess). Changing the default doesn't make sense?

I'd rather set a specific order at the view layer, where we need rather than changing the default order and unexpectedly affect other views.

hallelujah (approved):

PR #2599, opened 2021-09-07
[THREESCALE-7084] 3scale API will not be authenticated by the user_cookie

jsmadis:

HTTP request to the 3scale API won't be authenticated via user_session inside cookies.

Main changes:

- Analytics (frontend) endpoint is moved from path `/stats/...` to `/stats/api/...`
- New controllers for the analytics were created. (Copy of the REST API ones).
- REST API endpoints are no longer authenticated with session/cookie.

https://issues.redhat.com/browse/THREESCALE-7084


thomasmaas:
This change means all users will have to manually authenticale calls by pasting personal token in api docs?

jsmadis:
@thomasmaas yes, your assumption is correct.

doc/active_docs/Analytics API.json
```diff
@@ -1,7 +1,7 @@
 {
   "apis": [
     {
-      "path": "/stats/applications/{application_id}/usage.{format}",
+      "path": "/admin/api/stats/applications/{application_id}/usage.{format}",
```
hallelujah, line 4: This is a breaking change!

I am not so sure we want that

app/javascript/src/Stats/buyer/stats_application.jsx
```diff
@@ -43,13 +43,13 @@ export class StatsApplicationChartManager extends StatsUsageChartManager {
   }
 
   _metricUrl (id) {
-    return `/stats/applications/${id}/summary.json?version=2.0`
+    return `/stats/api/applications/${id}/summary.json?version=2.0`
```
hallelujah, line 46: So this is a temporary measure. we would need to clean that by properly announcing a deprecation notice
On the API side

`stats/applications` => `admin/api/stats/applications`

I wonder if we could create the routes already

config/routes.rb
```diff
@@ -444,6 +444,29 @@ def call(env)
 
   # These are API routes, beware
   namespace :stats do
+    namespace :api do
+      # horrible copy paste, but we need to create separate endpoints
```
hallelujah, line 448: Especially it is not a "real" API but a JS endpoint

hallelujah (approved):

PR #2604, opened 2021-09-09
🐛 Fix Edit Redirect URL form in Application's show page

josemigallas:

The Edit button is supposed to load a template that wasn't updated during the Applications Controllers refactoring.

edit_url

[THREESCALE-7501: Unable to edit Redirect URL in Admin Portal](https://issues.redhat.com/browse/THREESCALE-7501)

**Verification steps** 

See JIRA


thomasmaas:
Kind of hard to see what the fix is due to een to slim in same commit. Needs a test?

josemigallas:
> Kind of hard to see what the fix is due to erb to slim in same commit. Needs a test?

OK, you're right. I'll change it back to .erb so that changes are more obvious.

jsmadis (approved):

PR #2606, opened 2021-09-10
Fix ParameterType "date"

josemigallas:

Cucumber is not able to suggest new step definitions (a very useful feature IMO) because there is a parameter type with a "catch-all" regexp. As a result, Cucumber will always suggest to use it:
Screenshot 2021-09-10 at 10 05 28


hallelujah (approved):
Good catch thank you!!!!

hallelujah (approved):

PR #2612, opened 2021-09-15
Simplify AJAX utility module

josemigallas:


akostadinov (approved):

app/javascript/src/utilities/ajax.js
```diff
@@ -2,22 +2,25 @@
 
 export type Method = 'GET' | 'POST' | 'DELETE'
 
-const ajax = (url: string, method: Method, body?: URLSearchParams): Promise<Response> => {
+type FetchFunction = (url: string, method: Method, body?: URLSearchParams) => Promise<Response>
```
damianpm, line 5: Why body is always `URLSearchParams` ?

josemigallas: Because that's what the API endpoints expect, same as method being only 3 verbs. We will extend these types as we need. This module is not supposed to be an all-purpose solution but rather a handy helper for the few requests we do from React.

damianpm: ok

damianpm (approved):

hallelujah (approved):
lGTM

PR #2615, opened 2021-09-16
Refactor utils ajax 2

josemigallas:
Follows up and fixes #2612

### DIFF
--- a/app/javascript/src/utilities/ajax.js
+++ b/app/javascript/src/utilities/ajax.js
@@ -20,7 +20,7 @@ const _ajax = (headers: { [key: string]: string }) => {
   }
 }
 
-const ajax: FetchFunction = _ajax({ 'ContentType': 'application/x-www-form-urlencoded; charset=UTF-8' })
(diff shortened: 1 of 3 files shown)


hallelujah (approved):

app/javascript/src/utilities/ajax.js
```diff
@@ -4,23 +4,23 @@ export type Method = 'GET' | 'POST' | 'DELETE'
 
 type FetchFunction = (url: string, method: Method, body?: URLSearchParams) => Promise<Response>
 
-const _ajax = (headers: Object) => {
+const { signal, abort }: AbortController = new AbortController()
```
damianpm, line 7: Is this being used anywhere, or it's for the future?

josemigallas: The idea is to allow cancelling any request from our React components. However, I've been playing with it in the New Application Form PR and it doesn't quite work so I'll let each component have their own `AbortController`.

akostadinov (approved):

PR #2617, opened 2021-09-16
update audited to 5.0.2

akostadinov:
This should preserve Rails lazy initialization order.
Also fixes deprecation warnings with Rails 6.1 and adds support for 7

supersedes #2616

### DIFF
--- a/config/initializers/audited_hacks.rb
+++ b/config/initializers/audited_hacks.rb
@@ -109,32 +109,28 @@ module AuditedHacks
       self.disable_auditing if Rails.env.test?
 
       include InstanceMethods
+      class << self
+        prepend ClassMethods
+      end
     end
 
     def synchronous
-      original = Thread.current[:audit_hacks_synchronous]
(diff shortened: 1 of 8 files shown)


akostadinov:
This is also related to https://issues.redhat.com/browse/THREESCALE-7571 as it seems like `index_audits_on_auditable_id_and_auditable_type_and_version` index is redundant now.

config/initializers/audited_hacks.rb
```diff
@@ -109,32 +109,28 @@ def audited(options = {})
...
+      class << self
+        prepend ClassMethods
+      end
     end
 
     def synchronous
-      original = Thread.current[:audit_hacks_synchronous]
-
-      Thread.current[:audit_hacks_synchronous] = true
+      original = Thread.current.thread_variable_get(:audit_hacks_synchronous)
```
hallelujah, line 118: I think we can use Fiber local variable

hallelujah (approved):

PR #2622, opened 2021-09-20
[WIP] tag browser tests

akostadinov:
Why do these tests pass with chrome but not with rack?


akostadinov:
superseded by simpler #2619

PR #2625, opened 2021-09-24
[THREESCALE-7605] Fix type of policies config returned by the proxy

jsmadis:

With the Policies config refactoring, we introduced a bug where the type of policies_config is a string instead of an array of hashes. This PR fixes that and keeps the same format as it was used before.

https://issues.redhat.com/browse/THREESCALE-7605

### DIFF
--- a/app/representers/proxy_representer.rb
+++ b/app/representers/proxy_representer.rb
@@ -48,7 +48,7 @@ class ProxyRepresenter < ThreeScale::Representer
   property :jwt_claim_with_client_id_type, if: ->(*) { oidc? }
 
   def policies_config
-    represented.policies_config.to_json
+    represented.policies_config.as_json
   end
 
   class JSON < ProxyRepresenter


app/representers/proxy_representer.rb
```diff
@@ -47,10 +47,6 @@ class ProxyRepresenter < ThreeScale::Representer
   property :jwt_claim_with_client_id, if: ->(*) { oidc? }
   property :jwt_claim_with_client_id_type, if: ->(*) { oidc? }
 
-  def policies_config
-    represented.policies_config.to_json
```
akostadinov, line 51: Would changing this to `represented.policies_config.as_json` work for both - JSON and XML?

jsmadis: Imho that was the issue, we returned the JSON but we need to keep it compatible with older versions where we return an array of hashes.

akostadinov: Also I think `spec/acceptance/api/proxy_spec.rb` needs to be amended to catch this issue properly. A quick run of the test file with `#as_json` still has it passed but it also passed with `#to_json` so I don't know whether the end result is precisely what we expect or not :)

akostadinov: `#as_json` should return an array of maps. That's why I suggest it - to work with both representers. An oversight on my side to use `#to_json`.

akostadinov:
```
PRY-SHELL
◈────────────────────────────────────────────────────────────────────────────────◈

Frame number: 0/80

From: /home/avalon/ws/repos/porta/app/representers/proxy_representer.rb:52 ProxyRepresenter#policies_config:

    50: def policies_config
    51:   binding.pry_shell
 => 52:   represented.policies_config.as_json
    53: end

[1] pry(#<ProxyRepresenter::JSON>)> represented.policies_config.as_json
=> [{"name"=>"apicast",
  "version"=>"builtin",
  "configuration"=>{},
  "enabled"=>true}]
[2] pry(#<ProxyRepresenter::JSON>)> 
```

akostadinov: Another good place to add a test for this might be `ProxiesControllerTest#test_show`. On neither place the returned data format is checked.

akostadinov:
Updated #2627 by adding a test for the expected behavior.

PR #2629, opened 2021-09-28
acts-as-taggable-on updated to 8.1.0

akostadinov:
Having `acts_as_taggable` in models caused Rails 5.1 deprecation with
4.0.0

> DEPRECATION WARNING: Passing a class to the `class_name` is deprecated and will raise an ArgumentError in Rails 5.2. It eagerloads more classes than necessary and potentially creates circular dependencies. Please pass the class name as a string: `has_many :tag_taggings, class_name: 'ActsAsTaggableOn::Tagging'`

Also new version adds support for Rails 5.2+


akostadinov:
superseded by #2660

PR #2630, opened 2021-09-29
add CircleCI caching for compiled assets

akostadinov:
When no changes to cached assets are needed, on my laptop `rake
assets:precompile assets:clean` takes around 12 seconds while without it
takes almost 2 minutes.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,6 +31,7 @@ store-log-artifacts: &store-log-artifacts
 flow-type-key: &flow-typed-cache-key flow-typed-{{ checksum "yarn.lock" }}-5
 npm-cache-key: &npm-cache-key node-v10-{{ checksum "yarn.lock" }}-5
 bundle-cache-key: &bundle-cache-key v2-bundler-gems-{{ .Environment.DB }}-{{ arch }}-{{ .Branch }}-{{ checksum "Gemfile.lock" }}
+assets-cache-key: &assets-cache-key v1-asset-cache-{{ checksum "tmp/assets_related_checksums" }}
 
 save-flow-typed-cache: &save-flow-typed-cache
   save_cache:
@@ -52,6 +53,24 @@ save-npm-cache: &save-npm-cache
 restore-flow-typed-cache: &restore-flow-typed-cache
   restore_cache:
     key: *flow-typed-cache-key
+
+save-assets-cache: &save-assets-cache
+  save_cache:
+    key: *assets-cache-key
+    paths:
+      - public/assets
+      - public/packs-test
+      - tmp/cache/assets
+      - tmp/cache/webpacker
+
+restore-assets-cache: &restore-assets-cache
+  restore_cache:
+    key: *assets-cache-key
+
+generate-assets-checksums: &generate-assets-checksums
+  run:
+    name: Generate assets precompilation checksum data
+    command: git ls-tree -r HEAD $(<.circleci/asset_paths) > tmp/assets_related_checksums
 
 use-example-config-files: &use-example-config-files
   run:
@@ -427,6 +446,8 @@ jobs:
     steps:
       - checkout
       - *attach-to-workspace
+      - *generate-assets-checksums
+      - *restore-assets-cache
       - run:
           name: Precompile assets
           command: |
@@ -434,6 +455,7 @@ jobs:
             bundle exec rake assets:precompile NODE_ENV=test RAILS_ENV=test
           environment:
             RAILS_GROUPS: assets
+      - *save-assets-cache
       - persist_to_workspace:
           root: .
           paths:
--- a/.circleci/asset_paths
+++ b/.circleci/asset_paths
@@ -1,0 +1,17 @@
+Gemfile.lock
+yarn.lock
+.babelrc
+.postcssrc.yml
+app/javascript
+app/assets
+app/lib/pdf/images
+cdn
+config/compass.rb
+config/webpack
+lib/developer_portal/app/assets
+lib/developer_portal/app/views/developer_portal/css
+lib/developer_portal/app/views/developer_portal/javascripts
+lib/developer_portal/app/views/developer_portal/images
+vendor/active-docs/app/assets
+vendor/active-docs/vendor/assets
+vendor/assets


.circleci/config.yml
```diff
@@ -427,13 +443,15 @@ jobs:
     steps:
       - checkout
       - *attach-to-workspace
+      - *restore-assets-cache
       - run:
           name: Precompile assets
           command: |
             cp config/examples/*.yml config/
-            bundle exec rake assets:precompile NODE_ENV=test RAILS_ENV=test
+            bundle exec rake assets:precompile assets:clean NODE_ENV=test RAILS_ENV=test
```
akostadinov, line 451: clean step doesn't add a significant amount of time, less than a second as far as I see locally. It should help avoid cache bloat over time and is needed because of line 70.

.circleci/config.yml
```diff
@@ -53,6 +53,22 @@ restore-flow-typed-cache: &restore-flow-typed-cache
...
+      - public/packs-test
+      - tmp/cache/assets
+      - tmp/cache/webpacker
+
+restore-assets-cache: &restore-assets-cache
+  restore_cache:
+    keys:
+      - v1-asset-cache-{{ .Branch }}-{{ .BuildNum }}
+      # fallback to using the latest cache if no exact match is found
+      - v1-asset-cache-
```
akostadinov, line 70: Reusing cache from other builds and pull requests. My understanding is that [cache is per repo](https://circleci.com/docs/2.0/caching/#caching-and-open-source), so cache generated in upstream repo will not be used for PRs submitted from forked repos as well forked repos will not share caches with each other. 

If sharing caches between PRs in same repo is undesirable, then we can remove this line and share only between same git revision.

.circleci/config.yml
```diff
@@ -53,6 +53,23 @@ restore-flow-typed-cache: &restore-flow-typed-cache
...
+    paths:
+      - public/assets
+      - public/packs-test
+      - tmp/cache/assets
+      - tmp/cache/webpacker
+
+restore-assets-cache: &restore-assets-cache
+  restore_cache:
+    keys:
+      - v1-asset-cache-{{ .Branch }}-{{ .BuildNum }}
```
akostadinov, line 68: In [documentation example](https://circleci.com/docs/2.0/caching/#full-example-of-saving-and-restoring-cache) `arch` is also included in cache key. I see no reason for this given that assets should be arch independent. Let me know if I'm missing something.

Also I included `BuildNum` to help pick latest version of cache on the next line instead of picking random one based on branch name alphanumeric order.

.circleci/config.yml
```diff
@@ -31,6 +31,7 @@ store-log-artifacts: &store-log-artifacts
 flow-type-key: &flow-typed-cache-key flow-typed-{{ checksum "yarn.lock" }}-5
 npm-cache-key: &npm-cache-key node-v10-{{ checksum "yarn.lock" }}-5
 bundle-cache-key: &bundle-cache-key v2-bundler-gems-{{ .Environment.DB }}-{{ arch }}-{{ .Branch }}-{{ checksum "Gemfile.lock" }}
+assets-cache-key: &assets-cache-key v1-asset-cache-{{ .Branch }}-{{ .Revision }}
```
akostadinov, line 34: In [documentation example](https://circleci.com/docs/2.0/caching/#full-example-of-saving-and-restoring-cache) `arch` is also included in cache key. I see no reason for this given that assets should be arch independent. Let me know if I'm missing something.

The idea is to cache not more often than per revision tested. Ideally we can figure out from sources whether anything changed but it seems too complicated. Most likely need to make sure that `yarn.lock` and `Gemfile.lock` didn't change as well no other js file in our repo.

Still doable but maybe not worth the effort. WDYT?

hallelujah: Well doing so does not really help much, does it? You are unlikely to run the same tests more than once for each revision ... Even for different type of DB, or you would wait for assets precompilation first and cache it for other pipelines can reuse it... But that means you will not gain anything for it, since you add 2' overhead before been able to run any pipeline instead of having the pipelines run in parallel

.circleci/config.yml
```diff
@@ -53,6 +54,23 @@ restore-flow-typed-cache: &restore-flow-typed-cache
...
     key: *flow-typed-cache-key
 
+save-assets-cache: &save-assets-cache
+  save_cache:
+    key: *assets-cache-key
+    paths:
+      - public/assets
+      - public/packs-test
+      - tmp/cache/assets
+      - tmp/cache/webpacker
```
hallelujah, line 64: `tmp` should not be cached ... it is a bad idea ;)
Generally (it might not happen in this case) the directory have the process PID in them so it might be at best case useless, at worse error prone

akostadinov: If these dirs are not cached, they are re-created. So we save time by caching them. It is also in the CircleCI example. This is project tmp dir, not `/tmp`.

.circleci/config.yml
hallelujah: Yes it is cache for repo

But instead you should not have fallbacks, can lead to some unexpected errors. I prefer not caching than caching badly

.circleci/config.yml
akostadinov: Yeah, we will need to compute for every workflow.

akostadinov: But it is about sharing between revisions. If we compute some checksum to know whether something is changing, we will not use `revision` but share across any revisions.

hallelujah: Yes but how likely are you going to test the same revision more than once?

.circleci/config.yml
hallelujah: I get that but I think it is riskier because of _random_ PID generation or something we do not really know how sprockets, webpacker do while generating those directories

.circleci/config.yml
hallelujah: Also you can completely base your cache key on `app/assets` and `app/javascript` and other directories
The issue is when it comes to vendored assets ...

.circleci/config.yml
akostadinov: I removed catch-all key and left the branch catch latest key so we restore for the same branch at least. Is this ok with you? 

Even if assets changed since cache was created (e.g. a new commit was added to branch) things should still work because `assets:precompile` is still called. Let me know if you want me to remove also branch fallback.

.circleci/config.yml
akostadinov: I think many pull requests do not touch assets so it would be often the case when cache is valid. Also even partial cache would have some usefulness. But anyway, I agree that unlikely the benefit will be worth the additional complexity. That's why I left it like it is.

.circleci/config.yml
akostadinov: `assets:precompile` should handle whatever is in the cache. This is what other people use. Hope it will be stable. I also don't like that these tools use and require caching of multiple directories that are not even documented (or I didn't see them documented).

akostadinov: btw I also see people caching bootsnap https://github.com/jetthoughts/jt_tools/blob/master/lib/install/.circleci/config.yml I wonder if that could benefit us.

.circleci/config.yml
hallelujah: Well, in the contrary many PR touch assets :)
Look at how many PR @josemigallas or @damianpm are doing, well in general the @3scale/system-ux is doing many many PR here.

hallelujah (approved):

.circleci/config.yml
```diff
@@ -31,6 +31,7 @@ store-log-artifacts: &store-log-artifacts
 flow-type-key: &flow-typed-cache-key flow-typed-{{ checksum "yarn.lock" }}-5
 npm-cache-key: &npm-cache-key node-v10-{{ checksum "yarn.lock" }}-5
 bundle-cache-key: &bundle-cache-key v2-bundler-gems-{{ .Environment.DB }}-{{ arch }}-{{ .Branch }}-{{ checksum "Gemfile.lock" }}
+assets-cache-key: &assets-cache-key v1-asset-cache-{{ checksum "tmp/assets_related_checksums" }}
```
Bartuz, line 34: ohhh, I love it so much! Thank you <3

PR #2632, opened 2021-09-29
Cucumber tests won't use zync even if it's configured

jsmadis:

Disables zync in the cucumber tests.
When you have the zync configure locally (`ZYNC_ENDPOINT` env exists) the cucumber tests will fail.

### DIFF
--- a/config/examples/zync.yml
+++ b/config/examples/zync.yml
@@ -12,6 +12,7 @@ development:
 
 test:
   <<: *base
+  endpoint:
 
 production:
   <<: *base


akostadinov (dismissed):

akostadinov:
Does this only affect cucumber?

jsmadis:
@akostadinov AFAIK yes, other tests are running normally.

hallelujah (changes requested):
What is it trying to do? What is the error you get? It is probably you do not have the same configuration for your running zync and for your test. I am really doubtful of code change. This smells a bit :)

I think it is probably preferable that you set this to `endpoint:` for the `test` environment in your `config/zync.yml` This is a file you can customize for your own purpose.

https://github.com/3scale/porta/blob/master/config/examples/zync.yml

akostadinov:
I think that product configuration files should not affect tests as much as this is possible. I don't know if this is the best fix, just telling my opinion in general.

hallelujah:
@akostadinov The files are part of the project, we test them as default values especially when the configuration has `test`  namespace. Otherwise what is the point to have it? 
If we want to test something in particular like a real zync integration, then you should have zync running in order to do so.

Disabling zync everywhere in the test is making the test independent from the configuration files which is not what we want. We want sane defaults. A way to circumvent this is to unset your ZYNC_URL environment before running the test.

Below is to demonstrate that we really rely on those files, they are not just configuration, they are sane defaults.

https://github.com/3scale/porta/blob/2f9694005e270063bdf80a79915ff671ee9c2c98/test/unit/backend/storage_test.rb#L3

akostadinov:
```yaml
test:
  <<:*base
  endpoint:
```

makes sense to me, thank you

akostadinov:
@jsmadis , what do you think about changing config?

akostadinov (approved):

hallelujah (approved):

PR #2633, opened 2021-10-01
Improve Products index

josemigallas:

The page is visually identical but this PR introduces a series of improvements:

* Extracts ruby logic into presenter and decorator
* Refactors JS: typings, proper extension JSX, extracts `ToolbarSearch` for reusability
* Simplifies CSS
* Adds some tests

Part 1/2 of [THREESCALE-7646: Cleanup/refactor Products and Backends index logic](https://issues.redhat.com/browse/THREESCALE-7646)

**Before:**
Screenshot 2021-10-01 at 12 11 56

**After:**
Screenshot 2021-10-01 at 12 11 31

**Search input:**
Imagine the popover message reads: `To search, type at least 3 characters.`
https://user-images.githubusercontent.com/11672286/136042060-86e2937a-014a-48e1-99aa-ca504caafdac.mp4


Martouta:
Guuuuys, my account is still linked to 3scale in CircleCI 😂 

image

image

app/javascript/src/Common/components/ToolbarSearch.jsx
```diff
@@ -0,0 +1,98 @@
...
+          type="search"
+          aria-label="Search"
+          value={searchText}
+          onChange={setSearchText}
+          autoComplete="off"
+        />
+        <Button
+          variant={ButtonVariant.control}
+          aria-label="search button for search input"
+          isDisabled={searchText.length > 0 && searchText.length < 3}
```
damianpm, line 86:
```suggestion
          isDisabled={searchText.length < 3}
```
If you disable only when `searchText.length > 0` the button is clickable with empty input. 

And it's a bit strange that the button is clickable when first rendered, but becomes disabled after the first character, then is clickable again with 3 chars.
I think it's better if first rendered as disabled, and only enabled when reaching 3 chars.

josemigallas: However, we need to allow users to submit an "empty" search in order to show all results. If we disable the button with 0 chars, then they will be forced to refresh the page.

josemigallas: We could disable it as long as the search is invalid, for instance when searching the same thing twice. Thoughts @cntlsn ?

app/controllers/api/services_controller.rb
```diff
@@ -159,4 +154,8 @@ def authorize_action
   def can_create?
     can? :create, Service
   end
+
+  def presenter
+    @presenter ||= Api::ServicesIndexPresenter.new(current_user: current_user, params: params)
+  end
```
akostadinov, line 160: Where is this method used?

app/javascript/src/Common/components/ToolbarSearch.jsx
damianpm: > However, we need to allow users to submit an "empty" search in order to show all results. If we disable the button with 0 chars, then they will be forced to refresh the page.

How is that? Show all results (index page) is a user triggered action?

app/controllers/api/services_controller.rb
josemigallas: It's used in the view: https://github.com/3scale/porta/blob/2cb37251a80941ba29d2a45fecb116431f8a96c0/app/views/api/services/index.html.slim#L1

app/javascript/src/Common/components/ToolbarSearch.jsx
josemigallas: Resolving as we've discussed it elsewhere: button will be always enabled and it will show a popover with the message "needs 3 chars to search".

akostadinov (approved):

PR #2635, opened 2021-10-01
refactor/cleanup backend apis index

josemigallas:

Same as #2633 but for Backend APIs index.

* Extracts ruby logic into presenter and decorator
* Simplifies CSS
* Adds some tests

Part 2/2 of [THREESCALE-7646: Cleanup/refactor Products and Backends index logic](https://issues.redhat.com/browse/THREESCALE-7646)

**Before:**
Screenshot 2021-10-01 at 18 35 18

**After:**

Screenshot 2021-10-01 at 18 33 21


app/presenters/provider/admin/backend_apis_index_presenter.rb
```diff
@@ -0,0 +1,34 @@
...
+    {
+      'new-backend-path': new_provider_admin_backend_api_path,
+      backends: page_backend_apis.decorate.map(&:index_page_data).to_json,
+      'backends-count': page_backend_apis.total_entries.to_json
+    }
+  end
+
+  protected
+
+  # TODO: is .order needed?
```
akostadinov, line 24: technically needed or is the question about whether we want it?

app/presenters/provider/admin/backend_apis_index_presenter.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+class Provider::Admin::BackendApisIndexPresenter
+  include System::UrlHelpers.system_url_helpers
+
+  def initialize(current_account:, params: {})
+    @current_account = current_account
+    @pagination_params = { page: params[:page] || 1, per_page: params[:per_page] || 20 }
+    @search = ThreeScale::Search.new(params[:search] || params)
```
akostadinov, line 9: This seems a little bit vague. What kind of data we allow in API documentation? Can we be a little be more strict in setting search params?

app/presenters/provider/admin/backend_apis_index_presenter.rb
josemigallas: Technically speaking.

app/presenters/provider/admin/backend_apis_index_presenter.rb
akostadinov: If `:search` param is not passed, then whole `params` is being used. Shouldn't use provide `:search` param in call? Why fallback? And if we fall back, shouldn't we at least remove other known parameters like `:page` and `:per_page`.

I don't know how things work so excuse me if what I ask is obvious already.

app/presenters/provider/admin/backend_apis_index_presenter.rb
akostadinov: If we want them in this order, then I guess it is needed. My understanding is that we can't guarantee api columns are stored in the order of `updated_at`.

josemigallas: I just tested it locally and this is the query for `current_account.backend_apis.reorder(updated_at: :asc)`:
```sql
SELECT `backend_apis`.* FROM `backend_apis`
WHERE `backend_apis`.`account_id` = 2
ORDER BY `backend_apis`.`updated_at` ASC
```
Which I guess is correct since ORDER BY looks correct.

damianpm (approved):
JS side, LGTM

app/presenters/provider/admin/backend_apis_index_presenter.rb
akostadinov: Didn't you need descending order? i.e. last updated to show up first?

PR #2646, opened 2021-10-11
THREESCALE-7194 Fixed New Backend API internal error

nidhi-soni1104:

Fixed Internal error when creating new Backend.

**issue1**: https://issues.redhat.com/browse/THREESCALE-7194

**issue2**: https://issues.redhat.com/browse/THREESCALE-7479
 
**issue3:** https://issues.redhat.com/browse/THREESCALE-7523

### DIFF
--- a/app/helpers/vertical_nav_helper.rb
+++ b/app/helpers/vertical_nav_helper.rb
@@ -202,6 +202,7 @@ module VerticalNavHelper
   def service_nav_sections
     sections = []
     return sections unless @service
+
     sections << {id: :overview,      title: 'Overview',      path: admin_service_path(@service)} if can? :manage, :plans
     sections << {id: :monitoring,    title: 'Analytics',     items: service_analytics}           if can? :manage, :monitoring
     sections << {id: :applications,  title: 'Applications',  items: service_applications}        if (can? :manage, :plans) || (can? :manage, :applications)
@@ -258,6 +259,7 @@ module VerticalNavHelper
   def backend_api_nav_sections
     sections = []
     return sections unless @backend_api.persisted?
+
     sections << {id: :overview,         title: 'Overview',           path: provider_admin_backend_api_path(@backend_api)}
     sections << {id: :monitoring,       title: 'Analytics',          path: provider_admin_backend_api_stats_usage_path(@backend_api)} if can? :manage, :monitoring
     sections << {id: :methods_metrics,  title: 'Methods & Metrics',  path: provider_admin_backend_api_metrics_path(@backend_api)}
--- a/app/helpers/vertical_nav_helper.rb
+++ b/app/helpers/vertical_nav_helper.rb
@@ -257,7 +257,7 @@ module VerticalNavHelper
   # Backend APIs
   def backend_api_nav_sections
     sections = []
-    return sections unless @backend_api
+    return sections unless @backend_api.persisted?
     sections << {id: :overview,         title: 'Overview',           path: provider_admin_backend_api_path(@backend_api)}
     sections << {id: :monitoring,       title: 'Analytics',          path: provider_admin_backend_api_stats_usage_path(@backend_api)} if can? :manage, :monitoring
     sections << {id: :methods_metrics,  title: 'Methods & Metrics',  path: provider_admin_backend_api_metrics_path(@backend_api)}
--- a/app/helpers/vertical_nav_helper.rb
+++ b/app/helpers/vertical_nav_helper.rb
@@ -258,7 +258,7 @@ module VerticalNavHelper
   # Backend APIs
   def backend_api_nav_sections
     sections = []
-    return sections unless @backend_api.persisted?
+    return sections unless @backend_api&.persisted?
 
     sections << {id: :overview,         title: 'Overview',           path: provider_admin_backend_api_path(@backend_api)}
     sections << {id: :monitoring,       title: 'Analytics',          path: provider_admin_backend_api_stats_usage_path(@backend_api)} if can? :manage, :monitoring
--- a/test/helpers/vertical_nav_helper_test.rb
+++ b/test/helpers/vertical_nav_helper_test.rb
@@ -9,15 +9,19 @@ class VerticalNavHelperTest < ActionView::TestCase
 
     # if permitted
     stubs(can?: true)
-    assert_equal(backend_api_nav_sections.pluck(:title), ["Overview", "Analytics", "Methods & Metrics", "Mapping Rules"])
+    assert_equal(["Overview", "Analytics", "Methods & Metrics", "Mapping Rules"], backend_api_nav_sections.pluck(:title))
 
     # if not permitted
     stubs(can?: false)
-    assert_equal(backend_api_nav_sections.pluck(:title), ["Overview", "Methods & Metrics", "Mapping Rules"])
+    assert_equal(["Overview", "Methods & Metrics", "Mapping Rules"], backend_api_nav_sections.pluck(:title))
 
     # When backend_api is not persisted
     @backend_api = BackendApi.new
-    assert_equal(backend_api_nav_sections.pluck(:title), [])
+    assert_equal([], backend_api_nav_sections.pluck(:title))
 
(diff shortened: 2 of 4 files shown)


app/helpers/vertical_nav_helper.rb
```diff
@@ -257,7 +258,8 @@ def service_integration_items
   # Backend APIs
   def backend_api_nav_sections
     sections = []
-    return sections unless @backend_api
+    return sections unless @backend_api.persisted?
```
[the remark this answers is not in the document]
hallelujah: Could you create one then? Or do you think it is unecessary

This actually really solved something, we should at least add a test to make it clear.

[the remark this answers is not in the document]
damianpm: @nidhi-soni1104 
Now the Vertical menu is not crashing, but you are returning an empty array, then there is no menu at all:

damianpm (changes requested):

nidhi-soni1104:
@damianpm This is what expected from code, not to show anything in the nav bar if backend api fails.

damianpm:
> @damianpm This is what expected from code, not to show anything in the nav bar if backend api fails.
> 
> @3scale/system-ux let me know your feedback on this.

@nidhi-soni1104 @3scale/system-ux 
My point is:

- When you fill the "create new" form, there is no navbar.
- If creation succeeds, then you are redirected to the show page, and the menu is shown
- If creation fails, some errors are shown, and you are shown again the "create new" form

I think, if creation fails and you are shown again the same form, it should look the same (no navbar in both cases):

nidhi-soni1104:
@hallelujah @josemigallas Added test case for the method, and waiting for UX team to confirm UI part.

test/helpers/vertical_nav_helper_test.rb
```diff
@@ -0,0 +1,18 @@
...
+
+require 'test_helper'
+
+class VerticalNavHelperTest < ActionView::TestCase
+
+  test '#backend_api_nav_sections' do
+    @backend_api = FactoryBot.create(:backend_api)
+
+    # if permitted
+    stubs(can?: true)
```
hallelujah, line 11: This stub is not bad but it feels a bit off to me. So just asking, this depends on what you are testing here. Does it matter to know what are the exact permissions or not? If not then leave it. If yes, you will need to include cancan controller additions and helpers and define current_user and the correct user.

test/helpers/vertical_nav_helper_test.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class VerticalNavHelperTest < ActionView::TestCase
+
+  test '#backend_api_nav_sections' do
```
hallelujah, line 7: What does happen if: 

1. `@backend_api = nil`
2. `@backend_api.persisted? == false`

hallelujah (changes requested):
I think we lack some tests, not all the cases are covered

app/helpers/vertical_nav_helper.rb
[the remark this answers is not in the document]
josemigallas: The empty vertical menu might have something to do with the `activate_menu` method?. For actoin `#new` it calls `:backend_apis` but for `#show` it is https://github.com/3scale/porta/blob/ff78ef29249d2081983efa909b573a33ef3d6b50/app/controllers/provider/admin/backend_apis_controller.rb#L9

test/helpers/vertical_nav_helper_test.rb
nidhi-soni1104: yes permission is not needed here

test/helpers/vertical_nav_helper_test.rb
nidhi-soni1104: correct this test is missing and I will cover this.

app/helpers/vertical_nav_helper.rb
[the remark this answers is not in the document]
damianpm: Yes, I think if you add `activate_menu :backend_apis` in the create method, it will fix it
https://github.com/3scale/porta/blob/ff78ef29249d2081983efa909b573a33ef3d6b50/app/controllers/provider/admin/backend_apis_controller.rb#L22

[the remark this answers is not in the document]
nidhi-soni1104: @josemigallas @damianpm Let me check this.

damianpm:
@nidhi-soni1104 I've created a PR to mark as required the mandatory fields
https://github.com/3scale/porta/pull/2725

app/helpers/vertical_nav_helper.rb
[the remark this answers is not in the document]
nidhi-soni1104: I checked adding this to create action and it removed the vertical menu when create fails and I think it looks good. 

I am adding a new commit adding `activate_menu :backend_api` in create method,

test/helpers/vertical_nav_helper_test.rb
nidhi-soni1104: I don't think it will be nil ever, because the side nav bar open when the api creation fails which is case of falseor else created successfully,
 
if @backend_api is nil then persisted? method will also fails and give nil error.

hallelujah: Yes, exactly. This is unit test, meaning it is isolated from the context. So your argument is not valid in this case ;)

Even if the scenario described does not happen now, we introduce a potential error leading to a 500 to the customer :) You do not know the future, but you know that this case might happen. Again not in our case but out of context it does. So I do not think we want to take that risk.

FYI do not mistake unit testing (without context) and integration testing (with context).

test/helpers/vertical_nav_helper_test.rb
```diff
@@ -0,0 +1,23 @@
...
+    stubs(can?: true)
+    assert_equal(backend_api_nav_sections.pluck(:title), ["Overview", "Analytics", "Methods & Metrics", "Mapping Rules"])
+
+    # if not permitted
+    stubs(can?: false)
+    assert_equal(backend_api_nav_sections.pluck(:title), ["Overview", "Methods & Metrics", "Mapping Rules"])
+
+    # When backend_api is not persisted
+    @backend_api = BackendApi.new
+    assert_equal(backend_api_nav_sections.pluck(:title), [])
```
hallelujah, line 20:
```suggestion
    @backend_api = BackendApi.new
    assert_equal([], backend_api_nav_sections.pluck(:title))
    
    # What about this test?
    @backend_api = nil
    assert_equal([], backend_api_nav_sections)
```
Please fix above too, we always prefer expected vs actual in test unit
```ruby
assert_equal expected, actual
``
![image](https://user-images.githubusercontent.com/64276/142437407-54964f34-c592-4e78-96c1-2cf9b73ed8d6.png)

FYI, in RSpec we do the inverse:

```ruby
expect(actual).to eq expected
``

**Review by hallelujah — CHANGES_REQUESTED** [review]

File: test/helpers/vertical_nav_helper_test.rb (hunk 4 above)
Comment (nidhi-soni1104) [reply test/helpers/vertical_nav_helper_test.rb]: Changes done

File: test/helpers/vertical_nav_helper_test.rb [hunk 5]
```diff
@@ -0,0 +1,27 @@
...
+    assert_equal(["Overview", "Methods & Metrics", "Mapping Rules"], backend_api_nav_sections.pluck(:title))
+
+    # When backend_api is not persisted
+    @backend_api = BackendApi.new
+    assert_equal([], backend_api_nav_sections.pluck(:title))
+
+
+    # When backend_api is nil
+    @backend_api = nil
+    assert_equal([], backend_api_nav_sections.pluck(:title))
```
hallelujah, line 25: You do not have to use pluck(:title) when you know it is an empty Hash btw ;)
That is fine for now

hallelujah (approved):

damianpm (approved):
LGTM

PR #2647, opened 2021-10-12
[Rails upgrade 5.1] Fixes issues with EventStore::Event

jsmadis:

Fixes issues with`EventStore::Event` when rails are upgraded to 5.1.7.

### DIFF
--- a/app/lib/event_store/event.rb
+++ b/app/lib/event_store/event.rb
@@ -35,7 +35,7 @@ module EventStore
       end
 
       def load(str)
-        deserialize([YAML.load(str)]).first.symbolize_keys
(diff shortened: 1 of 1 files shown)


jsmadis:
@akostadinov 

> Do you have any idea though why it was not a problem with Rails 5.0 but became a problem with 5.1? 

Honestly, I don't know. From the code, I can tell that ActiveRecord is calling `deserialize`

```
      def deserialize(value)
        if default_value?(value)
          value
        else
          coder.load(super)
        end
      end
```
and then the `default_value?` is called

```
        def default_value?(value)
          value == coder.load(nil)
        end
```
And here is the issue when the `nil` is passed to the `load` method.

jsmadis:
Found other root causes.

PR #2653, opened 2021-10-15
✅ Remove render's nothing option

josemigallas:
`render nothing: true` is no longer available. We use `head` instead.
```
DEPRECATION WARNING: `:nothing` option is deprecated and will be removed in Rails 5.1. Use `head`
method to respond with empty response body. (called from destroy at 
/opt/app-root/src/project/app/controllers/admin/api/web_hooks_failures_controller.rb:39)

DEPRECATION WARNING: `:nothing` option is deprecated and will be removed in Rails 5.1. Use `head`
method to respond with empty response body. (called from destroy at 
/opt/app-root/src/project/app/controllers/admin/api/web_hooks_failures_controller.rb:39)
```

### DIFF
--- a/app/controllers/frontend_controller.rb
+++ b/app/controllers/frontend_controller.rb
@@ -48,7 +48,7 @@ class FrontendController < ApplicationController
   end
 
   def do_nothing_if_head
-    head :success if request.head?
+    head :ok if request.head?
   end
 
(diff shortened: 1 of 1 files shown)


akostadinov:
This is awesome, could you submit to master branch?

PR #2654, opened 2021-10-15
✅ Replaces deprecated ActiveModel::Errors#get

josemigallas:
Fixes deprecation:
```
Onboarding::ApiFormTest#test_invalid_backend DEPRECATION WARNING: ActiveModel::Errors#get is
deprecated and will be removed in Rails 5.1. To achieve the same use model.errors[:private_endpoint].
(called from block in validate_each at /opt/app-root/src/project/app/validators/nested_association_validator.rb:12)
```


akostadinov:
Nice! Same here. Could you submit to master?

PR #2660, opened 2021-10-19
acts-as-taggable-on updated to 8.1.0

akostadinov:
Having `acts_as_taggable` in models caused Rails 5.1 deprecation with
4.0.0

> DEPRECATION WARNING: Passing a class to the `class_name` is deprecated and will raise an ArgumentError in Rails 5.2. It eagerloads more classes than necessary and potentially creates circular dependencies. Please pass the class name as a string: `has_many :tag_taggings, class_name: 'ActsAsTaggableOn::Tagging'`

Also new version adds support for Rails 5.2+


db/migrate/20211013201609_add_missing_unique_indices.acts_as_taggable_on_engine.rb
```diff
@@ -0,0 +1,32 @@
...
+  disable_ddl_transaction! if System::Database.postgres?
+
+  def self.up
+    add_index ActsAsTaggableOn.tags_table, [:name, :tenant_id], unique: true, name: 'index_tags_on_name', **index_options
+
+    remove_index ActsAsTaggableOn.taggings_table, :tag_id if index_exists?(ActsAsTaggableOn.taggings_table, :tag_id)
+    remove_index ActsAsTaggableOn.taggings_table, name: 'taggings_taggable_context_idx' if index_exists?(ActsAsTaggableOn.taggings_table, :taggings_taggable_context_idx)
+    add_index ActsAsTaggableOn.taggings_table,
+              [:tag_id, :taggable_id, :taggable_type, :context, :tagger_id, :tagger_type],
+              unique: true, name: 'taggings_idx', **index_options
```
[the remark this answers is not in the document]
akostadinov: we delay this decision with https://issues.redhat.com/browse/THREESCALE-7898

app/lib/three_scale/semantic_form_builder.rb
```diff
@@ -31,7 +31,10 @@ def text_field_with_errors( method, opts = {})
     # Get a column object for a specified attribute method - if possible.
     #
     def column_for(method) #:nodoc:
-      @object.column_for_attribute(method) if @object.respond_to?(:column_for_attribute) && @object.try(:has_attribute?, method)
+      # our old formtastic expects nil as in Rails 4, not a null object
+      # https://github.com/rails/rails/pull/15878
+      column = super
+      ActiveRecord::ConnectionAdapters::NullColumn === column ? nil : column
```
hallelujah, line 37: Would be good to have a test so when we upgrade formtastic it will break and then make the appropriate changes here ;)

db/migrate/20211117094501_add_missing_unique_indices.acts_as_taggable_on_engine.rb
```diff
@@ -0,0 +1,33 @@
...
+else
+  class AddMissingUniqueIndices < ActiveRecord::Migration; end
+end
+AddMissingUniqueIndices.class_eval do
+  disable_ddl_transaction! if System::Database.postgres?
+
+  def self.up
+    add_index ActsAsTaggableOn.tags_table, [:name, :tenant_id], unique: true, name: 'index_tags_on_name', **index_options
+
+    # see https://issues.redhat.com/browse/THREESCALE-7898
```
hallelujah, line 13: 👍

db/migrate/20211117094502_add_missing_taggable_index.acts_as_taggable_on_engine.rb
```diff
@@ -0,0 +1,13 @@
+# This migration comes from acts_as_taggable_on_engine (originally 4)
+class AddMissingTaggableIndex < ActiveRecord::Migration[5.0]; end
+AddMissingTaggableIndex.class_eval do
+  # disable_ddl_transaction! if System::Database.postgres?
+
+  def self.up
+    rename_index ActsAsTaggableOn.taggings_table, 'index_taggings_on_taggable_id_and_taggable_type_and_context', 'taggings_taggable_context_idx'
```
hallelujah, line 7: So this rename is to be inline with the upstream naming right?

db/migrate/20211117094503_change_collation_for_tag_names.acts_as_taggable_on_engine.rb
```diff
@@ -0,0 +1,6 @@
+# This migration comes from acts_as_taggable_on_engine (originally 5)
+# This migration is added to circumvent issue #623 and have special characters
+# work properly
+class ChangeCollationForTagNames < ActiveRecord::Migration[4.2]
```
hallelujah, line 4: Should we add this file in our project then? It does nothing and in few years later I will be a bit confused

hallelujah (changes requested):

db/migrate/20211117094502_add_missing_taggable_index.acts_as_taggable_on_engine.rb
akostadinov: yes

db/migrate/20211117094503_change_collation_for_tag_names.acts_as_taggable_on_engine.rb
akostadinov: The point is to avoid migration to be auto-generated by the gem going forward.

app/lib/three_scale/semantic_form_builder.rb
akostadinov: see `#test_column_for_override_needed`

db/migrate/20211117094503_change_collation_for_tag_names.acts_as_taggable_on_engine.rb
hallelujah: How could it be autogenerated? You have to run a rails command no? And also we review so we should just discard it. It does not make sense to commit empty migration.

akostadinov: When we upgrade the gem and run generation of migrations, we will always have to look at it and discard it as unnecessary. So my intention is to just avoid having to look at it going forward. Is there any other way to achieve this?

akostadinov: I mean, when one does:
```
rake acts_as_taggable_on_engine:install:migrations
```

Then missing migrations are generated. And when this class is missing, then it will be generated. So one will have to investigate whether it is a new one or old one and whether it applies. Maybe not a huge task but I wanted to avoid this. WDYT?

hallelujah: I think it is unnecessary, you do not want to commit unnecessary code
We will anyway investigate if we need to run this command again in the future

akostadinov: We don't need this migration because it is a noop for our database state. So we will never need it.

akostadinov:
Added a test for ignoring `#tag_list` in `#as_json` as codecov rightfully complained. Things should be good to go now.

Gemfile
```diff
@@ -92,7 +92,7 @@ gem 'mimemagic', '~> 0.3.10'
 gem 'nokogiri', '~> 1.10.10'
 gem 'secure_headers', '~> 6.3.0'
 
-gem 'acts-as-taggable-on', '~> 4.0'
+gem 'acts-as-taggable-on', '~> 8.0'
```
josemigallas, line 95: Any reason why not 8.1 here?
```suggestion
gem 'acts-as-taggable-on', '~> 8.1'
```

app/models/concerns/tagging.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Tagging
```
josemigallas, line 3: Purely cosmetic suggestion
```suggestion
module Taggable
```

test/unit/cms/page_test.rb
```diff
@@ -86,4 +86,15 @@ def setup
     assert ! page.searchable?
   end
 
+  test "tags N+1 queries for tags" do
+    page = FactoryBot.create(:cms_file)
```
josemigallas, line 90:
```suggestion
    page = FactoryBot.create(:cms_page)
```

test/unit/cms/page_test.rb
```diff
@@ -86,4 +86,15 @@ def setup
...
+    page.expects(:tag_list).never
+    page.as_json
+  end
+
+  test "as_json returns tag_list when requested" do
+    page = FactoryBot.create(:cms_file)
```
josemigallas, line 96:
```suggestion
    page = FactoryBot.create(:cms_page)
```

josemigallas (changes requested):
Might this be a typo?

Gemfile
akostadinov: it doesn't matter, it picked up 8.1 anyway because the matcher means latest 8.x

test/unit/cms/page_test.rb
akostadinov: Good catch!

app/models/concerns/tagging.rb
akostadinov: We can discuss this. "Concern tagging" sounded a little bit more natural than "concern taggable". Also it is concerning the taggings - or rather avoid obtaining the taggings.

But I don't know, we can change. It's not important.

hallelujah (approved):

PR #2664, opened 2021-10-25
Remove some old unused assets

josemigallas:
I noticed we upgraded jquery to `1.7.1 -> 3.5.0` in dev portal but we still have the older one: https://github.com/3scale/porta/commit/6017b4a916df3b027601a4c455124133b2b4acda
There are other assets that apparently are not used anymore


akostadinov:
Did we get any feedback from production? Did we ask in the first place? :)

PR #2667, opened 2021-10-25
[api] Add CMS API to apidocs

hallelujah:
In order to prepare for support the CMS API

image


akostadinov (approved):
So we decided to support CMS api? Cool.

btw [sour](https://github.com/HakubJozak/sour) seems last touched 2013. [source2swagger](https://github.com/solso/source2swagger) last touched 2015. I wonder if we need to check for a replacement. Or maybe these gems work well and don't need fixing.

PR #2672, opened 2021-10-26
[email-async] Generate email async for lost password token

hallelujah:
We do not want to keep the web server busy sending email

Requires: https://github.com/3scale/deploy/pull/721

Tested and working

image

Part of THREESCALE-7574

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -244,9 +244,9 @@ class User < ApplicationRecord
   def generate_lost_password_token!
     if generate_lost_password_token
       if account.provider?
-        ProviderUserMailer.lost_password(self).deliver_now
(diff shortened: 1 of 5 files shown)


akostadinov (dismissed):

akostadinov (dismissed):
Looks nice.

akostadinov:
btw do we notify user of sending failures with this background sending? I wonder if we should or should not.

hallelujah:
@akostadinov I do not understand you. What kind of notification would you like the end-user to receive?

akostadinov:
You are right. Exposing whether an email has been sent or not is not good :/

akostadinov (approved):

PR #2674, opened 2021-10-26
Update/refactor tests at test/integration/master

josemigallas:
♻️ refactor a lot
🚨 fixes lint (rubocop)
🎨 improves code format

### DIFF
--- a/test/integration/master/api/providers_controller_integration_test.rb
+++ b/test/integration/master/api/providers_controller_integration_test.rb
@@ -6,7 +6,7 @@ class Master::Api::ProvidersControllerIntegrationTest < ActionDispatch::Integrat
 
   def setup
     @account_plan = master_account.default_account_plan
-    @account_plan.expects(:approval_required).returns(false)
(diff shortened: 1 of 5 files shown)


test/integration/master/api/providers_controller_integration_test.rb
```diff
@@ -1,10 +1,12 @@
+# frozen_string_literal: true
+
 require 'test_helper'
 
 class Master::Api::ProvidersControllerIntegrationTest < ActionDispatch::IntegrationTest
 
   def setup
     @account_plan = master_account.default_account_plan
-    @account_plan.update_attribute(:approval_required, false)
+    @account_plan.update(approval_required: false)
```
akostadinov, line 9: Given that DB should be truncated after test, this should be a good approach. I would suggest to use the `!` form though.

```suggestion
    @account_plan.update!(approval_required: false)
```

akostadinov (approved):

hallelujah (approved):

PR #2678, opened 2021-10-27
Fixed alert test

nidhi-soni1104:
Fixed Failing test case

### DIFF
--- a/test/unit/alert_test.rb
+++ b/test/unit/alert_test.rb
@@ -15,8 +15,8 @@ class AlertTest < ActiveSupport::TestCase
     FactoryBot.create(:limit_alert, level: 50)
     FactoryBot.create(:limit_alert, level: 80)
 
-    alerts = Alert.by_level(80)
-    assert_equal 1, alerts.count
+    alert = Alert.by_level(80).last
+    assert_equal alert.level, 80
   end
 
   context 'Alert#kind' do
--- a/test/unit/alert_test.rb
+++ b/test/unit/alert_test.rb
@@ -12,12 +12,8 @@ class AlertTest < ActiveSupport::TestCase
   should validate_numericality_of(:utilization)
 
   def test_by_level
-    FactoryBot.create(:limit_alert, level: 50)
-    FactoryBot.create(:limit_alert, level: 80)
-
-    alert = Alert.by_level(80).last
+    alert = FactoryBot.create(:limit_alert, level: 80)
     assert_equal [alert], Alert.by_level(80).to_a
-
   end
 
   context 'Alert#kind' do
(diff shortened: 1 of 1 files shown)


akostadinov (approved):
lgtm, hopefully @hallelujah can confirm this is what he meant as this is test code, not application code. Returned values are sometimes good to test. But checking return type of a model scope is over the top IMHO.

hallelujah (changes requested):
Yes ... well you can even test that the only record is the one you just created instead of testing the count. Which will make it more clearer IMO ;)

nidhi-soni1104:
> Yes ... well you can even test that the only record is the one you just created instead of testing the count. Which will make it more clearer IMO ;)

I think the count was also clear to me, but I updated the test and checked the actual created object.

test/unit/alert_test.rb
```diff
@@ -15,9 +15,8 @@ def test_by_level
     FactoryBot.create(:limit_alert, level: 50)
     FactoryBot.create(:limit_alert, level: 80)
 
-    alerts = Alert.by_level(80)
-    assert_kind_of Alert::ActiveRecord_Relation, alerts
-    assert_equal 1, alerts.count
+    alert = Alert.by_level(80).last
```
hallelujah, line 18: What about

```
assert_equal [alert], Alert.by_level(80).to_a
```

nidhi-soni1104: Done

hallelujah: well not really what I expected :) you need to modify the line 16 into

```
alert = FactoryBot.create(:limit_alert, level: 80)
```

hallelujah (changes requested):

nidhi-soni1104:
@hallelujah test meaning was same but its a Good suggestion, Changes are done.

hallelujah (approved):

test/unit/alert_test.rb
```diff
@@ -12,12 +12,8 @@ class AlertTest < ActiveSupport::TestCase
   should validate_numericality_of(:utilization)
 
   def test_by_level
-    FactoryBot.create(:limit_alert, level: 50)
-    FactoryBot.create(:limit_alert, level: 80)
-
-    alerts = Alert.by_level(80)
-    assert_kind_of Alert::ActiveRecord_Relation, alerts
-    assert_equal 1, alerts.count
+    alert = FactoryBot.create(:limit_alert, level: 80)
```
hallelujah, line 15: Wait, why did you remove the level 50 one? Does not that change the meaning of the test?

hallelujah (changes requested):

test/unit/alert_test.rb
nidhi-soni1104: I think No, because we are not even using it anywhere we have just created it, It was useful when we are checking count.

nidhi-soni1104: Cleared doubt with @akostadinov and sorted the PR. @hallelujah

akostadinov (approved):

hallelujah (approved):

PR #2680, opened 2021-10-27
replaced a removed method sanitize

nidhi-soni1104:
Fixed error by replacing the removed method sanitize

### DIFF
--- a/test/unit/rails_sqli_test.rb
+++ b/test/unit/rails_sqli_test.rb
@@ -8,7 +8,7 @@ class RailsSqliTest < ActiveSupport::TestCase
     # No error should be raised on that and the SQL
 
     CMS::Section.find_by(params)
-    sanitized = ActiveRecord::Base.sanitize params['system_name'].to_s
+    sanitized = ActiveRecord::Base.connection.quote params['system_name'].to_s
     assert_match sanitized, CMS::Section.where(params).to_sql
   end
 


test/unit/rails_sqli_test.rb
```diff
@@ -8,7 +8,7 @@ def test_action_controller_params_are_escaped
     # No error should be raised on that and the SQL
 
     CMS::Section.find_by(params)
-    sanitized = ActiveRecord::Base.sanitize params['system_name'].to_s
+    sanitized = ActiveRecord::Base.connection.quote params['system_name'].to_s
```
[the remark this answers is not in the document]
akostadinov: I don't know if we should keep it or not. It has to be tested to see if it serves any purpose with current rails or not. Actually it was Hery's suggestion to try delete it. It might be safe to leave it if there is no time to investigate.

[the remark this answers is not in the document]
nidhi-soni1104: @hallelujah If  we keep initializer file then we need to apply some workarounds given to fix this issue in particular test cases which are failing as I did in the commit, but If we keep initializer then my first commit is sufficient and fix the issue. Let me know you thoughts so we can fix and close this PR.

test/unit/rails_sqli_test.rb
```diff
@@ -7,13 +7,13 @@ def test_action_controller_params_are_escaped
     # This should fix # https://groups.google.com/forum/#!topic/rubyonrails-security/8CVoclw-Xkk
     # No error should be raised on that and the SQL
 
-    CMS::Section.find_by(params)
-    sanitized = ActiveRecord::Base.sanitize params['system_name'].to_s
-    assert_match sanitized, CMS::Section.where(params).to_sql
+    CMS::Section.find_by_system_name(params['system_name'].to_s)
```
hallelujah, line 10: `find_by_system_name` should not be used, those dynamic finders are deprecated!
Also you changed the meaning of the test.

Before we test that when passed a Hash in `where` we sanitize the passed parameters.
Now you already have them passed as a String ... which is not really what the CVE is about

test/unit/rails_sqli_test.rb
```diff
@@ -7,13 +7,13 @@ def test_action_controller_params_are_escaped
...
+    sanitized = ActiveRecord::Base.connection.quote params['system_name'].to_s
+    assert_match sanitized, CMS::Section.where(system_name: params['system_name'].to_s).to_sql
   end
 
   def test_quoting_acton_controller_parameters
     params = ActionController::Parameters.new('system_name' => {'0' => 'foo'})
-    master_account.sections.find_by_system_name(params[:system_name])
+    master_account.sections.find_by_system_name(params[:system_name].to_s)
```
hallelujah, line 17: This is not what we want ;)
What I meant is if you remove the initializer, does it pass? meaning it was already fixed by rails so we do not need it anymore.

hallelujah (changes requested):

test/unit/rails_sqli_test.rb
nidhi-soni1104: @hallelujah No, if we remove initializer then it fails and so we need to apply this workarounds. if we keep initializer file no workarounds are needed I think.

As I mentioned earlier my first commit fixed the issue in which we are not removing initialize filer. I can remove the second commit and this depreciation is fixed.

hallelujah: 2 subjects:

1. So removing the initializer will open a vulnerability that was not fixed in Rails, so keep it.
2. The point here was that your change is not testing anything...

nidhi-soni1104: @hallelujah for your 2 point, if we are keeping initializer, no changes are needed in test. All changes will be reverted and test remains same as it was before.

akostadinov (approved):

hallelujah (approved):

PR #2684, opened 2021-10-29
Update/refactor test/integration/user-management-api/account_plan_features_test

josemigallas:
♻️ refactor a lot
🚨 fixes lint (rubocop)
🎨 improves code format

### DIFF
--- a/test/integration/user-management-api/account_plan_features_test.rb
+++ b/test/integration/user-management-api/account_plan_features_test.rb
@@ -1,22 +1,23 @@
+# frozen_string_literal: true
+
 require 'test_helper'
 
 class Admin::Api::AccountPlanFeaturesTest < ActionDispatch::IntegrationTest
   def setup
-    @provider = FactoryBot.create :provider_account, :domain => 'provider.example.com'
-    @account_plan = FactoryBot.create :account_plan, :issuer => @provider
-    FactoryBot.create :feature, :featurable => @provider
+    @provider = FactoryBot.create(:provider_account, domain: 'provider.example.com')
+    @account_plan = FactoryBot.create(:account_plan, issuer: @provider)
+    @feat = FactoryBot.create(:feature, featurable: @provider)
 
     host! @provider.admin_domain
   end
 
+  attr_reader :feat
+
   test 'index' do
-    feat = FactoryBot.create :feature, :featurable => @provider
     @account_plan.features << feat
     @account_plan.save!
 
-    get(admin_api_account_plan_features_path(@account_plan.id),
-             :provider_key => @provider.api_key, :format => :xml)
-
+    get admin_api_account_plan_features_path(@account_plan.id), params: { provider_key: @provider.api_key, format: :xml }
     assert_response :success
 
     xml = Nokogiri::XML::Document.parse(@response.body)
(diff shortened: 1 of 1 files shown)


test/integration/user-management-api/account_plan_features_test.rb
```diff
@@ -25,52 +26,37 @@ def setup
...
-              :feature_id => feat.id,
-              :provider_key => @provider.api_key, :format => :xml)
-
+    post admin_api_account_plan_features_path(@account_plan.id), params: { feature_id: feat.id, provider_key: @provider.api_key, format: :xml }
     assert_response :success
 
     xml = Nokogiri::XML::Document.parse(@response.body)
-
-    assert xml.xpath('.//feature/id').children.first.text == feat.id.to_s
+    assert_equal xml.xpath('.//feature/id').children.first.text, feat.id.to_s
```
akostadinov, line 55: This seems to be safe here. But just FYI `#assert_equal` is not always safe. See #2627

test/integration/user-management-api/account_plan_features_test.rb
```diff
@@ -25,52 +26,37 @@ def setup
...
   pending_test 'enable existing feature'
 
   test 'disable feature' do
-    feat = FactoryBot.create :feature, :featurable => @provider
-
-    post(admin_api_account_plan_features_path(@account_plan.id),
-              :feature_id => feat.id,
-              :provider_key => @provider.api_key, :format => :xml)
-
+    post admin_api_account_plan_features_path(@account_plan.id), params: { feature_id: feat.id, provider_key: @provider.api_key, format: :xml }
```
akostadinov, line 51: Just curious, how are we disabling anything here? In docs I see this as creating.
image

akostadinov (approved):
One question. I assume everything is alright so approving anyway.

test/integration/user-management-api/account_plan_features_test.rb
```diff
@@ -25,52 +26,37 @@ def setup
...
     xml = Nokogiri::XML::Document.parse(@response.body)
-
-    assert xml.xpath('.//feature/id').children.first.text == feat.id.to_s
+    assert_equal xml.xpath('.//feature/id').children.first.text, feat.id.to_s
   end
 
-  pending_test 'disable non-existing feature'
-
+  test 'disable non-existing feature' do
+    post admin_api_account_plan_features_path(@account_plan.id), params: { feature_id: 'nope', provider_key: @provider.api_key, format: :xml }
```
hallelujah, line 59: Shouldn't it be `nil` or empty string `''` instead?

test/integration/user-management-api/account_plan_features_test.rb
hallelujah: Well create and enable are basically synonym in this context. I think doc is a bit confusing like you, but it is to create a link between a feature and a plan

test/integration/user-management-api/account_plan_features_test.rb
josemigallas: Do you mean `feature_id`? I think those are as good options as an inexistent ID, no?

hallelujah: Ah! well why not, not sure :)

test/integration/user-management-api/account_plan_features_test.rb
akostadinov: I was referring to test name `disable feature`. But I don't see us disable anything. That's why I asked. Maybe test name better be changed?

> doc is a bit confusing like you

Apologizes, fwiw my gf thinks the same :angel:

test/integration/user-management-api/account_plan_features_test.rb
josemigallas: It looks to me that they're exactly the same:
https://github.com/seattlerb/minitest/blob/f4f57afaeb3a11bd0b86ab0757704cb78db96cf4/lib/minitest/assertions.rb#L218

test/integration/user-management-api/account_plan_features_test.rb
josemigallas: It's very confusing yes, I even debugged a little and didn't found anything "disabled"... ultimately decided not to change it and just fix the deprecation. We can take a look at it again in the future.

jsmadis (approved):

hallelujah (approved):

PR #2686, opened 2021-10-29
Change `redirect_to :back` with new method `redirect_back`

nidhi-soni1104:

This PR removes the depreciated warning for redirect_to :back

https://issues.redhat.com/browse/THREESCALE-7759

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -66,7 +66,7 @@ class ApplicationController < ActionController::Base
   helper_method :sublayout
 
   def redirect_back_or_to(*fallback_params)
-    redirect_back(fallback_location: root_path)
+    redirect_to :back
   rescue ActionController::RedirectBackError
     redirect_to(*fallback_params)
   end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -66,7 +66,7 @@ class ApplicationController < ActionController::Base
   helper_method :sublayout
 
   def redirect_back_or_to(*fallback_params)
-    redirect_to :back
+    redirect_back(fallback_location: root_path)
   rescue ActionController::RedirectBackError
     redirect_to(*fallback_params)
   end
--- a/app/controllers/provider/admin/applications_controller.rb
+++ b/app/controllers/provider/admin/applications_controller.rb
@@ -103,7 +103,7 @@ class Provider::Admin::ApplicationsController < FrontendController
       redirect_to provider_admin_applications_path
     else
       flash[:notice] = 'Not possible to delete application'
-      redirect_back(fallback_location: root_path)
+      redirect_back(fallback_location: provider_admin_applications_path)
     end
   end
 
--- a/app/controllers/provider/admin/applications_controller.rb
+++ b/app/controllers/provider/admin/applications_controller.rb
@@ -103,7 +103,7 @@ class Provider::Admin::ApplicationsController < FrontendController
       redirect_to provider_admin_applications_path
     else
       flash[:notice] = 'Not possible to delete application'
-      redirect_to :back
+      redirect_back(fallback_location: root_path)
     end
   end
 
--- a/app/controllers/provider/admin/cms/versions_controller.rb
+++ b/app/controllers/provider/admin/cms/versions_controller.rb
@@ -25,7 +25,7 @@ class Provider::Admin::CMS::VersionsController < Provider::Admin::CMS::BaseContr
       redirect_to polymorphic_path([:edit, :provider, :admin, @page])
     else
       flash[:error] = "Problem reverting version"
-      redirect_to :back
+      redirect_back(fallback_location: root_path)
     end
   end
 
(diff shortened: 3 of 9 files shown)


app/controllers/application_controller.rb
```diff
@@ -66,7 +66,7 @@ def status
   helper_method :sublayout
 
   def redirect_back_or_to(*fallback_params)
-    redirect_to :back
+    redirect_back(fallback_location: root_path)
   rescue ActionController::RedirectBackError
     redirect_to(*fallback_params)
```
akostadinov, line 71: Can `redirect_back` with `fallback_location` as specified above, raise `ActionController::RedirectBackError`?

nidhi-soni1104: May be not, because I am really confused about the fallback path from application controller, So checked with root_path which also make fails  2 test cases, redirect_to(*fallback_params) is also not a solution as it will also make fails the test case.

Any suggestion what can be the correct fallback_location?

akostadinov: If it was me, I'd try to see what happens when `fallback_location` parameter is not set. Also I'd add a breakpoint to see what these `fallback_params` are and whether they can be passed as `fallback_location`.

nidhi-soni1104: I tried all but nothing worked even setting fallback_location to fallback_params

nidhi-soni1104: I think we don't need to do any change to this method, because `ActionController::RedirectBackError` will rescue the error and this will not fail so reverting back to how its was instead of giving any fallback location.

josemigallas: @akostadinov No I don't think `redirect_back` is supposed to raise such error because that's the reason why it was created. `fallback_location` is a required argument now.

josemigallas: @nidhi-soni1104 You should redirect_back to the route passed in `*fallback_params`. When no route is passed, then use `root_path` as default.

akostadinov: This was my guess, that `fallback_params` should contain information where to fallback to. Thank you.

app/controllers/provider/admin/applications_controller.rb
```diff
@@ -103,7 +103,7 @@ def destroy
       redirect_to provider_admin_applications_path
     else
       flash[:notice] = 'Not possible to delete application'
-      redirect_to :back
+      redirect_back(fallback_location: root_path)
```
hallelujah, line 106: Maybe we want to redirect to the list of applications `provider_admin_applications_path`? Or to the edit page of the application. @3scale/system-ux  what is your thought about it?

app/lib/forum_support/user_topics.rb
```diff
@@ -27,7 +27,7 @@ def create
 
       respond_to do |format|
         flash[:notice] = 'You have successfully subscribed to the thread.' if @user_topic.save
-        format.html { redirect_to :back }
+        format.html { redirect_back(fallback_location: @user_topic) }
```
hallelujah, line 30: Better use route helpers than resource

hallelujah (changes requested):

app/controllers/provider/admin/cms/versions_controller.rb
```diff
@@ -25,7 +25,7 @@ def revert
       redirect_to polymorphic_path([:edit, :provider, :admin, @page])
     else
       flash[:error] = "Problem reverting version"
-      redirect_to :back
+      redirect_back(fallback_location: root_path)
```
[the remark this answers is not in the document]
nidhi-soni1104: There is no edit route for this, may be we can redirect to show , index or root_path. I think redirecting to show 
path(**provider_admin_cms_template_version**) will be a better option if no test case fails.

nidhi-soni1104:
@hallelujah Updated redirections

akostadinov:
@nidhi-soni1104 , what happened with the `application_controller.rb` update?

nidhi-soni1104:
> @nidhi-soni1104 , what happened with the `application_controller.rb` update?

I have reverted it because redirectig back to fallback_params location and root_path both are failing specs.

akostadinov:
Maybe specs need to be updated? Or different value passed to `fallback_params`? What were the failures?

nidhi-soni1104:
> Maybe specs need to be updated? Or different value passed to `fallback_params`? What were the failures?

I think we are not changing the redirection to new path so specs should not fail, when we change the redirection it give failure to the wrong redirection path.

akostadinov:
ok, then can you pass the proper path to `fallback_params`?

nidhi-soni1104:
> fallback_params

I think the back redirection is from application controller so everytime path is not same and fallback_params are passed different from different actions.

nidhi-soni1104:
> > fallback_params
> 
> I think the back redirection is from application controller so everytime path is not same and fallback_params are passed different from different actions.

even I found the request.referer is also nil , which is suggested in many articles.

hallelujah (approved):

PR #2687, opened 2021-11-02
Removing MessageBus

hallelujah:
## Issues

It is disabled for really long time, and is hard to understand.

- issue with process getting terminated randmly

```
 Global messages on 12355 timed out, restarting process
13884 successfully terminated by `TERM` signal.
```

- UI was already removed when migrating to React Patternfly
- Zync queues were waiting for nothing and it was raising a MessageBusTimeout
- it is not used, and disabling it make it timeout somehow

Closes https://issues.redhat.com/browse/THREESCALE-7765

## Steps to reproduce with current HEAD 9ad04a4b9609dc031d55627e0964e08ee7d9c9be

- Use production environment
- disable message bus in config/zync.yml and config/message_bus.yml
- launch a rails server and wait a little bit more than 3 minutes
- see your rails server terminated with a message like the one above

### Explanation

- When disabling in the config we only disable the message bus client not the server. The message bus server is still waiting for some messages. See https://github.com/discourse/message_bus/blob/v2.0.9/lib/message_bus.rb#L473-L506
- Despite this config https://github.com/3scale/porta/commit/444e28aa68c4a13cb92628eac2af0e9d704505d5 there is no ping to the server, thus after the 3' mark MessageBus will restart the process.


josemigallas (commented):
I think we still have to remove it from licenses.xml.
We could remove message_bus.yml from `.gitignore` as well.

akostadinov (approved):
Looks nice. I just don't know where license file changes came from. Perhaps unrelated? But I guess they are auto-generated and correct.

hallelujah:
@akostadinov 
> Looks nice. I just don't know where license file changes came from. Perhaps unrelated? But I guess they are auto-generated and correct.

It was asked by @josemigallas I just regenerated the licenses.xml with the corresponding rake task
The thing is when we update dependencies we should run it so we have them up to date.
We had a check made by @didierofrivia but it looks like it does not work anymore ...

josemigallas (approved):

PR #2692, opened 2021-11-03
[THREESCALE-5010] sphinx 5.3 does not set default value

akostadinov:
Upgrading to Rails 5.1 benefits from Sphinx update to 5.3.0. In that version
Sphinx does not set default configuration file path value and starting
Sphinx locally in dev mode fails.

Emulating same behavior in the configuration file.

### DIFF
--- a/config/docker/thinking_sphinx.yml
+++ b/config/docker/thinking_sphinx.yml
@@ -1,5 +1,5 @@
 common: &sphinx
-  configuration_file: <%= ENV['THINKING_SPHINX_CONFIGURATION_FILE'] %>
+  configuration_file: <%= ENV.fetch('THINKING_SPHINX_CONFIGURATION_FILE') { Rails.root.join("config/#{Rails.env}.sphinx.conf")} %>
   big_document_ids: true
   html_strip: 1
   html_remove_elements: "style, script"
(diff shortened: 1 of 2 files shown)


hallelujah (commented):
How about in openshift and docker?

hallelujah (changes requested):
Some of those configuration files are also in 

https://github.com/3scale/porta/tree/master/config/docker
https://github.com/3scale/porta/tree/master/openshift/system/config

The logic is to be found in

https://github.com/3scale/porta/blob/2b1a24b175c9dbe6157c1ed3affa2cdce458ecce/Dockerfile#L22

https://github.com/3scale/porta/blob/2b1a24b175c9dbe6157c1ed3affa2cdce458ecce/openshift/system/Dockerfile#L59
https://github.com/3scale/porta/blob/2b1a24b175c9dbe6157c1ed3affa2cdce458ecce/openshift/system/Dockerfile#L62

and similarly in the productization Dockerfile

hallelujah:
>  I'm not sure what configuration file is it using there.

Look at the Dockerfile ;)
They override each other. if not in openshift then no worries it uses the config/docker one

hallelujah (approved):

PR #2701, opened 2021-11-04
Fixes profile test which saves profile withou account

jsmadis:

Fixes profile model to throw abort when validation fails.

### DIFF
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -107,7 +107,7 @@ class Profile < ApplicationRecord
   protected
 
   def validate_presence_of_account
-    !self.account.nil?
+    throw :abort if self.account.nil?
   end
 
   def fix_http str


test/unit/profile_test.rb
```diff
@@ -20,8 +20,9 @@ class ProfileTest < ActiveSupport::TestCase
     should 'not be saved without account' do
       #this complex tests are used because if we try to save with account_id == nil
       # a mysql ActiveRecord::StatementInvalid exception is raised
-      assert_nothing_raised { @invalid_profile.save }
-      assert_equal false, @invalid_profile.save
+      assert_raises(ActiveRecord::NotNullViolation) do
```
hallelujah, line 23: I think the test means the contrary: `it must not raise an error`
So my bet is that in the model, instead of catching `ActiveRecord::StatementInvalid` only we should also catch `ActiveRecord::NotNullViolation`

hallelujah (changes requested):

test/unit/profile_test.rb
hallelujah: Oh I see in fact it is an issue with callbacks

https://github.com/3scale/porta/blob/2b1a24b175c9dbe6157c1ed3affa2cdce458ecce/app/models/profile.rb#L110

This needs to `throw :abort`

jsmadis: Nice catch!
Thanks @hallelujah

hallelujah (approved):

PR #2702, opened 2021-11-04
Fix Application plans select

josemigallas:

Removes `buyerCanSelectPlan` parameter, which in fact means _developer can select plan_ and it shouldn't affect the New Application page.

[THREESCALE-7834: admin can't create application when developer is not allowed to pick a plan](https://issues.redhat.com/browse/THREESCALE-7834)

**Verification steps** 

See JIRA


akostadinov (approved):
On ruby side I only see some data removed from presenter so looks good. As long as page looks good, the js parts also seem ok to me.

hallelujah (approved):

damianpm (approved):

damianpm (changes requested):
LGTM js side, but cukes failing

damianpm (approved):

hallelujah (approved):

PR #2710, opened 2021-11-05
Update/refactor integrations/user-management-api/applications

josemigallas:
♻️ refactor a lot
🚨 fixes lint (rubocop)
🎨 improves code format

### DIFF
--- a/test/integration/user-management-api/applications_test.rb
+++ b/test/integration/user-management-api/applications_test.rb
@@ -17,9 +17,10 @@ class EnterpriseApiApplicationsTest < ActionDispatch::IntegrationTest
 
     @application_plan = FactoryBot.create(:application_plan, issuer: @provider.default_service)
     @application_plan.publish!
-    @buyer.buy! @application_plan
-
     @application = @buyer.buy! @application_plan
+
+    # Create additional apps to have more data
+    @buyer.buy! @application_plan
 
     host! @provider.admin_domain
   end
@@ -65,7 +66,7 @@ class EnterpriseApiApplicationsTest < ActionDispatch::IntegrationTest
         get admin_api_applications_path, params: params.merge({ service_id: @service.id })
 
         assert_response :success
-        assert_equal 2, Nokogiri::XML::Document.parse(@response.body).xpath("//applications").children.length
+        assert_equal @token.owner.accessible_cinstances.size, Nokogiri::XML::Document.parse(@response.body).xpath("//applications").children.length
       end
 
       should '#index with access to no service' do
--- a/test/integration/user-management-api/applications_test.rb
+++ b/test/integration/user-management-api/applications_test.rb
@@ -1,299 +1,287 @@
+# frozen_string_literal: true
+
 require 'test_helper'
 
 class EnterpriseApiApplicationsTest < ActionDispatch::IntegrationTest
-
   include TestHelpers::BackendClientStubs
   include TestHelpers::ApiPagination
 
   def setup
     stub_backend_get_keys
 
-    @provider = FactoryBot.create :provider_account, :domain => 'provider.example.com'
+    @provider = FactoryBot.create(:provider_account, domain: 'provider.example.com')
     @service = @provider.services.first
 
-    @buyer = FactoryBot.create(:buyer_account, :provider_account => @provider)
+    @buyer = FactoryBot.create(:buyer_account, provider_account: @provider)
     @buyer.buy! @provider.default_account_plan
 
-    @application_plan = FactoryBot.create :application_plan, :issuer => @provider.default_service
+    @application_plan = FactoryBot.create(:application_plan, issuer: @provider.default_service)
     @application_plan.publish!
     @buyer.buy! @application_plan
+
+    @application = @buyer.buy! @application_plan
 
     host! @provider.admin_domain
   end
 
   #TODO: test extra fields in indexes
 
-  # Access token
-
-  test 'index (access_token)' do
-    User.any_instance.stubs(:has_access_to_all_services?).returns(false)
-    user  = FactoryBot.create(:member, account: @provider, admin_sections: ['partners'])
-    token = FactoryBot.create(:access_token, owner: user, scopes: 'account_management')
-
-    get(admin_api_applications_path)
-    assert_response :forbidden
-    get(admin_api_applications_path, access_token: token.value)
-    assert_response :success
-    User.any_instance.expects(:member_permission_service_ids).returns([@service.id]).at_least_once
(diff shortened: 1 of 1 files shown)


test/integration/user-management-api/applications_test.rb
```diff
@@ -1,299 +1,287 @@
...
-    get admin_api_applications_path(:format => :xml), :provider_key => @provider.api_key
+      should '#index with access to some service' do
+        User.any_instance.expects(:member_permission_service_ids).returns([@service.id]).at_least_once
+        get admin_api_applications_path, params: params.merge({ service_id: @service.id })
 
-    assert_response :success
-    assert_applications @response.body, :backend => :oauth
-  end
+        assert_response :success
+        assert_equal 2, Nokogiri::XML::Document.parse(@response.body).xpath("//applications").children.length
```
akostadinov, line 68: Does this mean two applications? Where are 2 coming from?

test/integration/user-management-api/applications_test.rb
```diff
@@ -1,299 +1,287 @@
...
     end
 
-    should 'return 404 on non found app' do
-      get(find_admin_api_applications_path(:format => :xml),
-               :user_key => "SHAWARMA",
-               :provider_key => @provider.api_key)
+    pending_test 'index returns fields defined'
 
+    test 'return 404 on non found app' do
+      get find_admin_api_applications_path(format: :xml), params: params.merge({ user_key: "SHAWARMA" })
```
akostadinov, line 151: shawarma not found - not good

test/integration/user-management-api/applications_test.rb
hallelujah: It comes from the first setup https://github.com/3scale/porta/pull/2710/files#diff-f0116a6712202a02588a528293a86589c4898120efb2ca2e64fd22399dc26299R9-R25

Buyer buys two times ;)

hallelujah (approved):

akostadinov (dismissed):

test/integration/user-management-api/applications_test.rb
hallelujah: @akostadinov 
You mean no food!

test/integration/user-management-api/applications_test.rb
```diff
@@ -1,299 +1,287 @@
...
+    def setup
+      super
+      user = FactoryBot.create(:member, account: @provider, admin_sections: ['partners'])
+      @token = FactoryBot.create(:access_token, owner: user, scopes: 'account_management')
+    end
 
-    get admin_api_applications_path(:format => :xml), :provider_key => @provider.api_key
+    context 'with access to all services' do
+      setup do
+        User.any_instance.stubs(:has_access_to_all_services?).returns(true)
```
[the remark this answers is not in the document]
josemigallas: Maybe in practice they are the same, but not logic-wise. Or is it?
https://github.com/3scale/porta/blob/afabb64f5d543d5af46fba747e1ef83b9e933f43/app/models/user/permissions.rb#L98-L100

test/integration/user-management-api/applications_test.rb
josemigallas: Since this test is failing, I'll remove the hardcode number

test/integration/user-management-api/applications_test.rb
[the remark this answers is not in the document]
hallelujah: Yes not logic wise but this methods should have been tested in `test/unit/user_test.rb` (if not that is bad) so we can use one of the branch.

test/integration/user-management-api/applications_test.rb
```diff
@@ -17,10 +17,11 @@ def setup
 
     @application_plan = FactoryBot.create(:application_plan, issuer: @provider.default_service)
     @application_plan.publish!
-    @buyer.buy! @application_plan
-
     @application = @buyer.buy! @application_plan
 
+    # Create additional apps to have more data
+    @buyer.buy! @application_plan
```
akostadinov, line 23: Did you figure out what is the difference between 
```
@application = @buyer.buy! @application_plan
@buyer.buy! @application_plan
```
and
```
@buyer.buy! @application_plan
@application = @buyer.buy! @application_plan
```
and why it works one way but not the other?

test/integration/user-management-api/applications_test.rb
```diff
@@ -65,7 +66,7 @@ def setup
         get admin_api_applications_path, params: params.merge({ service_id: @service.id })
 
         assert_response :success
-        assert_equal 2, Nokogiri::XML::Document.parse(@response.body).xpath("//applications").children.length
+        assert_equal @token.owner.accessible_cinstances.size, Nokogiri::XML::Document.parse(@response.body).xpath("//applications").children.length
```
akostadinov, line 69: I actually liked `2` more because it tests something that we know, not something that might be broken in the same way in the ruby expression and the api call. :)

akostadinov (approved):

test/integration/user-management-api/applications_test.rb
hallelujah: I prefer the other version, 2 is like a magic constant, if you create more applications then this test fails but it should not :)

test/integration/user-management-api/applications_test.rb
josemigallas: It's the same, I just changed it for a better readability. One of the apps `@application` is used in the tests, so I moved its declaration to the top. The second one is just to have some more data, so I moved it to the bottom. But they do the same.

test/integration/user-management-api/applications_test.rb
josemigallas: The way I see it, this test is about "getting applications that are accessible by the buyer". We test the integration between user permission (token) and the endpoint.
Say we changed the user permissions, made it more strict, and now the buyer can only access 1 of the 2 applications. If the endpoint logic is outdated or broken, the test will pass, because we have a hardcoded 2 and nothing to tell us this 2 is no longer correct.
WDYT?

test/integration/user-management-api/applications_test.rb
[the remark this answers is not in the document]
hallelujah: In fact if you could set the stub on a particular instance of an object (User or Service Object) it is the way to go, but sadly you can't really do that with mocha, probably with rspec-mocks you could.

That makes me think that `has_access_to_all_services?` does not belong to the `User` model, but `Ability` or other service object `can? :access_all_services?`. Does it make sense though? :thinking:

hallelujah (approved):

PR #2712, opened 2021-11-08
Guard from truncate on nil NilClass

hallelujah:
Fixes THREESCALE-7300


akostadinov (approved):
Looks good.

wrt messages without subject and body - not nice, I guess we better keep if it worked in the past.

nidhi-soni1104 (approved):
LGTM

PR #2714, opened 2021-11-09
Run oracle database in detached mode 

jsmadis:

Makes sure that `docker start oracle-database` won't print errors if the container doesn't exist.
Also runs the container in detached mode.

Those changes are for better integration with porta-dev-tools


hallelujah (approved):

akostadinov:
Well, in the passed we discussed that not waiting for it to start may make other targets which include starting oracle to fail.

PR #2718, opened 2021-11-10
ignore IntelliJ temporary out dir

akostadinov:


hallelujah (commented):
You are the first having this issue :) Where does this `out` dir come from. Can you just ignore it globally?

```
git config --global core.excludesfile $HOME/.config/git/ignore && echo out >> $HOME/.config/git/ignore
```

or if you just want for the current repo:

```
echo out >> .git/info/exclude
```

I mean, it is rather strange to add it to the shared .gitignore while it only happens for just your use case. Not that I am against it, I find it strange not to use first your own config.

akostadinov:
If you run unit and integration tests with intelliJ, it will show up for everybody. 

https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore#L53
https://gist.github.com/alifa98/afecac889d14fff0ade363cc5e079fd7#file-gitignore-L70

hallelujah:
Yeah nobody before ran test with Intellij ;) Please ignore locally :)

PR #2719, opened 2021-11-11
THREESCALE-7852 fixed deprecated positional arguments in functional buyers test

nidhi-soni1104:

This PR fix deprecated positional arguments in functional buyers test.
Fixed rubocops

https://issues.redhat.com/browse/THREESCALE-7852

### DIFF
--- a/app/controllers/buyers/users_controller.rb
+++ b/app/controllers/buyers/users_controller.rb
@@ -92,8 +92,6 @@ class Buyers::UsersController < Buyers::BaseController
   end
 
   def redirect_back_or_show_detail
-    redirect_back(fallback_location: root_path)
-  rescue ActionController::RedirectBackError
-    redirect_to admin_buyers_account_user_path(account_id: user.account_id, id: user.id)
+    redirect_back(fallback_location: admin_buyers_account_user_path(account_id: user.account_id, id: user.id))
   end
 end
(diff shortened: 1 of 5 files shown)


app/controllers/buyers/users_controller.rb
```diff
@@ -92,7 +92,7 @@ def user_params
   end
 
   def redirect_back_or_show_detail
-    redirect_to(:back)
+    redirect_back(fallback_location: root_path)
   rescue ActionController::RedirectBackError
     redirect_to admin_buyers_account_user_path(account_id: user.account_id, id: user.id)
```
hallelujah, line 97: I think you should merge those lines

hallelujah (changes requested):

app/controllers/buyers/users_controller.rb
nidhi-soni1104: @hery I tried but it fails the test so I redirected it to root_path and which makes test passed.

hallelujah: What test is failing?
The rescue is the same as fallback ...

nidhi-soni1104: I tried with redirect_to but I think If I give this path as fallback then it will not fail and we can remove rescue .

nidhi-soni1104: @hallelujah Updated location

akostadinov (approved):

jsmadis (approved):

PR #2732, opened 2021-11-12
[THREESCALE-7884] Fix service top applications endpoint

jsmadis:

Fixes regression introduced by refactoring and upgrading everything to be rails 5.1 compatible.

https://issues.redhat.com/browse/THREESCALE-7884

### DIFF
--- a/app/controllers/stats/data/services_controller.rb
+++ b/app/controllers/stats/data/services_controller.rb
@@ -41,7 +41,7 @@ class Stats::Data::ServicesController < Stats::Data::BaseController
   ##~ op.parameters.add @parameter_metric_name
   #
   def top_applications
-    options = slice_and_use_defaults(params, :metric_name, :period, :since, :timezone)
+    options = slice_and_use_defaults(stats_params, :metric_name, :period, :since, :timezone)
 
     begin
       @data = @source.top_clients(options)
@@ -76,5 +76,13 @@ class Stats::Data::ServicesController < Stats::Data::BaseController
     rescue ActiveRecord::RecordNotFound
       render_error "Service not found", :status => :not_found
     end
+  end
+
+  def stats_params
+    required = %i[metric_name period since]
+
+    permitted_params = params.permit(required | %i[timezone])
+    permitted_params.require(required)
+    permitted_params
   end
 end
--- a/test/integration/stats/data/service_controller_test.rb
+++ b/test/integration/stats/data/service_controller_test.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class Stats::Data::ServicesControllerTest < ActionDispatch::IntegrationTest
+  def setup
+    @application = FactoryBot.create :cinstance
+    host! @application.provider_account.admin_domain
+    user = @application.provider_account.admins.first!
+    @token = FactoryBot.create(:access_token, owner: user, scopes: %w[stats], permission: 'rw').value
+  end
+
+  attr_reader :application, :token
+
+  test 'top applications' do
+    params = { access_token: token, since: '2021-08-30 00:00:00', period: :day, metric_name: 'hits' }
+    get top_applications_stats_data_services_path(application.service, format: :json), params: params
+    assert_response :success
+  end
+
+  test 'missing params' do
+    params = { access_token: token, since: '2021-08-30 00:00:00', period: :day, metric_name: 'hits' }
+    get top_applications_stats_data_services_path(application.service, format: :json), params: params.except(:since)
+    assert_response :bad_request
+
+    get top_applications_stats_data_services_path(application.service, format: :json), params: params.except(:period)
+    assert_response :bad_request
+
+    get top_applications_stats_data_services_path(application.service, format: :json), params: params.except(:metric_name)
+    assert_response :bad_request
+  end
+end


test/integration/stats/data/service_test.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class Stats::Data::ServicesControllerTest < ActionDispatch::IntegrationTest
```
josemigallas, line 5: What's this test doing and why should it fail?

akostadinov (approved):

test/integration/stats/data/service_test.rb
akostadinov: What do you mean with "why should it fail"? I don't understand the question.

test/integration/stats/data/service_test.rb
```diff
@@ -0,0 +1,20 @@
...
+    @application = FactoryBot.create :cinstance
+    host! @application.provider_account.admin_domain
+    user = @application.provider_account.admins.first!
+    @token = FactoryBot.create(:access_token, owner: user, scopes: %w[stats], permission: 'rw').value
+  end
+
+  attr_reader :application, :token
+
+  test 'top applications' do
+    params = { access_token: token, since: '2021-08-30 00:00:00', period: :day, metric_name: 'hits' }
```
akostadinov, line 16: I would suggest another test where `period` is not specified. `usage_params` has a special handling of it.

test/integration/stats/data/service_test.rb
jsmadis: @josemigallas It's testing basic requests to the `top_application` endpoint. Right now it fails with status code 400

josemigallas: I wonder what's the relation between the change in the controller and this test. If the test is right now failing with a 400, then we should also add a test case for that. The test must prove that the new changes are useful.

test/integration/stats/data/service_test.rb
josemigallas: @akostadinov that would be an answer for my question "why should it fail": because it's missing period.

test/integration/stats/data/service_test.rb
jsmadis: This test case covers it, we were missing a basic test case that would reveal the bug.

test/integration/stats/data/service_controller_test.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  test 'top applications' do
+    params = { access_token: token, since: '2021-08-30 00:00:00', period: :day, metric_name: 'hits' }
+    get top_applications_stats_data_services_path(application.service, format: :json), params: params
+    assert_response :success
+  end
+
+  test 'missing params' do
+    params = { access_token: token, period: :day, metric_name: 'hits' }
+    get top_applications_stats_data_services_path(application.service, format: :json), params: params
```
josemigallas, line 23: We can dry this by extracting `params` in [the test above](https://github.com/3scale/porta/pull/2732/files#diff-c1f1a7109b968f74461fa8fa1bb44594459e5be05ad77ec1214af9f842333b4eR16)
```suggestion
    get top_applications_stats_data_services_path(application.service, format: :json), params: params.except(:period)
```

app/controllers/stats/data/services_controller.rb
```diff
@@ -77,4 +77,12 @@ def set_source
       render_error "Service not found", :status => :not_found
     end
   end
+
+  def stats_params
+    required = %i[metric_name period since]
+
+    permitted_params = params.permit(required | %i[timezone])
+    permitted_params.require(required)
```
hallelujah, line 85: does it do anything?

jsmadis: Yes, it will fail the request if the required param is missing.

```
Required parameter missing: period
```

app/controllers/stats/data/services_controller.rb
```diff
@@ -77,4 +77,12 @@ def set_source
...
+    permitted_params
```
sergioifg94, line 86:
```suggestion
    params.permit(required | %i[timezone]).require(required)
```
I think this would be equivalent, and more concise

jsmadis: It wouldn't be the same because `.require(required)` will return an array of values for the required keys. We would lose other parameters that are not required but still permitted.

jsmadis: `.permit` filters only permitted parameters and we want to save that value.
https://github.com/rails/rails/blob/v5.0.7.2/actionpack/lib/action_controller/metal/strong_parameters.rb#L469

`.require` checks the parameters for required keys and returns them. We only want to check their presence.
https://github.com/rails/rails/blob/v5.0.7.2/actionpack/lib/action_controller/metal/strong_parameters.rb#L406

jsmadis: @hallelujah this explains your questions as well

sergioifg94: I see, thanks @jsmadis !

test/integration/stats/data/service_controller_test.rb
```diff
@@ -0,0 +1,32 @@
...
+    assert_response :success
+  end
+
+  test 'missing params' do
+    params = { access_token: token, since: '2021-08-30 00:00:00', period: :day, metric_name: 'hits' }
+    get top_applications_stats_data_services_path(application.service, format: :json), params: params.except(:since)
+    assert_response :bad_request
+
+    get top_applications_stats_data_services_path(application.service, format: :json), params: params.except(:period)
+    assert_response :bad_request
```
akostadinov, line 27: Can we have another test where `granularity` is present instead of `period`. Sorry for not being clear the first time.

From https://github.com/3scale/porta/blob/master/app/controllers/stats/api/base_controller.rb#L57
```ruby
permitted_params.require([:granularity, :since]) if permitted_params[:period].blank?
```

akostadinov, line 27: Can we have another test where `granularity` is present instead of `period`. Sorry for not being clear the first time.

From https://github.com/3scale/porta/blob/master/app/controllers/stats/api/base_controller.rb#L57
```ruby
permitted_params.require([:granularity, :since]) if permitted_params[:period].blank?
```

akostadinov (changes requested):

akostadinov (approved):
Actually approving for merge to avoid delays. Test can be added after.

test/integration/stats/data/service_controller_test.rb
jsmadis: I made a mistake in my first attempt I wanted to reuse the method which permits the params but it can't be used here.

This endpoint doesn't permit `granularity` as a parameter, it only permits `:metric_name, :period, :since, :timezone`
You can see it here: https://github.com/3scale/porta/blob/master/app/controllers/stats/data/services_controller.rb#L44

Which calls this: https://github.com/3scale/porta/blob/e1d7807c58d941abe34117c9d476449e5152ff5c/app/controllers/stats/data/base_controller.rb#L100

PR #2733, opened 2021-11-12
Fixes integration provider tests for rails 5.1 upgrade

jsmadis:

Fixes integration provider tests for rails 5.1 upgrade

### DIFF
--- a/app/views/provider/admin/dashboards/widgets/_service_hits.html.slim
+++ b/app/views/provider/admin/dashboards/widgets/_service_hits.html.slim
@@ -13,7 +13,7 @@ article.DashboardWidget id = widget.id
           ' In order to show Hits for the #{friendly_service_name(service)} you need to have at least one application sending traffic to it.
         p
           ' Consider
-          => link_to 'making some test calls', edit_admin_service_integration_path(widget_params['service_id'], :anchor => 'staging')
+          => link_to 'making some test calls', edit_admin_service_integration_path(widget_params[:service_id], :anchor => 'staging')
           | from the Integration page to get a feel for what you'd see here.
 
       - else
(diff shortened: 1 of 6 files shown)


app/views/provider/admin/dashboards/widgets/_service_hits.html.slim
```diff
@@ -13,7 +13,7 @@ article.DashboardWidget id = widget.id
           ' In order to show Hits for the #{friendly_service_name(service)} you need to have at least one application sending traffic to it.
         p
           ' Consider
-          => link_to 'making some test calls', edit_admin_service_integration_path(widget_params['service_id'], :anchor => 'staging')
```
hallelujah, line 16: Where did the `widget_params` come from in the old version? Are you sure you are not overriding anything here?

test/integration/provider/invoices_controller_test.rb
```diff
@@ -73,23 +73,23 @@ class Finance::Provider::InvoicesControllerTest < ActionDispatch::IntegrationTes
           Invoice.any_instance.stubs("transition_allowed?").returns(true)
           Invoice.any_instance.stubs("#{action}!").returns(true)
 
-          put url_for([action, :admin, :finance, @invoice]), :format => 'js'
+          put url_for([action, :admin, :finance, @invoice]), params: { format: 'js' }
```
hallelujah, line 76: Or we can put the format inside of the `url_for`

app/views/provider/admin/dashboards/widgets/_service_hits.html.slim
jsmadis: It comes from `widget_params = widget.params.symbolize_keys`

I saw an issue that when accessing service_id as `widget_params['service_id']` because keys were symbolized. That's why I changed that to a symbol.

Anyway, those 2 files are not used and will be deleted by @josemigallas

hallelujah (approved):

PR #2739, opened 2021-11-15
signup controller handles CORS by itself

akostadinov:
`rack-cors` gem handles all pre-flight requests but signup controller has its
own way of handling these. So we make rack-cors skip handling of requests
with the signup controller path.

Otherwise 3scale.net signup form gets broken.

To test you need cors enabled in `config/cors.yml`. The this is with the fix:
```
$ curl -v 'http://master-account.3scale.localhost:3000/p/signup?signup_origin=website-signup' -X OPTIONS -H 'User-Agent: Mozilla/5.0 (X11; Fedora; Linux x86_64; rv:94.0) Gecko/20100101 Firefox/94.0' -H 'Accept: */*' -H 'Accept-Language: en-US,en;q=0.7,bg;q=0.3' --compressed -H 'Access-Control-Request-Method: GET' -H 'Access-Control-Request-Headers: 3scale-origin,x-requested-with' -H 'Referer: https://www.3scale.net/' -H 'Origin: https://www.3scale.net' -H 'DNT: 1' -H 'Connection: keep-alive' -H 'Sec-Fetch-Dest: empty' -H 'Sec-Fetch-Mode: cors' -H 'Sec-Fetch-Site: same-site' -H 'Pragma: no-cache' -H 'Cache-Control: no-cache'
*   Trying ::1:3000...
* connect to ::1 port 3000 failed: Connection refused
*   Trying 127.0.0.1:3000...
* Connected to master-account.3scale.localhost (127.0.0.1) port 3000 (#0)
> OPTIONS /p/signup?signup_origin=website-signup HTTP/1.1
> Host: master-account.3scale.localhost:3000
> Accept-Encoding: deflate, gzip, br
> User-Agent: Mozilla/5.0 (X11; Fedora; Linux x86_64; rv:94.0) Gecko/20100101 Firefox/94.0
> Accept: */*
> Accept-Language: en-US,en;q=0.7,bg;q=0.3
> Access-Control-Request-Method: GET
> Access-Control-Request-Headers: 3scale-origin,x-requested-with
> Referer: https://www.3scale.net/
> Origin: https://www.3scale.net
> DNT: 1
> Connection: keep-alive
> Sec-Fetch-Dest: empty
> Sec-Fetch-Mode: cors
> Sec-Fetch-Site: same-site
> Pragma: no-cache
> Cache-Control: no-cache
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Date: Mon, 15 Nov 2021 21:21:34 GMT
< Connection: close
< Access-Control-Allow-Origin: https://www.3scale.net
< Access-Control-Allow-Methods: GET, POST
< Access-Control-Allow-Headers: Origin, Accept, Cookie, Content-Type, X-Requested-With, X-CSRF-Token, 3scale-Origin
< Access-Control-Allow-Credentials: true
< Content-Type: text/html
< Cache-Control: no-cache
< Set-Cookie: _system_session=TzxcFDSlkjxkvFDAzhjv3ZnUDM5b3F4eDBjWHc9PS0tVXZpU1VtaFhiQkZUU3hwTVRTTGQ5dz09--a4bb8c1f611b5d2e19ace44da48c9ce3920a9dbb; path=/; HttpOnly; SameSite=Lax
< X-Request-Id: 17d731b0-8e68-449b-a1fa-641ca5d56e56
< X-Runtime: 0.542218
< X-Served-By: Horak
< Vary: Accept-Encoding
< Content-Encoding: gzip
< X-Content-Type-Options: nosniff
< X-XSS-Protection: 1; mode=block
< 
* Closing connection 0
```

Without the fix:
```
curl -v 'http://master-account.3scale.localhost:3000/p/signup?signup_origin=website-signup' -X OPTIONS -H 'User-Agent: Mozilla/5.0 (X11; Fedora; Linux x86_64; rv:94.0) Gecko/20100101 Firefox/94.0' -H 'Accept: */*' -H 'Accept-Language: en-US,en;q=0.7,bg;q=0.3' --compressed -H 'Access-Control-Request-Method: GET' -H 'Access-Control-Request-Headers: 3scale-origin,x-requested-with' -H 'Referer: https://www.3scale.net/' -H 'Origin: https://www.3scale.net' -H 'DNT: 1' -H 'Connection: keep-alive' -H 'Sec-Fetch-Dest: empty' -H 'Sec-Fetch-Mode: cors' -H 'Sec-Fetch-Site: same-site' -H 'Pragma: no-cache' -H 'Cache-Control: no-cache'
*   Trying ::1:3000...
* connect to ::1 port 3000 failed: Connection refused
*   Trying 127.0.0.1:3000...
* Connected to master-account.3scale.localhost (127.0.0.1) port 3000 (#0)
> OPTIONS /p/signup?signup_origin=website-signup HTTP/1.1
> Host: master-account.3scale.localhost:3000
> Accept-Encoding: deflate, gzip, br
> User-Agent: Mozilla/5.0 (X11; Fedora; Linux x86_64; rv:94.0) Gecko/20100101 Firefox/94.0
> Accept: */*
> Accept-Language: en-US,en;q=0.7,bg;q=0.3
> Access-Control-Request-Method: GET
> Access-Control-Request-Headers: 3scale-origin,x-requested-with
> Referer: https://www.3scale.net/
> Origin: https://www.3scale.net
> DNT: 1
> Connection: keep-alive
> Sec-Fetch-Dest: empty
> Sec-Fetch-Mode: cors
> Sec-Fetch-Site: same-site
> Pragma: no-cache
> Cache-Control: no-cache
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Date: Mon, 15 Nov 2021 21:22:21 GMT
< Connection: close
< X-Frame-Options: DENY
< X-Content-Type-Options: nosniff
< X-XSS-Protection: 1; mode=block
< 
* Closing connection 0
```

### DIFF
--- a/lib/three_scale/middleware/cors.rb
+++ b/lib/three_scale/middleware/cors.rb
@@ -2,14 +2,13 @@
 
 module ThreeScale::Middleware
   class Cors < Rack::Cors
-    include Rails.application.routes.url_helpers
-
     def initialize(app, opts = {}, &block)
       super(app, opts) { set_config }
+      set_excludes
 
-      if block_given?
+      if block_given? # rubocop:disable Style/GuardClause why: keep copy from superclass
         if block.arity == 1
-          block.call(self)
+          block.call(self) # rubocop:disable Performance/RedundantBlockCall why: keep copy from superclass
         else
           instance_eval(&block)
         end
@@ -17,25 +16,18 @@ module ThreeScale::Middleware
     end
 
     def call(env)
-      return @app.call(env) if disabled? || signup_controller?(env)
+      return @app.call(env) if excluded?(env)
 
       super
     end
+
+    private
 
     def config
       Rails.configuration.three_scale.cors
     end
 
-    delegate :enabled, to: :config
-    alias enabled? enabled
-
-    def disabled?
-      !enabled?
-    end
-
-    private
-
-    def set_config
+    def set_config # rubocop:disable Metrics/CyclomaticComplexity, Metrics/AbcSize
       rules = config.allow.presence || []
 
       rules.map(&:symbolize_keys).each do |rule|
@@ -52,8 +44,67 @@ module ThreeScale::Middleware
       end
     end
 
-    def signup_controller?(env)
-      evaluate_path(env).start_with? provider_signup_path # '/p/signup'
+    def set_excludes
+      rules = config.exclude.presence || []
+
+      @excludes = rules.map do |rule|
+        ExcludeMatcher.for(rule.symbolize_keys)
+      end
     end
+
+    def excluded?(env)
+      @excludes.any? { |matcher| matcher.matches?(env) }
+    end
+  end
+
+  module ExcludeMatcher
+    class << self
+      def matchers
+        @matchers ||= Set.new
+      end
+
+      def add_matchers(*matcher_classes)
+        matchers.merge matcher_classes.flatten
+      end
+
+      def for(rule)
+        matcher = matchers.detect { |klass| klass.supports?(rule) }
+
+        raise ThreeScale::ConfigurationError, "Unsupported exclude specification: #{rule}" unless matcher
+
+        matcher.new(rule)
+      end
+    end
+
+    class PathRegexpMatcher
+      attr_reader :regexp
+
+      def initialize(rule)
+        regexp = rule[:path_regexp]
+        @regexp = regexp.is_a?(Regexp) ? regexp : /#{regexp}/
+      end
+
+      def matches?(env)
+        regexp.match? env["PATH_INFO"]
+      end
+
+      def self.supports?(rule)
+        rule[:path_regexp].present?
+      end
+    end
+
+    class PathPrefixMatcher < PathRegexpMatcher
+      def initialize(rule)
+        path = rule[:path_prefix]
+        segment_end_check = path.end_with?("/") ? "" : "(?:/|$)"
+        super({path_regexp: /^#{path}#{segment_end_check}/})
+      end
+
+      def self.supports?(rule)
+        rule[:path_prefix].present?
+      end
+    end
+
+    add_matchers PathPrefixMatcher, PathRegexpMatcher
   end
 end
--- a/lib/three_scale/middleware/cors.rb
+++ b/lib/three_scale/middleware/cors.rb
@@ -6,12 +6,12 @@ module ThreeScale::Middleware
       super(app, opts) { set_config }
       set_excludes
 
-      if block_given? # rubocop:disable Style/GuardClause why: keep copy from superclass
-        if block.arity == 1
-          block.call(self) # rubocop:disable Performance/RedundantBlockCall why: keep copy from superclass
-        else
-          instance_eval(&block)
-        end
+      return unless block_given?
+
+      if block.arity == 1
+        yield(self)
+      else
+        instance_eval(&block)
       end
     end
 
--- a/lib/three_scale/middleware/cors.rb
+++ b/lib/three_scale/middleware/cors.rb
@@ -15,7 +15,7 @@ module ThreeScale::Middleware
     end
 
     def call(env)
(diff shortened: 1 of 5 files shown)


akostadinov:
Failures seem to be because of shoulda instability.

hallelujah:
I wonder if we could just ignore some paths in our CORS lib instead, like adding a global setting in the config:

```yaml

ignore:
  - path: /p/signup
 ```

**Comment by akostadinov** [comment]
It is possible to make it configuration. But this particular exception is kind of code induced. And I find it more practical to have it handled in code as well.

Please ignore failing tests and merge if you agree to keep in code. Or let me know if you want me to make it configuration.

**Comment by hallelujah** [comment]
Well I prefer it is a configuration, later if we want to change for another forgotten controller we should be able to do that as well without redeploying another code

**Comment by akostadinov** [comment]
Ok, I'm checking how to do it. I don't like configuration that should by no use case be changed though. I don't mean if we find or implement something that needs it. I mean users should never change it.

**Comment by akostadinov** [comment]
How about testing, should we test the configuration file that it ignores signup controller or leave it to some other system?

**Comment by hallelujah** [comment]
About testing, this basically can be a gem on its own, adding more functionality to Rack::Cors
So think about it this way.

File: lib/three_scale/middleware/cors.rb [hunk 1]
```diff
@@ -2,6 +2,8 @@
 
 module ThreeScale::Middleware
   class Cors < Rack::Cors
+    include Rails.application.routes.url_helpers
```
hallelujah, line 5: You should not need this if you use the configuration

lib/three_scale/middleware/cors.rb
```diff
@@ -15,7 +17,7 @@ def initialize(app, opts = {}, &block)
     end
 
     def call(env)
-      return @app.call(env) unless enabled?
+      return @app.call(env) if disabled? || signup_controller?(env)
```
hallelujah, line 20:
```suggestion
      return @app.call(env) if disabled? || excluded_by_config?(env)
```

lib/three_scale/middleware/cors.rb
```diff
@@ -45,5 +51,9 @@ def set_config
         end
       end
     end
+
+    def signup_controller?(env)
+      evaluate_path(env).start_with? provider_signup_path # '/p/signup'
+    end
```
hallelujah, line 57:
```suggestion
    def excluded_by_config?(env)
      config.ignore.any?{ |rule| IgnoreMatcher.new(rule).matches?(env) }
    end
    
    class IgnoreMatcher
      def initialize(rule)
        @rule = rule
        # Add anything you could find useful, checks, exclusivity betwen path and regexp for e.g.
        # Below `@regexp`, `@path` can also be better as subclass RegexpMatcher, PathMatcher etc ...
        @regex = rule[:regexp]
        @path = rule[:path]
      end
      
      def matches?(env)
        request = ActionDispatch::Request.new(env)
        if @path.present?
           request.path == @path
        elsif @regexp.present?
          @regexp =~ request,path          
        end
      end
    end
```

hallelujah (changes requested):
I would do it with the configuration, In suggestions this is how I would do it. Not completely implemented but that is the idea

akostadinov:
Updated, please check again. I didn't notice your suggested changes so I implemented in a little bit different way.

One thing is that I wanted to use `start_with?` because it is much faster than regexp matching. And it will be performed for all requests so I thought that it is worth going for it. I can add regexp matcher as well if you want.

I also extended tests. I added a test for the example configuration files. I'm not sure if this was the proper way.

Another thing I changed is to completely avoid inserting the middleware when it is disabled. This bugged me from the beginning, that we always inserted it.

So let me know any suggestions that you have.

hallelujah (approved):

jsmadis (approved):

PR #2745, opened 2021-11-16
Fix MetricsController failing tests

josemigallas:
Hoping to fix failing tests from backend apis' MetricsController

### DIFF
--- a/test/integration/admin/api/backend_apis/metrics_controller_index_text.rb
+++ b/test/integration/admin/api/backend_apis/metrics_controller_index_text.rb
@@ -1,0 +1,45 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class Admin::Api::BackendApis::MetricsControllerIndexTest < ActionDispatch::IntegrationTest
+  def setup
+    @provider = FactoryBot.create(:simple_provider)
+    admin = FactoryBot.create(:admin, account: provider)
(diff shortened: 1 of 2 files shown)


test/integration/admin/api/backend_apis/metrics_controller_index_text.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    response_metrics = JSON.parse(response.body)['metrics']
+
+    assert_equal 2, response_metrics.length
+    response_metric_ids = response_metrics.map { |metric| metric.dig('metric', 'id') }
+    assert_same_elements backend_api.metrics.pluck(:id), response_metric_ids
+  end
+
+  test 'index can be paginated' do
+    backend_api = FactoryBot.create(:backend_api, account: provider)
```
akostadinov, line 35: seems like this can go to setup

akostadinov (approved):
I don't understand why a separate class is needed. Also a minor optimization comment. But in general, if this is to try things out, it is good to merge in my view.

test/integration/admin/api/backend_apis/metrics_controller_index_text.rb
josemigallas: All tests would share the same backend_api, wouldn't they? With same plans and all

PR #2755, opened 2021-11-19
[integration/api] Fixing test for the upgrade

jsmadis:

Fixing integration/api tests for the upgrade

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -65,10 +65,8 @@ class ApplicationController < ActionController::Base
   #
   helper_method :sublayout
 
-  def redirect_back_or_to(*fallback_params)
-    redirect_to :back
-  rescue ActionController::RedirectBackError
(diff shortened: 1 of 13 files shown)


app/controllers/application_controller.rb
```diff
@@ -65,10 +65,8 @@ def status
   #
   helper_method :sublayout
 
-  def redirect_back_or_to(*fallback_params)
-    redirect_to :back
-  rescue ActionController::RedirectBackError
-    redirect_to(*fallback_params)
+  def redirect_back_or_to(fallback_location)
```
akostadinov, line 68: This is changing accepted parameters type. Can't break expectations somewhere?

jsmadis: I checked the usage and every time the method was used we passed only a single parameter. So I don't think it will break anything.

Also, the tests passed

hallelujah (approved):

PR #2758, opened 2021-11-22
Fixes integration proxy configs tests to be compatible with rails 5.1

jsmadis:

Fixes proxy/configs_test to be compatible with rails 5.1


nidhi-soni1104 (approved):
LGTM

PR #2759, opened 2021-11-22
THREESCALE-7676: Allow tenant_id to be propagated from account

sergioifg94:

There is an issue where the `master` field of provider accounts is `null` by default. This results in triggers for the `backend_apis`, `alerts`, and `log_entries` tables, that set the `tenant_id` field from the related account **only** if the `master` field is `false` not setting the field.

This PR fixes it by adding a `before_create` callback to the Accounts model that sets the master field of provider accounts to `false` if it's not set before. For existing data, a new rake task (`fixes:tenant_id`) is added to set these missing fields.

https://issues.redhat.com/browse/THREESCALE-7676

**Verification steps** 

##### Callback

Create a new account with the `provider` field set to `true` and verify that `master` is set to `false` automatically

##### Task

Run the task and verify that for an existing provider account where the `master` field was `null`, it's set to `false`

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -99,7 +99,7 @@ class Account < ApplicationRecord
 
   before_validation(on: :create, if: :provider?) { generate_s3_prefix }
   before_validation(on: :create, if: :provider?) { generate_domains }
-  before_create :generate_site_access_code
+  before_create :generate_site_access_code, :set_default_master
 
   attr_protected :master, :provider, :buyer, :from_email, :vat_rate, :sample_data, :default_service_id, :s3_prefix,
                  :provider_account_id, :paid_at, :paid, :signs_legal_terms, :tenant_id, :default_account_plan_id,
@@ -578,6 +578,11 @@ class Account < ApplicationRecord
 
   def generate_site_access_code
     self.site_access_code ||= SecureRandom.hex(5) if provider?
+  end
+
+  def set_default_master
+    self.master = false if provider? && master.nil?
+    true
   end
 
   def destroy_all_users
--- a/app/workers/set_tenant_id_worker.rb
+++ b/app/workers/set_tenant_id_worker.rb
@@ -5,13 +5,9 @@ class SetTenantIdWorker < ApplicationJob
   def perform(provider)
     provider.update_column(:master, false)
 
-    {
-      BackendApi => :account_id,
-      LogEntry => :provider_id,
-      Alert => :account_id,
-    }.each do |model, relation|
-      model.where(relation => provider.id).find_each do |instance|
-        ModelTenantIdWorker.perform_later(instance, :tenant_id, provider)
+    [:backend_apis, :log_entries, :alerts].each do |relation|
+      provider.public_send(relation).where(tenant_id: nil).find_each do |instance|
+        ModelTenantIdWorker.perform_later(instance, provider.tenant_id)
       end
     end
   end
@@ -20,16 +16,16 @@ class SetTenantIdWorker < ApplicationJob
     unique :until_executed
 
     def perform(*)
-      Account.providers.not_master.find_each do |provider_batch|
-        SetTenantIdWorker.perform_later(provider_batch)
+      Account.providers.where(master: nil).find_each do |provider|
+        SetTenantIdWorker.perform_later(provider)
       end
     end
   end
 
   class ModelTenantIdWorker < ApplicationJob
 
-    def perform(object, attribute, provider)
-      object.update_column(attribute, provider.tenant_id)
+    def perform(object, tenant_id)
+      object.update_column(:tenant_id, tenant_id)
     end
 
   end
--- a/app/workers/set_tenant_id_worker.rb
+++ b/app/workers/set_tenant_id_worker.rb
@@ -14,7 +14,7 @@ class SetTenantIdWorker < ApplicationJob
       }.each do |model, relation|
         q = {relation => provider.id}
         
-        model.where(q).each do |instance|
+        model.where(relation => provider.id).find_each do |instance|
           instance.update!(tenant_id: provider.tenant_id)
         end
       end
--- a/app/workers/set_tenant_id_worker.rb
+++ b/app/workers/set_tenant_id_worker.rb
@@ -4,7 +4,6 @@ class SetTenantIdWorker < ApplicationJob
 
   def perform(provider_batch)
     provider_batch.each do |provider|
-      # Set the master field to false
       provider.master = false
       provider.save!
 
--- a/app/workers/set_tenant_id_worker.rb
+++ b/app/workers/set_tenant_id_worker.rb
@@ -25,7 +25,7 @@ class SetTenantIdWorker < ApplicationJob
     unique :until_executed
 
     def perform(*)
-      Account.providers.not_master.find_in_batches(batch_size: 100).each do |provider_batch|
+      Account.providers.not_master.find_each do |provider_batch|
         SetTenantIdWorker.perform_later(provider_batch)
       end
     end
--- a/app/workers/set_tenant_id_worker.rb
+++ b/app/workers/set_tenant_id_worker.rb
@@ -7,8 +7,6 @@ class SetTenantIdWorker < ApplicationJob
       provider.master = false
       provider.save!
 
-      # Update the backend_apis, log_entries and alerts to propagate the tenant_id from the updated
-      # provider
       {
         BackendApi => :account_id,
         LogEntry => :provider_id,
--- a/app/workers/set_tenant_id_worker.rb
+++ b/app/workers/set_tenant_id_worker.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+class SetTenantIdWorker < ApplicationJob
+
+  def perform(provider_batch)
+    provider_batch.each do |provider|
+      # Set the master field to false
+      provider.master = false
+      provider.save!
+
+      # Update the backend_apis, log_entries and alerts to propagate the tenant_id from the updated
+      # provider
+      {
+        BackendApi => :account_id,
+        LogEntry => :provider_id,
+        Alert => :account_id,
+      }.each do |model, relation|
+        q = {}
+        q[relation] = provider.id
+        
+        model.where(q).each do |instance|
+          instance.update!(tenant_id: provider.tenant_id)
+        end
+      end
+    end
+  end
+
+  class BatchEnqueueWorker < ApplicationJob
+    unique :until_executed
+
+    def perform(*)
+      Account.providers.not_master.find_in_batches(batch_size: 100).each do |provider_batch|
+        SetTenantIdWorker.perform_later(provider_batch)
+      end
+    end
+  end
+
+end
--- a/app/workers/set_tenant_id_worker.rb
+++ b/app/workers/set_tenant_id_worker.rb
@@ -12,8 +12,7 @@ class SetTenantIdWorker < ApplicationJob
         LogEntry => :provider_id,
         Alert => :account_id,
       }.each do |model, relation|
-        q = {}
-        q[relation] = provider.id
+        q = {relation => provider.id}
         
         model.where(q).each do |instance|
           instance.update!(tenant_id: provider.tenant_id)
--- a/app/workers/set_tenant_id_worker.rb
+++ b/app/workers/set_tenant_id_worker.rb
@@ -2,21 +2,16 @@
 
 class SetTenantIdWorker < ApplicationJob
 
-  def perform(provider_batch)
-    provider_batch.each do |provider|
-      provider.master = false
-      provider.save!
+  def perform(provider)
+    provider.update_column(:master, false)
 
-      {
-        BackendApi => :account_id,
-        LogEntry => :provider_id,
-        Alert => :account_id,
-      }.each do |model, relation|
-        q = {relation => provider.id}
-        
-        model.where(relation => provider.id).find_each do |instance|
-          instance.update!(tenant_id: provider.tenant_id)
-        end
+    {
+      BackendApi => :account_id,
+      LogEntry => :provider_id,
+      Alert => :account_id,
+    }.each do |model, relation|
+      model.where(relation => provider.id).find_each do |instance|
+        ModelTenantIdWorker.perform_later(instance, :tenant_id, provider)
       end
     end
   end
@@ -29,6 +24,14 @@ class SetTenantIdWorker < ApplicationJob
         SetTenantIdWorker.perform_later(provider_batch)
       end
     end
+  end
+
+  class ModelTenantIdWorker < ApplicationJob
+
+    def perform(object, attribute, provider)
+      object.update_column(attribute, provider.tenant_id)
+    end
+
   end
 
 end
--- a/lib/tasks/fixes.rake
+++ b/lib/tasks/fixes.rake
@@ -305,4 +305,9 @@ namespace :fixes do
     exec_batch.call(:providers, Account.providers, provider_states_transitions)
     exec_batch.call(:developers, Account.buyers, buyer_states_transitions)
   end
+
+  desc 'Fix tenant_id missing in some tables'
+  task :tenant_id => :environment do
+    SetTenantIdWorker::BatchEnqueueWorker.perform_later
+  end
 end


app/models/account.rb
```diff
@@ -580,6 +580,13 @@ def generate_site_access_code
     self.site_access_code ||= SecureRandom.hex(5) if provider?
   end
 
+  def set_default_master
+    if self.provider? && self.master == nil
```
hallelujah, line 584:
```suggestion
    if self.provider? && !self.master?
```

I think this is safer than nil check

app/models/account.rb
```diff
@@ -580,6 +580,13 @@ def generate_site_access_code
...
+      self.master = false
+    end
+    true
```
hallelujah, line 587: Oh because of old callback way to check for `false` you have to return `true`
Good catch ;)

db/migrate/20211117170147_set_tenant_id.rb
```diff
@@ -0,0 +1,17 @@
+class SetTenantId < ActiveRecord::Migration[5.0]
+  
+  def up
+    Account.where("provider = 1 AND master IS NULL").update_all(master: false)
```
hallelujah, line 4: The account table is quite big and is always used, an `UPDATE` will block the table for write.
This is not good in production as it will make web requests longer, probably killing the unicorn workers by timeout.

db/migrate/20211117170147_set_tenant_id.rb
```diff
@@ -0,0 +1,17 @@
...
+
+    {
+      BackendApi => :account,
+      LogEntry => :provider,
+      Alert => :account
+    }.each do |m, r|
+      m.joins(r).where(tenant_id: nil).where(accounts: { master: false }).each do |c|
+        c.update(tenant_id: c.send(r).tenant_id)
+      end
```
hallelujah, line 13: Yes the same here about locked `accounts` table.
For DML (Data Manipulation Language) we prefer a rake task and if the data to migrate are too big, we enqueue a job. We then delete it later mainly in another release (because of on-prem)

Example:

https://github.com/3scale/porta/blob/master/lib/tasks/services.rake#L28-L31

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20211109141544) do
+ActiveRecord::Schema.define(version: 20211117170147) do
```
hallelujah, line 13: We also have Oracle and Postgresql schema

https://github.com/3scale/porta/tree/master/db

app/models/account.rb
sergioifg94: I might be missing something but, wouldn't `!self.master?` evaluate to `true` if `master` is `nil` or `false`? In that case it would only be redundant in comparison to `self.master.nil?` as the body of the `if` statement sets the value to `false`.
I'm new to Ruby so there's a good chance I'll stand corrected 🙂

makikata: As `master` is a table column, it gets handy `StringInquirer` from ActiveRecord.
So anything that is _falsey_ value (`nil`,`false`, `''`) would return `false` and any _truthy_ value will return `true`

Here we want to check that the account is a provider but not master.

No worries about being new to Ruby ;) it is a perk from ActiveSupport https://api.rubyonrails.org/classes/ActiveSupport/StringInquirer.html

hallelujah: As `master` is a table column, it gets handy `StringInquirer` from ActiveRecord.
So anything that is _falsey_ value (`nil`,`false`, `''`) would return `false` and any _truthy_ value will return `true`

Here we want to check that the account is a provider but not master.

No worries about being new to Ruby ;) it is a perk from ActiveSupport https://api.rubyonrails.org/classes/ActiveSupport/StringInquirer.html

sergioifg94: > Here we want to check that the account is a provider **but not master**.

Actually what I want is to check that the account is a provider **but the master field is not set** (`NULL` in the database, which is what causes the triggers not to work as expected), and then set it to `false` by default.

app/workers/set_tenant_id_worker.rb
```diff
@@ -0,0 +1,18 @@
+class SetTenantIdWorker < ApplicationJob
+
+  def perform(*)
+    ActiveRecord::Base.transaction do
+      Account.where("provider = 1 AND master IS NULL").update_all(master: false)
```
hallelujah, line 5: `update_all` will lock the `accounts` table for as long as it is executed. This is dangerous as we have a live application:

1. If a request updates the data of an Account, it will wait until the lock is released or raise a Lock wait timeout exceeded error.
2. That error will return to the user as a 5xx which is not nice
3. it will slow down the application, eventually leading to a global Denial of Service

Some advices, make it simpler at the beginning, you can probably make it more optimized later but will you need it?

1. enqueue per provider the update
2. in the `SetTenantId#perform` do the update for the provider and the other models that depend on it but per object, having a loop is not bad, we can iterate from there

sergioifg94: Thanks @hallelujah I changed the `update_all` for smaller updates of batches of 100. That should be small enough to avoid significant locks. Wdyt?
```ruby
loop do
  ids = Account.where("provider = 1 AND master IS NULL").limit(100).to_a.map { |account| account.id }
  break if ids.empty?

  Account.where(id: ids).update_all(master: false)
end
```

app/workers/set_tenant_id_worker.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+class SetTenantIdWorker < ApplicationJob
+
+  def perform(*)
+    # Update the providers in batches of 100
+    loop do
+      ids = Account.where("provider = 1 AND master IS NULL").limit(100).to_a.map { |account| account.id }
```
hallelujah, line 8: You can use `Account.providers.not_master.find_each`
ActiveRecord::Query has handy magic methods ;)
Also `find_each` will use `find_in_batches`

db/migrate/20211117170147_set_tenant_id.rb
hallelujah: As mentionned in a previous comment you can look at a previous example:

https://github.com/3scale/porta/blob/master/lib/tasks/services.rake#L28-L31

Which will lead you to

https://github.com/3scale/porta/blob/1e77f3f0bd6471d0b98e1d09e6081b23a6eaa8f1/app/workers/create_default_proxy_worker.rb

app/workers/set_tenant_id_worker.rb
sergioifg94: I updated the PR following your suggestion, with only one difference: instead of enqueing a separate job for every single provider, I enqueue a job for batches of providers. I think the result would be the same but it wouldn't congest the queue as much

app/workers/set_tenant_id_worker.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+class SetTenantIdWorker < ApplicationJob
+
+  def perform(provider_batch)
+    provider_batch.each do |provider|
+      # Set the master field to false
```
josemigallas, line 7:
```suggestion
```

app/workers/set_tenant_id_worker.rb
```diff
@@ -0,0 +1,38 @@
...
+      provider.master = false
+      provider.save!
+
+      # Update the backend_apis, log_entries and alerts to propagate the tenant_id from the updated
+      # provider
```
josemigallas, line 12:
```suggestion
```

app/workers/set_tenant_id_worker.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+class SetTenantIdWorker < ApplicationJob
+
+  def perform(provider_batch)
+    provider_batch.each do |provider|
+      # Set the master field to false
+      provider.master = false
+      provider.save!
```
josemigallas, line 9: Why not `provider.update(master: false)`?

app/workers/set_tenant_id_worker.rb
```diff
@@ -0,0 +1,38 @@
...
+
+      # Update the backend_apis, log_entries and alerts to propagate the tenant_id from the updated
+      # provider
+      {
+        BackendApi => :account_id,
+        LogEntry => :provider_id,
+        Alert => :account_id,
+      }.each do |model, relation|
+        q = {}
+        q[relation] = provider.id
```
josemigallas, line 19: If `q` this doesn't depend on model/relation, can we move outside of the loop?
Also, maybe
```suggestion
        q = Hash(relation, provider.id)
```

josemigallas, line 19:
```suggestion
        q = Hash[relation, provider.id]
```

app/models/account.rb
```diff
@@ -580,6 +580,11 @@ def generate_site_access_code
     self.site_access_code ||= SecureRandom.hex(5) if provider?
   end
 
+  def set_default_master
+    self.master = false if provider? && master.nil?
```
josemigallas, line 584: So this statement translates to a very [IMO] confusing predicate:
```ruby
false if (self[:provider] || exists?(master: true)) && master.nil?
```

I self.master supposed to be a boolean? Isn't it a find method? https://github.com/3scale/porta/blob/master/app/models/account.rb#L300

Does this makes sense? `exists?(master: true) && master.nil?` They seem opposites to me. If so, it will always be `false` unless self[:provider].
```ruby
self.master = false unless self[:provider] && master.nil?
```

hallelujah: I really prefer not checking for nil as per my former review https://github.com/3scale/porta/pull/2759#discussion_r756221916 it does not make sense, it should be checking for `truthy` or `falsey` value.

And @josemigallas just find why it was a bad idea to check for nil

app/workers/set_tenant_id_worker.rb
akostadinov: will this raise on error?

app/workers/set_tenant_id_worker.rb
akostadinov:
```suggestion
        q = {relation => provider.id}
```

app/models/account.rb
akostadinov: > Actually what I want is to check that the account is a provider **but the master field is not set** (`NULL` in the database, which is what causes the triggers not to work as expected), and then set it to `false` by default. 

```suggestion
    self.master = false
```

How about this? In which situations will we want that to be initialized to true?

app/workers/set_tenant_id_worker.rb
akostadinov: According to docs, it will not raise on error.

app/workers/set_tenant_id_worker.rb
```diff
@@ -4,7 +4,6 @@ class SetTenantIdWorker < ApplicationJob
 
   def perform(provider_batch)
     provider_batch.each do |provider|
-      # Set the master field to false
```
akostadinov: :rofl:

app/models/account.rb
hallelujah: Why bother? that works fine ` if self.provider? && !self.master?`
Why even bother for checking `nil` values instead of both `nil` and `false` ? I do not see the point.

app/workers/set_tenant_id_worker.rb
```diff
@@ -0,0 +1,34 @@
...
+      provider.save!
+
+      {
+        BackendApi => :account_id,
+        LogEntry => :provider_id,
+        Alert => :account_id,
+      }.each do |model, relation|
+        q = {relation => provider.id}
+        
+        model.where(q).each do |instance|
```
hallelujah, line 17:
```suggestion
        model.where(relation => provider.id).find_each do |instance|
```

Prevent loading all objects in memory

app/workers/set_tenant_id_worker.rb
```diff
@@ -0,0 +1,34 @@
...
+        end
+      end
+    end
+  end
+
+  class BatchEnqueueWorker < ApplicationJob
+    unique :until_executed
+
+    def perform(*)
+      Account.providers.not_master.find_in_batches(batch_size: 100).each do |provider_batch|
```
hallelujah, line 28:
```suggestion
      Account.providers.not_master.find_each do |provider_batch|
```

hallelujah (changes requested):

app/models/account.rb
akostadinov: @hallelujah , then why not just 
```suggestion
    # prevent nil value for master
    self.master = master?
```

hallelujah: OK so I looked at it and it is used in `before_create`
We should just use:

```ruby
    self.master = provider? && master?
```

That will take into account all cases

app/workers/set_tenant_id_worker.rb
```diff
@@ -0,0 +1,34 @@
...
+    end
+  end
+
+  class BatchEnqueueWorker < ApplicationJob
+    unique :until_executed
+
+    def perform(*)
+      Account.providers.not_master.find_each do |provider_batch|
+        SetTenantIdWorker.perform_later(provider_batch)
+      end
```
hallelujah, line 30:
```suggestion
      Account.providers.not_master.find_each do |provider|
        SetTenantIdWorker.perform_later(provider)
      end
```

app/workers/set_tenant_id_worker.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+class SetTenantIdWorker < ApplicationJob
+
+  def perform(provider_batch)
```
hallelujah, line 5: You do not pass a batch here, you pass a single Account (provider) object

app/workers/set_tenant_id_worker.rb
hallelujah: I would use 

```ruby
provider.update_column(:master, false)
```

app/workers/set_tenant_id_worker.rb
```diff
@@ -0,0 +1,34 @@
...
+
+      {
+        BackendApi => :account_id,
+        LogEntry => :provider_id,
+        Alert => :account_id,
+      }.each do |model, relation|
+        q = {relation => provider.id}
+        
+        model.where(relation => provider.id).find_each do |instance|
+          instance.update!(tenant_id: provider.tenant_id)
```
hallelujah, line 18:
```suggestion
          instance.update!(tenant_id: provider.id)
```

app/workers/set_tenant_id_worker.rb
```diff
@@ -0,0 +1,34 @@
...
+      provider.save!
+
+      {
+        BackendApi => :account_id,
+        LogEntry => :provider_id,
+        Alert => :account_id,
+      }.each do |model, relation|
+        q = {relation => provider.id}
+        
+        model.where(relation => provider.id).find_each do |instance|
```
hallelujah, line 17: I prefer to create a class that sets the tenant_id for any model record.

```ruby
class ModelTenantIdWorker < ApplicationJob

  def perform(object, attribute, provider)
     object.update_column(attribute, provider.id)
  end
end
```


```

**Review by hallelujah — CHANGES_REQUESTED** [review]

File: app/models/account.rb (hunk 12 above)
Comment (akostadinov) [reply app/models/account.rb:584]: Now looking at this, it seems like the triggers can be fixed and avoid this question altogether. Presently in these 3 triggers we use `AND NOT master`. And on MySQL this means `Shows a record if the condition(s) is FALSE.`

But as Sergio noticed, `master` can also be `NULL`. But on PostgreSQL we use `master <> TRUE`. On Oracle we use `master <> 1`.

So I think on MySQL we can also use  `master <> TRUE`.

WDYT?

Comment (akostadinov) [reply app/models/account.rb:584]: We can also probably fix the 3 triggers to catch `NULL` values.

File: app/workers/set_tenant_id_worker.rb [hunk 20]
```diff
@@ -0,0 +1,37 @@
...
+
+  def perform(provider)
+    provider.update_column(:master, false)
+
+    {
+      BackendApi => :account_id,
+      LogEntry => :provider_id,
+      Alert => :account_id,
+    }.each do |model, relation|
+      model.where(relation => provider.id).find_each do |instance|
```
akostadinov, line 13:
```
      %i(backend_apis log_entries alerts).each do |relation|
        provider.public_send(relation).find_each do |instance|
         ....
```

app/workers/set_tenant_id_worker.rb
```diff
@@ -0,0 +1,37 @@
...
+        ModelTenantIdWorker.perform_later(instance, :tenant_id, provider)
```
akostadinov, line 14:
```suggestion
        ModelTenantIdWorker.perform_later(instance.class, instance.id, provider.id)
```

sergioifg94:
@akostadinov Thanks for the suggestions, I pushed a new commit after implementing them and testing that the solution still works

akostadinov:
Closing in favor of #2900, the only change is adding one suggestion from Hery and squashing.

akostadinov:
@sergioifg94, how did you test that the solution works? For me it fails when more than one provider accounts are created because we have a unique constraint on the master boolean field.

I'm asking you to know if I missed something. Seems like setting to `false` is not a viable option.

Again, I'm just asking, you don't have to fix it.

sergioifg94:
@akostadinov I tested it on the development database, I guess that because it's smaller I didn't run into that problem.

I wasn't aware of that constraint, but if only one row in the `accounts` table can have `master` set to `false`, that changes everything. It means as long as there's `backend_apis` that are related to an account that is not `master` the tenant_id will still be `null`. Maybe we need to propagate the `tenant_id` from accounts where `master` is `false` _or_ `null`

PR #2760, opened 2021-11-23
[email-async] Support entitlements

hallelujah:

It enqueues Entitlement support email in ActionMailer::DeliveryJob
Instead of sending it along the web request

It will suppress some 5xx errors when the SMTP server is not available at the time of the web request

Part of THREESCALE-7574

- Imported ActiveJob::Arguments and ActiveJob::Serializers from Rails 6.1.4.1
  in order to have Time serialization

I thought this is a good addition. ActiveJob arguments and serializers should be compatible with Rails 5.x

### DIFF
--- a/config/initializers/active_job_hooks.rb
+++ b/config/initializers/active_job_hooks.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+# TODO: Remove when migrated to Rails 6.1.4.1+
+#
+# This is extracted from Rails 6.1.4.1
+# Once we migrate there we can remove those files
+# Those are located in lib/active_job
+require 'active_job/arguments'
+require 'active_job/serializers'
--- a/app/services/support_entitlements_service.rb
+++ b/app/services/support_entitlements_service.rb
@@ -76,10 +76,10 @@ class SupportEntitlementsService
   end
 
   def notify_entitlements_assigned
-    AccountMailer.support_entitlements_assigned(account, notification_options).deliver_now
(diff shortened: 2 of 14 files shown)


akostadinov:
What happens without importing rails 6 serializers? As a whole I find it as acceptable to import these. I just think it is better to write a comment which version of Rails we can remove these files with and actually even better if we can do the version check in the initializer and raise in case this version is reached.

config/initializers/active_job_hooks.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+require "active_support/core_ext/hash"
```
akostadinov, line 3: Since this one appears to be loaded from standard libs, wouldn't file be loaded by itself or we need to do here. I guess we do but have to ask :)

hallelujah:
@akostadinov it is in the description of the PR


> **Special notes for your reviewer:**
> * Imported ActiveJob::Arguments and ActiveJob::Serializers from Rails 6.1.4.1
> in order to have Time serialization

akostadinov (approved):
I read this note. I was asking what happens without it. Are dates ignored or serialization fail or some other ill effect?

In any case PR is good to merge. I think it is preferable to import things from later Rails and have it working OOB in the future instead of implementing something custom now and then have to rework.

hallelujah:
Well if we don't have it it just fails to serialize Time, which is part of an argument of the support entitlement email

PR #2764, opened 2021-11-23
Billing summary should expire

hallelujah:
Issue:

Sometimes it does not unstore the results, because the batch callback was not called
The root cause is not yet found, this still ensures that Redis memory usage won't grow indefinitely

### DIFF
--- a/app/models/billing_summary.rb
+++ b/app/models/billing_summary.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 class BillingSummary
+  TTL = 1.day.to_i
 
   delegate :redis, to: :System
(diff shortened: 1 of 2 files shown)


app/models/billing_summary.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 class BillingSummary
+  TTL = 1.day.to_i
```
akostadinov, line 4: I

hallelujah: Yes summary is invalid.

akostadinov (approved):

PR #2767, opened 2021-11-24
[Params] Fixes devel portal contracts/service_plans/application controllers

jsmadis:
Fixes usage of params in the controller in the:

```
./lib/developer_portal/app/controllers/developer_portal/admin/contracts_controller.rb
./lib/developer_portal/app/controllers/developer_portal/admin/contracts/service_plans_controller.rb
./lib/developer_portal/app/controllers/developer_portal/admin/applications_controller.rb
```

### DIFF
--- a/lib/developer_portal/app/controllers/developer_portal/admin/applications_controller.rb
+++ b/lib/developer_portal/app/controllers/developer_portal/admin/applications_controller.rb
@@ -21,7 +21,7 @@ class DeveloperPortal::Admin::ApplicationsController < ::DeveloperPortal::BaseCo
 
   def index
     cinstances = current_account.bought_cinstances.includes(:service)
-                   .order_for_dev_portal.paginate(page: params[:page])
+                   .order_for_dev_portal.paginate(page: local_params[:page])
     collection = Liquid::Drops::Collection.for_drop(Liquid::Drops::Application).new(cinstances)
     pagination = Liquid::Drops::Pagination.new(cinstances, self)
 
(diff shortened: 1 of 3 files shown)


lib/developer_portal/app/controllers/developer_portal/admin/applications_controller.rb
```diff
@@ -169,14 +169,23 @@ def authorize_update_app
...
+    application_attributes.permit(*permitted_params).to_h
   end
 
   def fields_definitions
     FieldsDefinition.by_provider(site_account).by_target('Cinstance').pluck(:name)
   end
+
+  private
+
+  def local_params
```
akostadinov, line 188: Maybe name it `app_params` or something? Unless this `local_params` name is some convention I'm unaware of, in which case it is fine with me.

lib/developer_portal/app/controllers/developer_portal/admin/applications_controller.rb
```diff
@@ -169,14 +169,23 @@ def authorize_update_app
...
+    params.permit(%i[plan_id service_id page]).to_h
```
akostadinov, line 189:
```suggestion
    params.permit(%i[plan_id service_id page])
```

In file parameters seem to be used as parameters, not hash. So perhaps remove `to_h`.

akostadinov (changes requested):
Looks great, just 2 nitpicks.

lib/developer_portal/app/controllers/developer_portal/admin/applications_controller.rb
```diff
@@ -115,15 +115,15 @@ def applications
   def service
     @service ||= if single_service?
                    services.first
-                 elsif (service_id = params[:service_id])
+                 elsif (service_id = local_params[:service_id])
```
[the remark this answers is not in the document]
jsmadis: You are right, it's not. Those PR's were made when I had insufficient knowledge in the problematic area and are made to point stuff like this. Thanks for the comment.

jsmadis:

PR #2768, opened 2021-11-24
[Params] Fixes params usage in devel portal applications controllers

jsmadis:

Fixes params usage for rails 5.1

### DIFF
--- a/lib/developer_portal/app/controllers/developer_portal/admin/applications/alerts_controller.rb
+++ b/lib/developer_portal/app/controllers/developer_portal/admin/applications/alerts_controller.rb
@@ -61,15 +61,18 @@ class DeveloperPortal::Admin::Applications::AlertsController <  ::DeveloperPorta
   end
 
   def find_application
-    @cinstance = current_account.bought_cinstances.find(params[:application_id])
+    @cinstance = current_account.bought_cinstances.find(params.require(:application_id))
   end
 
   def collection
-    current_account.alerts.not_deleted.by_application(@cinstance).sorted.order_by(params[:sort], params[:direction])
+    current_account.alerts.not_deleted.by_application(@cinstance).sorted.order_by(local_params[:sort], local_params[:direction])
   end
 
   def resource
-    collection.find(params[:id])
+    collection.find(params.require(:id))
   end
 
+  def local_params
+    params.permit(%i[sort direction]).to_h
+  end
 end
--- a/lib/developer_portal/app/controllers/developer_portal/admin/applications/keys_controller.rb
+++ b/lib/developer_portal/app/controllers/developer_portal/admin/applications/keys_controller.rb
@@ -13,7 +13,7 @@ class DeveloperPortal::Admin::Applications::KeysController < DeveloperPortal::Ba
   end
 
   def create
-    @key = @cinstance.application_keys.add(params[:key])
+    @key = @cinstance.application_keys.add(params.permit!.to_h[:key])
 
     respond_to do |format|
       format.html do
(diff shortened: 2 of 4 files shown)


lib/developer_portal/app/controllers/developer_portal/admin/applications/alerts_controller.rb
```diff
@@ -61,15 +61,18 @@ def authorize_alerts
   end
 
   def find_application
-    @cinstance = current_account.bought_cinstances.find(params[:application_id])
+    @cinstance = current_account.bought_cinstances.find(params.require(:application_id))
```
hallelujah, line 64: What kind of warning errors does this solve? I am not aware of it. The only thing I could think of is:

```ruby
params = { application_id: {table: 'column'} 
```

So basically you might prevent from some SQL injection. AFAIK `params[:application_id]` still works if the parameter is a simple value (Integer, String ...)

lib/developer_portal/app/controllers/developer_portal/admin/applications/keys_controller.rb
```diff
@@ -13,7 +13,7 @@ def new
   end
 
   def create
-    @key = @cinstance.application_keys.add(params[:key])
+    @key = @cinstance.application_keys.add(params.permit!.to_h[:key])
```
hallelujah, line 16: Why `permit!` everything here?
Isn't `key` required?

Or if it is not required, then would that work better?

```suggestion
    @key = @cinstance.application_keys.add(params.permit(:key)[:key])
```

lib/developer_portal/app/controllers/developer_portal/admin/applications/alerts_controller.rb
```diff
@@ -61,15 +61,18 @@ def authorize_alerts
...
+    current_account.alerts.not_deleted.by_application(@cinstance).sorted.order_by(local_params[:sort], local_params[:direction])
   end
 
   def resource
-    collection.find(params[:id])
+    collection.find(params.require(:id))
   end
 
+  def local_params
+    params.permit(%i[sort direction]).to_h
```
akostadinov, line 76: Why do we need `#to_h` here?

lib/developer_portal/app/controllers/developer_portal/admin/applications/alerts_controller.rb
jsmadis: > What kind of warning errors does this solve? 

None, it tries to improve the logic of the parameters. If the parameter is missing it will return an error with the message `param is missing or the value is empty`.

> So basically you might prevent from some SQL injection. 

`params.require(:application_id)` would return unpermitted hash in this context. Which would raise an error if anyone tries to convert it to a hash.

I am not sure which method is better, I like the require because it makes raises the error if it's missing

lib/developer_portal/app/controllers/developer_portal/admin/applications/alerts_controller.rb
```diff
@@ -61,15 +61,18 @@ def authorize_alerts
...
-    current_account.alerts.not_deleted.by_application(@cinstance).sorted.order_by(params[:sort], params[:direction])
+    current_account.alerts.not_deleted.by_application(@cinstance).sorted.order_by(local_params[:sort], local_params[:direction])
   end
 
   def resource
-    collection.find(params[:id])
+    collection.find(params.require(:id))
   end
 
+  def local_params
```
sergioifg94, line 75:
```suggestion
  private

  def local_params
```

I think it's a good practice to make this private

jsmadis:

PR #2769, opened 2021-11-24
[Params] Fixes params usage in devel portal admin controllers

jsmadis:

Fixes usage of params in the service contract controller.

### DIFF
--- a/lib/developer_portal/app/controllers/developer_portal/admin/service_contracts_controller.rb
+++ b/lib/developer_portal/app/controllers/developer_portal/admin/service_contracts_controller.rb
@@ -24,7 +24,7 @@ class DeveloperPortal::Admin::ServiceContractsController < DeveloperPortal::Base
   end
 
   def create
-    @service_contract = current_account.bought_service_contracts.build params[:service_contract]
+    @service_contract = current_account.bought_service_contracts.build service_contract_params
     @service_contract.plan = site_account.service_plans.find params[:service_contract][:plan_id]
 
     if @service_contract.save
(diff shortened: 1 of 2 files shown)


lib/developer_portal/app/controllers/developer_portal/admin/service_contracts_controller.rb
```diff
@@ -82,4 +82,9 @@ def find_plan
     end
   end
 
+  def service_contract_params
+    internal_attributes = %w[created_at updated_at accepted_at]
+    attributes = current_account.fields.for(ServiceContract) - internal_attributes
```
josemigallas, line 87: This returns `["created_at", "updated_at", "description", "name", "redirect_url", "extra_fields", "create_origin", "first_traffic_at", "first_daily_traffic_at", "service_id", "accepted_at"]`

I think we just need to permit `description name redirect_url extra_field`.

Do you know where the form this controller handles is?

jsmadis: That's why I am extracting the internal_attributes (I should also remove the service_id field)
So it ends up as `["description", "name", "redirect_url", "extra_fields", "create_origin", "first_traffic_at", "first_daily_traffic_at", "service_id"]`>

I was checking the API docs for the application and we have update fields for `first_traffic_at` and `first_daily_traffic_at` so that's why I left it here.

> Do you know where the form this controller handles is?

I have no idea, I couldn't find it on the developer portal. I think this is some old controller which is not used anymore but I am not sure.

jsmadis: I will go with permitting only the `description name redirect_url extra_fields`. I will close this PR and create a new one since there is a change request from Hery and it's blocking the merging.

PR #2770, opened 2021-11-24
Remove "context" from files under test/

josemigallas:
This is a first step to remove `shoulda-context` from porta.

### DIFF
--- a/test/unit/user_test.rb
+++ b/test/unit/user_test.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 require 'test_helper'
 
 class UserTest < ActiveSupport::TestCase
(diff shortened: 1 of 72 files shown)


test/unit/user_test.rb
```diff
@@ -19,16 +21,11 @@ class UserTest < ActiveSupport::TestCase
...
-    # will be saved in DB as 1512012308
-    Timecop.freeze(Time.parse('2017-11-23 03:25:08 UTC +00:00'))
   end
 
-  def teardown
-    Timecop.return
-  end
+  # def teardown
+  #   Timecop.return
+  # end
```
akostadinov, line 28: why comment?

josemigallas: There are probably many things left undone so let's split this into smaller PRs to make it easier for everyone.

PR #2771, opened 2021-11-24
[Params] Fixes app/controllers/master controllers

nidhi-soni1104:
Fixes usage of params in the controller
Fixed rubocop offenses

### DIFF
--- a/app/controllers/admin/api/applications_controller.rb
+++ b/app/controllers/admin/api/applications_controller.rb
@@ -79,7 +79,7 @@ class Admin::Api::ApplicationsController < Admin::Api::BaseController
       applications.joins(:service).where("(services.backend_version <> '1' AND cinstances.application_id = ?)", app_id).first!
 
                      else
-      applications.find(params[:application_id] || params[:id])
+      applications.find(params.require(:application_id) || params.require(:id))
     end
   end
 
--- a/app/controllers/admin/api/buyers_applications_controller.rb
+++ b/app/controllers/admin/api/buyers_applications_controller.rb
@@ -20,7 +20,7 @@ class Admin::Api::BuyersApplicationsController < Admin::Api::BuyersBaseControlle
   ##~ op.parameters.add @parameter_account_id_by_id_name
   #
   def index
-    respond_with(applications.by_state(params.require(:state)))
+    respond_with(applications.by_state(params[:state]))
   end
 
   ##~ op = e.operations.add
@@ -47,11 +47,11 @@ class Admin::Api::BuyersApplicationsController < Admin::Api::BuyersBaseControlle
   def create
     application = applications.new(user_account: buyer, plan: application_plan, create_origin: "api")
     application.unflattened_attributes = application_params
-    application.user_key = local_params[:user_key] if local_params[:user_key]
-    application_id_param = local_params[:application_id]
+    application.user_key = params[:user_key] if params[:user_key]
+    application_id_param = params[:application_id]
     application.application_id = application_id_param if application_id_param
 
-    Array(local_params[:application_key]).each do |key|
+    Array(params[:application_key]).each do |key|
       application.application_keys.build(value: key)
     end
 
@@ -98,7 +98,7 @@ class Admin::Api::BuyersApplicationsController < Admin::Api::BuyersBaseControlle
   #
   def update
     application.unflattened_attributes = flat_params
-    user_key_param = local_params[:user_key]
+    user_key_param = params[:user_key]
     application.user_key = user_key_param if user_key_param
 
     application.save
@@ -140,7 +140,7 @@ class Admin::Api::BuyersApplicationsController < Admin::Api::BuyersBaseControlle
   ## op.parameters.add :name => "app_id", :description => "app_id of the application (for app_id/app_key and oauth authentication modes).", :dataType => "string", :allowMultiple => false, :required => false, :paramType => "query"
   #
   def find
-    application = buyer.bought_cinstances.joins(:service).where("(services.backend_version = '1' AND cinstances.user_key = ?) OR (services.backend_version <> '1' AND cinstances.application_id = ?)", local_params[:user_key], local_params[:app_id]).first!
+    application = buyer.bought_cinstances.joins(:service).where("(services.backend_version = '1' AND cinstances.user_key = ?) OR (services.backend_version <> '1' AND cinstances.application_id = ?)", params[:user_key], params[:app_id]).first!
 
     respond_with application
   end
@@ -268,12 +268,6 @@ class Admin::Api::BuyersApplicationsController < Admin::Api::BuyersBaseControlle
     application.resume
 
     respond_with application
-  end
-
-  private
-
-  def local_params
-    params.permit(%i[user_key application_id application_key app_id]).to_h
   end
 
   protected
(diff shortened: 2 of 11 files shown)


app/controllers/admin/api/applications_controller.rb
```diff
@@ -79,7 +79,7 @@ def application
       applications.joins(:service).where("(services.backend_version <> '1' AND cinstances.application_id = ?)", app_id).first!
 
                      else
-      applications.find(params[:application_id] || params[:id])
+      applications.find(params.require(:application_id) || params.require(:id))
```
jsmadis, line 82: this will break the logic

You can't require one parameter and use another if he is missing.
If you pass the params without `application_id` it will raise an Error, but it should use the `params[:id]` value.

app/controllers/admin/api/buyers_applications_controller.rb
```diff
@@ -46,7 +48,8 @@ def create
     application = applications.new(user_account: buyer, plan: application_plan, create_origin: "api")
     application.unflattened_attributes = application_params
     application.user_key = params[:user_key] if params[:user_key]
-    application.application_id = params[:application_id] if params[:application_id]
+    application_id_param = params[:application_id]
```
jsmadis, line 51: This is not fixing the usage of the params.

app/controllers/admin/api/buyers_applications_controller.rb
```diff
@@ -95,7 +98,8 @@ def show
   #
   def update
     application.unflattened_attributes = flat_params
-    application.user_key = params[:user_key] if params[:user_key]
+    user_key_param = params[:user_key]
```
jsmadis, line 101: This as well.

app/controllers/master/api/finance/billing_jobs_controller.rb
```diff
@@ -45,6 +45,6 @@ def provider
   end
 
   def billing_params
-    @billing_params ||= params.permit(:provider_id, :account_id, :date)
+    @billing_params ||= params.permit(%i[provider_id account_id date])
```
jsmadis, line 48: This will break in rails 5.1.

You need to return it as Hash

app/controllers/master/events/imports_controller.rb
```diff
@@ -13,6 +15,7 @@ def create
   private
 
   def check_shared_secret
-    head(403) unless params[:secret] == Events.shared_secret
+    permitted_params = params.permit(%i[secret])
+    head(403) unless permitted_params[:secret] == Events.shared_secret
```
jsmadis, line 19: This will break in rails 5.1
```suggestion
    permitted_params = params.permit(:secret).to_h
    head(403) unless permitted_params[:secret] == Events.shared_secret
```

app/controllers/master/service_discovery/auth_controller.rb
```diff
@@ -11,7 +11,7 @@ def show
   protected
 
   def self_domain_url(self_domain)
-    options = params.permit(:code, :referrer, :state).merge(host: self_domain, controller: 'provider/admin/service_discovery/auth', action: :show)
+    options = params.permit(%i[code referrer state]).merge(host: self_domain, controller: 'provider/admin/service_discovery/auth', action: :show).to_h
```
jsmadis, line 14:
```suggestion
    options = params.permit(%i[code referrer state]).to_h.merge(host: self_domain, controller: 'provider/admin/service_discovery/auth', action: :show)
```

jsmadis (changes requested):

app/controllers/admin/api/buyers_applications_controller.rb
josemigallas, line 51: Why don't we use `require` here and [here](https://github.com/3scale/porta/pull/2771/files#diff-6d426f7a2ef18db9c2765aac3763814c7c2a8eca35c260faa1e98f31523a3f29R101)?
```suggestion
    application_id_param = params.require(:application_id)
```

nidhi-soni1104: This is code climate issue

app/controllers/admin/api/buyers_applications_controller.rb
nidhi-soni1104: This is too due to code climate issue

app/controllers/admin/api/buyers_applications_controller.rb
jsmadis: @josemigallas it would break the logic. `application_id` is not a required parameter. (Application can also have only user_key without  app_id)

You can also see it from:

```
 ##~ op.parameters.add :name => "application_id", :description => "App ID or Client ID (for OAuth and OpenID Connect authentication modes) of the application to be created.", :dataType => "string", :allowMultiple => false, :required => false, :paramType => "query"
```

app/controllers/admin/api/buyers_applications_controller.rb
jsmadis: I am talking about `params[:user_key]`. That's what we are fixing.

Since the `user_key` is not permitted it will produce `unable to convert unpermitted parameters to hash (ActionController::UnfilteredParameters)`

app/controllers/admin/api/applications_controller.rb
nidhi-soni1104: applications.find(params.require(:application_id) || params[:id])

Are you expecting this?

app/controllers/admin/api/buyers_applications_controller.rb
nidhi-soni1104: Yes, I tried permitting all params and then require but tests failed when I did so.

def local_params
    params.permit!.to_h
  end
 
 user_key_param = local_params.require(:user_key)

jsmadis: That's not correct usage. 
You can call `require` to variable which type is `ActionController::StrongParameters`.

In this case, `local_params` returns a `Hash` with all permitted parameters. We are using it because we can accept dynamic data, we might not know all the possible parameters, and so on...

In this context it would be used as: 

```
def local_params
  params.permit!.to_h
end
 
user_key_param = local_params[:user_key]
```


We are using `require` when the params MUST contain the parameter. That's why we are using it for getting the values of id's.

app/controllers/admin/api/applications_controller.rb
jsmadis: No, you still break the logic there with the `require`

```suggestion
      id_params = params.permit(%i[application_id id]).to_h
      applications.find(id_params[:application_id] || id_params[:id])
```

hallelujah: See my comment in this PR https://github.com/3scale/porta/pull/2778#discussion_r756826572

app/controllers/admin/api/buyers_applications_controller.rb
hallelujah: What code climate issue? Probably a rubocop one right?
Rubocop is a script with an algorithm that helps detect bad practice. Sometimes it does not apply because it is not perfect. In every case, ask yourself if it makes sense to correct those (following good practice). Maybe it is not necessary in this case, aka false positive.

@nidhi-soni1104  we do not know which cop it violates, could you please write more details for the reviewers to understand and help you better?

nidhi-soni1104: @hallelujah PR is failed due to code climate issue as it says params[:application_id] is repeated twice  so I fixed it.

app/controllers/admin/api/buyers_applications_controller.rb
nidhi-soni1104: Okay I got it.

PR #2774, opened 2021-11-24
[Params] Fixes params usage in devel portal (rest of the controllers)

jsmadis:

Fixes params usage in the controllers.


jsmadis:

PR #2775, opened 2021-11-25
[Params] Fixes params in controllers/provider controlllers

nidhi-soni1104:
Fixes usage of params in the controllers/provider controlllers
Fixed rubocop offenses

### DIFF
--- a/app/controllers/provider/invitee_signups_controller.rb
+++ b/app/controllers/provider/invitee_signups_controller.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 class Provider::InviteeSignupsController < FrontendController
   skip_before_action :login_required
 
@@ -10,8 +12,7 @@ class Provider::InviteeSignupsController < FrontendController
 
   layout 'provider/login'
 
-  def show
-  end
+  def show; end
 
   def create
     @user.admin_sections = domain_account.provider_can_use?(:service_permissions) ? [] : ['monitoring']
@@ -40,7 +41,7 @@ class Provider::InviteeSignupsController < FrontendController
   end
 
   def find_invitation
-    @invitation = domain_account.invitations.pending.find_by_token!(invitation_token)
+    @invitation = domain_account.invitations.pending.find_by!(token: invitation_token)
   end
 
   def can_create?
@@ -49,7 +50,7 @@ class Provider::InviteeSignupsController < FrontendController
   end
 
   def build_user
-    @user = @invitation.make_user(params[:user] || {})
+    @user = @invitation.make_user(params.require(:user) || {})
 
     # This is just a sanity guard added when splitting invitation
     # controllers. Remove when SURE.
--- a/app/controllers/provider/passwords_controller.rb
+++ b/app/controllers/provider/passwords_controller.rb
@@ -1,11 +1,13 @@
+# frozen_string_literal: true
+
 class Provider::PasswordsController < FrontendController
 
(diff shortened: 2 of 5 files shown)


app/controllers/provider/invitee_signups_controller.rb
```diff
@@ -49,7 +50,7 @@ def can_create?
   end
 
   def build_user
-    @user = @invitation.make_user(params[:user] || {})
+    @user = @invitation.make_user(params.require(:user) || {})
```
hallelujah, line 53: This is not the same behaviour.

`params.require` raises an exception if it is not present.
What we need here is list all attributes 
```ruby
    permitted_attributes = ????? # Need to list all of the permitted ones to create a User from an invitation
    user_attributes = params.permit(user: permitted_attributes) || {}
    @user = @invitation.make_user(user_attributes)
```

app/controllers/provider/passwords_controller.rb
```diff
@@ -28,15 +30,15 @@ def destroy
   end
 
   def show
-    if params[:password_reset_token].present? && session[:password_reset_token].blank?
+    if params.require(:password_reset_token).present? && session[:password_reset_token].blank?
```
hallelujah, line 33: `params.require` raises an exception if present

app/controllers/provider/passwords_controller.rb
```diff
@@ -28,15 +30,15 @@ def destroy
...
       new_token = @user.generate_lost_password_token
       session[:password_reset_token] = new_token
       redirect_to provider_password_path
     end
   end
 
   def update
-    user = params[:user]
+    user = params.require(:user)
```
hallelujah, line 41: You would need to use permit correctly here

```suggestion
    user = params.permit(user: %i[password password_confirmation])
```

app/controllers/provider/sessions_controller.rb
```diff
@@ -87,12 +88,13 @@ def auth_params
   end
 
   def sso_params
-    params.permit(:token, :expires_at, :redirect_url, :system_name, :code).merge(request: request)
+    params.permit(%i[token expires_at redirect_url system_name code]).merge(request: request).to_h
```
hallelujah, line 91: Why `to_h`? It is already permitted so it should not be an issue. If it is because of a warning with unsecure redirect then OK

app/controllers/provider/sessions_controller.rb
```diff
@@ -87,12 +88,13 @@ def auth_params
...
   end
 
   def session_return_to
-    if params[:return_to]
-      return_to = safe_return_to(params[:return_to])
+    return_to_params = params.require(:return_to)
```
hallelujah, line 95: params.require will raise an error if empty. It is unlikely that you want that behaviour

app/controllers/provider/signups_controller.rb
```diff
@@ -21,7 +23,7 @@ def show # original iframe form
     @provider = master.providers.build
     @user     = @provider.users.build_with_fields
     @plan     = plan
-    @signup_origin = default_params[:origin] || default_params[:signup_origin]
+    @signup_origin = default_params[:origin]|| default_params[:signup_origin]
```
hallelujah, line 26: Typo :)

```suggestion
    @signup_origin = default_params[:origin] || default_params[:signup_origin]
```

hallelujah (changes requested):

PR #2778, opened 2021-11-25
[params] Fixes controllers/partners controllers

nidhi-soni1104:
Fixes usage of params in the controllers/partners controlllers
Fixed rubocop offenses and codeclimate issues

### DIFF
--- a/app/controllers/partners/providers_controller.rb
+++ b/app/controllers/partners/providers_controller.rb
@@ -8,7 +8,7 @@ class Partners::ProvidersController < Partners::BaseController
     signup_result = Signup::ProviderAccountManager.new(Account.master).create(signup_params) do |result|
       account = result.account
       account.signup_mode!
-      account.subdomain = "#{params[:subdomain]}-#{partner_system_name}"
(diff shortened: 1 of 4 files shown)


hallelujah (changes requested):
Introducing `params.require` changes the behavior see your tests are failing.
I think it is fine to NOT follow rubocop sometimes ;) You can manually disable it by a comment

app/controllers/partners/providers_controller.rb
```diff
@@ -56,27 +56,27 @@ def account_params
   def user_params
     {
       signup_type: partner.signup_type,
-      password: params[:password].presence || SecureRandom.hex,
-      email: params[:email],
-      first_name: params[:first_name],
-      last_name: params[:last_name],
+      password: params.require(:password).presence || SecureRandom.hex,
```
jsmadis, line 59: If the password is not present we will use `SecureRandom.hex`.

If you use `require` it will raise an error which destroys the logic here. 
You need to permit it and check it presence.

PR #2782, opened 2021-11-25
[Params] Fixes params usage in the sites controllers

jsmadis:

Fixes usage of params in the controllers to be compatible with rails 5.1.


jsmadis:

PR #2788, opened 2021-11-25
[Params] Refactors params usage in the admin api controllers

jsmadis:

Refactors usage of params to have controllers more secure.

### DIFF
--- a/app/controllers/admin/api/buyers_applications_controller.rb
+++ b/app/controllers/admin/api/buyers_applications_controller.rb
@@ -94,7 +94,7 @@ class Admin::Api::BuyersApplicationsController < Admin::Api::BuyersBaseControlle
   ##~ op.parameters.add @parameter_extra
   #
   def update
-    application.unflattened_attributes = flat_params
+    application.unflattened_attributes = flat_params.permit(application_attributes)
     application.user_key = params[:user_key] if params[:user_key]
 
     application.save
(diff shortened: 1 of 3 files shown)


app/concerns/metric_params.rb
```diff
@@ -10,11 +10,11 @@ module MetricParams
     private
 
     def create_params
-      params.fetch(:metric).permit(DEFAULT_PARAMS | %i[system_name])
+      params.fetch(:metric).permit(DEFAULT_PARAMS | %i[system_name]).to_h
```
hallelujah, line 13: I do not get why we need `to_h`
Could you explain?

jsmadis: I have made a mistake with first assumption of the issue. It's not necessary

app/controllers/admin/fields_definitions_controller.rb
```diff
@@ -76,8 +75,14 @@ def sort
 
   private
 
+  DEFAULT_PARAMS = %i[target label required hidden read_only choices_for_views].freeze
+
   def field_definition_params
-    params[:fields_definition] || {}
+    params.require(:fields_definition).permit(DEFAULT_PARAMS | %i[name])
```
akostadinov, line 81: `field_definition_params` is also used in `#new`. I guess that `:fields_definition` will not be provided in such call so the `require` call will fail. Have you checked that?

btw, will it be safe to remove `attr_protected :account_id, :tenant_id` from `app/models/fields_definition.rb`?

app/controllers/admin/api/buyers_applications_controller.rb
```diff
@@ -282,11 +282,12 @@ def application_plan
   end
 
   def application_params
-    flat_params.slice(*application_attributes)
+    flat_params.permit(application_attributes | %w[user_key application_id plan_id])
```
akostadinov, line 285: Seems like `plan_id` was not needed in the past. Would you explain why add it now?

PR #2791, opened 2021-11-26
Remove `context` and `should` from invoices_test

josemigallas:

#### Affected files

* test/integration/finance/api/invoices_test.rb

#### JIRA

### DIFF
--- a/test/integration/finance/api/invoices_test.rb
+++ b/test/integration/finance/api/invoices_test.rb
@@ -6,7 +6,7 @@ module Finance::Api
   class InvoicesTestCommonCases < ActionDispatch::IntegrationTest
     def setup
       @provider = FactoryBot.create(:provider_account)
-      @buyer = FactoryBot.create(:buyer_account, :provider_account => @provider)
+      @buyer = FactoryBot.create(:buyer_account, provider_account: @provider)
       @provider.create_billing_strategy
       @provider.save!
       @key = @provider.api_key
@@ -20,7 +20,7 @@ module Finance::Api
     end
 
     test 'deny access if finance module is disabled' do
-      without_finance = FactoryBot.create(:provider_account, :billing_strategy => nil)
+      without_finance = FactoryBot.create(:provider_account, billing_strategy: nil)
       host! without_finance.self_domain
       get "/api/#{@context}invoices.xml?provider_key=#{without_finance.api_key}"
       assert_response :forbidden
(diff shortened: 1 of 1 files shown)


test/integration/finance/api/invoices_test.rb
```diff
@@ -1,210 +1,199 @@
...
+    end
+
+    test "have vat_rate if non negative" do
+      setup_existing_invoices_test
+      @buyer.vat_rate = 10.0
+      @buyer.save
+
+      get "/api/#{@context}invoices.xml?provider_key=#{@key}"
 
-      context 'with existing invoices' do
```
akostadinov, line 60: for `      context 'with existing invoices' do` you could create a sub-class to hold the setup method but current approach with `#setup_existing_invoices_test` is reasonable. Just saying.

test/integration/finance/api/invoices_test.rb
```diff
@@ -1,210 +1,199 @@
...
+        FactoryBot.create(:invoice, provider_account: @provider, buyer_account: buyer)
       end
+    end
+  end
+
+  class InvoicesNoScopeTest < InvoicesTestCommonCases
+    def setup
+      super
+      @context = ''
+    end
```
akostadinov, line 184: nice!

test/integration/finance/api/invoices_test.rb
```diff
@@ -1,210 +1,199 @@
...
+    def setup
+      super
+      @context = "accounts/#{@buyer.id}/"
     end
+  end
+
+  def self.runnable_methods
+    return [] if self == InvoicesTestCommonCases
 
+    super
```
akostadinov, line 197: Actually is it possible to rename `InvoicesTestCommonCases` to `InvoicesUnscopedTest`, then `InvoicesBuyersScopeTest` inherit from `InvoicesUnscopedTest`. In this way it will not be needed to deal with implementation details like `#runnable_methods`. WDYT?

akostadinov (approved):
I have one suggested improvement but it is not important. Please, in case you decide change things before merging, add new commits instead of amending so it is easier to see only the changes. It is very hard to review the whole thing again.

test/integration/finance/api/invoices_test.rb
josemigallas: Absolutely, good catch!

test/integration/finance/api/invoices_test.rb
josemigallas: No, you're right, sub-classes is better in this case.

hallelujah (approved):

akostadinov (approved):

PR #2793, opened 2021-11-26
Remove `context` and `should` from integration/stats

josemigallas:

#### Affected files

* test/integration/stats/applications_test.rb
* test/integration/stats/services_test.rb

#### JIRA

### DIFF
--- a/test/integration/stats/services_test.rb
+++ b/test/integration/stats/services_test.rb
@@ -11,6 +11,7 @@ class Stats::ServicesTest < ActionDispatch::IntegrationTest
     Stats::Base.storage.flushdb
 
     host! @provider_account.admin_domain
+    provider_login_with @provider_account.admins.first.username, 'supersecret'
   end
 
   def teardown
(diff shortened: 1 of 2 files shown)


test/integration/stats/services_test.rb
```diff
@@ -53,64 +51,59 @@ def teardown
...
-      assert_content_type 'application/json'
-      assert_json_contains(
-      "period" => {"name"=>"day",
-      "granularity"=>"hour",
-      "since"=>Time.zone.parse("2009-12-11T07:00:00+12:00"),
-      "until"=>Time.zone.parse("2009-12-12T07:59:59+12:00"),
-      "timezone"=>"Asia/Kamchatka"}
-      )
-    end
+    provider_login_with @provider_account.admins.first.username, 'supersecret'
```
akostadinov, line 96: These seem identical with `with simple plan and cinstance, retrieve data with account timezone`, perhaps sub-class them?

akostadinov (changes requested):
Looks great. I have just a small DRY request.

test/integration/stats/services_test.rb
josemigallas: You're right. I've spotted another line shared by all test cases that I'll move to the setup.

josemigallas: And the problem with grouping in sub-classes is that either they inherit from the main class, sharing setup/teardown and *running every test twice* or they inherit from `ActionDispatch::IntegrationTest` and we need duplicate setups

hallelujah (approved):

test/integration/stats/services_test.rb
hallelujah: Exactly this

> And the problem with grouping in sub-classes is that either they inherit from the main class, sharing setup/teardown and running every test twice or they inherit from ActionDispatch::IntegrationTest and we need duplicate setups

Subclassing is sometimes not what you want to do, let's not be too picky and accept even if you have to duplicate things.
Remember the goal is to get rid of shoulda, not refactoring nicely. Of course if can make it better, this is always welcome.

akostadinov (approved):

PR #2790, opened 2021-11-26
[THREESCALE-7898] tag list presentation

akostadinov:
partial fix for https://issues.redhat.com/browse/THREESCALE-7898

Not sure how to add instructions for users.

I need to add some tests later.

PR #2798, opened 2021-11-29
Add hint for Tag list input

damianpm:

- Adding a hint for the **Tag list** input, in Developer portal **Pages** and **Files**
- Add tests for both pages
 
Closes https://github.com/3scale/porta/pull/2790

Screenshot 2021-11-29 at 12 10 46

Jira issues:

https://issues.redhat.com/browse/THREESCALE-7898
https://issues.redhat.com/browse/THREESCALE-7935

**Steps to reproduce:**

- Go to **Audience > Developer Portal > Content**
- Select a file (eg: mouse.jpg)
- Write some tags
- Click save

### DIFF
--- a/features/old/cms/pages.feature
+++ b/features/old/cms/pages.feature
@@ -39,7 +39,7 @@ Feature: CMS Pages
         | Path           | /potato       |
         | Content type   | text/css      |
         | Handler        | markdown      |
-        | Tag list       | potato salad |
+        | Tag list       | potato, salad |
 
        And I press "Publish"
       Then I should see "Page saved and published"
--- a/app/views/provider/admin/cms/files/_form.html.erb
+++ b/app/views/provider/admin/cms/files/_form.html.erb
@@ -3,5 +3,5 @@
   <%= f.input :path, input_html: {class: 'cms-path-autocomplete'}  %>
   <%= f.input :downloadable, :hint => "Checked sets the content-disposition to 'attachment.'" %>
   <%= f.input :attachment, :as => :file %>
-  <%= f.input :tag_list %>
+  <%= f.input :tag_list, :input_html => { :value => @file.tag_list.join(", ") } %>
 <% end %>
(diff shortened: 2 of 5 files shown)


akostadinov:
That's cool. Could you add some tests that tags are properly saved and later shown comma separated? e.g. in `features/old/cms/pages.feature` , for files I don't see an existing file, maybe I just can't spot.

hallelujah (approved):

features/old/cms/pages.feature
```diff
@@ -39,7 +39,7 @@ Feature: CMS Pages
         | Path           | /potato       |
         | Content type   | text/css      |
         | Handler        | markdown      |
-        | Tag list       | potato salad |
+        | Tag list       | potato, salad |
```
hallelujah, line 42: So before it was space separated list?
If so then this is a major change, we should verify that we do not break customer workflow

hallelujah (changes requested):

features/old/cms/pages.feature
akostadinov: Before it showed space separated, but when you saved, it was saves as a single tag. So customer saves as "tag1, tag2, tag3", then on page reload it appeared as "tag1 tag2 tag3" and if saved again would save as a single tag "tag1 tag2 tag3". Basically now we avoid customer messing up tags by mistake and document how to split tags properly.

akostadinov: see https://issues.redhat.com/browse/THREESCALE-7898

akostadinov:
Looks great! Can you try to write some File UI test? I now found features/provider/cms/files.feature, maybe inside it.

akostadinov (dismissed):
Looks nice. Thank you! could you point pull request to master branch and close mine?

damianpm:
@akostadinov @hallelujah PR updated, please take a look

akostadinov (approved):
Thank you!

PR #2799, opened 2021-11-29
Remove `context` and `should` from fields_definition_test

josemigallas:

#### Affected files

* test/unit/fields_definition_test.rb

#### JIRA

### DIFF
--- a/test/unit/fields_definition_test.rb
+++ b/test/unit/fields_definition_test.rb
@@ -2,272 +2,276 @@
 
 require 'test_helper'
 
-def new_field_definition(**opts)
-  FieldsDefinition.new(account: @provider, target: 'User', name: 'new_field', label: 'lal', **opts)
-end
-
 class FieldsDefinitionTest < ActiveSupport::TestCase
-  setup do
-    @provider = FactoryBot.create(:provider_account)
+  class OtherTest < FieldsDefinitionTest
+    setup do
+      @provider = FactoryBot.create(:provider_account)
+    end
+
+    test 'visible_for?' do
+      provider_user = @provider.first_admin
+
+      buyer_user = FactoryBot.create(:buyer_account).first_admin
+      other_provider_user = FactoryBot.create(:provider_account).first_admin
+
+      #
+      # With a hidden field
+      #
+      fd = new_field_definition(hidden: true)
+
+      # For a buyer
+      assert_not fd.visible_for?(buyer_user)
+
+      # For other provider account
+      assert_not fd.visible_for?(other_provider_user)
+
+      # For the owner
+      assert fd.visible_for?(provider_user)
+
+      #
+      # With a visible field
+      #
+      fd = new_field_definition(hidden: false)
+
+      # For a buyer
+      assert fd.visible_for?(buyer_user)
+
+      # For other provider account
+      assert fd.visible_for?(other_provider_user)
+
+      # For the owner
+      assert fd.visible_for?(provider_user)
+    end
+
+    test 'not set unregistered target' do
+      fd = FieldsDefinition.new
+      assert FieldsDefinition.targets.exclude?("Foo")
+
+      fd.target = "Foo"
+      assert fd.target.nil?
+    end
+
+    test "not in database modificable fields not allow hidden AND required" do
+      fd = new_field_definition(hidden: true, required: true)
+      assert_not fd.valid?
+      fd.hidden = false
+      assert fd.valid?
+    end
+
+    test "not in database modificable fields not allow readonly AND required" do
+      fd = new_field_definition(read_only: true, required: true)
+      assert_not fd.valid?
+      fd.read_only = false
+      assert fd.valid?
+    end
+
+    test "not in database modificable fields allow choices property" do
+      a = new_field_definition(choices: %w[a b])
+      assert a.valid?
+    end
   end
 
-  test 'visible_for?' do
-    provider_user = @provider.first_admin
+  class ValidationTest < FieldsDefinitionTest
+    def setup
+      @provider = FactoryBot.create(:provider_account)
+      FieldsDefinition.where(account_id: @provider.id).delete_all # removing default ones
 
-    buyer_user = FactoryBot.create(:buyer_account).first_admin
-    other_provider_user = FactoryBot.create(:provider_account).first_admin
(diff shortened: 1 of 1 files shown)


test/unit/fields_definition_test.rb
```diff
@@ -1,32 +1,40 @@
+# frozen_string_literal: true
+
 require 'test_helper'
 
+def new_field_definition(**opts)
+  FieldsDefinition.new(account: @provider, target: 'User', name: 'new_field', label: 'lal', **opts)
+end
```
akostadinov, line 7: why is this outside classes? Can it be in one of the classes or in a helper class or something?

test/unit/fields_definition_test.rb
```diff
@@ -38,262 +46,228 @@ class FieldsDefinitionTest < ActiveSupport::TestCase
...
-      assert_raise ActiveRecord::RecordInvalid do
-        FieldsDefinition.create!(:account => @provider, :name => 'org_name', :target => 'Account', :label => 'bar2')
-      end
+    @fd = new_field_definition(label: 'lal', target: 'User', name: 'first_name', choices: %w[a b])
+  end
+
+  test 'validate presence of' do
+    assert validate_presence_of :name
+    assert validate_presence_of :label
+    assert validate_presence_of :target
```
akostadinov, line 88: where do these checks come from?

test/unit/fields_definition_test.rb
```diff
@@ -38,262 +46,228 @@ class FieldsDefinitionTest < ActiveSupport::TestCase
...
-      FieldsDefinition.delete_all
-      FactoryBot.create(:fields_definition, :account => @provider, :name => "hidden",
-              :hidden => true)
-      FactoryBot.create(:fields_definition, :account => @provider, :name => "read_only",
-              :read_only => true)
-      @public_field = FactoryBot.create(:fields_definition, :account => @provider,
-                              :name => "public")
-      @buyer = FactoryBot.create :buyer_account, :provider_account => @provider
-    end
+  test "builtin fields and are modifiable fields not allow hidden AND required" do
```
akostadinov, line 155: I can't parse this test name, do you know what it should mean? :)

akostadinov:
```suggestion
  test "builtin modifiable fields not allow hidden AND required" do
```

test/unit/fields_definition_test.rb
```diff
@@ -38,262 +46,228 @@ class FieldsDefinitionTest < ActiveSupport::TestCase
...
+    a = new_field_definition(hidden: true, required: true)
+    assert_not a.valid?
+  end
 
-    should 'return all fields for provider users' do
-      assert FieldsDefinition.editable_by(@provider.users.first) ==
-	           FieldsDefinition.all
-    end
+  test "builtin fields and are modifiable fields not allow readonly AND required" do
```
akostadinov, line 160:
```suggestion
  test "builtin modifiable fields not allow readonly AND required" do
```

test/unit/fields_definition_test.rb
```diff
@@ -38,262 +46,228 @@ class FieldsDefinitionTest < ActiveSupport::TestCase
...
+    a = new_field_definition(read_only: true, required: true)
+    assert_not a.valid?
+  end
 
-    should 'return only public fields to buyer users' do
-      assert FieldsDefinition.editable_by(@buyer.users.first) == [@public_field]
-    end
+  test "builtin fields and are modifiable fields allow choices property only on string fields" do
```
akostadinov, line 165:
```suggestion
  test "builtin modifiable fields allow choices property only on string fields" do
```

akostadinov (changes requested):

test/unit/fields_definition_test.rb
josemigallas: From the gem `shoulda-matchers`

test/unit/fields_definition_test.rb
josemigallas: Absolutely no idea haha. I think we should probably find a better test name? Like, what's a _modifiable field_ anyway? And what's _builtin_ and what isn't.
I guess we could simply say _field definition_? @hallelujah any ideas/opinions?

test/unit/fields_definition_test.rb
akostadinov: Aren't we removing that? Or not?

test/unit/fields_definition_test.rb
josemigallas: I don't think this should go in a helper, it's too silly of a method. I would put it all inside a `FieldsDefinitionTest` class or module, but then I'd need to rename the current [`FieldsDefinitionTest`](https://github.com/3scale/porta/pull/2799/files#diff-e8951b1c6f578bd6a73ca3802c1353151dbc036e0e630e6e69ab87875f8cc015R9) to something like `OtherFieldsDefinitionTest` 🤮

akostadinov: What you have done seems reasonable to me. While you can define a new sub-class with the method and have other sub-classes inherit from it, naming would again be tricky. btw I think you can make the method `private`, not that it matters in this case.

test/unit/fields_definition_test.rb
josemigallas: Only `shoulda-context` at the moment.

akostadinov (approved):
Thank you!

test/unit/fields_definition_test.rb
akostadinov: Actually I'm not sure. Now I saw unstable:
https://app.circleci.com/pipelines/github/3scale/porta/19182/workflows/b9d5fa7b-eebb-4ba6-b839-171b8c555b19/jobs/229034/tests

Also I see a deprecation warning from `shoulda-matchers`:
```
~/.asdf/installs/ruby/2.6.7/lib/ruby/gems/2.6.0/gems/shoulda-matchers-2.8.0/lib/shoulda/matchers/active_model/validate_inclusion_of_matcher.rb:251: warning: BigDecimal.new is deprecated; use BigDecimal() method instead.
```

Investigating.

akostadinov: Ok, the failures were not caused by shoulda [1] but by test order. The deprecation is an issue.

[1] https://github.com/3scale/porta/pull/2820/commits/ad30893024dc0fd2f0b5c7310566c4e13060d242

PR #2800, opened 2021-11-29
Prawn memory leak THREESCALE-8120

hallelujah:
Update prawn gem to fix https://github.com/prawnpdf/prawn/issues/424

- Update prawn gem to 2.2.2
- Fixes prawnpdf/prawn#424
- Fixes license issues prawnpdf/prawn#970. TTF fonts are not distributed in the bundled gem anymore


josemigallas (dismissed):

app/lib/pdf/report.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-require 'prawn/core'
+require 'prawn'
```
akostadinov, line 3: Isn't this required automatically?

akostadinov (dismissed):

Gemfile.lock
```diff
@@ -1605,9 +1595,15 @@ GEM
...
-      prawn-core
-    prawn-layout (0.2.1)
+    prawn (2.4.0)
+      pdf-core (~> 0.9.0)
+      ttfunk (~> 1.7)
+    prawn-markup (0.3.3)
+      nokogiri
+      prawn
+      prawn-table
+    prawn-table (0.2.2)
```
akostadinov, line 1605: 4 years without commits, doesn't [prawn-markup](https://github.com/puzzle/prawn-markup) handle tables well enough?

Just asking, probably you already checked.

akostadinov (approved):
If it works, looks awesome. Just this `prawn-table` seems abandoned and a potential future liability but if it is necessary, then it's fine.

josemigallas:
@akostadinov assigning this to you since you were the most involved besides Hery. I think there is no rush but we should not forget about it either.

Gemfile.lock
```diff
@@ -1605,9 +1595,14 @@ GEM
       net-http-persistent
     polyamorous (1.3.3)
       activerecord (>= 3.0)
-    prawn-format (0.2.1)
+    prawn (2.4.0)
```
akostadinov, line 1598: just a note to myself this only supports ruby 2.5+

Gemfile.lock
akostadinov: Note to myself - prawn-markup depends on prawn-table. Also prawn-table had seen some care this year.

PR #2802, opened 2021-11-30
[params] fixes providers/plans_controller.rb controller

nidhi-soni1104:
Fixes usage of params in the auth controller

### DIFF
--- a/app/controllers/master/providers/plans_controller.rb
+++ b/app/controllers/master/providers/plans_controller.rb
@@ -18,7 +18,7 @@ class Master::Providers::PlansController < Master::Providers::BaseController
 
 
   def find_new_plan
-    @new_plan = Account.master.application_plans.stock.find(params[:plan_id])
+    @new_plan = Account.master.application_plans.stock.find(params.permit!.to_h[:plan_id])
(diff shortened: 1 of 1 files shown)


app/controllers/master/providers/plans_controller.rb
```diff
@@ -18,7 +18,7 @@ def update
 
 
   def find_new_plan
-    @new_plan = Account.master.application_plans.stock.find(params[:plan_id])
+    @new_plan = Account.master.application_plans.stock.find(params.permit!.to_h[:plan_id])
```
hallelujah, line 21: I would prefer

```suggestion
    @new_plan = Account.master.application_plans.stock.find(params.permit(:plan_id)[:plan_id])
```

hallelujah (changes requested):

app/controllers/master/providers/plans_controller.rb
nidhi-soni1104: Okay

nidhi-soni1104:
@hallelujah Changes done. I  have added .to_h as we discussed that we need to convert to hash but I think its not necessary always as it worked without converting it to hash .

hallelujah (approved):

PR #2804, opened 2021-11-30
[params] Fixes events/imports_controller.rb

nidhi-soni1104:
Fixes usage of params in the imports controller

### DIFF
--- a/app/controllers/master/events/imports_controller.rb
+++ b/app/controllers/master/events/imports_controller.rb
@@ -13,7 +13,7 @@ class Master::Events::ImportsController < Master::BaseController
   private
 
   def check_shared_secret
-    permitted_params = params.permit(%i[secret]).to_h
+    permitted_params = params.permit(%i[secret])
     head(403) unless permitted_params[:secret] == Events.shared_secret
   end
 end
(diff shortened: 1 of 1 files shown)


app/controllers/master/events/imports_controller.rb
```diff
@@ -13,6 +13,7 @@ def create
   private
 
   def check_shared_secret
-    head(403) unless params[:secret] == Events.shared_secret
+    permitted_params = params.permit(%i[secret]).to_h
```
hallelujah, line 16: Why `to_h`?

hallelujah (changes requested):

app/controllers/master/events/imports_controller.rb
nidhi-soni1104: @hallelujah we (@jsmadis  and @josemigallas )discussed in the call that its safe to convert to hash , so if we permit we convert it to hash.

nidhi-soni1104: @hallelujah PR resolved

akostadinov:
Just curious, which test failed without this change? I wonder if it is needed.

hallelujah (approved):

nidhi-soni1104:
@akostadinov Not sure about the test, these are few files listed by Jakub which caused faliures due to params unpermitted issues so fixing those listed files.

PR #2805, opened 2021-11-30
[params] Fixes api/providers_controller.rb

nidhi-soni1104:
Fixes usage of params in the api providers controller controller

### DIFF
--- a/app/controllers/master/api/providers_controller.rb
+++ b/app/controllers/master/api/providers_controller.rb
@@ -187,7 +187,7 @@ class Master::Api::ProvidersController < Master::Api::BaseController
   end
 
   def get_partner
-    @partner = Partner.find_by(system_name: params.require(:partner))
+    @partner = Partner.find_by(system_name: params.permit(:partner)[:partner])
   end
 
   def get_application_plan
--- a/app/controllers/master/api/providers_controller.rb
+++ b/app/controllers/master/api/providers_controller.rb
@@ -154,7 +154,7 @@ class Master::Api::ProvidersController < Master::Api::BaseController
   private
 
   def provider_account
-    @provider_account ||= current_account.providers.without_deleted(!action_includes_deleted_providers?).find(params.require(id))
+    @provider_account ||= current_account.providers.without_deleted(!action_includes_deleted_providers?).find(params.require(:id))
   end
 
   def action_includes_deleted_providers?
@@ -187,12 +187,12 @@ class Master::Api::ProvidersController < Master::Api::BaseController
   end
 
   def get_partner
-    @partner = Partner.find_by_system_name(params.permit!.to_h[:partner])
+    @partner = Partner.find_by(system_name: params.permit!.to_h[:partner])
   end
 
   def get_application_plan
     application_plans = Account.master.application_plans.where(partner_id: @partner.try!(:id))
-    @application_plan = application_plans.find_by_system_name(params.permit!.to_h[:application_plan])
+    @application_plan = application_plans.find_by(system_name: params.permit!.to_h[:application_plan])
     raise ActiveRecord::RecordNotFound if @application_plan.blank?
   end
 
--- a/app/controllers/master/api/providers_controller.rb
+++ b/app/controllers/master/api/providers_controller.rb
@@ -169,7 +169,7 @@ class Master::Api::ProvidersController < Master::Api::BaseController
 
   def update_params
     permitted_params = provider_account.scheduled_for_deletion? ? %i[state_event] : UPDATE_PARAMS
-    params.require(:account).permit(permitted_params).to_h
+    params.require(:account).permit(permitted_params)
   end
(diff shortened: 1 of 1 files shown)


app/controllers/master/api/providers_controller.rb
```diff
@@ -169,7 +169,7 @@ def ensure_master_with_plans
 
   def update_params
     permitted_params = provider_account.scheduled_for_deletion? ? %i[state_event] : UPDATE_PARAMS
-    params.require(:account).permit(permitted_params)
+    params.require(:account).permit(permitted_params).to_h
```
hallelujah, line 172: Why do we need `to_h` ?

app/controllers/master/api/providers_controller.rb
```diff
@@ -187,16 +187,16 @@ def plans
   end
 
   def get_partner
-    @partner = Partner.find_by_system_name(params[:partner])
+    @partner = Partner.find_by(system_name: params.permit!.to_h[:partner])
```
hallelujah, line 190:
```suggestion
    @partner = Partner.find_by(system_name: params.require(:partner))
```

app/controllers/master/api/providers_controller.rb
```diff
@@ -187,16 +187,16 @@ def plans
...
   end
 
   def get_application_plan
     application_plans = Account.master.application_plans.where(partner_id: @partner.try!(:id))
-    @application_plan = application_plans.find_by_system_name(params[:application_plan])
+    @application_plan = application_plans.find_by(system_name: params.permit!.to_h[:application_plan])
```
hallelujah, line 195:
```suggestion
    @application_plan = application_plans.find_by(system_name: params.require(:application_plan))
```

hallelujah (changes requested):

app/controllers/master/api/providers_controller.rb
nidhi-soni1104: (yes)

app/controllers/master/api/providers_controller.rb
nidhi-soni1104: :+1:

app/controllers/master/api/providers_controller.rb
nidhi-soni1104: not needed actually

app/controllers/master/api/providers_controller.rb
```diff
@@ -187,16 +187,16 @@ def plans
...
 
   def get_partner
-    @partner = Partner.find_by_system_name(params[:partner])
+    @partner = Partner.find_by(system_name: params.require(:partner))
   end
 
   def get_application_plan
     application_plans = Account.master.application_plans.where(partner_id: @partner.try!(:id))
-    @application_plan = application_plans.find_by_system_name(params[:application_plan])
+    @application_plan = application_plans.find_by(system_name: params.require(:application_plan))
```
hallelujah, line 195:
```suggestion
    @application_plan = application_plans.find_by(system_name: params.permit(:application_plan)[:application_plan])
```

hallelujah (changes requested):
To keep the same behaviour

hallelujah (approved):

app/controllers/master/api/providers_controller.rb
```diff
@@ -187,16 +187,16 @@ def plans
   end
 
   def get_partner
-    @partner = Partner.find_by_system_name(params[:partner])
+    @partner = Partner.find_by(system_name: params.permit(:partner)[:partner])
```
akostadinov, line 190: Do we need `permit`? `Partner.find_by(system_name: params[:partner])` should be fine as far as I can tell.

app/controllers/master/api/providers_controller.rb
```diff
@@ -187,16 +187,16 @@ def plans
...
   end
 
   def get_application_plan
     application_plans = Account.master.application_plans.where(partner_id: @partner.try!(:id))
-    @application_plan = application_plans.find_by_system_name(params[:application_plan])
+    @application_plan = application_plans.find_by(system_name: params.permit(:application_plan)[:application_plan])
```
akostadinov, line 195: permit seems not needed here as well

hallelujah: I think it is OK. It will prevent some SQLi if `params[:application_plan]` is not a simple value but a Hash ;)
https://github.com/3scale/porta/blob/b0046173d5d11e13c48e7da9fceab7cdc7dd776b/config/initializers/rails_sqli.rb

hallelujah (approved):

app/controllers/master/api/providers_controller.rb
akostadinov: In Rails 5.1 it should not be treated as a Hash in case it is not permitted. We can test that.

PR #2809, opened 2021-12-01
[params] Fixes master/api/services_controller

nidhi-soni1104:
Fixes usage of params in the service controller

### DIFF
--- a/app/controllers/master/api/services_controller.rb
+++ b/app/controllers/master/api/services_controller.rb
@@ -31,10 +31,10 @@ class Master::Api::ServicesController < Master::Api::BaseController
   private
 
   def service
-    @service ||= provider.accessible_services.find(params[:id])
+    @service ||= provider.accessible_services.find(params.permit!.to_h[:id])
   end
 
   def provider
-    @provider ||= Account.providers.find(params[:provider_id])
+    @provider ||= Account.providers.find(params.permit!.to_h[:provider_id])
   end
 end
--- a/app/controllers/master/api/services_controller.rb
+++ b/app/controllers/master/api/services_controller.rb
@@ -35,6 +35,6 @@ class Master::Api::ServicesController < Master::Api::BaseController
   end
 
   def provider
-    @provider ||= Account.providers.find(params.permit!.to_h[:provider_id])
+    @provider ||= Account.providers.find(params.require(:provider_id))
   end
 end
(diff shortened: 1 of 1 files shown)


app/controllers/master/api/services_controller.rb
```diff
@@ -31,10 +31,10 @@ def destroy
   private
 
   def service
-    @service ||= provider.accessible_services.find(params[:id])
+    @service ||= provider.accessible_services.find(params.permit!.to_h[:id])
```
hallelujah, line 34: Why not

```suggestion
    @service ||= provider.accessible_services.find(params.require(:id))
```

app/controllers/master/api/services_controller.rb
```diff
@@ -31,10 +31,10 @@ def destroy
...
   end
 
   def provider
-    @provider ||= Account.providers.find(params[:provider_id])
+    @provider ||= Account.providers.find(params.permit!.to_h[:provider_id])
```
hallelujah, line 38: Same 
```suggestion
    @provider ||= Account.providers.find(params.require(:provider_id))
```

hallelujah (changes requested):

app/controllers/master/api/services_controller.rb
nidhi-soni1104: @hallelujah I preferred permit because what if we reuire id and its missing so may be can break the code.

nidhi-soni1104: params.require raises an exception if it is not present.

As you commented on one of my PR

akostadinov: In this case all usages of `#service` appear to expect it NOT to be `nil`. That's why if we raise early when parameter is missing, that makes more sense than leaving the error for later. And the error message will be more meaningful.

akostadinov: I think that also API documentation can be useful to see what parameters we require. If parameter is required by documentation, in most cases we should better require it in code as well.

nidhi-soni1104: @akostadinov checking API documentation is a good option.

nidhi-soni1104:
@hallelujah Changes done

hallelujah (approved):

PR #2811, opened 2021-12-01
[params] Fixes session controller

nidhi-soni1104:
Fixes usage of params in the session controller

### DIFF
--- a/app/controllers/partners/sessions_controller.rb
+++ b/app/controllers/partners/sessions_controller.rb
@@ -3,7 +3,7 @@ class Partners::SessionsController < Partners::BaseController
   skip_before_action :authenticate!
 
   def openid
-    authenticate_with_open_id(params[:openid_url]) do |result, identity_url, registration|
+    authenticate_with_open_id(params.require(:openid_url)) do |result, identity_url, registration|
       if result.successful?
         @user = User.find_by!(open_id: identity_url)
         @account = @user.account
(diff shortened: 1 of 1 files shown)


app/controllers/partners/sessions_controller.rb
```diff
@@ -3,15 +3,16 @@ class Partners::SessionsController < Partners::BaseController
...
+    authenticate_with_open_id(params.require(:openid_url)) do |result, identity_url, registration|
       if result.successful?
         @user = User.find_by!(open_id: identity_url)
         @account = @user.account
         sso_token = SSOToken.new user_id: @user.id
         sso_token.protocol = 'http' unless request.ssl?
         sso_token.account = @account
         sso_url = sso_token.sso_url!(@account.external_admin_domain)
-        sso_url << "&return_to=#{params[:return_to]}" if params[:return_to].present?
+        return_to_param = params.require(:return_to)
```
hallelujah, line 14: This will raise if `params[:return_to]` is not present

hallelujah (changes requested):

app/controllers/partners/sessions_controller.rb
nidhi-soni1104: @hallelujah we are checking if return_to_param[:return_to].present?

hallelujah: I mean `params.require` is raising if it is not present

See https://api.rubyonrails.org/v5.0/classes/ActionController/Parameters.html#method-i-require

nidhi-soni1104: yes agreed so we can use this without require as it was used before.

`        sso_url << "&return_to=#{params[:return_to]}" if params[:return_to].present?
`

PR #2815, opened 2021-12-01
Remove `context` and `should` and from profile_test

josemigallas:

#### Affected files

* test/unit/profile_test.rb

#### JIRA

### DIFF
--- a/test/unit/profile_test.rb
+++ b/test/unit/profile_test.rb
@@ -1,7 +1,8 @@
+# frozen_string_literal: true
+
 require 'test_helper'
 
 class ProfileTest < ActiveSupport::TestCase
-
   should have_db_column :logo_file_name
   should have_db_column :logo_content_type
   should have_db_column :logo_file_size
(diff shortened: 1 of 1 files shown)


test/unit/profile_test.rb
```diff
@@ -12,139 +13,102 @@ class ProfileTest < ActiveSupport::TestCase
...
+    # assert_not Profile.new(company_size: nil).individual_profile?
+  end
 
-      should 'return false for profiles with company_size set to blank' do
-        assert_equal false, Profile.new(:company_size => "").individual_profile?
-      end
-    end
+  test '#individual_profile? method should return false for profiles with company_size set to blank' do
+    # TODO: validation says company_size can't be blank, so is this test necessary/valid?
+    assert_not Profile.new(company_size: '').individual_profile?
```
akostadinov, line 82: it is possible that we have outdated data in DB so model might be loaded from DB with such data... hypothetically :)

But yes - it makes little sense.

akostadinov (approved):

akostadinov:
failure is caused by #2821

PR #2816, opened 2021-12-01
[params] Fixes partner providers controller

nidhi-soni1104:
Fixes usage of params in the partners providers controller

### DIFF
--- a/app/controllers/partners/providers_controller.rb
+++ b/app/controllers/partners/providers_controller.rb
@@ -8,7 +8,7 @@ class Partners::ProvidersController < Partners::BaseController
     signup_result = Signup::ProviderAccountManager.new(Account.master).create(signup_params) do |result|
       account = result.account
       account.signup_mode!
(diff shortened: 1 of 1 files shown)


app/controllers/partners/providers_controller.rb
```diff
@@ -68,19 +68,22 @@ def user_params
...
+    if @selected_plan = application_plans.find_by(system_name: permitted_params[:application_plan])
       @selected_plan
     else
       application_plans.first
     end
   end
 
+  def permitted_params
+    params.permit!.to_h
+  end
```
hallelujah, line 88: We should list all needed params instead

hallelujah (changes requested):

app/controllers/partners/providers_controller.rb
nidhi-soni1104: @hallelujah changes done

hallelujah (approved):

PR #2818, opened 2021-12-01
[params] Fixes partners base controller

nidhi-soni1104:
Fixes usage of params in the partners base controller

### DIFF
--- a/app/controllers/partners/base_controller.rb
+++ b/app/controllers/partners/base_controller.rb
@@ -9,7 +9,7 @@ class Partners::BaseController < ApplicationController
   attr_reader :partner
 
   def authenticate!
-    @partner = Partner.find_by_api_key(params[:api_key])
+    @partner = Partner.find_by(api_key: params.permit(:api_key)[:api_key])
     render(plain: 'unauthorized', status: 401) unless @partner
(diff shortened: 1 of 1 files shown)


app/controllers/partners/base_controller.rb
```diff
@@ -9,7 +9,7 @@ class Partners::BaseController < ApplicationController
   attr_reader :partner
 
   def authenticate!
-    @partner = Partner.find_by_api_key(params[:api_key])
+    @partner = Partner.find_by(api_key: params.permit!.to_h[:api_key])
```
hallelujah, line 12: We can use `require` here
```ruby
    @partner = Partner.find_by(api_key: params.require(:api_key))
```

or `permit`, better to keep the behaviour
```suggestion
    @partner = Partner.find_by(api_key: params.permit(:api_key)[:api_key])
```

hallelujah (changes requested):

app/controllers/partners/base_controller.rb
nidhi-soni1104: @hallelujah done

hallelujah (approved):

PR #2819, opened 2021-12-01
[params] Fixes partners user controler

nidhi-soni1104:
Fixes usage of params in the partners user controller

### DIFF
--- a/app/controllers/partners/users_controller.rb
+++ b/app/controllers/partners/users_controller.rb
@@ -3,30 +3,30 @@ class Partners::UsersController < Partners::BaseController
   before_action :find_account
 
   def index
-    @users = @account.users.page(params[:page])
-    @users = @users.where(open_id: params[:open_id]) if params[:open_id]
+    @users = @account.users.page(permitted_params[:page])
+    @users = @users.where(open_id: permitted_params[:open_id]) if permitted_params[:open_id]
     render json: @users
   end
 
   def show
-    @user = @account.users.find(params[:id])
+    @user = @account.users.find(permitted_params[:id])
     render json: @user
   end
 
   def destroy
-    @user = @account.users.find(params[:id])
+    @user = @account.users.find(permitted_params[:id])
     @user.destroy
     render json: {success: true}
   end
 
   def create
     @user = @account.users.build
-    @user.email = params[:email]
+    @user.email = permitted_params[:email]
     @user.password = SecureRandom.hex
-    @user.first_name = params[:first_name].presence
-    @user.last_name = params[:last_name].presence
-    @user.open_id = params[:open_id].presence
-    @user.username = params[:username]
+    @user.first_name = permitted_params[:first_name].presence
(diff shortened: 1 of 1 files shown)


app/controllers/partners/users_controller.rb
```diff
@@ -40,6 +41,10 @@ def create
   private
 
   def find_account
-    @account = @partner.providers.find(params[:provider_id])
+    @account = @partner.providers.find(permitted_params[:provider_id])
+  end
+
+  def permitted_params
+    params.permit!.to_h
```
hallelujah, line 48: We should list all needed/accepted keys instead.

hallelujah (changes requested):

app/controllers/partners/users_controller.rb
nidhi-soni1104: @hallelujah changes done

hallelujah (dismissed):
LGTM

2 other people from @3scale/system to approve too :)

app/controllers/partners/users_controller.rb
```diff
@@ -3,30 +3,31 @@ class Partners::UsersController < Partners::BaseController
   before_action :find_account
 
   def index
-    @users = @account.users.page(params[:page])
-    @users = @users.where(open_id: params[:open_id]) if params[:open_id]
+    @users = @account.users.page(permitted_params[:page])
```
akostadinov, line 6: I don't see `:page` permitted in `#permitted_params`. This will always return `nil` as far as I can tell.

akostadinov (changes requested):

app/controllers/partners/users_controller.rb
nidhi-soni1104: @akostadinov Changes done.

hallelujah (approved):

app/controllers/partners/users_controller.rb
```diff
@@ -40,6 +41,10 @@ def create
   private
 
   def find_account
-    @account = @partner.providers.find(params[:provider_id])
+    @account = @partner.providers.find(permitted_params[:provider_id])
+  end
+
+  def permitted_params
+    params.permit(%i[api_key email first_name last_name open_id username provider_id id page])
```
hallelujah, line 48: I do not see why `api_key` is listed here

hallelujah (changes requested):

app/controllers/partners/users_controller.rb
nidhi-soni1104: when I checked params api_key is there so I listed here to not miss any key.

hallelujah: What did you check? We do not use `permitted_params[:api_key]` or pass `permitted_params` in `User.new` or `user.update` so it should not be dangerous at all.

nidhi-soni1104: I checked running specs like what all params will come
And permitted all the listed ones

PR #2822, opened 2021-12-01
Remove `context` from the last batch of tests

josemigallas:

#### Affected files

* ...
* ...

#### JIRA

### DIFF
--- a/test/unit/finance/charging_test.rb
+++ b/test/unit/finance/charging_test.rb
@@ -15,8 +15,7 @@ class Finance::ChargingTest < ActiveSupport::TestCase
   end
 
   class BuyerNotPayingMonthlyTest < Finance::ChargingTest
-    def setup
-      super
+    setup do
       @buyer_account.not_paying_monthly!
       @invoice = @provider_account.billing_strategy.create_invoice!(buyer_account: @buyer_account)
       @invoice.line_items << LineItem.new(cost: 100)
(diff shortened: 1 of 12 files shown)


test/unit/three_scale/search_test.rb
```diff
@@ -1,171 +1,157 @@
...
-      buyer_zb = FactoryBot.create(:buyer_account, :provider_account => provider, :country => zb)
+    buyer_it = FactoryBot.create(:buyer_account, provider_account: provider, country: it)
+    buyer_zb = FactoryBot.create(:buyer_account, provider_account: provider, country: zb)
 
-      assert_equal [buyer_it, buyer_zb], provider.buyer_accounts.order_by('countries.name', :asc)
-    end
+    assert_equal [buyer_it, buyer_zb], provider.buyer_accounts.order_by('countries.name', :asc)
   end
 
+  class FooClass < ApplicationController
```
akostadinov, line 146: why change parent?

akostadinov (changes requested):
I'm gonna try a few minor changes.

test/unit/finance/charging_test.rb
```diff
@@ -1,89 +1,92 @@
...
-      @plan = FactoryBot.create(:application_plan, :issuer => @provider_account.default_service)
+    @buyer_account = FactoryBot.create(:buyer_account, provider_account: @provider_account)
+    @cinstance = @buyer_account.buy!(@plan)
+  end
 
-      @buyer_account = FactoryBot.create(:buyer_account, :provider_account => @provider_account)
-      @cinstance = @buyer_account.buy!(@plan)
+  class BuyerNotPayingMonthlyTest < Finance::ChargingTest
+    def setup
+      super
```
akostadinov, line 19: I just checked. When inheriting from a class, the setup callbacks registered with the superclass are executed before the setup method. So no need for super. Because of this is is also alright just to use `setup do` in the subclass.

akostadinov (approved):

PR #2824, opened 2021-12-01
[params] Fixes providers session controller

nidhi-soni1104:
Fixes usage of params in the provider session controller

### DIFF
--- a/app/controllers/provider/sessions_controller.rb
+++ b/app/controllers/provider/sessions_controller.rb
@@ -87,12 +87,13 @@ class Provider::SessionsController < FrontendController
   end
 
   def sso_params
-    params.permit(:token, :expires_at, :redirect_url, :system_name, :code).merge(request: request)
+    params.permit(%i[token expires_at redirect_url system_name code]).merge(request: request)
   end
 
   def session_return_to
-    if params[:return_to]
-      return_to = safe_return_to(params[:return_to])
+    return_to_params = params.permit(:return_to)[:return_to]
+    if return_to_params
+      return_to = safe_return_to(return_to_params)
       session[:return_to] = return_to if return_to.present?
     end
   end


app/controllers/provider/sessions_controller.rb
```diff
@@ -87,12 +87,12 @@ def auth_params
   end
 
   def sso_params
-    params.permit(:token, :expires_at, :redirect_url, :system_name, :code).merge(request: request)
+    params.permit(%i[token expires_at redirect_url system_name code]).to_h.merge(request: request)
```
hallelujah, line 90: Is the `to_h` because of a warning with security on user provided params?
Otherwise it is not needed

hallelujah (changes requested):

app/controllers/provider/sessions_controller.rb
nidhi-soni1104: @hallelujah Changes done

app/controllers/provider/sessions_controller.rb
```diff
@@ -87,12 +87,13 @@ def auth_params
...
-    params.permit(:token, :expires_at, :redirect_url, :system_name, :code).merge(request: request)
+    params.permit(%i[token expires_at redirect_url system_name code]).merge(request: request)
   end
 
   def session_return_to
-    if params[:return_to]
-      return_to = safe_return_to(params[:return_to])
+    return_to_params = params.permit(:return_to)[:return_to]
+    if return_to_params
+      return_to = safe_return_to(return_to_params)
```
akostadinov, line 96: Was there any failure that this fixed?

hallelujah: I think it is rubocop complaining only.

hallelujah (approved):

PR #2825, opened 2021-12-02
[params] Fixes provider invitee signups controller

nidhi-soni1104:
Fixes usage of params in the provider invitee signups controller

### DIFF
--- a/app/controllers/provider/invitee_signups_controller.rb
+++ b/app/controllers/provider/invitee_signups_controller.rb
@@ -50,7 +50,7 @@ class Provider::InviteeSignupsController < FrontendController
 
   def build_user
     permitted_params = params.permit(user: %i[password username])
-    @user = @invitation.make_user(permitted_params[:user] || {})
+    @user = @invitation.make_user(permitted_params[:user])
 
     # This is just a sanity guard added when splitting invitation
     # controllers. Remove when SURE.
--- a/app/controllers/provider/invitee_signups_controller.rb
+++ b/app/controllers/provider/invitee_signups_controller.rb
@@ -49,7 +49,7 @@ class Provider::InviteeSignupsController < FrontendController
   end
 
   def build_user
-    @user = @invitation.make_user(params[:user] || {})
+    @user = @invitation.make_user(params.permit![:user].to_h || {})
 
     # This is just a sanity guard added when splitting invitation
     # controllers. Remove when SURE.
--- a/app/controllers/provider/invitee_signups_controller.rb
+++ b/app/controllers/provider/invitee_signups_controller.rb
@@ -49,7 +49,8 @@ class Provider::InviteeSignupsController < FrontendController
   end
 
   def build_user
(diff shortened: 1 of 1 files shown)


app/controllers/provider/invitee_signups_controller.rb
```diff
@@ -49,7 +49,7 @@ def can_create?
   end
 
   def build_user
-    @user = @invitation.make_user(params[:user] || {})
+    @user = @invitation.make_user(params.permit![:user].to_h || {})
```
hallelujah, line 52: I would prefer having a list of permitted attributes that can be passed there than permit everything

nidhi-soni1104: @hallelujah Done

app/controllers/provider/invitee_signups_controller.rb
```diff
@@ -49,7 +49,8 @@ def can_create?
   end
 
   def build_user
-    @user = @invitation.make_user(params[:user] || {})
+    permitted_params = params.permit(user: %i[password username])
```
hallelujah, line 52: Are we sure the params are only those ones?

app/controllers/provider/invitee_signups_controller.rb
```diff
@@ -49,7 +49,8 @@ def can_create?
...
+    @user = @invitation.make_user(permitted_params[:user] || {})
```
hallelujah, line 53:
```suggestion
    @user = @invitation.make_user(permitted_params[:user])
```
No need of `|| {}` since `#permit` always return an `ActionController::Parameters`

hallelujah (changes requested):

app/controllers/provider/invitee_signups_controller.rb
nidhi-soni1104: I checked running specs and found these two params only

hallelujah: The tests are probably lacking, we have more than those 2 fields, those are dynamic fields. You should look at the `show.html.erb` template

https://github.com/3scale/porta/blob/64b2cf8af119b89a480c7a7dd6d9fac1f06afc9d/app/views/provider/invitee_signups/show.html.erb#L22

And those are defined here https://github.com/3scale/porta/blob/64b2cf8af119b89a480c7a7dd6d9fac1f06afc9d/app/lib/fields/form_builder.rb#L13

which uses https://github.com/3scale/porta/blob/master/app/lib/fields/fields.rb#L241-L244

And you collect the `name` of those fields to permit here

nidhi-soni1104: @hallelujah I checked the files and forms you mentioned, but not able to collect the filed names as they are not mentioned clearly.

can we permit all attributes instead? Or any easy and sure way I can collect name of fields names

hallelujah: I described how to collect those names above basically `object.defined_fields.collect(&:name)`

PR #2827, opened 2021-12-02
[params] Fixes provider passwords controller

nidhi-soni1104:
Fixes usage of params in the provider passwords controller

### DIFF
--- a/app/controllers/provider/passwords_controller.rb
+++ b/app/controllers/provider/passwords_controller.rb
@@ -28,7 +28,7 @@ class Provider::PasswordsController < FrontendController
   end
 
   def show
-    if params.permit!.to_h[:password_reset_token].present? && session[:password_reset_token].blank?
+    if password_params[:password_reset_token].present? && session[:password_reset_token].blank?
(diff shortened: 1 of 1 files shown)


app/controllers/provider/passwords_controller.rb
```diff
@@ -28,15 +28,15 @@ def destroy
   end
 
   def show
-    if params[:password_reset_token].present? && session[:password_reset_token].blank?
+    if params.permit!.to_h[:password_reset_token].present? && session[:password_reset_token].blank?
```
hallelujah, line 31: You probably want to have

```ruby
def password_params
  @password_params ||= params.permit(:password_reset_token, {user: { :password, :password_confirmation})
end
```

Then reuse it in `#update` and `#show` actions

hallelujah (changes requested):

app/controllers/provider/passwords_controller.rb
nidhi-soni1104: @hallelujah Changes done

PR #2830, opened 2021-12-06
transient CAS login test failure

akostadinov:
rarely
`FDeveloperPortal::LoginControllerTest#test_cas_is_not_displayed_on_login_page`
may fail, I suspect because of the random content of `authenticity_token`.

Example failure:
https://app.circleci.com/pipelines/github/3scale/porta/19117/workflows/7f60c74d-94d0-48cc-98df-0de12187f9ea/jobs/228400/parallel-runs/1

### DIFF
--- a/test/functional/developer_portal/login_controller_test.rb
+++ b/test/functional/developer_portal/login_controller_test.rb
@@ -16,7 +16,8 @@ class DeveloperPortal::LoginControllerTest < DeveloperPortal::ActionController::
     get :new
 
     assert_response 200
-    assert !@response.body.include?("CAS")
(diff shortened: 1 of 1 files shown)


hallelujah:
We want to move from functional tests (Rails has removed them long ago), and integration tests do not have authenticity_token :)
Probably is the best bet instead of fixing this with regexp

akostadinov:
Moving to integration tests sounds reasonable. But I'd rather not spend time on it now. Can we merge or should I leave as is for the time being?

hallelujah (approved):

PR #2831, opened 2021-12-06
[params] Fixes dashboard service base controller

nidhi-soni1104:
Fixes usage of params in the dashboard service base controller


akostadinov (approved):

akostadinov:
btw does this resolve any issue or it just fails early when parameter is not supplied? What is the purpose of the change?

nidhi-soni1104:
@akostadinov Not Sure about faliure but  there are files sorted by Jakub which causes faliures so we are fixing them all .

PR #2837, opened 2021-12-09
cover bill variable tests with timecop

akostadinov:
In the past `bill for variable` test failed on 1st of the month. Now using `Timecop` to cover some other edge cases as well.

### DIFF
--- a/test/unit/finance/no_variable_cost_test.rb
+++ b/test/unit/finance/no_variable_cost_test.rb
@@ -1,16 +1,51 @@
 require 'test_helper'
 
 class Finance::NoVariableCostClass < ActiveSupport::TestCase
+  attr_reader :contract, :account, :app_plan
 
-  def test_bill_for_variable
-    contract = FactoryBot.build_stubbed(:contract)
-    account  = FactoryBot.build_stubbed(:simple_account)
-    app_plan = FactoryBot.build_stubbed(:application_plan)
+  setup do
+    @contract = FactoryBot.build_stubbed(:contract)
+    @account  = FactoryBot.build_stubbed(:simple_account)
+    @app_plan = FactoryBot.build_stubbed(:application_plan)
+
+    @org_tz = ENV["TZ"]
+  end
+
+  teardown do
+    ENV["TZ"] = @org_tz
+  end
+
+  test "bill for variable" do
+    Timecop.travel(15.days.ago) if Time.now.mday == 1
 
     contract.stubs(:provider_account).returns(account)
     account.stubs(:provider_can_use?).returns(true)
 
     contract.expects(:save).returns(true)
+    contract.notify_observers(:bill_variable_for_plan_changed, app_plan)
+  end
+
+  test "no variable billing on 1st day of month local time" do
+    # this changes server and client local time to China Standard Time (+8)
+    ENV["TZ"] = "Asia/Shanghai"
+
+    # this is 4 hours before beginning of month in UTC and 4 after local
+    Timecop.travel(Time.zone.now.beginning_of_month - 4.hours)
+
+    contract.stubs(:provider_account).returns(account)
+    account.stubs(:provider_can_use?).returns(true)
+
+    contract.expects(:save).never
+    contract.notify_observers(:bill_variable_for_plan_changed, app_plan)
+  end
+
+  test "no variable billing if last billed until is today" do
+    contract.stubs(:provider_account).returns(account)
+    Timecop.freeze
+    contract.stubs(:variable_cost_paid_until).returns(Time.now.to_date)
+    account.stubs(:provider_can_use?).returns(true)
+
+    contract.expects(:save).never
     contract.notify_observers(:bill_variable_for_plan_changed, app_plan)
   end
 end


test/unit/finance/no_variable_cost_test.rb
```diff
@@ -1,16 +1,51 @@
...
     contract.stubs(:provider_account).returns(account)
     account.stubs(:provider_can_use?).returns(true)
 
     contract.expects(:save).returns(true)
     contract.notify_observers(:bill_variable_for_plan_changed, app_plan)
   end
+
+  test "no variable billing on 1st day of month local time" do
+    # this changes server and client local time to China Standard Time (+8)
+    ENV["TZ"] = "Asia/Shanghai"
```
akostadinov, line 30: Tthe test makes sure that billing depends on timezone. If somebody changes `#instant_bill_plan_change` to use `Time.zone.now` instead of the current `Time.now` the test will fail.

Presently no test will fail by such a change.

That is - make sure we retain current behavior until we really want it changed.

I chose this timezone as it is stable 8+ hours AFAIK no DTS but ±1 hour will not affect test outcome. If we rely on local machine timezone, I don't see how to properly test this.

hallelujah: OK so you want it to be timezone aware ...
You can use `Time.zone = 'Asia/Shanghai'` see https://api.rubyonrails.org/v5.0/classes/ActiveSupport/TimeZone.html

test/unit/finance/no_variable_cost_test.rb
```diff
@@ -1,16 +1,51 @@
...
+
+    # this is 4 hours before beginning of month in UTC and 4 after local
+    Timecop.travel(Time.zone.now.beginning_of_month - 4.hours)
```
hallelujah, line 33: Well hopefully there is no DST otherwise it is relatively ±1 hour, otherwise this comment is false. :)
You are comparing UTC (which is absolute) with timezone which are affected by daylight saving.

test/unit/finance/no_variable_cost_test.rb
```diff
@@ -1,16 +1,51 @@
...
+    contract.notify_observers(:bill_variable_for_plan_changed, app_plan)
+  end
+
+  test "no variable billing if last billed until is today" do
+    contract.stubs(:provider_account).returns(account)
+    Timecop.freeze
+    contract.stubs(:variable_cost_paid_until).returns(Time.now.to_date)
+    account.stubs(:provider_can_use?).returns(true)
+
+    contract.expects(:save).never
```
hallelujah, line 48: IMO expect...never is not an expectation. Here what you probably mean is that some database records are not mutated.

hallelujah (changes requested):
So there are 2 specifications:

- changing the same day does not generate variable costs
- changing the 1st of the month does not generate variable costs

I think the latter is a bug but need to be confirmed exactly what is the setup:

1. variable_cost_paid_until set to last day of previous month, let 1 day pass so the test is done at the beginning of the current month, then change plan
2.  variable_cost_paid_until set to the 15th of current month, let 1 day pass , so current date is the 16th, then change plan

IMO both cases should yield the same results.... as 1. is a particular case of 2

test/unit/finance/no_variable_cost_test.rb
hallelujah: Oh I get it now why you use the environment variable ENV['TZ']
It is because of local time.

test/unit/finance/no_variable_cost_test.rb
```diff
@@ -1,16 +1,51 @@
...
+
+    @org_tz = ENV["TZ"]
+  end
+
+  teardown do
+    ENV["TZ"] = @org_tz
+  end
+
+  test "bill for variable" do
+    Timecop.travel(15.days.ago) if Time.now.mday == 1
```
hallelujah, line 19: I do not get that :)
Why not set a fixed date instead?

hallelujah (changes requested):

test/unit/finance/no_variable_cost_test.rb
akostadinov: I chose such a time difference so that DTS will not affect the outcome. Also IIRC China timezone does not use DTS at all.

akostadinov: It is intentional - to make sure that billing uses local time and not UTC. If we used UTC in billing, then test would fail.

test/unit/finance/no_variable_cost_test.rb
akostadinov: https://github.com/3scale/porta/blob/master/app/lib/finance/variable_cost.rb#L35-L40

```ruby
        if invoice.used?
          self.save(:validate => false)
          true
        else
          false
        end
```

Original test used the assertion whether contract was saved or not. If you have better suggestions for assertions, I'll be happy to change.

test/unit/finance/no_variable_cost_test.rb
akostadinov: I can use fixed time. But we caught the issue because test failed at some certain time. So I wanted to let it run at unexpected times and timezones so that we have a chance to catch unexpected behavior. Only at the times we know it should fail, alter time to something else. Let me know.

test/unit/finance/no_variable_cost_test.rb
akostadinov: https://www.timeanddate.com/time/change/china/beijing

DST is no longer in use.

akostadinov: Comment should be always correct :)

hallelujah (approved):

test/unit/finance/no_variable_cost_test.rb
akostadinov: Yes, I'll look at improving this.

PR #2841, opened 2021-12-13
Refactor BillVariableForPlanChangedTest and BillingOverserver

josemigallas:


hallelujah (dismissed):

test/unit/finance/bill_variable_for_plan_changed_test.rb
```diff
@@ -3,49 +3,56 @@
...
     # this is 4 hours before beginning of month in UTC and 4 after local
     Timecop.travel(Time.zone.now.beginning_of_month - 4.hours)
 
-    contract.stubs(:provider_account).returns(account)
-    account.stubs(:provider_can_use?).returns(true)
-
     contract.expects(:save).never
     contract.notify_observers(:bill_variable_for_plan_changed, app_plan)
+
+    ENV["TZ"] = org_tz
```
akostadinov, line 50: this should happen in `tierdown` not as part of test code. Otherwise on a failed test we will end up with a changed timezone.

test/unit/finance/bill_variable_for_plan_changed_test.rb
```diff
@@ -3,49 +3,56 @@
...
+    Timecop.travel(Time.now.beginning_of_month) # rubocop:disable Rails/TimeZone we don't use timezones in billing
+
+    contract.expects(:save).never
+    contract.notify_observers(:bill_variable_for_plan_changed, app_plan)
   end
 
-  test "bill for variable" do
-    Timecop.travel(15.days.ago) if Time.now.mday == 1 # rubocop:disable Rails/TimeZone we don't use timezones in billing
+  test "bill for variable before first day" do
+    Timecop.travel(Time.now.beginning_of_month - 15.days) # rubocop:disable Rails/TimeZone we don't use timezones in billing
```
akostadinov, line 25: `bill for variable after first day` and `bill for variable before first day` is one and the same thing.

Just current month is different. So we don't need two tests to check what happens in the middle of the month.

Also like to run the test in different days of month to have chance to catch any other special cases. I'll leave Hery to advise whether to use this or leave `Timecop.travel(15.days.ago) if Time.now.mday == 1`. Maybe we can iterate over every day of the month so we have full coverage.

josemigallas: Waiting every day of the month in order to possibly find a bug doesn't sound right, considering we can force the time using `Timecop`. I think there are 2 strong scenarios here:  when it is the first day and when it isn't.

The 3rd edge case is interaction between UTC and local time which is already covered.

And we probably should work a bit more with `variable_cost_paid_until`.

> Maybe we can iterate over every day of the month so we have full coverage

Although a bit overkill, it would make sense.

> bill for variable after first day and bill for variable before first day is one and the same thing.

You're right I was thinking about `variable_cost_paid_until` when I wrote this. It is definitely redundant.

test/unit/finance/bill_variable_for_plan_changed_test.rb
```diff
@@ -3,49 +3,56 @@
...
     @app_plan = FactoryBot.build_stubbed(:application_plan)
 
-    @org_tz = ENV["TZ"]
+    contract.stubs(:provider_account).returns(account)
+    account.stubs(:provider_can_use?).with(:instant_bill_plan_change).returns(true)
   end
 
-  teardown do
-    ENV["TZ"] = @org_tz
+  test "bill for variable on first day" do
```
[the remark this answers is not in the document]
josemigallas: Not really. If we consider time zones they are different cases: here it is the 1st day of the month, in the other case it is the day before.

test/unit/finance/bill_variable_for_plan_changed_test.rb
```diff
@@ -3,25 +3,33 @@
...
 
   teardown do
     ENV["TZ"] = @org_tz
   end
 
-  test "bill for variable" do
-    Timecop.travel(15.days.ago) if Time.now.mday == 1 # rubocop:disable Rails/TimeZone we don't use timezones in billing
+  test "bill for variable on 1st day of month UTC" do
+    ENV["TZ"] = "UTC"
+    Timecop.travel(Time.now.beginning_of_month) # rubocop:disable Rails/TimeZone we don't use timezones in billing
```
akostadinov, line 25: We don't care about UTC, we care about local time. In the other test, timezone is changed to make sure calculations are done in local time instead of UTC. If local time is UTC, then we test nothing of that.

Current month in UTC should not matter at all. Perhaps if you use a timezone which is -6 for example, and set current month to current month UTC, then make sure that contract is billed as if it is previous month, it would be provide usefulness.

test/unit/finance/bill_variable_for_plan_changed_test.rb
akostadinov: There is another special case which is the day at a previous plan change (this is also covered). We can never be sure whether some time a new special case can be introduced. We have a lot of uncontrolled variance in tests. Like time zones, hour of day, etc. So it is not really much different.

I think a loop from 2nd to last day of month would be a good options that would be best of both worlds.

hallelujah: To settle it we can just put a fix date we know is an issue ;)

hallelujah (approved):

hallelujah (approved):

test/unit/finance/bill_variable_for_plan_changed_test.rb
```diff
@@ -3,25 +3,32 @@
...
-  test "bill for variable" do
-    Timecop.travel(15.days.ago) if Time.now.mday == 1 # rubocop:disable Rails/TimeZone we don't use timezones in billing
+  test "bill for variable on 1st day of month UTC" do
+    Timecop.travel(Time.now.beginning_of_month) # rubocop:disable Rails/TimeZone we don't use timezones in billing
 
-    contract.stubs(:provider_account).returns(account)
-    account.stubs(:provider_can_use?).returns(true)
+    contract.expects(:save).never
+    contract.notify_observers(:bill_variable_for_plan_changed, app_plan)
+  end
```
akostadinov, line 28: This one seems to replicate `no variable billing on 1st day of month local time`, with the difference that local timezone on test machine is used instead of a hardcoded one. If you intend to use UTC, then it needs to be set. Default on our test machines is +1 I think.

It should pass either way, just title is suggesting UTC while test runs with whatever timezone happens to be on current machine.

josemigallas: Yes I got confused when rebasing. I wanted to remove the whole UTC thing since we don't care about that and I guess I should have removed the whole test case.

PR #2843, opened 2021-12-13
[params] Fixes Outbox controller

josemigallas:
Fixes failure:
```
    Scenario: Sending a message

Then the "To" field should be fixed to "bob"

Message:

    unable to convert unpermitted parameters to hash (ActionController::UnfilteredParameters)
/opt/app-root/src/project/vendor/bundle/ruby/2.6.0/gems/actionpack-5.1.7/lib/action_controller/metal/strong_parameters.rb:265:in `to_h'
/opt/app-root/src/project/vendor/bundle/ruby/2.6.0/gems/activerecord-5.1.7/lib/active_record/inheritance.rb:218:in `subclass_from_attributes'
...
/opt/app-root/src/project/app/controllers/provider/admin/messages/outbox_controller.rb:45:in `build_message'
...
features/old/messages/provider_side.feature:27:in `the "To" field should be fixed to "bob"'
```
Link to build: https://app.circleci.com/pipelines/github/3scale/porta/19248/workflows/2546831f-88c2-4728-a620-bf9cd2b9341e/jobs/229561/tests#failed-test-0

### DIFF
--- a/app/controllers/provider/admin/messages/outbox_controller.rb
+++ b/app/controllers/provider/admin/messages/outbox_controller.rb
@@ -42,7 +42,7 @@ class Provider::Admin::Messages::OutboxController < FrontendController
 
   private
 
-  PERMITTED_PARAMS = %i[id page to subject body origin].freeze
+  PERMITTED_PARAMS = [:id, :to, :page, { message: %i[subject body origin] }].freeze
 
   def build_message
     @message = current_account.messages.build(message_params)
@@ -71,7 +71,7 @@ class Provider::Admin::Messages::OutboxController < FrontendController
   end
 
   def message_params
-    params.fetch(:message, {}).merge(:origin => "web").permit(PERMITTED_PARAMS)
+    permitted_params.fetch(:message, {}).merge(:origin => "web")
   end
 
   def permitted_params
--- a/app/controllers/provider/admin/messages/outbox_controller.rb
+++ b/app/controllers/provider/admin/messages/outbox_controller.rb
@@ -1,16 +1,18 @@
 # frozen_string_literal: true
 
 class Provider::Admin::Messages::OutboxController < FrontendController
-  before_action :build_message, :only => %i[new create]
   activate_menu :buyers, :messages, :sent_messages
+
+  delegate :messages, to: :current_account
 
   def new
     activate_menu :buyers, :messages, :inbox
+    @message = messages.build({})
     @message.to recipients
   end
 
   def destroy
-    @message = current_account.messages.find(message_id_param)
+    @message = messages.find(message_id_param)
     @message.hide!
 
     flash[:notice] = 'Message was deleted.'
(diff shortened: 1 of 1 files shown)


app/controllers/provider/admin/messages/outbox_controller.rb
```diff
@@ -64,10 +67,14 @@ def recipient_ids
   end
 
   def mass_message?
-    current_account.provider? && params[:to].blank?
+    current_account.provider? && permitted_params[:to].blank?
   end
 
   def message_params
-    params.fetch(:message, {}).merge(:origin => "web")
+    params.fetch(:message, {}).merge(:origin => "web").permit(PERMITTED_PARAMS)
```
hallelujah, line 74: I do not think `page` is an attribute of a Message ;)

hallelujah (dismissed):

app/controllers/provider/admin/messages/outbox_controller.rb
```diff
@@ -64,10 +67,14 @@ def recipient_ids
   end
 
   def mass_message?
-    current_account.provider? && params[:to].blank?
+    current_account.provider? && permitted_params[:to].blank?
   end
 
   def message_params
-    params.fetch(:message, {}).merge(:origin => "web")
+    permitted_params.fetch(:message, {}).merge(:origin => "web")
```
akostadinov, line 74: I believe we don't need to permit `origin` if we are overwriting it here.

josemigallas: You're right. In fact, I think there shouldn't be a general permitted params but instead each param should be retrieved separately

app/controllers/provider/admin/messages/outbox_controller.rb
```diff
@@ -64,10 +68,22 @@ def recipient_ids
...
+  def message_id_param
+    params.require(:id)
+  end
+
+  def pagination_params
+    { page: params.permit(:page)[:page] }
+  end
+
+  def to_param
+    params.permit(:to).fetch(:to)
```
hallelujah, line 87: Using `#fetch` will raise a `KeyError` if it is not there. How do you rescue that? We do not want to display a 5xx to the user

hallelujah (changes requested):

app/controllers/provider/admin/messages/outbox_controller.rb
akostadinov: does original approach where we use `params[:to].blank?` inside `@mass_message?` fail for whatever reason?

josemigallas: @hallelujah `:to` should always be there, tests will fail otherwise. In any case, it should fail gracefully:

Screenshot 2021-12-23 at 09 18 13

@akostadinov I don't think so, do you think it's wrong to permit it first?

josemigallas: Removed `param_to`.

hallelujah (approved):

jsmadis (approved):

PR #2844, opened 2021-12-13
Upgrades bullet gem to 6.1.5

jsmadis:

Upgrades bullet gem to a newer version. The old version is incompatible with rails 5.1.7. It results in errors like: `Wrong number of arguments (given 1, expected 2)`. 
The issue in the bullet Github: https://github.com/flyerhzm/bullet/issues/389

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -161,7 +161,7 @@ gem 'slim-rails', '~> 3.2'
 gem 'draper', '~> 3.0'
 
 group :development do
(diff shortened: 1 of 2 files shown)


akostadinov:
Does this one support  5.0?

akostadinov (approved):

jsmadis:
Yes it does. 

`Bullet gem now supports activerecord >= 4.0 and mongoid >= 4.0.`

PR #2847, opened 2021-12-13
[params] Update `send_emails_controller`s

josemigallas:
Fixes _Send mass email to application owners_ tests:
https://app.circleci.com/pipelines/github/3scale/porta/19248/workflows/2546831f-88c2-4728-a620-bf9cd2b9341e/jobs/229561/tests#failed-test-2

https://app.circleci.com/pipelines/github/3scale/porta/19248/workflows/2546831f-88c2-4728-a620-bf9cd2b9341e/jobs/229561/tests#failed-test-5

https://app.circleci.com/pipelines/github/3scale/porta/19248/workflows/2546831f-88c2-4728-a620-bf9cd2b9341e/jobs/229561/tests#failed-test-13

**Note about new scenarios**
The application currently allows sending emails without a subject. This is problematic since the email can't be opened without it (see image below). However wrong this may be, it should be included in the tests. I create a JIRA to address this functionality https://issues.redhat.com/browse/THREESCALE-8009.
Screenshot 2021-12-23 at 10 19 42

### DIFF
--- a/app/controllers/buyers/accounts/bulk/base_controller.rb
+++ b/app/controllers/buyers/accounts/bulk/base_controller.rb
@@ -1,7 +1,11 @@
 # frozen_string_literal: true
 
 class Buyers::Accounts::Bulk::BaseController < Buyers::BulkBaseController
-  before_action :find_accounts
+  before_action :accounts, only: :create
+
+  helper_method :accounts
+
+  def create; end
 
   protected
 
@@ -9,12 +13,12 @@ class Buyers::Accounts::Bulk::BaseController < Buyers::BulkBaseController
     :partners
   end
 
-  def find_accounts
-    @accounts = collection.decorate
+  def accounts
+    @accounts ||= collection.decorate
   end
 
   def collection
-    current_account.buyers.where(id: selected_ids_param)
+    @collection ||= current_account.buyers.where(id: selected_ids_param)
   end
 
   def errors_template
--- a/app/controllers/buyers/accounts/bulk/base_controller.rb
+++ b/app/controllers/buyers/accounts/bulk/base_controller.rb
@@ -14,7 +14,7 @@ class Buyers::Accounts::Bulk::BaseController < Buyers::BulkBaseController
   end
 
   def collection
-    current_account.buyers.where(id: permitted_params[:selected])
+    current_account.buyers.where(id: selected_ids_param)
   end
 
   def errors_template
--- a/app/controllers/buyers/accounts/bulk/base_controller.rb
+++ b/app/controllers/buyers/accounts/bulk/base_controller.rb
@@ -1,12 +1,12 @@
-class Buyers::Accounts::Bulk::BaseController < FrontendController
+# frozen_string_literal: true
 
-  before_action :authorize_bulk_operations
+class Buyers::Accounts::Bulk::BaseController < Buyers::BulkBaseController
   before_action :find_accounts
 
   protected
 
-  def authorize_bulk_operations
-    authorize! :manage, :partners
+  def scope
+    :partners
   end
 
   def find_accounts
@@ -14,12 +14,10 @@ class Buyers::Accounts::Bulk::BaseController < FrontendController
   end
 
   def collection
-    current_account.buyers.where(id: params[:selected])
+    current_account.buyers.where(id: permitted_params[:selected])
   end
 
-  def handle_errors
-    if @errors.present?
-      render 'buyers/accounts/bulk/shared/errors', :status => :unprocessable_entity, formats: [:html]
-    end
+  def errors_template
+    'buyers/accounts/bulk/shared/errors'
   end
 end
--- a/app/controllers/buyers/accounts/bulk/change_states_controller.rb
+++ b/app/controllers/buyers/accounts/bulk/change_states_controller.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 class Buyers::Accounts::Bulk::ChangeStatesController < Buyers::Accounts::Bulk::BaseController
   ACTIONS = %w{ approve make_pending reject }
 
@@ -6,10 +8,9 @@ class Buyers::Accounts::Bulk::ChangeStatesController < Buyers::Accounts::Bulk::B
   end
 
   def create
-    @action = ( ACTIONS & [params[:change_states][:action]] ).first
+    @action = ( ACTIONS & [change_state_action_param] ).first
     return unless @action.present?
 
-    @errors = []
     @accounts = @accounts.to_a.reject do |account|
       !account.public_send("can_#{@action}?")
     end
@@ -21,4 +22,9 @@ class Buyers::Accounts::Bulk::ChangeStatesController < Buyers::Accounts::Bulk::B
     handle_errors
   end
 
+  private
+
+  def change_state_action_param
+    params.require(:change_states).require(:action)
+  end
 end
--- a/app/controllers/buyers/accounts/bulk/change_states_controller.rb
+++ b/app/controllers/buyers/accounts/bulk/change_states_controller.rb
@@ -3,28 +3,20 @@
 class Buyers::Accounts::Bulk::ChangeStatesController < Buyers::Accounts::Bulk::BaseController
   ACTIONS = %w{ approve make_pending reject }
 
-  def new
-    @actions = ACTIONS.map {|a| [a.humanize, a] }
-  end
+  before_action :humanized_actions, only: :new
+
+  def new; end
 
   def create
-    @action = ( ACTIONS & [change_state_action_param] ).first
-    return unless @action.present?
-
-    @accounts = @accounts.to_a.reject do |account|
-      !account.public_send("can_#{@action}?")
-    end
-
-    @accounts.each do |account|
-      @errors << account unless account.public_send(@action)
-    end
-
+    change_states
     handle_errors
   end
 
   private
 
-  def change_state_action_param
-    params.require(:change_states).require(:action)
+  alias subjects accounts
+
+  def actions
+    ACTIONS
   end
 end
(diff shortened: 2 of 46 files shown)


app/controllers/buyers/accounts/bulk/send_emails_controller.rb
```diff
@@ -1,18 +1,24 @@
...
       @errors << message unless message.save && message.deliver!
     end
 
     handle_errors
   end
 
+  private
+
+  def send_emails_params
+    params.fetch(:send_emails).permit!
```
akostadinov, line 22: On the one hand it worked seemingly without protection in the past. On the other hand, perhaps we need to filter out some parameters from message. Maybe check docs.

hallelujah: I tend to agree with you to filter but it is also OK to do the filtering in another PR later ;) In the latter case, open an issue to not forget about it and schedule it

This will really pose an issue when we remove `protected_attributes` gem

app/controllers/buyers/bulk_base_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+  before_action :authorize_bulk_operations
+
+  protected
+
+  def authorize_bulk_operations
+    authorize! :manage, scope
+  end
+
+  def permitted_params
+    params.premit(:selected)
```
hallelujah, line 13:
```suggestion
    params.permit(:selected)
```

If it does not fail then it looks like there is no test case for this :)

app/controllers/buyers/bulk_base_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+  end
+
+  def handle_errors
+    render errors_template, status: :unprocessable_entity, formats: [:html] if @errors.present?
+  end
+
+  def errors_template
+    raise NoMethodError
```
hallelujah, line 21: I like to add a message for this

```suggestion
    raise NoMethodError, "Please define `#errors_template` method in #{self.class}`
```

hallelujah (changes requested):

app/controllers/buyers/accounts/bulk/send_emails_controller.rb
akostadinov: Probably we need to revisit all `#permit!` usages.

akostadinov: I added a line to https://issues.redhat.com/browse/THREESCALE-5010

app/controllers/buyers/bulk_base_controller.rb
```diff
@@ -2,22 +2,46 @@
...
+  def new; end
+
+  def create
+    @errors = nil
+
+    recipients.each do |recipient|
+      message = current_account.messages.build send_emails_params
+      message.to = recipient
+
+      @errors << message unless message.save && message.deliver!
```
akostadinov, line 16: `@errors` supposed to be an array but I see it initialized as `nil`, how would `<<` work with this? Can we have some test to make sure errors do work?

app/controllers/buyers/bulk_base_controller.rb
```diff
@@ -2,22 +2,46 @@
...
+    end
+
+    handle_errors
+  end
 
   protected
 
+  def initializa_errors
+    @errors = nil
```
akostadinov, line 25: typo in method name, and again, shouel errors be `nil` or an empty array `[]`?

app/controllers/buyers/bulk_base_controller.rb
```diff
@@ -2,22 +2,46 @@
 
 class Buyers::BulkBaseController < FrontendController
   before_action :authorize_bulk_operations
+  before_action :initializa_errors, only: :create
+
+  def new; end
+
+  def create
+    @errors = nil
```
akostadinov, line 10: if we init `@errors` in `#initializa_errors`, then probably no need to set it here? Or remove the `before_action`?

app/controllers/buyers/bulk_base_controller.rb
```diff
@@ -4,37 +4,31 @@ class Buyers::BulkBaseController < FrontendController
...
     authorize! :manage, scope
   end
 
-  def permitted_params
-    params.permit(selected: [], send_emails: %i[subject body])
-  end
-
   def send_emails_params
-    permitted_params.fetch(:send_emails)
+    params.permit(send_emails: %i[subject body]).fetch(:send_emails)
```
akostadinov, line 31: Looks great! Just one question - how did you know only `subject` and  `body` are supported?

akostadinov:
```suggestion
    params.require(:send_emails).permit(:subject, :body)
```

app/controllers/buyers/bulk_base_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+      message = current_account.messages.build send_emails_params
+      message.to = recipient
+
+      @errors << message unless message.save && message.deliver!
+    end
+  end
+
+  protected
+
+  def initializa_errors
```
hallelujah, line 22:
```suggestion
  def initialize_errors
```

app/controllers/buyers/bulk_base_controller.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+class Buyers::BulkBaseController < FrontendController
+  before_action :authorize_bulk_operations
+  before_action :initializa_errors, only: :create
```
hallelujah, line 5:
```suggestion
  before_action :initialize_errors, only: :create
```

app/controllers/buyers/bulk_base_controller.rb
hallelujah: Looking at the form in the template probably

hallelujah (changes requested):

app/controllers/buyers/bulk_base_controller.rb
josemigallas: Spanglish alert 🚨

app/controllers/buyers/bulk_base_controller.rb
akostadinov: That's good, just asking. Also if we can have a simple test for these `@errors`, would be good as it is easy to overlook something and break them otherwise.

app/controllers/buyers/accounts/bulk/send_emails_controller.rb
josemigallas: I updated all params usages and hopefully now it's done correctly.

app/controllers/buyers/accounts/bulk/change_states_controller.rb
```diff
@@ -21,4 +22,9 @@ def create
     handle_errors
   end
 
+  private
+
+  def change_state_action_param
+    params.require(:change_states).require(:action)
```
hallelujah, line 28: Do we want to fail hard if the `action` is not sent?
This will raise an `ActionController::ParameterMissing`, and  will display a generic page instead of the bulk action page with errors.

app/controllers/buyers/applications/bulk/change_states_controller.rb
```diff
@@ -22,4 +23,10 @@ def create
     handle_errors
   end
 
+  private
+
+  def change_state_action_param
+    params.require(:change_states).require(:action)
```
hallelujah, line 29: We need to make sure if we do not select any action that `params.require(:change_states) == {}`

hallelujah (changes requested):
Maybe `params.require` here is not appropriate, unless you are sure that it won't raise anything

app/controllers/buyers/accounts/bulk/change_states_controller.rb
josemigallas: I will keep the same behavior with a blank option, maybe we upgrade the whole bulk actions with Patternfly in the future.

app/controllers/buyers/accounts/bulk/base_controller.rb
```diff
@@ -1,25 +1,23 @@
...
+class Buyers::Accounts::Bulk::BaseController < Buyers::BulkBaseController
   before_action :find_accounts
 
   protected
 
-  def authorize_bulk_operations
-    authorize! :manage, :partners
+  def scope
+    :partners
   end
```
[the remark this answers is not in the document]
josemigallas: What is this for anyway?

app/controllers/buyers/bulk_base_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+  def initialize_errors
+    @errors = []
+  end
+
+  def authorize_bulk_operations
+    authorize! :manage, scope
+  end
+
+  def selected_ids_param
+    params.require(:selected)
```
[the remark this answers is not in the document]
josemigallas: It doesn't fail hard:
Screenshot 2021-12-22 at 09 26 08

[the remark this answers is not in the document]
josemigallas: And it shouldn't be possible to submit a change without any item selected (the bulk actions menu will hide) this is only possible by manipulating the DOM with the browser dev toold.

app/controllers/buyers/bulk_base_controller.rb
josemigallas: The simple test for @errors turned up in an important cleanup/refactoring. But I think the outcome is much better than what we currently have.

app/controllers/buyers/accounts/bulk/base_controller.rb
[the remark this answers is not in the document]
hallelujah: I do not know :)

hallelujah (approved):

jsmadis (approved):

PR #2852, opened 2021-12-15
⚰️ Cleanup partial apicast

josemigallas:
https://issues.redhat.com/browse/THREESCALE-7992

### DIFF
--- a/app/views/api/integrations/apicast/configuration_driven/_apicast.html.slim
+++ b/app/views/api/integrations/apicast/configuration_driven/_apicast.html.slim
@@ -7,9 +7,6 @@ section.Configuration class=('Configuration--is-promoted' unless deployment_opti
         h4
           => icon('cubes')
           = @service.name
-      dd.u-dl-definition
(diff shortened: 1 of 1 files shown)


akostadinov:
Who should review this?

josemigallas:
> Who should review this?

Ah! An unsuspecting volunteer!

akostadinov:
I just don't understand how is this to be tested. What is the effect of the change?

josemigallas:
@akostadinov This is supposed to be dead code that should have been removed as part of [THREESCALE-3759](https://issues.redhat.com/browse/THREESCALE-3759). Not sure how to test it though, I guess all related tests were removed.

Maybe @hallelujah @guicassolato know more.

akostadinov (dismissed):
Ok. If it doesn't result in any side effects in the UI, it should be fine.

josemigallas:
> Ok. If it doesn't result in any side effects in the UI, it should be fine.

Well, it turns out it does breaks the UI 😅

PR #2855, opened 2021-12-15
[WIP] Intercept emails to use custom username/password

hallelujah:
Part of THREESCALE-7232


app/models/email_configuration.rb
```diff
@@ -0,0 +1,30 @@
...
+      # domain: nil,
+      # authentication: nil
+    }.reverse_merge(ActionMailer::Base.smtp_settings)
+  end
+
+  protected
+
+  def set_tenant_id
+    self.tenant_id ||= account&.id if account&.provider?
+  end
```
hallelujah, line 29: This is to get rid of DB triggers

app/models/email_configuration.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+class EmailConfiguration < ApplicationRecord
+
+  validates :email, :password, :user_name, :authentication, :domain, :tls, :address, length: { maximum: 255 }
+
+  validates :email, :password, :user_name, presence: true
+  validates :email, format: { with: Authentication::RE_EMAIL_OK, message: Authentication::MSG_EMAIL_BAD }
```
hallelujah, line 8: I have a gem to do email validation, which was quite famous :) https://github.com/hallelujah/valid_email (contributors needed :) )
I am not sure about using this validation and how it plays with on-premises, I can imagine that people want to send them to internal aliases which do not have domain parts. Well it is not defined, it is for SaaS only so we can be at ease with validating like this first.

app/models/email_configuration.rb
```diff
@@ -0,0 +1,45 @@
...
+  validates :email, uniqueness: { case_sensitive: false, message: "This email address is already used." }
```
hallelujah, line 9: I do not know much about email address RFC but are they really case insensitive?

But a quick search tells me:

> Are email addresses case sensitive RFC?
> Yes, according to RFC 2821, the local mailbox (the portion before `@`) is considered case-sensitive. However, e-mail addresses are not usually case-sensitive because of the difficulties and confusion it would cause between users, the server, and the administrator.Jun 7, 2019

app/models/email_configuration.rb
```diff
@@ -0,0 +1,45 @@
...
+  validates :address, :domain, format: { with: Authentication::RE_DOMAIN_OK }, if: -> { address }
+  validates :authentication, inclusion: { in: %w[plain login cram_md5] }, if: -> { authentication }
+  validates :tls, inclusion: { in: %w[starttls starttls_auto tls no] }, if: -> { tls }
```
hallelujah, line 12: I would not do that in this first PR but in a second one. Not even displaying the field

db/migrate/20211215120415_create_email_configurations.rb
```diff
@@ -0,0 +1,31 @@
...
+    # Oracle and Postgres don't have case insensitive collations, we will rely on app logic for it.
+    collation_ci = case
+                 # when System::Database.postgres?
+                 #   supported only with version 13+
+                 #   { collation: "und-x-icu" }
+                 when System::Database.mysql?
+                   { collation: "utf8_general_ci" }
+                 else
+                   {}
+                 end
```
hallelujah, line 15: I think this will use the config/database.yml options by default. Also no need to add it for every column.
I say this is not needed

db/schema.rb
```diff
@@ -470,6 +470,19 @@
     t.index ["owner_type", "owner_id"], name: "index_deleted_objects_on_owner_type_and_owner_id", using: :btree
   end
 
+  create_table "email_configurations", id: :bigint, force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_bin" do |t|
```
hallelujah, line 473: We should not use that but utf8_mb4 ... another story... please do not add it.

akostadinov: The issue with `mb4` is that it also counts what is the MySQL connection encoding. I did a quick experiment with `mb4` encoding and collation for the tags pull request. The issue was that 4 byte characters fail to work because connection is non-mb4.

I tested that if connection is `mb4` encoding, then mb3 still works. But I was worried about unexpected side effects from switching to a different connection encoding and decided to just keep broken utf8 :disappointed:

akostadinov:
I updated again. It is mostly the same + validation refinements and fixes and more tests.

One major thing I changed is that global config is merged only when nothing but credentials are set for the EmailConfiguration. The idea is that server settings should be always consistent, only credentials can differ.

Otherwise a user may set server address and TLS but if global config contains `port`, then that port might be not the default port that user expected. Same with tls settings.

Still controller is untouched.

db/schema.rb
hallelujah: But we do not want to enforce encoding anyway. It should be on the config level for all the tables

akostadinov: For the record, I didn't commit `db/schema.rb` :) But even if I did, it would have been the same.

This seems to be default configuration in on-prem database. When schema is dumped, charset and collation are dumped with it. For individual columns, that is skipped if it matches table config.

One thing that I don't like is that 4mb and non-4mb collations need to be applied accordingly to column encoding. So we can't add a `_ci` collation in an encoding agnostic way :/

db/migrate/20211215120415_create_email_configurations.rb
akostadinov: It is to add `_ci` collation for columns that should be case insensitive. Unfortunately seems to work only on MySQL. Some fields are better off with `_bin` collation some with `_ci`. It mostly matters for unique indices. Otherwise we need to ensure uniqueness on the app level which is not so reliable.

Do you want no `_ci` indices?

app/models/email_configuration.rb
akostadinov: They are not but we don't validate them according to spec anyway. See
https://github.com/3scale/porta/blob/master/app/lib/authentication.rb#L16

I don't think emails with only case difference is a good idea. If we need at some point, I assume we can easily relax thta restriction. Easier than ensuring uniqueness later at least.

app/models/email_configuration.rb
akostadinov: I'm using what we already have in other places we set email addresses. I believe these need to stay consistent.

Your gem is awesome and I'm starring it to use later.

app/models/email_configuration.rb
akostadinov: I didn't think about this one. Wouldn't it set `tenant_id` to master account and in trigger we set to `null` in such case. Perhaps we need consistent behavior here.

db/schema.rb
akostadinov: You can see that we don't add it in migration:
https://github.com/3scale/porta/pull/2855/files#diff-b3fb527ac8c4b8ef169d45e9d95097c1e6c60ff72abdf3d340fd7bd2ac78e546R17

app/controllers/provider/admin/account/email_configurations_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  def index
+  end
+
+  def new
+    @email_configuration = EmailConfiguration.new
+  end
+
+  def edit
+    @email_configuration = account.email_configurations.find(params[:id])
```
akostadinov, line 16: if we are master, we probably need to search all configurations, not only owned by provider.

akostadinov: pagination

app/controllers/provider/admin/account/email_configurations_controller.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+
+class Provider::Admin::Account::EmailConfigurationsController < Provider::Admin::Account::BaseController
+
+  # We only enable it for master for now, feel free to roll it out for any provider
+  before_action :ensure_master_domain, :email_configurations_enabled?
```
josemigallas, line 6: This is already defined in a parent class Provider::Admin::BaseController: https://github.com/3scale/porta/blob/25fa9bc53bdd3eeb8f2a9c8e882f771c85c3c3cb/app/controllers/provider/admin/base_controller.rb#L2

josemigallas, line 6: Could we inherit from https://github.com/3scale/porta/blob/25fa9bc53bdd3eeb8f2a9c8e882f771c85c3c3cb/app/controllers/provider/base_controller.rb instead that includes this already?

app/controllers/provider/admin/account/email_configurations_controller.rb
josemigallas: I think you mean `pagination` for the index. We will probably add a presenter for this.

app/controllers/provider/admin/account/email_configurations_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  def edit
+    @email_configuration = account.email_configurations.find(params[:id])
+  end
+
+  def update
+    @email_configuration = account.email_configurations.find(params[:id])
+
+    if @email_configuration.update(configuration_params)
+      redirect_to action: :index
```
josemigallas, line 23: Some flash messages would be nice I guess: _Email configuration created_, _Email configuration could not be updated_, etc.

app/controllers/provider/admin/account/email_configurations_controller.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+
+class Provider::Admin::Account::EmailConfigurationsController < Provider::Admin::Account::BaseController
```
josemigallas, line 3: Actually, is this correctly scoped? Shouldn't in be `Provider::EmailConfigurationsController` since it's only for the master provider?

akostadinov: The thing is that in the future we may enable this for other providers as well.

wrt scoping - I'm not sure about this. I don't understand the convention very well. `Provider::Admin` sounds good to me because only provider admins should be ever able to manage this. `Account` - I don't know what is this about. So if you think another scope makes more sense - I'm very open to it.

josemigallas: Yeah, the convention about scoping is very loose. `Provider::Admin` makes sense, we could remove the `Account` part.

I think the idea ultimately is to share the same structure between UI and controllers. So if the scope were `Provider::Admin::Account` the URL should be like `p/admin/accounts/:id/email_configuration`

app/controllers/provider/admin/account/email_configurations_controller.rb
josemigallas: Actually, if this will be available for providers in the future, we don't want to inherit from that. It's fine like this, explicit, so that we can remove it when we make it available for all providers (instead of changing the parent class again). WDYT?

app/controllers/provider/admin/account/email_configurations_controller.rb
akostadinov: Agreed - depending on UI follow Rails conventions.

akostadinov (approved):

PR #2857, opened 2021-12-16
Update product metrics

josemigallas:
### Description
This PR does 2 things:
* Fixes n+1 query in MetricsIndexPresenter by using `includes` to load proxy_rules for each metric.
* Update old 🥒 tests:
  * Combines related both metrics and methods tests in the file (they are all Metric)
  * Uses standard names (create vs. creation, delete vs. deletion)
  * Replaces nukes and zombies for Italian dishes
  * Updates scenarios so that they take into account the new 2 tabs
  * Removes a couple of redundant steps related to ParameterType `provider`
  * DRYs step `provider should have metric` by using ParameterType `should`

### Jira
[THREESCALE-7987: Add test for metrics index](https://issues.redhat.com/browse/THREESCALE-6872)


akostadinov:
Steps look good. It is hard to reason about the test changes though. It is not possible to see what changed from original version. Could you try to submit functional changes separate from refactoring?

josemigallas:
> Steps look good. It is hard to reason about the test changes though. It is not possible to see what changed from original version. Could you try to submit functional changes separate from refactoring?

I don't think there are that many functional changes, it's mostly refactoring. I reordered the commits so that it's easier to understand the changes (hopefully).

I also added a changes summary in the description.

app/presenters/api/metrics_index_presenter.rb
```diff
@@ -37,7 +37,8 @@ def create_path
   protected
 
   def raw_collection
-    @raw_collection ||= metrics? ? top_level_metrics : method_metrics
+    collection = metrics? ? top_level_metrics : method_metrics
+    @raw_collection ||= collection.includes(:proxy_rules)
```
akostadinov, line 41: There is a nitpick here, that if `@raw_collection` is set, no need for the previous statement but it doesn't matter

josemigallas: I assumed that's a lazy statement? No data is loaded. Feel free to suggest an improvement.

hallelujah, line 41:
```suggestion
    @raw_collection = (metrics? ? top_level_metrics : method_metrics).includes(:proxy_rules)
```

josemigallas: Ok that was easy 😅

features/api/metrics/index.feature
```diff
@@ -16,37 +16,40 @@ Feature: Product > Metrics index
...
+
+    Scenario: New method button
+      Then I should see "Add a method"
+
+    Scenario: Methods and mapping rules
+      Given method "Diavola" is not mapped
+      And method "Carbonara" is mapped
+      When I change to tab "Methods"
+      Then I should be able to add a mapping rule to "Diavola"
+      But I should not be able to add a mapping rule to "Carbonara"
```
guicassolato, line 35: This is confusing for one who is not looking at the UI. Technically, you _should_ be able to map "Carbonara", even when it's already been mapped. It's only the link that changes from "Add a mapping rule" to a "✓" sign.

features/api/metrics/update.feature
```diff
@@ -0,0 +1,39 @@
...
+    And I press "Update Method"
+    Then I should see "Burrata Panzanella"
+    And method "Burrata Panzanella" should have the following:
+      | Friendly name | Burrata Panzanella |
+      | Unit          | hit                |
+
+  Scenario: Can't change system name of default metric
+    Given I change to tab "Metrics"
+    When I follow "Hits"
+    Then I should not see field "metric[system_name]"
```
guicassolato, line 28: This is new, right? Because right now the field is there; it's just disabled, yet still visible.

features/support/parameter_types.rb
```diff
@@ -184,7 +184,14 @@ def provider_by_name(name)
...
   name: 'metric',
   regexp: /metric "([^"]*)"/,
-  transformer: ->(name) { Metric.find_by!(system_name: name) }
+  transformer: ->(name) {
+    metrics = Metric.where(parent_id: nil)
+    metric = metrics.find_by(friendly_name: name) || metrics.find_by(system_name: name)
+
+    raise ActiveRecord::RecordNotFound, "Couldn't find metric '#{name}'" unless metric
+
+    metric
```
guicassolato, line 193:
```suggestion
    metric or raise(ActiveRecord::RecordNotFound, "Couldn't find metric '#{name}'")
```

features/support/parameter_types.rb
```diff
@@ -184,7 +184,14 @@ def provider_by_name(name)
 ParameterType(
   name: 'metric',
   regexp: /metric "([^"]*)"/,
-  transformer: ->(name) { Metric.find_by!(system_name: name) }
+  transformer: ->(name) {
+    metrics = Metric.where(parent_id: nil)
+    metric = metrics.find_by(friendly_name: name) || metrics.find_by(system_name: name)
```
guicassolato, line 189: Searching across all metrics in the table by system_name is less prone to finding the wrong one, than by friendly_name, yet even so not risk free. This seems to increase the risk of fetching the wrong metric when there are other tests running in parallel on the same db. Do we really want it to be unscoped?

josemigallas: I just think it's far more readable to use the friendly name in the scenarios. I wouldn't call it risky but... 🤷🏻‍♂️

features/api/metrics/create.feature
```diff
@@ -0,0 +1,38 @@
...
+    And I fill in "System name" with "antipasti"
+    And I fill in "Unit" with "order"
+    And I fill in "Description" with "The number of antipasti dishes ordered"
+    And I press "Create Metric"
+    Then I should see "Antipasti"
+    And the provider should have metric "Antipasti"
+    And metric "Antipasti" should have the following:
+      | Friendly name | Antipasti                              |
+      | Unit          | order                                  |
+      | Description   | The number of antipasti dishes ordered |
```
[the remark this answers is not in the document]
josemigallas: Why not!

features/api/metrics/update.feature
hallelujah: Probably use a step where a hidden field is expected, if it is the case

josemigallas: Yeah good catch

josemigallas: Yes, for capybara to "not see" equals "see hidden or disabled". But there is this other step `see field X disabled` which is more specific.

features/support/parameter_types.rb
josemigallas: @akostadinov  @hallelujah WDYT?

hallelujah (approved):

features/support/parameter_types.rb
hallelujah: I think this one is fine. There should not be other tests running in parallel on the same table, there were parallel tests before and they are scoped on DB name level.

On new Rails version it is also the case https://guides.rubyonrails.org/testing.html#parallel-testing-with-processes

> When parallelizing tests, Active Record automatically handles creating a database and loading the schema into the database for each process. The databases will be **suffixed with the number corresponding to the worker.** For example, if you have 2 workers the tests will create test-database-0 and test-database-1 respectively.

features/api/metrics/delete.feature
```diff
@@ -0,0 +1,32 @@
...
+    And the following methods:
+      | Carbonnara |
+    And I go to the metrics and methods page
+
+  Scenario: Delete a method from the index page
+    Given I change to tab "Methods"
+    When I follow "Carbonnara"
+    And I press "Delete" and I confirm dialog box
+    Then I should not see metric "Carbonnara"
+    And the provider should not have metric "Cabonnara"
```
damianpm, line 20: Typo, missing "r"
In fact, is Carbonara, with single ;)

damianpm (changes requested):

damianpm (approved):
LGTM

PR #2858, opened 2021-12-16
New Metrics and Methods page for Backend API using PF4

josemigallas:

Re-implements Metrics index page with React and Patterrnfly styles. **DISCLAIMER**: it's url-params-based so page has to load everytime tab/page changes.

Old:
Screenshot 2021-12-16 at 15 42 33

New:

Screenshot 2021-12-16 at 15 19 06
Screenshot 2021-12-16 at 15 39 02

Search implemented:
Screenshot 2021-12-16 at 15 38 56

[THREESCALE-6872: [3scale][2.11][HI-prio] Improve list of Methods and Metrics in Backend context](https://issues.redhat.com/browse/THREESCALE-6872)

[THREESCALE-7027: [UI Strings] Improve list of Methods and Metrics in Backend context](https://issues.redhat.com/browse/THREESCALE-7027)


app/controllers/api/metrics_controller.rb
```diff
@@ -68,11 +52,7 @@ def destroy
       flash[:error] = 'The Hits metric cannot be deleted'
     end
 
-    respond_to do |format|
-      format.html do
-        redirect_to admin_service_metrics_path(@service, tab: "#{method_or_metric}s")
-      end
-    end
+    redirect_to admin_service_metrics_path(@service, tab: "#{method_or_metric}s")
```
akostadinov, line 55: Why do we get rid of `respond_to`? We do the same regardless of format requested?

akostadinov:
Looks good to me.

app/controllers/api/metrics_controller.rb
josemigallas: Rails responds with HTML by default and we don't require any other format for these actions, so I figured we could reduce the verbosity.

akostadinov: if another format is requested, will the error message change after this update?

akostadinov: well, this is a potential issue. What is the old and new message?

josemigallas: Sorry I meant the opposite, the response is exactly the same.

akostadinov:
One question about 
> DISCLAIMER: it's url-params-based so page has to load everytime tab/page changes.

Did it need reload previously?

josemigallas:
> One question about
> 
> > DISCLAIMER: it's url-params-based so page has to load everytime tab/page changes.
> 
> Did it need reload previously?

No, it maintains the same behaviour. This warning was to maintain UX's (a.k.a Thomas 😂 ) expectations low, because at some point the mention they would like everything to be reactive and AJAX based. But it's not going to happen for now.

akostadinov (approved):

PR #2859, opened 2021-12-16
Quick starts integration

jschuler:
WIP proof-of-concept for adding quick starts to 3scale

This POC adds the quick starts drawer and the resources page to 3scale.
The code is not meant to be production ready at this point, it is merely at a point to showcase that it is possible to integrate the quick starts library into 3scale.

For the quick starts drawer styling and interaction to work properly, the main content needs to be wrapped. In `QuickStartsContext.jsx` there is a `useEffect` which moves DOM elements around to achieve the proper nesting. This is a bit hackish but appears to work alright. Ideally we could wrap the main container in the quick starts react context, but I have not found a way yet to do so (I have also not done ruby/rails work before so this is probably possibly).

~~There are also styling issues, since the quick starts library depends on newer versions of react-core and patternfly:
Quick starts has a peer dependency on `"@patternfly/react-core": ">=4.115.2"`, which uses `"@patternfly/patternfly": "4.102.2"`.~~
~~However, 3scale uses~~
```
"@patternfly/patternfly": "2.40.6",
"@patternfly/react-core": "^3.120.5",
```

~~For this POC, I have added the following two stylesheets and commented out the PF4Styles/base:~~
```
link rel="stylesheet" href="https://unpkg.com/@patternfly/patternfly@4.102.2/patternfly.css"/
link rel="stylesheet" href="https://unpkg.com/@patternfly/patternfly@4.102.2/patternfly-addons.css"/
```
~~but of course it causes other styling issues, since 3scale uses older versions of PatternFly.~~

~~I think either 3scale will have to update their version of PatternFly, or we need to figure out a way
to bundle a subset of PF classes/styles (namespaced to the quick start components) with the library.~~

**I have updated the POC with a newer version of the quickstarts library, which now ships a set of CSS that contains the needed PatternFly**

Demo:

https://user-images.githubusercontent.com/869106/146456403-f124028b-6eb4-4cc0-b003-0bc4a1d9cb34.mov


junezhang:
@jschuler Thanks for creating this PoC. From a design perspective, I don't have any questions except the line height is small. 
Just one question regarding how to maintain the quick start content, when reviewing the files, if the Content team maintains the quick start, they just need to take care of the files in porta/app/javascript/src/QuickStarts/resources/*.jsx right? or any other ways?

deernano:
Great PoC! Just the margin between each line could be larger ?

jschuler:
Thanks for the comments, I will be updating the quickstarts library to provide additional (optional) patternfly stylesheets that consumers like 3scale can pull in which should resolve the CSS issues. Although not ideal, this way 3scale can keep their older version of PF and we load additional stylesheets for quickstarts, without anything clashing hopefully.

jschuler:
I moved the quick starts link out of the context menu, and into the help menu, seemed more appropriate

Screen Shot 2022-01-13 at 4 02 30 PM

PR #2860, opened 2021-12-20
[BugFix] Tag list: order alphabetically

damianpm:
**What this PR does / why we need it**:

Tag list items are being ordered differently by `pgsql` and `mysql`, making some test fail.
To workaround this issue we can **sort** the tags.

Follow-up of https://github.com/3scale/porta/pull/2798

PR #2863, opened 2021-12-21
tag check step to ignore order of tags

akostadinov:
Replaces https://github.com/3scale/porta/pull/2860


hallelujah (approved):
I think that is good too.
Either we specify that tags are always displayed in alphabetical order or this.

PR #2864, opened 2021-12-24
THREESCALE-8013 Clean bulk action "Delete" up

lvillen:

This PR deletes all comented lines referring delete action on files: 
  - applications index (app/views/shared/applications/_listing.html.slim)
  - accounts index (app/views/buyers/accounts/index.html.erb)
  - service contracts index (app/views/buyers/service_contracts/index.html.slim)

Fixes # THREESCALE-8013
https://issues.redhat.com/browse/THREESCALE-8013


josemigallas:
Hi @lvillen thanks a lot for your PR!

Please remove as well all related code (controllers, tests, views, javascript). Basically everything under `bulk/deletes'` and  `bulk_operations/deletes`. The following commands might be useful:
- `$ ls -al ./**/bulk/delete**`
- `$ ls -al ./**/bulk_operation/delete**`

Similarly you could look up for other usages of the delete bulk action in other files by running
* `$ ls -al ./app/**/bulk_operations**`

And finally we should remove any routes, you could run something like this to verify:
- `$ rails routes |grep _bulk_delete` to verify.

I think the only one left is at  I think there is only one left at: https://github.com/3scale/porta/blob/6613ba1f9348e25de1e3a89ff08b975a859453f1/config/routes.rb#L947.

josemigallas (approved):

jsmadis (approved):

josemigallas:
> Are the below lines also relevant or we want to keep them?
> 
> ```shell
> $ rg -i bulk | grep -i delete
> app/views/provider/admin/messages/_bulk_operations.html.slim:    = bulk_action 'Delete', new_provider_admin_messages_bulk_trash_path(scope: scope), 
> app/views/provider/admin/messages/bulk/trash/new.js.erb:      <%= form.commit_button 'Delete', button_html: { data: { 'disable-with' => 'Running bulk action...' }} %>
> features/old/messages/provider_side.feature:    And I press "Delete" within "#bulk-operations"
> features/old/messages/provider_side.feature:    When I press "Delete" within "#bulk-operations"
> ```

Messages is the only thing that still supports bulk deletion

PR #2868, opened 2022-01-04
[params] Settings controller

josemigallas:
**Description**
~Unlike Hash, Params doesn't have `with_indiferent_access` so it has to be parsed when passed to `update_attribute`. However, since params are not permitted, `to_h` will fail.~
We need to upgrade `protected_attributes_continued` https://github.com/3scale/porta/pull/2870

**Steps to reproduce**
Go to Audience > Billing > Charging & Gateway and change charging settings

**Before**
Screenshot 2022-01-04 at 09 55 56

**After**
Screenshot 2022-01-04 at 09 56 15

### DIFF
--- a/app/controllers/finance/provider/settings_controller.rb
+++ b/app/controllers/finance/provider/settings_controller.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 class Finance::Provider::SettingsController < Finance::Provider::BaseController
 
   before_action :set_strategy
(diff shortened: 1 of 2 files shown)


jsmadis:
The root cause of this issue is an old version of `protected_attributes_continued` gem.

We are using version `1.3.0` which is not compatible with rails `5.1.7` especially with the Parameters class.

This line: https://github.com/westonganger/protected_attributes_continued/blob/v1.3.0/lib/active_record/mass_assignment_security/nested_attributes.rb#L53 
Results in the exception and it's fixed in the following versions. In order to fix this, we need to update the `protected_attributes_continued` gem to a higher version.

akostadinov:
I think the changes are good. But probably update model not to use `protected_attributes` because they become redundant as far as I can say.

jsmadis (approved):

PR #2869, opened 2022-01-04
[Params] Fixing sudo controllers when using symbolize keys

jsmadis:

Fixes the sudo controllers usage when using rails 5.1 version. 
We need to convert the params to hash before calling `symbolize_keys`

### DIFF
--- a/app/controllers/provider/admin/sudo_controller.rb
+++ b/app/controllers/provider/admin/sudo_controller.rb
@@ -2,8 +2,7 @@ class Provider::Admin::SudoController < FrontendController
 
   helper_method :sudo
 
-  def show
-  end
+  def show; end
 
   def create
     if sudo.correct_password?(current_password)
@@ -30,7 +29,7 @@ class Provider::Admin::SudoController < FrontendController
   end
 
   def sudo
-    @_sudo ||= ::Sudo.new(sudo_params.symbolize_keys)
+    @sudo ||= ::Sudo.new(sudo_params.to_h.symbolize_keys)
   end
 
 end


app/controllers/provider/admin/sudo_controller.rb
```diff
@@ -30,7 +29,7 @@ def current_password
   end
 
   def sudo
-    @_sudo ||= ::Sudo.new(sudo_params.symbolize_keys)
+    @sudo ||= ::Sudo.new(sudo_params.to_h.symbolize_keys)
```
akostadinov, line 32: Could you explain this change?

jsmadis: It fixes this test in the rails upgrade branch 

```log
failed Regenerate provider key

    Scenario: Regenerate provider key

And I enter the admin password in "Current password"

Message:

    undefined method `symbolize_keys' for #<ActionController::Parameters:0x0000000013cb8be0> (NoMethodError)
/opt/app-root/src/project/app/controllers/provider/admin/sudo_controller.rb:33:in `sudo'
/opt/app-root/src/project/app/controllers/provider/admin/sudo_controller.rb:9:in `create'
/opt/app-root/src/project/vendor/bundle/ruby/2.6.0/gems/actionpack-5.1.7/lib/action_controller/metal/basic_implicit_render.rb:4:in `send_action'
/opt/app-root/src/project/vendor/bundle/ruby/2.6.0/gems/actionpack-5.1.7/lib/abstract_controller/base.rb:186:in `process_action'
/opt/app-root/src/project/vendor/bundle/ruby/2.6.0/gems/actionpack-5.1.7/lib/action_controller/metal/rendering.rb:30:in `process_action'
/opt/app-root/src/project/vendor/bundle/ruby/2.6.0/gems/actionpack-5.1.7/lib/abstract_controller/callbacks.rb:20:in `block in process_action'
/opt/app-root/src/project/vendor/bundle/ruby/2.6.0/gems/activesupport-5.1.7/lib/active_support/callbacks.rb:108:in `block in run_callbacks'
/opt/app-root/src/project/vendor/bundle/ruby/2.6.0/gems/audited-5.0.2/lib/audited/sweeper.rb:16:in `around'
/opt/app-root/src/project/vendor/bundle/ruby/2.6.0/gems/activesupport-5.1.7/lib/active_support/callbacks.rb:117:in `block in run_callbacks'
/opt/app-root/src/project/vendor/bundle/ruby/2.6.0/gems/audited-5.0.2/lib/audited/sweeper.rb:16:in `around'
```

akostadinov: I was asking more about `@_sudo` vs `@sudo`. Sorry for not being clear :)

Is it because of style complaints or some other reason. Can `@sudo` interfere with any other function?

jsmadis: Then it was this warning:

```
RuboCop: Memoized variable `@_sudo` does not match method name `sudo`. Use `@sudo` instead. [Naming/MemoizedInstanceVariableName]
```
AFAIK it should interfere since it is only a class variable.

akostadinov (approved):

app/controllers/provider/admin/sudo_controller.rb
akostadinov: Instance variables are passed to views. https://thepugautomatic.com/2013/05/locals/

If you think that is not used anywhere - ok. I personally prefer to keep it `@_sudo` to indicate this is not to be used in views. I don't care so much about this rubocop warning. But leave it up to you.

jsmadis: I double-check that and `@sudo`  is not used anywhere.

Also, I think that `_` as name prefix doesn't mean that the instance variable is private so it shouldn't matter.

PR #2878, opened 2022-01-10
Remove whatwg-fetch polyfill

josemigallas:

`whatwg-fetch` is a polyfill for `fetch` that we no longer need. We have been using `fetch` for a while so it should be safe to remove.

**Verification steps** 

`fetchData` is used in `ActiveDocs v3 autocomplete`, `policies`, `service discovery` and `inline stats`. Everything should be working as usual.


nidhi-soni1104 (approved):

didierofrivia (approved):
I'd say it's pretty safe, given we are no longer giving support to old browsers... 
https://caniuse.com/fetch
https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API

PR #2885, opened 2022-01-11
Integration controller permit proxy_rules_attributes

jsmadis:

Cherry-pick from rails branch. This change is needed for the rails 5.1.


jsmadis:
This commit can't be a part of the master yet. It requires rails version 5.1

PR #2887, opened 2022-01-13
Message controller parameters permit

jsmadis:
Permitting only parameters described in API documentation.

Cherry-pick from rails branch that can be merged to actual master.

### DIFF
--- a/app/controllers/admin/api/messages_controller.rb
+++ b/app/controllers/admin/api/messages_controller.rb
@@ -18,7 +18,7 @@ class Admin::Api::MessagesController < Admin::Api::BuyersBaseController
   ##~ op.parameters.add :name => "body", :description => "Text to send", :dataType => "string", :required => true, :paramType => "query", :allowMultiple => false
   #
   def create
-    message = current_account.messages.build params[:message]
+    message = current_account.messages.build params[:message].permit(:body)
     message.to = buyer
     message.deliver!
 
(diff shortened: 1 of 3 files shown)


app/controllers/admin/api/messages_controller.rb
```diff
@@ -18,7 +18,7 @@ class Admin::Api::MessagesController < Admin::Api::BuyersBaseController
   ##~ op.parameters.add :name => "body", :description => "Text to send", :dataType => "string", :required => true, :paramType => "query", :allowMultiple => false
   #
   def create
-    message = current_account.messages.build params[:message]
+    message = current_account.messages.build params[:message].permit(:body)
```
akostadinov, line 21: we permit user to set body but not subject?

spec/acceptance/api/message_spec.rb
```diff
@@ -9,9 +9,14 @@
...
     post '/admin/api/accounts/:account_id/messages.:format', action: :create do
       parameter :body, 'Message Body'
+      parameter :subject, 'Message Subject'
       let(:body) { 'This is some text' }
+      let(:subject) { 'custom subject' }
 
-      after { serializable.body.should == body }
+      after do
+        serializable.body.should == body
+        serializable.subject.should == subject
```
akostadinov, line 18: I'm confused here again. Is `subject` here the parameter subject or the `let(:subject)`? And if it is `let(:subject)`, then why parameter `subject` is ignored?

akostadinov (approved):

PR #2891, opened 2022-01-18
Sphinx index only non-deleted accounts

akostadinov:
we may also need to add a callback when state of account changes to `scheduled_for_deletion`. But this probably will quickly reduce index size.

### DIFF
--- a/test/unit/account/search_test.rb
+++ b/test/unit/account/search_test.rb
@@ -65,28 +65,6 @@ class Account::SearchTest < ActiveSupport::TestCase
     assert_equal 'fo\/o', Account.search_ids('fo/o').query
   end
 
-  test 'search with query by email' do
-    ThinkingSphinx::Test.rt_run do
-      perform_enqueued_jobs(only: SphinxIndexationWorker) do
-        provider = FactoryBot.create(:simple_provider)
-        buyers = FactoryBot.create_list(:simple_buyer, 2, provider_account: provider)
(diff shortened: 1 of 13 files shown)


test/unit/account/search_test.rb
```diff
@@ -77,6 +77,10 @@ class Account::SearchTest < ActiveSupport::TestCase
         buyers = provider.buyer_accounts.scope_search(:query => 'foo@bar.co.example.com')
         expected_buyer = User.find_by!(email: 'foo@bar.co.example.com').account
         assert_equal [expected_buyer], buyers
+
+        # incremental index update check
+        provider.smart_destroy
+        assert_equal [], provider.buyer_accounts.scope_search(:query => 'foo@bar.co.example.com')
```
josemigallas, line 83: This is totally expected because the buyer has been removed. What we need to assert is whether the index still exists.
I've been using this to get the number of indexed records:
```
conn = ThinkingSphinx::Connection.new
res = conn.execute("show index account_core status")
res.map(&:values).to_h["indexed_documents"]
```
Maybe there's a way to query the index directly but at least we should assert that `indexed_documents` is short by the number of buyers deleted (2 in this case).

josemigallas (changes requested):

test/unit/account/search_test.rb
akostadinov: This is searching the index. If my indexer changes are not applied, then the record is still found.

akostadinov: Anyway, updated to catch regardless of record existence.

PR #2893, opened 2022-01-24
Improve CMS API docs

sergioifg94:

Improve CMS API docs: Fix typos and update summary of operations in order to align with current styling

**Verification steps** 

image

### DIFF
--- a/doc/active_docs/CMS API.json
+++ b/doc/active_docs/CMS API.json
@@ -1,4 +1,7 @@
 {
+  "resourcePath": "/admin/api/cms/templates",
+  "swaggerVersion": "1.1",
+  "apiVersion": "1.0",
   "apis": [
     {
       "path": "/admin/api/cms/files.xml",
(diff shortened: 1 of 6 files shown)


doc/active_docs/CMS API.json
```diff
@@ -6,8 +6,9 @@
       "operations": [
         {
           "httpMethod": "GET",
-          "summary": "List all files",
+          "summary": "File List",
```
guicassolato, line 9: We usually do not touch these files directly, but change instead in https://github.com/3scale/porta/blob/master/app/controllers/admin/api/cms/files_controller.rb#L37

...and then:

```
doc:swagger:generate:cms
```

...which generates https://github.com/3scale/porta/blob/427633bf033c58076987f04d75e0b525cbe9c35f/doc/active_docs/cms.json#L9.

Some minor details might differ on the base branch you are working on (i.e. proper `docs/active_docs/CMS API.json` file now), but the overall flow to generate the docs remains the same.

guicassolato (changes requested):

doc/active_docs/CMS API.json
sergioifg94: Thanks for pointing that our @guicassolato I updated the PR with the changes in the controllers to generate the equivalent documentation, using `source2swagger` in a similar way as the other controllers are documented

akostadinov (dismissed):
Looks awesome.

sergioifg94:
@akostadinov I updated the PR to base on master as we discussed yesterday. Also verified that the documentation is still not customer facing

akostadinov:
incorporated into #2667, will merge together with rails branch so closing this one. Thanks a lot!

PR #2900, opened 2022-02-07
fix: Allow tenant_id to be propagated from account

akostadinov:
supersedes #2759

fixes THREESCALE-7676


akostadinov:
the solution to set `master` to `false` does not work because of the unique constraint. Only one record can be `true`, one `false` and everything else must be `null`. So we need to update the triggers to match `null` condition as well.

PR #2903, opened 2022-02-08
Quick starts

josemigallas:

[THREESCALE-8283: Integrate Quick Starts into 3scale](https://issues.redhat.com/browse/THREESCALE-8283)


junezhang:
LGTM

app/views/provider/admin/quick_starts/show.html.slim
```diff
@@ -0,0 +1,12 @@
+#quick-start-catalog-container
+
+javascript:
+  // Hack of the year, part 2: For QuickStartCatalogPage to work we need to render it as a children
+  // of QuickStartContainer, we can't just render it in another pack here. This moves the node from
+  // within QuickStartContainer to #quick-start-catalog-container
```
akostadinov, line 6: Was any non-hack solution discussed?

app/views/shared/provider/_header.html.slim
```diff
@@ -28,6 +28,9 @@ header.pf-c-page__header id="user_widget" class='Header' class=('Header--master'
         - if saas?
           li.PopNavigation-listItem
             = fixed_width_icon_link_to "What's new?", 'leaf', '//access.redhat.com/articles/3107441#newfeaturesenhancements', class: 'PopNavigation-link', target: '_blank'
+        - if provider_can_use?(:quick_starts)
+          li.PopNavigation-listItem
+            = fixed_width_icon_link_to 'Quick starts', 'book', provider_admin_quick_starts_path, class: 'PopNavigation-link'
```
akostadinov, line 33: I thought this should be one word `Quickstarts`.

app/controllers/provider/admin/quick_starts_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+  activate_menu :quick_starts
+
+  layout 'provider'
+
+  def show; end
+
+  private
+
+  def hide_quick_starts
+    raise ActionController::RoutingError, '' unless provider_can_use?(:quick_starts)
```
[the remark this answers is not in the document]
josemigallas: It sounds like it should be for the entire product, yes.

test/functional/provider/admin/quick_starts_controller_test.rb
```diff
@@ -0,0 +1,34 @@
...
+    rolling_updates_on
+    provider = FactoryBot.create(:provider_account)
+    login_provider(provider)
+  end
+
+  class RollingUpdateOnTest < self
+    def setup
+      super
+      rolling_update(:quick_starts, enabled: true)
+    end
```
akostadinov, line 16: If you use `setup do`, then no need to call `super`.

app/views/provider/admin/quick_starts/show.html.slim
josemigallas: No, not really, but if we don't want to use a hack we need to implement the whole thing from scratch. Patternfly-react already takes care of everything BUT it's meant to be used in a React application, which we don't have.
The non-hack option is to use plain React (html/css) and implement ourselves the business logic... it's a no-go.

features/provider/header.feature
```diff
@@ -0,0 +1,12 @@
...
+  Background:
+    Given a provider is logged in
+
+  Scenario: Help menu dropdown
+    Then the help menu should have the following items:
+      | Customer Portal  |
+      | 3scale API Docs  |
+      | Liquid Reference |
+      | What's new?      |
+      | Quick starts     |
```
akostadinov, line 12: Should this depend on rolling updates?

josemigallas: Yes indeed, good catch

features/provider/admin/quick_starts.feature
```diff
@@ -0,0 +1,40 @@
...
+      Given I have rolling updates "quick_starts" disabled
+      And a provider is logged in
+
+    Scenario: Header help menu does not have the option
+      Then I should not be able to go to the quick start catalog from the help menu
+
+    Scenario: Following a Quick Start does not show it
+      Given I am following a quick start
+      When I go anywhere else
+      Then I won't be able to see the quick start
```
akostadinov, line 40: how come one can follow a quickstart if they are disabled? Maybe I'm missing something :)

app/views/shared/provider/_header.html.slim
akostadinov: https://en.wiktionary.org/wiki/quickstart

but I don't know, I see it written as people want everywhere

features/provider/admin/quick_starts.feature
josemigallas: It's a bit of a hacked test... the progress of a quick start is stored in a cookie. So, even if the user can't start a quick start manually, if they add the appropriate cookie the quick start would show up 😂 
Test scenario: Say we enable Quick starts, some customers use them, find a bug and we rollback and disable them. Users would still have their cookies stored.. so Quick start could accidentally show up.

app/controllers/provider/admin/quick_starts_controller.rb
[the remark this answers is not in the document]
akostadinov: In this case better use a feature like in email-configurations feature. But why do we need that at all? Why should we want to turn it on/off? Don't we want to just have it there and that's it. Unconditionally?

features/provider/admin/quick_starts.feature
josemigallas: It's just a theoretical scenario, to have more test coverage

app/views/shared/provider/_header.html.slim
josemigallas: Not to Patternfly guys it isn't 😂 https://quickstarts.netlify.app
Anyway, I renamed all our files and code references to quickstart instead of quick_start. The microcopy we leave it to Docs.

deernano:
UXD create a doc to review the quick starts:
https://docs.google.com/document/d/1_3esBC2Y1URDNYsFGLMTnQgAEcrAdZiBJGdjWEoQqWs/edit

Overall it looks good , just some tiny issues. Please take a look. THX!

akostadinov (approved):
Looks good. If it was me I wouldn't make this feature conditional, but it is totally fine.

PR #2904, opened 2022-02-08
THREESCALE-7676 fix tenant_id triggers

akostadinov:
supersedes #2759 and #2900 

We can't set `Account#master` to `false` because there is a database unique index on that field. It should be `NULL` by design. That's why fix triggers.

### DIFF
--- a/app/workers/set_tenant_id_worker.rb
+++ b/app/workers/set_tenant_id_worker.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+class SetTenantIdWorker < ApplicationJob
+
+  def perform(provider)
+    provider.update_column(:master, false)
+
+    [:backend_apis, :log_entries, :alerts].each do |relation|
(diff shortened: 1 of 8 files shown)


app/workers/set_tenant_id_worker.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+class SetTenantIdWorker < ApplicationJob
+
+  def perform(provider)
+    provider.update_column(:master, false)
```
akostadinov, line 6: This needs to be removed, `master` should stay null because of a unique index on the column. Just not to forget.

app/workers/set_tenant_id_worker.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    [:backend_apis, :log_entries, :alerts].each do |relation|
```
akostadinov, line 8: use `%i`

PR #2906, opened 2022-02-09
[params] Update `symbolized_settings` for rails 5.2

josemigallas:


akostadinov:
merged already to rails

PR #2910, opened 2022-02-14
🥒 Tests for Backend APIs

josemigallas:

There are some JIRA issues reporting errors in the Backend API forms and even though they should be fixed there are no tests to back them up.

[THREESCALE-8303: Update/improve Backend API related cukes](https://issues.redhat.com/browse/THREESCALE-8303)


features/step_definitions/web_steps.rb
```diff
@@ -32,6 +32,7 @@ def with_scope(locator)
 end
 
 When /^(?:|I )follow( invisible)? "([^"]*)"(?: within "([^"]*)")?$/ do |invisible, link, selector|
+
```
josemigallas, line 35:
```suggestion
```

features/provider/admin/backend_apis/show.feature
```diff
@@ -0,0 +1,16 @@
+@javascript
+Feature: Backend API overview page
+
+  Background:
+    Given a provider is logged in
+
+  Scenario: Products used by backend table
+    Given a backend api that is being used by a product
+    When an admin is in the backend over page
```
josemigallas, line 9:
```suggestion
    When an admin is in the backend overview page
```

features/provider/dashboard/widgets.feature
```diff
@@ -1,22 +1,14 @@
...
+    When an admin needs to find a product or backend quickly
+    Then the most recently updated products and backends can be found in the dashboard
 
-  @javascript
-  Scenario: Backends widget
-    Given I should see "My Fancy Backend API" in the apis dashboard backends widget
-    And I should see "My Regular Backend API" in the apis dashboard backends widget
+  Scenario: Add new products and backends quickly
+    When an admin needs a new product or backend quickly
+    Then products and backends can be created from the dashboard
```
akostadinov, line 14:
```suggestion
    When an admin goes to the admin dashboard page
    Then products can be created from the dashboard
    And backends can be created from the dashboard
```

akostadinov (commented):
Just suggestions. If you have better idea, that would be nice.

features/provider/admin/backend_apis/new.feature
```diff
@@ -0,0 +1,21 @@
+@javascript
+Feature: Backend API new page
+
+  Background:
+    Given a provider is logged in
+
+  Scenario: Create a new Backend API
+    When an admin goes to the backend apis page
+    Then they can create a new backend api
+    And be redirected to the new backend api overview page
```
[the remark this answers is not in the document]
josemigallas:
```cucumber
And they create a new backend api
Then they are redirected to the new backend api overview page
```

features/provider/dashboard/widgets.feature
josemigallas:
```cucumber
When an admin is at the dashboard
Then products can be created from the dashboard
And backends can be created from the dashboard
```

features/provider/admin/backend_apis/edit.feature
```diff
@@ -0,0 +1,29 @@
+@javascript
+Feature: Backend API edit page
+
+  Background:
+    Given a provider is logged in
+
+  Scenario: Edit an existing backend API
+    Given an admin is at a backend api edit page
+    Then they will be able to edit the name, description and base url
+    But will not be able to edit its system name
```
akostadinov, line 10: The scenario looks really nice. I have a concern about step conflict and discoverability. Because step `they will be able to edit the name, description and base url` is probably very backend API specific but does not indicate that it is about backend APIs. So maybe one will want to create same step for somehting else.

I don't know, does it make sense? How about 
```suggestion
    Then they will be able to edit backend api name, description and base url
    But will not be able to edit backend api system name
```

If you think that discoverability and step collisions will not be a problem, then leaving them as they are is ok.

features/provider/admin/backend_apis/edit.feature
```diff
@@ -0,0 +1,29 @@
...
+    Given an admin is at a backend api edit page
+    Then they will be able to update it with an existing name and url
+
+  Scenario: Delete a backend API used by product
+    Given a backend api that is being used by a product
+    When an admin tries to delete the backend api from its edit page
+    Then it is not possible to delete the backend
+
+  Scenario: Delete a backend API not used by any product
+    Given a backend api that is not being used by any products
```
akostadinov, line 27: This seems to be another step dup for `a backend`. Or does something more?

features/provider/admin/backend_apis/edit.feature
```diff
@@ -0,0 +1,29 @@
...
+    When an admin tries to delete the backend api from its edit page
+    Then it is possible to delete the backend
```
akostadinov, line 29: Like above, all steps in this file, if they are not reusable for something other than backend api pages, perhaps put somehow `backend api` in them to improve discoverability and avoid collisions.

features/provider/admin/backend_apis/edit.feature
josemigallas: These particular steps were intended like this. It should work in any form, as long as it uses those fields exactly.

features/provider/admin/backend_apis/edit.feature
josemigallas: No, I guess they're the same. I wanted to make it explicit that no product is using the backend, so maybe it should add an assertion to make sure of this. In that case they wouldn't be duplicates.

features/provider/admin/backend_apis/new.feature
```diff
@@ -0,0 +1,21 @@
...
+    And they create a new backend api
+    Then they are redirected to the new backend api overview page
+
+  Scenario: Form validation
+    When an admin is creating a new backend api
+    Then it is not possible to create it without a name, a valid url or system name
+    But it is possible to create it without system name
+
+  Scenario: Create a Backend API with duplicate fields
+    Given a backend api
```
akostadinov, line 18: Another `a backend api`

features/provider/admin/backend_apis/edit.feature
akostadinov: I think at every place this step required the backend not to be used.

My request is to get all `Given a backend` step to use the same format. Other than that LGTM, now scenarios are much clearer.

features/provider/admin/backend_apis/edit.feature
josemigallas: But it this case we're good, no? All 3 steps refer to a backend.

features/provider/admin/backend_apis/edit.feature
josemigallas: There are 4 step the creates a backend, we can split them so that `Given a backend` is always used first:
```ruby
Given "a backend( api)" do
Given "an admin is at a backend api edit page" do
Given "a backend api that {is} being used by a(ny) product(s)" do |used|
Given 'a backend that is unavailable' do
```

```cucumber
Given a backend api
And an admin is at the backend edit page

Given a backend api
And the backend is (not) used by a product

Given a backend api
And the backend api is unavailable
```
WDYT?

Bonus question: are backends used by default? The default product always has a backend but any further new products/backends are created independently.

akostadinov (approved):

PR #2912, opened 2022-02-14
Fix deprecated usage of `attribute_changed?` in callbacks

josemigallas:
Rails changed the behavior of `attribute_changed?` in after and before callbacks. Older versions will throw deprecation warnings indicating which methods to use instead.
```
DEPRECATION WARNING: The behavior of `attribute_changed?` inside of after callbacks will be changing in the next version of Rails. The new return value will reflect the behavior of calling the method after `save` returned (e.g. the opposite of what it returns now). To maintain the current behavior, use `saved_change_to_attribute?` instead.
```
This PR reviews usages of `attribute_changed?` and replace it with `saved_change_to_attribute?` when necessary.
More info:
* https://rubyinrails.com/2019/04/08/rails-changed-behavior-of-attribute-changed-in-callbacks/
* https://github.com/rails/rails/issues/29035#issuecomment-423085336

[THREESCALE-8109: The behavior of `attribute_changed?` will be changing](https://issues.redhat.com/browse/THREESCALE-8109)

### DIFF
--- a/app/models/proxy.rb
+++ b/app/models/proxy.rb
@@ -106,7 +106,7 @@ class Proxy < ApplicationRecord
   after_save :publish_events
   before_destroy :publish_events
 
-  after_save :track_apicast_version_change, if: :apicast_configuration_driven_changed?
+  after_save :track_apicast_version_change, if: :saved_change_to_apicast_configuration_driven?
 
   alias_attribute :production_endpoint, :endpoint
   alias_attribute :staging_endpoint, :sandbox_endpoint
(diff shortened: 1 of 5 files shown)


app/models/proxy.rb
```diff
@@ -269,7 +269,8 @@ def self.credentials_collection
   end
 
   def set_correct_endpoints?
-    apicast_configuration_driven_changed? || new_record?
+    # Verified
```
akostadinov, line 272: is this telling something or just personal note that it was tested?

josemigallas: Personal note, I thought I removed all of them 😓

akostadinov (approved):

PR #2913, opened 2022-02-15
THREESCALE-8108 removed config to fix deprecation

nidhi-soni1104:

To fix depreciation
**DEPRECATION WARNING**: ActiveSupport.halt_callback_chains_on_return_false= is deprecated

https://issues.redhat.com/browse/THREESCALE-8108


akostadinov (changes requested):
@nidhi-soni1104 , before this can be merged, please explain what are the consequences of changing this as well how did you make sure that it is not an issue with our existing callbacks.

This is an important far-reaching change and it deserves time for a thorough review of what can and cannot be affected by it.

akostadinov:
This options seems not deprecated but removed already in Rails 5.1. See rails/rails@9e98f3f7e61 and rails/rails@4e63ce53fc2.
In such case that option already does nothing and we can just remove it. It seems that no issues have been observed since we upgraded to Rails 5.1

@jkeam , am I missing anything?

akostadinov (approved):
Approving this, we're gonna test it as part of rails upgrade.

PR #2915, opened 2022-02-16
Prevent email from being sent in bulk to Subscriptions without subject/body

nidhi-soni1104:

When sending emails from the bulk operations emails for subscriptions can be sent if body and subject is empty
make those field required

Fixed subtask 

Main task : https://issues.redhat.com/browse/THREESCALE-8009

SubTask: https://issues.redhat.com/browse/THREESCALE-8145

### DIFF
--- a/features/old/services/subscriptions/bulk_operations/send_emails.feature
+++ b/features/old/services/subscriptions/bulk_operations/send_emails.feature
@@ -35,10 +35,7 @@ Feature: Mass email bulk operations
     And I fill in "Body" with ""
     And I press "Send"
     Then I should see "Selected Service Subscriptions"
-      And "jane@me.us" should not receive an email with the following subject:
-      """
-        Nothing to say
-      """
+      And a clear email queue
 
   Scenario: Emails can't be sent without subject
     Given provider "foo.3scale.localhost" has "service_plans" visible
@@ -50,10 +47,7 @@ Feature: Mass email bulk operations
     And I fill in "Body" with "Did I forget to add a subject?"
     And I press "Send"
     Then I should see "Selected Service Subscriptions"
-      And "jane@me.us" should not receive an email with the following body:
-      """
-        Did I forget to add a subject?
-      """
+      And a clear email queue
 
   Scenario: Send mass email to application owners
       And provider "foo.3scale.localhost" has "service_plans" visible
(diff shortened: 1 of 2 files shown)


features/old/services/subscriptions/bulk_operations/send_emails.feature
```diff
@@ -25,34 +25,27 @@ Feature: Mass email bulk operations
...
     And I press "Send email"
     And I fill in "Subject" with "Nothing to say"
     And I fill in "Body" with ""
-    And I press "Send" and I confirm dialog box within colorbox
-    Then I should see "Action completed successfully"
-    And "jane@me.us" should receive an email with the following body:
-    """
-    """
+    And I press "Send"
+    Then I should see "Selected Service Subscriptions"
```
josemigallas, line 37: Could we assert instead `And "jane@me.us" should not receive an email with the following body:`? The test has turned a bit vague.

josemigallas: This goes also for the previous PR.

josemigallas (changes requested):
We could add better tests

features/old/services/subscriptions/bulk_operations/send_emails.feature
nidhi-soni1104: @josemigallas  I have updated test case

Now test is clear i think,

```
And I press "Send"
Then I should see "Selected Service Subscriptions"
  And a clear email queue
```

These line explains that we after pressing send button we still remain on same page where subscriptions are selected and email queue is also clear which tends to no email shoot.

Also if I say about the UI part when user don't enter body and subject now they are required filed so new action will be trigerred.

nidhi-soni1104:
> We could add better tests

If still better updates need let me know, or else I will do same changes for other PR's

features/old/services/subscriptions/bulk_operations/send_emails.feature
```diff
@@ -25,34 +25,29 @@ Feature: Mass email bulk operations
...
+      And a clear email queue
```
josemigallas, line 38: This is clearing the email queue rather than asserting it's empty, see: https://github.com/3scale/porta/blob/f6e98d1d8a66f28d13a91ebba8767e12814509ae/features/step_definitions/email_steps.rb#L49-L51

nidhi-soni1104: uisng **no emails have been sent** will be correct, because this is what I am checking.

nidhi-soni1104: @josemigallas Updated

josemigallas (approved):

PR #2922, opened 2022-02-25
Reduce times `by_cookie_token` query is executed

josemigallas:
It seems like this method it being called many times even tho the updated values are already `nil`.

### DIFF
--- a/app/lib/authentication/by_cookie_token.rb
+++ b/app/lib/authentication/by_cookie_token.rb
@@ -51,11 +51,10 @@ module Authentication
       # always be deleted together.
       #
       def forget_me
-        self.remember_token_expires_at = nil
-        self.remember_token            = nil
-        self.class.where(id: id).update_all(:remember_token_expires_at => nil, :remember_token => nil) # TODO: replace by update_columns
-      end
+        return if remember_token_expires_at.nil? && remember_token.nil?
 
+        update_columns(remember_token_expires_at: nil, remember_token: nil) # rubocop:disable Rails/SkipsModelValidations
+      end
     end # instance methods
   end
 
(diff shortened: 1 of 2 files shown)


app/lib/authentication/by_cookie_token.rb
```diff
@@ -50,12 +50,11 @@ def refresh_token
       # client-side (browser cookie) and server-side (this remember_token) must
       # always be deleted together.
       #
-      def forget_me
-        self.remember_token_expires_at = nil
-        self.remember_token            = nil
-        self.class.where(id: id).update_all(:remember_token_expires_at => nil, :remember_token => nil) # TODO: replace by update_columns
-      end
+      def new_forget_me
```
josemigallas, line 53: Is this equivalent to the method above or am I missing somehting?

app/lib/authentication/by_cookie_token.rb
```diff
@@ -50,12 +50,11 @@ def refresh_token
...
+        return if remember_token_expires_at.nil? && remember_token.nil?
 
+        update(remember_token_expires_at: nil, remember_token: nil)
```
josemigallas, line 56: Is this equivalent to the old implementation or am I missing somehting?

app/lib/authentication/by_cookie_token.rb
```diff
@@ -51,11 +51,10 @@ def refresh_token
       # always be deleted together.
       #
       def forget_me
-        self.remember_token_expires_at = nil
-        self.remember_token            = nil
```
akostadinov, line 55: Shouldn't we still set these?

josemigallas: Isn't it done by the update? If not... we should probably test it. Somehow.

akostadinov: I don't think `#update_clomuns` will update current object.

akostadinov: maybe it does

josemigallas: Ah, apparently it does. I'll add a test for this

akostadinov (approved):

PR #2929, opened 2022-03-07
Update deprecated usage of `attribute_changed?` in callbacks

josemigallas:
#2912 was removed from `rails` before deploying so here we are again.

### What this PR does / why we need it
Rails changed the behavior of `attribute_changed?` in after and before callbacks. Older versions will throw deprecation warnings indicating which methods to use instead.
```
DEPRECATION WARNING: The behavior of `attribute_changed?` inside of after callbacks will be changing in the next version of Rails. The new return value will reflect the behavior of calling the method after `save` returned (e.g. the opposite of what it returns now). To maintain the current behavior, use `saved_change_to_attribute?` instead.
```
This PR reviews usages of `attribute_changed?` and replace it with `saved_change_to_attribute?` when necessary.

### More info:
* [Rails Changed Behavior of Attribute Changeed in Callbacks](https://rubyinrails.com/2019/04/08/rails-changed-behavior-of-attribute-changed-in-callbacks/)
* [Conversion Table](https://github.com/rails/rails/pull/25337#issuecomment-225166796)

### Which issue(s) this PR fixes

[THREESCALE-8109: The behavior of `attribute_changed?` will be changing](https://issues.redhat.com/browse/THREESCALE-8109)

### TODO
- [ ] Double check changes with the conversion table.
- [ ] Test with debugger, add some test cases if possible


akostadinov (commented):
Amazing work. Most places are clearly solid as far as I can tell.

We can merge these immediately.

About the other cases, we can investigate more of course. But also we can for a week keep the original and add a log message when the new one is used so that we see how the deployed cluster actually works. I'm concerned that tests may not call the methods always at the time actual code does.

app/models/proxy.rb
```diff
@@ -278,7 +280,11 @@ def publish_events
...
 
-  DEPLOYMENT_OPTION_CHANGED = ->(record) { record.changed_attributes.key?(:deployment_option) }
+  DEPLOYMENT_OPTION_CHANGED = ->(record) {
+    raise 'crap' if record.saved_changes.key?(:deployment_option) != record.changed_attributes.key?(:deployment_option)
+
+    record.saved_change_to_attribute?(:deployment_option)
+  }
 
   def deployment_option_changed?
     [ self, service ].any?(&DEPLOYMENT_OPTION_CHANGED)
```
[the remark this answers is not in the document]
josemigallas: I think I've tried every new method in every scenario and `saved_change_to_attribute?` is the only that passes all tests and scenarios. I will double check this one just in case.

josemigallas:
> Amazing work. Most places are clearly solid as far as I can tell.
> 
> We can merge these immediately.
> 
> About the other cases, we can investigate more of course. But also we can for a week keep the original and add a log message when the new one is used so that we see how the deployed cluster actually works. I'm concerned that tests may not call the methods always at the time actual code does.

Yeah, and I think some cases used it "wrong". I think we should analyze the business logic behind each one and decide which method is the proper one. Like, the vat update is supposed to trigger an update to all invoices, so after_save makes sense.

app/models/proxy.rb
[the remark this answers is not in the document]
akostadinov: maybe we can inspect the backlogs when it fails to see where it comes from and whether it is legit.

app/models/proxy.rb
```diff
@@ -278,10 +278,10 @@ def publish_events
     nil
   end
 
-  DEPLOYMENT_OPTION_CHANGED = ->(record) { record.changed_attributes.key?(:deployment_option) }
-
   def deployment_option_changed?
-    [ self, service ].any?(&DEPLOYMENT_OPTION_CHANGED)
+    [self, service].any? do |record|
+      record.respond_to?(:will_save_change_to_deployment_option?) && record.will_save_change_to_deployment_option?
```
josemigallas, line 283:
```suggestion
      record.will_save_change_to_attribute?(:deployment_option)
```

app/events/oidc/proxy_changed_event.rb
```diff
@@ -21,6 +21,9 @@ def self.create(proxy)
   def self.valid?(proxy)
     service = proxy.try(:service)
     return unless service
-    service.backend_version.oauth? || service.backend_version_change&.include?('oauth')
+
+    raise 'crap' if !!(service.backend_version.oauth? || service.backend_version_change_to_be_saved&.include?('oauth') || service.saved_change_to_backend_version&.include?('oauth')) != !!(service.backend_version.oauth? || service.backend_version_change&.include?('oauth'))
+
+    service.backend_version.oauth? || service.backend_version_change_to_be_saved&.include?('oauth') || service.saved_change_to_backend_version&.include?('oauth')
```
josemigallas, line 27:
```suggestion
    # TODO: second assertion is probably useless, or not. Revisit this in the future.
    service.backend_version.oauth? || service.backend_version_change_to_be_saved&.include?('oauth') || service.saved_change_to_backend_version&.include?('oauth')
```

app/events/oidc/proxy_changed_event.rb
```diff
@@ -21,6 +21,9 @@ def self.create(proxy)
   def self.valid?(proxy)
     service = proxy.try(:service)
     return unless service
-    service.backend_version.oauth? || service.backend_version_change&.include?('oauth')
+
+    raise 'crap' if !!(service.backend_version.oauth? || service.backend_version_change_to_be_saved&.include?('oauth') || service.saved_change_to_backend_version&.include?('oauth')) != !!(service.backend_version.oauth? || service.backend_version_change&.include?('oauth'))
```
josemigallas, line 25:
```suggestion
```

app/models/proxy.rb
```diff
@@ -278,15 +278,16 @@ def publish_events
...
-    [ self, service ].any?(&DEPLOYMENT_OPTION_CHANGED)
+    [self, service].any? do |record|
+      record.changed_attributes.key?(:deployment_option)
+    end
   end
 
   # We want to autosave when Service#deployment_option changed
   def changed_for_autosave?
-    deployment_option_changed? or super
+    association(:service).loaded? && service.will_save_change_to_attribute?(:deployment_option) or super
```
akostadinov, line 290: First of all `#deployment_option_changed?` is overriding the deprecated method of the `attribute_changed?` family. So I believe we want to just get rid of it.

I have tried `changed_for_autosave?` method and it returns `false` inside an `after-save` callback as well after a `save` call. So I believe equivalent is `will_save_change..`.

Another thing I checked is that `Proxy` lacks a `:deployment_option` attribute at all. So this seems to be some old code from the distant past when that attribute was being migrated from `Proxy` to `Service`.

That's why I leave only check for `service`.

A small optimization is to check whether `association(:service)` was loaded to avoid a database call in case it was not. If object was not loaded, then it can't have any changes applied so that would be a worthless call.

app/models/proxy.rb
```diff
@@ -278,15 +278,16 @@ def publish_events
     nil
   end
 
+  # TODO: remove unused method, there is no such attribute
```
josemigallas, line 281: Is it used here?
https://github.com/3scale/porta/blob/de8c865c1022ca7d6b981fdb9bae14e2806def03/app/models/service.rb#L494

app/models/proxy.rb
josemigallas: Makes sense!

test/unit/proxy_test.rb
```diff
@@ -659,6 +659,28 @@ def test_set_correct_endpoints
     @proxy.destroy
   end
 
+  test '#changed_for_autosave?' do
+    # no changes
+    assert_not @proxy.changed_for_autosave?
+
+    # deployment option changed
+    @proxy.expects(:deployment_option_changed?).returns(true)
```
josemigallas, line 667:
```suggestion
    @proxy.service.expects(:will_save_change_to_attribute?).with(:deployment_option).returns(true)
```

test/unit/service_test.rb
```diff
@@ -532,12 +532,6 @@ class AsynchronousDeletionOfService < ActiveSupport::TestCase
     end
   end
 
-  test 'don not call update endpoint if deployment option has not changed' do
-    service = FactoryBot.create(:simple_service, deployment_option: 'hosted')
-    service.expects(:deployment_option_changed).times(0)
```
akostadinov, line 537: this method does not exist so the test doesn't make any sense

lib/tasks/proxy.rake
```diff
@@ -33,11 +33,6 @@ namespace :proxy do
     end
   end
 
-  desc 'Fixing nil endpoint for hosted'
-  task :set_correct_endpoint_hosted => :environment do
-    Service.includes(:proxy).where(proxies: {endpoint: nil}, deployment_option: 'hosted').find_each(&:deployment_option_changed)
```
akostadinov, line 38: this seems to be an old fix that no longer is necessary. Not necessary because such method does not exit since long time.

akostadinov (approved):

akostadinov:
Will merge provided no errors. But we need to test with debug statements before deploying to master.

PR #2935, opened 2022-03-10
THREESCALE-8141 Prevent emails from being sent without subject and/or body for accounts

nidhi-soni1104:

make those field required

Which issue(s) this PR fixes

Main task : https://issues.redhat.com/browse/THREESCALE-8009

SubTask: https://issues.redhat.com/browse/THREESCALE-8141

### DIFF
--- a/features/old/accounts/bulk_operations/send_email.feature
+++ b/features/old/accounts/bulk_operations/send_email.feature
@@ -28,7 +28,7 @@ Feature: Mass email bulk operations
     And I fill in "Subject" with "Nothing to say"
     And I press "Send"
     Then I should see "Selected Accounts"
-      And "jane@me.us" should receive no emails
+    And "jane@me.us" should receive no emails
 
   Scenario: Emails can't be sent without subject
     And I am on the accounts admin page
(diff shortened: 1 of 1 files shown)


josemigallas (approved):

josemigallas (dismissed):

nidhi-soni1104:
> Looks good but indentation is screwed.


@akostadinov This is how it is supposed to be I think because other examples are also same so intended this way.

josemigallas:
> > Looks good but indentation is screwed.
> 
> @akostadinov This is how it is supposed to be I think because other examples are also same so intended this way.

No, it is screwed 😂 not a big deal because these tests are obsolete anyway. But it would be nice to use the proper indentation.

nidhi-soni1104:
> > > Looks good but indentation is screwed.
> > 
> > 
> > @akostadinov This is how it is supposed to be I think because other examples are also same so intended this way.
> 
> No, it is screwed joy not a big deal because these tests are obsolete anyway. But it would be nice to use the proper indentation.

Okay then let me reindent

nidhi-soni1104:
@akostadinov @josemigallas Changes are done related to indentation

akostadinov (approved):

PR #2943, opened 2022-03-15
fix NoMethodError: undefined method for nil:NilClass

akostadinov:
We observed error from ActionMailer background job

>  NoMethodError · undefined method `buyer?' for nil:NilClass

It seems to appear when account got deleted before delivery happens.

This is just logging the error. But PR is in WIP status because
I try to make sure that we don't try to deliver such messages when
for example an account is deleted. So far I don't see such occasion though.


josemigallas (approved):
This makes total sense to me and everything points to the account being deleted before the mail was delivered.

PR #2945, opened 2022-03-16
Attributes are now Parameters in user_controller's #update_attributes

josemigallas:

`attributes` used to be a hash, now it's ActionController::Parameters.. and sometimes it's still a hash 🤷🏻‍♂️ .

Error found https://app.bugsnag.com/3scale-networks-sl/system/errors/623154afea8d6b0008c052c5?event_id=623154af0092ebfb070a0000&i=sk&m=nw

### DIFF
--- a/app/controllers/provider/admin/account/users_controller.rb
+++ b/app/controllers/provider/admin/account/users_controller.rb
@@ -36,19 +36,22 @@ class Provider::Admin::Account::UsersController < Provider::Admin::Account::Base
     @users ||= end_of_association_chain.but_impersonation_admin.paginate(page: params[:page]).decorate
   end
 
-  def permitted_params
-    # TODO: this should be handled by cancancan
-    permitted = current_account.provider_can_use?(:service_permissions) ? %i[role member_permission_ids member_permission_service_ids] : %i[role member_permission_ids]
-    params.require(:user).permit(permitted)
-  end
-
   def update_resource(user, attributes)
     # FIXME: in rails 3, we're getting an array
     attributes = attributes.first
 
+    # After the rails 5.1 upgrade, attributes comes as ActionController::Parameters except when they are empty
+    attributes = attributes.permit!.to_h unless attributes.is_a?(Hash)
+
+    protected_attributes = attributes.extract!(*User::Permissions::ATTRIBUTES)
+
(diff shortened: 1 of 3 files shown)


app/controllers/provider/admin/account/users_controller.rb
```diff
@@ -36,19 +36,19 @@ def collection
...
-    # And in rails 5, it can be ActionController::Parameters or hash
-    attributes = ActionController::Parameters.new(attributes) unless attributes.is_a?(ActionController::Parameters)
-    protected_attributes = attributes.permit(User::Permissions::ATTRIBUTES)
-
-    protected_attributes.extract!(:member_permission_service_ids) unless current_account.provider_can_use?(:service_permissions)
-
     user.class.transaction do
       user.assign_attributes(attributes)
-      user.assign_attributes(protected_attributes, without_protection: can?(:update_role, user))
+      user.assign_attributes(attributes, without_protection: can?(:update_role, user))
```
akostadinov, line 51: I think we should extract `member_permission_service_ids` and don't save with it without the permission check. Something like:
```ruby
member_permission_service_ids = attributes.delete(:member_permission_service_ids)
```
then
```ruby
      user.assign_attributes(attributes)
      user.assign_attributes({member_permission_service_ids: member_permission_service_ids}, without_protection: can?(:update_role, user))
```

Otherwise we are saving without protection all attributes.

Although it is presently double work - we permit params, then `protected_attributes` gem still should take care. So we can remove protected attributes gem from model and make sure we follow same rules as protected attributes does. But that's more work.

akostadinov: Actually not sure we should use a Hash for the second assignment :(

josemigallas: I'm thinking about doing the 💩 approach:
```diff
-protected_attributes = attributes.extract!(*User::Permissions::ATTRIBUTES)
+protected_attributes = attributes.permit!.to_h.extract!(*User::Permissions::ATTRIBUTES)
```
Because by permitting only ATTRIBUTES we're gonna miss the rest of params (name, etc). And at least we maintain the current logic and avoid introducing new errors.

akostadinov (approved):

PR #2947, opened 2022-03-18
🐛 Fix undefined method strip with extra_fields to_xml

josemigallas:
JIRA: [THREESCALE-8082: Expected string in extra_fields.rb](https://issues.redhat.com/browse/THREESCALE-8082)

### DIFF
--- a/app/lib/fields/extra_fields.rb
+++ b/app/lib/fields/extra_fields.rb
@@ -101,6 +101,8 @@ module Fields::ExtraFields
       encode_string_extra_field(value)
     when Integer
       value.to_s
+    when Array
+      value.map { |v| encode_extra_field(v) }
     else
       value
     end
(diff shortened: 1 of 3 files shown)


akostadinov:
If there is no way to store integers except for direct database queries, maybe just fix database and avoid this `to_s`? I wonder..

test/unit/extra_fields_test.rb
```diff
@@ -177,10 +178,11 @@ def setup
     end
 
     # Regression: https://app.bugsnag.com/3scale-networks-sl/system/errors/61eefe2bd365260008097f85
-    test 'extra_field is an integer' do
+    test 'extra_field should not be store as an integer' do
```
akostadinov, line 181:
```suggestion
    test 'extra_field should not be stored as an integer' do
```

app/lib/fields/extra_fields.rb
```diff
@@ -99,6 +99,10 @@ def encode_extra_field(value)
     case value
     when String
       encode_string_extra_field(value)
+    when Integer
+      value.to_s
+    when Array
+      value.map { |v| encode_extra_field(v) }
```
akostadinov, line 105: Why do we need an array here? Should we allow arrays at all?

josemigallas: Yes, extra_fields support multiple selection. See https://github.com/3scale/system/pull/2752

PR #2949, opened 2022-03-18
THREESCALE-7676: limit number of jobs for tenant fixes

akostadinov:
First require user to enter relations that need fixing so that
they can be fixed individually.

Additionally perform updating in batches to reduce overal number of
jobs.

Also set the used queue to "low".

https://issues.redhat.com/browse/THREESCALE-7676

### DIFF
--- a/lib/tasks/multitenant/tenants.rake
+++ b/lib/tasks/multitenant/tenants.rake
@@ -20,6 +20,11 @@ namespace :multitenant do
       puts(query.any? ? 'Some of the tenants haven\t been suspended' : 'All the right tenants have been suspended')
     end
 
+    desc 'Fix in the background tenant_id missing in alerts, log entries and backend apis'
+    task :fix_missing_in_background => :environment do |_task, relations|
+      SetTenantIdWorker::BatchEnqueueWorker.perform_later(*relations.to_a)
+    end
+
     desc 'Fix empty or corrupted tenant_id in accounts'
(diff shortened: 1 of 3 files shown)


lib/tasks/multitenant/tenants.rake
```diff
@@ -20,6 +20,11 @@ namespace :multitenant do
       puts(query.any? ? 'Some of the tenants haven\t been suspended' : 'All the right tenants have been suspended')
     end
 
+    desc 'Fix in the background tenant_id missing in alerts, log entries and backend apis'
+    task :fix_missing_in_background => :environment do |_task, relations|
+      SetTenantIdWorker::BatchEnqueueWorker.perform_later(*relations.to_a)
+    end
+
```
nidhi-soni1104, line 27: I have just a confusion, how you are getting relations here and is that the reason to  move from fixes.rake to tenants.rake

akostadinov: I just found `fixes.rake` to be more appropriate. The relations user must pass when calling the task. For example
```
$ rake multitenant:tenants:fix_missing_in_background[backend_apis]
```

nidhi-soni1104: we are expecting realtion to be this array 
 
` relations = ["backend_apis", "log_entries", "alerts"]`

akostadinov: yes, one or more of these

akostadinov:
`Metrics` also have a faulty trigger because it only accounts for `Service` metrics and not `BackendApi` ones.

PR #2951, opened 2022-03-24
performant latest proxy_configs SQL query

akostadinov:
wanted to make #2921 use code and not direct SQL but failed. I found a simpler way to replace the existing scope though that seems also fast. I have not done proper performance comparison. I just thought having it as a scope has it's benefits.

btw this is related to rzane/baby_squeel#123, hopefully we can avoid direct SQL usage at some point


josemigallas:
Added to https://github.com/3scale/porta/pull/2921
Kudos! 🥇

PR #2955, opened 2022-03-30
THREESCALE-8115 added autocomplete off

nidhi-soni1104:

Autocomplete attribute is set to "on" for the credentials fields in the login page

Which issue(s) this PR fixes

https://issues.redhat.com/browse/THREESCALE-8115

### DIFF
--- a/app/javascript/src/LoginPage/loginForms/Login3scaleForm.jsx
+++ b/app/javascript/src/LoginPage/loginForms/Login3scaleForm.jsx
@@ -95,10 +95,11 @@ class Login3scaleForm extends React.Component<Props, State> {
         id='new_session'
         acceptCharset='UTF-8'
         method='post'
+        autoComplete="off"
       >
         <HiddenInputs/>
-        <TextField inputProps={usernameInputProps}/>
-        <PasswordField inputProps={passwordInputProps} />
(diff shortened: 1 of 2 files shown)


nidhi-soni1104:
> https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion#the_autocomplete_attribute_and_login_fields
> 
> _Even without a master password, in-browser password management is generally seen as a net gain for security. Since users do not have to remember passwords that the browser stores for them, they are able to choose stronger passwords than they would otherwise._
> 
> _For this reason, many modern browsers do not support autocomplete="off" for login fields._
> 
> Seems to me the jira issue is incorrect.
Then need to find alternate if we can but generally we cant mess up with browser setting related things

secretsurf:
I don't think this code is wrong though. It will work on browsers that support it and that's all we can do. I would merge this. What do you think, @thomasmaas?

nidhi-soni1104:
Closing this PR for now

nidhi-soni1104:
@thomasmaas @secretsurf Can we merge this?

nidhi-soni1104:
As per the comment mentioned by cat in Jira I am merging this PR and closing JIRA as well

PR #2956, opened 2022-04-01
Metric.service attribute is deprecated and can be NULL for metrics of BackendApi

akostadinov:


josemigallas (approved):
We should add some test for UsageLimitRepresenter

PR #2958, opened 2022-04-05
test representation of all usage limit type

akostadinov:
additionally `service` attribute of Metric is obsolete so stop using it
see bbd7051d92

One issue here is that still production code expects `service` attribute to be set when the owner is a service. So we can't fix all tests to stop referring to `#service`. One option is to fix the production code. But maybe better for now to still set `service` attribute in the factory.

At the same we shouldn't leave production code to rely on the `service` because we don't validate and database does not enforce that attribute. So this is just a step to removing reliance on it.

### DIFF
--- a/test/unit/metric_test.rb
+++ b/test/unit/metric_test.rb
@@ -56,8 +56,8 @@ class MetricTest < ActiveSupport::TestCase
   test 'system_name is not case sensitive' do
     service = FactoryBot.create(:simple_service)
 
-    metric_one = FactoryBot.create(:metric, service: service, system_name: 'frags')
-    metric_two = FactoryBot.create(:metric, service: service)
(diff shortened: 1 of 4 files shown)


test/unit/metric_test.rb
```diff
@@ -88,9 +88,9 @@ def test_destroyable?
 
   test 'fill owner' do
     service = FactoryBot.create(:simple_service)
-    service_metric = FactoryBot.build(:metric, service: service)
+    service_metric = service.metrics.build(system_name: "meth1")
```
akostadinov, line 91: This will test that metrics created the old way (with `service` instead of an `owner`) will automatically set `owner` before validation. Since Factory changed to use new style, we build directly which is IMO the correct way given that factory can change any time.

josemigallas (approved):

akostadinov:
superseded by #3006

PR #2960, opened 2022-04-06
THREESCALE-8347 fixed warnings Can't mass-assign protected attributes

nidhi-soni1104:

WARNING: Can't mass-assign protected attributes 
Added all the attributes which are mass assigned and removed duplicate from protected 

https://issues.redhat.com/browse/THREESCALE-8347

### DIFF
--- a/app/models/metric.rb
+++ b/app/models/metric.rb
@@ -1,6 +1,5 @@
 # TODO: parameter name was deprecated on 21.02.2014 and should be removed at one point.
 class Metric < ApplicationRecord
-  include ActiveModel::MassAssignmentSecurity
   include Backend::ModelExtensions::Metric
   include SystemName
   include BackendApiLogic::MetricExtension
@@ -53,22 +52,17 @@ class Metric < ApplicationRecord
     where.has { ((owner_type == 'Service') & owner_id.in(provider.services.pluck(:id))) | ((owner_type == 'BackendApi') & owner_id.in(provider.backend_apis.pluck(:id))) }
   }
 
-  def assign_attributes(values, options = {})
-    sanitize_for_mass_assignment(values, options[:as]).each do |key, value|
-      send("#{key}=", value)
-    end
-  end
-
   # Create one of the predefined, default metrics.
   #
   # == Arguments
   #
   # +type+:: Which default metric to create. Currently only :hits are supported.
   def self.create_default!(type, attributes = {})
-    metric = Metric.new
-    metric.assign_attributes({:friendly_name => 'Hits', :system_name => 'hits', :unit => 'hit',
-                                :description => 'Number of API hits'})
-    metric.service_id = attributes[:service_id]
+    id = attributes[:service_id]
+    attributes.delete(:service_id)
+    metric = new(attributes.merge(:friendly_name => 'Hits', :system_name => 'hits', :unit => 'hit',
+                                  :description => 'Number of API hits'))
+    metric.service_id = id
     metric.save!
     metric
   end
--- a/app/models/metric.rb
+++ b/app/models/metric.rb
@@ -54,8 +54,8 @@ class Metric < ApplicationRecord
   }
 
   def assign_attributes(values, options = {})
-    sanitize_for_mass_assignment(values, options[:as]).each do |k, v|
-      send("#{k}=", v)
+    sanitize_for_mass_assignment(values, options[:as]).each do |key, value|
+      send("#{key}=", value)
     end
   end
 
--- a/app/models/metric.rb
+++ b/app/models/metric.rb
@@ -26,7 +26,9 @@ class Metric < ApplicationRecord
 
   acts_as_tree
 
-  attr_protected :service_id, :parent_id, :tenant_id, :audit_ids
+  attr_protected :parent_id, :tenant_id, :audit_ids
+  attr_accessible :service_id, :friendly_name, :system_name, :unit, :description
+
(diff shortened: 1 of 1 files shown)


app/models/metric.rb
```diff
@@ -26,7 +26,9 @@ class Metric < ApplicationRecord
 
   acts_as_tree
 
-  attr_protected :service_id, :parent_id, :tenant_id, :audit_ids
```
josemigallas, line 29: There must be a reason why `service_id` is protected, you can't just unprotect it.

app/models/metric.rb
```diff
@@ -26,7 +26,9 @@ class Metric < ApplicationRecord
...
+  attr_protected :parent_id, :tenant_id, :audit_ids
+  attr_accessible :service_id, :friendly_name, :system_name, :unit, :description
```
josemigallas, line 30: > To start from an all-closed default and enable attributes as needed, have a look at [attr_accessible](https://apidock.com/rails/ActiveModel/MassAssignmentSecurity/ClassMethods/attr_accessible).

Meaning you either define protected OR accessible attributes but not both.

app/models/metric.rb
nidhi-soni1104: @josemigallas yes agreed and we keep id's as protected attributes only but warning also raise for the same.

josemigallas:
If the application is trying to assign a protected attribute, IMO it is not about allowing that attribute but to remove it from the assignment. `service_id` is protected for a reason.
Besides, tests are 🔴, maybe caused by this changes.
Also, always back these kind of changes up with some sort of test.

app/models/metric.rb
nidhi-soni1104: yes, we can keep one  and that's the reason I removed service_id from protected because this attribute is used while doing mass assignment.

here the attributes  for **attr_protected** and **attr_accessible** are different. 

I think you are saying not to use both together on same model?

nidhi-soni1104:
> If the application is trying to assign a protected attribute, IMO it is not about allowing that attribute but to remove it from the assignment. `service_id` is protected for a reason. Besides, tests are red_circle, maybe caused by this changes. Also, always back these kind of changes up with some sort of test.


> This is the reason I got confusion here it they are assigning service_id differently but still raise the warning

```
def self.create_default!(type, attributes = {})
    metric = new(attributes.merge(:friendly_name => 'Hits', :system_name => 'hits', :unit => 'hit',
                                  :description => 'Number of API hits'))
    metric.service_id = attributes[:service_id]
    metric.save!
    metric
  end
```

nidhi-soni1104:
> > If the application is trying to assign a protected attribute, IMO it is not about allowing that attribute but to remove it from the assignment. `service_id` is protected for a reason. Besides, tests are red_circle, maybe caused by this changes. Also, always back these kind of changes up with some sort of test.
> 
> > This is the reason I got confusion here it they are assigning service_id differently but still raise the warning because we save all together at the end
> 
> ```
> def self.create_default!(type, attributes = {})
>     metric = new(attributes.merge(:friendly_name => 'Hits', :system_name => 'hits', :unit => 'hit',
>                                   :description => 'Number of API hits'))
>     metric.service_id = attributes[:service_id]
>     metric.save!
>     metric
>   end
> ```

nidhi-soni1104:
Resolving all conversations and checking out for better solution .

nidhi-soni1104:
@josemigallas @akostadinov 
Can I please have some suggestion to fix this.
All tests are failing and MassAssignmentSecurity is not working fine here

source from where I got this is
https://apidock.com/rails/ActiveModel/MassAssignmentSecurity/ClassMethods/attr_protected

I cannot see warning now but test keeps failing

app/models/metric.rb
```diff
@@ -51,14 +53,21 @@ class Metric < ApplicationRecord
...
+    end
+  end
+
   # Create one of the predefined, default metrics.
   #
   # == Arguments
   #
   # +type+:: Which default metric to create. Currently only :hits are supported.
   def self.create_default!(type, attributes = {})
-    metric = new(attributes.merge(:friendly_name => 'Hits', :system_name => 'hits', :unit => 'hit',
```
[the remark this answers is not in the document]
nidhi-soni1104: sDid you mean removing from the attribute which are received as params in attributes = {} hash ?

app/models/metric.rb
```diff
@@ -57,9 +58,11 @@ class Metric < ApplicationRecord
   #
   # +type+:: Which default metric to create. Currently only :hits are supported.
   def self.create_default!(type, attributes = {})
+    id = attributes[:service_id]
+    attributes.delete(:service_id)
```
josemigallas, line 62: We're in `metric` so `id` would normally mean the Metric's ID. Better be explicit when declaring variables.
```suggestion
    service_id = attributes.delete(:service_id)
```

app/models/metric.rb
```diff
@@ -57,9 +58,11 @@ class Metric < ApplicationRecord
...
     metric = new(attributes.merge(:friendly_name => 'Hits', :system_name => 'hits', :unit => 'hit',
                                   :description => 'Number of API hits'))
-    metric.service_id = attributes[:service_id]
+    metric.service_id = id
```
josemigallas, line 65:
```suggestion
    metric.service_id = service_id
```

josemigallas (changes requested):

josemigallas (approved):

PR #2962, opened 2022-04-11
Final touches for Quickstarts

josemigallas:

Once this PR is merged, we should be able to deploy to master and ship this new feature.

Changes:
- Removes feature config
- Prevent Quickstarts from rendering when not necessary
- Cleanup / Refactor

[THREESCALE-8283: Integrate Quick Starts into 3scale](https://issues.redhat.com/browse/THREESCALE-8283)

### DIFF
--- a/app/controllers/frontend_controller.rb
+++ b/app/controllers/frontend_controller.rb
@@ -37,6 +37,8 @@ class FrontendController < ApplicationController
 
   layout :pick_buyer_or_provider_layout
 
+  helper_method :quickstarts_presenter
+
   private
 
   def disable_x_content_type
@@ -167,5 +169,9 @@ class FrontendController < ApplicationController
 
   def set_display_currency
     ThreeScale::MoneyHelper.display_currency = current_account.currency if current_account
+  end
+
+  def quickstarts_presenter
+    @quickstarts_presenter ||= QuickstartsPresenter.new(@backend_api || @service || current_account.default_service)
   end
 end
--- a/app/javascript/packs/quickstarts/container.js
+++ b/app/javascript/packs/quickstarts/container.js
@@ -17,4 +17,12 @@ document.addEventListener('DOMContentLoaded', () => {
     links: safeFromJsonString<Array<[string, string, string]>>(links) || [],
     renderCatalog: safeFromJsonString<boolean>(renderCatalog)
   }, containerId)
+
+  // HACK of the year: We need QuickStartContainer to wrap the whole #wrapper for the Drawer to work properly.
+  const wrapperContainer = document.getElementById('wrapper')
+  const quickStartsContainer = document.querySelector('.pfext-quick-start-drawer__body')
+
+  if (wrapperContainer && quickStartsContainer) {
+    quickStartsContainer.after(wrapperContainer)
+  }
 })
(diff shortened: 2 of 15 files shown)


app/javascript/src/QuickStarts/utils/progressTracker.js
```diff
@@ -0,0 +1,13 @@
+// @flow
+
+/**
+ * This is a collection of wrappers around localStorage to ease getting information about Quickstarts state
+ */
+
+function getActiveQuickstart (): null | string {
```
jkeam, line 7: I think you can do something like:

```
function getActiveQuickstart(): null | string {
  const quickstartId = localStorage.getItem('quickstartId')
  return quickstartId || null
}
```

From MDN (https://developer.mozilla.org/en-US/docs/Web/API/Storage/getItem), it returns: `A string containing the value of the key. If the key does not exist, null is returned.`

I don't think there's a need to `JSON.parse` as it looks like it's an ID you are saving and then retrieving.

And no need to check for length as the `getItem` call will already return either the string or null if the key does not exist.  The reason for the short circuit logic in the `return` is to catch the edge case of `quickstartId` being an empty string.

app/javascript/packs/quickstarts/container.js
```diff
@@ -1,9 +1,14 @@
 // @flow
 
 import { QuickStartContainerWrapper as QuickStartContainer } from 'QuickStarts/components/QuickStartContainer'
+import { getActiveQuickstart } from 'QuickStarts/utils/progressTracker'
 import { safeFromJsonString } from 'utilities/json-utils'
 
 document.addEventListener('DOMContentLoaded', () => {
+  if (getActiveQuickstart == null) {
```
jkeam, line 8: I might be missing something, but can you invoke the method without parenthesis?

Shouldn't it be:

```javascript
if (getActiveQuickstart() == null) {
  return 
}
```

Or more concisely I'd probably write it like this:

```javascript
if (!getActiveQuickstart()) {
  return
}
```

app/controllers/frontend_controller.rb
```diff
@@ -168,4 +170,15 @@ def render_wrong_domain_error
...
+  def quickstarts_presenter
+    current_api = if @backend_api&.id.present?
+                    @backend_api
+                  elsif @service&.id.present?
+                    @service
+                  else
+                    current_account.default_service
+                  end
+    @quickstarts_presenter ||= QuickstartsPresenter.new(current_api)
+  end
```
jkeam, line 183: Stylistic comment, thoughts on short circuiting if `quickstarts_presenter` exists? 
 This then avoids the logic around `current_api` unnecessarily on multiple invocations of this.

Also not sure if you can drop the `present?` call and simply check for the existence of the `id`?

```ruby
def quickstarts_presenter
  return @quickstarts_presenter if @quickstarts_presenter

  current_api = if @backend_api&.id
                  @backend_api
                elsif @service&.id
                  @service
                else
                  current_account.default_service
                end
  @quickstarts_presenter = QuickstartsPresenter.new(current_api)
end
```

josemigallas: I like to use `present` for more readability but I think it's redundant in this case.

PR #2965, opened 2022-04-13
THREESCALE-8371 fixed warning of protected attribute

nidhi-soni1104:

WARNING: Can't mass-assign protected attributes for CMS::Builtin::Section: parent_id

Which issue(s) this PR fixes

https://issues.redhat.com/browse/THREESCALE-8371

### DIFF
--- a/app/lib/logic/cms.rb
+++ b/app/lib/logic/cms.rb
@@ -60,7 +60,7 @@ module Logic
       end
 
       def create_cms_assets
-        self.builtin_sections.create!(title: 'Root', system_name: 'root', parent_id: nil)
+        self.builtin_sections.create!(title: 'Root', system_name: 'root')
         # TODO: add SimpleLayout logic here
       end
     end


app/lib/logic/cms.rb
```diff
@@ -60,7 +60,7 @@ def cms_toolbar_intro_visible?
       end
 
       def create_cms_assets
-        self.builtin_sections.create!(title: 'Root', system_name: 'root', parent_id: nil)
```
josemigallas, line 63: Where are protected attributes defined? Why was `parent_id` added in the first place if it was protected? 🤔 It's a bit strange that this attribute is not set afterwards..

nidhi-soni1104: Protected attributes are defined no where but this module is used in a model and there also nothing describes, One strange thing is parent_id: nil is passed so what is the sense of passing any attribute nil so I removed it.

nidhi-soni1104: @josemigallas can we merge this?

nidhi-soni1104: not  tested resetting DB let me test this way

PR #2966, opened 2022-04-14
avoid sphinx indexing tests issues

akostadinov:
In sphinx tests we search all classes for being indexed. If this test ran first, then there was an extra unexpected class `MetricWithFiber`. So now removing this class in `teardown`.

### DIFF
--- a/test/unit/backend/model_extensions/metric_test.rb
+++ b/test/unit/backend/model_extensions/metric_test.rb
@@ -56,21 +56,17 @@ class Backend::ModelExtensions::MetricTest < ActiveSupport::TestCase
   end
 
   test 'sync backend metric data multiple times under race condition' do
-    class MetricWithFiber < ::Metric
-      def destroy
-        super
-        Fiber.yield
-      end
-    end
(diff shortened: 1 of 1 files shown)


jkeam:
I think for `MetricWithFiber`, since it's just a throwaway class, we should instead create an anonymous class.  I created a gist with an example of what I mean:

https://gist.github.com/jkeam/ddc5fca16816be645b2c73b88c29b5ee

akostadinov:
@jkeam , thank you. Actually both ways do not work which can be observed by running

```
bundle exec rails test:run TEST="test/unit/backend/model_extensions/metric_test.rb test/unit/indices/indexing_test.rb" TESTOPTS="--seed 13236 --verbose"
```

I've got confused that my change had effect. But I updated PR to patch the objects. Let me know if you know a nicer way.

josemigallas (approved):

jkeam:
Sounds good!  I am currently out of town until Thursday so haven’t had much
time to look at this.

PR #2967, opened 2022-04-22
metrics delete race simplification

akostadinov:
The test is doing nothing because essentially we don't protect against a race so callback executes twice. We do not need it to execute twice so there is no point to test that. If it executed only once, that would be better but then this is not the case.

### DIFF
--- a/app/lib/backend/model_extensions/metric.rb
+++ b/app/lib/backend/model_extensions/metric.rb
@@ -11,6 +11,7 @@ module Backend
           # - 2 threads are trying to destroy a metric – T1 and T2
           # - T1 succeeds and T2 gets `persisted? => false` inside `ActiveRecord::Persistence#destroy` due to https://apidock.com/rails/v4.2.7/ActiveRecord/Core/sync_with_transaction_state, thus making `@_trigger_destroy_callback = true`
           # - Because of `m.destroyed? == true`, both T1 and T2 will invoke the callback
+          # TODO: this condition seems to do nothing, maybe delete it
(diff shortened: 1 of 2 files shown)


app/lib/backend/model_extensions/metric.rb
```diff
@@ -11,6 +11,7 @@ def self.included(base)
           # - 2 threads are trying to destroy a metric – T1 and T2
           # - T1 succeeds and T2 gets `persisted? => false` inside `ActiveRecord::Persistence#destroy` due to https://apidock.com/rails/v4.2.7/ActiveRecord/Core/sync_with_transaction_state, thus making `@_trigger_destroy_callback = true`
           # - Because of `m.destroyed? == true`, both T1 and T2 will invoke the callback
+          # TODO: this condition seems to do nothing, maybe delete it
           after_commit :sync_backend, if: ->(m) { (m.persisted? || m.destroyed?) && !m.changed? }
```
akostadinov, line 15: @guicassolato , hi, do you think we can just remove this condition? Do you see any way that it can help?

jkeam:
I think this looks good.  As we discussed, the work that's being scheduled to be done later by the `BackendMetricWorker` looks to be idempotent, it's either a save or delete.  In this case, the deletion of an already existing object shouldn't be a problem.

jkeam (approved):

app/lib/backend/model_extensions/metric.rb
jkeam: Hi @guicassolato!  Was wondering if you saw this comment, thanks!

PR #2968, opened 2022-04-22
more deprecation fixes

akostadinov:


jkeam:
While I understand the purpose of this PR, I'm not a fan of the entire `symbolize.rb` class in general.  I hate changing the way Rails works at such a fundamental level and forcing it to return symbols instead of the actual datatype that Active Record is mapping it to.  I see a few cons with this class:

1. You hide the actual datatype and force everything to be a symbol
2. You make upgrading Rails harder as now you have to overload all of these attributes accessor methods to turn them into symbols to keep things consistent, (for example: `saved_changes`, `attribute_in_database`, `changes_to_save`, `attributes_in_database`)

I see the original commit for this class (https://github.com/3scale/porta/commit/688adc03fa872e7976dbecf505cd6442dc6a3aba) was to replace the Symbolize gem.  I think the proper fix at the time, as well as now, is to not keep adding more attributes accessor methods to keep up with all the new methods added by Rails, but instead to kill this class entirely.

akostadinov:
I agree, just didn't want to start this process. We produced dev3 build without this addition. We may as well just leave it be for awhile and see if everything works. Then we can remove the `symbolize` keyword from all models that presently use **one by one** it to avoid major disruption and be easier to track if something fails. Then once no `symbolize` method is used anywhere, we can remove this class. WDYT?

akostadinov:
btw if anybody wanted an attribute to be a symbol, they could just create a custom ActiveRecord type. Would be much more straightforward.

josemigallas (approved):

josemigallas:
Tentatively approved, I'll let you guys ultimately decide

jkeam:
I was thinking we don't need this pull request at all if there are no errors?

akostadinov:
Agreed. Thanks for the input.

PR #2969, opened 2022-04-25
THREESCALE-8330 Updated queries to improve performance

nidhi-soni1104:

Improve performance when deleting an application

https://issues.redhat.com/browse/THREESCALE-8330?page=com.docminer.jira.issue-links%3Acom.kintosoft.jira.links.tab-panel

### DIFF
--- a/app/lib/applications_controller_methods.rb
+++ b/app/lib/applications_controller_methods.rb
@@ -52,8 +52,7 @@ module ApplicationsControllerMethods
   end
 
   def find_cinstance
-    @cinstance = accessible_not_bought_cinstances.includes(plan: %i[service original plan_metrics pricing_rules])
-                                                 .find(params[:id])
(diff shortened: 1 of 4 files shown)


app/lib/applications_controller_methods.rb
```diff
@@ -52,8 +52,7 @@ def find_applications
   end
 
   def find_cinstance
-    @cinstance = accessible_not_bought_cinstances.includes(plan: %i[service original plan_metrics pricing_rules])
-                                                 .find(params[:id])
+    @cinstance = accessible_not_bought_cinstances.find(params[:id])
```
jkeam, line 55: Assuming you never reference these associations later (service, original, plan_metrics, pricing_rules)?  Bc if you do, the includes will save you additional queries to get those objects.

PR #2971, opened 2022-04-26
[THREESCALE-5725] FIPS compatibility

akostadinov:
Ability to deploy on a FIPS enabled infrastructure.

- [x] <del>check if [rotating cookie digest](https://guides.rubyonrails.org/v5.2/security.html#rotating-encrypted-and-signed-cookies-configurations) will work in 5.1</del>
- [x] check if markdown in dev portal will work in FIPS mode, `redcarpet` docs state that perl + MD5 are required for it (see redcarpet-3.5.1/test/fixtures/benchmark.md) /create a new page in dev portal and select markdown handler/
- [x] check for rails 7 string truncations with the SHA256 update for all changes from original rails/rails#31289
- [x] investigate/implement paperclip rehashing (investigate whether we use S3 too and whether rehash will be needed there as well, and see next point too)
- [x] `aws-sdk` usage may be unsupported with FIPS, see aws/aws-sdk-ruby#2645, maybe that is for paperclip attachments
- [x] validate active-job-uniqueness is working (check Sidekick for similar concerns)
- [x] <del>check whether we support HTTP digest auth and whether change can affect it on non-fips systems (apparently it will on FIPS), see `rack-2.1.4/lib/rack/auth/digest/md5.rb` and `nonce.rb`</del>
- [x] update `oauth2` to latest where MD5 is not used (actually monkey patched latest stable)
- [x] update [dalli](https://github.com/petergoldstein/dalli) to some version with petergoldstein/dalli#724 and investigate migration requirements
- [x] `compass-core` uses MD5, investigate whether this is a runtime problem or only during build, there is a failed PR to get rid of MD5, see Compass/compass#1266 (also some MD5 references in `compass/sass_extensions/sprites`)
- [x] multiple `activemerchant` billing/gateways depend on MD5, check that we don't care about them

**PS** See output of  `rg -ipuuu md5 $(bundle show --paths)` here: [md5_references.txt](https://github.com/3scale/porta/files/8585691/md5_references.txt). For color output open by `less -R md5_references.txt`.
**PPS** Compiling ruby on a FIPS machine can result in ruby that uses MD5 internally, not through OpenSSL, so that may workaround the issue provided we consider existing usages as acceptable. I just read a comment somewhere, so this has to be verified.


app/models/cms/file.rb
```diff
@@ -16,7 +16,8 @@ class CMS::File < ApplicationRecord
   has_attached_file :attachment,
                     url: Rails.application.config.cms_files_path,
                     s3_url_options: ->(file) { file.s3_url_options },
-                    validate_media_type: false
+                    validate_media_type: false,
+                    adapter_options: { hash_digest: Digest::SHA256 }
```
akostadinov, line 20: For this we need to rehash all attachments which might be an issue for updating existing deployments. See [paperclip docs](https://www.rubydoc.info/gems/paperclip/6.1.0#checksum-fingerprint).

app/models/profile.rb
```diff
@@ -44,10 +44,15 @@ class Profile < ApplicationRecord
...
-    :default_url => '/assets/3scale-logo.png'.freeze
+    styles: {
+      large: '300x300>'.freeze,
+      medium: '150x150>'.freeze,
+      thumb: '100x100>'.freeze,
+      invoice: ['200x50>'.freeze, :png].freeze }.freeze,
+    url: ':url_root/:account_id/:class/:attachment/:style/:basename.:extension'.freeze,
+    s3_permissions: 'public-read'.freeze,
+    default_url: '/assets/3scale-logo.png'.freeze,
+    adapter_options: { hash_digest: Digest::SHA256 }
```
akostadinov, line 55: For this we need to rehash all attachments which might be an issue for updating existing deployments. See [paperclip docs](https://www.rubydoc.info/gems/paperclip/6.1.0#checksum-fingerprint).

app/models/invoice.rb
```diff
@@ -39,7 +39,7 @@ class InvalidInvoiceStateException < RuntimeError; end
   has_many :payment_transactions, -> { oldest_first }, dependent: :nullify, inverse_of: :invoice
   has_many :payment_intents, dependent: :destroy, inverse_of: :invoice
 
-  has_attached_file :pdf, url: ':url_root/:class/:id/:attachment/:style/:basename.:extension'
+  has_attached_file :pdf, url: ':url_root/:class/:id/:attachment/:style/:basename.:extension', adapter_options: { hash_digest: Digest::SHA256 }
```
akostadinov, line 42: For this we need to rehash all attachments which might be an issue for updating existing deployments. See [paperclip docs](https://www.rubydoc.info/gems/paperclip/6.1.0#checksum-fingerprint).

config/initializers/active_job_uniqueness.rb
```diff
@@ -23,7 +23,7 @@
 
       # Digest method for lock keys generating. Expected to have `hexdigest` class method.
       #
-      # config.digest_method = OpenSSL::Digest::MD5
+      config.digest_method = Digest::SHA256
```
akostadinov, line 26: We need to test active-job-uniqueness working especially make sure Redis max length not reached.

app/models/cms/file.rb
jkeam: Ya agree there.  Looks like it can be fixed with something like:

```ruby
CLASS='CMS::File' ATTACHMENT=attachment rake paperclip:refresh:fingerprints
```

so that might be something we need to add to the docs.

app/models/invoice.rb
jkeam: Yep, same as above:

```ruby
CLASS=Invoice ATTACHMENT=pdf rake paperclip:refresh:fingerprints
```

app/models/profile.rb
jkeam:
```ruby
CLASS=Profile ATTACHMENT=logo rake paperclip:refresh:fingerprints
```

config/initializers/md5_replace.rb
```diff
@@ -0,0 +1,6 @@
+# TODO: configure ActiveSupport::Digest after Rails 5.2 update
+
+raise "setup ActiveSupport::Digest on Rails 5.2+" if Gem::Version.new(Rails.version) >= Gem::Version.new('5.2')
+
+::Digest.send :remove_const, :MD5
+::Digest::MD5 = ::Digest::SHA256
```
jkeam, line 6: Monkey patches like this make me cringe for the simple fact of confusion.  Now the `MD5` class will actually perform a `SHA256`.  Since this is all throwaway code though, I think it's ok, but instead of a `TODO` I'd use a `FIXME`.  `TODO` tells me an optimization or feature might be worth implementing in the future where a `FIXME` tells me this is definitely not the right way, but we are ok with it for now.

akostadinov:
For the record, we can probably use cookie rotation to avoid invalidation but not sure it is available in 5.1. To be investigated:
https://guides.rubyonrails.org/v5.2/security.html#rotating-encrypted-and-signed-cookies-configurations

config/initializers/md5_replace.rb
akostadinov: We will not do this. Too risky of bad confusion provided so many MD5 references in other gems.

config/application.rb
```diff
@@ -44,8 +44,12 @@ class Application < Rails::Application
     config.action_controller.forgery_protection_origin_check = false
     # Make `form_with` generate non-remote forms. Defaults true in Rails 5.1 to 6.0
     config.action_view.form_with_generates_remote_forms = false
+
     # Make Ruby preserve the timezone of the receiver when calling `to_time`.
     config.active_support.to_time_preserves_timezone = false
+    # Use a modern approved hashing function
+    # config.active_support.hash_digest_class = OpenSSL::Digest::SHA256
+    ActiveSupport::Digest.hash_digest_class = OpenSSL::Digest::SHA256 # option above should work with Rails 6.x
```
jkeam, line 52: What are your thoughts around `ActiveSupport::Digest.hexdigest` (https://github.com/rails/rails/blob/5-2-stable/activesupport/lib/active_support/digest.rb).  This is the method:

```ruby
def hexdigest(arg)
  hash_digest_class.hexdigest(arg)[0...32]
end
```

`OpenSSL::Digest::MD5` generates a string of length 32, but `OpenSSL::Digest::SHA256` generates a string of length 64.  So truncating it has no impact on the effectiveness of MD5, but might for SHA256.  I'm not sure we have a choice in the matter since MD5 is a non-starter, but wanted to still drop a note here.  Maybe we can look at Rails 6.x and see what other changes went into ensuring that SHA256 worked.

config/initializers/dalli.rb
```diff
@@ -0,0 +1,23 @@
+# Patches dalli to use SHA256 checksum instead of MD5.
+# Once we stop supporting Ruby 2.4, we should upgrade dalli
+# and see if any configuration is needed for same effect.
+#
+# key digests were actually fixed with
+# https://github.com/petergoldstein/dalli/commit/74b2625f11ff56dd6f55c7316bc115bc5a29be5f
+
+Dalli::Client.prepend(Module.new do
```
jkeam, line 8: This is some pretty fancy key patching going on here.  It's easy enough to follow, but thoughts on adding a test?  I know it's temporary throw away code, but there's a saying `the most permanent code is temporary code` so might be worth adding a test?

Some examples of how to do this:
https://stackoverflow.com/questions/25147236/how-do-you-test-your-config-initializer-scripts-with-rspec-in-rails

Basically the idea is to create an extra class or similar containing a static method and then test that static method in a test.

config/initializers/oauth2.rb
```diff
@@ -0,0 +1,17 @@
+OAuth2::MACToken.prepend(Module.new do
+  # this is a copy of original method but without using MD5 to
+  # enable FIPS environment compatibility.
+  # see https://github.com/oauth-xx/oauth2/pull/587
+  def header(verb, url)
```
jkeam, line 5: I'm not as worried about a test for this method since you noted the exact place you pulled this exact code from and that codebase is fully tested.

jkeam (changes requested):
Just a few comments.  Not all of them have direct action; some are just notes and/or questions.

config/application.rb
akostadinov: The change already went into 5.2 and I couldn't spot any complaints about hash length. I also went over same pieces of code in Rails 7 for the purpose you describe. One issue was the memcache store that we switched to and is mostly broken regardless of hash size. I have rails/rails#45245 for that but still it works well enough for our use cases.

config/initializers/oauth2.rb
akostadinov: Yes, also probably we don't use mac token as it is a very old draft standard. At least I hope not used. See linked pull request.

app/models/invoice.rb
akostadinov: We are not fingerprinting.

app/models/profile.rb
akostadinov: we are not fingerprinting in fact

josemigallas:
@akostadinov about the compass TODO, if it compiles and images are built without errores I wouldn't worry because as it is said in the issue https://github.com/Compass/compass/pull/1266 it's not used for authentication, only for css assets.

config/initializers/dalli.rb
akostadinov: From what I later read, this is an outdated draft standard so most likely we don't use it. And is removes in upcoming gem versions. Perhaps it was useless to monkey patch it. I don't find it worth spending more time on it.

jkeam (dismissed):
looks good to me

PR #2972, opened 2022-04-28
Fix Sphinx version for community builds for OpenShift

guicassolato:
The Dockerfile to build for OpenShift based on CentOS7 is broken since https://github.com/3scale/porta/commit/ccc114103bad6bfff296e409c3cbe0a78ee806ba upgraded the reference of the Sphinx version to '3.4.1' without providing a corresponding RPM package to replace the existing one based on Sphinx v2.2.11-1.

This PR fixes the Sphinx version now including the necessary package to build with Sphinx 3.4.1.


openshift/system/Dockerfile
```diff
@@ -27,11 +23,16 @@ RUN yum-config-manager --save --setopt=cbs.centos.org_repos_sclo7-rh-ruby26-rh-c
...
               rh-nodejs12 \
-    && rpm -i /tmp/sphinxsearch.rpm \
-    && rm /tmp/*.rpm \
     && yum install -y epel-release \
     && yum -y clean all
 
+# Install sphinx search
+ARG SPHINX_VERSION=3.4.1
+COPY openshift/system/sphinx-${SPHINX_VERSION}-efbcc65-linux-amd64.tar.gz /tmp/sphinxsearch.tar.gz
+RUN tar zxf /tmp/sphinxsearch.tar.gz -C /tmp \
```
guicassolato, line 32: Worth replacing this with a curl/wget command that downloads the package at every build, to avoid hosting the 20MB file in the code base? Granted it won't ever be yanked from source of course.

openshift/system/Dockerfile
```diff
@@ -69,6 +70,7 @@ RUN export ${BUNDLER_ENV} >/dev/null \
     && umask 0002 \
     && cd /opt/system \
     && npm install -g yarn \
+    && yarn config set network-timeout 300000 \
```
josemigallas, line 73: Why this timeout?

openshift/system/Dockerfile
```diff
@@ -27,11 +23,16 @@ RUN yum-config-manager --save --setopt=cbs.centos.org_repos_sclo7-rh-ruby26-rh-c
...
-    && rpm -i /tmp/sphinxsearch.rpm \
-    && rm /tmp/*.rpm \
     && yum install -y epel-release \
     && yum -y clean all
 
+# Install sphinx search
+ARG SPHINX_VERSION=3.4.1
+COPY openshift/system/sphinx-${SPHINX_VERSION}-efbcc65-linux-amd64.tar.gz /tmp/sphinxsearch.tar.gz
+RUN tar zxf /tmp/sphinxsearch.tar.gz -C /tmp \
+    && cp /tmp/sphinx-${SPHINX_VERSION}/bin/* /usr/bin/ \
```
josemigallas, line 33: Is `tar + cp` a manual installation similar to what `rpm` does? Any particular reason why we're not using an rpm anymore?

akostadinov: I don't see RPM download options available anymore.

guicassolato: > Is `tar + cp` a manual installation

It is.

> reason why not using an rpm

Because there's no official RPM available for Sphinx Search [v3](sphinxsearch.com/downloads/current) as we used to have for [v2](http://sphinxsearch.com/downloads/archive).

openshift/system/Dockerfile
guicassolato: It looks like it was an intermittent issue. Removed.

openshift/system/Dockerfile
guicassolato: Done.

akostadinov (approved):
Thanks a lot!

PR #2976, opened 2022-04-29
use #saved_changes in place of #previous_changes

akostadinov:
In these case both methods are equivalent and in Rails 5.2+ they will be equivalent in all conditions. But updating to `#saved_changes` to prevent deprecation warnings as well because this is method with more clear intentions and behavior. At least I don't have to think how to write `previous` properly anymore.

Namely in Rails 5.1, `#saved_changes` and `#previous_changes` differ in behavior only in `after_save` callback where `#previous_changes` returns nothing. In 5.2 `#previous_changes` will behave exactly as `#saved_changes`.

### DIFF
--- a/app/models/access_token.rb
+++ b/app/models/access_token.rb
@@ -165,7 +165,7 @@ class AccessToken < ApplicationRecord
   end
 
   def show_value?(*)
-    previous_changes.include?(:value)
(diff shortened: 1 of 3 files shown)


jkeam (approved):
Looks good to me!

PR #2979, opened 2022-05-04
build community image

akostadinov:
after #2972 lets also build the image!


akostadinov:
Need to merge so I can test it but please provide input!

PR #2980, opened 2022-05-05
Container image

akostadinov:
test and optimize image build, follow up on #2979 

Please comment on the whole file, I merged the last one just to be able to try out.

### DIFF
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -3,7 +3,12 @@ on:
   # push:
   #   branches:
   #   - master
-  workflow_dispatch: {}
+  workflow_dispatch:
+    inputs:
+      platforms:
+        description: Platforms to build for
+        required: false
+        default: 'linux/amd64'
 
   # pull_request:
   # branches: [ master ]
@@ -18,7 +23,7 @@ jobs:
   docker:
     runs-on: ubuntu-latest
     steps:
-      # Allow multi-target builds, uncomment this to use `platfotms` below
+      # Allow multi-target builds
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v1
       - name: Set up Docker Buildx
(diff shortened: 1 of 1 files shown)


guicassolato (commented):
Not using quay.io?

akostadinov:
@guicassolato , I don't know where it is appropriate to put it there. Do you have suggestions?

akostadinov:
@guicassolato , thank you. One thing that seems not possible to do is multi-arch build with quay. While with GHA we can do it. Not sure if anybody needs though.

.github/workflows/container-image.yml
```diff
@@ -18,7 +18,7 @@ jobs:
   docker:
     runs-on: ubuntu-latest
     steps:
-      # Allow multi-target builds
+      # Allow multi-target builds, uncomment this to use `platfotms` below
```
jkeam, line 21: When you say "uncomment this to use `platforms` below", what is "this" referring to?

.github/workflows/container-image.yml
```diff
@@ -1,8 +1,8 @@
 name: Build porta container image
 on:
-  push:
-    branches:
-      - master
+  # push:
+  #   branches:
+  #   - master
```
jkeam, line 5: Why is this commented out?

.github/workflows/container-image.yml
akostadinov: There is an element `platforms` further down the file.

.github/workflows/container-image.yml
akostadinov: Because I'm not yet sure we want to have it build for every push to master. Or just have ability to trigger manually for whatever branch we want only when we need.

akostadinov:
Merging as I need it for testing purposes. We can tune how it works later.

PR #2983, opened 2022-05-09
THREESCALE-8009 prevent outbox emails from being sent empty

nidhi-soni1104:
**hat this PR does / why we need it:**

make those field required

In admin portal - /p/admin/messages/outbox/new - page under Audience > Messages > Inbox/Sent messages

https://issues.redhat.com/browse/THREESCALE-8009

https://issues.redhat.com/browse/THREESCALE-8392

### DIFF
--- a/app/views/provider/admin/messages/outbox/new.html.erb
+++ b/app/views/provider/admin/messages/outbox/new.html.erb
@@ -20,7 +20,8 @@
     </li>
 
     <%= form.input :subject, :as => :string, :label => 'Subject:', input_html: { required: true } %>
-    <%= form.input :body, :input_html => {:rows => 10}, :label => 'Body:', input_html: { required: true } %>
+    <%= form.input :body, :input_html => {:rows => 10, required: true}, :label => 'Body:' %>
+
   <% end %>
   <%= form.buttons do %>
     <% if request.xhr? %>
--- a/app/views/provider/admin/messages/outbox/new.html.erb
+++ b/app/views/provider/admin/messages/outbox/new.html.erb
@@ -3,30 +3,30 @@
   <h2>Compose</h2>
 <% end %>
 
-<%= semantic_form_for @message, :url => provider_admin_messages_outbox_index_path,
-  :html => {:class => if request.xhr? then 'remote' end, :id => "message-form"} do |form| %>
+<%= semantic_form_for @message, url: provider_admin_messages_outbox_index_path,
+  html: { class: if request.xhr? then 'remote' end, id: "message-form" } do |form| %>
 
   <%= form.inputs do %>
     <li>
       <label>To:</label>
       <% if current_account.provider? && params[:to].nil? && @message.to.size == current_account.buyer_accounts.size %>
-        <%= text_field_tag '', "All Users", :readonly => true %>
+        <%= text_field_tag '', "All Users", readonly: true %>
         <p class="inline-hints">
           If you want to send a message to a particular user, go to the <%= link_to 'accounts page', admin_buyers_accounts_path %>, click on the user and then click on the send message icon.
         </p>
       <% else %>
-        <%= text_field_tag 'to', @message.to.map(&:org_name).join(', '), :readonly => true  %>
+        <%= text_field_tag 'to', @message.to.map(&:org_name).join(', '), readonly: true %>
       <% end %>
     </li>
 
-    <%= form.input :subject, :as => :string, :label => 'Subject:', input_html: { required: true } %>
-    <%= form.input :body, :input_html => {:rows => 10, required: true}, :label => 'Body:' %>
+    <%= form.input :subject, as: :string, label: 'Subject:', input_html: { required: true } %>
+    <%= form.input :body, input_html: { rows: 10, required: true }, label: 'Body:' %>
 
   <% end %>
   <%= form.buttons do %>
     <% if request.xhr? %>
       <li class="link">
-        <%= link_to 'Cancel', provider_admin_messages_root_path, :class => 'fancybox-close' %>
+        <%= link_to 'Cancel', provider_admin_messages_root_path, class: 'fancybox-close' %>
       </li>
     <% end %>
     <%= form.button 'Send', button_html: { class: 'pf-c-button pf-m-primary' } %>
--- a/app/views/provider/admin/messages/outbox/new.html.erb
+++ b/app/views/provider/admin/messages/outbox/new.html.erb
@@ -19,8 +19,8 @@
       <% end %>
     </li>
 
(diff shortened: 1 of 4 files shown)


josemigallas (changes requested):
Is the test in the right file? It looks to me this file is for applications' bulk action, whereas the fix is for messages/outbox.

On the other hand, I think the steps are mixed up: first *arrange* conditions (subject/body filled/empty) then *act* on the component (perform test, click send) and finally *assert* by verifying no email was sent.

nidhi-soni1104:
> Is the test in the right file? It looks to me this file is for applications' bulk action, whereas the fix is for messages/outbox.
> 
> On the other hand, I think the steps are mixed up: first _arrange_ conditions (subject/body filled/empty) then _act_ on the component (perform test, click send) and finally _assert_ by verifying no email was sent.

this is also related to bulk message so I added test in same file.
Okay let me update the test conditions accordingly

nidhi-soni1104:
@josemigallas Sequence is updated.

app/views/provider/admin/messages/outbox/new.html.erb
```diff
@@ -19,8 +19,9 @@
       <% end %>
     </li>
 
-    <%= form.input :subject, :as => :string, :label => 'Subject:' %>
-    <%= form.input :body, :input_html => {:rows => 10}, :label => 'Body:' %>
+    <%= form.input :subject, :as => :string, :label => 'Subject:', input_html: { required: true } %>
+    <%= form.input :body, :input_html => {:rows => 10, required: true}, :label => 'Body:' %>
+
```
jkeam, line 24: I see this also being done this way in app/views/buyers/accounts/bulk/send_emails/new.html.erb, app/views/buyers/applications/bulk/send_emails/new.html.erb, app/views/buyers/service_contracts/bulk/send_emails/new.html.erb, but if you look at app/views/buyers/users/edit.html.erb line 13, you will see you can do `required: true` like this:

```
<%= form.input :password, required: true, input_html: { autocomplete: 'off' } %>
```

This also looks like the preferred way as per the docs also:
https://github.com/formtastic/formtastic

nidhi-soni1104: yes both are correct ways , but if input_html is already used there so we can keep required: true inside it also.
I have seen it in my earlier experiences that when we use it inside it input_html it works for all browsers.

nidhi-soni1104: @jkeam @josemigallas if all looks good to you guys now can we merge this?

josemigallas:
> this is also related to bulk message so I added test in same file

AFAIK this has nothing to do with bulk actions. This is related to `Outbox::Messages` and it has its own controller, routes and views. The tests belong to `features/admin/messages/outbox.feature`.

features/provider/admin/messages/outbox.feature
```diff
@@ -0,0 +1,27 @@
...
+    Given a default application plan "Basic" of provider "foo.3scale.localhost"
+      And provider "foo.3scale.localhost" has multiple applications enabled
+
+    Given a following buyers with applications exists:
+      | name | provider        | applications        |
+      | bob  | foo.3scale.localhost | BobApp              |
+      | jane | foo.3scale.localhost | JaneApp, JaneAppTwo |
+
+    Given current domain is the admin domain of provider "foo.3scale.localhost"
+      And I don't care about application keys
```
josemigallas, line 18: Applications is a different scope and I don't thing it is related to outbox messages. All these steps can probably go away.
```suggestion
    Given a provider is logged in
```

features/provider/admin/messages/outbox.feature
```diff
@@ -0,0 +1,27 @@
+@javascript
```
josemigallas, line 1: Is this needed? I don't recall any javascript is used.

features/provider/admin/messages/outbox.feature
```diff
@@ -0,0 +1,27 @@
...
+    Given a following buyers with applications exists:
+      | name | provider        | applications        |
+      | bob  | foo.3scale.localhost | BobApp              |
+      | jane | foo.3scale.localhost | JaneApp, JaneAppTwo |
+
+    Given current domain is the admin domain of provider "foo.3scale.localhost"
+      And I don't care about application keys
+
+  Scenario: Outbox Message can't be sent without subject/body
+    Given I am logged in as provider "foo.3scale.localhost"
```
josemigallas, line 21: This is already in the Background

features/provider/admin/messages/outbox.feature
```diff
@@ -0,0 +1,27 @@
...
+    And I am on the outbox compose page
+    And a clear email queue
```
josemigallas, line 23: This should go in the Background

nidhi-soni1104:
> The test needs to be a bit more concise:
> 
> * Use as few Background steps as possible
> * Missing scenario for empty body
> * Missing scenario for sending a message

The syntax I followed is from the last feature specs I have worked and kept it same. I have committed  the suggestion but spec is not working well. I will check like what all things are needed to make it work and will commit it updated.

features/provider/admin/messages/outbox.feature
nidhi-soni1104: ok

features/provider/admin/messages/outbox.feature
```diff
@@ -0,0 +1,27 @@
...
+    And I fill in "Body" with "There is no Subject to this email"
+    And I press "Send"
+    Then I should see "Compose"
```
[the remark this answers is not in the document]
nidhi-soni1104: Okay I can check both same path and no email sent thing

features/provider/admin/messages/outbox.feature
```diff
@@ -0,0 +1,17 @@
...
+  I want to have an internal messaging system at my disposal
+
+  Background:
+    Given a provider is logged in
+
+  Scenario: Outbox Message can't be sent without subject/body
+    Given I am logged in as provider "foo.3scale.localhost"
+    And I am on the outbox compose page
+    And a clear email queue
+    And I fill in "Body" with "There is no Subject to this email"
```
lvillen, line 14: I find this line a little confusing. Maybe we should explicitly say Subject is empty?

nidhi-soni1104:
@josemigallas I have updated the test cases

- Used as few Background steps as possible, removing buyers and application wont make it work so these are need to be added.
- Added Missing scenario for empty body and empty subject
- updated assertion  and checked it with path from which we started and stayed on the same path
- Removed @javascript tag as it was not needed

Let me know if this still needs any update

josemigallas (approved):

PR #2986, opened 2022-05-11
THREESCALE-3927 remove rspec api documentation

nidhi-soni1104:

(Update our own docs to OAS 3.0)

removed gem rspec_api_documentation and its support code

https://issues.redhat.com/browse/THREESCALE-3927

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -203,7 +203,6 @@ group :test do
   gem 'rspec-rails', '~> 4.1', require: false # version 5.x is needed for Rails 6
 
   # Reason to use the fork: https://github.com/kucaahbe/rspec-html-matchers/pull/21
-  gem 'rspec_api_documentation'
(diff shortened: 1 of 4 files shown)


jkeam (approved):
looks good to me

api-docs/swagger/v1/swagger.json
```diff
@@ -0,0 +1,11 @@
+---
```
josemigallas, line 1: Should this file be a `.yml`?

josemigallas:
Can we see any results from this new gem?

api-docs/swagger/v1/swagger.json
josemigallas: But it is a `yaml` file, so it should have the `.yml` extension.

nidhi-soni1104: agreed let me do these changes and then let see how it goes

nidhi-soni1104: extensions updated

PR #2992, opened 2022-05-30
[WIP] Upgrade to formtastic 4

josemigallas:


app/views/provider/admin/user/notification_preferences/_form.html.slim
```diff
@@ -7,11 +7,11 @@
 
       ol
 
-        = f.input :enabled_notifications, as: :check_boxes_with_hints, required: false,
```
akostadinov, line 10: is `required` intentionally removed?

test/unit/three_scale/semantic_form_builder_test.rb
```diff
@@ -36,13 +36,13 @@ def test_column_for_override_needed
     assert_not_empty(semantic_form_for(file, url: '') { |f| f.input :tag_list; f.input :date })
 
     assert_raises message do
-      semantic_form_for(file, url: '', builder: ::Formtastic::SemanticFormBuilder) do |f|
+      semantic_form_for(file, url: '', builder: ::Formtastic::FormBuilder) do |f|
```
akostadinov, line 39: Still raises :( Sad.

app/inputs/check_boxes_with_hints_input.rb
```diff
@@ -0,0 +1,14 @@
...
+class CheckBoxesWithHintsInput < Formtastic::Inputs::CheckBoxesInput
+  def choice_html(choice)
+    super(choice) <<
+    template.content_tag(:p, hint(choice), class: 'inline-hints')
+  end
+
+  private
+
+  def hint(choice)
+    options[:member_hint].call(choice.is_a?(Array) ? choice.last : choice)
```
akostadinov, line 12: Why do we have it array sometimes, in case we don't need it as an array?

app/inputs/price_input.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+# Adds input followed by currency string and displays only two
+# decimals of the value.
+#
+# FIXME: if the price has more than 2 decimals than they are cut
+# on display - clicking save then damages the data!
```
akostadinov, line 7: Is this different than before?

app/views/forums/public/topics/show.html.erb
```diff
@@ -74,7 +74,7 @@
       <%= render 'forums/public/posts/anonymous', :form => form, :post => @post %>
     <% end %>
 
-    <%= form.buttons do %>
+    <%= form.actions do %>
```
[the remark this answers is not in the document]
josemigallas: Oh, indeed!

app/views/provider/admin/user/notification_preferences/_form.html.slim
josemigallas: Inputs are not required by default so I thought we could get rid of some redundancy.

test/unit/three_scale/semantic_form_builder_test.rb
```diff
@@ -87,17 +72,31 @@ def test_column_for_override_needed
     buffer = TestOutputBuffer.new
     buffer.concat(
       semantic_form_for(dummy, :url => '', as: :dummy) do |f|
-        f.input(:title, inline_errors: :list).html_safe +
+        f.input(:title).html_safe +
```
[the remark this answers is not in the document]
josemigallas: `inline_errors` was removed:
https://github.com/formtastic/formtastic/blob/abc0c94e3637a03675c6ae31f2f0df15eaee0ab1/CHANGELOG#L107

app/inputs/check_boxes_with_hints_input.rb
josemigallas: I left it as it was, we don't need it because we're using this input only in one place so the choice is always the same format. I think it was done this way to follow formtastic implementation:
https://github.com/formtastic/formtastic/blob/24812e62ccc2bb8958acdedece276e204836c74b/lib/formtastic/inputs/base/choices.rb#L47-L56

app/inputs/price_input.rb
josemigallas: Nothing has changed as far as I know but I haven't looked into this TBH, it's just copied like it was.

PR #3003, opened 2022-07-01
why does swagger response contain this extraneous object?

SohumB:
Hi there,

I'd have created an issue, but I couldn't figure out how to do so. I think issues are disabled on this repo?

I'm not entirely sure why the code I've deleted in this PR (`injectServerToResponseBody`) exists, but, as a side effect? it seems to be modifying the actual response body the user sees in the dev portal UI. 

image

This is quite clearly surprising and not desired! What would be the best way to fix this?

Thanks!

### DIFF
--- a/app/javascript/packs/OAS3Autocomplete.js
+++ b/app/javascript/packs/OAS3Autocomplete.js
@@ -73,32 +73,16 @@ const injectAutocompleteToResponseBody = (responseBody: ResponseBody | string, a
   return res
 }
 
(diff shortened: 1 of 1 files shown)


josemigallas:
Hi @SohumB, thanks for your interest in our project. We handle issues in JIRA https://issues.redhat.com/projects/THREESCALE/summary

I can't answer right now, I'll need dig in a little bit but in the meantime feel free to create a JIRA issue for better visibility.

PR #3005, opened 2022-07-06
♻️  Refactors plans_base_presenter

lvillen:

Using update_plans_controllers refactor branch, this PR continues refactoring to solve the Default plans / Table plans problems described at THREESCALE-8066 Jira Ticket.

[THREESCALE-8066](https://issues.redhat.com/browse/THREESCALE-8066) Default plan card disappears when searching for plans. 

**Verification steps** 

> 1. Go to Application Plans.
> 2. Search for an Application plan.
> 3. Confirm that Default plan card is still available and that the dropdown has information.


josemigallas (changes requested):
Merge conflicts still pending :)

test/unit/presenters/plan_presenters_test.rb
```diff
@@ -9,8 +9,8 @@ class PlanPresentersTest < ActiveSupport::TestCase
     assert_same_elements plans.order(created_at: :desc), presenter({ sort: 'created_at', direction: 'desc' }).plans
```
josemigallas, line 9: This test is wrong. See that plans are ordered by created_at and yet in your implementation you reordered by name. So this test is a false positive.
When sorting matter we should use `assert_equal` instead.

app/presenters/plans_base_presenter.rb
```diff
@@ -13,14 +13,18 @@ def initialize(service:, collection:, params: {})
...
 
-  def paginated_plans
-    @paginated_plans ||= plans.paginate(pagination_params)
-  end
-
   def plans
     @plans ||= collection.not_custom
-                         .reorder(sorting_params)
-                         .scope_search(search)
+                         .reorder('name ASC')
```
josemigallas, line 18: Instead of reordering by `name asc` to later reorder by sorting_params (which btw default value is name asc)
I would create a different collection for the default_plans select and make plans private.
```ruby
private

def plans
  @plan ||= collection.not_custom
end
```

josemigallas (changes requested):

PR #3012, opened 2022-07-19
[remove_forum] Remove admin forums

josemigallas:


nidhi-soni1104:
we can remove routes  if any

nidhi-soni1104 (commented):
We can remove routes related to this if any.

josemigallas:
> We can remove routes related to this if any.

Problem with removing routes is that it would involve removing a lot more files, since the route definitions are nested:
https://github.com/3scale/porta/blob/a00780cb7d00c665d694ae6b789af7ce25746729/config/routes.rb#L1011-L1027

And editing the routes.rb file to ignore these methods seems unnecessary since they are guarded by https://github.com/3scale/porta/blob/f05be97a4bf1ffdd3391e42217c05d7e9e2d5f94/app/controllers/forums/admin/hiden_forum_controller.rb#L9

PR #3018, opened 2022-07-20
Fix sorting test failures in postgres and Oracle CI

josemigallas:


app/presenters/plans_base_presenter.rb
```diff
@@ -8,7 +8,7 @@ def initialize(service:, collection:, params: {})
     @collection = collection
     @pagination_params = { page: params[:page] || 1, per_page: params[:per_page] || 20 }
     @search = ThreeScale::Search.new(params[:search])
-    @sorting_params = "#{params[:sort].presence || 'name'} #{params[:direction].presence || 'asc'}"
+    @sorting_params = "plans.#{params[:sort].presence || 'name'} #{params[:direction].presence || 'asc'}"
```
akostadinov, line 11: Basically postgres needed to see the table name in the sorting string? How did it work for other presenters where `sorting_params` is used in a similar way?

josemigallas: Not postgres, this is Oracle. It throws this error:
```
OCIError: ORA-00918: column ambiguously defined: 
              SELECT * FROM (
                SELECT raw_sql_.*, rownum raw_rnum_
                FROM (SELECT  "PLANS".* FROM "PLANS" INNER JOIN "SERVICES" ON "PLANS"."ISSUER_ID" = "SERVICES"."ID" WHERE "PLANS"."TYPE" IN ('ServicePlan') AND "SERVICES"."ACCOUNT_ID" = :a1 AND "PLANS"."ISSUER_TYPE" = :a2 AND "PLANS"."ORIGINAL_ID" = :a3 ORDER BY name asc ) raw_sql_ WHERE rownum <= (:a4 + :a5) ) WHERE raw_rnum_ > :a4
```

Maybe it's because of the particular complexity of this query that Oracle gets confused. For other presenters supporting sorting there should be tests already and we haven't seen this error in the logs or anything so I think we're good. This should highlight yet more the importance of exhaustive testing.

akostadinov: ahha, it is a join. makes some sense

akostadinov (dismissed):

PR #3021, opened 2022-07-22
Fix "Delete Account" scenario failing in Oracle

josemigallas:
[THREESCALE-8549](https://issues.redhat.com/browse/THREESCALE-8549)	Fix random failure "failed Delete an account plan" in oracle

### DIFF
--- a/features/step_definitions/flash_steps.rb
+++ b/features/step_definitions/flash_steps.rb
@@ -1,3 +1,22 @@
+# frozen_string_literal: true
+
 Then /^I should see the flash message "([^"]*)"$/ do |flash_text|
-  find(:xpath, "//div[@id='flash-messages'] | //div[@id='flashWrapper']").text.should have_content(flash_text)
+  assert_flash flash_text
+end
+
+def assert_flash(message)
+  # Sometimes the flash hides before the a test asserts it and fails. Therefore, we target that the flash element
+  # is rendered with the proper message, regardless of visibility.
+
+  # For admin portal see app/views/shared/provider/_flash.html.erb
+  if has_css?('#flashWrapper', visible: :all)
+    wrapper = find('#flashWrapper span')
+  # For dev portal see app/assets/javascripts/flash-buyer.js
+  elsif has_css?('#flash-messages')
+    wrapper = find('#flash-messages .alert .container')
+  else
+    raise "No flash messages container has been found"
+  end
+
+  wrapper.assert_text(:all, message)
 end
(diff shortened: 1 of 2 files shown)


akostadinov (dismissed):

nidhi-soni1104 (approved):
LGTM

mayorova (dismissed):
Looks good now after rebase :+1:

features/step_definitions/flash_steps.rb
```diff
@@ -1,3 +1,11 @@
+# frozen_string_literal: true
+
 Then /^I should see the flash message "([^"]*)"$/ do |flash_text|
-  find(:xpath, "//div[@id='flash-messages'] | //div[@id='flashWrapper']").text.should have_content(flash_text)
+  assert_flash(flash_text)
+end
+
+def assert_flash(flash_text)
+  # Sometimes the flash hides before the a test asserts it and fails. Therefore, we target that the flash element
+  # is rendered with the proper message, regardless of visibility.
```
akostadinov, line 9: do you know why?

akostadinov (dismissed):

akostadinov (dismissed):

features/step_definitions/flash_steps.rb
```diff
@@ -1,3 +1,22 @@
...
+
+  # For admin portal see app/views/shared/provider/_flash.html.erb
+  if has_css?('#flashWrapper', visible: :all)
+    wrapper = find('#flashWrapper .flash-message')
+  # For dev portal see app/assets/javascripts/flash-buyer.js
+  elsif has_css?('#flash-messages')
+    wrapper = find('.alert .container')
+  else
+    assert false, "No flash messages container has been found"
```
akostadinov, line 18:
```suggestion
    raise "No flash messages container has been found"
```

features/step_definitions/flash_steps.rb
josemigallas: I am not sure but it seems to be the test browser is a bit slow and it takes a while to render. It's like the page is rendered after a few seconds and the flash is displayed and hidden at the same time, before this step is used.
I haven't found a reliable way to reproduce this... it seems to happen only in Circleci but not when running with ssh.

akostadinov: we use small runners so being slow is expected :/
On the other hand, with oracle the runner is big. But then oracle is heavy.

akostadinov (approved):

PR #3028, opened 2022-07-28
Incorrect behaviour after Product update with invalid public base URL

lvillen:

[THREESCALE-8557 Incorrect behaviour after Product update with invalid public base URL](https://issues.redhat.com/browse/THREESCALE-8557)

**Verification steps** 

1. Go to Product Settings page (/apiconfig/services/{service_id}/settings)
2. Write invalid URL to Staging or Production public base URL text field
3. Click Update Product button at the bottom of the page

### DIFF
--- a/app/views/api/services/settings.html.slim
+++ b/app/views/api/services/settings.html.slim
@@ -3,4 +3,4 @@
 = semantic_form_for @service, url: admin_service_path(@service) do |form|
   = render partial: 'api/services/forms/integration_settings', locals: { form: form, service: @service }
   = form.actions do
-    = form.commit_button 'Update Product', button_html: { name: 'update_settings' }
+    = form.commit_button 'Update Product', button_html: { name: 'update_settings', value: 'true' }


mayorova:
Just a weird thing I noticed - the error message and the input label are swapped, compared to master.

**master**:
Screenshot from 2022-07-28 22-41-19

**fips**:
Screenshot from 2022-07-28 22-41-30

Not sure if it is critical, better check with @3scale/qe I guess.

JaurbanRH:
It would be best if it will be the same as before otherwise it will break our automatic tests

lvillen:
Apparently, at previous Formtastic versions the default inline_order was: 
https://github.com/formtastic/formtastic/blob/8741d76a95b7adf29d39df41f6d4f015cdd74473/lib/formtastic.rb#L36

But they have changed that into this:
https://github.com/formtastic/formtastic/blob/e34baba470d2fda75bf9748cff8898ee0ed29075/lib/formtastic/inputs/base/wrapping.rb#L12

So, I propose to use the default inline_order to avoid more monkey patching.

app/views/api/services/settings.html.slim
```diff
@@ -3,4 +3,4 @@
 = semantic_form_for @service, url: admin_service_path(@service) do |form|
   = render partial: 'api/services/forms/integration_settings', locals: { form: form, service: @service }
   = form.actions do
-    = form.commit_button 'Update Product', button_html: { name: 'update_settings' }
+    = form.commit_button 'Update Product', button_html: { name: 'update_settings', value: 'true' }
```
akostadinov, line 6: Looks good but what does this do?

lvillen: It ensures that when the error occurs, it doesn't redirect to the edit template, as you can see here: 

https://github.com/3scale/porta/blob/0018732a81e13aec4dfb3ab871dc7f0bcfcbd4d7/app/controllers/api/services_controller.rb#L72

akostadinov: So `value: 'true'` is just to make `#present?` return true in case there are no fields? Could you explain a little bit?

lvillen: Yes, this is another thing related to the Formtastic upgrade. At the previous version, that commit button was returning `true` by default, but after the upgrade we have to add the `value: 'true'` so the user stays at `:settings`, which was part of the issue, the user being redirected to `:edit` when sending the form with an invalid URL.

akostadinov (approved):
This brings form to the behavior of `master` so lets merge it. Investigate later the hiccup with validation when type of proxy is changed but invalid URLs set.

PR #3029, opened 2022-07-28
Unable to change product authentication method

lvillen:

[THREESCALE-8570 Unable to change product authentication method](https://issues.redhat.com/browse/THREESCALE-8570)

**Verification steps** 

1. Navigate to Product setting page (/apiconfig/services/{product_id}/settings).
2. Change authentication method.
3. Click Update Product button at the bottom of the page.
4. You get 'Product information updated' message but authentication method won't change.

### DIFF
--- a/test/integration/api/services_controller_test.rb
+++ b/test/integration/api/services_controller_test.rb
@@ -90,33 +90,37 @@ class Api::ServicesControllerTest < ActionDispatch::IntegrationTest
       assert hosted_option.attribute('checked').present?
     end
 
-    test 'shows authentication options' do
+    test 'shows and update authentication options' do
       # Default value
-      service.stubs(:backend_version)
       get settings_admin_service_path(service)
 
       page = Nokogiri::HTML::Document.parse(response.body)
       default = page.at_css('#service_proxy_authentication_method_1')
       assert default.attribute('checked').present?
-      service.unstub(:backend_version)
 
-      # API Key (user_key)
-      service.update!(backend_version: '1')
-      get settings_admin_service_path(service)
+      # App_Id and App_Key Pair 
+      put admin_service_path(service), params: update_params.deep_merge(service: { backend_version: '2' })
 
-      page = Nokogiri::HTML::Document.parse(response.body)
-      api_key_option = page.at_css('#service_proxy_authentication_method_1')
-      assert api_key_option.attribute('checked').present?
-      service.unstub(:backend_version)
+      assert_equal 'Product information updated.', flash[:notice]
 
-      # App_ID and App_Key Pair
-      service.update!(backend_version: '2')
       get settings_admin_service_path(service)
 
       page = Nokogiri::HTML::Document.parse(response.body)
       app_id_app_key_option = page.at_css('#service_proxy_authentication_method_2')
+
       assert app_id_app_key_option.attribute('checked').present?
-      service.unstub(:backend_version)
+
+      # API Key (user_key)
+      put admin_service_path(service), params: update_params.deep_merge(service: { backend_version: '1' })
+
+      assert_equal 'Product information updated.', flash[:notice]
+
+      get settings_admin_service_path(service)
+
+      page = Nokogiri::HTML::Document.parse(response.body)
+      api_key_option = page.at_css('#service_proxy_authentication_method_1')
+
+      assert api_key_option.attribute('checked').present?
     end
 
     test 'update the settings' do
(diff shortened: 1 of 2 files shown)


akostadinov:
```
🐚 rg -U "form.input[^=]+name:"
app/views/provider/signups/_user.html.slim
1:= form.inputs name: '', for: @user do |user|

app/views/provider/admin/user/personal_details/edit.html.slim
8:  = form.inputs name: 'User Information' do
14:    = form.inputs name: "Provide your current password and update your personal details" do

app/views/provider/admin/webhooks/_form.html.slim
1:= form.inputs name: 'Behaviour' do
8:= form.inputs name: 'Webhooks endpoint', class: 'inputs' do
17:= form.inputs name: 'Events that trigger webhooks', class: 'inputs' do
22:          = form.inputs name: 'Accounts' do
29:          = form.inputs name: 'Users' do
34:          = form.inputs name: 'Applications' do
45:          = form.inputs name: 'Keys' do

app/views/provider/admin/accounts/new.html.slim
4:  = form.inputs name: 'User Information', for: [:user, @user] do |user|
8:  = form.inputs name: 'Organization Information' do

app/views/buyers/accounts/edit.html.slim
12:    = form.inputs name: 'Contact Details', class: 'inputs' do

app/views/api/services/forms/_integration_settings.html.slim
18:    = form.input :proxy_authentication_method,
19:            name: 'service[backend_version]',
```

test/integration/api/services_controller_test.rb
```diff
@@ -90,6 +90,35 @@ class SettingsTest < self
       assert hosted_option.attribute('checked').present?
     end
 
+    test 'shows authentication options' do
+      # Default value
+      service.stubs(:backend_version).returns(nil)
```
akostadinov, line 95:
```suggestion
      service.stubs(:backend_version)
```

if you return `nil`, no need to specify it

test/integration/api/services_controller_test.rb
```diff
@@ -90,6 +90,35 @@ class SettingsTest < self
...
+      service.unstub(:backend_version)
+
+      # API Key (user_key)
+      service.update!(backend_version: '1')
+      get settings_admin_service_path(service)
+
+      page = Nokogiri::HTML::Document.parse(response.body)
+      api_key_option = page.at_css('#service_proxy_authentication_method_1')
+      assert api_key_option.attribute('checked').present?
+      service.unstub(:backend_version)
```
akostadinov, line 110: This `unstub` seems to be redundant because that is not `stubbed` beforehand. Do you need stubbing in this test case?

test/integration/api/services_controller_test.rb
```diff
@@ -178,6 +207,19 @@ class SettingsTest < self
...
+      get settings_admin_service_path(service)
+
+      backend_version = service.backend_version
+      assert_equal '1', backend_version
+
+      backend_version = '2'
+
+      put admin_service_path(service), params: update_params
+      assert_equal 'Product information updated.', flash[:notice]
+      assert_equal '2', backend_version
```
akostadinov, line 220: This is checking that `backend_version` equals `2` but we set this on line 216. What is the intended purpose of this assertion?

test/integration/api/services_controller_test.rb
lvillen: Yes, since I do 

https://github.com/3scale/porta/blob/b96010ed6062d18d4541dd58de4700f434188c91/test/integration/api/services_controller_test.rb#L95

If I don't have that `unstab` the test fails. 

On the other hand, though it appears with the `returns(nil)`, I'm deleting it.

test/integration/api/services_controller_test.rb
lvillen: I understand that it may seem redundant but I want to confirm it doesn't go back to the default option, since that was also part of the issue.

akostadinov: But nothing can change this local variable `backend_version`. No expression between setting it to `2` and assetion can change it. Perhaps you want to check the value from the response? Or reload from database?

lvillen: Yep, that sounds right... 😞 I'll fix it so I can check the value from the response.

test/integration/api/services_controller_test.rb
akostadinov: But on line 101 you already unstubbed.

akostadinov: Also what happens if you don't stub that in the first place?

akostadinov (approved):

PR #3032, opened 2022-08-08
Fix tests failing in circleci

josemigallas:


test/integration/sites/spam_protection_test.rb
```diff
@@ -11,11 +11,12 @@ def setup
   end
 
   def test_edit
+    Recaptcha.expects(:captcha_configured?).returns(false).twice
```
akostadinov, line 14: setting expectations twice in scenario does not work as intuitively it appears.

akostadinov:
cherry picked, thank you!

PR #3041, opened 2022-08-12
🔥 Remove IE specific code

josemigallas:
Internet explorer is no longer maintained/supported by Microsoft. 3scale has never officially supported however there are some hacks here and there to make it work. This is a technical debt that degrades our codebase and we should clean it up.

[THREESCALE-8613: Clean up IE11 code](https://issues.redhat.com/browse/THREESCALE-8613)

Our server still gets requests from IE and Windows Phone browsers but most must should come from automated test. If any customer still uses IE we announced end of support in 2020.


josemigallas:
According to Kibana there were no logs during the last year with an user agent that points to Internet Explorer [1]
Screenshot 2022-08-29 at 13 03 29

[1] - see https://www.whatismybrowser.com/guides/the-latest-user-agent/internet-explorer

thomasmaas (approved):

lvillen (dismissed):

lib/developer_portal/app/views/developer_portal/javascripts/3scale_v2.js
```diff
@@ -286,7 +286,7 @@ var hljs=new function(){function e(e){return e.replace(/&/gm,"&amp;").replace(/<
       return rails.blankInputs(form, specifiedSelector, true); // true specifies nonBlank
     },
 
-    // Helper function, needed to provide consistent behavior in IE
+    // Helper function
```
lvillen, line 289: I am a little confused with the comment deleted, I guess we still need this helper function but not specifically for IE, right? 😄

vendor/active-docs/app/assets/javascripts/active-docs/form_handler.js.erb
```diff
@@ -333,8 +333,6 @@
 
   function getHost() {
     var base = basePathComponents();
-    // This is to make IE happy.
-    // https://connect.microsoft.com/IE/feedback/details/817343/ie11-scripting-value-of-htmlanchorelement-host-differs-between-script-created-link-and-link-from-document
```
akostadinov, line 337: Comment removed but no code change. Was this intentional?

lib/developer_portal/app/views/developer_portal/javascripts/3scale_v2.js
akostadinov: Good question.

app/assets/javascripts/swagger-ui/threescale.coffee.erb
```diff
@@ -100,8 +88,6 @@ class ThreeScaleHttp
       !@enableApiDocsProxy() || @desiredOrigin() == @locationOrigin()
 
   desiredOrigin: ->
-     # This is to make IE happy.
-     # https://connect.microsoft.com/IE/feedback/details/817343/ie11-scripting-value-of-htmlanchorelement-host-differs-between-script-created-link-and-link-from-document
```
mayorova, line 104: Seems that only the comment is being removed here, but not the actual code that is specific to IE.
Checking the history... https://github.com/3scale/system/commit/09ee0fa2031ae26bd2d3338204a9498b0aa136a5

app/assets/javascripts/swagger-ui2/api_docs_proxy.coffee.erb
```diff
@@ -57,17 +52,7 @@ class ApiDocsProxy
...
-    # In some browsers like IE pathname do not includes (/) slash at the
-    # begining of the string
-    pathname = @linkElement.pathname
-    if pathname.length > 0 and pathname.indexOf('/') != 0
-      pathname = '/' + pathname
-    pathname
-
   desiredOrigin: ->
-     # This is to make IE happy.
-     # https://connect.microsoft.com/IE/feedback/details/817343/ie11-scripting-value-of-htmlanchorelement-host-differs-between-script-created-link-and-link-from-document
```
mayorova, line 70: Same as above

lib/developer_portal/app/views/developer_portal/javascripts/3scale_v2.js
josemigallas: No, I don't think we do. Suggestions for refactoring it welcome!

vendor/active-docs/app/assets/javascripts/active-docs/form_handler.js.erb
josemigallas: Yes. The link doesn't exist anymore and I don't think it's worth investigating.

lib/developer_portal/app/views/developer_portal/javascripts/3scale_v2.js
akostadinov: I think question is whether we need to delete this function now.

josemigallas: I don't think we need this helper function. I couldn't come up with a way to refactor de code in order to delete it.

vendor/active-docs/app/assets/javascripts/active-docs/form_handler.js.erb
josemigallas: I'll take a look, thanks @mayorova !

lib/developer_portal/app/views/developer_portal/javascripts/3scale_v2.js
mayorova: The whole code block seems to be coming from smth called jquery_ujs: https://github.com/rails/jquery-ujs/blob/4bdd8724d41a4dc84aa3f9b29381adb47e6ee215/src/rails.js#L361

I'd keep both the code and the comment, it does not do any harm

akostadinov: Agreed, if IE specific code is not removed, then comment should stay. Otherwise somebody will be baffled at to why is this code like that instead of something else.

lvillen (dismissed):

nidhi-soni1104 (approved):

lvillen (dismissed):

akostadinov (dismissed):
For what it's worth, looks good to me. But don't rely on my review on UI.

josemigallas:
Force pushed because I the commits where signed.

PR #3042, opened 2022-08-12
🚧 Make Quickstarts not an optional feature

josemigallas:
⚠️ Quickstarts were never visible in production SaaS, see https://github.com/3scale/deploy/pull/733/files So we should not remove the feature config until we're sure it works properly.

Quickstarts was added behind a FeatureConfig, just in case we broke production and needed to roll back quickly. Later it was refactored into a FeatureConfig, but now that it's been tested and approved it doesn't make that much sense to keep it hidden.

[THREESCALE-8615: Make quickstarts not an optional feature](https://issues.redhat.com/browse/THREESCALE-8615)


jlledom (commented):
I suggest to update `config/examples/feaures.yml` to remove `features.quickstarts.enabled: true`, since by default porta tries to load the `quickstarts` feature which doesn't exist anymore.

PR #3046, opened 2022-08-30
THREESCALE-6745: Fix ReferenceError in /p/admin/dashboard

thalesmiguel:

This PR attempts to fix the error we receive sometimes on Bugsnag (please check the ticket for the error).

Fixes [THREESCALE-6745: ReferenceError in /p/admin/dashboard](https://issues.redhat.com/browse/THREESCALE-6745)

**Verification steps** 
How to reproduce the error:
- Make sure that [app/javascript/packs/dashboard.js](https://github.com/3scale/porta/blob/master/app/javascript/packs/dashboard.js) is loaded before [app/views/provider/admin/dashboards/_ajax_load.html.slim](https://github.com/3scale/porta/blob/master/app/views/provider/admin/dashboards/_ajax_load.html.slim);
  - This can be accomplished by using a `setTimeout` in [app/javascript/packs/dashboard.js](https://github.com/3scale/porta/blob/master/app/javascript/packs/dashboard.js)

When [app/javascript/packs/dashboard.js](https://github.com/3scale/porta/blob/master/app/javascript/packs/dashboard.js) is ~somehow~ loaded before [app/views/provider/admin/dashboards/_ajax_load.html.slim](https://github.com/3scale/porta/blob/master/app/views/provider/admin/dashboards/_ajax_load.html.slim), an `Uncaught ReferenceError: dashboardWidget is not defined` exception is raised. This could happen as the request that calls `.../dashboards/_ajax_load.html.slim` is async.
To try to mitigate that, this PR removes the event listener from `.../packs/dashboard.js` as it only sets window variables and thus shouldn't have the need to wait for DOM content to be loaded.

Is this the best solution? I highly doubt that 😅 

Another idea I was exploring:

1. Have `dashboard.js` send a custom event when it's loaded:
    ```javascript
      window.dashboardWidget = dashboardWidget
      
     const widget_loaded_event = new Event('widget-loaded')
     document.dispatchEvent(widget_loaded_event)
    })
    ```
2. Have `_ajax_load.html.slim` listen to the custom event instead of `DOMContentLoaded`:
    ```javascript
    document.addEventListener('widget-loaded', function () {
      let widgetPath = #{json widget.path}

      dashboardWidget.loadAudienceWidget(widgetPath)
    })
    ```
This would assure us that the load order is preserved **BUT**, if `dashboard.js` sends the event before the event listener is in place, I believe `_ajax_load.html.slim` would not be executed. Unless there's a way to guarantee it...

Anyways, I'm opening the PR so we may have a discussion on how to best implement a solution for this case :v:


app/javascript/packs/dashboard.js
```diff
@@ -1,7 +1,5 @@
 import * as dashboardWidget from 'Dashboard/ajax-widget'
 import { render as renderChartWidget } from 'Dashboard/chart'
 
-document.addEventListener('DOMContentLoaded', () => {
-  window.dashboardWidget = dashboardWidget
-  window.renderChartWidget = renderChartWidget
-})
+window.dashboardWidget = dashboardWidget
+window.renderChartWidget = renderChartWidget
```
akostadinov, line 5: If these exist before content was loaded, then maybe fix makes sense. But I'll let UI experts chime in.

thalesmiguel:
Another possible approach: Stop setting `window.dashboardWidget` on `dashboard.js` and set it on the `_ajax_load` partial. `dashboardWidget` is only being used by the partial, so we should be able to remove it from `window` without problems. Also, we'd be sure that this error could not happen again :v:

PR #3048, opened 2022-09-02
Update setup steps (`asdf` as default now)

josemigallas:
Updates our installation guide:
* For latest macOS and Silicon chip
* Some fixes to Fedora
* asdf is the recommended version manager now for ruby, Node.js and Python.

[ASDF](https://asdf-vm.com/guide/getting-started.html#global)

Any proper dev environment requires a version manager tool, ruby and node in the case of porta. Throughout the years the team has tried a bunch of them: nvm, rvm, rbenv... and recently the majority seems to favor `asdf` which main advantage is it allows managing both packages.

Moreover, there's been times when a conflict between versions between local and prod has caused us problems.

My proposal is to promote `asdf` as the preferred version manager for our project. This way we're all in the same page and we can mitigate potential configuration issues. It will also help using the versions we're supposed to thanks to `.tool-versions`.

### DIFF
--- a/.tool-versions.sample
+++ b/.tool-versions.sample
@@ -1,2 +1,3 @@
-ruby 2.6.6
-nodejs 14.15.4
+ruby 2.6.10
+nodejs 12.22.12
+python 3.10.7 2.7.18 # Meant for M1 macs
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -54,159 +54,157 @@ to stop the application, run:
 make dev-stop
 ```
 
-## Manual setup on Mac OS X (10.13 - 12.3.1)
+## Manual setup on MacOS (12.5.1)
 
 ### Prerequisites
 
-#### Ruby version
+#### Package and runtime version management
 
-The project supports **2.6.x**.
+To add all required missing packages, make sure you have [Homebrew](https://brew.sh/) installed in your machine.
 
-Verify you have a proper version by running on your terminal:
-
-```bash
-ruby -v
+To manage the required runtimes, such as Ruby and Node.js, we recommend using [asdf](https://asdf-vm.com/guide/getting-started.html#global). Install it with brew:
+```
+brew install asdf
 ```
 
-> We recommend using [rbenv](https://github.com/rbenv/rbenv) or [rvm](https://rvm.io/) to install any ruby version.
+> Don't forget to add to your ~/.zshrc:
+>
+> ```bash
+> echo -e "\n. $(brew --prefix asdf)/libexec/asdf.sh" >> ${ZDOTDIR:-~}/.zshrc
+> ```
 
-Example with RVM:
-```bash
-gpg --keyserver hkp://pool.sks-keyservers.net --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3 7D2BAF1CF37B13E2069D6956105BD0E739499BDB
-\curl -sSL https://get.rvm.io | bash -s stable
-rvm install 2.6
+And use the provided `.tool-versions.sample` file to use the appropriate versions.
+```
+ln -s .tool-versions.sample .tool-versions
 ```
 
-#### Node version
+#### Python (only M1 macs)
+The project requires Python 2.7.18. However, it is not included anymore in Apple macs with Silicon. We recommend to handle Python installation with `asdf`:
 
-The project supports **Version: 12.X.X**.
+```
+asdf plugin add python
 
-You might want to use [nvm](https://github.com/creationix/nvm/) to install and work with specific Node versions:
-
-```bash
-curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.11/install.sh | bash
+asdf install
 ```
 
-Restart the terminal and install Node:
+#### Ruby and Node.js
 
-```bash
-nvm install 12 && nvm use 12
+The project supports **[ruby 2.6.x](https://www.ruby-lang.org/en/downloads/)** and **[Node.js 12](https://nodejs.org/en/download/)**.
+The recommended way to install them is with `asdf`:
+
 ```
+asdf plugin add ruby
+asdf plugin add nodejs
 
-###### As an alternative for Mac OS, and if you don't want multiple Node versions, you could use homebrew:
-
-```bash
-brew install node@12
+asdf install
 ```
 
 #### Xcode
 
-Install Xcode from the App Store.
-You can download all Xcode versions from [Apple's developer site](https://developer.apple.com/download/more/?name=Xcode).
+Install Xcode from the App Store. Then run the following command from your terminal to install Command Line Tools:
+```
+xcode-select —install
+```
+
+> Older versions of Xcode are available at [Apple's developer site](https://developer.apple.com/download/all/?q=xcode).
 
 #### Dependencies
 
-Make sure you have [Homebrew](https://brew.sh/) in your machine to install the following dependencies:
-
-```shell
-brew tap homebrew/cask
-brew install chromedriver --cask
-brew install imagemagick@6 mysql@5.7 gs pkg-config openssl geckodriver postgresql memcached
+```
+brew install
+brew install chromedriver imagemagick@6 mysql@5.7 gs pkg-config openssl geckodriver postgresql@14 memcached
 brew link mysql@5.7 --force
 brew link imagemagick@6 --force
 brew services start mysql@5.7
 ```
 
-Optionally, depending on your needs you can launch memcached and postgresql services
-
-```shell
-brew services start memcached postgresql
+For M1 macs you will also need:
+```
+brew install pixman cairo pango
 ```
 
+> Depending on your needs, you may want use `postgresql` instead of `mysql`.
+>
+> ```
+> brew services start postgresql@14
+> ```
+
 #### Sphinx Search
-
-[Sphinx](http://sphinxsearch.com/) has to be installed with **mysql@5.7** and compiled from source:
-
-```shell
-sed -i '' -e 's|depends_on "mysql"|depends_on "mysql@5.7"|g' /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/sphinx.rb
-brew install --build-from-source sphinx
+Install [Sphinx](http://sphinxsearch.com/) for **mysql@5.7** with Homebrew:
+```
+brew install sphinx
 ```
 
 #### Redis
 
 [Redis](https://redis.io) has to be running for the application to work. The easiest way to do it is in a [Docker](https://www.docker.com/) container by simply running:
 
-```shell
+```
 docker run -d -p 6379:6379 redis
 ```
 
 Alternatively, you can run Redis directly on your machine by using `brew`:
 
-```shell
+```
 brew install redis
 brew services start redis
 ```
 
 ### Setup
 
-#### Eventmachine
-
-Eventmachine has to be installed with `--with-cppflags=-I/usr/local/opt/openssl/include`. Simply run:
-
-```shell
-bundle config build.eventmachine --with-cppflags=-I/usr/local/opt/openssl/include
-```
-
 #### Config files
 
 Copy example config files from the examples folder:
 
-```shell
+```
 cp config/examples/* config/
 ```
 
-#### Bundle
+#### Bundler
 
-On MacOS 10.15 or newer, first configure the bundle config with:
-```shell
-bundle config --global build.eventmachine --with-cppflags=-I/usr/local/opt/openssl/include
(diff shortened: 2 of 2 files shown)


lvillen (commented):
I completely agree in using `asdf` as ruby and nodejs manager. In fact, I'm already using it and I share the same thoughts, having just one tool to manage both languages simplify things and can help us avoid issues. 

Also, great idea using the `.tool_versions` file to track versions, simplify installation process and, as you say, help us stay in the same page. 

Nicely done! 💯

lvillen (dismissed):

thalesmiguel (commented):
> This way we're all in the same page and we can mitigate potential configuration issues

+1 to that!

Could you please elaborate a little on this?
> ...there's been times when a conflict between versions between local and prod has caused us problems.

.tool-versions
```diff
@@ -0,0 +1,2 @@
+ruby 2.6.8
```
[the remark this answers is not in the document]
josemigallas: Sounds good to me! Do you know what those files are for? I've always disregarded them tbh 🤷🏻‍♂️

josemigallas:
>Could you please elaborate a little on this?
>
>...there's been times when a conflict between versions between local and prod has caused us problems.

It happened a while ago and I can't quite remember, but in involved node (10 in production image vs 16 in somebody's local) and there were some bugs happening after deploying.

I think something with ruby happened as well.

By not relegating version management we're also causing a lot of configuration headache.

INSTALL.md
```diff
@@ -214,24 +197,30 @@ yarn install
...
+apt install curl git
+git clone https://github.com/asdf-vm/asdf.git ~/.asdf --branch v0.10.2
 
-The project supports **Version: 12.X.X**.
+asdf plugin add ruby https://github.com/asdf-vm/asdf-ruby.git
+asdf plugin add nodejs https://github.com/asdf-vm/asdf-nodejs.git
+```
 
-```bash
-dnf module install nodejs:12
```
akostadinov, line 234: I prefer to keep info about native way to install nodejs as well. We should not enforce one to use asdf. We can make it easy though.

.gitignore
```diff
@@ -26,7 +26,6 @@ C:\\nppdf32Log\\debuglog.txt
 .zeus.sock
 .deploy
 .s3cfg
-.tool-versions
```
akostadinov, line 29: I don't want us to need to go through hoops to use something different than asdf.

My suggestion is to keep this ignored. But create a file `.tool-versions.default` or  `.tool-versions.example`.

If one wants to stay on the default, then `ln -s .tool-versions.default .tool-versions` will do the job.

INSTALL.md
```diff
@@ -54,49 +54,32 @@ to stop the application, run:
...
+to install a plugin for each individual package:
 
-Example with RVM:
-```bash
-gpg --keyserver hkp://pool.sks-keyservers.net --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3 7D2BAF1CF37B13E2069D6956105BD0E739499BDB
-\curl -sSL https://get.rvm.io | bash -s stable
-rvm install 2.6
+```sh
+brew install asdf
+asdf plugin add ruby https://github.com/asdf-vm/asdf-ruby.git
```
mayorova, line 76: The URL is optional, so can be just `asdf plugin add ruby`, and the same for nodejs

.tool-versions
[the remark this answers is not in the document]
thalesmiguel: Those `.sample` files are usually there so developers can kick start projects.
Since we tell `git` to ignore version manager files such as `.tool-versions` and `.ruby-version`, we put samples in the repo instead.

.gitignore
josemigallas: If `.tool-versions` is used only by `asdf` and we're using as default version manager, what's the matter in leaving it included? Won't other tools such as rbenv work if this file is present?

akostadinov: For example if I use asdf for ruby but install node from other channels, it may mess-things up. Changing the file locallly for whatever reason to try different versions will require care not to add this to commits. Switching branches, rebasing, etc. with a custom tools versions might be annoying as git will complain that there is a modified file.

For me this is local setup of everybody and I prefer the `.sample` approach. If everybody wants to have it by default, then fine. But I don't like it.

josemigallas: I get that, however, we should have a way to enforce Ruby and Node versions and I think this tool-versions + asdf strategy is an easy win. I don't mind the `.sample` approach, in fact it is like that already... missing any mention about it in the docs though 🙄

akostadinov: We had some place where we fail in case ruby is not the right version. I guess it would also be possible for node.

akostadinov: I don't see it but we can add such check somewhere.

josemigallas: I reverted to `.sample` file, resolving conversation.

.tool-versions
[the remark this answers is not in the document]
josemigallas: Following Aleksander's preference I'm gonna leave the `.sample` file and the version files ignored.

.gitignore
akostadinov: if we are back to `.sample`, then we should still ignore `.tool-versions`.

lvillen (dismissed):
LGTM! 🚀

thalesmiguel (approved):
🚀

INSTALL.md
```diff
@@ -54,189 +54,198 @@ to stop the application, run:
...
 
-> You can use [rbenv](https://github.com/rbenv/rbenv) or [rvm](https://rvm.io/) to install ruby.
+[asdf](https://asdf-vm.com/guide/getting-started.html#global) is a convenient tool version manager that reads
+from the `.tool-versions` file included in this project. This way you won't have to worry about the versioning
+of these packages and it will override the system's ruby version. After installing the required dependencies
+and asdf you need to install a plugin for each individual package:
 
-#### Node version
+```
+apt install curl git
```
mayorova, line 229: `apt` ? :thinking: 
for Fedora should be `dnf`

INSTALL.md
```diff
@@ -54,189 +54,198 @@ to stop the application, run:
...
+git clone https://github.com/asdf-vm/asdf.git ~/.asdf --branch v0.10.2
 
-The project supports **Version: 12.X.X**.
+asdf plugin add ruby https://github.com/asdf-vm/asdf-ruby.git
```
mayorova, line 232: Same as for Mac, the `.git` URLs are optional for the default plugins

INSTALL.md
```diff
@@ -54,189 +54,198 @@ to stop the application, run:
...
+asdf plugin add nodejs https://github.com/asdf-vm/asdf-nodejs.git
+```
 
-```bash
-dnf module install nodejs:12
+Finally, download the specific versions and use them in the local folder:
 ```
+asdf install
```
mayorova, line 238: Missing `ln -s .tool-versions.sample .tool-versions` step?

INSTALL.md
```diff
@@ -283,52 +292,52 @@ sudo systemctl restart redis
...
 ```
 
 #### Node packages
 
 Install [Yarn](https://yarnpkg.com/):
 
-```bash
-brew install yarn
+```
+sudo dnf install yarnpkg
```
mayorova, line 320: I can't find this package. Maybe `npm install --global yarn` ?

INSTALL.md
```diff
@@ -54,189 +54,198 @@ to stop the application, run:
...
-
-```bash
-ruby -v
+To manage the required runtimes, such as Ruby and Node.js, we recommend using [asdf](https://asdf-vm.com/guide/getting-started.html#global). Install it with brew:
+```
+brew install asdf
 ```
 
-> We recommend using [rbenv](https://github.com/rbenv/rbenv) or [rvm](https://rvm.io/) to install any ruby version.
+> Don't forget to add to your ~/.zshrc:
```
mayorova, line 70: Seems a bit specific, not everyone might be using ZSH. Maybe just point to https://asdf-vm.com/guide/getting-started.html#_3-install-asdf

INSTALL.md
josemigallas: According to https://developer.fedoraproject.org/tech/languages/nodejs/nodejs.html it should be available since Fedora 29.

INSTALL.md
josemigallas: This instruction is included in brew's output so it's just a reminder. We could say something like this instead:
```suggestion
> Don't forget to check for any *Caveats* in `brew`'s output.
```

On the other hand, it is specific because ZSH is the default shell in macOS.
The link is already present in the second paragraph (click on 'asdf') so anybody with different setup can check specific steps for it.

INSTALL.md
josemigallas: Actually the whole section is still using the old format, I will update it.

INSTALL.md
mayorova: Oh, OK, I forgot that ZSH was default on macOS.

INSTALL.md
mayorova: Hmm, now I see it! :upside_down_face:  Don't know what went wrong the first time I tried :grimacing:

INSTALL.md
```diff
@@ -54,189 +54,197 @@ to stop the application, run:
...
-
-The project supports **2.6.x**.
+#### Package and runtime version management
 
-Verify you have a proper version by running on your terminal:
+To manage the required runtimes, such as Ruby and Node.js, we recommend using [asdf](https://asdf-vm.com/guide/getting-started.html#global). Install it directly from Github:
+```
+dnf install curl git
+git clone https://github.com/asdf-vm/asdf.git ~/.asdf --branch v0.10.2
+```
```
jlledom, line 221: Instead of this, I would say "Follor the instructions at [asdf sire](https://asdf-vm.com/guide/getting-started.html#_2-download-asdf) to install it". Since it needs to add some lines to `.bashrc` which doesn't appear in this guide, and maybe they change that in future releases. So better just link there.

INSTALL.md
jlledom: I couldn't install the dependencies using the `yarnpkg` package in `dnf`. Don't remember the error, but I had to uninstall it and use the yarn package at `npm install --global yarn`. That worked. Fedora 36.

INSTALL.md
josemigallas: What about:
>To manage the required runtimes, such as Ruby and Node.js, we recommend using [asdf](https://asdf-vm.com/guide/introduction.html). See its [getting started](https://asdf-vm.com/guide/getting-started.html) guide in order to get specific instructions for your SHELL and installation method.

.tool-versions.sample
```diff
@@ -1,2 +1,3 @@
-ruby 2.6.6
-nodejs 14.15.4
+ruby 2.6.8
+nodejs 14.20.0
+python 3.10.6 2.7.18
```
akostadinov, line 3: we require python?

akostadinov: ok, it is a mac thing

akostadinov: I'm just concerned that new users may be confused and install python from asdf on fedora, which could be confusing later. But I don't know how to explain this well.

INSTALL.md
jlledom: That's perfect

.tool-versions.sample
josemigallas: Python: some packaged required it as a dependency, not sure if postgres or mysql so yes we require it. We could add a comment to the right saying "For M1 macs missing python".

Node 14 vs 12: good catch!! I need to use node 14 because of the new Mac chip but it should be 12.

INSTALL.md
```diff
@@ -338,7 +342,7 @@ Command above will seed the application and create the Master, Provider & Develo
 
 Alternatively you can set environment variables before running command to have predictable setup. For more you can look at `db/seed.rb`.
 
-```bash
+```sh
```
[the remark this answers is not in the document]
josemigallas: Same thing, it's just for consistency throughout the file 😅 TOC hitting me you know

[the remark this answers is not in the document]
josemigallas: In fact, I'm gonna change all to bash which was the one being used.

lvillen (approved):

.tool-versions.sample
akostadinov: should we maybe upgrade to 14 for all. 12 not supported in Fedora 36 anymore. Also in RHEL8 I'm not sure or RHEL9.

josemigallas: It's already planned: https://issues.redhat.com/browse/THREESCALE-7331
I'd stick with node 12 in our local env unless strictly necessary.

PR #3054, opened 2022-09-08
⚰️ Flow to TS

josemigallas:
[THREESCALE-8721: Migrate from Flow to Typescript](https://issues.redhat.com/browse/THREESCALE-8721)


.circleci/config.yml
```diff
@@ -28,17 +28,10 @@ store-log-artifacts: &store-log-artifacts
     path: log
     destination: log
 
-flow-type-key: &flow-typed-cache-key flow-typed-{{ checksum "yarn.lock" }}-5
```
akostadinov, line 31: do we want to cache TS compilation results or something?

.circleci/config.yml
```diff
@@ -497,13 +485,10 @@ jobs:
           command: |
             bundle exec rake doc:swagger:validate:all
             bundle exec rake doc:swagger:generate:all
-      - *restore-flow-typed-cache
       - run:
-          name: Eslint & Flow
+          name: Eslint
           command: |
-            yarn flow:install
             yarn lint
```
akostadinov, line 491: maybe we want to check TS typing somewhere here? Or as a parallel job.

josemigallas: Typings are check during compilation, no need for extra steps. For instance before running tests.

.circleci/config.yml
josemigallas: I don't think so. What was cached before were the typings, not actual code. Typescript's typings are stored inside `node_modules` so they are already cached.

.circleci/config.yml
akostadinov: I read that compiling with babel does not check types. So is we use babel instead of ts for compiling, we will need a step to check types. Or will we stop using babel?

.circleci/config.yml
akostadinov: sounds good

.circleci/config.yml
josemigallas: First the code is transpiled into JS and JSX by typescript, via ts-loader, then it is polyfilled by babel. Removing this dependencies from Babel is what @lvillen is working on anyway.

akostadinov: Where is the compilation/transpilation step done?

josemigallas: Rake task is invoked here by circleci:https://github.com/3scale/porta/blob/4f138ded4460aaab9158009ca513c46f0fa93c0c/.circleci/config.yml#L457

Which runs webpacker, which runs webpack, which uses the ts-loader which transpiles all TS and TSX:https://github.com/3scale/porta/blob/4f138ded4460aaab9158009ca513c46f0fa93c0c/config/webpack/environment.js#L13-L16

Once all is parsed into JS / JSX, it runs the babel-loader to apply polyfills and plugins (such as React):https://github.com/3scale/porta/blob/4f138ded4460aaab9158009ca513c46f0fa93c0c/config/webpack/environment.js#L5-L6

But we've been reading about this and we don't actually need any polyfills so probably Typescript can handle all transpilation by itself, including React.

mayorova:
Create Application from Account is broken:
Screenshot from 2022-10-20 14-45-49


Screenshot from 2022-10-20 14-46-01

Screenshot from 2022-10-20 14-46-08

jlledom (dismissed):

lvillen (dismissed):

PR #3055, opened 2022-09-09
Add info about Rails cache to INSTALL.md

jlledom:

It adds a new section to INSTALL.md to show how to enable the rails cache for development

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -318,15 +318,15 @@ Run [Yarn](https://www.yarnpkg.com/) to install all the required dependencies:
 yarn install
 ```
 
-## Enable caching
+## Rails cache
 
-Caching is disabled by default in Rails. It can be switched on/off by running:
+Rails cache is enabled by default for development, and uses a memcached instance that must be listening at `localhost:11211`.
+However, you might need to disable it in your environment. It can be switched off by updating `config/cache_store.yml` to this:
 
 ```bash
-bundle exec rails dev:cache
+development:
+  - :null_store
 ```
-
-For caching to work, a memcached instance must be running and listening at `localhost:11211`
 
 ## Setup Database
 
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -318,6 +318,16 @@ Run [Yarn](https://www.yarnpkg.com/) to install all the required dependencies:
 yarn install
 ```
 
+## Enable caching
+
+Caching is disabled by default in Rails. It can be switched on/off by running:
+
+```bash
+bundle exec rails dev:cache
+```
+
+For caching to work, a memcached instance must be running and listening at `localhost:11211`
+
 ## Setup Database
 
 Finally, initialize the database with some seed data by running:
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -28,21 +28,14 @@ Rails.application.configure do
 
   config.public_file_server.enabled = true
 
-  # Enable/disable caching. By default caching is disabled.
-  # Run rails dev:cache to toggle caching.
-  if Rails.root.join('tmp', 'caching-dev.txt').exist?
-    config.action_controller.perform_caching = true
+  # Enable caching.
+  config.action_controller.perform_caching = true
 
-    # when we don't set this, the default from application config is used
-    # config.cache_store = :memory_store
-    config.public_file_server.headers = {
-      'Cache-Control' => "public, max-age=#{2.days.to_i}"
-    }
-  else
-    config.action_controller.perform_caching = false
-
-    config.cache_store = :null_store
-  end
+  # when we don't set this, the default from application config is used
+  # config.cache_store = :memory_store
+  config.public_file_server.headers = {
+    'Cache-Control' => "public, max-age=#{2.days.to_i}"
+  }
 
   # Store uploaded files on the local file system (see config/storage.yml for options)
   # config.active_storage.service = :local


akostadinov:
It is enabled by default in `config/examples/cache_store.yml`. The example configs is something everybody is expected to use unless they have special needs.

jlledom:
AFAIK it's disabled by default, because it only uses that configuration when the file `tmp/caching-dev.txt` exists; otherwise it uses `:null-store`. See https://github.com/3scale/porta/blob/master/config/environments/development.rb#L31

akostadinov:
We have it set also in `config/application.rb`. I'm not sure which option takes precedence.

jlledom:
I think it takes the value from `config/application.rb` when it's enabled, and sets the value to `:null_store` when disabled. See https://github.com/3scale/porta/blob/master/config/environments/development.rb#L36

This is what I see in my local environment

```
$ bundle exec rails dev:cache
Development mode is now being cached.
$ bundle exec rails c
Loading development environment (Rails 5.2.8)
[1] pry(main)> Rails.application.config.cache_store
=> [:mem_cache_store, ["localhost:11211"], {}]

$ bundle exec rails dev:cache
Development mode is no longer being cached.
$ bundle exec rails c
Loading development environment (Rails 5.2.8)
[1] pry(main)> Rails.application.config.cache_store
=> :null_store
```
So the default value depends on the presence of `tmp/caching-dev.txt` when you run rails for the first time after cloning the repo, which I assume is false since that folder is empty in the repo.

akostadinov:
I understand that this is the default for Rails, but is we want to have it enbaled in general, why don't we change config to make it always enabled only depending on the configuration file?

Enabling by reading the docs doesn't seem very reliable and then if one loses the `tmp` dir or setup new environment, it would be easy to forget.

jlledom:
In my opinion is better to work having it enabled, but if we update the config file to enable it by default, then the `bundle exec rails dev:cache` won't have any effect anymore and the only way to disable the cache locally would to update the config file again and have the file bothering in the git index and potentially being commited by mistake. If somebody in the team knows how to enable the cache by default without leaving any trace in the git index, I'd go for it.

akostadinov:
`/config/cache_store.yml` is in `.gitignore` already.

jlledom:
If I'm not wrong, this is how it works:

1. [cache_store.yml](https://github.com/3scale/porta/blob/e7845856a42cc58a85376568c11b1df72e5cb56b/config/examples/cache_store.yml#L1): Contains the date for the default cache store, in this case memcached.
2. [application.rb](https://github.com/3scale/porta/blob/e7845856a42cc58a85376568c11b1df72e5cb56b/config/application.rb#L155): Reads `cache_store.yml` to get the cache params and configures a default cache store in rails. This doesn't mean it's enabled, only configured.
3. [development.rb](https://github.com/3scale/porta/blob/e7845856a42cc58a85376568c11b1df72e5cb56b/config/environments/development.rb#L31): This enables or disables the default cache store based on the presence of the file `tmp/caching-dev.txt`

So AIUI updating cache_store.yml would be useful to switch to another cache store different from memcache, but not for enable or disable it.

akostadinov:
If we use `:null_store`, does it matter that `config.action_controller.perform_caching = true`?

My line of thinking is that with `null_store` there will be no caching anyway. Should be very rare if ever to need `config.action_controller.perform_caching = false` for this to be a problem.

Or do you see more benefits to keep using `rails dev:cache` as your original suggestion goes?

akostadinov:
btw we may implement logic to disable caching when `null_store` is used but that would be an over-engineering as I see it.

jlledom:
You're right. We can leave it always enabled in `development.rb`, and then use `cache_store.yml` to set the store to `memcache` or null, using `memcache` as default as it's now. That would work and not mess with the git index. So I'd say we have two options:

- Leaving as it is:
  - Pros: It can be enabled/disabled by `rails dev:cache` which is the standard way and documented in the official Rails docs
  - Cons: Disabled by default
- Enable it by default and allow contributors to disable by through `cache_store.yml`
  - Pros: Enabled by default
  - Cons: Not the standard way

I prefer to follow the standard but don't have a strong opinion. Should we ask the team?

akostadinov:
I don't see it as a problem. This standard does not make so much sense to me in case we want some other default. Also the rake task can be overridden to fails and print a message that you control caching by editing the file. It should not confuse anybody in that way.

You were the driver behind this. So you decide what to do. If you want, I'm ok to just go ahead with the change. If you in doubt, asking the rest of the team is good with me too.

jlledom:
In that case, I prefer to follow the standard. I think that's always the best choice, and enabling the cache it's just one command, not a big deal.

akostadinov:
When I wrote
> If you want, I'm ok to just go ahead with the change. If you in doubt, asking the rest of the team is good with me too.

I meant that if you want, you can make this enabled by default. I didn't mean that doc-only change is fine with me. Sorry if I was not clear.

I believe nobody would notice this doc and remember to do it. I don't see any benefits in requiring a manual step. 

I differentiate between a standard functionality in rails and a "standard". It is standard in Rails that memcached is **not** used in dev mode, yet you want it enabled. One can argue that it's against the standard. So you want us to stop following the standard but do it manually and have to think about it, instead of making things work as we want OOB. I know for sure that if something depends on my memory, then I can guarantee you that immediately or very soon I stop thinking about it and forget it. And I don't know people that always remember everything, although many people are better "rememberers" than me but some are even worse.

That's why I don't agree with a doc-only change.

jlledom:
I pushed the changes to have it enabled by default

akostadinov (approved):

PR #3057, opened 2022-09-09
revert back to Sphinx 2.2.11

akostadinov:
Sphinx 3.x is non-FLOSS so we can't use it.
It also lacks builds for non-x86 archs.
We can update in the future to Manticore.
Especially when we update to RHEL8 base image,
2.2.11 fails to work with MySQL 8.


akostadinov:
Build succeeded here: https://quay.io/repository/3scale/porta/build/3dd7bf5e-d858-4f00-a4a7-23618cb9a52c

mayorova (approved):
Thank you @akostadinov ! :+1:

PR #3058, opened 2022-09-12
THREESCALE-8393 Prevent message from being sent without subject  from dev portal

nidhi-soni1104:

Prevent message from being sent without subject  from dev portal

Fixes [THREESCALE-8393: Prevent message from being sent without subject  from dev portal](https://issues.redhat.com/browse/THREESCALE-8393 )

**Verification steps**
In the developer portal - /admin/messages/new  - page under Messages > Compose
1. try sending message without subject.

Notes:
1. As per discussion its decided to make subject as required field and so added validation
2.I have removed some unwanted test as added new in controller test


test/functional/developer_portal/admin/messages/outbox_controller_test.rb
```diff
@@ -39,4 +39,22 @@ def test_index
...
+    assert msg.valid?
+  end
+
+  test "will not create messages with only body and no subject" do
+    buyer = FactoryBot.create :buyer_account, :provider_account => @provider
+
+    msg = Message.new
+    msg.sender_id = buyer.id
+    msg.body = "I am body"
+    assert !msg.valid?
```
mayorova, line 58:
```suggestion
    assert_not msg.valid?
```

app/models/message.rb
```diff
@@ -32,6 +32,7 @@ class Message < ApplicationRecord
 
   validates :sender_id, presence: true
   validates :type, :origin, length: { maximum: 255 }
+  validates :subject, presence: true
```
mayorova, line 35: A bit scary to add such a validation, when we know that there might be many messages without subject in the DB.
The good thing is that it seems that once created the Message objects are not modified, so maybe this is fine.

mayorova:
I've left a comment on the [JIRA](https://issues.redhat.com/browse/THREESCALE-8393?focusedCommentId=20945478&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-20945478)

I am not too sure that having empty subject is actually a problem. To be honest, I would keep it as it is, but let's wait for feedback.

app/models/message.rb
nidhi-soni1104: yes but as per the discussion we need to make it mandatory and somehow it makes sense because we should never send empty messages some validation is must

nidhi-soni1104:
> I've left a comment on the [JIRA](https://issues.redhat.com/browse/THREESCALE-8393?focusedCommentId=20945478&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-20945478)
> 
> I am not too sure that having empty subject is actually a problem. To be honest, I would keep it as it is, but let's wait for feedback.

yes let see what others have to say for the same

app/models/message.rb
akostadinov: It makes sense to enforce good etiquette I guess.

test/unit/message_test.rb
```diff
@@ -141,8 +141,8 @@ class MessageTest < ActiveSupport::TestCase
     assert_equal 1, message.errors[:sender_id].size
   end
 
-  test 'A message not require a subject' do
-    message = FactoryBot.build(:message, subject: nil, sender: FactoryBot.create(:simple_account))
+  test 'A message require a subject' do
+    message = FactoryBot.build(:message, subject: "I am subject", sender: FactoryBot.create(:simple_account))
```
thalesmiguel, line 145: This is testing that an object of `Message` with subject AND sender (and maybe something else from the Factory) is valid, which's more than the description lets us know.
Maybe a shorter way to test this is to instantiate a `Message` without a `subject` and check that it's invalid AND that there's an error message for `subject`.

app/models/message.rb
nidhi-soni1104: umm I think the requirement is to make subject required field and yes may be I can cofirm if character limit needed

test/functional/developer_portal/admin/messages/outbox_controller_test.rb
```diff
@@ -39,4 +39,22 @@ def test_index
...
+    assert msg.valid?
+  end
+
+  test "will not create messages with only body and no subject" do
+    buyer = FactoryBot.create :buyer_account, :provider_account => @provider
+
+    msg = Message.new
+    msg.sender_id = buyer.id
+    msg.body = "I am body"
+    assert_not msg.valid?
```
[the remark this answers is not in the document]
nidhi-soni1104: assert_not and refute both can be used but mostly in 3scale we have used assert_not

test/unit/message_test.rb
nidhi-soni1104: I think this test failed because there were no subject and if we have added validation of required filed so I think we must enter it in test also so if in future we change something atleast things reflect what failed or not

app/workers/message_worker.rb
```diff
@@ -8,6 +8,7 @@ def self.enqueue(recipients, attributes)
   def perform(recipients, attributes)
     message = Message.new(attributes)
     message.sender_id = attributes["sender_id"]
+    message.subject = attributes["subject"]
```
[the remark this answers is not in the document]
akostadinov: We still use the `protected_attributes` gem so it is possible that these attributes are protected and need individual assignment. I don't know if this is the case! Just saying.

But then if this **is** the case, to avoid warnings in log, we should filter them out when mass-assigning.

test/functional/developer_portal/admin/messages/outbox_controller_test.rb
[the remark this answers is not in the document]
akostadinov: codeclimate complains about `refute`, maybe we configured it like that. That's why we use `assert_not`.

app/workers/message_worker.rb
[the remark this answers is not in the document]
nidhi-soni1104: yes true sender_id is the protected one and so assigned it this way and similar to that subject as well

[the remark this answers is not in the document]
akostadinov: Ok, to avoid warnings in server log, you'd have to remove first these from `attributes` in the `::new` call.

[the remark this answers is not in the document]
nidhi-soni1104: but is it really needed?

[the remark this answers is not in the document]
akostadinov: Is it needed to avoid unnecessary warnings? I think yes. Reducing noise in logs helps catch actual things that need to be fixed.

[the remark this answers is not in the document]
nidhi-soni1104: done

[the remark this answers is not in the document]
nidhi-soni1104: @akostadinov Can we resolve this too?

app/models/message.rb
nidhi-soni1104: I have received comment from team on JIRA

[Nidhi Soni](https://issues.redhat.com/secure/ViewProfile.jspa?name=nsoni%40redhat.com) I think it is not necessary to add a minimal character limit. (Everything  message > 0 would be fine)

So I am resolving this conversation as we have already made subject required and other than that works fine for them.

test/unit/message_test.rb
nidhi-soni1104: I have added one more test which cover your explained scenario as well

lib/developer_portal/app/controllers/developer_portal/admin/messages/outbox_controller.rb
```diff
@@ -34,21 +34,27 @@ def destroy
 
 
   def create
-    @message.enqueue! :to => current_account.provider_account.id
-    @notice = 'Message was sent.'
+    if @message.subject.present?
+      enqueue_message
+    else
+      flash[:error] = 'Please fill subject.'
```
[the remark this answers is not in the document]
nidhi-soni1104: Correct check updated

nidhi-soni1104:
@mayorova @thalesmiguel can we merge this PR now?

features/provider/admin/messages/outbox.feature
```diff
@@ -1,24 +0,0 @@
-Feature: Outbox messages
```
mayorova, line 1: Why is the whole file removed?

nidhi-soni1104: These test are covered here test/functional/developer_portal/admin/messages/outbox_controller_test.rb which makes more sense

app/models/message.rb
```diff
@@ -32,11 +32,12 @@ class Message < ApplicationRecord
...
 
   after_save :update_recipients
   after_initialize :default_values
 
-  attr_protected :sender_id, :tenant_id
+  attr_protected :tenant_id
```
mayorova, line 40: I think this is dangerous. Although this issue is present in the previous implementation also.
Basically, by passing `message[sender_id]` in a form, the user can send messages as if they were from another account. I think we need to discard this field, and set it through the `current_account`.

app/views/provider/admin/messages/outbox/index.html.slim
```diff
@@ -21,7 +21,7 @@ h2 Sent Messages
       - @messages.each do |message|
         tr id="message_#{message.id}"
           td.select= bulk_select_one message
-          td = link_to message.subject, provider_admin_messages_outbox_path(message)
+          td = link_to message_subject(message), provider_admin_messages_outbox_path(message)
```
mayorova, line 24: I have added this to show the `(No subject)` message with the link, rather than not showing nothing. This will be only applied for already existing messages without subject/body.

See the discussion starting from https://issues.redhat.com/browse/THREESCALE-8393?focusedCommentId=20974783&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-20974783

mayorova (changes requested):

lib/developer_portal/app/controllers/developer_portal/admin/messages/outbox_controller.rb
```diff
@@ -34,21 +34,27 @@ def destroy
 
 
   def create
-    @message.enqueue! :to => current_account.provider_account.id
-    @notice = 'Message was sent.'
+    if @message.valid?
+      enqueue_message
+    else
+      flash[:error] = 'Please fill subject.'
```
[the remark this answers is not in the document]
nidhi-soni1104: Sure, Let me do that

app/models/message.rb
mayorova: Well, I mean that we should not permit `sender_id` to be taken from the form data (i.e. coming from the user).

test/functional/developer_portal/admin/messages/outbox_controller_test.rb
```diff
@@ -39,4 +39,22 @@ def test_index
     assert assigned_drop_variables.include?('messages')
     assert assigned_drop_variables.include?('pagination')
   end
+
+  test "will create messages with only subject and not body" do
```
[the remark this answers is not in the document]
mayorova: Move these new tests to `test/unit/message_test.rb`, and in `outbox_controller_test.rb` add another test, which checks the controller functionality

app/models/message.rb
nidhi-soni1104: So we have made it attr_protected

lib/developer_portal/app/controllers/developer_portal/admin/messages/outbox_controller.rb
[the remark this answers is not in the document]
nidhi-soni1104: Done

test/functional/developer_portal/admin/messages/outbox_controller_test.rb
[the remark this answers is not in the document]
nidhi-soni1104: Okay can do that
But these test are enough to check I think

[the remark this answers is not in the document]
nidhi-soni1104: Updated are done

features/provider/admin/messages/outbox.feature
mayorova: But the tests are slightly different. I wouldn't remove the tests that already exist and validate functionality.

mayorova: Actually, this test seems to be for the **provider** outbox page, rather than the **developer portal**. And this test breaks with the change.

mayorova:
In [this comment](https://github.com/3scale/porta/pull/3058#discussion_r988883708) I was actually referring to having something like the following:
devportal_outbox_error

In fact, I tried it in this branch: https://github.com/3scale/porta/compare/THREESCALE-8393-stop-blank-message...mayorova:porta:THREESCALE-8393-stop-blank-message-alternative?expand=1

However, I found a couple of issues with it:

- when using `render` and not `redirect_to`, it messes up with the tabs menu, due to the fact that the URL points to the form submit URL.
- as this is a developer portal template (cutomizable), the change in the template would only apply for the newly generated developer portals, and not for the exiting ones (in the existing ones no errors would be shown on the page).

So, probably the solution with flash error is OK (although not too consistent with some other forms in the developer portal).

test/functional/developer_portal/admin/messages/outbox_controller_test.rb
```diff
@@ -39,4 +39,18 @@ def test_index
     assert assigned_drop_variables.include?('messages')
     assert assigned_drop_variables.include?('pagination')
   end
+
+  test "Not Creates invalid message" do
+    buyer = FactoryBot.create :buyer_account, :provider_account => @provider
+    res = post :create, params: { message: { subject: nil, :body => "message with nil subject" }, :to => buyer.id }
```
mayorova, line 45: Here and on line 52 `res` variable is not used.

Also, I'm wondering if we need to run `MessageWorker.drain`.

mayorova:
I've changed `app/controllers/provider/admin/messages/outbox_controller.rb`, as otherwise it was possible to create a message with empty body from the admin portal (bypassing frontend validation), which would cause the Sidekiq queue filling with invalid jobs, that would be stuck there and keep failing.

We probably need a test for this controller too.

test/functional/developer_portal/admin/messages/outbox_controller_test.rb
nidhi-soni1104: Yes we can remove res and yes we can run MessageWorker.drain, that woul drain and make test  more clear

Changes are done

nidhi-soni1104:
> In [this comment](https://github.com/3scale/porta/pull/3058#discussion_r988883708) I was actually referring to having something like the following: devportal_outbox_error
> 
> In fact, I tried it in this branch: https://github.com/3scale/porta/compare/THREESCALE-8393-stop-blank-message...mayorova:porta:THREESCALE-8393-stop-blank-message-alternative?expand=1
> 
> However, I found a couple of issues with it:
> 
> * when using `render` and not `redirect_to`, it messes up with the tabs menu, due to the fact that the URL points to the form submit URL.
> * as this is a developer portal template (cutomizable), the change in the template would only apply for the newly generated developer portals, and not for the exiting ones (in the existing ones no errors would be shown on the page).
> 
> So, probably the solution with flash error is OK (although not too consistent with some other forms in the developer portal).

Yes I also Tried similar things and so I added flash error message. May be we can go with flash message now.

features/old/messages/provider_side.feature
```diff
@@ -20,19 +20,11 @@ Feature: Provider side messages
   Scenario: Sending a message
     Given account "bob" has no messages
 
-    And I go to the provider dashboard
-    And I navigate to the accounts page
-    And I follow "bob"
-    And I follow "Send message"
-    Then the "To" field should be fixed to "bob"
+    And I am on the outbox compose page
```
mayorova, line 23: I'm afraid this "fix" is not the right one :confused: 
The original test was validating sending messages for individual account, it's done in a pop-up, not the outbox compose page.
In fact, by failing this test caught a real bug (that I introduced :upside_down_face:  ), see the screenshot - a successful message sending got a "null" flash message.

See the top-right corner of the screenshot.
Screenshot from 2022-10-18 17-53-55

mayorova (changes requested):

features/old/messages/provider_side.feature
nidhi-soni1104: @mayorova which user is this? can you share ur?

yes screenshots looks right, I checked your fix also looks good

nidhi-soni1104: you reverted the test and I think its same as previous only message is changed which is assigned through variable

mayorova: @nidhi-soni1104 That happens for any user.

Yes, I reverted the test, because it's actually valid.

Indeed, I put the flash message in an instance variable. The reason is that in on this send message form (which is a pop-up), the flash messages are handled differently, and they are expected to be in a variable - no actual reload takes place on that page, it's done through JavaScript.

test/unit/message_test.rb
```diff
@@ -167,6 +174,25 @@ class MessageTest < ActiveSupport::TestCase
     buyer_sent_message.restore_for!(provider)
     assert_not provider_received_message.reload.hidden?
   end
+
+  test "will create messages with only subject and not body" do
```
mayorova, line 178: We moved these two tests from the controller tests, but above there are two other new tests that already test the same, so these two are redundant.

test/unit/message_test.rb
```diff
@@ -141,11 +141,18 @@ class MessageTest < ActiveSupport::TestCase
     assert_equal 1, message.errors[:sender_id].size
   end
 
-  test 'A message not require a subject' do
-    message = FactoryBot.build(:message, subject: nil, sender: FactoryBot.create(:simple_account))
+  test 'A message require a subject' do
```
mayorova, line 144: This is actually tested in the test on line 133

lib/developer_portal/app/controllers/developer_portal/admin/messages/outbox_controller.rb
```diff
@@ -32,23 +32,28 @@ def destroy
...
-        flash[:notice] = @notice
+        flash[:notice] = 'Message was sent.'
         redirect_to admin_messages_root_path
       end
-
       format.js
     end
   end
 
-  private
```
mayorova, line 50: Why this `private` was removed?

test/functional/developer_portal/admin/messages/outbox_controller_test.rb
```diff
@@ -39,4 +39,22 @@ def test_index
...
+    assert_not_includes "message with nil subject", msg.body
+  end
+
+  test "Creates valid message" do
+    buyer = FactoryBot.create :buyer_account, :provider_account => @provider
+    post :create, params: { message: { subject: "Valid Message", :body => "message with subject" }, :to => buyer.id }
+
+    MessageWorker.drain
+    assert msg = Message.last
+    assert "message with subject", msg.body
```
mayorova, line 58: this assert with always pass, because your're asserting here that `"message with subject"` is truthy :upside_down_face:

mayorova (approved):
LGTM

features/old/messages/provider_side.feature
nidhi-soni1104: Oh good to  know this

nidhi-soni1104:
Can I merge this PR now??

mayorova:
@nidhi-soni1104 whenever the tests pass

PR #3061, opened 2022-09-14
🚧 THREESCALE-7132: Add support for STS authentication for S3 bucket used in 3Scale

thalesmiguel:

This PR adds a patch for Paperclip so that we're able to pass AWS `session_token`s to it. The idea is to support the use of temporary AWS tokens to access our S3 buckets.

Implements https://issues.redhat.com/browse/THREESCALE-7132.

**Verification steps** 

Testing in local dev env:
1. Add valid credentials to `config/amazon_s3.yml`;
2. Perform a request using Paperclip;
          a.  Could be a file upload or a check if a file exists;
3. Assert that the `session_token` is used on the request;

This is the first draft of what we could use to support STS without migrating from Paperclip to ActiveStorage;
We still have to better understand how this `session_token` will be present to the Porta application, which most probably will result in further development.

Furthermore, I'd like to share a couple lines of code used to test AWS session tokens:

```ruby
# Use access_key_id, secret_access_key and region to instantiate an Aws::STS::Client
# and get the temporary session_token
sts = Aws::STS::Client.new(CMS::S3.credentials.merge(region: CMS::S3.region))
session = sts.get_session_token(duration_seconds: 15*60)
# Use the access_key_id, secret_access_key and temporary session_token (and the region again...)
# to access S3.
# NOTE: This "session" variable includes the access_key_id and secret_access_key
s3 = Aws::S3::Client.new(credentials: session, region: CMS::S3.region)
# Call whatever S3 api just to test if it's working properly
s3.get_bucket_tagging(bucket: 'some_bucket')
#=> #<struct Aws::S3::Types::GetBucketTaggingOutput...

# Also, Paperclip instantiates an Aws::S3::Resource to perform calls. This piece of code tests that a session_token is usable:
resource = Aws::S3::Resource.new(
  access_key_id: session.credentials.access_key_id,
  secret_access_key: session.credentials.secret_access_key,
  region: CMS::S3.region,
  session_token: session.credentials.session_token
)
resource.client.get_bucket_tagging(bucket: 'edge-flexo.3scale.net.3scale.net')

```


app/lib/cms/s3.rb
```diff
@@ -35,7 +35,7 @@ def region
     end
 
     def credentials
-      config.slice(:access_key_id, :secret_access_key) if enabled?
+      config.slice(:access_key_id, :secret_access_key, :session_token) if enabled?
```
thalesmiguel, line 38: I guess this `session_token` value will actually come from another place, instead of a config file, as the value should be dynamic...
Depending on how it goes, we might need a "dynamic_credentials" of sort of even perform a request to AWS ourselves to get the credentials.

thalesmiguel:
Another point to consider is that, if the role_arn is provided, we could use something like:
```ruby
role_credentials = Aws::AssumeRoleCredentials.new(
  client: Aws::STS::Client.new(CMS::S3.credentials.merge(region: CMS::S3.region)),
  role_arn: "some::nice::arn",
  role_session_name: "a-session-name"
)

# Paperclip things...
Aws::S3::Resource.new(role_credentials)
# Paperclip things...
```

lib/paperclip/storage/s3.rb
```diff
@@ -0,0 +1,34 @@
...
+              userinfo += ":#{http_proxy_password}" if http_proxy_password
+              proxy_opts[:userinfo] = userinfo
+            end
+            config[:proxy_uri] = URI::HTTP.build(proxy_opts)
+          end
+
+          config[:use_accelerate_endpoint] = use_accelerate_endpoint?
+
+          # Add :session_token as a valid argument
+          [:access_key_id, :secret_access_key, :credential_provider, :credentials, :session_token].each do |opt|
```
akostadinov, line 25: **If I remember correctly**, `session_token` is the temporary thing that needs to be updated so we don't like to use it directly. I think there was a credentials class that will read token regularly from a file so we can just pass the `credentials` object and let it figure out when and how to read the token.

In this way no monkey patching would be needed:
https://github.com/thoughtbot/paperclip/blob/main/lib/paperclip/storage/s3.rb#L251

app/lib/cms/s3.rb
akostadinov: If there is a standard path for this token, then we can just detect it I guess.

lib/paperclip/storage/s3.rb
thalesmiguel: And I think you're right... Paperclip [instantiates Aws::S3::Resource](https://github.com/thoughtbot/paperclip/blob/c769382c9b7078f3d1620b50ec2a70e91ba62ec4/lib/paperclip/storage/s3.rb#L261) which, in turn, [instantiates Aws::S3::Client](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/S3/Client.html#initialize-instance_method). Just passing a hash inside of credentials with `:access_key_id`, `:secret_access_key` and `:session_token`, could be enough.
Let me check that out 👀

akostadinov: Actually it is in my JIRA comment. I believe you need to create a `AssumeRoleWebIdentityCredential` and pass it as the `credentials` parameter.

thalesmiguel: Following up on the tests:

- Passing `credentials` down to `Aws::S3::Resource` in the hopes to avoid the monkey patch
    ```ruby
    sts = Aws::STS::Client.new(CMS::S3.credentials.merge(region: CMS::S3.region))
    session = sts.get_session_token(duration_seconds: 15*60)
    resource = Aws::S3::Resource.new(credentials: session, region: CMS::S3.region)
    
    resource.client.get_bucket_tagging(bucket: 'edge-flexo.3scale.net.3scale.net') #=> Works
    ```

I was checking [Aws::STS::Client#assume_role_with_web_identity](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/STS/Client.html#assume_role_with_web_identity-instance_method) and, since it also returns the credentials, we should be OK to use it in the same fashion as the example above. Now, for us to do that, we'll need more params that I'm not really sure we have or where to get from. Namely `role_arn`, `role_session_name` and `web_identity_token`:
```ruby
resp = client.assume_role_with_web_identity({
  role_arn: "arnType", # required
  role_session_name: "roleSessionNameType", # required
  web_identity_token: "clientTokenType", # required
  provider_id: "urlType",
  policy_arns: [
    {
      arn: "arnType",
    },
  ],
  policy: "sessionPolicyDocumentType",
  duration_seconds: 1,
})
```
Any ideas on those?

akostadinov: I don't know about the other parameters. Hopefully whoever requested that, will know. The point of `AssumeRoleWebIdentityCredential` is that it should update automatically when token expires. Otherwise we will have to implement that somehow.

thalesmiguel: > The point of AssumeRoleWebIdentityCredential is that it should update automatically when token expires.

Yes. But from what I got from the documentation, this should also be true for many of the other methods (APIs) available in `Aws::STS::Client`, such as `assume_role` and `get_session_token`. What's still unclear to me is how the "mounted file that contains signed token(which is automatically rotated every few minutes)" (described in the ticket) would play a part in here 🤔

akostadinov: I haven't looked now, from my memory only `AssumeRoleWebIdentityCredential` reads token from a file the way feature is described. All other options need  other credentials or don't update automatically when token changed.

thalesmiguel: And you're right yet again!
I just double checked the [AssumeRoleWebIdentityCredentials](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/AssumeRoleWebIdentityCredentials.html) **class** and it accepts a `web_identity_token_file` param in its initializer that sounds closer to what the ticket says. I'll keep this one close to mind.
Thank you!

thalesmiguel:
I just pushed 2 new commits to this branch:

- 370e1cd to revert the first assumptions on this. Turns out we won't need a monkey patch as Paperclip already supports a `credentials` argument;
- c9a57b8 to implement a call to `Aws::AssumeRoleWebIdentityCredentials` using the `web_identity_token_file` and `role_name` arguments that should be provided on a configuration file on OpenShift deployments.

Next steps:

1. Figure out where to mount the necessary volume that stores the file we need;
2. Find a way to manually test this integration with a proper credentials file;

PR #3063, opened 2022-09-15
THREESCALE-7540 fix master admin tenant name issue

nidhi-soni1104:

 This issue is for tenant (account in master portal) and not for account in admin portal. Because tenant URL is created from tenant name hence those special characters are forbidden.

https://issues.redhat.com/browse/THREESCALE-7540

**Verification steps** 

mentioned in Jira

Note for reviewer:
Validation was already there but error message is not shown so I have added. 
image

### DIFF
--- a/app/lib/signup/domains_builder.rb
+++ b/app/lib/signup/domains_builder.rb
@@ -15,7 +15,7 @@ module Signup
     end
 
     def generate_subdomain
-      result = org_name.tr("_", "-").to_s.parameterize
+      result = org_name.to_s.tr('_', '-').parameterize
 
       while result.present? && invalid_subdomain_condition.call(subdomain: result)
         base, sequence = result.match(/\A(.*?)(?:\-(\d+))?\z/).to_a[1, 2]
--- a/app/lib/signup/domains_builder.rb
+++ b/app/lib/signup/domains_builder.rb
@@ -15,7 +15,7 @@ module Signup
     end
 
     def generate_subdomain
-      result = org_name.to_s.parameterize
+      result = org_name.gsub("_", "-").to_s.parameterize
 
       while result.present? && invalid_subdomain_condition.call(subdomain: result)
         base, sequence = result.match(/\A(.*?)(?:\-(\d+))?\z/).to_a[1, 2]
--- a/app/lib/signup/domains_builder.rb
+++ b/app/lib/signup/domains_builder.rb
@@ -15,7 +15,7 @@ module Signup
     end
 
     def generate_subdomain
-      result = org_name.gsub("_", "-").to_s.parameterize
+      result = org_name.tr("_", "-").to_s.parameterize
 
       while result.present? && invalid_subdomain_condition.call(subdomain: result)
         base, sequence = result.match(/\A(.*?)(?:\-(\d+))?\z/).to_a[1, 2]
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -260,8 +260,9 @@ class Account < ApplicationRecord
 
   # TODO: multitenant. enable it?
   # validates_uniqueness_of :s3_prefix
-  validates :org_name, format: { with: /\A.*[a-zA-Z0-9]+.*\z/, message: :invalid_format },
-                       presence: true, length: { maximum: 255 }
+  validates :org_name, presence: true, length: { maximum: 255 },
+                       format: { with: /\A.*[a-zA-Z0-9]+.*\z/, message: :invalid_format }
+
 
   validates :org_legaladdress, :domain, :telephone_number, :site_access_code,
             :billing_address_name, :billing_address_address1, :billing_address_address2, :billing_address_city,
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -167,7 +167,7 @@ class Account < ApplicationRecord
   has_many :sso_authorizations, through: :users
   has_many :user_sessions, through: :users
 
-  validates_format_of :org_name, with: /\A.*[a-zA-Z0-9]+.*\z/
+  validates :org_name, format: /\A.*[a-zA-Z0-9]+.*\z/
 
   alias_attribute :name, :org_name
 
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -167,7 +167,7 @@ class Account < ApplicationRecord
   has_many :sso_authorizations, through: :users
   has_many :user_sessions, through: :users
 
-  validates :org_name, format: /\A.*[a-zA-Z0-9]+.*\z/
+  validates :org_name, format: { with: /\A.*[a-zA-Z0-9]+.*\z/, message: :invalid_format }
 
   alias_attribute :name, :org_name
 
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -167,8 +167,6 @@ class Account < ApplicationRecord
   has_many :sso_authorizations, through: :users
   has_many :user_sessions, through: :users
 
-  validates :org_name, format: { with: /\A.*[a-zA-Z0-9]+.*\z/, message: :invalid_format }
-
   alias_attribute :name, :org_name
 
   has_one :onboarding
@@ -262,7 +260,8 @@ class Account < ApplicationRecord
 
   # TODO: multitenant. enable it?
   # validates_uniqueness_of :s3_prefix
-  validates :org_name, presence: true, length: { maximum: 255 }
+  validates :org_name, format: { with: /\A.*[a-zA-Z0-9]+.*\z/, message: :invalid_format },
+                       presence: true, length: { maximum: 255 }
 
   validates :org_legaladdress, :domain, :telephone_number, :site_access_code,
             :billing_address_name, :billing_address_address1, :billing_address_address2, :billing_address_city,
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -167,6 +167,8 @@ class Account < ApplicationRecord
   has_many :sso_authorizations, through: :users
   has_many :user_sessions, through: :users
 
+  validates_format_of :org_name, with: /\A.*[a-zA-Z0-9]+.*\z/
+
   alias_attribute :name, :org_name
 
   has_one :onboarding
--- a/app/views/provider/admin/accounts/new.html.slim
+++ b/app/views/provider/admin/accounts/new.html.slim
@@ -1,6 +1,11 @@
 h2 Create new Account
 = semantic_form_for @provider, builder: Fields::FormBuilder, url: provider_admin_accounts_path do |form|
 
+  - if @provider.errors.any?
+    ul style = "color: red"
+      - @provider.errors.full_messages.each do |message|
+        li= h message
+
   = form.inputs name: 'User Information', for: [:user, @user] do |user|
     = user.user_defined_form
     = user.input :password, required: true
--- a/app/views/provider/admin/accounts/new.html.slim
+++ b/app/views/provider/admin/accounts/new.html.slim
@@ -1,11 +1,5 @@
 h2 Create new Account
 = semantic_form_for @provider, builder: Fields::FormBuilder, url: provider_admin_accounts_path do |form|
-
-  - if @provider.errors.any?
-    ul style = "color: red"
-      - @provider.errors.full_messages.each do |message|
-        li= h message
-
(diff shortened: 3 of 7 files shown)


app/views/provider/admin/accounts/new.html.slim
```diff
@@ -1,6 +1,11 @@
 h2 Create new Account
 = semantic_form_for @provider, builder: Fields::FormBuilder, url: provider_admin_accounts_path do |form|
 
+  - if @provider.errors.any?
+    ul style = "color: red"
```
mayorova, line 5: We shouldn't be adding custom error messages, but rather use the default form validation styles.

validation

Apparently, the validation here is on the internal field (`domain` and `self_domain` I guess), that's why the highlighting doesn't work.

mayorova (changes requested):
@nidhi-soni1104 After investigating this issue further, I think that rather than showing the validation error (it would be quite tricky in this case because it is related to the internal `subdomain` field, and we probably don't want to add a validation for the underscore character in the Org Name) we should just replace `_` with `-` when generating a subdomain from the Org Name, as it is done for other special characters.

Please see the full description with links to code in [this JIRA comment](https://issues.redhat.com/browse/THREESCALE-7540?focusedCommentId=20961524&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-20961524).

Also, we would need to have some tests that cover different scenarios of provider creation with special characters (including underscore) in Org Name.

nidhi-soni1104:
> @nidhi-soni1104 After investigating this issue further, I think that rather than showing the validation error (it would be quite tricky in this case because it is related to the internal `subdomain` field, and we probably don't want to add a validation for the underscore character in the Org Name) we should just replace `_` with `-` when generating a subdomain from the Org Name, as it is done for other special characters.
> 
> Please see the full description with links to code in [this JIRA comment](https://issues.redhat.com/browse/THREESCALE-7540?focusedCommentId=20961524&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-20961524).
> 
> Also, we would need to have some tests that cover different scenarios of provider creation with special characters (including underscore) in Org Name.

@mayorova Agreed what you explained, so you want to add similar  validation for org_name in account model. 
The issue mentioned in Jira is to show error if it fails to create account and So I did same. Let see what they say more after your explanation on the JIRA

app/views/provider/admin/accounts/new.html.slim
nidhi-soni1104: @mayorova  This issue is for tenant (account in master portal) and not for account in admin portal. The screenshot you attached is from admin portal?

mayorova: @nidhi-soni1104 The screenshot is from the master portal, the form creating a tenant.

mayorova:
@nidhi-soni1104 No, actually quite the opposite, rather than adding a validation on Org Name field, I suggest to accept underscores in the Org Name, and transform them to `-` for generating the subdomain (the same way it is done for all other special characters)

nidhi-soni1104:
@mayorova changes done from my side

nidhi-soni1104:
> @nidhi-soni1104 I would suggest to add a meaningful message on Org Name `invalid` error to `en.yml`, e.g. `"should contain at least one alphanumeric character"`
> 
> Maybe a screenshot of the form with the new validation message in the PR would be useful.

Sure Let me update PR with message and screenshot

nidhi-soni1104:
@mayorova Validation is updated with message and below is the screenshot.

image

Also here is the screenshot of organization valid names


image

test/unit/signup/result_test.rb
```diff
@@ -105,7 +105,8 @@ class ResultTest < ActiveSupport::TestCase
       refute user.persisted?
       refute account.persisted?
       assert_includes signup_result.errors[:account], 'Organization/Group Name can\'t be blank'
-      assert_equal 1, signup_result.errors.full_messages.length
+      assert_includes signup_result.errors[:account], 'Organization/Group Name invalid'
+      assert_equal 2, signup_result.errors.full_messages.length
```
mayorova, line 109: maybe makes sense to validate the error message text also?

nidhi-soni1104: Umm I have added both check we can remove also .I Have added so one should know like which are the two failing assertions.

nidhi-soni1104: No I think because we have now added a custom message in en.yml and every time update that message this test will fail and we have to update here also which doesn't makes sense. 
I am updating and removing this assertion. Thanks for pointing this out.

nidhi-soni1104: @mayorova changes done from my side

app/lib/signup/domains_builder.rb
```diff
@@ -15,7 +15,7 @@ def generate
     end
 
     def generate_subdomain
-      result = org_name.to_s.parameterize
+      result = org_name.tr("_", "-").to_s.parameterize
```
mayorova, line 18:
```suggestion
      result = org_name.to_s.tr('_', '-').parameterize
```

nidhi-soni1104: Let me check this on my  local then I will commit this suggestion

test/unit/signup/result_test.rb
mayorova: But I don't think just checking the length of the errors is meaningful enough. What errors are they? Are they the ones we are expecting?

The fact that a change in `en.yml` make the test fail - I'd say it is good. If we accidentally or erroneously modify something in the i18n file - the test will capture it.

mayorova: I'm suggesting to add the following (fixing the current test and adding a new one):
```suggestion
    test '#save does not save and #errors return the errors when the account is blank' do
      @account = FactoryBot.build(:account, org_name: nil)
      signup_result.save
      refute user.persisted?
      refute account.persisted?
      assert_includes signup_result.errors[:account], "Organization/Group Name can't be blank"
    end

    test '#save does not save and #errors return the errors when the account is invalid' do
      @account = FactoryBot.build(:account, org_name: '?-!_$')
      signup_result.save
      refute user.persisted?
      refute account.persisted?
      assert_includes signup_result.errors[:account], "Organization/Group Name should contain at least one alphanumeric character"
    end
```

test/unit/account_test.rb
```diff
@@ -746,6 +746,16 @@ def test_smart_destroy_master
...
   end
+
+  test "allow underscore in org_name" do
+    account = FactoryBot.build(:simple_account, org_name: 'name@._+1')
+    assert account.valid?
+  end
+
+  test "there should be at least one alphanum character anywhere in the org_name" do
+    account = FactoryBot.build(:simple_account, org_name: '.?+@')
+    assert_not account.valid?
```
mayorova, line 757: I would also add this check here:
```
assert_includes account.errors.messages[:org_name], "should contain at least one alphanumeric character"
```

mayorova (changes requested):
I'm also missing a test to validate that the underscores are correctly replaced with `-`.
There is a test `test/unit/account/provider_domains_test.rb` which seems to be a good fit for this

test/unit/signup/result_test.rb
jlledom: I think writing two tests and checking for the two messages is OK, but then we would need to add tests for other fields like username and email. But maybe that's out of scope in this PR

nidhi-soni1104: > But I don't think just checking the length of the errors is meaningful enough. What errors are they? Are they the ones we are expecting?
> 
> The fact that a change in `en.yml` make the test fail - I'd say it is good. If we accidentally or erroneously modify something in the i18n file - the test will capture it.
Agreed

nidhi-soni1104: > I'm suggesting to add the following (fixing the current test and adding a new one):
> 
> ```
>     test '#save does not save and #errors return the errors when the account is blank' do
>       @account = FactoryBot.build(:account, org_name: nil)
>       signup_result.save
>       refute user.persisted?
>       refute account.persisted?
>       assert_includes signup_result.errors[:account], "Organization/Group Name can't be blank"
>     end
> 
>     test '#save does not save and #errors return the errors when the account is invalid' do
>       @account = FactoryBot.build(:account, org_name: '?-!_$')
>       signup_result.save
>       refute user.persisted?
>       refute account.persisted?
>       assert_includes signup_result.errors[:account], "Organization/Group Name should contain at least one alphanumeric character"
>     end
> ```

This can be also good idea

test/unit/account_test.rb
nidhi-soni1104: Okay if we are adding there we can add here to so if it fails it will fail together

nidhi-soni1104:
> test/unit/account/provider_domains_test.rb
Yes we can add that
which test in test/unit/account/provider_domains_test.rb?

test/unit/signup/result_test.rb
nidhi-soni1104: > I think writing two tests and checking for the two messages is OK, but then we would need to add tests for other fields like username and email. But maybe that's out of scope in this PR

yes they are not something related to my change

app/models/account.rb
```diff
@@ -167,6 +167,8 @@ def schedule_backend_sync_worker
   has_many :sso_authorizations, through: :users
   has_many :user_sessions, through: :users
 
+  validates :org_name, format: { with: /\A.*[a-zA-Z0-9]+.*\z/, message: :invalid_format }
```
mayorova, line 170: By the way, I've just noticed there is another validation for `:org_name` on line 265:
```
validates :org_name, presence: true, length: { maximum: 255 }
```

I think we need to either merge the two validations, or at least to put them together, because it's hard to see then the complete set of rules for this field.

nidhi-soni1104: Yes may  be we can merge them
Let me check and update the PR

test/unit/account_test.rb
nidhi-soni1104: Done changes so resolving

test/unit/signup/result_test.rb
nidhi-soni1104: Updated so resolving

mayorova (approved):

app/lib/signup/domains_builder.rb
```diff
@@ -15,7 +15,7 @@ def generate
     end
 
     def generate_subdomain
-      result = org_name.to_s.parameterize
+      result = org_name.to_s.tr('_', '-').parameterize
 
       while result.present? && invalid_subdomain_condition.call(subdomain: result)
         base, sequence = result.match(/\A(.*?)(?:\-(\d+))?\z/).to_a[1, 2]
```
akostadinov, line 21: should we use `Authentication::RE_DOMAIN_OK` instead?

app/models/account.rb
```diff
@@ -260,7 +260,8 @@ def bought?(plan)
 
   # TODO: multitenant. enable it?
   # validates_uniqueness_of :s3_prefix
-  validates :org_name, presence: true, length: { maximum: 255 }
+  validates :org_name, format: { with: /\A.*[a-zA-Z0-9]+.*\z/, message: :invalid_format },
```
akostadinov, line 263: `Authentication::RE_DOMAIN_OK`?

akostadinov (changes requested):

app/lib/signup/domains_builder.rb
mayorova: Not sure how different these regexes are, but I think we don't need to touch it in this PR.

mayorova:
@nidhi-soni1104 Looks good to me :+1: I'll open another JIRA for the length validation.

app/lib/signup/domains_builder.rb
akostadinov: I didn't think much before writing. Yeah, it has some grouping and stuff that obviously will not work OOB. So we can merge with this.

akostadinov (dismissed):

mayorova (changes requested):
Sorry, I spotted another potential problem.

When the org name is converted to subdomain, we need to ensure that the subdomain is not starting or ending with special characters.

UPDATE: Nevermind, `.parameterize` takes care for that :wink:

mayorova (approved):

akostadinov (approved):

PR #3064, opened 2022-09-15
Test N+1 in /buyers/accounts index

jlledom:

Adds a test to check N+1 is solved in the `/buyers/accounts` screen

[THREESCALE-8530](https://issues.redhat.com/browse/THREESCALE-8530)

Not sure if I did this correctly, I filtered by table `users` and requested the `json` format to reduce  the number of observed queries. Also set the `slope` to 3 because that's the number of queries for each account that I observed in previous tests.

### DIFF
--- a/app/decorators/account_decorator.rb
+++ b/app/decorators/account_decorator.rb
@@ -3,7 +3,23 @@
 class AccountDecorator < ApplicationDecorator
   delegate :display_name, :email, to: :admin_user, prefix: true
 
+  def first_admin
+    @first_admin ||= (first_admin_cached || super)
+  end
+
   private
+
+  def first_admin_cached
+    @first_admin_cached ||= admins&.first
+  end
+
+  def admins
+    return nil unless users.loaded?
+
+    @admins ||= users.select do |user|
+      user.role == :admin && user.username != ThreeScale.config.impersonation_admin['username']
+    end
+  end
 
   def admin_user
     @admin_user ||= (super || User.new).decorate
--- a/app/decorators/account_decorator.rb
+++ b/app/decorators/account_decorator.rb
@@ -4,22 +4,11 @@ class AccountDecorator < ApplicationDecorator
   delegate :display_name, :email, to: :admin_user, prefix: true
 
   def first_admin
-    @first_admin ||= (first_admin_cached || super)
+    # when users are eager loaded, filter manually to save an extra query
+    @first_admin ||= users.loaded? ? users.find { |user| user.admin? && !user.impersonation_admin? } : super
   end
 
   private
-
-  def first_admin_cached
-    @first_admin_cached ||= admins&.first
-  end
-
-  def admins
-    return nil unless users.loaded?
-
-    @admins ||= users.select do |user|
-      user.role == :admin && user.username != ThreeScale.config.impersonation_admin['username']
-    end
-  end
 
   def admin_user
     @admin_user ||= (super || User.new).decorate
--- a/app/decorators/account_decorator.rb
+++ b/app/decorators/account_decorator.rb
@@ -4,8 +4,8 @@ class AccountDecorator < ApplicationDecorator
   delegate :display_name, :email, to: :admin_user, prefix: true
 
   def first_admin
-    # when users are eager loaded, filter manually to save an extra query
-    @first_admin ||= users.loaded? ? users.find { |user| user.admin? && !user.impersonation_admin? } : super
+    # when admin_users is eager loaded, filter manually to save an extra query
+    @first_admin ||= admin_users.loaded? ? admin_users.find { |user| !user.impersonation_admin? } : super
   end
 
   private
--- a/app/decorators/account_decorator.rb
+++ b/app/decorators/account_decorator.rb
@@ -3,11 +3,6 @@
 class AccountDecorator < ApplicationDecorator
   delegate :display_name, :email, to: :admin_user, prefix: true
 
-  def first_admin
-    # when admin_users is eager loaded, filter manually to save an extra query
-    @first_admin ||= admin_users.loaded? ? admin_users.find { |user| !user.impersonation_admin? } : super
-  end
-
   private
 
   def admin_user
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -140,7 +140,7 @@ class User < ApplicationRecord
 
   scope :latest, -> {limit(5).order(created_at: :desc)}
 
-  scope :but_impersonation_admin, -> { where(['username <> ?', ThreeScale.config.impersonation_admin['username']]) }
+  scope :but_impersonation_admin, -> { where(['users.username <> ?', ThreeScale.config.impersonation_admin['username']]) }
   scope :impersonation_admins, -> { where(username: ThreeScale.config.impersonation_admin['username']) }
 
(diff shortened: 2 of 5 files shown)


test/integration/buyers/accounts_controller_test.rb
```diff
@@ -391,4 +391,18 @@ class NotLoggedInTest < ActionDispatch::IntegrationTest
...
+
+  class ManyAccountsTest < ActionDispatch::IntegrationTest
+    def setup
+      @provider = FactoryBot.create(:provider_account)
+      login! @provider
+    end
+
+    test 'no N+1 problems in the index' do
+      populate = ->(n) { FactoryBot.create_list(:buyer_account, n, provider_account: @provider) }
+      assert_perform_linear_number_of_queries(slope: 3, populate: populate, matching: /users/) do
```
akostadinov, line 403: Shouldn't we `assert_perform_constant_number_of_queries`? Why aren't all buyers queried by a single query?

Just asking, I didn't look at the code. Just this is how I would expect endpoint to work.

akostadinov:
While on it, could you include JIRA issue in commit comment? This helps when looking at git history without need to go to and rely on github.

test/integration/buyers/accounts_controller_test.rb
akostadinov: btw in the past I had to use `warmup` to avoid issues caused by caching or who knows what.

jlledom: I think you're right and I misunderstood this. But I don't understand why the amount of queries to the users table keeps incrementing linearly after solving the N+1 and bullet stopped complaining. It's possible that there are more N+1 problems in this screen not detected by bullet?. I'll investigate further.

jlledom:
I found the problem here. There was another N+1 in that screen, not detected by Bullet, at this line:

[app/views/buyers/accounts/_search_results.html.slim#L58](https://github.com/3scale/porta/blob/0e946cc325b60719fc3eff9820174b39da278765/app/views/buyers/accounts/_search_results.html.slim#L58)

Accessing `first_admin` filters by scopes `admin` and `but_impersonation_admin`. Since the scopes are defined as a `where` clause, a new SQL query is executed every time `first_admin` is called.

To fix this, I modified the account decorator to apply the filter manually when the users association happens to be already loaded, avoiding the extra query, and then including the association in the main query in the account index presenter.

app/models/user.rb
```diff
@@ -140,7 +140,7 @@ def self.search_states
 
   scope :latest, -> {limit(5).order(created_at: :desc)}
 
-  scope :but_impersonation_admin, -> { where(['username <> ?', ThreeScale.config.impersonation_admin['username']]) }
+  scope :but_impersonation_admin, -> { where(['users.username <> ?', ThreeScale.config.impersonation_admin['username']]) }
```
akostadinov, line 143: why is this needed?

app/decorators/account_decorator.rb
```diff
@@ -3,8 +3,24 @@
...
+  def first_admin_cached
+    @first_admin_cached ||= admins&.first
+  end
+
+  def admins
+    return nil unless users.loaded?
+
+    @admins ||= users.select do |user|
+      user.role == :admin && user.username != ThreeScale.config.impersonation_admin['username']
+    end
```
akostadinov, line 21: maybe add a method `#admin?` for the `Account`

app/decorators/account_decorator.rb
```diff
@@ -3,8 +3,24 @@
 class AccountDecorator < ApplicationDecorator
   delegate :display_name, :email, to: :admin_user, prefix: true
 
+  def first_admin
+    @first_admin ||= (first_admin_cached || super)
+  end
```
akostadinov, line 8:
```suggestion
  def first_admin
    @first_admin ||= users.loaded? ? users.find { |user| user.admin? && user.impersonation_admin? }  : super
  end
```

Where `user.impersonation_admin?` needs to be implemented in `User`.

akostadinov (changes requested):

app/models/user.rb
jlledom: I found some tests failing because username was an ambiguous column name: https://app.circleci.com/pipelines/github/3scale/porta/21537/workflows/052cd6c3-150c-44f3-a943-a83579956da4/jobs/249337

app/decorators/account_decorator.rb
jlledom: I used this code, and I also esu the `:admin_users` association instead of `:users`, since `first_admin` only searches among admins.

app/models/user.rb
akostadinov:
```suggestion
  scope :but_impersonation_admin, -> { where.has { username <> ThreeScale.config.impersonation_admin['username'] }}
```
What I'm concerned with is hardcoding table name might be a problem with more complicated joins or sub-queries. I'm not sure that's a real risk but this is my reasoning :man_shrugging:

app/presenters/buyers/accounts_index_presenter.rb
```diff
@@ -14,7 +14,7 @@ def initialize(provider:, params: {})
 
   def buyers
     @buyers ||= provider.buyer_accounts
-                        .includes([:admin_user])
+                        .includes(%i[admin_users admin_user])
```
akostadinov, line 17: why is `admin_user` not enough?

app/models/user.rb
jlledom: done

app/presenters/buyers/accounts_index_presenter.rb
jlledom: I think it's enough, and all the first_admin additional logic was redundant, I removed it and the n+1 test passes

app/presenters/buyers/accounts_index_presenter.rb
```diff
@@ -14,7 +14,7 @@ def initialize(provider:, params: {})
 
   def buyers
     @buyers ||= provider.buyer_accounts
-                        .includes([:admin_user])
+                        .includes(%i[admin_user])
```
akostadinov, line 17: Everything looks awesome now. Just a nitpick. I prefer such minor style changes to stay out because they make git blame more complicated. I don't know whether you use some special tool to help with that.

test/integration/buyers/accounts_controller_test.rb
```diff
@@ -391,4 +391,18 @@ class NotLoggedInTest < ActionDispatch::IntegrationTest
...
+
+  class ManyAccountsTest < ActionDispatch::IntegrationTest
+    def setup
+      @provider = FactoryBot.create(:provider_account)
+      login! @provider
+    end
+
+    test 'no N+1 problems in the index' do
+      populate = ->(n) {FactoryBot.create_list(:buyer_account, n, provider_account: @provider)}
+      assert_perform_constant_number_of_queries(populate: populate, matching: /users/) do
```
akostadinov, line 403: do we need the `matching` argument?

jlledom: I think so, I'm using it because the N+1 problem was querying the users table. If I remove it, then it fails b/c the number of queries to the plans table also increments linearly

akostadinov (approved):
awesome work!

please squash when you decide to merge

PR #3065, opened 2022-09-15
use libgd instead of ImageMagick

akostadinov:
**Note:** DEPLOY ONLY AFTER COORDINATION WITH PRODUCT TEAM!

In RHEL8+ ImageMagick is no longer provided due to being hard
to maintain and prone to security issues.

That's why we implement an image processor based on the `libgd` backed
`local-fastimage_resize` gem for logo processing instead of the
Paperclip's default ImageMagick based processor `Paperclip::Thumbnail`.

We remove support for uploading GIF profile logos in the process.
Logo is used when generating invoices. But prawn only supports JPEGs
and PNGs.

With ImageMagick we could specify target format for the `:invoice`
style to be `png` regardless of source image format.
`local-fastimage_resize` only resizes between same formats though.

Provided very low adoption of GIF logos, it is not worth implementing
such support for `local-fastimage_resize` although it shouldn't be too
hard if need arise.

<del>Note: existing attachments don't need to be processed because the
`:invoice` style will already be PNG. Make sure NOT to re-process all
attachments though. I don't see a reason to re-process them.</del>
Note: existing **png** and **gif** attachments don't need to be processed because the
`:invoice` style will already be PNG. **JPEGs** will need though as style for them changes.

fixes [THREESCALE-8579](https://issues.redhat.com/browse/THREESCALE-8579)


jlledom (approved):

test/integration/provider/admin/account/logos_controller_test.rb
```diff
@@ -85,7 +107,7 @@ def logo_file
     @logo_file ||= fixture_file_upload(file_fixture('hypnotoad.jpg'))
   end
 
-  def countries_yaml_file
-    @countries_yaml_file ||= fixture_file_upload(file_fixture('countries.yml'))
+  def small_logo_file
+    @logo_file ||= fixture_file_upload(file_fixture('small.png'))
```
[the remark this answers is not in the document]
akostadinov: copy paste error :)
Thanks!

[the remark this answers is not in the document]
josemigallas:
```suggestion
    @small_logo_file ||= fixture_file_upload(file_fixture('small.png'))
```

INSTALL.md
```diff
@@ -237,7 +237,7 @@ dnf module install nodejs:12
 #### Dependencies
 
 ```shell
-sudo yum install patch autoconf automake bison libffi-devel libtool libyaml-devel readline-devel sqlite-devel zlib-devel openssl-devel ImageMagick ImageMagick-devel mysql-devel postgresql-devel chromedriver memcached
+sudo yum install patch autoconf automake bison libffi-devel libtool libyaml-devel readline-devel sqlite-devel zlib-devel openssl-devel gd-devel mysql-devel postgresql-devel chromedriver memcached
```
josemigallas, line 240: We should update macOS setup steps. I can help you test it.

josemigallas (changes requested):

INSTALL.md
akostadinov: Good point!

I see two lines: 
```
brew install imagemagick@6 mysql@5.7 gs pkg-config openssl geckodriver postgresql memcached
brew link imagemagick@6 --force
```

I think we need to replace `imagemagick@6` with `gd`. But doe we need `brew-link`? Could you try this out?

akostadinov: I updated instructions for brew. Test, pretty please :pleading_face:

test/unit/tasks/fixes_test.rb
```diff
@@ -0,0 +1,24 @@
...
+      profile.logo.post_processing = false
+      file_fixture("hypnotoad.jpg").open("rb") { |io| profile.logo = io }
+      profile.save!
+
+      tmpfile = Tempfile.new(["logo-", ".jpg"])
+      assert_raises(Errno::ENOENT) { profile.logo.copy_to_local_file :invoice, tmpfile.path }
+
+      execute_rake_task 'fixes.rake', 'fixes:regenerate_jpeg_invoice_logo'
+      profile.logo.copy_to_local_file :invoice, tmpfile.path
+    end
```
jlledom, line 16: I would add a new test to ensure it's not reprocessing gifs and pngs

jlledom (approved):

test/unit/tasks/fixes_test.rb
akostadinov: done

jlledom (approved):

mayorova:
I guess we also need to add `gd` to the [Dockerfile](https://github.com/3scale/porta/blob/master/openshift/system/Dockerfile) ?

mayorova:
There is also a conflict in `INSTALL.md`

app/models/profile.rb
```diff
@@ -42,14 +42,16 @@ class Profile < ApplicationRecord
   belongs_to :account
   delegate :s3_provider_prefix, to: :account
 
+  LOGO_STYLES = ->(logo) { {large: '300x300>'.freeze, medium: '150x150>'.freeze, thumb: '100x100>'.freeze, invoice: (logo.content_type.end_with?("gif") ? ['200x50>'.freeze, :png].freeze : '200x50>'.freeze)}.freeze }
```
mayorova, line 45: For the sake of readability and style, maybe make this multi-line?
Also, maybe remove `.freeze` and add `# frozen_string_literal: true`?
Finally, maybe use single quotes for strings.

mayorova:
Maybe we could do something to improve UX, in this branch uploading a `.gif` file shows
logo-content-type-invalid

Maybe we could add some message saying which formats are supported (especially as we are dropping gifs).

Or probably we can leave it for another PR, or even leave it as it is - as at the moment for all other non-supported types of logos we are showing the same error.

app/models/profile.rb
akostadinov: I don't want to change whole profile. We have a number of files without it. To reduce scope of possible issues.

akostadinov: made multi-line

which single quote do you mean?

mayorova: I meant more like:
```
  LOGO_STYLES = ->(logo) {
    {
      large: '300x300>'.freeze,
      medium: '150x150>'.freeze,
      thumb: '100x100>'.freeze,
      invoice: (logo.content_type.end_with?('gif') ? ['200x50>'.freeze, :png].freeze : '200x50>'.freeze)
    }.freeze
  }
```

Just easier to read what styles are available and what are their sizes.
Re: quotes I meant `"gif"` => `'gif'`, for consistency.

INSTALL.md
```diff
@@ -240,8 +239,9 @@ asdf install
 
 #### Dependencies
 
+<<<<<<< HEAD
```
mayorova, line 242: :eyes:

lib/tasks/fixes.rake
```diff
@@ -113,6 +113,17 @@ namespace :fixes do
     end
   end
 
+  desc "Regenerate JPEG logos :invoice style"
+  task :regenerate_jpeg_invoice_logo => :environment do
+    Profile.where(logo_content_type: "image/jpeg").find_each do |profile|
+      begin
+        profile.logo.reprocess!(:invoice)
```
mayorova, line 120: I am wondering if we are affected by this bug: https://github.com/thoughtbot/paperclip/pull/2637

I have tested this successfully though, so not sure if this may cause any issues. But this is to be aware of for SaaS migration.

mayorova:
Great job @akostadinov ! :clap: 
LGTM :+1:

lib/tasks/fixes.rake
```diff
@@ -113,6 +113,17 @@ namespace :fixes do
     end
   end
 
+  desc "Regenerate JPEG logos :invoice style"
+  task :regenerate_jpeg_invoice_logo => :environment do
+    Profile.where(logo_content_type: "image/jpeg").find_each do |profile|
```
nidhi-soni1104, line 118: What about png extension?

akostadinov: not needed

app/models/profile.rb
akostadinov: makes sense, thanks!

lib/tasks/fixes.rake
akostadinov: I think we should be fine. Since we are not running an external process like it was done by ImageMagic, I think in case of a trouble, the process will just fail. Also if target style only is truncated, that shouldn't be an issue.

But I will add as a precaution to backup the S3 bucket beforehand. Good catch!

akostadinov: I managed to test this by editing `gems/paperclip-6.1.0/lib/paperclip/storage/s3.rb` and making sure an empty local file is being generated when copying the file locally.

The bug is present - paperclip overwrites the target image with an empty file.

But we have nothing to worry because:
1. it only overwrites the styles being reprocessed, this means that original logo will not be affected
2. because the `invoice` style will be missing anyway, overwriting it with a zero byte file will not matter (no data will be lost)

What we have to do would be to check for any broken invoice logos after running the task.

Gemfile
```diff
@@ -94,9 +94,11 @@ gem 'baby_squeel', '~> 1.4.3'
 gem 'ransack', '2.4.1' # we can remove line when stop using ruby 2.4
 gem 'browser', '~> 5.0.0' # we can update to lts when we stop using ruby 2.4
 gem 'diff-lcs', '~> 1.2'
+gem 'fastimage'
 gem 'hiredis', '~> 0.6.3'
 gem 'httpclient', github: 'mikz/httpclient', branch: 'ssl-env-cert'
 gem 'json-schema', git: 'https://github.com/3scale/json-schema.git'
+gem 'local-fastimage_resize', require: 'fastimage/resize'
```
thalesmiguel, line 101: We're missing the minimum gem versions here.

app/models/profile.rb
```diff
@@ -117,6 +126,10 @@ def fix_http str
 
   delegate :provider_id_for_audits, :to => :account, :allow_nil => true
 
+  def will_save_change_to_logo?
+    logo.dirty?
```
thalesmiguel, line 130: I'm not entirely sure about the logic behind this `dirty?` method but, by checking the usage in this class, couldn't we use `ActiveRecord::AttributeMethods::Dirty`'s `has_changes_to_save?` or even `will_save_change_to_attribute?(:logo)`?

lib/paperclip/gd_image_processor.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+module Paperclip
+  class GDImageProcessor < Processor
+    def make
+      original_geometry = Geometry.new(*FastImage.size(file))
+      computed_geometry = original_geometry.resize_to(target_geometry_str)
+      file.rewind
+      original_geometry.to_s[/\d+x\d+/] == computed_geometry.to_s[/\d+x\d+/] ? temp_copy(file) : FastImage.resize(file, computed_geometry.width, computed_geometry.height)
```
thalesmiguel, line 9: This line got big. Aren't we using some linter to standardise some style guides?

test/unit/pdf/finance/invoice_generator_test.rb
```diff
@@ -27,18 +25,27 @@ class Pdf::Finance::InvoiceGeneratorTest < ActiveSupport::TestCase
...
+  test 'should generate valid PDF content with jpg logo and line items' do
+    test_generate_with_logo_and_lines(file_fixture("wide.jpg").to_s)
+  end
+
+  test 'should generate valid PDF content with png logo and line items' do
+    test_generate_with_logo_and_lines(file_fixture("wide.png").to_s)
+  end
+
+  private
+  def test_generate_with_logo_and_lines(image_path)
```
thalesmiguel, line 37: Another one related with style guides: We're missing a blank line after `private`.

thalesmiguel (commented):
Just left a couple comments there :v:

Gemfile
akostadinov: We don't really have a minimum as long as it works. But it is also in `Gemfile.lock`. I don't see how we can go below that.

app/models/profile.rb
```diff
@@ -42,14 +42,23 @@ class Profile < ApplicationRecord
...
   delegate :s3_provider_prefix, to: :account
 
+  LOGO_STYLES = ->(logo) {
+    {
+      large: '300x300>'.freeze,
+      medium: '150x150>'.freeze,
+      thumb: '100x100>'.freeze,
+      invoice: (logo.content_type.end_with?('gif') ? ['200x50>'.freeze, :png].freeze : '200x50>'.freeze)
+    }.freeze
+  }
```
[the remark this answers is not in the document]
akostadinov: We can do later. I don't want to add this to the file to avoid something unexpected. Other than that I agree. We have a lot of files without it.

app/models/profile.rb
akostadinov: `logo` is not a normal attribute. We set some DB fields and paperclip uses them to process `logo`.

https://stackoverflow.com/questions/9029002/how-to-detect-if-a-paperclip-attachment-was-changed-in-after-save-callback

lib/paperclip/gd_image_processor.rb
akostadinov: We've been using long lines in porta. We can discuss with team changing our style if you see issues with that.

test/unit/pdf/finance/invoice_generator_test.rb
akostadinov: I think that the only way to enforce such requirement is to configure code-climate to complain about it. We don't have this consistent through codebase. It also makes sense to me to require an empty line. But I have some recollection that actually codeclimate complains about having it.

btw same with long lines. If we want this style, we need to setup codeclimate. I'm fine with long lines though in general.

```
  coffeelint:
    enabled: true
    checks:
      max_line_length:
        enabled: false
```

rubocop:
```
Layout/LineLength:
  Max: 99
  Enabled: false

Layout/EmptyLinesAroundAccessModifier:
  Enabled: false
```

I don't know historic reasons for these. As I said we can always discuss.

lib/paperclip/gd_image_processor.rb
thalesmiguel: I agree to bring this kind of code style discussions to the team. After checking some of the code in this repo, I can see only improvements coming from a discussed/agreed style guide.

test/unit/pdf/finance/invoice_generator_test.rb
thalesmiguel: There's a plugin https://docs.codeclimate.com/docs/rubocop for CodeClimate so that it reads from our `.rubocop` file.
Also, there are other ways of doing it, like having overcommit setup... Let's have a discussion round with the team about this as well :v:

app/models/profile.rb
akostadinov: It is not:
```
[1] pry(main)> Profile.take.logo.has_changes_to_save?
NoMethodError: undefined method `has_changes_to_save?' for #<Paperclip::Attachment:0x0000000007c523b0>
```

Gemfile
akostadinov: Done. Actually I noticed that we do not need `fastimage`. Because it uses another gem - `local-fastimage`.

akostadinov:
> I think the PR should also include `THREESCALE-8579` in the beginning of the title, for the scraping to be performed correctly.

I hate pull request messages with issue number. Just personal thing. Having the issue reference in the commit message is understood both from JIRA and errata. So it is fine.

lib/paperclip/gd_image_processor.rb
jlledom, line 9: Question: why the regular expression? is it possible that the geometry is returned in any other format than WxH ?

akostadinov: There can be a modifyer character(s) at the end of target geometry.

app/models/profile.rb
```diff
@@ -42,14 +42,23 @@ class Profile < ApplicationRecord
   belongs_to :account
   delegate :s3_provider_prefix, to: :account
 
+  LOGO_STYLES = ->(logo) {
+    {
+      large: '300x300>'.freeze,
+      medium: '150x150>'.freeze,
+      thumb: '100x100>'.freeze,
+      invoice: (logo.content_type.end_with?('gif') ? ['200x50>'.freeze, :png].freeze : '200x50>'.freeze)
+    }.freeze
```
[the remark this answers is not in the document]
akostadinov: We do. It doesn't come from imagemagic but from the `Geometry` class calculations. So it is taken into account.

test/unit/pdf/finance/invoice_generator_test.rb
```diff
@@ -26,19 +26,27 @@ class Pdf::Finance::InvoiceGeneratorTest < ActiveSupport::TestCase
...
     logo_file = File.open(file_fixture('wide.jpg'), 'rb')
-    @data.expects(:with_logo).yields(logo_file)
+    @data.expects(:with_logo).yields(image_path)
```
mayorova, line 40: `image_path` > `logo_file`

akostadinov: :man_facepalming:

mayorova (dismissed):
LGTM

thalesmiguel (dismissed):

jlledom (dismissed):

josemigallas:
@thalesmiguel Didn't you have any issues installing `local-fastimage_resize` on your mac? Merging this PR would essentially break my local env.

thalesmiguel:
I managed to install the `local-fastimage_resize` gem with `gem install local-fastimage_resize -- --with-opt-dir=$(brew --prefix gd)`. After that, `bundle [install]` went smoothly and I was able to run the application.
@akostadinov I can give it another try after the merge conflicts are solved if you want.

osx-manual-setup.md
```diff
@@ -58,7 +58,7 @@ xcode-select -—install
 ### Dependencies
 
 ```
-brew install chromedriver imagemagick@6 gs pkg-config openssl geckodriver sphinx mysql@5.7 postgresql@14
+brew install chromedriver imagemagick@6 gs pkg-config openssl geckodriver sphinxgd mysql@5.7 postgresql@14
```
josemigallas, line 61: Wait, shouldn't we remove imagemagick and add gd?
```suggestion
brew install chromedriver libgd gs pkg-config openssl geckodriver sphinxgd mysql@5.7 postgresql@14
```

akostadinov: in the next one #3165

akostadinov: But wait isn't package called `gd`? 
https://formulae.brew.sh/formula/gd

PR #3066, opened 2022-09-21
THREESCALE-7132: Add support for AWS STS authentication

thalesmiguel:

This PR implements STS credentials that can be used to access S3 buckets instead of the IAM credentials used by default.

Implements https://issues.redhat.com/browse/THREESCALE-7132.

**Some context first**
- With the current IAM-related ENV variables (`AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`), calls to AWS should still work; 
- If the new STS-related ENV variables (`AWS_ROLE_ARN` and `AWS_WEB_IDENTITY_TOKEN_FILE`) are **also** present, we'll still call AWS with the IAM credentials;

**Verification steps** 
- With IAM credentials
  - Assert that the files we store on S3 are still available;
    - Examples of files we store on S3: `Invoice#pdf`, `Profile#logo`, `File#attachment`;
- With both IAM and STS credentials;
  - Assert that IAM credentials are used instead of the STS ones;
- With STS credentials (the new stuff)
  - Locally:
    - Have a `token` file somewhere in your local machine;
    - Set the `AWS_WEB_IDENTITY_TOKEN_FILE` so it points to the `token` file;
    - Set a valid `AWS_ROLE_ARN`;
    - Assert that the files we store on S3 are still accessible;
  - RHOAM:
    - The `token` file should already be available from an Operator;
    - Assert that the ENV variables are set;
    - Assert that the files we store on S3 are still accessible;


app/services/aws/credentials_service.rb
```diff
@@ -0,0 +1,53 @@
...
+
+    # @return [Hash] Depending on the available connections, this method may return:
+    # { credentials: Aws::STS::Types::AssumeRoleWithWebIdentityResponse } if STS credentials are available;
+    # { web_identity_token_file: 'foo', role_name: 'bar' } if the values are present;
+    # An empy Hash if no configurations are available;
+    def call
+      if sts_params.any? && sts_params.all? { |_key, value| value.present? } && web_identity_token_file_exists?
+        { credentials: sts_credentials }
+      else
+        aws_credentials
```
laurafitzgerald, line 27: small naming suggestion
```suggestion
        iam_credentials
```

app/services/aws/credentials_service.rb
```diff
@@ -0,0 +1,53 @@
...
+
+    ROLE_SESSION_NAME = '3scale-porta'
+
+    def initialize(params)
+      @params = params
+    end
+
+    # @return [Hash] Depending on the available connections, this method may return:
+    # { credentials: Aws::STS::Types::AssumeRoleWithWebIdentityResponse } if STS credentials are available;
+    # { web_identity_token_file: 'foo', role_name: 'bar' } if the values are present;
```
akostadinov, line 21: I read is as if  `{ web_identity_token_file: 'foo', role_name: 'bar' }` can be a return value. While from code it seems like that if `{ web_identity_token_file: 'foo', role_name: 'bar' }` is present, then return value is `{ credentials: Aws::STS::Types::AssumeRoleWithWebIdentityResponse }`.

May need some rewording.

akostadinov: btw I expected that we will check for the token file at a predefined location instead of requiring configuration. I guess it depends how operator implements this.

akostadinov: wrt `role_name`, is this something that administrator assigns to us or is it something that we always need? I mean if we need a specific role that is global to any S3 account, maybe we should hardcode the proper value. I don't know how it works so excuse my ignorance.

test/unit/services/aws/credentials_service_test.rb
```diff
@@ -0,0 +1,69 @@
...
+    Aws::AssumeRoleWebIdentityCredentials.expects(:new).never
+
+    assert Aws::CredentialsService.call(params_for_aws_credentials), {
+      access_key_id: params_for_aws_credentials[:access_key_id],
+      secret_access_key: params_for_aws_credentials[:secret_access_key]
+    }
+  end
+
+  test '#call does not try to assume role if the web_identity_token_file does not exist' do
+    params_for_sts_credentials = full_params.slice(:web_identity_token_file, :role_name)
```
akostadinov, line 46: seems like this can be extracted as a method

app/lib/cms/s3.rb
```diff
@@ -35,7 +35,9 @@ def region
     end
 
     def credentials
-      config.slice(:access_key_id, :secret_access_key) if enabled?
+      return unless enabled?
```
akostadinov, line 38: Maybe this behavior already has a test but good to add such if not.

app/services/aws/credentials_service.rb
```diff
@@ -0,0 +1,53 @@
...
+    # An empy Hash if no configurations are available;
+    def call
```
briangallagher, line 23: Hi  @Thales Martinez  @Nidhi Soni  we're hoping to release STS support for RHOAM in the next release, 1.27. To do that we will have the 3scale work around in place i.e. providing an IAM key/secret in an OpenShift secret to 3scale. 
When STS support for 3scale does come out, RHOAM will provide an OpenShift secret with an ARN representing the role to be used for the STS support. 
The question is, will this process be seamless? Can we go from the IAM based secret to the STS based secret without any manual intervention. Will 3scale automatically switch from IAM to STS as long as the correct secret is provided?

thalesmiguel: Hello @briangallagher 👋 
With this implementation, 3Scale will try to use STS credentials if a `web_identity_token_file` file is provided, which **should** be seamless. Now, I can't really guarantee that there won't be any manual intervention for us to get access to this file... Maybe @raelga can help us with this.

app/services/aws/credentials_service.rb
thalesmiguel: >btw I expected that we will check for the token file at a predefined location instead of requiring configuration. I guess it depends how operator implements this.

The file should be on a predefined location and we should know its location through configuration. As explained in [here](https://github.com/openshift/cloud-credential-operator/blob/master/docs/sts.md), we should have access to both `web_identity_token_file` and `role_name` from OpenShift.

>wrt role_name, is this something that administrator assigns to us or is it something that we always need? I mean if we need a specific role that is global to any S3 account, maybe we should hardcode the proper value.

Good question! `role_name` is the ARN that should be provided by OpenShift.

app/lib/cms/s3.rb
thalesmiguel: Look at that... It wasn't tested at all. Added in fd980729e.

app/services/aws/credentials_service.rb
thalesmiguel: Much better name. Changed in bce3c14e5. Thank you!

config/examples/amazon_s3.yml
```diff
@@ -4,15 +4,24 @@ development:
...
+  role_session_name: ""
 
 test: &test
   access_key_id: invalid
   secret_access_key: invalid
   bucket: test-my-s3-bucket
   region: "us-east-1"
+  web_identity_token_file: ""
+  role_name: ""
+  role_session_name: ""
```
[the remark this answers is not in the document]
thalesmiguel: I added those in there so we have a sample (more like a placeholder) of the possible credentials we use to access S3. So, as we have the possibility to use an `access_key_id` and `secret_access_key`, I also wanted to make it clear we can use `web_identity_token_file` and `role_name`.

test/unit/services/aws/credentials_service_test.rb
thalesmiguel: I agree. Extracted in 00e94efb4 :v:

app/services/aws/credentials_service.rb
briangallagher: @thalesmiguel thanks. Can this scenario be tested. The BU are reluctant to roll out STS support for RHOAM is there is complex manual steps required when STS becomes available in 3scale

thalesmiguel: @raelga could you please clarify this?

raelga: To test the functionality, we a deployed development OCP4.9 cluster with STS:

Screenshot 2022-09-28 at 17 58 04
https://console-openshift-console.apps.dev-eng-ocp4-sts.dev.3sca.net/k8s/ns/3scale-sts-rael/pods/aws-cli/terminal

The role has been created with `ccoctl` using the follwing `credentialsrequest` to allow S3 access to the test bucket:

```yaml
apiVersion: cloudcredential.openshift.io/v1
kind: CredentialsRequest
metadata:
  name: system-aws-creds
  namespace: openshift-cloud-credential-operator
spec:
  secretRef:
    name: system-aws-creds
    namespace: 3scale-sts-rael
  providerSpec:
    apiVersion: cloudcredential.openshift.io/v1
    kind: AWSProviderSpec
    statementEntries:
      - effect: Allow
        action:
          - "s3:*"
        resource: "arn:aws:s3:::3scale-dev-eng-system-assets"
      - effect: Allow
        action:
          - "s3:*"
        resource: "arn:aws:s3:::3scale-dev-eng-system-assets/*"
  serviceAccountNames:
    - default
    - 3scale-sts-test
 ```

@thalesmiguel  feel free to test porta with the STS feature in the new cluster, the development team has full admin privileges to this cluster using Github authentication. The cluster will be destroyed once the testing is completed.

Comment (laurafitzgerald) [reply app/services/aws/credentials_service.rb:23]: Normally cco is turned to manual mode when the cluster is STS enabled. Is the case that this was tested on non STS enabled cluster? I think it's fine as long as it was possible to confirm that the component given the secret was indeed using it in order to authenticate using STS rather than IAM any IAM credentials.

Comment (raelga) [reply app/services/aws/credentials_service.rb:23]: This was tested in a STS enabled cluster with CCO in manual mode.

The `CredentialsRequest`  is not actually deployed in the cluster, it just was used to generate the required role with `ccoctl` in the AWS account and generate the secret that needs to be deployed in the namespace.

File: app/services/aws/credentials_service.rb [hunk 7]
```diff
@@ -8,46 +8,58 @@ def call(params = {})
       end
     end
 
-    attr_reader :params
+    AuthenticationTypeError = Class.new(StandardError)
+    TokenNotFoundError = Class.new(StandardError)
```
akostadinov, line 12: https://rubystyle.guide/#single-line-classes

app/services/aws/credentials_service.rb
```diff
@@ -8,46 +8,58 @@ def call(params = {})
...
-      params.slice(:web_identity_token_file, :role_name)
+    def iam_credentials
+      @iam_credentials ||= params.slice(*IAM_KEYS)
     end
 
-    def web_identity_token_file_exists?
-      File.exist?(sts_params[:web_identity_token_file])
+    def sts_credentials
+      @sts_credentials ||= params.slice(*STS_KEYS).tap do |credentials|
+        credentials[:role_session_name] ||= DEFAULT_ROLE_SESSION_NAME
```
akostadinov, line 65: I think it makes sense to have a default path for `:web_identity_token_file` too. And then use `File.exists?` to validate that this file exists.

This will make it work configuration-less which is great. Setup cluster properly and skip adding any configuration values.

test/unit/services/aws/credentials_service_test.rb
```diff
@@ -7,45 +7,46 @@ class Aws::CredentialsServiceTest < ActiveSupport::TestCase
...
   end
 
+  test '#call returns IAM credentials when available' do
+    Aws::AssumeRoleWebIdentityCredentials.expects(:new).never
+
+    assert Aws::CredentialsService.call(iam_auth_params), {
+      access_key_id: iam_auth_params[:access_key_id],
+      secret_access_key: iam_auth_params[:secret_access_key]
+    }
+  end
```
akostadinov, line 17: Would be good to have a test that verifies IAM preference over STS.

test/unit/services/aws/credentials_service_test.rb
```diff
@@ -0,0 +1,76 @@
...
+      region: 'region'
+    }
+  end
+
+  def iam_auth_params
+    full_params.slice(:access_key_id, :secret_access_key)
+  end
+
+  def sts_auth_params
+    full_params.slice(:region, :role_arn, :role_session_name, :web_identity_token_file)
```
akostadinov, line 70: I wonder, can't we do some call to get `region` instead of requiring a config value?

akostadinov (commented):
Nicely done!

some comments below

test/unit/services/aws/credentials_service_test.rb
thalesmiguel: Agreed! Added in 16f2b72.

app/services/aws/credentials_service.rb
thalesmiguel: Updated in f391f58.

app/services/aws/credentials_service.rb
thalesmiguel: I'm not sure on what could be a valid default value for it.
Operator will add a projected volume with the file, which means it'll be available for OpenShift deployments. A realistic value here would be `/var/run/secrets/openshift/serviceaccount/token`, which doesn't look to make sense with the application.
I'd say it's better not to assume it at all to avoid some level of coupling. The application simply receives the arguments and tries to understand which auth type it'll use, while guarding itself from crashes that could come from the external world (`File.exist?`).
What do you think?

test/unit/services/aws/credentials_service_test.rb
thalesmiguel: Do you mean something like "calling Amazon to figure out the region"?
`region` is a mandatory value we send to S3 even with IAM authentication.

akostadinov: Yes, IIRC there was a a call to check region of a bucket.

app/services/aws/credentials_service.rb
akostadinov: I think that we will have a documented way how to configure and there should be a way to figure out these settings  I think by cluster config or convention. Our app is deployed only in OpenShift so I don't see an issue in case that value is openshift specific.

If `File.exist?` crashes, then we will have a lot of bigger issues already. Also if there were no IAM credentials, then no STS credentials, then app will `raise` anyway. If `File.exists?` raises it will not make things much worse IMO.

For me ultimate goal is to make user experience seamless - need to do as little as possible.

For region - if we required it before, it can wait for another time. But this is a new functionality and is better to not introduce this later.

test/unit/services/aws/credentials_service_test.rb
thalesmiguel: I was looking into it and found [get-bucket-location](https://docs.aws.amazon.com/cli/latest/reference/s3api/get-bucket-location.html#get-bucket-location), which's part of the S3 API. It's also exposed [in the AWS Gem](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/S3/Client.html#get_bucket_location-instance_method).
The catch here is that we need an [S3 Client](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/S3/Client.html#initialize-instance_method) to perform this call, which depends on a `region` as well.
Now, before digging deeper on this one, is this really something we'd like to do? My reasoning is that the `region` is a required argument that needs to be provided in order for the application to communicate with S3, not that far from the actual credentials, in the sense that all values are mandatory to proceed.
Right now we request that the `region` is provided for IAM authentication and the same will still be true for STS auth, so I think it's not necessary to be changed... Also, even if this call is possible somehow, do we need to add yet another external call to the app?

app/services/aws/credentials_service.rb
thalesmiguel: > For me ultimate goal is to make user experience seamless - need to do as little as possible.

This is the point. Let me confirm with OPS what'll be the default path for the token :v:

test/unit/services/aws/credentials_service_test.rb
akostadinov: If it was a new requirement, I would insist. Given it has been like this before, we can keep it.

PS one more call, provided we will do a lot of calls, one more or less would not bother me personally.

config/examples/amazon_s3.yml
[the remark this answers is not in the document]
akostadinov: We have them under production already. Seems redundant to keep them empty in there.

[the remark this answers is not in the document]
thalesmiguel: This is outdated as there was a change in `config/examples/amazon_s3.yml` after this PR was opened.
I updated this branch and added the new values to it: https://github.com/3scale/porta/blob/f391f58dc2f1e8b2e8d5de26c5d54dde957e5007/config/examples/amazon_s3.yml

app/services/aws/sts/assume_role_web_identity_service.rb
```diff
@@ -0,0 +1,51 @@
...
+
+      attr_reader :params
+
+      def cached_call
+        cached_value = ::Rails.cache.read(cache_key)
+        return cached_value if cached_value.present?
+
+        sts_response = yield
+
+        ::Rails.cache.write(cache_key, sts_response, expires_in: cache_expires_in(sts_response.expiration))
```
akostadinov, line 35: shouldn't we use just some class attribute/variable instead of using memcached? Do you see any benefits with that?

thalesmiguel: I don't quite understand the suggestion of class attribute/variable 😬 
I added this cache so that we don't call AWS more than we need. Without it, each time we call AWS (for S3), we'll first have to call STS to get the temporary credentials. With the cache, we won't perform calls to STS until the cache expires, which's at the same time the temporary credentials expire.

thalesmiguel: The credentials we get from `Aws::AssumeRoleWebIdentityCredentials` are temporary. They become invalid after 1 hour (by default) and, when that happens, we need to request for credentials again, using the token that is refreshed after X time as well.
Now, by adding cache, we'll only hit AWS again when we know the credentials will expire, as STS tells us the expiration date/time.
Those credentials will be used each time we have to communicate with S3 to fetch the assets we have stored there. If I understood your suggestion correct, by keeping the state of the credentials in a variable, we'd be able to use it only until the transaction finishes. Once a new transaction starts (page refresh?), we'd have to call AWS again.

app/services/aws/credentials_service.rb
thalesmiguel: An update on this: @eguzki and I talked about whether or not the path to the token file will be configurable by our users. Since this will be an internal process, the operator will be responsible for creating the volume, putting the token file in the correct place and also provide the application with the correct location of the file, without the need of our users to set paths.
As this is related with OPS configuration, I'd avoid adding the default value to the application. Agreed?

app/services/aws/sts/assume_role_web_identity_service.rb
akostadinov: On https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/AssumeRoleWebIdentityCredentials.html I read:
> An auto-refreshing credential provider that assumes a role via STS::Client#assume_role_with_web_identity.

Are you sure it expires and new object has to be created?

thalesmiguel: > Are you sure it expires and new object has to be created?

I'm sure it expires (check [here](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/STS/Client.html#assume_role_with_web_identity-instance_method) please) now that the new object has to be created is debatable. If we can keep the same object all the time, it'll be able to automatically go to AWS to fetch new credentials in case those expire.
Let me try to do something here... I'm thinking about a singleton 🤔

akostadinov: yes, token expires. Just I was left with the impression that the credentials provider will silently handle that under the hood.

Would be nice to have it as a singleton of some sort. This is what I mean to keep it into a class instance variable.

thalesmiguel: >Just I was left with the impression that the credentials provider will silently handle that under the hood

And it does indeed. And in quite a nice way... When we're using the credentials provided by `assume_role_with_web_identity`, if those are expired, the object makes sure to retrieve new credentials.
I added another commit (06d1aff) to remove the cache (yay!) in favour of a class that'll keep the `Aws::AssumeRoleWebIdentityCredentials` instance in memory. With this, unless that application instance changes, we should always be using the same object :v:

Thank you, @akostadinov! ... Could you please double check again 😬 ?

test/unit/cms/s3_test.rb
```diff
@@ -64,4 +64,16 @@ class CMS::S3Test < ActiveSupport::TestCase
     assert CMS::S3.options[:endpoint]
     assert_not CMS::S3.options[:use_fips_endpoint]
   end
+
+  test "#credentials returns nil when not CMS::S3 is not enabled" do
```
nidhi-soni1104, line 68: redentials returns nil when CMS::S3 is not enabled would be better

akostadinov (commented):
Very nice, thank you!

Find a few minor comments. I'm just wondering, having two services just to get credentials, is this a little bit of an overkill?

test/unit/cms/s3_test.rb
thalesmiguel: Much better indeed. It even makes sense now 😅 
Updated in 386b70e.

test/unit/services/aws/sts/assume_role_web_identity_service_test.rb
```diff
@@ -0,0 +1,64 @@
...
+    assume_role_web_identity_service_instance.config(sts_auth_params)
+
+    assert assume_role_web_identity_service_instance.identity_credentials, sts_credentials
+  end
+
+  private
+
+  def sts_auth_params
+    {
+      web_identity_token_file: '/path/to/token',
```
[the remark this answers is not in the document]
thalesmiguel: This is just a random/placeholder path for test purposes and could be anything, as we're not actually looking for a real file during tests.

test/unit/services/aws/credentials_service_test.rb
```diff
@@ -0,0 +1,76 @@
...
+  end
+
+  def sts_credentials
+    @sts_credentials ||= Aws::Credentials.new(nil, nil)
+  end
+
+  def assume_role_web_identity_service_instance
+    Aws::Sts::AssumeRoleWebIdentityService.instance
+  end
```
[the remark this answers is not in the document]
thalesmiguel: `Aws::Sts::AssumeRoleWebIdentityService.instance` is used multiple times in the test, so I added this method just so there's no code duplication.

test/unit/services/aws/sts/assume_role_web_identity_service_test.rb
```diff
@@ -0,0 +1,64 @@
...
+class Aws::Sts::AssumeRoleWebIdentityServiceTest < ActiveSupport::TestCase
+  setup do
+    File.stubs(:exist?).returns(true)
+    assume_role_response.class.any_instance.stubs(:expiration).returns(1.minute.from_now)
+    assume_role_response.stubs(:credentials).returns(sts_credentials)
+  end
+
+  test '#config sets params' do
+    assert(
+      assume_role_web_identity_service_instance.config(sts_auth_params).as_json['params'].keys.sort,
```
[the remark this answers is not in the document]
thalesmiguel: I agree it's more readable. I usually avoid using `send` so I don't call private methods directly. Now, since this is a test, I'm OK with the change. Updated in c003f85 :v:

test/unit/services/aws/credentials_service_test.rb
```diff
@@ -0,0 +1,76 @@
...
+    assume_role_web_identity_service_instance.expects(:identity_credentials).never
+
+    assert Aws::CredentialsService.call(full_params), iam_auth_params
+  end
+
+  test '#call raises an error if the web_identity_token_file does not exist' do
+    token_not_found_error = Aws::Sts::AssumeRoleWebIdentityService::TokenNotFoundError
+    File.stubs(:exist?).with(sts_auth_params[:web_identity_token_file]).returns(false)
+
+    assume_role_web_identity_service_instance.expects(:config).raises(token_not_found_error)
```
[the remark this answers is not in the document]
thalesmiguel: Indeed! And it's actually a duplicated test now, as I moved this error to [Aws::Sts::AssumeRoleWebIdentityService](https://github.com/thalesmiguel/porta/blob/support-aws-sts/test/unit/services/aws/sts/assume_role_web_identity_service_test.rb#L23-L29). Removed in bbcb342.

app/services/aws/credentials_service.rb
```diff
@@ -0,0 +1,64 @@
...
+    #
+    # @param [Hash] Hash containing AWS credentials related keys
+    # @return [Hash<Symbol, String>] with access_key_id and secret_access_key when IAM credentials are provided
+    # @return [Hash<Symbol, Aws::AssumeRoleWebIdentityCredentials>] when STS credentials are provided
+    # @raise [AuthenticationTypeError] if not enough params are provided
+    def call
+      if valid_params_with(iam_params, IAM_KEYS)
+        iam_params
+      elsif valid_params_with(sts_params, STS_KEYS)
+        sts_temporary_security_credentials
```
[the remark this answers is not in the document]
thalesmiguel: Update in bbcb342. Thank you!

nidhi-soni1104:
PR looks good to me lets the test pass then I will approve this

config/application.rb
```diff
@@ -261,6 +261,15 @@ def try_config_for(*args)
       config.three_scale.oauth2 = config_for(:oauth2)
     end
 
+    initializer :config_sts_client, after: :load_configs do
+      config.sts_client = CMS::STS::AssumeRoleWebIdentity.instance.tap do |sts_client|
```
mayorova, line 265: Feels a bit weird to initialize `AssumeRoleWebIdentity` here, without actually knowing whether it will be used or not - chances are it will not, as maybe IAM credentials will be used, so this singleton is there for nothing.
Not a big deal, but hmm...

thalesmiguel: The idea in here is to have the ``Aws::STS::Client`` instance available throughout all application as part of the app's configuration. So, in here, we just create an object with the necessary configurations (maybe even missing something) and leave it in memory to be used or not.
Now, as a config, there shouldn't be any problem if we don't use it at all. Like how we don't use the `paperclip_defaults` config or even the `s3` one, in case we never hit the buckets to get/put assets (for CMS).

That was the line of thought 😬 Makes any sense?

mayorova: Yep, I get the idea, and that's what I don't like too much :upside_down_face: 
Indeed, we store config for S3, even though we might be using filesystem, and not S3.

That's the fact that it's actually an object instance, and not just some key-value hash that bugs me.

akostadinov: Given `AssumeRoleWebIdentity` is a singleton, we can store the value as a `class_attrubute` (or a class-instance variable).

thalesmiguel: >That's the fact that it's actually an object instance, and not just some key-value hash that bugs me.

I myself am not a big fan of globals but, for this case where we need to be sure this Client object never changes, so that the credentials can be fetch automatically, I think it makes sense. Specially since it's a Singleton.
For what it's worth, `Rails.application.config.sts_client` is pretty much a Hash (`.as_json` 😬) until credentials are requested, which'll only happen when STS is used over IAM.

test/unit/services/aws/credentials_service_test.rb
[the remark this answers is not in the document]
akostadinov: DRY shouldn't be taken to the extreme. There are multiple sources on the internet that can be read about it. This one looks reasonable to me [1].

In this case I commented because I had to scroll down to see what this method did exactly and then I see just a single call. So it doesn't reduce complexity. Just adds some indirection. It also doesn't cache or benefit us as far as I can see.

As I wrote, just a nitpick, so not really concerned whether it stays or goes. I like theoretical discussions though :)

[1] https://betterprogramming.pub/when-dry-doesnt-work-go-wet-6befda0444bf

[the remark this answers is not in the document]
thalesmiguel: I wouldn't call this extreme DRY or anything of sorts 😅
This is a Singleton so there's really no harm in calling it multiple times but that would mean, for this test, calling  `CMS::STS::AssumeRoleWebIdentity.instance` 7 times, which looks even odder to me.
This link you shared has an interesting view on how "bad abstractions" can have an impact on the future, when new criteria appears and the old abstract code needs to be changed. I don't think it applies to this piece of code on the test we're discussing, as it's just one verbose named method that returns an instance of a Singleton. If anything, this method is nothing more than a local helper.

As I mentioned, "I added this method **just** so there's no code duplication.". Now, if you think it brings more complexity than benefits, I'd be glad to remove it. Or maybe the method's name is the problem here? 🤔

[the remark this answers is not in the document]
akostadinov: "problem" is a strong word here. Just as author you know what this method does. And I as a reader got annoyed that I have to lookup what it did only to see it is just singular call. I don't worry about repeating 7 times as it is easy to `sed` it if it ever needs to change, which is unlikely.

So just giving you the perspective of a reader. And my preference. But it is absolutely fine to keep :)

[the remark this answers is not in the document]
thalesmiguel: Another thought on this: IF this was an (R)spec instead of a (Mini)test, my first idea would be to add `CMS::STS::AssumeRoleWebIdentity.instance` to a `let` variable.
Maybe this would be clearer if, instead of this private method, I'd declare an instance variable in this test's `setup` method 🤔.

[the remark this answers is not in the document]
thalesmiguel: Changed it in afe6d15 :v:

config/application.rb
thalesmiguel: >Given AssumeRoleWebIdentity is a singleton, we can store the value as a class_attrubute (or a class-instance variable).

Looks good. Updated in c89d43e.

test/unit/services/aws/credentials_service_test.rb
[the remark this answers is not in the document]
thalesmiguel: Aaaaand removed in c89d43e, as the usage of `class_attribute` in the Singleton removed the necessity of working directly with an instance when setting attributes :v:

app/services/aws/credentials_service.rb
```diff
@@ -0,0 +1,60 @@
...
+      end
+    end
+
+    private
+
+    attr_reader :params
+
+    def valid_params_with(credentials, keys)
+      return false if credentials.empty?
+      return false if keys.difference(credentials.keys).any?
```
josemigallas, line 43:
```suggestion
      return false if credentials.empty? || keys.difference(credentials.keys).any?
```

app/lib/cms/sts/assume_role_web_identity.rb
```diff
@@ -0,0 +1,39 @@
...
+            web_identity_token_file: web_identity_token_file
+          )
+        end
+      end
+
+      private
+
+      attr_reader :params
+
+      def validate_configuration
```
josemigallas, line 28: Not a big deal again but "validation" is a bit misleading. This only checks the token presence, not its validity.

app/lib/cms/sts/assume_role_web_identity.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+module CMS
+  module STS
+    class AssumeRoleWebIdentity
+      include Singleton
+      # :reek:Attribute
```
josemigallas: Is it worth adding a comment here?

app/services/aws/credentials_service.rb
```diff
@@ -0,0 +1,60 @@
...
+
+      credentials.all? { |_key, value| value.present? }
+    end
+
+    def sts_credetials
+      { credentials: ::Rails.application.config.sts_client.instance.identity_credentials }
```
akostadinov, line 49:
```suggestion
      { credentials: CMS::STS::AssumeRoleWebIdentity.instance.identity_credentials }
```

app/lib/cms/sts/assume_role_web_identity.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+module CMS
+  module STS
+    class AssumeRoleWebIdentity
+      include Singleton
+
+      class_attribute :region, :role_arn, :role_session_name, :web_identity_token_file
```
akostadinov, line 8: my suggestion was to set `identity_credentials` as a class_attribute, everything else is already stored in config as far as I can tell. Is there any advantage to keep these as an attribute but store `identity_credentials` in Rails configuration? i.e. the point was to avoid polluting Rails configuration with with custom objects. And set it regardless of whether we use that or not... as explained in @mayorova's comment.

app/lib/cms/sts/assume_role_web_identity.rb
akostadinov: I understood it as validating that configuration options are set. But given it does only that - check token presence, what you say actually makes sense!

app/lib/cms/sts/assume_role_web_identity.rb
thalesmiguel: Changed in 18c48a2.

app/lib/cms/sts/assume_role_web_identity.rb
thalesmiguel: Changed in 18c48a2.

app/services/aws/credentials_service.rb
thalesmiguel: Changed in 7dc1b13.

app/lib/cms/sts/assume_role_web_identity.rb
thalesmiguel: Changes applied in e6d084e.
Now the `CMS::STS::AssumeRoleWebIdentity` Singleton gets its data from ENV variables (through Rails.application.config.s3), as we discussed :v:

config/application.rb
thalesmiguel: Updated it again in e6d084e. Please check https://github.com/3scale/porta/pull/3066#discussion_r997180912.

app/services/aws/credentials_service.rb
```diff
@@ -0,0 +1,59 @@
...
+    end
+
+    # Returns either IAM or STS credentials to be used on calls to S3
+    #
+    # @param [Hash] Hash containing AWS credentials related keys
+    # @return [Hash<Symbol, String>] with access_key_id and secret_access_key when IAM credentials are provided
+    # @return [Hash<Symbol, Aws::AssumeRoleWebIdentityCredentials>] when STS credentials are provided
+    # @raise [AuthenticationTypeError] if not enough params are provided
+    def call
+      if valid_params_with(iam_params, IAM_KEYS)
```
jlledom, line 28: Shouldn't STS be the default one, and use IAM only when no STS is provided? that's what the [comment](https://github.com/3scale/porta/pull/3066/files#r980422584) above suggests, but the PR description and this code says the opposite.

app/lib/cms/sts/assume_role_web_identity.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+module CMS
```
jlledom, line 3: I'd need some help on understanding project folders, why is this under `CMS`? Is not that for the developer portal? and why under the `lib` folder? This is not a library but part of the project, it seems to me this would be better placed at `services/aws`

app/lib/cms/sts/assume_role_web_identity.rb
thalesmiguel: This magic comment was removed in c89d43e :v:

openshift/system/.env
```diff
@@ -1,6 +1,8 @@
 # S3 access
 AWS_ACCESS_KEY_ID=
 AWS_SECRET_ACCESS_KEY=
+AWS_ROLE_ARN=
+AWS_WEB_IDENTITY_TOKEN_FILE=
```
jlledom, line 5: `AWS_REGION` would be defined here as well.

app/services/aws/credentials_service.rb
akostadinov: Using IAM is something more manual than the automatic STS provided by operator. At least this is my understanding. So to me it makes more sense to give priority to the things users manually do/configure (i.e. IAM with priority).

app/lib/cms/sts/assume_role_web_identity.rb
akostadinov: This is not a service. More of a helper class. I think path was chosen because `s3.rb` is already there.

My guess without checking git log is that `s3.rb` is already there for historic reasons. 

For me `app/lib` is the proper location. And I prefer `s3.rb` and this one to stay together for consistency. Now I guess `s3.rb` can also be moved along with any other related classes if there are such. I wouldn't make it mandatory for this PR though.

app/services/aws/credentials_service.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+module Aws
+  class CredentialsService
+    class << self
+      def call(params = {})
+        new(params).call
+      end
+    end
```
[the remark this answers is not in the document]
akostadinov: This is how ruby service objects convention goes. See for example this [article](https://dev.to/joker666/ruby-on-rails-pattern-service-objects-b19). If you see a way to make it nicer though, maybe elaborate a little more.

app/lib/cms/sts/assume_role_web_identity.rb
```diff
@@ -0,0 +1,39 @@
...
+
+      def identity_credentials
+        @identity_credentials ||= begin
+          check_token_presence
+
+          Aws::AssumeRoleWebIdentityCredentials.new(
+            region: config[:region],
+            role_arn: config[:role_arn],
+            role_session_name: config[:role_session_name],
+            web_identity_token_file: config[:web_identity_token_file]
```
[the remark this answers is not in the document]
akostadinov: Correct me if I'm wrong but when we are here, we are inside the `CMS::S3.credentials` call. This should still allow us to get the rest of the values though from it. So this suggestion makes sense to me unless I miss something and we go into an infinite loop.

app/services/aws/credentials_service.rb
jlledom: makes sense

app/services/aws/credentials_service.rb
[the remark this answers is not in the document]
jlledom: I see, but it's not following the pattern because there's not an `ApplicationService` base class with the static call to be reused from all the services.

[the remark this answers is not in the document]
akostadinov: There doesn't seem to be such a base class in Rails. Neither in porta codebase. I don't think it is a mandatory part of the pattern. It makes sense to have one and I'm also fine not to have if the duplication is just one method.

Also for me it is fine if `CMS::S3` is the single entrypoint to getting S3 configuration so this could be another helper class (not a service) that shouldn't be used by anything but `CMS::S3`. On the other hand STS credentials are potentially useful to other uses, not only S3 so it is also acceptable to be a service class.

app/lib/cms/sts/assume_role_web_identity.rb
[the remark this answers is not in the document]
thalesmiguel: For this change, we'd have to expose `role_arn`, `role_session_name` and `web_identity_token_file` on `CMS::S3`. I'm wondering if that'd be the best approach, since none of those values would be usable without each other... Maybe expose a "sts_arguments" with all those values?

With the current implementation, the single source of truth is the `amazon_s3.yml` config file, witch we access through the Application variable.

No real preference from my side... What do you think would be better?

app/services/aws/credentials_service.rb
```diff
@@ -0,0 +1,59 @@
...
+    #
+    # @param [Hash] Hash containing AWS credentials related keys
+    # @return [Hash<Symbol, String>] with access_key_id and secret_access_key when IAM credentials are provided
+    # @return [Hash<Symbol, Aws::AssumeRoleWebIdentityCredentials>] when STS credentials are provided
+    # @raise [AuthenticationTypeError] if not enough params are provided
+    def call
+      if valid_params_with(iam_params, IAM_KEYS)
+        iam_params
+      elsif valid_params_with(sts_params, STS_KEYS)
+        sts_credetials
```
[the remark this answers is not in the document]
thalesmiguel: Fixed on b5d33a7 🙈

app/services/aws/credentials_service.rb
```diff
@@ -0,0 +1,59 @@
...
+      def call(params = {})
+        new(params).call
+      end
+    end
+
+    class AuthenticationTypeError < StandardError
+    end
+
+    IAM_KEYS = %i[access_key_id secret_access_key].freeze
+    STS_KEYS = %i[region role_arn role_session_name web_identity_token_file].freeze
```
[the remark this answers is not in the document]
thalesmiguel: You're right! Removed on 052c6f1.

openshift/system/.env
thalesmiguel: And we were missing some others too. Fixed on c2d9b5d. Thank you!

app/services/aws/credentials_service.rb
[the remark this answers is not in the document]
thalesmiguel: Regarding the service pattern, one thing I did in the past was to have a `Callable` Module that was added in whatever Service Object that didn't had the need to keep state (which was all of them, if memory serves right). It was something simple like:
```ruby
module Callable
  extend ActiveSupport::Concern

  included do
    private_class_method :new
  end

  class_methods do
    def call(*args)
      new(*args).call
    end
  end
end
```
Do you think it would make sense to add such a Module to Porta to be used on Service Objects?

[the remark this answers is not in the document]
jlledom: I think would be an improvement.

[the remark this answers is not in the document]
thalesmiguel: @jlledom added in 20ea64e :v:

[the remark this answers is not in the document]
jlledom: Cool. The [article](https://dev.to/joker666/ruby-on-rails-pattern-service-objects-b19) also includes a `block` parameter.

```
class ApplicationService
  def self.call(*args, &block)
    new(*args, &block).call
  end
end
```
I think it's worth adding it, so we can handle all kinds of calls.

app/lib/cms/sts/assume_role_web_identity.rb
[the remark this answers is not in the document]
jlledom: I think either are good. I'd prefer exposing the three properties in different methods, but I don't know why 😅. It just looks nicer to me.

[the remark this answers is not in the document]
thalesmiguel: Updated in 17a7ad1.

jlledom (approved):

akostadinov (approved):

test/unit/cms/s3_test.rb
```diff
@@ -104,9 +104,9 @@ class CMS::S3Test < ActiveSupport::TestCase
   end
 
   test "#web_identity_token_file returns nil when CMS::S3 is not enabled" do
-    CMS::S3.stubs(:enabled?).returns(:false)
+    CMS::S3.stubs(:enabled?).returns(false)
 
-    assert CMS::S3.web_identity_token_file, nil
+    assert_equal CMS::S3.web_identity_token_file, nil
```
akostadinov, line 109: `assert_nil`

thalesmiguel: Changed in 4bdc9fb.

akostadinov (approved):

test/unit/services/aws/credentials_service_test.rb
```diff
@@ -0,0 +1,61 @@
...
+
+    CMS::STS::AssumeRoleWebIdentity.instance.expects(:identity_credentials).returns(sts_credentials)
+
+    assert Aws::CredentialsService.call(sts_auth_params), { credentials: sts_credentials }
+  end
+
+  test '#call returns IAM credentials when both authentication types are available' do
+    CMS::STS::AssumeRoleWebIdentity.instance.expects(:identity_credentials).never
+
+    assert Aws::CredentialsService.call(full_params), iam_auth_params
```
mayorova, line 28: This test and the one on line 22 are not right :upside_down_face:

PR #3070, opened 2022-09-28
THREESCALE-6745: Fix ReferenceError in /p/admin/dashboard

thalesmiguel:

This PR attempts to fix the error we receive sometimes on Bugsnag (please check the ticket for the error).

Fixes [THREESCALE-6745: ReferenceError in /p/admin/dashboard](https://issues.redhat.com/browse/THREESCALE-6745)

**Verification steps** 
How to reproduce the error:
- Open the dashboard and assert that all data is still presented there without errors.

**Developer notes**
While working on this PR, it was noted that we could create two new React components to display the NewAccounts and PotentialUpgrades widgets, replacing the now "old way" of using Rails templates.
During the development, it was noted that we still had unreferenced code related with the old dashboard and that, after the changes proposed in this PR, more code could be removed. Thus, the 63f5633 commit was added to remove such code.


app/controllers/provider/admin/dashboards_controller.rb
```diff
@@ -19,7 +19,8 @@ def show
   end
 
   include DashboardTimeRange
-  helper_method :current_range, :previous_range, :backend_apis_presenter, :products_presenter
+  helper_method :current_range, :previous_range, :new_accounts_presenter, :potential_upgrades_presenter,
+    :backend_apis_presenter, :products_presenter
```
josemigallas, line 23: May I suggest reordering these items so the changes are the fewest possible:
```suggestion
  helper_method :current_range, :previous_range, :backend_apis_presenter, :products_presenter,
    :new_accounts_presenter, :potential_upgrades_presenter,
```

app/javascript/packs/dashboard_new_accounts_widget.js
```diff
@@ -0,0 +1,17 @@
+import { NewAccountsWidgetWrapper } from 'Dashboard/components/NewAccountsWidget'
+import { safeFromJsonString } from 'utilities'
+
+const containerId = 'new-accounts-widget'
+
+document.addEventListener('DOMContentLoaded', () => {
+  const container = document.getElementById(containerId)
```
josemigallas, line 7: I recommend asserting the container exists like it's done in other packs. Flow will warn you anyway. 👮🏻‍♂️

app/javascript/src/Dashboard/components/NewAccountsWidget.jsx
```diff
@@ -0,0 +1,79 @@
...
+    },
+    currentRangeAdminBuyersAccount: {
+      url: string,
+      value: string
+    }
+  },
+  percentualChange: number
+}
+
+const NewAccountsWidget = ({ chartData, newAccountsTotal, hasHistory, links, percentualChange }: Props) => {
```
lvillen, line 37: I think we might be missing some typing here, right? We should specify NewAccountsWidget is a React Component, don't we?

app/javascript/src/Dashboard/components/PotentialUpgradesWidget.jsx
```diff
@@ -0,0 +1,64 @@
...
+    url: string
+  }>,
+  incorrectSetUp: boolean,
+  links: {
+    adminServiceApplicationPlans: string,
+    settingsAdminService: string
+  }
+}
+
+const PotentialUpgradesWidget = ({ violations, incorrectSetUp, links }: Props) => {
```
lvillen, line 25: Same as with NewAccountsWidget, shouldn't we specify this is a React Component?

app/controllers/provider/admin/dashboards_controller.rb
thalesmiguel: I'd prefer to leave as it is so that the order of the helper methods and the order of the method definitions are the same.

app/javascript/src/Dashboard/components/NewAccountsWidget.jsx
thalesmiguel: Right! But what the what now 😅?
What do you mean by "We should specify NewAccountsWidget is a React Component"?

app/javascript/src/Dashboard/components/NewAccountsWidget.jsx
```diff
@@ -0,0 +1,79 @@
...
+  const [isLoading, setIsLoading] = useState(true)
+
+  useEffect(() => {
+    renderChartWidget(document.getElementById('dashboard-widget-new_accounts'), chartData)
+    setIsLoading(false)
+  }, [])
+
+  return (
+    <article className='DashboardWidget' id='dashboard-widget-new_accounts'>
```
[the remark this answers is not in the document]
thalesmiguel: Updated in cf8f31b.

app/javascript/src/Dashboard/components/NewAccountsWidget.jsx
```diff
@@ -0,0 +1,79 @@
...
+      <div className='Dashboard-chart c3' data-chart/>
+      <header className='DashboardWidget-badge'>
+        <h1 className='DashboardWidget-title'>
+          <a className='DashboardWidget-link' href="">
+            <strong data-title-count="true">{newAccountsTotal}</strong> Signups
+          </a>
+        </h1>
+        <div className='DashboardWidget-percentageInfo' data-toggle-visibility='true'>
+          {
+            !hasHistory
```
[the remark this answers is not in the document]
thalesmiguel: Agreed! Updated in cf8f31b.

app/javascript/packs/dashboard_new_accounts_widget.js
thalesmiguel: Cool! And how do we do that? Is there an example in code I can check out?

app/javascript/packs/dashboard_new_accounts_widget.js
```diff
@@ -0,0 +1,17 @@
+import { NewAccountsWidgetWrapper } from 'Dashboard/components/NewAccountsWidget'
```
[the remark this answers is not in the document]
thalesmiguel: Updated in cf8f31b.

app/javascript/src/Dashboard/components/NewAccountsWidget.jsx
lvillen: Well, there is a way to type React Components, but I see that what we usually do is [type the render](https://flow.org/en/docs/react/types/#toc-react-node). However, I don't see any Dashboard components with this typing. In any case, the way we do it is the following:

```suggestion
const NewAccountsWidget = ({ chartData, newAccountsTotal, hasHistory, links, percentualChange }: Props): React.Node => {
```

thalesmiguel: Got it! Updated in [4eac31d](https://github.com/3scale/porta/pull/3070/commits/4eac31d5a2412d7ae81a6c241818037650e07468). Thank you!

app/javascript/src/Dashboard/components/NewAccountsWidget.jsx
```diff
@@ -0,0 +1,79 @@
...
+    setIsLoading(false)
+  }, [])
+
+  return (
+    <article className="DashboardWidget" id="dashboard-widget-new_accounts">
+      {isLoading && <Spinner size="md" isSVG className="pf-u-ml-md DashboardWidget-spinner" />}
+      <div className="Dashboard-chart c3" data-chart/>
+      <header className="DashboardWidget-badge">
+        <h1 className="DashboardWidget-title">
+          <a className="DashboardWidget-link" href="">
```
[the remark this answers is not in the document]
thalesmiguel: Ah yes! And a ternary was inverted as well. Fixed both in c35cdc4.

app/javascript/src/Dashboard/components/PotentialUpgradesWidget.jsx
```diff
@@ -0,0 +1,64 @@
...
+              violations.map(violation =>
+                <li className='DashboardWidgetList-item' key={violation.account_id}>
+                  <a href={violation.url} className='DashboardWidgetList-link'>
+                    {violation.account_name} has {violation.alerts_count} { pluralize('Usage Limit alert', violation.alerts_count) }
+                  </a>
+                </li>
+            )}
+          </ol>
+          : (incorrectSetUp)
+            ? <>
```
josemigallas, line 48: No need for a fragment here, wrap with parenthesis instead.

josemigallas (commented):
Overall it looks pretty good, although I have not reviewed line by line, I would like to test this in my local and play a bit with it.

Besides some minor comments, the new React components are missing tests and unfortunately it seems they're gonna be complex due to the amount of props and inner logic but they shouldn't be hard to implement, just tedious (I would recommend taking NewApplication specs as the example to follow...)

I see there's a bunch of features that mention "dashboard widget" which is good, although I would take a look and try to add a couple more scenarios, if needed.

app/javascript/packs/dashboard_new_accounts_widget.js
josemigallas: Actually, let's favor this other one, better for debugging https://github.com/3scale/porta/blob/01bf13b1a866bde19075b2f3c341def9a7c6427d/app/javascript/packs/backend_api_metrics_index.js#L13-L15

app/javascript/src/Dashboard/components/PotentialUpgradesWidget.jsx
thalesmiguel: I tried the parenthesis out but I got some nasty errors...

app/javascript/packs/dashboard_potential_upgrades_widget.js
```diff
@@ -0,0 +1,15 @@
+import { PotentialUpgradesWidgetWrapper } from 'Dashboard/components/PotentialUpgradesWidget'
```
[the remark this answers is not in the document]
thalesmiguel: Added in 9015281.

app/javascript/packs/dashboard_new_accounts_widget.js
thalesmiguel: Nice stuff!
Added in 73c8a61.

app/javascript/src/Dashboard/components/PotentialUpgradesWidget.jsx
```diff
@@ -0,0 +1,64 @@
...
+  }>,
+  incorrectSetUp: boolean,
+  links: {
+    adminServiceApplicationPlans: string,
+    settingsAdminService: string
+  }
+}
+
+const PotentialUpgradesWidget = ({ violations, incorrectSetUp, links }: Props): React.Node => {
+  return (
```
josemigallas, line 26:
```suggestion
const PotentialUpgradesWidget = ({ violations, incorrectSetUp, links }: Props): React.Node => (
```

app/javascript/src/Dashboard/components/PotentialUpgradesWidget.jsx
josemigallas: I see. In any case, this JSX is a bit hard to read and I would extract a couple of components from there. For instance I would extract the item into a `PotentialUpgradeWidgetItem` component.

app/javascript/src/Dashboard/components/PotentialUpgradesWidget.jsx
```diff
@@ -0,0 +1,64 @@
...
+    <article className='DashboardWidget' id='dashboard-widget-potential_upgrades'>
+      <h1 className='DashboardWidget-title'>
+        Potential Upgrades
+      </h1>
+      <span className='DashboardWidget-intro'>
+        Accounts that hit their Usage Limits in the last 30 days
+      </span>
+
+      {
+        violations.some((violation) => violation)
```
josemigallas, line 36:
```suggestion
        violations.length > 0
```

app/javascript/src/Dashboard/components/PotentialUpgradesWidget.jsx
```diff
@@ -0,0 +1,64 @@
...
+            {
+              violations.map(violation =>
+                <li className='DashboardWidgetList-item' key={violation.account_id}>
+                  <a href={violation.url} className='DashboardWidgetList-link'>
+                    {violation.account_name} has {violation.alerts_count} { pluralize('Usage Limit alert', violation.alerts_count) }
+                  </a>
+                </li>
+            )}
+          </ol>
+          : (incorrectSetUp)
```
josemigallas, line 47: Seems to me a guard condition... why check this only if violations is empty? I assume it was like that before but we could improve it, maybe.
IMO `violations` should be `nil` if not setup correctly and the guard could be something like.
```jsx
if (incorrectSetUp) {
  return (
    <>
      <p> In order to show... </p>
      <p> ... </p>
    </>
  )
}

return ( ... )

File: app/javascript/src/Dashboard/components/PotentialUpgradesWidget.jsx (hunk 12 above)
Comment (thalesmiguel) [reply app/javascript/src/Dashboard/components/PotentialUpgradesWidget.jsx]: Changed in b89bf8e.

File: app/javascript/src/Dashboard/components/PotentialUpgradesWidget.jsx (hunk 11 above)
Comment (thalesmiguel) [reply app/javascript/src/Dashboard/components/PotentialUpgradesWidget.jsx]: Changed in b89bf8e.

File: app/javascript/src/Dashboard/components/PotentialUpgradesWidget.jsx (hunk 9 above)
Comment (thalesmiguel) [reply app/javascript/src/Dashboard/components/PotentialUpgradesWidget.jsx]: Changed in b89bf8e.

File: app/javascript/src/Dashboard/components/PotentialUpgradesWidget.jsx (hunk 13 above)
Comment (thalesmiguel) [reply app/javascript/src/Dashboard/components/PotentialUpgradesWidget.jsx]: I changed the order in b89bf8e but, since there's an elsif and an else, I didn't manage to scape the ternary.
Should I create a function that receives **all** arguments, check what should be returned and then call it in the JSX? Is there a better way to do it? 😬

File: spec/javascripts/Dashboard/components/NewAccountsWidget.spec.jsx [hunk 14]
```diff
@@ -0,0 +1,80 @@
...
+
+it('should render', () => {
+  const wrapper = mountWrapper(defaultProps)
+  expect(wrapper.exists()).toBe(true)
+})
+
+describe('when there is historical data', () => {
+  const wrapper = mountWrapper(defaultProps)
+
+  it('should render past data', () => {
```
lvillen, line 60: Since when `hasHistory` have two options, I'd say we could test both options: `u-plus` and `u-minus`.

spec/javascripts/Dashboard/components/NewAccountsWidget.spec.jsx
```diff
@@ -0,0 +1,80 @@
...
+
+const mountWrapper = (props) => mount(<NewAccountsWidget {...{ ...defaultProps, ...props }}/>)
+
+it('should render', () => {
+  const wrapper = mountWrapper(defaultProps)
+  expect(wrapper.exists()).toBe(true)
+})
+
+describe('when there is historical data', () => {
+  const wrapper = mountWrapper(defaultProps)
```
lvillen, line 58: I'd include this line (and applies to every case where the `wrapper` is built this way) inside every `test` rather than inside the `describe` or I'd use a `beforeEach` to create the `wrapper` and an `afterEach` to clean it since we can't be sure if these tests will grow in complexity some time in the future.

spec/javascripts/Dashboard/components/PotentialUpgradesWidget.spec.jsx
```diff
@@ -0,0 +1,62 @@
...
+    expect(wrapper.text().includes('In order to show Potential Upgrades, add 1 or more usage limits to')).toBe(true)
+    expect(wrapper.text().includes('Web Alerts for Admins of this Account of 100%')).toBe(true)
+  })
+})
+
+describe('when the setup is correct', () => {
+  const correctSetUpProps = { ...defaultProps, incorrectSetUp: false }
+
+  describe('and there are violations', () => {
+    const wrapper = mountWrapper(correctSetUpProps)
```
lvillen, line 46: Same as with the other `spec`: I'd include this line (and applies to every case where the `wrapper` is built this way) inside every `test` rather than inside the `describe` or I'd use a `beforeEach` to create the `wrapper` and an `afterEach` to clean it.

spec/javascripts/Dashboard/components/PotentialUpgradesWidget.spec.jsx
```diff
@@ -0,0 +1,62 @@
+// @flow
+
+import React from 'react'
+
+import { PotentialUpgradesWidget } from 'Dashboard/components/PotentialUpgradesWidget'
+import { mount } from 'enzyme'
+
+let props
+const defaultProps = {
```
lvillen, line 9: Missing Props typings.

spec/javascripts/Dashboard/components/PotentialUpgradesWidget.spec.jsx
```diff
@@ -0,0 +1,62 @@
...
+    }
+  ],
+  incorrectSetUp: true,
+  links: {
+    adminServiceApplicationPlans: '/apiconfig/services/1/application_plans',
+    settingsAdminService: '/apiconfig/services/1/settings#web_provider'
+  }
+}
+
+const mountWrapper = (props) => mount(<PotentialUpgradesWidget {...{ ...defaultProps, ...props }}/>)
```
lvillen, line 26: Missing types.

spec/javascripts/Dashboard/components/NewAccountsWidget.spec.jsx
```diff
@@ -0,0 +1,80 @@
+// @flow
+
+import React from 'react'
+
+import { NewAccountsWidget } from 'Dashboard/components/NewAccountsWidget'
+import { mount } from 'enzyme'
+
+let props
+const defaultProps = {
```
lvillen, line 9: Missing Props type.

spec/javascripts/Dashboard/components/NewAccountsWidget.spec.jsx
```diff
@@ -0,0 +1,80 @@
...
+    },
+    lastDayInRangeAdminBuyersAccount: {
+      url: '/buyers/accounts?search%5Bcreated_within%5D%5B%5D=2022-01-03&search%5Bcreated_within%5D%5B%5D=2022-01-03',
+      value: '0'
+    }
+  },
+  percentualChange: 50
+}
+
+const mountWrapper = (props) => mount(<NewAccountsWidget {...{ ...defaultProps, ...props }}/>)
```
lvillen, line 50: Missing types.

thalesmiguel: Added in 4b1b900.

spec/javascripts/Dashboard/components/NewAccountsWidget.spec.jsx
thalesmiguel: Added in 4b1b900.

spec/javascripts/Dashboard/components/PotentialUpgradesWidget.spec.jsx
thalesmiguel: Added in 4b1b900.

spec/javascripts/Dashboard/components/PotentialUpgradesWidget.spec.jsx
thalesmiguel: Added in 4b1b900.

spec/javascripts/Dashboard/components/PotentialUpgradesWidget.spec.jsx
thalesmiguel: Good! Added in 4b1b900.

spec/javascripts/Dashboard/components/NewAccountsWidget.spec.jsx
thalesmiguel: I ended up adding it for each test in 4b1b900.

spec/javascripts/Dashboard/components/NewAccountsWidget.spec.jsx
thalesmiguel: Good! Added in 4b1b900.

spec/javascripts/Dashboard/components/PotentialUpgradesWidget.spec.jsx
```diff
@@ -0,0 +1,61 @@
...
+
+    expect(wrapper.text().includes('In order to show Potential Upgrades, add 1 or more usage limits to')).toBe(true)
+    expect(wrapper.text().includes('Web Alerts for Admins of this Account of 100%')).toBe(true)
+  })
+})
+
+describe('when the setup is correct', () => {
+  describe('and there are violations', () => {
+    it('should render a list of violations', () => {
+      let props: Props = { ...defaultProps, incorrectSetUp: false }
```
lvillen, line 45: As said in the previous file, I suggest changing `let` for `const` since these props won't be changing:

```suggestion
      const props: Props = { ...defaultProps, incorrectSetUp: false }
```

spec/javascripts/Dashboard/components/NewAccountsWidget.spec.jsx
```diff
@@ -0,0 +1,96 @@
...
+    it('should render past data', () => {
+      const wrapper = mountWrapper(defaultProps)
+
+      expect(wrapper.find('.DashboardWidget-link.u-plus').exists()).toBe(true)
+    })
+  })
+
+  describe('and the percentual change is negative or 0', () => {
+    it('should render past data', () => {
+      let props: Props = { ...defaultProps, percentualChange: 0 }
```
lvillen, line 68: Since these props won't change, I'd suggest using `const` instead of `let`. Applies to all `let props`.

```suggestion
      const props: Props = { ...defaultProps, percentualChange: 0 }
```

spec/javascripts/Dashboard/components/PotentialUpgradesWidget.spec.jsx
```diff
@@ -0,0 +1,61 @@
...
+      const wrapper = mountWrapper(props)
+
+      expect(wrapper.find('.DashboardWidgetList').exists()).toBe(true)
```
lvillen, line 48: Is difficult for me to tell but it seems something might be happening here with spaces / indentation. I'd say is worth to take a look.

spec/javascripts/Dashboard/components/PotentialUpgradesWidget.spec.jsx
```diff
@@ -0,0 +1,61 @@
...
+      expect(wrapper.find('.DashboardWidgetList-item').exists()).toBe(true)
+    })
+  })
+
+  describe('and there are no violations', () => {
+    let props: Props = { ...defaultProps, incorrectSetUp: false, violations: 0 }
+    const wrapper = mountWrapper(props)
```
lvillen, line 55: I suggest to include lines 54 and 55 inside the test and would change `let` for `const`.

thalesmiguel: Updated in bf00c23.

spec/javascripts/Dashboard/components/NewAccountsWidget.spec.jsx
thalesmiguel: Updated in bf00c23.

spec/javascripts/Dashboard/components/PotentialUpgradesWidget.spec.jsx
thalesmiguel: Updated in bf00c23.

thalesmiguel:
Added 63f5633 to remove unused code from the "old dashboard". Those look to be leftovers from https://github.com/3scale/porta/pull/2234.

thalesmiguel:
Tested on a test cluster. Things look good.
There's also a BUG fix in here, where the total amount of new accounts in the period is displayed incorrectly.
For a total of 5 new accounts in the period:

(Before)
Screenshot 2022-11-02 at 10 32 57

(After)
Screenshot 2022-11-02 at 10 32 10

app/javascript/src/Dashboard/components/NewAccountsWidget.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+  useEffect(() => {
+    renderChartWidget(chartWidget, chartData)
+    setIsLoading(false)
+  }, [])
+
+  return (
+    <article className="DashboardWidget" id="dashboard-widget-new_accounts">
+      {isLoading && <Spinner className="pf-u-ml-md DashboardWidget-spinner" size="md" />}
+      <div data-chart className="Dashboard-chart c3" />
```
josemigallas, line 82: I don't think `data-chart` is used anywhere.

app/views/provider/admin/dashboards/widgets/_chart.html.erb
```diff
@@ -1,7 +0,0 @@
-<div class="Dashboard-chart" data-chart>
-  <% if widget.loaded? %>
```
josemigallas, line 2: After taking a deeper look at how widgets currently work, I think `loaded` just means whether the data is available or not. I don't see any indication of an async process, so rendering a spinner like in https://github.com/3scale/porta/blob/7f8fcd9276bd3708a10bbda07e9f14ffd4301985/app/views/provider/admin/dashboards/widgets/_potential_upgrades.html.slim#L26 does little sense to me. To me it would be more appropriate to show a "data not available" sign instead.

app/javascript/src/Dashboard/components/NewAccountsWidget.tsx
```diff
@@ -0,0 +1,116 @@
...
+interface LastDayInRangeAdminBuyersAccount {
+  url: string;
+  value: string;
+}
+
+interface Links {
+  previousRangeAdminBuyersAccount: PreviousRangeAdminBuyersAccount;
+  currentRangeAdminBuyersAccount: CurrentRangeAdminBuyersAccount;
+  lastDayInRangeAdminBuyersAccount: LastDayInRangeAdminBuyersAccount;
+}
```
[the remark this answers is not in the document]
thalesmiguel: Donno... I think there was a comment in this PR to split things into interfaces. I can put it back into an one piece.

app/javascript/src/Dashboard/components/PotentialUpgradesWidget.tsx
```diff
@@ -0,0 +1,72 @@
...
+                // eslint-disable-next-line react/jsx-key
+                <li className="DashboardWidgetList-item">
+                  <a className="DashboardWidgetList-link" href={violation.url}>
+                    {violation.accountName} has {violation.alertsCount} { pluralize('Usage Limit alert', violation.alertsCount) }
+                  </a>
+                </li>
+              )
+            )}
+          </ol>
+        )
```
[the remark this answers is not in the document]
thalesmiguel: Heeeey... One elsif shouldn't categorize as hadouken 😆

PR #3075, opened 2022-10-12
[WIP] ⚰️ remove core-js from code and dependencies

lvillen:

Removes `core-js` and `babel/polyfills` from dependencies and removes all references and uses of them inside the code. This will suppose a 12MB lighter bundle. 

This PR intends to make a lighter bundle and a more maintainable code removing libraries that are not really used with modern browsers.

**Verification steps** 


.babelrc
```diff
@@ -1,27 +1,10 @@
 {
   "plugins": [
     "@babel/plugin-syntax-dynamic-import",
-    "@babel/plugin-transform-runtime",
-    "@babel/plugin-proposal-class-properties",
-    "@babel/plugin-transform-template-literals",
-    "@babel/plugin-proposal-object-rest-spread",
-    "@babel/plugin-transform-destructuring"
+    "@babel/plugin-transform-runtime"
```
josemigallas, line 4: Why do we need `plugin-transform-runtime` and `plugin-syntax-dynamic-import`?

.babelrc
```diff
@@ -1,27 +1,10 @@
...
-        "targets": {
-          "node": "current",
-          "ie": "11",
-          "firefox": "67",
-          "chrome": "75",
-          "edge": "44"
-        }
-      }
-    ],
+    "@babel/preset-env",
```
josemigallas, line 7: Why removing present-env config? I think we could target core-js 3 now

josemigallas (changes requested):
Licenses file needs to be updated

josemigallas:
Let's target only supported browsers to `.babelrc`:
```diff
 {
+  "targets": {
+    // From "Red Hat 3scale API Management Supported Configurations" https://access.redhat.com/articles/2798521
+    "chrome": "101",
+    "firefox": "100",
+    "edge": "101"
+  },
   "plugins": [
     "@babel/plugin-syntax-dynamic-import",
     "@babel/plugin-transform-runtime"
   ],
  ```

File: .babelrc (hunk 2 above)
Comment (josemigallas) [reply .babelrc:7]: `targets` can be moved to the top level (I left a comment) but we still need useBuiltIns and corejs:
```suggestion
    [
      "@babel/preset-env",
      {
        "useBuiltIns": "entry",
        "corejs": "3.25"
      }
    ],
```

josemigallas:
Just for the record, this PR only removes redundant dependencies. `core-js` is still used by `babel/preset-env` but we won't need to manage it.

.babelrc
lvillen: All the other plugins we had there are already included at the `@babel/preset-env`, those two aren't. But let me give some details:

About `plugin-transform-runtime`, this one, according to [Babel docs](https://babeljs.io/docs/en/babel-plugin-transform-runtime), avoid helpers duplication and I think it's a good idea to keep it.

About `plugin-syntax-dynamic-import`, this is the one that allows us to use `import` syntax and though [Babel docs](https://babeljs.io/docs/en/babel-plugin-syntax-dynamic-import) says it's included at `preset-env`, I ran a small script to check which plugins were included and this one wasn't.

.babelrc
lvillen: Well, we can discuss the `preset-env` config, but with these changes the idea is to let `preset-env` handle things with core-js instead of having it installed. 

As it is right now, `preset-env` is following browserslist default configuration, which is: `> 0.5%, last 2 versions, Firefox ESR, not dead`. More information about this can be found [here](https://github.com/browserslist/browserslist#queries). Maybe we could discuss which browsers we want to target.

PR #3077, opened 2022-10-19
Fix Braintree form

josemigallas:
Crappy solution. We leave the "good one" for when we have time.
 
> **What this PR does / why we need it**:
> 
> The JS at Braintree edit form is faulty, probably due to inconsistent jQuery version or unexpected upgrade.
> 
> The Braintree js SDK is also being downloaded via CDN, even though it's already included as a NPM package. For that reason the JS has been moved into a pack (to be handled by webpack and have access to NPM modules).
> 
> Moreover, this form is already implemented by React and used in the Dev portal but this this bug is a blocker we're not gonna spend time refactoring the whole page.
> 
> **Which issue(s) this PR fixes** 
> 
> [THREESCALE-8808: SaaS customers not able to add/edit Payment details](https://issues.redhat.com/browse/THREESCALE-8808)
> 
> **Special notes for your reviewer**:
> Braintree test data in Bitwarden.
> 
> https://user-images.githubusercontent.com/11672286/196646555-c0184fce-67bb-42e0-b615-6ceb48281107.mp4
> 
>

### DIFF
--- a/app/views/provider/admin/account/payment_gateways/braintree_blue/edit.html.slim
+++ b/app/views/provider/admin/account/payment_gateways/braintree_blue/edit.html.slim
@@ -94,11 +94,12 @@ p
   = form.actions do
     = form.commit_button 'Save credit card'
 
+/ TODO: All is duplicated. Move this into a pack and use braintree from node_modules
 script src="https://js.braintreegateway.com/web/3.69.0/js/client.min.js"
 script src="https://js.braintreegateway.com/web/3.69.0/js/hosted-fields.min.js"
 javascript:
-  var $form = $('#new_customer');
-  var submit = document.querySelector('input[type="submit"]');
+  const form = document.querySelector('form#new_customer');
+  const submit = document.querySelector('button[type="submit"]');
 
   if (typeof braintree !== 'undefined') {
     braintree.client.create({
@@ -147,17 +148,16 @@ javascript:
         }
 
         submit.removeAttribute('disabled');
-
-        $form.on('submit', function(event) {
+        form.addEventListener('submit', function(event) {
           event.preventDefault();
+          event.stopPropagation();
           hostedFieldsInstance.tokenize(function(tokenizeErr, payload) {
             if (tokenizeErr) {
               console.error(tokenizeErr);
               return;
             }
-
-            $('#braintree_nonce').val(payload['nonce']);
-            $form.get(0).submit();
+            document.querySelector('input#braintree_nonce').setAttribute('value', payload.nonce);
+            form.submit();
           });
         });
       });


app/javascript/packs/braintree_edit_form.js
```diff
@@ -0,0 +1,34 @@
...
+  const client = await createBraintreeClient(braintree.client, authorization)
+  const instance = await createHostedFieldsInstance(braintree.hostedFields, client, hostedFieldOptions)
+
+  submit.removeAttribute('disabled')
+  form.addEventListener('submit', (event: Event) => {
+    event.preventDefault()
+    event.stopPropagation()
+    instance.tokenize((tokenizeErr, payload) => {
+      if (tokenizeErr) {
+        error('Some of the credit card details are wrong. Please update them.')
```
thalesmiguel, line 28: We're setting a fixed error message in here instead of using `tokenizeErr`, as in the past. Why is that?

josemigallas: `tokenizeErr` is an object containing info too cryptic for users. On the other hand, I opted for showing an alert in the UI because right now the page fails without any kinds of feedback. While playing with it I realized there's only 2 errors Braintree will return:
* All your fields are empty
* Some of your fields are wrong
But of course it only refers to the credit card fields, which are the ones being handled by Braintree. Simply printing the "all your fields are wrong" built-in message will be very confusing for the user.

thalesmiguel (dismissed):
🚀

lvillen (approved):

app/views/provider/admin/account/payment_gateways/braintree_blue/edit.html.slim
```diff
@@ -94,11 +94,12 @@ p
   = form.actions do
     = form.commit_button 'Save credit card'
 
+/ TODO: All is duplicated. Move this into a pack and use braintree from node_modules
```
thalesmiguel, line 97: If memory serves right, this single bar is a comment in `Slim`. Still weird to me 😬

lvillen (approved):

app/views/provider/admin/account/payment_gateways/braintree_blue/edit.html.slim
```diff
@@ -94,11 +94,12 @@ p
...
 script src="https://js.braintreegateway.com/web/3.69.0/js/client.min.js"
 script src="https://js.braintreegateway.com/web/3.69.0/js/hosted-fields.min.js"
 javascript:
-  var $form = $('#new_customer');
-  var submit = document.querySelector('input[type="submit"]');
+  const form = document.querySelector('form#new_customer');
+  const submit = document.querySelector('button[type="submit"]');
```
thalesmiguel, line 102: Maybe dumb question, but are we selecting the same element as before in here?

thalesmiguel (commented):
Couple questions there 😬

app/views/provider/admin/account/payment_gateways/braintree_blue/edit.html.slim
josemigallas: Comment, correct

app/views/provider/admin/account/payment_gateways/braintree_blue/edit.html.slim
josemigallas: Same elements just more specific selectors.

thalesmiguel (approved):

PR #3079, opened 2022-10-19
🚧🧠🌴 Fix Braintree 3DS in Dev portal (and much more)

josemigallas:
Related:
* https://github.com/3scale/porta/pull/3167
* https://github.com/3scale/porta/pull/3176
* https://github.com/3scale/porta/pull/3206

[THREESCALE-8967: Adding Braintree SCA card fails](https://issues.redhat.com/browse/THREESCALE-8967)

We load braintree-web SDK from node_modules AND as a cdn. Besides, the whole braintree flow is implemented in a convoluted way and it's hard to maintain.

First, this PR cleans up the pack used in developer portal that renders a React component which uses braintree-web imported from node_modules.

Second, it removes the dependency from braintre-web CDN by moving the script from admin portal's view into a pack and importing the SDK from node_nodules.

**TODO**:
- [x] 3D secure flow not working at the moment

**verification**
* Admin portal: http://provider-admin.lvh.me:3000/p/admin/account/braintree_blue/edit
* Dev portal: http://provider.lvh.me:3000/admin/account/braintree_blue/edit

Testing credit cards:

with 3DS disabled | 4111111111111111
-- | --
3DS successful with no challenge | 4000000000001000
3DS successful with challenge | 4000000000001091
3DS unsuccessful with challenge | 4000000000001109

**NOTE**: both forms: payment details (admin portal) and credit card details (dev portal), are basically the same. They share the same logic and have the same fields. However, one is renders via formtastic in a slim template ([edit.html.slim](https://github.com/3scale/porta/blob/29089e62578e0105a14ffb568d0f030b6fde1977/app/views/provider/admin/account/payment_gateways/braintree_blue/edit.html.slim#L1)) whereas the other is a React component ([BraintreeForm.tsx](https://github.com/3scale/porta/blob/29089e62578e0105a14ffb568d0f030b6fde1977/app/javascript/src/PaymentGateways/braintree/BraintreeForm.tsx#L1)). The only problem is mainly the differences between class names. Example:

Admin portal (PF4 style)
```html
<li class="string input required stringish" id="customer_first_name_input">
  <label for="customer_first_name" class="label">First name<abbr title="required">*</abbr></label>
  <input id="customer_first_name" type="text" value="Pepe" name="customer[first_name]">
</li>
```

Dev portal (dev portal style)
```html
<li class="string optional form-group" id="customer_first_name_input">
  <label class="col-md-4 control-label" for="customer_first_name">First name *</label>
  <input required="" class="col-md-6 form-control" id="customer_first_name" name="customer[first_name]" value="Pepe">
</li>
```

This to me looks like repetition and we should be able to use React in both (or none..) but it goes out of scope of this PR.


akostadinov:
Can't really review but looks much better than before.

app/views/provider/admin/account/payment_gateways/braintree_blue/edit.html.slim
```diff
@@ -92,74 +92,7 @@ p
...
-            'color': 'red'
-          },
-          'input.valid': {
-            'color': 'green'
-          }
-        },
-        fields: {
-          number: {
-            selector: '#customer_credit_card_number',
-            placeholder: '4111 1111 1111 1111'
```
akostadinov, line 133: does this come from automatically now?

jlledom (dismissed):

app/javascript/src/PaymentGateways/braintree/useBraintreeHostedFields.ts
```diff
@@ -0,0 +1,69 @@
+import { useEffect, useState } from 'react'
+
+import { create3DSecure, getClient, getHostedFields } from 'PaymentGateways/braintree/braintree'
+
+import type { BillingAddress } from 'PaymentGateways/braintree/types'
+import type { HostedFields, HostedFieldsFieldDataFields } from 'braintree-web/modules/hosted-fields'
+
+type CustomHostedFields = HostedFields & {
+  getNonce: (BillingAddress: BillingAddress) => Promise<string>;
```
josemigallas, line 9:
```suggestion
  getNonce: (billingAddress: BillingAddress) => Promise<string>;
```

app/javascript/src/PaymentGateways/braintree/utils/formValidation.ts
```diff
@@ -0,0 +1,51 @@
...
+
+const validateForm = (formRef: unknown): Record<keyof typeof VALIDATION_CONSTRAINTS, string[]> | undefined => {
+  const formErrors = validate(formRef, VALIDATION_CONSTRAINTS) as (Record<keyof typeof VALIDATION_CONSTRAINTS, string[]> | undefined)
+  return formErrors
+}
+
+// const validateSingleField = (value: unknown, type: keyof typeof VALIDATION_CONSTRAINTS): string[] | undefined => {
+//   const fieldError = validate.single(value, VALIDATION_CONSTRAINTS[type]) as (string[] | undefined)
+//   return fieldError
+// }
```
josemigallas, line 46: oops

app/views/provider/admin/account/payment_gateways/braintree_blue/edit.html.slim
josemigallas: Now it comes from https://github.com/3scale/porta/pull/3079/files#diff-e5280eb9a5ca2994755960970425ea7ff092eef504a1a954682533ea1d1cc75c and it's the same as in the Dev portal form.
This was copy-pasted from a Braintree docs example.

app/views/provider/admin/account/payment_gateways/braintree_blue/edit.html.slim
```diff
@@ -92,74 +92,7 @@ p
...
-              return;
-            }
-            document.querySelector('input#braintree_nonce').setAttribute('value', payload.nonce);
-            form.submit();
-          });
-        });
-      });
-    })
-  }
+span#braintree_data data-authorization="#{j @braintree_authorization}"
```
josemigallas, line 97: TODO: move this to the helper

app/javascript/src/utilities/test-utils.ts
```diff
@@ -86,13 +86,23 @@ function assertInputs (wrapper: ReactWrapper<unknown>, inputs: { id: string; pre
...
 
+/**
+ * Use when mounting a component with promises or side effects, such as BraintreeForm#onSubmit
+ */
+async function waitForPromises (): Promise<void> {
+  await act(async () => {
+    await new Promise((resolve) => setTimeout(resolve))
+  })
+}
+
```
jlledom, line 97: May you explain this? I don't understand it. AIUI this won't return a `Promise` b/c the only call in the function is preceded by an `await`. Besides, there's no `return` keyword. What does `act()` do?

josemigallas: `async / await` is syntactic sugar for wrapping something with a promise. Not returning means the method returns `undefined`. But if method awaits something inside, then it has to be marked as async and it will wrap this undefined inside a Promise, therefore it will return `Promise<void>`. In this scenario, a `return` statement is redundant.

Does that make sense?

jlledom: Yes, but still don't understand the code. What does `act()` do? and why calling `setTimeout`?

jlledom (dismissed):
I don't feel able to review all the TS code, so I cloned the branch and tried to make it work locally. Everything seems to work fine for me, so approved.

app/javascript/src/utilities/test-utils.ts
josemigallas: Anything that triggers a change in the component has to be wrapper around [act](https://reactjs.org/docs/test-utils.html#act).

`setTimeout` is used to add a delay to the promise so that it doesn't resolve immediately.

PR #3086, opened 2022-10-25
THREESCALE-6851: Stripe webhook: Acknowledge reception even for payments we don't care about

jlledom:

When a client is using Stripe to accept payments from other applications besides 3scale, all payment events are sent through all webhooks, so we receive them even when we don't care. So far, we were returning a 404 in that situation as the payment intent was not in our DB, but that's causing problems to clients since Stripe disables a webhook if it returns 404 too many times. This PR makes it return a 204 to acknowledge the reception in order to prevent Stripe to disable the webhook.

[THREESCALE-6851](https://issues.redhat.com/browse/THREESCALE-6851)

**Verification steps** 

The easiest way to generate a stripe payment out of 3scale is to use the [stripe  cli](https://stripe.com/docs/stripe-cli) tool.

1. Log into the Stripe dashboard
2. Add a [Local listener](https://dashboard.stripe.com/test/webhooks/create?endpoint_location=local):
```
stripe listen --events payment_intent.succeeded --forward-to http://provider-admin.3scale.localhost:3000/api/payment_callbacks/stripe_callbacks?access_token=ACCESS_TOKEN
```
4. From another terminal, send a `payment_intent.succeeded` event:
```
stripe trigger payment_intent.succeeded
```
6. A `[204]` code should be returned

### DIFF
--- a/app/controllers/finance/api/payment_callbacks/stripe_callbacks_controller.rb
+++ b/app/controllers/finance/api/payment_callbacks/stripe_callbacks_controller.rb
@@ -21,8 +21,8 @@ class Finance::Api::PaymentCallbacks::StripeCallbacksController < Finance::Api::
 
   # Undocumented endpoint used for update callbacks of async-authorized payment transactions (mostly due to SCA regulations)
   def create
-    service = begin
-      Finance::StripePaymentIntentUpdateService.new(current_account, stripe_event)
+    begin
+      service =  Finance::StripePaymentIntentUpdateService.new(current_account, stripe_event)
     rescue ActiveRecord::RecordNotFound
       # Returning 204 to acknowledge reception even for payments we don't care about.
       #
@@ -34,10 +34,10 @@ class Finance::Api::PaymentCallbacks::StripeCallbacksController < Finance::Api::
       # too often.
       #
       # https://issues.redhat.com/browse/THREESCALE-6851
-      nil
+      return head(:no_content)
     end
 
-    return head(:no_content) if service.blank? || service.call
+    return head(:no_content) if service.call
 
     exception = StripeCallbackError.new('Cannot update Stripe payment intent')
     report_error(exception, event: stripe_event, payment_intent: service.payment_intent)
--- a/app/controllers/finance/api/payment_callbacks/stripe_callbacks_controller.rb
+++ b/app/controllers/finance/api/payment_callbacks/stripe_callbacks_controller.rb
@@ -24,7 +24,15 @@ class Finance::Api::PaymentCallbacks::StripeCallbacksController < Finance::Api::
     service = begin
       Finance::StripePaymentIntentUpdateService.new(current_account, stripe_event)
     rescue ActiveRecord::RecordNotFound
-      # Returning 204 to acknowledge reception even for payments we don't care about
+      # Returning 204 to acknowledge reception even for payments we don't care about.
+      #
+      # There are some clients who are using their Stripe account not only for 3Scale,
+      # but also for other services they provide. When they receive any payment, Stripe
+      # will call all webhooks, no matter where the payment comes from, and we'll
+      # receive a call for a payment not managed by us. Stripe expects us to return 204
+      # in this situation, since they will remove the webhook if it returns error codes
+      # too often.
+      #
       # https://issues.redhat.com/browse/THREESCALE-6851
       nil
     end
--- a/app/controllers/finance/api/payment_callbacks/stripe_callbacks_controller.rb
+++ b/app/controllers/finance/api/payment_callbacks/stripe_callbacks_controller.rb
@@ -26,7 +26,6 @@ class Finance::Api::PaymentCallbacks::StripeCallbacksController < Finance::Api::
     rescue ActiveRecord::RecordNotFound
       # Returning 204 to acknowledge reception even for payments we don't care about
       # https://issues.redhat.com/browse/THREESCALE-6851
-      Rails.logger.info("Acknowledging reception: '#{stripe_event.id}'")
       return head(:no_content)
     end
 
--- a/app/controllers/finance/api/payment_callbacks/stripe_callbacks_controller.rb
+++ b/app/controllers/finance/api/payment_callbacks/stripe_callbacks_controller.rb
@@ -21,15 +21,15 @@ class Finance::Api::PaymentCallbacks::StripeCallbacksController < Finance::Api::
 
   # Undocumented endpoint used for update callbacks of async-authorized payment transactions (mostly due to SCA regulations)
   def create
-    begin
-      service = Finance::StripePaymentIntentUpdateService.new(current_account, stripe_event)
+    service = begin
+      Finance::StripePaymentIntentUpdateService.new(current_account, stripe_event)
     rescue ActiveRecord::RecordNotFound
       # Returning 204 to acknowledge reception even for payments we don't care about
       # https://issues.redhat.com/browse/THREESCALE-6851
-      return head(:no_content)
+      nil
     end
 
-    return head(:no_content) if service.call
+    return head(:no_content) if service.blank? || service.call
 
     exception = StripeCallbackError.new('Cannot update Stripe payment intent')
     report_error(exception, event: stripe_event, payment_intent: service.payment_intent)
--- a/app/controllers/finance/api/payment_callbacks/stripe_callbacks_controller.rb
+++ b/app/controllers/finance/api/payment_callbacks/stripe_callbacks_controller.rb
@@ -26,7 +26,7 @@ class Finance::Api::PaymentCallbacks::StripeCallbacksController < Finance::Api::
     rescue ActiveRecord::RecordNotFound
       # Returning 204 to acknowledge reception even for payments we don't care about.
       #
-      # There are some clients who are using their Stripe account not only for 3Scale,
+      # There are some clients who are using their Stripe account not only for 3scale,
       # but also for other services they provide. When they receive any payment, Stripe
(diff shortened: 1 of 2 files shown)


test/integration/finance/api/payment_callbacks/stripe_callbacks_controller_test.rb
```diff
@@ -61,11 +61,11 @@ class CreateTest < self
     end
 
     test 'cannot find payment intent' do
-      stripe_event = self.stripe_event(type: 'payment_intent.requires_action', payment_intent_data: { id: 'non-existent-payment-intent-id' })
+      stripe_event = self.stripe_event(type: 'payment_intent.succeeded', payment_intent_data: { id: 'non-existent-payment-intent-id' })
```
jlledom, line 64: I think this test was wrong, it was supposed to check the controller returned a 404 when the payment is not found, but instead of that it was sending a `requires_action` event, which is not allowed. The test passed because not allowed events also return a 404, but the error was being returned before even checking whether the payment existed or not. Now it really checks the payment and returns a 204 when it doesn't exits.

app/controllers/finance/api/payment_callbacks/stripe_callbacks_controller.rb
```diff
@@ -21,7 +21,14 @@ class MissingStripeEndpointSecret < StripeCallbackError; end
 
   # Undocumented endpoint used for update callbacks of async-authorized payment transactions (mostly due to SCA regulations)
   def create
-    service = Finance::StripePaymentIntentUpdateService.new(current_account, stripe_event)
+    begin
+      service = Finance::StripePaymentIntentUpdateService.new(current_account, stripe_event)
+    rescue ActiveRecord::RecordNotFound
+      # Returning 204 to acknowledge reception even for payments we don't care about
+      # https://issues.redhat.com/browse/THREESCALE-6851
+      Rails.logger.info("Acknowledging reception: '#{stripe_event.id}'")
```
mayorova, line 29: As we are not caring about these event, I am not sure we should even print this message in the log.
Or, if we do, I would provide more information, because it would be way too cryptic, I think (reception of what? there would be just some random ID).
I'd probably opt for not printing anything.

In fact, I am not sure we need to wait the result from `service.call` to return `204`...
That's what Stripe says about it: https://stripe.com/docs/webhooks#acknowledge-events-immediately

> Your endpoint must quickly return a successful status code (`2xx`) prior to any complex logic that could cause a timeout. For example, you must return a `200` response before updating a customer’s invoice as paid in your accounting system.

I'd say as soon as we populate `stripe_event` successfully, we can return `204`. What do you think @jlledom ?

jlledom: Then we should call `service.call` in a a background job or another thread? and who is receiving the exception in the case it fails?

app/controllers/finance/api/payment_callbacks/stripe_callbacks_controller.rb
```diff
@@ -21,7 +21,13 @@ class MissingStripeEndpointSecret < StripeCallbackError; end
...
   # Undocumented endpoint used for update callbacks of async-authorized payment transactions (mostly due to SCA regulations)
   def create
-    service = Finance::StripePaymentIntentUpdateService.new(current_account, stripe_event)
+    begin
+      service = Finance::StripePaymentIntentUpdateService.new(current_account, stripe_event)
+    rescue ActiveRecord::RecordNotFound
+      # Returning 204 to acknowledge reception even for payments we don't care about
+      # https://issues.redhat.com/browse/THREESCALE-6851
+      return head(:no_content)
+    end
```
thalesmiguel, line 30: We could remove the double `return head(:no_content)` and the variable assignment inside of the block with something like
```ruby
def create
  service = begin
    Finance::StripePaymentIntentUpdateService.new(current_account, stripe_event)
  rescue ActiveRecord::RecordNotFound
    SomeThing
  end

  return head(:no_content) if `service is SomeThing` || service.call
  ...
```
Which doesn't look like the best thing possible...

The odd thing for me is that `Finance::StripePaymentIntentUpdateService` can raise an `ActiveRecord::RecordNotFound` when initialised, instead of a custom error, for example.

thalesmiguel: Aside from this controller, `Finance::StripePaymentIntentUpdateService` is only called [here](https://github.com/3scale/porta/blob/master/lib/developer_portal/app/controllers/developer_portal/admin/account/invoices_controller.rb#L40). Do we need to perform the rescue there as well?
Maybe we should change `Finance::StripePaymentIntentUpdateService` so it raises an error when `call` is executed or even set an `errors` attribute to it, if we get an `ActiveRecord::RecordNotFound` when it's initialised, which would make life easier to work with its exceptions, in the sense it'd be handled by the service itself.

test/integration/finance/api/payment_callbacks/stripe_callbacks_controller_test.rb
thalesmiguel: I agree. Also, [this test](https://github.com/3scale/porta/blob/master/test/integration/finance/api/payment_callbacks/stripe_callbacks_controller_test.rb#L55-L61) is checking already what happens when an invalid event is received already.

app/controllers/finance/api/payment_callbacks/stripe_callbacks_controller.rb
jlledom: I made some changes based on your comment.

About `Finance::StripePaymentIntentUpdateService` begin called from [here](https://github.com/3scale/porta/blob/master/lib/developer_portal/app/controllers/developer_portal/admin/account/invoices_controller.rb#L40), we don't need to add the rescue there. That rescue only makes sense in the webhook b/c we want to return a `20X` status code to Stripe for it to not disable the webhook. But out of the webhook we don't want to mask the exception. Anyway, the `RecordNotFound` exception comes from [/app/services/finance/stripe_payment_intent_update_service.rb#L7](https://github.com/3scale/porta/blob/3017562449d3275840cae6afe338a01c1dcca899/app/services/finance/stripe_payment_intent_update_service.rb#L7), as a result of sending it an non existing `payment_intent`, that is, trying to pay an invoice not in our DB. But when called from [/app/controllers/developer_portal/admin/account/invoices_controller.rb#L40](https://github.com/3scale/porta/blob/master/lib/developer_portal/app/controllers/developer_portal/admin/account/invoices_controller.rb#L40), the line above already takes the proper invoice and fails if it doesn't exist, so it will always instantiate the service with a valid `payment_intent`.

thalesmiguel (approved):
🚀

app/controllers/finance/api/payment_callbacks/stripe_callbacks_controller.rb
mayorova: @jlledom I think for now we can leave it as it is. This controller is not known for causing troubles (except for the issue that is fixed in this PR), so probably we don't need to complicate the things, and we can leave it as it is.

mayorova (dismissed):

jlledom:
I added a comment to better explain the situation

app/controllers/finance/api/payment_callbacks/stripe_callbacks_controller.rb
```diff
@@ -24,7 +24,15 @@ def create
     service = begin
       Finance::StripePaymentIntentUpdateService.new(current_account, stripe_event)
     rescue ActiveRecord::RecordNotFound
-      # Returning 204 to acknowledge reception even for payments we don't care about
+      # Returning 204 to acknowledge reception even for payments we don't care about.
+      #
+      # There are some clients who are using their Stripe account not only for 3Scale,
```
mayorova, line 29: 3scale is with lowecase `s` !!!

app/controllers/finance/api/payment_callbacks/stripe_callbacks_controller.rb
```diff
@@ -21,9 +21,15 @@ class MissingStripeEndpointSecret < StripeCallbackError; end
...
   # Undocumented endpoint used for update callbacks of async-authorized payment transactions (mostly due to SCA regulations)
   def create
-    service = Finance::StripePaymentIntentUpdateService.new(current_account, stripe_event)
+    service = begin
+      Finance::StripePaymentIntentUpdateService.new(current_account, stripe_event)
+    rescue ActiveRecord::RecordNotFound
+      # Returning 204 to acknowledge reception even for payments we don't care about
+      # https://issues.redhat.com/browse/THREESCALE-6851
+      nil
+    end
```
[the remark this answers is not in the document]
jlledom: yes, but codeclimate complains about calling return `head(:no_content)` twice

app/controllers/finance/api/payment_callbacks/stripe_callbacks_controller.rb
jlledom: Fixed

app/controllers/finance/api/payment_callbacks/stripe_callbacks_controller.rb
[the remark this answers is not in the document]
akostadinov: We can approve that manually in case it gets in the way. If we agree that something is more readable or better in another way, we should do it instead of staying hostages to a tool.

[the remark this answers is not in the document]
jlledom: Since there's not a consensus I decided to apply this changes

mayorova (approved):

PR #3091, opened 2022-10-28
📝 Update setup document

josemigallas:
I recently manage to run the project with node 12 and python2 on a M1 mac, so I'm updating the docs according to that.

On the other hand, I think it's a good idea to extract specific setup guides into their own files because it's kinda confusing to have all together in the same document.

In general the guides are untouched except for small improvements here and there.


josemigallas:
@thalesmiguel since you're the only one using an M1 mac besides me, I would appreciate you read through the installation doc and try using `asdf` with the provided `.tool-versions`.

fedora_manual_setup.md
```diff
@@ -0,0 +1,114 @@
+# Manual setup on Fedora (34)
```
jlledom, line 1:
```suggestion
# Manual setup on Fedora (36)
```
I got Fedora 36 and this manual is valid

fedora_manual_setup.md
```diff
@@ -0,0 +1,114 @@
...
+> ```
+> dnf module install nodejs:12
+> ```
+
+## Dependencies
+
+```
+sudo yum install patch autoconf automake bison libffi-devel libtool libyaml-devel readline-devel sqlite-devel zlib-devel openssl-devel ImageMagick ImageMagick-devel mysql-devel postgresql-devel chromedriver memcached
+
+sudo systemctl restart memcached
```
jlledom, line 34: Not sure about installing memcached, b/c I use it in a container, and the APP will work even without it anyway

fedora_manual_setup.md
```diff
@@ -0,0 +1,114 @@
...
+```
+
+Alternatively, you can run Redis directly on your machine by using `yum`:
+
+```
+sudo yum install redis
+
+sudo systemctl restart redis
+```
+
```
jlledom, line 71:
```suggestion

## Memcached

TODO: A brief explanation here...

docker run -d -p 11211:11211 memcached
```

Better recommend using docker? like any other dependency, If we don't recommend to install the database, which is required, and suggest to use docker instead, why not doing the same with memcached, which is optional?

fedora_manual_setup.md
```diff
@@ -0,0 +1,114 @@
...
+## Eventmachine
+
+Eventmachine has to be installed with `--with-cppflags=-I/usr/include/openssl/`. Simply run:
```
jlledom, line 74: What's this for? I'm not sure I installed this.

fedora_manual_setup.md
```diff
@@ -0,0 +1,114 @@
...
+
+```
+sudo yum install patch autoconf automake bison libffi-devel libtool libyaml-devel readline-devel sqlite-devel zlib-devel openssl-devel ImageMagick ImageMagick-devel mysql-devel postgresql-devel chromedriver memcached
+
+sudo systemctl restart memcached
+```
+
+## Database (Postgres / MySQL / Oracle)
+
+Postgres, MySQL or Oracle has to be running for the application to work. The easiest way to do it is in a [Docker](https://www.docker.com/) container by simply running:
```
jlledom, line 39:
```suggestion
Postgres, MySQL or Oracle has to be running for the application to work. The easiest way to do it is in a [Podman](https://podman.io/) container by simply running:
```
In all this document, I think we should replace "Docker" by "Podman" or "Container". To follow RH guidelines.

fedora_manual_setup.md
```diff
@@ -0,0 +1,114 @@
...
+```
+
+## Yarn (1.x)
+
+To manage our JavaScript packages we use [Yarn](https://yarnpkg.com/). It is recommended to install it with NPM:
+```
+npm install --global yarn
+```
+
+> See other ways to install it, such as with `brew`, at https://classic.yarnpkg.com/en/docs/install.
```
jlledom, line 100: Why linking to the installer for versions lower than 2.0?

INSTALL.md
```diff
@@ -8,319 +8,16 @@ Follow these instructions to set up a development environment, build and deploy
...
-> you can fix it setting the flags:
->
-> ```bash
-> bundle config --local build.mysql2 --with-ldflags="-L$(brew --prefix openssl)/lib" --with-cppflags="-I$(brew --prefix openssl)/include"
-> ```
->
-> and run `bundle install` again.
-
-#### Yarn
+[Manual setup on MacOS](./osx-manual-setup.md)
```
thalesmiguel, line 18: This path doesn't correspond with the file created for the OSX docs.

INSTALL.md
```diff
@@ -8,319 +8,16 @@ Follow these instructions to set up a development environment, build and deploy
...
-npm install --global yarn
-```
-
-
-Run [Yarn](https://www.yarnpkg.com/) to install all the required dependencies:
-
-```
-yarn install
-```
+[Manual setup on Fedora](./fedora-manual-setup.md)
```
thalesmiguel, line 20: This path doesn't correspond with the file created for the Fedora docs.

.tool-versions.sample
```diff
@@ -1,3 +1,3 @@
 ruby 2.6.10
 nodejs 12.22.12
-python 3.10.7 2.7.18 # Meant for M1 macs
+python 2.7.18
```
akostadinov, line 3: what requires python 2?

fedora_manual_setup.md
josemigallas: I don't use it either. Actually, there is no step or instruction to start it up as part of macOS setup.

fedora_manual_setup.md
josemigallas: I for instance try to avoid using Docker due to memory consumption but I see your point. I will add this new section and maybe mention the 2 options. I need to do the same for the macOS document.

fedora_manual_setup.md
josemigallas: This probably fixed some issue in the past but nobody knew why. We can maybe trace the commit back and see what happened.

fedora_manual_setup.md
josemigallas: Because we still use 1.x

jlledom (approved):

.tool-versions.sample
josemigallas: Something failed at some point with python 3, now I don't know what it was anymore. There are too many pieces in this crazy puzzle.

oxs_manual_setup.md
```diff
@@ -0,0 +1,152 @@
...
+ln -s .tool-versions.sample .tool-versions
+```
+
+### Ruby and Node.js
+
+The project supports **[ruby 2.6.x](https://www.ruby-lang.org/en/downloads/)** and **[Node.js 12](https://nodejs.org/en/download/)**.
+The recommended way to install them is with `asdf`:
+
+```
+asdf plugin add ruby
```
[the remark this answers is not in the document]
josemigallas: I was giving me the headaches until one day it worked and I've been happily using `2.6.8` since. I'm going to uninstall it and start over 🤞🏻.

thalesmiguel (approved):

fedora_manual_setup.md
josemigallas: I'm trying to replace all instructions with Podman but it's not as simplye as I thought. I will leave this change to someone who actually knows how Podman works, if that's OK.

jlledom: I'll do it myself today. Don't merge this

jlledom (approved):

PR #1427, opened 2019-11-18
Updated with dark nav 

dlabaj:
Updated vertical nav to be dark nav.

Screen Shot 2019-11-18 at 3 40 40 PM

PR #687, opened 2019-03-06
Feature: Custom Policies

didierofrivia:
**What this PR does / why we need it**:

New feature: Custom Policies.

**Which issue(s) this PR fixes** 

TODO: add Jira issue

**Verification steps** 
* Visit Account -> Integration -> Custom Policies
* Profit

**Special notes for your reviewer**:
Make it ugly, make it work first. There are some things to polish, specially regarding the way we get the policies from different sources, we should normalize that. Regarding more JS stuff, it's not super heavily tested, but the most important features and behaviour is there. There's also some use of weak flow types (like `any`) for objects like `Window` which doesn't have a specific flow type yet.

PR #3092, opened 2022-10-28
➖ Remove `canvas`

josemigallas:

From time to time, `canvas.node` throws an error that mess Circleci or yarn install or whatever. Example:
```
FAIL  spec/javascripts/utilities/json-utils.spec.ts
  ● Test suite failed to run

    /lib64/libstdc++.so.6: version `CXXABI_1.3.9' not found (required by /opt/app-root/src/project/node_modules/canvas/build/Release/canvas.node)

      at Object.<anonymous> (node_modules/canvas/lib/bindings.js:3:18)

 FAIL  spec/javascripts/Settings/components/AuthenticationSettingsFieldset.spec.tsx 
```
TL;DR: Well it turns out we don't even use this dependency 🤷🏻‍♂️ .

Some history though:

When Policies were added (#687), instead of using the existing testing library for React `enzyme`, a new package `jsdom` was added to be used literally twice, one to test `Editor` and another for `CSRFToken` (https://github.com/3scale/porta/pull/687/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519).

Some months later, for reasons unknown really, `canvas` was added as a direct dependency in #1436. It was supposed to be required for #1427 but for all we know that PR was never merged into `master` 🙃.

After years (yes, years) of unnecessary headaches and bad developer experience due to this `canvas` dependency I tried to fix it by removing the also-unnecessary `jsdom` from the tests in #2426. Little did I know `canvas` was only added in the first place by the it.

And finally we end up here where `canvas` is still wreaking havoc: https://app.circleci.com/pipelines/github/3scale/porta/21952/workflows/a75222f7-92a8-4f5a-8ecb-15a97f53c7be/jobs/252157

### DIFF
--- a/doc/dependency_decisions.yml
+++ b/doc/dependency_decisions.yml
@@ -630,4 +630,9 @@
 - - :approve
   - WHY=0BSD
   - *16
-
+- - :remove_package
+  - canvas
+  - :who: 
+    :why: 
+    :versions: []
+    :when: 2022-10-28 11:26:31.331728000 Z
--- a/package.json
+++ b/package.json
@@ -28,7 +28,6 @@
     "@babel/preset-react": "^7.13.13",
     "babel-eslint": "^10.1.0",
     "babel-jest": "^24.9.0",
-    "canvas": "^2.5.0",
     "css-loader": "^2.1.1",
     "empty": "^0.10.1",
     "enzyme": "^3.3.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1289,21 +1289,6 @@
   dependencies:
     stream "^0.0.2"
 
-"@mapbox/node-pre-gyp@^1.0.0":
-  version "1.0.5"
-  resolved "https://registry.yarnpkg.com/@mapbox/node-pre-gyp/-/node-pre-gyp-1.0.5.tgz#2a0b32fcb416fb3f2250fd24cb2a81421a4f5950"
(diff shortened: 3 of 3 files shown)


jlledom (dismissed):

thalesmiguel (approved):
🚀

doc/dependency_decisions.yml
```diff
@@ -630,4 +630,9 @@
 - - :approve
   - WHY=0BSD
   - *16
-
+- - :remove_package
+  - canvas
+  - :who: 
+    :why: 
+    :versions: []
+    :when: 2022-10-28 11:26:31.331728000 Z
```
akostadinov, line 638: why do we need a decision to remove package canvas?

josemigallas: If I understood correctly, this file keeps track of what packages are added and removed. Since we don't use canvas anymore I thought it was best to remove it. There is an entry approving it:
https://github.com/josemigallas/porta/blob/61d162aeb78f7768d6f1f783e3610404149be533/doc/dependency_decisions.yml#L581-L587

akostadinov: I think we need to add/remove there individual packages only when their licenses are not approved by default. Strange that some packages were individually approved. Honestly I'd rather remove them from file than making it longer. We have git history anyway.

josemigallas: Unfortunately there is not much information about it https://github.com/pivotal/LicenseFinder#decisions-file. It's supposed to be handled by `license_finder` alone 🤷🏻‍♂️ but I don't really understand the importance of it

PR #3093, opened 2022-10-28
📄 Update `license_finder` and licenses file

josemigallas:
I still don't understand why licenses don't fail, but here we go.

Licenses tasks have been renamed to:
```
rails --tasks|grep license

rails licenses:compliance
rails licenses:report
```


lib/tasks/ci/licenses.rake
```diff
@@ -1,27 +1,26 @@
...
+  desc 'Run compliance task and generates the license report if complies'
+  task run: :environment do
+    Rake::Task['license_finder:report'].invoke if Rake::Task['license_finder:compliance'].invoke
+  end
+
+  desc 'Check license compliance of dependencies'
+  task compliance: :environment do
+    $stdout.puts 'Checking license compliance...'
+    unless system('license_finder')
+      warn '*** License compliance test failed  ***'
```
akostadinov, line 14: Shouldn't this be an `error`?

akostadinov:
Why remove `licenses.xml`? Honest question, if this report doesn't serve any purpose, that could be actually good. We better understand why it was put in the repo in the first place though.

lib/tasks/ci/licenses.rake
```diff
@@ -1,27 +1,26 @@
...
-        exit 1
-      end
-    end
-    desc 'Generates a report with the dependencies and their licenses'
-    task :report do
-      STDOUT.puts 'Generating report...'
-      licenses_xml_path = Rails.root.join('doc', 'licenses', 'licenses.xml')
-      exec("license_finder report --format=xml --save #{licenses_xml_path}")
+  desc 'Run compliance task and generates the license report if complies'
+  task run: :environment do
```
akostadinov, line 6:
```suggestion
  task :run do
```
I don't think we need to depend on the `environment` task for licene_finder`.

lib/tasks/ci/licenses.rake
```diff
@@ -1,27 +1,26 @@
...
+    Rake::Task['license_finder:report'].invoke if Rake::Task['license_finder:compliance'].invoke
```
akostadinov, line 7: actually if you want to achieve this, you can do
```
  task run: %i(compliance report) do
  end
```

.circleci/config.yml
```diff
@@ -491,7 +491,7 @@ jobs:
       - run:
           name: Licences check
           command: |
-            bundle exec rake ci:license_finder:run
+            bundle exec rake license_finder:compliance
```
jlledom, line 494: Calling `compliance` instead of `run` skips the call to `report` when `compliance` fails, is that intended?

lib/tasks/ci/licenses.rake
josemigallas: Rubocop 🤷🏻‍♂️

akostadinov: what do you mean?

lib/tasks/ci/licenses.rake
```diff
@@ -1,27 +1,25 @@
...
-      exec("license_finder report --format=xml --save #{licenses_xml_path}")
+  desc 'Run compliance task and generates the license report if complies'
+  task run: %i[compliance report] do
+  end
+
+  desc 'Check license compliance of dependencies'
+  task :compliance do
+    $stdout.puts 'Checking license compliance...'
+    unless system('license_finder')
+      $stdout.puts '*** License compliance test failed  ***'
```
jlledom, line 13: better print to `stderr` IMO.

lib/tasks/ci/licenses.rake
josemigallas: `error` doesn't seem to exist and `warn` is identical to `$stdout.puts` apparently, so I'm gonna stick to the original.

.circleci/config.yml
josemigallas: We will keep the report in the repo. What I don't see is why circleci should generate it again? The job is to _check_ the licenses, not generating a new report. Am I missing something else?

josemigallas:
> Why remove `licenses.xml`? Honest question, if this report doesn't serve any purpose, that could be actually good. We better understand why it was put in the repo in the first place though.

It hasn't been removed but massively cleaned up, so it seems. That's the result of me running `license_finder:run` on my machine.

jlledom (approved):

.gitignore
```diff
@@ -27,6 +27,7 @@ C:\\nppdf32Log\\debuglog.txt
 .deploy
 .s3cfg
 .tool-versions
+doc/licenses/temp_licenses.xml
```
akostadinov, line 30: maybe use `tmp/something` for this? Or maybe even better Ruby's `Tempfile` for the temporary report.

lib/tasks/licenses.rake
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+namespace :licenses do
+
+  report_path = Rails.root.join('doc/licenses/licenses.xml')
+  temp_report_path = Rails.root.join('doc/licenses/temp_licenses.xml')
```
akostadinov, line 6: here would be a good place to use `Tempfile`

lib/tasks/licenses.rake
```diff
@@ -0,0 +1,31 @@
...
+    $stdout.puts 'Checking license report is up to date...'
+
+    exec "license_finder report --format=xml --save #{temp_report_path}"
+
+    unless system("diff #{report_path} #{temp_ci_report_path} --brief")
+      $stderr.puts '*** License report outdated. Please generate again. ***' # rubocop:disable Style/StderrPuts
+      exit 1
+    end
+
+    $stdout.puts 'Checking license compliance...'
```
akostadinov, line 25: `$stderr` everywhere.

lib/tasks/licenses.rake
```diff
@@ -0,0 +1,31 @@
...
+    exec("license_finder report --format=xml --save=#{report_path}")
+  end
+
+  desc 'Check license compliance of dependencies'
+  task :compliance do
+    $stdout.puts 'Checking license report is up to date...'
+
+    exec "license_finder report --format=xml --save #{temp_report_path}"
+
+    unless system("diff #{report_path} #{temp_ci_report_path} --brief")
```
akostadinov, line 20: instead of using an external programs, better use https://ruby-doc.org/stdlib-1.9.3/libdoc/fileutils/rdoc/FileUtils.html#method-c-identical-3F

lib/tasks/licenses.rake
```diff
@@ -0,0 +1,31 @@
...
+      $stderr.puts '*** License report outdated. Please generate again. ***' # rubocop:disable Style/StderrPuts
+      exit 1
+    end
+
+    $stdout.puts 'Checking license compliance...'
+    unless system('license_finder action_items')
+      $stderr.puts '*** License compliance test failed  ***' # rubocop:disable Style/StderrPuts
+      exit 1
+    end
```
akostadinov, line 29: I would put compliance check before report comparison.

config/license_finder.yml
```diff
@@ -0,0 +1,7 @@
+---
+project_name: Porta
+dependencies_file_dir: './doc/license'
+decisions_file: './doc/licenses/dependency_decisions.yml'
```
akostadinov, line 4: is it intentional to have `./doc/license` vs `./doc/licenses`?

lib/tasks/licenses.rake
josemigallas: WDYM? `stderr` is to put when the task fails, or is it?

config/license_finder.yml
josemigallas: No, it's a typo

lib/tasks/licenses.rake
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+namespace :licenses do
+
+  report_path = Rails.root.join('doc/licenses/licenses.xml')
+
+  desc 'Generates a report with the dependencies and their licenses'
+  task :report do
+    $stdout.puts 'Generating report...'
+    exec license_finder_report
```
akostadinov, line 10: btw why use `exec` instead of `system` here?
```suggestion
    system license_finder_report, exception: true
```

While as a last task, `exec` makes sense because we don't run anything after, if we have multiple rake tasks on the command line, this will terminate execution of any further list tasks.

Maybe unlikely but I think it is proper way to do.

P.S. `exception: true` is available from ruby 2.6 but we all use 2.6+ and we don't run this task on legacy.

akostadinov (changes requested):
Thanks, just some details.

lib/tasks/licenses.rake
```diff
@@ -0,0 +1,36 @@
...
+
+  desc 'Check license compliance of dependencies'
+  task :compliance do
+    $stdout.puts 'Checking license compliance...'
+    unless system('license_finder action_items')
+      $stderr.puts '*** License compliance test failed  ***' # rubocop:disable Style/StderrPuts
+      exit 1
+    end
+
+    $stdout.puts 'Checking license report is up to date...'
```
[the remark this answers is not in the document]
josemigallas: >  I see that rubocop recommends to use warn "message" instead of $stderr.puts. So maybe use that instead?
😆 
https://github.com/3scale/porta/pull/3093#discussion_r1008057770

lib/tasks/licenses.rake
josemigallas: Short question: it was like this before 😄 I'll change it.

lib/tasks/licenses.rake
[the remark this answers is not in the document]
akostadinov: I didn't know it will complain :man_facepalming:

akostadinov (dismissed):
Thanks a lot, this was super useful!

lib/tasks/licenses.rake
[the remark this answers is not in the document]
jlledom: But don't we want to present the string 'Checking license report is up to date...' to the user? otherwise the task would have no standard output. I agree not everything in `stderr` must be an error, but in this case, if I run the task and redirect the `ERR` output to a file or `/dev/null`, I won't see anything.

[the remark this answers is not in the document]
josemigallas: I agree with both but since these messages are meant for us developer and they really don't give much value... I'd leave it in `stderr` if not remove them altogether.

jlledom (approved):

jlledom (dismissed):

config/license_finder.yml
```diff
@@ -0,0 +1,6 @@
+---
+project_name: Porta
```
akostadinov, line 2: Where did you see this? In README I only see `license_finder project_name add Porta` which alters `dependency_decisions.yml`, not the config file

config/license_finder.yml
```diff
@@ -0,0 +1,6 @@
...
+dependencies_file_dir: './doc/licenses'
```
akostadinov, line 3: btw setting also the dependency decisions file worked for me
```
🐚 cat config/license_finder.yml 
---
project_name: Porta
dependencies_file_dir: './doc/licenses'
decisions_file: './doc/licenses/decisions.yml'
enabled_package_managers:
  - bundler
  - yarn
```
When I ran some commands, they were applied to the specified decisions file. If you want to have its path changed, the above worked.

config/license_finder.yml
josemigallas: It is used as the page title when generating a report in HTML format.

config/license_finder.yml
josemigallas: Yes, problem is with Github in this case, it complains about unsolvable conflicts in `dependency_decisions.yaml` even though it's just a file name change. Moreover the default location is `doc/dependency_decisions.yml` so it's better not to change it after all.

akostadinov: IMO, it's a strange location. It's more for `config` something but I'm fine with keeping it where it is.

lib/tasks/licenses.rake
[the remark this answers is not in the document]
akostadinov: @jlledom , stderr is still presented to user. The point is that in shell, the data that one would pipe into simething else goes to stdout. And any other information messages should go to stderr. I don't know of a formal standard but this is what I have observed. So stderr is still presented to user.

config/license_finder.yml
akostadinov: It doesn't work for me locally. Whatever I put there, project name is always `porta`.

lib/tasks/licenses.rake
```diff
@@ -0,0 +1,36 @@
...
+    unless identical?(report_path, temp_report_path)
+      warn '*** License report outdated. Please run "rails licenses:report" ***'
+      exit 1
+    end
+  end
+
+  private
+
+  def license_finder_report(path)
+    "license_finder report --format=xml --save=#{path}"
```
akostadinov, line 34: btw we don't have to run an external process to generate the report. We can do

```ruby
require 'license_finder'
lf = LicenseFinder::CLI::Main.new
lf.options = { "format" => "html", "save" => "/tmp/api.html" }
lf.report
```

I think it is preferable to avoid calling external programs. But if you strongly prefer keep calling it like this, I will not insist :innocent:

akostadinov (dismissed):
Thanks a lot!

akostadinov (approved):

PR #3094, opened 2022-10-31
THREESCALE-8443 prevent Mesages being sent without a subject or body in reply to message form under Messages > Inbox > Message

nidhi-soni1104:

prevent Messages being sent without a subject or body in reply to message form under Messages > Inbox > Message

https://issues.redhat.com/browse/THREESCALE-8443

**Verification steps**
Emails can be sent without a subject or body in reply to message form under Messages > Inbox > Message

e.g. path /p/admin/messages/inbox/{message_id}
try sending message without subject.

### DIFF
--- a/app/controllers/provider/admin/messages/inbox_controller.rb
+++ b/app/controllers/provider/admin/messages/inbox_controller.rb
@@ -30,6 +30,19 @@ class Provider::Admin::Messages::InboxController < FrontendController
     end
   end
 
+  def reply
+    reply = @message.reply
+    reply.attributes = message_params
+
+    if reply.save && reply.deliver
+      flash[:notice] = 'Reply was sent.'
+      redirect_to action: :index
+    else
+      flash[:error] = reply.errors.full_messages.to_sentence
+      redirect_to provider_admin_messages_inbox_path(@message)
+    end
+  end
+
   private
 
   def message_params
@@ -38,13 +51,5 @@ class Provider::Admin::Messages::InboxController < FrontendController
 
   def find_message
     @message = current_account.received_messages.find(params.require(:id)).decorate
-  end
-
-  def send_reply(reply)
-    reply.save
-    reply.deliver
-
-    flash[:notice] = 'Reply was sent.'
-    redirect_to action: :index
   end
 end
--- a/app/controllers/provider/admin/messages/inbox_controller.rb
+++ b/app/controllers/provider/admin/messages/inbox_controller.rb
@@ -22,11 +22,16 @@ class Provider::Admin::Messages::InboxController < FrontendController
     reply = @message.reply
     reply.attributes = message_params
 
-    reply.save!
-    reply.deliver!
+    if reply.valid?
+      reply.save!
+      reply.deliver!
 
-    flash[:notice] = 'Reply was sent.'
-    redirect_to action: :index
+      flash[:notice] = 'Reply was sent.'
+      redirect_to action: :index
+    else
+      flash[:error] = reply.errors.full_messages.to_sentence
+      redirect_to provider_admin_messages_inbox_path(@message)
+    end
   end
 
   private
--- a/app/controllers/provider/admin/messages/inbox_controller.rb
+++ b/app/controllers/provider/admin/messages/inbox_controller.rb
@@ -23,11 +23,7 @@ class Provider::Admin::Messages::InboxController < FrontendController
     reply.attributes = message_params
 
     if reply.valid?
-      reply.save!
-      reply.deliver!
-
-      flash[:notice] = 'Reply was sent.'
-      redirect_to action: :index
+      send_reply(reply)
     else
       flash[:error] = reply.errors.full_messages.to_sentence
       redirect_to provider_admin_messages_inbox_path(@message)
@@ -42,5 +38,13 @@ class Provider::Admin::Messages::InboxController < FrontendController
 
   def find_message
     @message = current_account.received_messages.find(params.require(:id)).decorate
+  end
+
+  def send_reply(reply)
+    reply.save!
+    reply.deliver!
+
+    flash[:notice] = 'Reply was sent.'
+    redirect_to action: :index
   end
 end
--- a/app/controllers/provider/admin/messages/inbox_controller.rb
+++ b/app/controllers/provider/admin/messages/inbox_controller.rb
@@ -41,8 +41,8 @@ class Provider::Admin::Messages::InboxController < FrontendController
   end
 
   def send_reply(reply)
-    reply.save!
-    reply.deliver!
+    reply.save
+    reply.deliver
 
     flash[:notice] = 'Reply was sent.'
     redirect_to action: :index
--- a/app/controllers/provider/admin/messages/inbox_controller.rb
+++ b/app/controllers/provider/admin/messages/inbox_controller.rb
@@ -22,18 +22,6 @@ class Provider::Admin::Messages::InboxController < FrontendController
     reply = @message.reply
     reply.attributes = message_params
 
-    if reply.valid?
-      send_reply(reply)
-    else
-      flash[:error] = reply.errors.full_messages.to_sentence
-      redirect_to provider_admin_messages_inbox_path(@message)
-    end
-  end
-
-  def reply
-    reply = @message.reply
-    reply.attributes = message_params
-
     if reply.save && reply.deliver
       flash[:notice] = 'Reply was sent.'
       redirect_to action: :index
--- a/test/functional/provider/admin/messages/inbox_controller_test.rb
+++ b/test/functional/provider/admin/messages/inbox_controller_test.rb
@@ -55,6 +55,7 @@ class Provider::Admin::Messages::InboxControllerTest < ActionController::TestCas
 
     MessageWorker.drain
     msg = Message.last
+    
     assert_equal flash[:notice], 'Reply was sent.'
     assert 'sent', msg.state
     assert_equal 'message with subject', msg.body
@@ -65,7 +66,7 @@ class Provider::Admin::Messages::InboxControllerTest < ActionController::TestCas
     post :reply, params: { message: { subject: nil, body: 'message with nil subject' }, id: @message.id }
 
     MessageWorker.drain
-    msg = Message.find_by(id: @message.message_id)
+    msg = Message.last
     assert_not_equal 'message with nil subject', msg.body
   end
 end
--- a/test/functional/provider/admin/messages/inbox_controller_test.rb
+++ b/test/functional/provider/admin/messages/inbox_controller_test.rb
@@ -49,23 +49,23 @@ class Provider::Admin::Messages::InboxControllerTest < ActionController::TestCas
     assert_select '#export-to-csv', false, 'Export all Messages'
   end
 
-  test 'Prevent sending reply without subject' do
+  test "creates valid reply" do
     login_as(@admin)
-    reply = @message.reply
-    reply.update_attributes(body: 'Reply Body', subject: nil)
+    post :reply, params: { message: { subject: "Valid Message", :body => "message with subject" }, id: @message.id }
 
     MessageWorker.drain
-    assert msg = Message.last
-    assert_not_equal "Reply Body", msg.body
+    msg = Message.last
+    assert "sent" msg.state
+    assert_equal "message with subject", msg.body
   end
 
-  test 'Allow sending reply with subject' do
+  test "not creates invalid message" do
     login_as(@admin)
-    reply = @message.reply
-    reply.update_attributes(body: 'Reply Body', subject: 'Reply Subject')
+    post :reply, params: { message: { subject: nil, body: "message with nil subject" }, id: @message.id }
 
     MessageWorker.drain
-    assert msg = Message.last
-    assert_equal "Reply Subject", msg.subject
+    msg = Message.last
+    assert_not_equal "message with nil subject", msg.body
   end
+
 end
--- a/test/functional/provider/admin/messages/inbox_controller_test.rb
+++ b/test/functional/provider/admin/messages/inbox_controller_test.rb
@@ -48,4 +48,24 @@ class Provider::Admin::Messages::InboxControllerTest < ActionController::TestCas
     assert_select 'title', "Inbox - Index | Red Hat 3scale API Management"
     assert_select '#export-to-csv', false, 'Export all Messages'
   end
+
+  test 'Prevent sending reply without subject' do
+    login_as(@admin)
+    reply = @message.reply
+    reply.update_attributes(body: 'Reply Body', subject: nil)
+
+    MessageWorker.drain
+    assert msg = Message.last
+    assert_not_equal "Reply Body", msg.body
+  end
+
+  test 'Allow sending reply with subject' do
+    login_as(@admin)
+    reply = @message.reply
+    reply.update_attributes(body: 'Reply Body', subject: 'Reply Subject')
+
+    MessageWorker.drain
+    assert msg = Message.last
+    assert_equal "Reply Subject", msg.subject
+  end
 end
--- a/test/functional/provider/admin/messages/inbox_controller_test.rb
+++ b/test/functional/provider/admin/messages/inbox_controller_test.rb
@@ -51,7 +51,7 @@ class Provider::Admin::Messages::InboxControllerTest < ActionController::TestCas
 
   test "creates valid reply" do
     login_as(@admin)
-    post :reply, params: { message: { subject: "Valid Message", :body => "message with subject" }, id: @message.id }
+    post :reply, params: { message: { subject: "Valid Message", body: "message with subject" }, id: @message.id }
 
     MessageWorker.drain
     msg = Message.last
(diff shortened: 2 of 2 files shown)


app/controllers/provider/admin/messages/inbox_controller.rb
```diff
@@ -22,11 +22,16 @@ def reply
     reply = @message.reply
     reply.attributes = message_params
 
-    reply.save!
-    reply.deliver!
-
-    flash[:notice] = 'Reply was sent.'
-    redirect_to action: :index
+    if reply.valid?
```
jlledom, line 25: This prevents sending messages without a subject, but by the title I understood that messages without a body should fail as well. Am I right?

test/functional/provider/admin/messages/inbox_controller_test.rb
```diff
@@ -48,4 +48,24 @@ def test_index_system_message
...
     assert_select '#export-to-csv', false, 'Export all Messages'
   end
+
+  test 'Prevent sending reply without subject' do
+    login_as(@admin)
+    reply = @message.reply
+    reply.update_attributes(body: 'Reply Body', subject: nil)
+
+    MessageWorker.drain
+    assert msg = Message.last
```
jlledom, line 58: why this assert?

test/functional/provider/admin/messages/inbox_controller_test.rb
```diff
@@ -48,4 +48,24 @@ def test_index_system_message
...
+    assert_not_equal "Reply Body", msg.body
```
jlledom, line 59: Shouldn't it better to assert `message.valid?`

test/functional/provider/admin/messages/inbox_controller_test.rb
```diff
@@ -48,4 +48,24 @@ def test_index_system_message
...
+  end
+
+  test 'Allow sending reply with subject' do
+    login_as(@admin)
+    reply = @message.reply
+    reply.update_attributes(body: 'Reply Body', subject: 'Reply Subject')
+
+    MessageWorker.drain
+    assert msg = Message.last
+    assert_equal "Reply Subject", msg.subject
```
jlledom, line 69: Same here, is it not more accurate to assert `message.valid?`

test/functional/provider/admin/messages/inbox_controller_test.rb
```diff
@@ -48,4 +48,24 @@ def test_index_system_message
     assert_select 'title', "Inbox - Index | Red Hat 3scale API Management"
     assert_select '#export-to-csv', false, 'Export all Messages'
   end
+
+  test 'Prevent sending reply without subject' do
```
thalesmiguel, line 52: This is the `Provider::Admin::Messages::InboxControllerTest` but we're not calling any of its actions on those 2 new tests. For those two, we should call `post :reply` and assert that things work as expected.

test/functional/provider/admin/messages/inbox_controller_test.rb
```diff
@@ -48,4 +48,24 @@ def test_index_system_message
...
+    login_as(@admin)
+    reply = @message.reply
+    reply.update_attributes(body: 'Reply Body', subject: nil)
```
thalesmiguel, line 55: `@message.reply` [creates a new Message instance](https://github.com/3scale/porta/blob/master/app/models/message_recipient.rb#L75-L79) with subject and body based on the `@message`. By using `update_attributes` on it, if there are no errors, the message will be stored on the DB, which doesn't look like something we want to do, as the code that was updated on `Provider::Admin::Messages::InboxController` calls `Message#reply` as well, creating a new Message instance as well.

app/controllers/provider/admin/messages/inbox_controller.rb
```diff
@@ -22,11 +22,16 @@ def reply
...
+      reply.save!
+      reply.deliver!
```
thalesmiguel, line 27: We're calling `save!` and `deliver!` with a bang ("!") that would raise errors in case of failure and I couldn't find any rescues for those cases.
Maybe it goes like:
- We'll not see a `ActiveRecord::RecordInvalid` here because, for a reply to be created, the message should be stored on the DB. Said message should be valid (unless we have invalid messages there, of course);
- We'll not see a `StateMachines::InvalidTransition` because the just created reply will have the `unsent` initial state;

That to say that those bangs could be removed and we could do something like

```ruby
if reply.save && reply.delive
  ...
else
  ...
end

Comment (thalesmiguel) [reply app/controllers/provider/admin/messages/inbox_controller.rb:27]: Notice that this could only be done if we add the body validation to the Message model.

File: test/functional/provider/admin/messages/inbox_controller_test.rb (hunk 5 above)
Comment (thalesmiguel) [reply test/functional/provider/admin/messages/inbox_controller_test.rb:52]: We should also have a test in case there's no body.

File: app/controllers/provider/admin/messages/inbox_controller.rb (hunk 1 above)
Comment (nidhi-soni1104) [reply app/controllers/provider/admin/messages/inbox_controller.rb:25]: This is subtask, and we have already added validation on message subject which check if message is valid or not . same goes for reply also, as reply is also a message and as per discussion with team its only required to make subject as mandatory field and not body

File: test/functional/provider/admin/messages/inbox_controller_test.rb (hunk 2 above)
Comment (nidhi-soni1104) [reply test/functional/provider/admin/messages/inbox_controller_test.rb:58]: Okay may be we can remove assert and just keep msg

File: test/functional/provider/admin/messages/inbox_controller_test.rb (hunk 3 above)
Comment (nidhi-soni1104) [reply test/functional/provider/admin/messages/inbox_controller_test.rb:59]: Sure we can check that but why you think that this is not better , this test also check what we want to check

File: test/functional/provider/admin/messages/inbox_controller_test.rb (hunk 5 above)
Comment (nidhi-soni1104) [reply test/functional/provider/admin/messages/inbox_controller_test.rb:52]: Agreed to first comment can add test which will call post :reply but this test also do similar check as we drain the message and assert on the latest one

For the second comment we dont have any sort of validation on body we allow both nil and with text also so what sense does it make to add a separate test for it

File: app/controllers/provider/admin/messages/inbox_controller.rb (hunk 7 above)
Comment (nidhi-soni1104) [reply app/controllers/provider/admin/messages/inbox_controller.rb]: Yes I think you are correct we already checked if reply is valid or not then only we are saving and delivering it.
i will remove the bangs we don't even need any if else conditions as well.

Comment (nidhi-soni1104) [reply app/controllers/provider/admin/messages/inbox_controller.rb]: we don't  need any kind of validation on body which is already cleared with the team on the JIRA itself

Comment (nidhi-soni1104) [reply app/controllers/provider/admin/messages/inbox_controller.rb]: Done . Removed Bangs

File: test/functional/provider/admin/messages/inbox_controller_test.rb (hunk 4 above)
Comment (nidhi-soni1104) [reply test/functional/provider/admin/messages/inbox_controller_test.rb:69]: Let me check this also and update test

File: test/functional/provider/admin/messages/inbox_controller_test.rb (hunk 5 above)
Comment (thalesmiguel) [reply test/functional/provider/admin/messages/inbox_controller_test.rb:52]: > we dont have any sort of validation on body we allow both nil and with text also so what sense does it make to add a separate test for it

From my understanding of the linked ticket, we should not allow emails to be sent without a subject or without a body.
If that's not the case, we should change the ticket and this PR's title and description.

File: test/functional/provider/admin/messages/inbox_controller_test.rb (hunk 6 above)
Comment (nidhi-soni1104) [reply test/functional/provider/admin/messages/inbox_controller_test.rb:55]: Ummm you are bit wrong here because we have added validation on model for subject can't be nil so here as you said that @message.reply creates a new instance then when update attribute is called on same and if any validation fail it will rollback it because reply is not saved yet  and will not save it.

Here attaching screenshot for you to understand it better

image

File: test/functional/provider/admin/messages/inbox_controller_test.rb (hunk 4 above)
Comment (jlledom) [reply test/functional/provider/admin/messages/inbox_controller_test.rb:69]: You're right, I misunderstood this test. I don't understand how checking the last record in the messages table gives us information about the `reply` message being delivered or not. Couldn't we use the `state` attribute to assert it? That would make the test easier to understand.

File: test/functional/provider/admin/messages/inbox_controller_test.rb (hunk 6 above)
Comment (thalesmiguel) [reply test/functional/provider/admin/messages/inbox_controller_test.rb:55]: Yes. That's what I meant with "... By using update_attributes on it, **if there are no errors**, the message will be stored on the DB..."

The point here is that this file contains the tests for `Provider::Admin::Messages::InboxController` but we're not really testing the controller actions.
For example, on the test `'renders index page with export option for admins'`, we call `get :index` to call the controller action.
As per https://github.com/3scale/porta/pull/3094/files#r1009591764, we should call `post :reply` for the new tests and assert they behave as expected.

Also, if it makes sense, part of the code you added in here could be used as unit tests 🎉

File: test/functional/provider/admin/messages/inbox_controller_test.rb (hunk 5 above)
Comment (nidhi-soni1104) [reply test/functional/provider/admin/messages/inbox_controller_test.rb]: No we had a word with the team and then only we decided only to keep subject mandatory and not body because there are lot of  hardcoded Messages in the code with no body and so they just want subject to be the required filed

here is the link to the JIRA you can read comments for the calrity(https://issues.redhat.com/browse/THREESCALE-8393)

File: test/functional/provider/admin/messages/inbox_controller_test.rb (hunk 6 above)
Comment (nidhi-soni1104) [reply test/functional/provider/admin/messages/inbox_controller_test.rb]: yes I got your point and test are updated

File: test/functional/provider/admin/messages/inbox_controller_test.rb [hunk 8]
```diff
@@ -48,4 +48,24 @@ def test_index_system_message
     assert_select 'title', "Inbox - Index | Red Hat 3scale API Management"
     assert_select '#export-to-csv', false, 'Export all Messages'
   end
+
+  test "creates valid reply" do
```
thalesmiguel, line 52: As a default, we should try to use single quotes for strings that are not interpolated.

test/functional/provider/admin/messages/inbox_controller_test.rb
```diff
@@ -48,4 +48,24 @@ def test_index_system_message
...
+    login_as(@admin)
+    post :reply, params: { message: { subject: "Valid Message", :body => "message with subject" }, id: @message.id }
+
+    MessageWorker.drain
+    msg = Message.last
+    assert "sent", msg.state
+    assert_equal "message with subject", msg.body
```
thalesmiguel, line 59: Instead of only going to the DB to fetch a message and test it out, we could also have assertions that depend on the code that's called in the controller.
Things like
```ruby
assert_equal flash[:notice], 'Reply was sent.'
assert_equal response.redirect_url, 'some url we're expecting...'
```
and so on.

test/functional/provider/admin/messages/inbox_controller_test.rb
```diff
@@ -48,4 +48,24 @@ def test_index_system_message
     assert_select 'title', "Inbox - Index | Red Hat 3scale API Management"
     assert_select '#export-to-csv', false, 'Export all Messages'
   end
+
+  test "creates valid reply" do
+    login_as(@admin)
+    post :reply, params: { message: { subject: "Valid Message", :body => "message with subject" }, id: @message.id }
+
+    MessageWorker.drain
+    msg = Message.last
```
thalesmiguel, line 57: This could cause an intermittent failed test, in case we're running tests in parallel and another unexpected message becomes the last one.
We could change this to
```ruby
message = Message.find_by(sender: @message.receiver)
```
instead to make it safer.

app/controllers/provider/admin/messages/inbox_controller.rb
```diff
@@ -22,11 +22,12 @@ def reply
...
-    reply.deliver!
-
-    flash[:notice] = 'Reply was sent.'
-    redirect_to action: :index
+    if reply.valid?
+      send_reply(reply)
+    else
+      flash[:error] = reply.errors.full_messages.to_sentence
+      redirect_to provider_admin_messages_inbox_path(@message)
+    end
```
thalesmiguel, line 30: We could remote the `send_reply` method by rewriting this like
```ruby
def reply
  reply = @message.reply
  reply.attributes = message_params

  if reply.save && reply.deliver
    flash[:notice] = 'Reply was sent.'
    redirect_to action: :index
  else
    flash[:error] = reply.errors.full_messages.to_sentence
    redirect_to provider_admin_messages_inbox_path(@message)
  end
end
```
"Advantage" is to remove one method.

thalesmiguel (changes requested):
We should change this ticket's name and description. Probably the ticket linked to it as well, for future reference.

This PR leads us believe that, after it's merged, emails will not be sent if they don't have a subject or if they don't have a body where, in reality, this PR adds a message explaining why the emails (that we're **already** preventing to be send) weren't sent.

"Display error message for emails without subjects" would fit this PR better.

test/functional/provider/admin/messages/inbox_controller_test.rb
```diff
@@ -48,4 +48,24 @@ def test_index_system_message
     assert_select 'title', "Inbox - Index | Red Hat 3scale API Management"
     assert_select '#export-to-csv', false, 'Export all Messages'
   end
+
+  test "creates valid reply" do
+    login_as(@admin)
+    post :reply, params: { message: { subject: "Valid Message", :body => "message with subject" }, id: @message.id }
```
jlledom, line 54:
```suggestion
    post :reply, params: { message: { subject: "Valid Message", body: "message with subject" }, id: @message.id }
```
Just a tiny change to use the same syntax as in the rest of the line.

test/functional/provider/admin/messages/inbox_controller_test.rb
```diff
@@ -48,4 +48,24 @@ def test_index_system_message
...
+    assert_equal "message with subject", msg.body
+  end
+
+  test "not creates invalid message" do
+    login_as(@admin)
+    post :reply, params: { message: { subject: nil, body: "message with nil subject" }, id: @message.id }
+
+    MessageWorker.drain
+    msg = Message.last
+    assert_not_equal "message with nil subject", msg.body
```
jlledom, line 68: also `assert "unsent", msg.state` here?

nidhi-soni1104: unsent is the initial state, and if validations fails we don't create object in db  so we check that there is  not last message with our subject which we have sent recently


Messages can be in 1 of 3 states:
# * +unsent+ - The message has not yet been sent.  This is the *initial* state.
# * +queued+ - The message has been queued for future delivery.
# * +sent+ - The message has been sent.

app/controllers/provider/admin/messages/inbox_controller.rb
nidhi-soni1104: this will fails due to code climate

nidhi-soni1104:
> We should change this ticket's name and description. Probably the ticket linked to it as well, for future reference.
> 
> This PR leads us believe that, after it's merged, emails will not be sent if they don't have a subject or if they don't have a body where, in reality, this PR adds a message explaining why the emails (that we're **already** preventing to be send) weren't sent.
> 
> "Display error message for emails without subjects" would fit this PR better.

Good suggestion or may be I can add it in description,  also this is for messages not for emails

test/functional/provider/admin/messages/inbox_controller_test.rb
nidhi-soni1104: Updated assertions

test/functional/provider/admin/messages/inbox_controller_test.rb
nidhi-soni1104: done

test/functional/provider/admin/messages/inbox_controller_test.rb
jlledom: > instead we can do this
> msg = Message.find_by(id: @message.message_id)

You can't because that would return the original message and we want the reply.

app/controllers/provider/admin/messages/inbox_controller.rb
thalesmiguel: > this will fails due to code climate

Do you know what issue CodeClimate has with this? I don't think it hurts Rubocop nor Reek 🤔 

 > we must check if message is valid then only we will save and get it delivered

We can check it with `valid?` as we're doing of course. I just want to point out that by using `if reply.save && reply.deliver`, if `reply.save` fails (due to validation), it will return `false` and `reply.deliver` won't even be executed.

nidhi-soni1104: Okay agreed and updated, let me just check if any rubocop or reek fails

test/functional/provider/admin/messages/inbox_controller_test.rb
nidhi-soni1104: yes I think getting last message is the correct approach we have used the same in other test cases as well

nidhi-soni1104:
@jlledom @thalesmiguel Can we merge this now??

test/functional/provider/admin/messages/inbox_controller_test.rb
thalesmiguel: Fine by me :v:

thalesmiguel (approved):
I'd still change this PR's name and/or the ticket AND the commit message once it's merged to reflect what it's really changing in the code.
We're not preventing messages to be send without a subject because this validation is already on place. With this PR, we'll start showing an error to the application's users, instead of raising an application error.

PR #3096, opened 2022-11-01
clean compatibility code in SphinxIndexationWorker

akostadinov:
we needed this part when we changed enqueuing from passing an object
to passing a class and id. It was only needed until old jobs were
drained.

So we have this code in 2.12 as well 2.13, it should be fine
to get rid of it in 2.14


mayorova (approved):
:+1:

thalesmiguel (approved):
🚀

PR #3099, opened 2022-11-03
THREESCALE-5442: Redact sensitive parameters in logs

jlledom:

Some sensitive parameters are being dumped to the logs without any masking. This PR adds those parameters to the filtering list.

[THREESCALE-5442](https://issues.redhat.com/browse/THREESCALE-5442)

**Verification steps** 

1. Send a request including one of the sensitive parameters:
  `curl http://provider-admin.3scale.localhost:3000//admin/api/accounts.xml?access_token=supersecrettoken`
3. Check how it¡s been traced in the logs
  
image

This filters parameters being sent to the app, but the issue also mentions how some environmental variables with sensitive values are being disclosed in the the pods logs. Those logs are generated by the container engine and are out of our scope. The only way I can think of workarounding it is by sending the sensitive values to the pods mapping the secrets to an `.env` file instead of environmental variables. I added a [comment](https://issues.redhat.com/browse/THREESCALE-5442?focusedCommentId=21195261&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-21195261) in Jira to explain how to do it.

### DIFF
--- a/config/initializers/filter_parameter_logging.rb
+++ b/config/initializers/filter_parameter_logging.rb
@@ -3,8 +3,8 @@
 
 # Configure sensitive parameters which will be filtered from the log file.
 Rails.application.config.filter_parameters += %i[activation_code cms_token credit_card credit_card_auth_code
-                                   credit_card_authorize_net_payment_profile_token credit_card_expires_on
-                                   credit_card_partial_number crypted_password janrain_api_key lost_password_token
-                                   password password_digest payment_gateway_options payment_service_reference salt
-                                   site_access_code sso_key user_key access_token invitation_token encoded_token
-                                   encoded_key variation_key]
+                                                 credit_card_authorize_net_payment_profile_token credit_card_expires_on
+                                                 credit_card_partial_number crypted_password janrain_api_key lost_password_token
+                                                 password password_digest payment_gateway_options payment_service_reference salt
+                                                 site_access_code sso_key user_key access_token invitation_token encoded_token
+                                                 encoded_key variation_key]
--- a/config/initializers/filter_parameter_logging.rb
+++ b/config/initializers/filter_parameter_logging.rb
@@ -6,5 +6,4 @@ Rails.application.config.filter_parameters += %i[activation_code cms_token credi
                                                  credit_card_authorize_net_payment_profile_token credit_card_expires_on
                                                  credit_card_partial_number crypted_password janrain_api_key lost_password_token
                                                  password password_digest payment_gateway_options payment_service_reference salt
-                                                 site_access_code sso_key user_key access_token invitation_token encoded_token
-                                                 encoded_key variation_key]
+                                                 site_access_code sso_key user_key access_token]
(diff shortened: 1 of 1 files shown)


thalesmiguel:
Looks good to me. We just need to fix the CodeClimate warning.

nidhi-soni1104 (approved):
LGTM, Just fix code climate

config/initializers/filter_parameter_logging.rb
```diff
@@ -3,7 +3,8 @@
...
 # Configure sensitive parameters which will be filtered from the log file.
 Rails.application.config.filter_parameters += %i[activation_code cms_token credit_card credit_card_auth_code
-                                   credit_card_authorize_net_payment_profile_token credit_card_expires_on
-                                   credit_card_partial_number crypted_password janrain_api_key lost_password_token
-                                   password password_digest payment_gateway_options payment_service_reference salt
-                                   site_access_code sso_key user_key]
+                                                 credit_card_authorize_net_payment_profile_token credit_card_expires_on
+                                                 credit_card_partial_number crypted_password janrain_api_key lost_password_token
+                                                 password password_digest payment_gateway_options payment_service_reference salt
+                                                 site_access_code sso_key user_key access_token invitation_token encoded_token
```
akostadinov, line 9: I'm not sure what is `encoded_token` but do we need to filter it?

jlledom: You got me, I don't know what `invitation_token`, `encoded_token`, `encoded_key` and `variation_key` are for. But they sound sensitive to me so better safe than sorry.

akostadinov: I think we better figure out as logs are usful and can help debugging issues.

jlledom: `invitation_token` is used in the link sent by mail to invited users
`encoded_token`, `encoded_key` and `variation_key` are rails internals for activestorage

akostadinov: Where these requested in JIRA? How did you decide these need to be filtered out in logs?

Information of encrypted_key and token I find here: https://www.zerodayinitiative.com/blog/2019/6/20/remote-code-execution-via-ruby-on-rails-active-storage-insecure-deserialization

Seems like they are signed by a secret but are not a secret themselves. They indicate what was being GET or PUT. And this is very important when debugging issues.

`variation_key` seems to be which type of blob is requested. Also useful to see in the logs and not something I would filter from system admin. it is described here if I'm not mistaken https://edgeapi.rubyonrails.org/classes/ActiveStorage/Variant.html 

Now should invitation token be filtered from logs. I wouldn't do it unless somebody requested that with a good argument.

jlledom: Done

akostadinov (dismissed):

config/initializers/filter_parameter_logging.rb
akostadinov: Cool. Approved. Just to clarify, when I ask, it doesn't mean that I'm against. If there is a clear rationale behind adding more stuff, I'll be all for it.

jlledom: Yeah, I forgot to answer. It's not clear to me either in the Jira are requesting to filter all sensitive parameters or only `access_token`. I assumed the more parameters filtered the better. So I ran `rake routes` and added to the list all parameters that seemed sensitive to me, like the ones having `key` or `token` in the name. However, API keys like `service_token`, `provider_key` or `app_key` doesn't appear in the routes output but they seem sensitive to me. So I would either filter only `access_token` or `[access_token, service_token, provider_key, app_key]`

nidhi-soni1104 (dismissed):

nidhi-soni1104:
I have manually started your js and my_sql build you can merge PR one this passed

akostadinov (approved):

PR #3103, opened 2022-11-09
THREESCALE-3927 Update Swagger to Open API 3.0

nidhi-soni1104:

Update our own docs(sourcetoswagger) to OAS 3.0(Open API 3.0)

Main Task: https://issues.redhat.com/browse/THREESCALE-3927

There is a subtask: https://issues.redhat.com/browse/THREESCALE-8904 which is **NOT** part of this PR.

Outstanding issues:
1. **Serialization does not work for `POST /transactions.xml` in Service Management API, see details in https://github.com/3scale/porta/pull/3103#issuecomment-1413955116**
**UPDATE:** this is fixed in https://github.com/3scale/porta/pull/3294

2. Auto-complete does not work for all cases:
  - doesn't work in the request body, see https://github.com/3scale/porta/pull/3103#issuecomment-1470120151
  - also doesn't work for metrics in Service Management API, because the examples are provided as JSON objects, the metrics are not defined as separate fields
  I (@mayorova) think this is OK - we may outline this as a limitation in the release notes.
3. The `/p/admin/api_docs/account_data` endpoint is called multiple times - for every API, because it is part of the existing auto-complete logic, which is called for every SwaggerUI invocation. We may want to refactor this part, so that the account data is pulled just one time, and then reused by all of the APIs.

I (@mayorova) think that we need to somehow fix issue 1, and leave 2 and 3 for later.

Also, cleaning up the code and JSON specs that become "dead" after this upgrade will be done in another PR.

**TODO:**
- [x] Tests for the `ThreeScaleApiDocs` JS module
- [x] (Optional) rename the files in `doc/active_docs` to follow snake case (e.g. `service_management_api.json` ?


Gemfile
```diff
@@ -247,6 +247,8 @@ group :development, :test do
   # for `rake doc:liquid:generate` and similar
   gem 'source2swagger', git: 'https://github.com/3scale/source2swagger'
   gem 'unicorn-rails'
+
+  gem 'rswag'
```
jlledom, line 251: You are not going to need this anymore. The library is to generate the jsons from the code, but we decided to leave that for another [issue ](https://issues.redhat.com/browse/THREESCALE-8904)

Gemfile.lock
```diff
@@ -680,6 +680,20 @@ GEM
...
+    rswag-api (2.7.0)
+      railties (>= 3.1, < 7.1)
+    rswag-specs (2.7.0)
+      activesupport (>= 3.1, < 7.1)
+      json-schema (>= 2.2, < 4.0)
+      railties (>= 3.1, < 7.1)
+      rspec-core (>= 2.14)
+    rswag-ui (2.7.0)
+      actionpack (>= 3.1, < 7.1)
+      railties (>= 3.1, < 7.1)
```
jlledom, line 696: For the same reason, you don't need any change in `Gemfile.lock`

config/initializers/rswag_api.rb
```diff
@@ -0,0 +1,14 @@
...
+  # NOTE: If you're using rswag-specs to generate Swagger, you'll need to ensure
+  # that it's configured to generate files in the same folder
+  c.swagger_root = Rails.root.to_s + '/swagger'
+
+  # Inject a lambda function to alter the returned Swagger prior to serialization
+  # The function will have access to the rack env for the current request
+  # For example, you could leverage this to dynamically assign the "host" property
+  #
+  #c.swagger_filter = lambda { |swagger, env| swagger['host'] = env['HTTP_HOST'] }
+end
```
jlledom, line 14: Remove this file too, it's  generated by le rswag library, which we are not going to use

config/initializers/rswag_ui.rb
```diff
@@ -0,0 +1,16 @@
...
+  # NOTE: If you're using rspec-api to expose Swagger files
+  # (under swagger_root) as JSON or YAML endpoints, then the list below should
+  # correspond to the relative paths for those endpoints.
+
+  c.swagger_endpoint '/api-docs/v1/swagger.yaml', 'API V1 Docs'
+
+  # Add Basic Auth in case your API is private
+  # c.basic_auth_enabled = true
+  # c.basic_auth_credentials 'username', 'password'
+end
```
jlledom, line 16: Remove the file

config/routes.rb
```diff
@@ -16,6 +16,8 @@ def call(env)
 
 Rails.application.routes.draw do
 
+  mount Rswag::Ui::Engine => '/api-docs'
+  mount Rswag::Api::Engine => '/api-docs'
```
jlledom, line 20: Also remove this changes, also generated by `rswag`

spec/swagger_helper.rb
```diff
@@ -0,0 +1,43 @@
...
+      ]
+    }
+  }
+
+  # Specify the format of the output Swagger file when running 'rswag:specs:swaggerize'.
+  # The swagger_docs configuration option has the filename including format in
+  # the key, this may want to be changed to avoid putting yaml in json files.
+  # Defaults to json. Accepts ':json' and ':yaml'.
+  config.swagger_format = :json
+end
```
jlledom, line 43: Remove this file

swagger/v1/swagger.json
```diff
@@ -0,0 +1,19 @@
...
+    {
+      "url": "https://{defaultHost}",
+      "variables": {
+        "defaultHost": {
+          "default": "www.example.com"
+        }
+      }
+    }
+  ]
+}
```
jlledom, line 19: Remove this file

jlledom (changes requested):

nidhi-soni1104:
Yes I will remove rswag completely now

nidhi-soni1104:
@jlledom Rswag is removed completely

spec/swagger_helper.rb
jlledom: This is still not removed

jlledom (changes requested):

nidhi-soni1104:
The list of API is updated to open API 3.0 shown on the UI 

Listed API's are
1.Analytics API
2.Billing API
3.Service Management API
4.Account Management API
5. Policy Registry API

Here is the screenshot attached to check how they will look 

Billing API
image


Policy Registry API
image

jlledom:
I observed that `account_data.json` is being called once per call. could we avoid that?

image

jlledom:
I think this tags should be fixed:

image

app/views/provider/admin/api_docs/show.html.erb
```diff
@@ -12,11 +13,42 @@
...
-$(function(){
-   ThreeScale.APIDocs.account_type = 'provider';
-   ThreeScale.APIDocs.host = "<%=j base_url.gsub(%r{/$}, '') %>";
-   ThreeScale.APIDocs.init([]);
-});
+<div class='api-docs-wrap' id = "analytics-api-docs-wrap"></div>
+<div class='api-docs-wrap' id = "billing-api-docs-wrap"></div>
+<div class='api-docs-wrap' id = "service-management-api-docs-wrap"></div>
+<div class='api-docs-wrap' id = "account-management-api-docs-wrap"></div>
+<div class='api-docs-wrap' id = "policy-registry-api-docs-wrap"></div>
```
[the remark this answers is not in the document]
nidhi-soni1104: @jlledom You mean we need to make dynamic urls according to this API endpoint

jlledom:
> I think the `-s20.json` files might have initially been added to upgrade to Swagger 2.0, but if we're upgrading to OpenAPI 3.0 already, then I think these files are not needed (and the code that is used to load these files either). There must be also a rolling upgrade, and if we are just upgrading the docs without rolling upgrade, then we will need to delete that code as well (as it's otherwise convert into dead code).
> 

Yes, there's some code to show some specs in OpenAPI v2 format. That code should be cleaned up. There also a javascript library `ThreeScale.APIDocs` that is currently being used to load the docs in OpenAPI v1 format, which could probably be removed too. I was going to suggest this later, when my current change requests were done.

> Also, I don't know what @3scale/system-ruby think, but I personally don't like spaces if filenames (even less as these files will eventually get loaded by the browser via URL), so maybe we could use this opportunity to rename the specification files and use hyphens instead (I'd prefer hyphens to underscores to be more URL-friendly).

Agree

app/views/provider/admin/api_docs/show.html.erb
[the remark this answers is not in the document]
nidhi-soni1104: Changes are pushed from my end let see what else we can update in JS code

app/javascript/packs/provider_active_docs.js
```diff
@@ -0,0 +1,16 @@
...
+const accountDataUrl = '/p/admin/api_docs/account_data.json'
+
+window.SwaggerUI = (args, serviceEndpoint) => {
+  const responseInterceptor = (response) => autocompleteOAS3(response, accountDataUrl, serviceEndpoint)
+
+  SwaggerUI({
+    ...args,
+    responseInterceptor
+  })
+}
```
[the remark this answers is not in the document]
nidhi-soni1104: because active docs was no longer on updated master and so we created this

app/controllers/api_docs/services_controller.rb
```diff
@@ -115,7 +115,7 @@ def index
   def show
     system_name = params[:id].to_sym
     api_file = (api_files.fetch(system_name) { raise ActiveRecord::RecordNotFound }).dup
-    api_file['apis'] = exclude_forbidden_endpoints(api_file['apis'])
+    # api_file['apis'] = exclude_forbidden_endpoints(api_file['apis'])
```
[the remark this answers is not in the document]
nidhi-soni1104: Let me check this again today

app/views/provider/admin/api_docs/show.html.erb
```diff
@@ -12,11 +13,31 @@
...
-   ThreeScale.APIDocs.account_type = 'provider';
-   ThreeScale.APIDocs.host = "<%=j base_url.gsub(%r{/$}, '') %>";
-   ThreeScale.APIDocs.init([]);
-});
-</script>
+<div class='api-docs-wrap' id = "analytics-api-docs"></div>
+<div class='api-docs-wrap' id = "billing-api-docs"></div>
+<div class='api-docs-wrap' id = "service-management-api-docs"></div>
+<div class='api-docs-wrap' id = "account-management-api-docs"></div>
+<div class='api-docs-wrap' id = "policy-registry-api-docs"></div>
```
jlledom, line 20: I'd prefer to not create a static list of containers. Different APIs are returned for different users, so there will be always some empty container in the code. Also, if we add a new API in the future, we'll need to add a new line here for it. In order to avoid that, it'd be preferable to create the containers dynamically with javascript, one for each API returned by the services call. That is, remove this entire list and replace it by a single `<div id="api-containers">` and then, in the callback for the successful ajax call (`success`) add a new `<div class='api-docs-wrap' id="{api-name}"></div>` to `#api-containers` for each API in the response.

app/views/provider/admin/api_docs/show.html.erb
```diff
@@ -12,11 +13,31 @@
...
+             async: false,
+             success : function(text)
+             {
+                 response = text.apis;
+                 var len = response.length
+                 for (var i = 0; i < len; i++)
+                  path.push(response[i].path)
+             }
+    });
+    path.forEach(function (url) {
```
jlledom, line 37: `path` will be empty at this point, b/c `path` is filled by the `success` callback when the ajax call gets a response. This should be moved inside the `success` callback.

jlledom (changes requested):

app/javascript/packs/provider_active_docs.js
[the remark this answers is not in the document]
jlledom: The only difference between this one and the original is the value of `accountDataUrl`. You could remove the duplicity storing both values in `active_docs.js` and creating a new boolean parameter `provider` for `window.SwaggerUI` at `active_docs.js` to tell the module either to load the provider url or the other one.

Anyway, what's exactly the error if we include the original `active_docs.js` without changes? I think we should be able to use it.

app/views/provider/admin/api_docs/show.html.erb
jlledom: I've seen that the `async` option in the ajax call is set to `false`, this makes it all synchronous, so what I said above is wrong. Anyway, not sure it's the best idea. Opinions?

app/views/provider/admin/api_docs/show.html.erb
nidhi-soni1104: Good Suggestion . Again JS challenge for me

app/views/provider/admin/api_docs/show.html.erb
```diff
@@ -1,9 +1,3 @@
-<%= stylesheet_link_tag 'active-docs/application.css' %>
-<%= javascript_tag do %>
-  window.enableApiDocsProxy = <%= Rails.configuration.three_scale.active_docs_proxy_disabled.blank? %>;
-<% end %>
-<%= javascript_include_tag 'active-docs/application.js' %>
-
```
jlledom, line 6: You probably need to remove the `css`, but do you need to remove the `js` includes? is not that causing any error in the console?

nidhi-soni1104: Yes I also thought to removed only CSS, but removing JS also not showed me any difference so I thought to remove it and then see if works well without these or not

I think I need to keep this part of code, So I will add this again
```
<%= javascript_tag do %>
  window.enableApiDocsProxy = <%= Rails.configuration.three_scale.active_docs_proxy_disabled.blank? %>;
<% end %>
```

jlledom: yes, you shouldn't remove it unless you know what is it for.

nidhi-soni1104: yes added again

app/views/provider/admin/api_docs/show.html.erb
nidhi-soni1104: Done and pushed

app/controllers/api_docs/services_controller.rb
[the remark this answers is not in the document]
jlledom: Any update on this? I don't think we can merge this line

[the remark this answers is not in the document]
nidhi-soni1104: Yes I have checked this today, we are not receiving anything now in api_file['apis']  and that is why this code fails.
Previously we are receiving the content of the api that need to be shown on the view in full format but now we are doing that part from view i.e from the js part of code I have written.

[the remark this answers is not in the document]
nidhi-soni1104: I would say we can remove this line.Let me know your views. if you remember  Joan you only asked me to comment this line and this is the reason behind this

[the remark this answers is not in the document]
jlledom: We couldn't remove the line, this line was ecluding some endpoint for the particular case of being master in an on_premises installation. I fixed it.

jlledom:
> I observed that `account_data.json` is being called once per call. could we avoid that?
> 
> image

Could you work next on fixing this?

nidhi-soni1104:
> hreeScale.APIDocs

sure

nidhi-soni1104:
@jlledom 
I have checked the reason of keep loading account.json for mutliple time is due to every time we create a new dynamic div, we make a call to swagger to list the API and there this is the url given used as one of the parameter so that is the reason it keeps load and I tried but I don't think I found something which make render dynamic div and listing API in open API format and not load it every time.

doc/active_docs/Service Management API (on-premises).json
```diff
@@ -1,693 +0,0 @@
-{
```
jlledom, line 1: I don't think you c an remove this one

jlledom (changes requested):

doc/active_docs/Service Management API (on-premises).json
nidhi-soni1104: Is this like this file will be used on-premise

But there we two files and one with no environment mentioned so considering that a generic one I preferred to  keep that and remove this

jlledom: I'd say it's being used from here:

https://github.com/3scale/porta/blob/df1b415ad931e42ef5d9f51b88223076a74b6319/app/controllers/api_docs/services_controller.rb#L57

Also, the other files you deleted belong to the `:new_provider_documentation` rolling update. If you want to do a proper clean you need to remove the code for all the rolling update, check how the controller looks for the rolling update before deciding which view it loads:

https://github.com/3scale/porta/blob/df1b415ad931e42ef5d9f51b88223076a74b6319/app/controllers/provider/admin/api_docs_controller.rb#L7

I think the cleanup should be done in another PR, better not mix things.

nidhi-soni1104: this sounds a better idea to clean it up later

test/integration/api_docs/services_controller_test.rb
```diff
@@ -88,7 +87,7 @@ def test_show_service_management
         # ignore stats endpoint
         name.to_s.match(name_or_path_regex) && name.to_s !~ /stats/
       }
-      assert_equal actual_backed_api_routes.length, JSON.parse(response.body)['apis'].each_with_object(Set.new, &select_endpoints).length
+      assert_equal actual_backed_api_routes.length, JSON.parse(response.body)['paths'].select{ |url| url.exclude?('stats') &&  url.include?('backend_')}.length
```
jlledom, line 90: I don't quite understand this, but fixing this test is simpler than this. Consider that with the new format, instead of having:

```
'apis': [
  {
    'path': '/path/to/resource.xml'
    ...
  }
]
```

what we have is:

```
'paths': [
  '/url/to/resource.xml': {
    ...
  }
]
```

Knowing that, you only need to adapt the test to this new structure:

```suggestion
      assert_equal actual_backed_api_routes.length, JSON.parse(response.body)['paths'].each_with_object(Set.new, &select_endpoints).length
```

and 

`path = api[0]`

At line `81`

test/integration/api_docs/services_controller_test.rb
```diff
@@ -105,11 +104,11 @@ def test_show_onprem_account_management_api
       select_endpoint = Proc.new { |api| api['path'] == '/admin/api/account_plans/{id}.xml' }
       ThreeScale.config.stubs(onpremises: false)
       get '/api_docs/services/account_management_api.json'
-      assert_not_empty JSON.parse(response.body)['apis'].select(&select_endpoint)
+      assert_not_empty JSON.parse(response.body)['paths'].keys
```
jlledom, line 107: Same as above. At line 104:

`select_endpoint = Proc.new { |api| api == '/admin/api/account_plans/{id}.xml' }`

Then here:

```suggestion
      assert_not_empty JSON.parse(response.body)['paths'].select(&select_endpoint)
```

test/integration/api_docs/services_controller_test.rb
```diff
@@ -105,11 +104,11 @@ def test_show_onprem_account_management_api
...
 
       ThreeScale.config.stubs(onpremises: true)
       get '/api_docs/services/account_management_api.json'
-      assert_empty JSON.parse(response.body)['apis'].select(&select_endpoint)
+      assert_empty JSON.parse(response.body)['paths'].select{ |url|  url.exclude?('admin')}
```
jlledom, line 111: And here:

```suggestion
      assert_empty JSON.parse(response.body)['paths'].select(&select_endpoint)
```

test/integration/api_docs/services_controller_test.rb
```diff
@@ -134,8 +133,7 @@ def test_index_and_show
           show_result = JSON.parse(response.body)
 
           assert_response :success
-          assert show_result.has_key?('basePath')
-          assert show_result.has_key?('apis')
+          assert show_result.has_key?('paths')
```
jlledom, line 136: Maybe add the assertion for` "openapi": "3.X.X"` here as well?

jlledom (changes requested):

app/controllers/api_docs/services_controller.rb
```diff
@@ -115,17 +115,16 @@ def index
...
   end
 
   private
 
-  def exclude_forbidden_endpoints(apis)
-    apis.select do |api|
-      path = api['path']
-      !master_on_premises? || path.exclude?('plan')
+  def exclude_forbidden_endpoints(paths)
+    paths.select do |url|
```
mayorova, line 126: Now that there is just a single line, I'd suggest to make it a one-liner:
```
paths.select { |url| url.exclude?('plan') }
```

app/javascript/packs/provider_active_docs.ts
```diff
@@ -0,0 +1,18 @@
+// We can define the 3scale plugins here and export the modified bundle
+import SwaggerUI from 'swagger-ui'
+import 'swagger-ui/dist/swagger-ui.css'
+
+import { autocompleteOAS3 } from 'ActiveDocs/OAS3Autocomplete'
+
+import 'ActiveDocs/swagger-ui-3-patch.scss'
+
+const accountDataUrl = '/p/admin/api_docs/account_data.json'
```
mayorova, line 9: This file seems to be the exact copy of `app/javascript/packs/active_docs.ts`, only the `accountDataUrl` value changes.

I have no idea how these packs work, but I am wondering if there is way to parameterize this somehow in order to avoid duplicate code.

Maybe @josemigallas can help.

mayorova:
I think the style has much room for improvement...

Add some padding here:
Screenshot from 2023-01-27 12-11-16
 
And tweak the color scheme here:
Screenshot from 2023-01-27 12-11-07

Also, now there's too much whitespace, and it's not easy to distinguish the different APIs, there is no big title like in the current version.
Screenshot from 2023-01-27 12-41-36

Seems that the only way to tell what API you are looking at is by this small URL, e.g. `/api_docs/services/service_management_api.json`


Also, I am wondering - maybe we can use this opportunity to rename the spec files and remove whitespaces.

doc/active_docs/Service Management API (on-premises).json
mayorova: It's OK to clean up in a different PR, but I think this PR should not touch the `*-s20.json` files, and should just leave them as they are, because:
1. They are not used unless `new_provider_documentation: true`
2. If `new_provider_documentation: true` the feature becomes broken
3. This is confusing :)

So, I think we need to restore the original contents of these files, and we can remove them later in a different PR.

mayorova:
Also, I think a nice-to-have would be to use YAML instead of JSON, I think it's easier to read, write and maintain.

mayorova:
Another nice-to-have would be to provide the schemas of the returned objects, but that's a lot of work, we can probably create a JIRA for that (if it does not already exist)

akostadinov:
I would just request the list of items to be cleaned up later to be put into the commit message when we finally merge.

mayorova (changes requested):
Some other issues that I've seen, specifically with the Service Management API:

1. the requests are sent to the admin portal (from which the Active Docs are hosted, e.g. locally http://provider-admin.3scale.localhost:3000), while they should be sent to the Service Management API (e.g. https://su1.3scale.net in production, or something like http://localhost:3001)
2. there are 9 endpoints in the current spec, and only 5 in the upgraded one. The reason is that in the current there are two endpoints for different auth methods: API key (user_key) and App Id / App Key. The upgraded Swagger only seems to be supporting `user_key`.
3. it is not possible to set the following parameters properly through the UI:
    - usage
    - log
    - transactions (for the POST)

They all have a nested structure, the UI should support setting those.

Related to this point, there is also some auto-complete on the "metrics" field (note that in the spec it has `"threescale_name": "metric_names"`, that we may want to bring back.

mayorova:
Account Management and CMS APIs now have tags set for all operations, but the other APIs don't, so all endpoints for them are shown in the **default** group.

The APIs are very small, so it's probably not worth grouping the endpoints, but maybe just set the same tag for all of them, just to replace "default" with something more meaningful and less confusing (e.g. "Analytics", "Master", "Policy", "Billing", etc.)

mayorova:
I've just checked, and OpenAPI v3 will not support having the same path multiple times (so, we can't replicate the current setup with *either* `user_key` param **or** `app_id` and `app_key`.
See this thread for example: https://stackoverflow.com/questions/40495880/how-to-define-different-query-parameters-for-same-path-in-openapi-swagger

Here people suggest some hacks to add this behavior (double slashes or other extra characters on the path), but I think we shouldn't do this nasty tricks.
I'd rather allow both `user_key`, `app_id` and `app_key` params as non-required, and put a clear description on the operations explaining that either `user_key` or a pair `app_id` and `app_key` **MUST** be provided.
Also, the Summary must be reviewed and fixed, as it is not correct now for some endpoints.

test/integration/api_docs/services_controller_test.rb
nidhi-soni1104: Agreed

test/integration/api_docs/services_controller_test.rb
```diff
@@ -61,8 +61,7 @@ def test_index_and_show
         show_result = JSON.parse(response.body)
 
         assert_response :success
-        assert show_result.has_key?('basePath')
-        assert show_result.has_key?('apis')
+        assert show_result.has_key?('paths')
```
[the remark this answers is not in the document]
nidhi-soni1104: Okay let me check and add this

test/integration/api_docs/services_controller_test.rb
```diff
@@ -62,6 +62,7 @@ def test_index_and_show
 
         assert_response :success
         assert show_result.has_key?('paths')
+        assert_equal show_result["openapi"], "3.0.1"
```
jlledom, line 65: Do you think you could assert for a regular expression that matches `/^3\.[0-9]+\.[0-9]+$/` ?

jlledom (changes requested):

test/integration/api_docs/services_controller_test.rb
nidhi-soni1104: but is that really needed to make it so complex when we can directly check the version

jlledom: I think it's good to have, so the test still works for future 3.x.x versions of the spec. The implementation is in fact trivial: only calling `assert_match` instead of `assert_equal`. The hardest part is to get to the regular expression itself, but I'm already providing it.

On the other hand, remember that in Minitest assertions the expected value comes first, so even if you want to use `assert_equal`, it should be `assert_equal "3.0.1", show_result["openapi"]`

nidhi-soni1104: Okay let me add and update this test

nidhi-soni1104:
> Account Management and CMS APIs now have tags set for all operations, but the other APIs don't, so all endpoints for them are shown in the **default** group.
> 
> The APIs are very small, so it's probably not worth grouping the endpoints, but maybe just set the same tag for all of them, just to replace "default" with something more meaningful and less confusing (e.g. "Analytics", "Master", "Policy", "Billing", etc.)

This part is done and pushed

test/integration/api_docs/services_controller_test.rb
nidhi-soni1104: added

test/integration/api_docs/services_controller_test.rb
```diff
@@ -135,7 +135,8 @@ def test_index_and_show
 
           assert_response :success
           assert show_result.has_key?('paths')
-          assert_equal show_result["openapi"], "3.0.1"
+          debugger
```
jlledom, line 138: Breakpoint forgotten here by mistake

nidhi-soni1104: sorry

mayorova:
I've updated the `Service Management API.json` specification, the `transactions.xml` endpoint expects parameters in the **body**, NOT the request params as it was specified.

However, we still face two issues:
1. The server still points to the admin portal URL, but I guess it's just some JS logic on our side that needs to be updated to replace the URL with the one we want.
2. The `transactions` array is not serialized properly...
    The object should be URL-encoded, in the current Swagger this is how the body is generated:
    ```
    'service_token=asdf&service_id=1&transactions%5B0%5D%5Buser_key%5D=first-key&transactions%5B0%5D%5Busage%5D%5Bhits%5D=1&transactions%5B1%5D%5Buser_key%5D=second-key&transactions%5B1%5D%5Busage%5D%5Bhits%5D=1'
    ```
    or decoded:
    ```
    'service_token=asdf&service_id=1&transactions[0][user_key]=first-key&transactions[0][usage][hits]=1&transactions[1][user_key]=second-key&transactions[1][usage][hits]=1'
    ```
    
    While in OpenAPI v3, the objects are represented as JSON (which is fine and works for `usage`), and they do not get serialized the same way, instead the request body looks like this:
    ```
    'service_token=asdf&service_id=1&transactions=%7B%0A%20%20%22user_key%22%3A%20%22first-key%22%2C%0A%20%20%22usage%22%3A%20%7B%0A%20%20%20%20%22hits%22%3A%201%0A%20%20%7D%0A%7D&transactions=%7B%0A%20%20%22user_key%22%3A%20%22second-key%22%2C%0A%20%20%22usage%22%3A%20%7B%0A%20%20%20%20%22hits%22%3A%201%0A%20%20%7D%0A%7D'
    ```
    or decoded:
    ```
    'service_token=asdf&service_id=1&transactions={
        "user_key": "first-key",
        "usage": {
          "hits": 1
        }
      }&transactions={
        "user_key": "second-key",
        "usage": {
          "hits": 1
        }
      }'
    ```
    
    After trying different combinations of settings I haven't managed to make it work as expected, so I am assuming that's just not supported. Here's what Swagger [doc](https://swagger.io/docs/specification/describing-request-body/) says about **Complex Serialization in Form Data**:
    > The serialization rules provided by the `style` and `explode` keywords **only have defined behavior for arrays of primitives and objects with primitive properties**. For more complex scenarios, such as nested arrays or JSON in form data, you need to use the `contentType` keyword to specify the media type for encoding the value of a complex field.
    I tried `contentType: application/x-www-form-urlencoded` for `transactions`, but it still doesn't work :cry:

jlledom:
> However, we still face two issues:
> 
>     1. The server still points to the admin portal URL, but I guess it's just some JS logic on our side that needs to be updated to replace the URL with the one we want.

Yes, the JS part of this PR needs changes to fix this URL issue and:

- active_docs.js/service_active_docs.js duplication
- multiple account.json loading

>     2. The `transactions` array is not serialized properly...
>        The object should be URL-encoded

Why?

>        While in OpenAPI v3, the objects are represented as JSON

I'm not sure I understood the problem correctly, I set `content-type: application/json` and it seems to work:

image

Is that not correct?

mayorova:
@jlledom 

> I'm not sure I understood the problem correctly, I set content-type: `application/json` and it seems to work:

What's the RAW request body though?

jlledom:
> @jlledom
> 
> > I'm not sure I understood the problem correctly, I set content-type: `application/json` and it seems to work:
> 
> What's the RAW request body though?

The same:

image

Is that correct?

mayorova:
@jlledom No, it's not, unfortunately... :(

At this moment, backend only supports form data, see https://github.com/3scale/apisonator/blob/7bcc83fcc351f353a97cf0447cc45405bff1cc3f/lib/3scale/backend/listener.rb#L494-L495

The data in your example must be represented like this:
```
service_token=string&service_id=0&transactions%5B0%5D%5Buser_key%5D=example&transactions%5B0%5D%5Busage%5D%5Bhits%5D=1&transactions%5B1%5D%5Buser_key%5D=example&transactions%5B1%5D%5Busage%5D%5Bhits%5D=1
```

which stands for

```
service_token=string&service_id=0&transactions[0][user_key]=example&transactions[0][usage][hits]=1&transactions[1][user_key]=example&transactions[1][usage][hits]=1
```

nidhi-soni1104:
I have Checked and updated Account Management API permission api, which works well from UI and generate UI also same as it was on master

Now let me check the issue where  server still points to the admin portal URL

jlledom:
> @jlledom No, it's not, unfortunately... :(
> 
> At this moment, backend only supports form data, see https://github.com/3scale/apisonator/blob/7bcc83fcc351f353a97cf0447cc45405bff1cc3f/lib/3scale/backend/listener.rb#L494-L495
> 

Now I understand the problem! thanks for the clarification. I'll spend some time trying to figure this out. Anyway if it were impossible to fix, this would only affect the API Docs, right?

nidhi-soni1104:
while cheking the API's I found a new thing and putting it here to remember, with old version it show list of top five application and whatever we choose its ID gets filled in the box and I think this is something done by java-script 

image

doc/active_docs/Account Management API.json
```diff
@@ -9935,5 +9929,14 @@
       }
     }
   },
-  "components": {}
+  "components": {
+    "schemas":{
+      "Permissions":{
+        "type": "array",
+        "items": {
+          "$ref": "#/components/schemas/Permissions"
```
mayorova, line 9937: I don't get it... The items reference the same schema, but what are the elements?... 

Swagger UI apparently interprets it as an array of strings, but this definition does not look right.

doc/active_docs/Account Management API.json
```diff
@@ -2647,10 +2644,7 @@
             "style": "form",
             "explode": false,
             "schema": {
-              "type": "array",
-              "items": {
-                "type": "object"
-              }
+              "$ref": "#/components/schemas/Permissions"
```
mayorova, line 2647: `allowed_sections` has a predefined list of values, so an enum would be suitable here.

doc/active_docs/Account Management API.json
```diff
@@ -1,8604 +1,9942 @@
...
+            "schema": {
+              "type": "string"
+            }
+          },
+          {
+            "name": "user_id",
+            "in": "query",
+            "description": "ID of the account user.",
+            "schema": {
+              "type": "ref"
```
mayorova, line 336: There are plenty of `"type": "ref"` entries, this is not valid, all ocurrences need to be fixed.

nidhi-soni1104: yes agreed I have fixed one so if anything wrong here can be fixed and then rest can be fixed

doc/active_docs/Account Management API.json
nidhi-soni1104: you mean defining properties with type and description

doc/active_docs/Account Management API.json
nidhi-soni1104: not clear with this point

nidhi-soni1104: done

nidhi-soni1104:
@mayorova  Issue related to Service Management API is resolved and pushed

.byebug_history
```diff
@@ -0,0 +1,10 @@
+exit
```
mayorova, line 1: I guess this was pushed by mistake?

nidhi-soni1104: oh yes

nidhi-soni1104: let me remove this

nidhi-soni1104: done

mayorova:
I've pushed some changes to the Account Management API:
- removed `"format": "int32"`, as this is for **signed** integers, and ours are unsigned
- fixed the invalid `"type": "ref"`
- added the note ` (use Cinstance for Application)` for the description `Target entity of fields definition.` (just a clarification for the users)

The format of the tags is not something important, this is just how https://editor.swagger.io exported the file to JSON from YAML.

mayorova:
@nidhi-soni1104 I've noticed that for PUT endpoints (I've noticed it on a couple, but I am suspecting ALL PUT endpoints are affected).

The original specification puts the parameters into **body** (which is expected for PUT), while the v3 puts it into the query parameters, which is not correct. Can you please fix it?

In OpenAPI v3 the definition of the body parameters is very different from query parameters, you can check this documentation which has some examples as well https://swagger.io/docs/specification/describing-request-body/

nidhi-soni1104:
> @nidhi-soni1104 I've noticed that for PUT endpoints (I've noticed it on a couple, but I am suspecting ALL PUT endpoints are affected).
> 
> The original specification puts the parameters into **body** (which is expected for PUT), while the v3 puts it into the query parameters, which is not correct. Can you please fix it?
> 
> In OpenAPI v3 the definition of the body parameters is very different from query parameters, you can check this documentation which has some examples as well https://swagger.io/docs/specification/describing-request-body/

Sure I will check this, but the thing is what changes you found affected in API is it something reflect in URL formation or response?

mayorova:
> but the thing is what changes you found affected in API is it something reflect in URL formation or response?

I don't understand the question, sorry. What do you mean?
If you are asking what changed in 
https://github.com/3scale/porta/pull/3103/commits/b8823c62626b445d000fc8bdade9c694ce8ee024 exactly - the spec is now valid (it wasn't, even though it was rendering correctly) and some data types are fixed.

> What I have seen is this in query parameters are for all API POST PUT PATCH and even in GET also

Query parameters might be supported in PUT, POST and wherever, but if it's not what our API implements (i.e. if it expects the data in the body, and not query parameters), then it's not valid.

nidhi-soni1104:
> > but the thing is what changes you found affected in API is it something reflect in URL formation or response?
> 
> I don't understand the question, sorry. What do you mean? If you are asking what changed in [b8823c6](https://github.com/3scale/porta/commit/b8823c62626b445d000fc8bdade9c694ce8ee024) exactly - the spec is now valid (it wasn't, even though it was rendering correctly) and some data types are fixed.

No actually this change is clear to me
> 
> > What I have seen is this in query parameters are for all API POST PUT PATCH and even in GET also
> 
> Query parameters might be supported in PUT, POST and wherever, but if it's not what our API implements (i.e. if it expects the data in the body, and not query parameters), then it's not valid.

If we se URL's all our api are expecting query parameters, May be some will expect request body  but how we will differentiate it?

nidhi-soni1104:
> > > but the thing is what changes you found affected in API is it something reflect in URL formation or response?
> > 
> > 
> > I don't understand the question, sorry. What do you mean? If you are asking what changed in [b8823c6](https://github.com/3scale/porta/commit/b8823c62626b445d000fc8bdade9c694ce8ee024) exactly - the spec is now valid (it wasn't, even though it was rendering correctly) and some data types are fixed.
> 
> No actually this change is clear to me
> 
> > > What I have seen is this in query parameters are for all API POST PUT PATCH and even in GET also
> > 
> > 
> > Query parameters might be supported in PUT, POST and wherever, but if it's not what our API implements (i.e. if it expects the data in the body, and not query parameters), then it's not valid.
> 
> If we se URL's all our API are expecting query parameters, May be some will expect request body but how we will differentiate it?

and Query params are common parameters.

nidhi-soni1104:
@mayorova 

**New curl**

curl -X 'PUT' \
  'http://provider-admin.3scale.localhost:3000/admin/api/fields_definitions/{id}.json' \
  -H 'accept: */*' \
  -H 'Content-Type: application/x-www-form-urlencoded' \
  -d 'position=sadas&hidden=true&target=Account&choices%5B%5D=one,two&label=asdsad&read_only=true&id=1&access_token=sdfds&required=true'


**Old Curl**

  curl -v  -X PUT "http://provider-admin.3scale.localhost:3000/admin/api/fields_definitions/1.json" -d 'access_token=asd&target=Account&label=sd&required=true&hidden=true&read_only=true&position=xcfsd&choices%5B%5D=one&choices%5B%5D=sadasd'

mayorova:
@nidhi-soni1104 these requests are not identical

1. `/admin/api/fields_definitions/{id}.json` - this is not right, the {id} needs to be replaced with the real ID of the field definition, the `id` parameter needs to be a **path** parameter, not in the body.

2. Note that the choices array is not represented in the same way
`choices%5B%5D=one,two` vs `choices%5B%5D=one&choices%5B%5D=two`. I am not sure whether the first one would work, have you checked? To have the same representation as the second one, I think maybe `explode: true` can help.

3. When some fields are not filled, they are sent as empty ones, e.g. `position=&hidden=&read_only=` while in the current spec they are not included in the request. I don't know what effect it has on the actual API controller, but I'd say they shouldn't be included if they are empty. But if you confirm that it doesn't matter whether they are there or not, maybe we can accept it.

nidhi-soni1104:
@mayorova for point two I would say these can be given both ways, so I changed this way

image

mayorova:
@nidhi-soni1104 The path param is the easiest, just move the "id" param back from the body to "parameters"

```
        "parameters": [
          {
            "name": "id",
            "in": "path",
            "description": "ID of the fields definition.",
            "required": true,
            "schema": {
              "type": "integer"
            }
          }
```

nidhi-soni1104:
> @nidhi-soni1104 The path param is the easiest, just move the "id" param back from the body to "parameters"
> 
> ```
>         "parameters": [
>           {
>             "name": "id",
>             "in": "path",
>             "description": "ID of the fields definition.",
>             "required": true,
>             "schema": {
>               "type": "integer"
>             }
>           }
> ```
> 
> 

> @nidhi-soni1104 The path param is the easiest, just move the "id" param back from the body to "parameters"
> 
> ```
>         "parameters": [
>           {
>             "name": "id",
>             "in": "path",
>             "description": "ID of the fields definition.",
>             "required": true,
>             "schema": {
>               "type": "integer"
>             }
>           }
> ```

Done, Here is the updated URL

curl -X 'PUT' \
  'http://provider-admin.3scale.localhost:3000/admin/api/fields_definitions/1.json' \
  -H 'accept: */*' \
  -H 'Content-Type: application/x-www-form-urlencoded' \
  -d 'access_token=sd&target=Account&label=sdf&read_only=true&required=true&position=sfd&hidden=true&choices%5B%5D=sdf'

mayorova:
OK (although it seems it's not pushed yet).

Maybe you can go ahead and update the requestBody parameter for all POST and PUT calls.

If there are some small tweaks still needed, we can apply them later.

But please check the `explode` parameter, I think it will be needed in multiple places.

nidhi-soni1104:
> OK (although it seems it's not pushed yet).
> 
> Maybe you can go ahead and update the requestBody parameter for all POST and PUT calls.
> 
> If there are some small tweaks still needed, we can apply them later.
> 
> But please check the `explode` parameter, I think it will be needed in multiple places.

ok

nidhi-soni1104:
@jlledom and @mayorova  I don't know what is wrong with cucumber test cases  they were passed yesterday, I have fixed failed integration tests and API's are also updated from my end.
it would be great if we can review and merge this PR and then create other Jira's to fix new or other issues

mayorova:
@nidhi-soni1104 Maybe try merging the latest master, I think there might be a fix for your failing test https://github.com/3scale/porta/pull/3192

mayorova:
Another feature that has disappeared with this upgrade is auto-complete for service tokens, service IDs, and other IDs.

I don't know how critical it is, and whether it is worth to fix it here, or to do it in another PR.

nidhi-soni1104:
@damianpm You mean creating a New PR with [3scale:master] as the base branch

nidhi-soni1104:
> Another feature that has disappeared with this upgrade is auto-complete for service tokens, service IDs, and other IDs.
> 
> I don't know how critical it is, and whether it is worth to fix it here, or to do it in another PR.

I think you are taking about this, I have added this in previous comment .
we don't know now how we can fix this with the current library, but yes we will do it in the next part if sounds good to you

https://user-images.githubusercontent.com/84672731/217036470-4a61b314-c1f1-4661-b26d-cb89f87e5455.png

nidhi-soni1104:
@mayorova I have rebased this branch with master, now your commits it show partially verified

nidhi-soni1104:
> There is an issue with the response object, it adds the `"servers"` element, which should not be there (the backend doesn't have it, so I guess it's the JS customization of Swagger UI that does that). Screenshot from 2023-03-01 16-48-16

servers is attribute defined in Open API 3.0 format if you will see its defined in each API

For eg

```
"openapi": "3.0.1",
  "info": {
    "title": "Policy Registry API",
    "description": "API for managing the APIcast Policy Registry.",
    "version": "1.0"
  },
  "servers": [
    {
      "url": "/"
    }
  ],
```

mayorova:
@nidhi-soni1104 yes, that's in the specification, the screenshot is for the **response body**, "servers" should not be there.

nidhi-soni1104:
> response body



> @nidhi-soni1104 yes, that's in the specification, the screenshot is for the **response body**, "servers" should not be there.


Okay so that  need to be checked if its the part of open API 3.0 or we can change something to not have servers in response body

nidhi-soni1104:
@mayorova For me also css worked in commn.css but what you pushed looks more good to me.

mayorova:
> There is an issue with the response object, it adds the "servers" element, which should not be there (the backend doesn't have it, so I guess it's the JS customization of Swagger UI that does that).

That's actually a global issue, that affects customer's OpenAPI v3 specifications also, there is a JIRA for this: https://issues.redhat.com/browse/THREESCALE-8612

I'll handle it in another PR.

mayorova:
We still have 2 issues:
1. Autocomplete is broken, see examples of how it works now in production.
For some things (metrics) I think it is not possible to add this auto-complete, but for others should be possible.

**Service tokens (Service Management API)**
Screenshot from 2023-03-06 11-08-55

**Service ID (Service Management API)**
Screenshot from 2023-03-06 11-09-21

**Metrics (Service Management API)**
Screenshot from 2023-03-06 11-09-42

**Account ID (Account Management API)**
Screenshot from 2023-03-06 11-10-08

**Application plan IDs (Account Management API)**
Screenshot from 2023-03-06 11-10-36

There might be others.

2. We still have an issue with `transactions.xml` endpoint and how it serializes the request body. This is explained in [this comment](https://github.com/3scale/porta/pull/3103#issuecomment-1413955116) - point 2.

nidhi-soni1104:
This is screenshot where removed the servers from the response body, but what I found in code that it was explicitly added.

image

mayorova:
Currently, autocomplete only works for the query parameters, and not the ones in the request body.

There are actually some limitations at the moment for the examples in the request body: https://swagger.io/docs/specification/adding-examples/
For example, only a single example (not multiple) is allowed, which kind of makes our autocomplete (which allows picking a value from a list) pretty useless.

So, I think at the moment it doesn't make sense to implement this auto-complete.
We may create a JIRA to add this in future, or we can leave it as it is, and a JIRA can be created if some customer raises this.
We need, however, to include this in the release notes.

app/views/provider/admin/api_docs/show.html.erb
```diff
@@ -1,9 +1,8 @@
-<%= stylesheet_link_tag 'active-docs/application.css' %>
 <%= javascript_tag do %>
   window.enableApiDocsProxy = <%= Rails.configuration.three_scale.active_docs_proxy_disabled.blank? %>;
 <% end %>
+<%= javascript_pack_tag 'provider_active_docs' %>
 <%= javascript_include_tag 'active-docs/application.js' %>
```
josemigallas, line 5: Having these 2 together...



It reeks of technical debt. I see most of the imports are scripts injected into the global scope (💩), also old jquery https://github.com/3scale/porta/blob/25fa9bc53bdd3eeb8f2a9c8e882f771c85c3c3cb/vendor/active-docs/vendor/assets/javascripts/active-docs/vendor.js#L1 
But who knows what's old and what's still used. Maybe some day we can clean it up.

app/javascript/packs/provider_active_docs.ts
```diff
@@ -0,0 +1,20 @@
...
+  const containerId = 'api-containers'
+  const container = document.getElementById(containerId)
+
+  if (!container) {
+    throw new Error('The target ID was not found: ' + containerId)
+  }
+
+  const { baseUrl = '', apiDocsPath = '', apiDocsAccountDataPath = '' } = container.dataset
+
+  await renderApiDocs(container, apiDocsPath, baseUrl, apiDocsAccountDataPath)
```
josemigallas, line 19: This is not awaiting anything, so it's better to explicitly ignore promise:
```suggestion
  void renderApiDocs(container, apiDocsPath, baseUrl, apiDocsAccountDataPath)
```
No need to make this block async and disable the eslint rule.

===
Ideally we should have here a loading/loaded/fail mechanism, and we would use `then` and `catch` and the promise wouldn't have to be ignore. But since we don't have that, let's just ignore it 😄

app/javascript/src/ActiveDocs/OAS3Autocomplete.ts
```diff
@@ -134,5 +148,6 @@ export const autocompleteInterceptor = (response: SwaggerUIResponse, accountData
   if (specUrl !== response.url) {
     return response
   }
+  console.log(JSON.stringify(response))
```
josemigallas, line 151: 🧹

app/javascript/src/Types/SwaggerTypes.ts
```diff
@@ -1,4 +1,18 @@
...
   value: string;
 }[]>
+
+export interface ApiDocsService {
+  name: string;
+  system_name: string;
+  description: string;
+  path: string;
+  service_endpoint: string;
+}
```
josemigallas, line 14: Missing blank line, for consistentcy
```suggestion
}

```

josemigallas:
> 3\. The `/p/admin/api_docs/account_data` endpoint is called multiple times - for every API, because it is part of the existing auto-complete logic, which is called for every SwaggerUI invocation. We may want to refactor this part, so that the account data is pulled just one time, and then reused by all of the APIs.

If this is handled by [autocomplete.coffee](https://github.com/3scale/porta/blob/25fa9bc53bdd3eeb8f2a9c8e882f771c85c3c3cb/app/assets/javascripts/swagger-ui2/autocomplete.coffee) then it definitely can be refactored. However, I strongly suggest to do it after this PR is merged!

app/javascript/src/ActiveDocs/ThreeScaleApiDocs.ts
```diff
@@ -0,0 +1,32 @@
...
+import type { ApiDocsServices } from 'Types/SwaggerTypes'
+
+import { autocompleteInterceptor } from './OAS3Autocomplete'
+
+const getApiSpecUrl = (baseUrl: string, specPath: string): string => {
+  return `${baseUrl.replace(/\/$/, '')}/${specPath}`
+}
+
+const appendSwaggerDiv = (container: HTMLElement, id: string): void => {
+  $(container).append($(`<div class="api-docs-wrap" id="${id}" ></div>`))
```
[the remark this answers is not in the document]
mayorova: Damn! At first it was exactly like the suggestion says - without jQuery, but to me it looked ugly :grimacing: 

But yes,that's true, in this case we can actually delegate stuff to the server when generating the page. I'll refactor.

app/javascript/packs/provider_active_docs.ts
mayorova: I assumed that as `renderApiDocs` itself is async (as it waits for an async function inside), we needed to wait for its also.

app/views/provider/admin/api_docs/show.html.erb
mayorova: Good catch. Actually, I don't know whether we need that here, most likely we don't.

app/javascript/src/ActiveDocs/ThreeScaleApiDocs.ts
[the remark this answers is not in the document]
mayorova: Well, after having a closer look...

The list of APIs is actually dynamic, and depends on whether the current portal is master or provider, and on some other provider permissions.

All that logic is in https://github.com/3scale/porta/blob/master/app/controllers/api_docs/services_controller.rb and it's quite entangled.

We could potentially simplify and move at least parts of it to https://github.com/3scale/porta/blob/master/app/controllers/provider/admin/api_docs_controller.rb and https://github.com/3scale/porta/blob/master/app/views/provider/admin/api_docs/show.html.erb but I would prefer not to do this at this point.

app/views/provider/admin/api_docs/show.html.erb
mayorova: Hm, I thought I commented on this, but don't see the comment now.
I think we might not need this here, actually, so I'll remove for now.

mayorova:
@josemigallas 
> If this is handled by [autocomplete.coffee](https://github.com/3scale/porta/blob/25fa9bc53bdd3eeb8f2a9c8e882f771c85c3c3cb/app/assets/javascripts/swagger-ui2/autocomplete.coffee) then it definitely can be refactored. However, I strongly suggest to do it after this PR is merged!

I believe [autocomplete.coffee](https://github.com/3scale/porta/blob/25fa9bc53bdd3eeb8f2a9c8e882f771c85c3c3cb/app/assets/javascripts/swagger-ui2/autocomplete.coffee) is for Swagger v2. v3 is handed by `app/javascript/src/ActiveDocs/OAS3Autocomplete.ts`

But is it also used for the customers' Active Docs (not only for **our** 3scale docs), so I am also reluctant to refactor this now - or maybe also do it in another PR.

app/javascript/packs/provider_active_docs.ts
jlledom, line 1: I'd like to ensure this file it's actually needed. There's an `active_docs.ts` which if I remember correctly, was being called originally. I don't remember the details, but I know the only problem with that file was the value of `accountDataUrl` had to be different for this PR. I'm sure that problem could be solved by adding some parameter somewhere, instead of creating a new file. Sorry for the lack of precision, but I don't remember the details. Now this `provider_active_docs.ts` is completely different to that original `active_docs.ts`, so, do we still need the old one?

app/javascript/src/ActiveDocs/OAS3Autocomplete.ts
jlledom, line 1: In my tests, the Access Token is not autocompleted, should it? I checked and in the previous version it shows a popup saying `"paste a personal access token with the correct permissions"`, shouldn't that popup appear for OpenAPI3 docs as well?

app/views/provider/admin/api_docs/show.html.erb
```diff
@@ -1,8 +1,4 @@
-<%= stylesheet_link_tag 'active-docs/application.css' %>
-<%= javascript_tag do %>
-  window.enableApiDocsProxy = <%= Rails.configuration.three_scale.active_docs_proxy_disabled.blank? %>;
-<% end %>
-<%= javascript_include_tag 'active-docs/application.js' %>
+<%= javascript_pack_tag 'provider_active_docs' %>
```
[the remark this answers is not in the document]
jlledom: What's that proxy for?

jlledom (commented):
This PR has become so long I lost the track, sorry if I mention something already fixed. Some additional comments:

- Multiple calls to `account_data` are not fixed yet, is it going to be merged in another PR?
- In the forms, I see the `Cancel` button instead of the `Try it out`, and the forms are always enabled. Is that intentional?

mayorova:
@jlledom 

> Multiple calls to account_data are not fixed yet, is it going to be merged in another PR?

This would require a significant (IMO) refactoring in the `OAS3Autocomplete.ts` module, which will also affect `service_active_docs.ts` and `active_docs.ts` packs (along with the newly added `provider_active_docs.ts`, so I would prefer to leave it for later, to avoid this PR becoming even more complex.
 
> In the forms, I see the Cancel button instead of the Try it out, and the forms are always enabled. Is that intentional?

Yes, I changed this behavior by setting [tryItOutEnabled: true](https://github.com/3scale/porta/pull/3103/files#diff-78d4b14487564cb474b2638642cd81f8cbc2626033cfe769bb0fe1f6524af4f6R31) in the SwaggerUI config. Having to click "Try it out" was annoying, and having it enabled by default is more similar to the previous active docs behavior.

app/views/provider/admin/api_docs/show.html.erb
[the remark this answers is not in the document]
mayorova: ApiDocs proxy is a feature available only in SaaS. It makes the requests that are made from SwaggerUI pass through a proxy that has CORS enabled. This is for APIs that do not have CORS enabled - and the feature allows these requests to work OK from the admin or developer portal.
The proxy is based on nginx and its source code is here https://github.com/3scale/cors-proxy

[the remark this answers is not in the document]
mayorova: We don't need it here, because:
1. I think 3scale backend (for Service Mgmt API) has CORS already enabled
2. Requests to system API are using the same domain, so CORS does not apply here

app/controllers/api_docs/services_controller.rb
```diff
@@ -2,6 +2,8 @@
 
 class ApiDocs::ServicesController < FrontendController
 
+  include ApplicationHelper
```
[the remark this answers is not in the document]
mayorova: Actually, yes, `helpers.base_url` also works.

app/controllers/api_docs/services_controller.rb
```diff
@@ -41,7 +43,7 @@ def initialize(name, system_name)
 
     def json
       parsed_content = JSON.parse(file_content)
-      parsed_content['basePath'] = backend_base_host if backend_api?
+      parsed_content['servers'] = [{ 'url' => backend_api? ? backend_base_host : '/' }]
```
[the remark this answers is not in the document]
mayorova: Yes, this is quite ugly, but I don't think we can't do what you suggest...

1. The variable `request` that is used for calculating `base_url` is not available at this point (it is available from the controller actions, but ApiFile is initialized in another place
2. The files are read from file system one time, but then can be accessed from different tenants, so we need to change the base URL for the tenant that it was requested from.
In the original implementation I think rather than the full URL just the path was used.
I needed to set the full URL to allow this check: https://github.com/3scale/porta/pull/3103/files#diff-4f8bdb6882899231dfd4039aec865573cb6a79e61de460f4425e45b5aa966439R148

But in fact, I can adjust that check and maybe assume that if a path (and not the full URL) is used, then the request is for the OAS spec. Not sure if that makes sense though.
Then we wouldn't need to replace `/` with the `base_url`. And also we may skip setting the servers to `/` in the first place, because they are already like that in the JSON files (but I kind of prefered to make this even more explicit).

WDYT?

app/javascript/packs/provider_active_docs.ts
mayorova: `active_docs.ts` is for the providers' own active docs. This is used when a single SwaggerUI is loaded on the page.

For `provider_active_docs.ts` we need to do two steps:
1. Call `/api_docs/services` endpoint that has the list of all specs that should be rendered (depends on whether it's master etc.)
2. For every spec from the response above render a SwaggerUI div

So, the logic is not the same, so yes, I think we do need a separate file.

Also, indeed `active_docs.ts` is used for the developer portal, and the account data URL is different (`/api_docs/account_data.json` vs `/p/api_docs/account_data.json` that is used in the admin portal), but as I said this is not the only change.

app/javascript/src/ActiveDocs/OAS3Autocomplete.ts
mayorova: Yes, good catch.

I actually removed autocomplete for the Access Token explicitly here: https://github.com/3scale/porta/pull/3103/commits/b40039273cc30bf85b53b9821637ba50d52f65ea

In the old UI it didn't look bad, but in the new autocomplete based on examples it looked really ugly and confusing, as it was not really an "example", nor it was inserting any values to the field. IMO the message is not too informative either. We can make it part of the `access_token` field description if needed, but I don't think it's required.

app/controllers/api_docs/services_controller.rb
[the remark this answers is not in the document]
mayorova: Fixed in https://github.com/3scale/porta/pull/3103/commits/81a49b9dfa305f0aa3a60adb3b0877e0b26e7435

app/controllers/api_docs/services_controller.rb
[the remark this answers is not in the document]
mayorova: @jlledom Yes, but as I explained previously, the full URL is needed for this check:  https://github.com/3scale/porta/pull/3103/files#diff-4f8bdb6882899231dfd4039aec865573cb6a79e61de460f4425e45b5aa966439R148

This is the way we know whether the request is for fetching the spec itself, or it's the API "try it out" call.

BTW, I recently saw that the `requestInterceptor` does distinguish calls for spec and API calls, I am wondering whether we could take advantage of that somehow, and pass this info from the requestInterceptor to the responseInterceptor.
But I am not sure I want to complicate the things even further...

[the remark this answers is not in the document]
jlledom: Hi, now I (think I) understand the problem. I tried rewriting the `autocompleteInterceptor` function like this:

```
export const autocompleteInterceptor = (response: SwaggerUIResponse, accountDataUrl: string, serviceEndpoint: string): Promise<Response> | SwaggerUIResponse => {
  try {
    const isSpec = !!JSON.parse(response.text).servers
    if (isSpec) {
      return response
    }
  // eslint-disable-next-line no-empty
  } catch {}

  return autocompleteOAS3(response, accountDataUrl, serviceEndpoint)
}
```

So the way to distinguish between spec and API responses is to deliberately look for something we expect to be in the spec response, for instance: it must be JSON and it must contain the servers attribute, but anything else could work as well. By doing it this way, I could:

- Remove the `update_base_url` method from `services_controller.rb`
- Simplify the `json` method from `services_controller.rb`
- Remove the specUrl parameter for `autocompleteInterceptor`

No sure if this is better 😅 , but at least `services_controller.rb` is simpler. I tried with two tenants and works fine. I didn't try the backend request bc I don't know how to fill the params.

>  But I am not sure I want to complicate the things even further...

Of curse you want.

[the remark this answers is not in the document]
mayorova: Well, I don't think this method is reliable @jlledom ...

The customer's API might have and endpoint returning a JSON, containing `"servers"` field, and then it would be a false positive for `isSpec`.

app/javascript/src/ActiveDocs/OAS3Autocomplete.ts
```diff
@@ -122,17 +136,16 @@ const autocompleteOAS3 = async (response: SwaggerUIResponse, accountDataUrl: str
...
-  if (specUrl !== response.url) {
-    return response
+export const autocompleteRequestInterceptor = (request: SwaggerUIRequest, accountDataUrl: string, serviceEndpoint: string): Promise<SwaggerUIRequest> | SwaggerUIRequest => {
+  if (request.loadSpec) {
+    request.responseInterceptor = (response: SwaggerUIResponse) => autocompleteOAS3(response, accountDataUrl, serviceEndpoint)
   }
-  return autocompleteOAS3(response, accountDataUrl, serviceEndpoint)
-}
+  return request
+}
```
jlledom, line 151: No new line

jlledom (approved):
This is better 😎 👍

mayorova (approved):

jlledom:
> 
> > Multiple calls to account_data are not fixed yet, is it going to be merged in another PR?
> 
> This would require a significant (IMO) refactoring in the `OAS3Autocomplete.ts` module, which will also affect `service_active_docs.ts` and `active_docs.ts` packs (along with the newly added `provider_active_docs.ts`, so I would prefer to leave it for later, to avoid this PR becoming even more complex.
> 

After all the changes you made on `OAS3Autocomplete.ts`, is this still hard to do?

app/javascript/src/ActiveDocs/OAS3Autocomplete.ts
```diff
@@ -100,7 +114,7 @@ export interface Response extends SwaggerUIResponse {
   text: string;
 }
 
-const autocompleteOAS3 = async (response: SwaggerUIResponse, accountDataUrl: string, serviceEndpoint: string): Promise<Response> => {
+export const autocompleteOAS3 = async (response: SwaggerUIResponse, accountDataUrl: string, serviceEndpoint: string): Promise<Response> => {
```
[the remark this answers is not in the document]
mayorova: > It seems this is being exported only for tests. IMO this has became a private function now, so it shouldn't be exported. Tests shouldn't condition the code, they should adapt to the code.

That is right, and I agree with you that ideally the tests should adapt to the code, and not the other way around. On the other hand, when we code, we generally think about making the code testable (splitting codes in smaller units, adding dependency injection mechanism to allow for easier mocking of the dependencies, etc.).

In this case, I would really prefer to test the function `autocompleteOAS3` itself, and not the only "public" function `autocompleteRequestInterceptor`.

I am not sure if there is another _easy_ way to do this, without exporting the function (there are ways, but not too straightforward).

> In this case, the tests to assert `autocompleteOAS3` logic should be included as tests for the **'fetching OpenAPI spec'** context of `autocompleteRequestInterceptor` test, since that's the 'unit' being exported.

I am not sure I understand what you mean :upside_down_face: 

@josemigallas what do you think?

mayorova:
> After all the changes you made on `OAS3Autocomplete.ts`, is this still hard to do?

I don't see it as a trivial task, but I do agree that this might be actually a good time to fix it, as we are changing the interface for all three related packs anyway. I'll give it a spin tomorrow.

app/javascript/src/ActiveDocs/OAS3Autocomplete.ts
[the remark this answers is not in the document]
jlledom: I meant basically move `'should inject servers to the spec'` and `'should autocomplete fields of OpenAPI spec with x-data-threescale-name property'` to line 100 after `'should update the response interceptor'`. But OK.

mayorova:
@jlledom done in https://github.com/3scale/porta/pull/3103/commits/e6a77555d0da71469c99ec97f58b16a92c3f80d2

I'm wondering if @josemigallas will approve :grimacing:

app/javascript/src/ActiveDocs/OAS3Autocomplete.ts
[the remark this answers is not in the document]
mayorova: @jlledom  like this? https://github.com/3scale/porta/pull/3103/commits/98ff758d69a9f501d0736ceb6a5f5b19292a3496

jlledom (approved):

PR #3110, opened 2022-11-15
🥒 Clean up credit card steps a bit

josemigallas:
Needed for https://github.com/3scale/porta/pull/3079


app/views/provider/admin/account/payment_gateways/braintree_blue/edit.html.slim
```diff
@@ -60,8 +60,8 @@ p
 
 = semantic_form_for current_account.customer, as: :customer, url: hosted_success_provider_admin_account_braintree_blue_path(next_step: params[:next_step]), id: 'customer_form' do |form|
   = form.inputs "Personal Details" do
-    = form.input :first_name, required: true
-    = form.input :last_name, required: true
+    = form.input :first_name, required: true, label: 'First name'
+    = form.input :last_name, required: true, label: 'Last name'
```
akostadinov, line 64: should we use some kind of i18n here? I don't know when we should and shouldn't use historically.

josemigallas: Good suggestion. We should use it everywhere IMHO.

features/support/hooks.rb
```diff
@@ -181,8 +181,25 @@
...
+  # stub_request(:get, 'https://payments.sandbox.braintree-api.com/graphql')
+  #   .to_return(status: 200, body: '', headers: {})
+  # stub_request(:post, %r{https://origin-analytics-sand.sandbox.braintree-api.com/.+})
+  #   .to_return(status: 200, body: '', headers: {})
+  # stub_request(:get, %r{https://api.sandbox.braintreegateway.com:443/merchants/.+/client_api/v1/})
+  #   .to_return(status: 200, body: '', headers: {})
+  # stub_request(:post, "https://api.sandbox.braintreegateway.com/merchants/my-payment-gw-mid/customers")
+  #   .to_return(status: 200, body: "", headers: {})
+  # stub_request(:delete, %r{@sandbox.braintreegateway.com/merchants/.+/customers/valid_code})
+  #   .to_return(status: 200, body: '', headers: {})º
```
jlledom, line 193: Why adding commented code?

config/locales/en.yml
```diff
@@ -1095,6 +1096,10 @@ en:
       usage_limit:
         value: Max. value
 
+      user:
+        first_name: 'First name'
+        last_name: 'Last name'
```
akostadinov, line 1101: How is this used? I see `= form.input :first_name, required: true, label: 'First name'` above directly set.

secretsurf:
open

config/locales/en.yml
josemigallas: I think at some point I mixed both approaches. The label is to be picked from `en.yaml` by formtastic, so the lines you refer are probably outdated. I will update it.

PR #3112, opened 2022-11-16
THREESCALE-8903 - Prevent notifications to be sent to non active users

thalesmiguel:

This PR removes non-active users from the list of possible email receivers.

Fixes [THREESCALE-8903](https://issues.redhat.com/browse/THREESCALE-8903).

**Verification steps** 

1. Configure SMTP Settings on 3scale in case it is not configured yet
2. Create a New Temporary Email account (i.e using mailhog)
3. As an already existing Admin (Provider) user - User A - log into the admin portal and invite the new user (Provider) - User B - with the email created in the step above
4. User A should create some new Developer Accounts and Applications
5. User B will receive email notifications triggered by the above events
6. User A should suspend User B by using the Account Management API User Suspend (provider account)
7. User A should create a new Developer Account and a new Application
8. Observe that notifications are still received on the mail box of the Temp Mail account of User B

Another way of fixing this would be to go through the application's permissions and figure out why non active users still retain some permissions to receive notifications, for example.
We already have [another ticket](https://issues.redhat.com/browse/THREESCALE-6318) that touches permissions based on Services (products), in which I expect to implement a better solution for this.
Right now, the query we perform to get a list of Users is too broad and it's also done in the last possible moment of the email message workflow, leaving the question to send or not the notification in the hast of Notification#should_deliver?. Maybe there's a way to do this so that we don't need to create more instances of Notification than we currently do.

### DIFF
--- a/app/workers/process_notification_event_worker.rb
+++ b/app/workers/process_notification_event_worker.rb
@@ -43,7 +43,7 @@ class ProcessNotificationEventWorker
     end
 
     parallelize do
-      provider.users.but_impersonation_admin.find_each do |user|
+      provider.users.active.but_impersonation_admin.find_each do |user|
         UserNotificationWorker.perform_async(user.id, event.event_id, event.system_name)
       end
     end
--- a/test/workers/process_notification_event_worker_test.rb
+++ b/test/workers/process_notification_event_worker_test.rb
@@ -21,7 +21,7 @@ class ProcessNotificationEventWorkerTest < ActiveSupport::TestCase
     provider     = FactoryBot.create(:simple_provider)
     event        = Invoices::InvoicesToReviewEvent.create_and_publish!(provider)
     notification = NotificationEvent.create_and_publish!(:invoices_to_review, event)
-    user         = FactoryBot.create(:simple_admin, account: provider)
+    user         = FactoryBot.create(:simple_admin, state: :active, account: provider)
 
     user.create_notification_preferences!(preferences: { invoices_to_review: true })
 
--- a/test/workers/process_notification_event_worker_test.rb
+++ b/test/workers/process_notification_event_worker_test.rb
@@ -50,6 +50,29 @@ class ProcessNotificationEventWorkerTest < ActiveSupport::TestCase
     refute @worker.create_notifications(notification)
   end
 
+  def test_only_send_notifications_for_active_users
+    provider = FactoryBot.create(:simple_provider)
+    system_name = 'application_created'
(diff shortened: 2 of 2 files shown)


thalesmiguel:
> we should disable any operations going on in the background with suspended users that result in notifications

Could you please specify this?
From what I could gather, those background jobs that result in notifications are defined in [EventStore::Repository](https://github.com/3scale/porta/blob/master/app/lib/event_store/repository.rb), which ultimately calls [ProcessNotificationEventWorker](https://github.com/3scale/porta/blob/master/app/workers/process_notification_event_worker.rb), where the query to get all users happens.

> If we just silence notifications but keep things going on in the background

Are there other places where those background jobs are defined? What are the "things going on in the background"?

I'd really appreciate more info here 👍

akostadinov:
I don't know what are the background things. The point is that something happens to create notifications. So we better stop these things than just silencing the result of them.

thalesmiguel:
> I don't know what are the background things. The point is that something happens to create notifications. So we better stop these things than just silencing the result of them.

The notifications are create in [ProcessNotificationEventWorker](https://github.com/3scale/porta/blob/master/app/workers/process_notification_event_worker.rb), where there's a query that [fetches all users from a given provider](https://github.com/3scale/porta/blob/master/app/workers/process_notification_event_worker.rb#L46). One notification is created for each user found and they are persisted and sent if the checks pass. This PR limits the range of the query so we avoid even trying to create the notifications for non active users.
The events that trigger it are defined in [EventStore::Repository](https://github.com/3scale/porta/blob/master/app/lib/event_store/repository.rb). Those look like events that could interest a large amount of users inside an Account and, as I see it, we have to direct the messages to the ones that are able and are willing to receive them. Sounds like permissions? 🤔 

I added it to this PR's description that there's another open ticket that will touch email notifications and permissions as well and that that one would require more work than this. The idea was to have this simpler fix deployed before a bigger refactor.

akostadinov:
> Those look like events that could interest a large amount of users inside an Account

According to THREESCALE-8903, the account is suspended. Then what happened to trigger an event? Why are events triggered for suspended accounts?

thalesmiguel:
> According to THREESCALE-8903, the account is suspended. Then what happened to trigger an event? Why are events triggered for suspended accounts?

Actually, it says a admin **User** is suspended, not the account. So one of those we see in
Screenshot 2022-11-17 at 11 20 41

mayorova (approved):
Well done :+1:

akostadinov (approved):
We discussed my initial concern in a call and JIRA was clarified. All seems nice.

PR #3115, opened 2022-11-17
MAS Pipeline: Fix senging parameters to the reused workflow

jlledom:
The pipeline failed (surprisingly), now we'll have to wait another three weeks to see if this worked.


.github/workflows/fast-forward-manual.yml
```diff
@@ -14,4 +14,7 @@ on:
 jobs:
   call-ff:
     uses: ./.github/workflows/fast-forward-branch.yml
+    with:
+      ref: ${{ github.event.inputs.ref }}
+      to_branch: ${{ github.event.inputs.to_branch }}
```
akostadinov, line 19: You removed `github.event` from other places. Why keep it here?

.github/workflows/fast-forward-branch.yml
```diff
@@ -4,11 +4,9 @@ on:
     inputs:
       ref:
         description: 'The branch name or commit to fast-forward from'
-        default: 'master'
         type: string
       to_branch:
         description: 'The branch name to fast-forward to'
-        default: 'managed-services'
```
akostadinov, line 11: why remove the defaults?

.github/workflows/fast-forward-manual.yml
jlledom: That was the source of the problem. The Github event in this case is `workflow_dispatch`, that is, manually triggered, and `github.event.inputs` are the inputs sent when manually triggering the pipeline. Now check this link: https://stackoverflow.com/a/71458860 

This file `fast-forward-manual.yml` is geting some event inputs, but in the scheduled one `fast-forward-managed-services.yml` the 'event' is automatic and gets no inputs, those empty inputs where inherited by the reusable workflow.

So in `fast-forward-manual.yml` we want to get the inputs from the event. But in the reusable workflow `fast-forward-branch.yml` we want to get the inputs from `input` context, which contains the parameters sent to the workflow call.

.github/workflows/fast-forward-branch.yml
jlledom: I just changed my mind and thought it was better to remove the defaults so it fails if the inputs are not being sent correctly, which is what was happening.

.github/workflows/fast-forward-manual.yml
akostadinov: in stackoverflow it is `github.event.input.some_other_value`. That means `input`, not `inputs`. But given that worked before, I guess it is fine.

akostadinov (approved):

.github/workflows/fast-forward-manual.yml
jlledom: The [contexts](https://docs.github.com/en/actions/learn-github-actions/contexts) are called `github.event.inputs` and `inputs` in the docs. The stackoverflow answer has a typo on it.

PR #3127, opened 2022-12-06
Fix httpclient fork

gsaslis:

Updates the fork used for the `httpclient` ruby gem

same exact code on both forks - the 3scale one having the benefit that it doesn't depend on an external (nowadays) contributor.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -95,7 +95,7 @@ gem 'ransack', '2.4.1' # we can remove line when stop using ruby 2.4
 gem 'browser', '~> 5.0.0' # we can update to lts when we stop using ruby 2.4
 gem 'diff-lcs', '~> 1.2'
 gem 'hiredis', '~> 0.6.3'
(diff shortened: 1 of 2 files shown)


mayorova (approved):

mayorova (changes requested):
The `ssl-env-cert` branches in the two repos are not actually equal:
https://github.com/mikz/httpclient/compare/ssl-env-cert...3scale:httpclient:ssl-env-cert

Once Gemfile.lock is regenerated, the code will not be the same.

We need to investigate which one is desired.

gsaslis:
Great catch @mayorova !! 👏🏼 

I've updated the PR to point to the expected git commit hash

akostadinov (approved):

akostadinov:
They used `require_relative` instead of `require` to fix an issue affecting cachito. It is valid change and follows community practices as far as I've investigated earlier.

gsaslis:
@mayorova I think you need to approve in order for this to be merged. ;)

mayorova (approved):

PR #3134, opened 2022-12-14
Prevent alerts to be sent to users who don't have access to the related service

thalesmiguel:

This PR updates how we grant the permission to `:show AlertRelatedEvent`s. Right now, any provider on an account has this permission, even though they might not have permissions for a service the alert refers to. With this PR, we're adding a check on service level to prevent the issue reported on the linked ticket. 

fixes [THREESCALE-6318](https://issues.redhat.com/browse/THREESCALE-6318)

**Verification steps** 

Verification steps copied from the linked ticket:
1) As Admin user, create a Product i.e. Test API, an application and application plan. Set a limit of 2 requests per minute.
2) Create a member i.e “Tom” user with the following rights:
```
    Accounts – Applications
    Analytics
    Integration & Application Plans
    Disable “All current and future APIs” and check 1 APi, i.e. Test API. 
```
3) Sing in at the admin portal as Tom user and enable the following:
    - Go to the Gear icon - Settings > Notification Preferences > Usage Alerts. Check both “Alert usage warning” and “Alert usage violation”. Then Click Update Notificaton Preferences. See screenshot Notification_preferences.png.
    - Go to Product Cars > Applications > Settings - Usage Rules > Alerts, check both “Show Web Alerts to Admins of this Account” and “Send Email Alerts to Admins of this Account”, from 50% to 100%, see screenshot Alerts.png. Both alerts are necessary due to this issue [THREESCALE-1870](https://issues.redhat.com/browse/THREESCALE-1870).
4) Repeat the same steps 2 & 3 for another user, i.e. Marta, however don't allow access to Test API but to another API. See screenshot Marta.png.
5) Make a request to Test API until hit the limit. Check the emails and you’ll see that Marta has received an alert email regarding Test Api which hasn’t access.

To test this, I ended-up having to run `Events.async_fetch_backend_events!` on a Rails console to force System to fetch the alert event from Apisonator. This might be happening due to a misconfiguration on my local environment but, just for the sake of sharing, I wanted to add this note to the PR.

### DIFF
--- a/config/abilities/provider_any.rb
+++ b/config/abilities/provider_any.rb
@@ -41,8 +41,11 @@ Ability.define do |user|
     end
 
     if user.has_permission?(:monitoring)
-      can [:show], AlertRelatedEvent
+      can [:show], AlertRelatedEvent do |event|
+        user.has_access_to_service?(event.try(:service))
+      end
     end
+
     can [:show], Reports::CsvDataExportEvent do |event|
       user.admin? && event.recipient.try!(:id) == user.id
     end
--- a/config/abilities/provider_any.rb
+++ b/config/abilities/provider_any.rb
@@ -1,7 +1,8 @@
-# here we define abilities for all users of provider account, for members use
-# provider_member.rb
-Ability.define do |user|
-  if user && user.account.provider?
+# frozen_string_literal: true
+
+# here we define abilities for all users of provider account, for members use provider_member.rb
+Ability.define do |user| # rubocop:disable Metrics/BlockLength
+  if user&.account&.provider?
     account = user.account
 
     account.settings.switches.each do |switch, value|
@@ -17,17 +18,15 @@ Ability.define do |user|
     cannot :update_permissions, User, &:admin?
 
     # Can't destroy or update role of himself.
-    cannot [:destroy, :update_role], user
+    cannot %i[destroy update_role], user
 
     # Services
-    can [:read, :show, :edit, :update], Service, user.accessible_services.where_values_hash
+    can %i[read show edit update], Service, user.accessible_services.where_values_hash
 
     #
     # Events
     #
-    if user.has_permission?(:finance)
-      can [:show], BillingRelatedEvent
-    end
+    can [:show], BillingRelatedEvent if user.has_permission?(:finance)
 
     if user.has_permission?(:partners)
       can [:show], AccountRelatedEvent do |event|
@@ -47,7 +46,7 @@ Ability.define do |user|
     end
 
     can [:show], Reports::CsvDataExportEvent do |event|
-      user.admin? && event.recipient.try!(:id) == user.id
+      user.admin? && event.recipient&.id == user.id
     end
 
     can :admin, :social_logins if account.settings.branding.visible?
@@ -75,13 +74,13 @@ Ability.define do |user|
     end
 
     can :reply, Topic do |topic|
-      forum = topic.forum || topic.category.try!(:forum)
+      forum = topic.forum || topic.category&.forum
       forum.account == account
     end
 
     if account.provider_can_use?(:new_notification_system)
-      can [:show, :edit, :update], NotificationPreferences, user_id: user.id
-      can [:show, :update], NotificationPreferences, &:new_record?
+      can %i[show edit update], NotificationPreferences, user_id: user.id
+      can %i[show update], NotificationPreferences, &:new_record?
     end
 
     if account.partner?
@@ -101,8 +100,6 @@ Ability.define do |user|
 
     can :manage, AccessToken, owner_id: user.id
 
-    if can?(:manage, :partners) && can?(:manage, account)
-      can :export, :data
-    end
+    can :export, :data if can?(:manage, :partners) && can?(:manage, account)
   end
 end
--- a/test/unit/abilities/provider_any_test.rb
+++ b/test/unit/abilities/provider_any_test.rb
@@ -69,6 +69,49 @@ module Abilities
       assert_cannot ability, :show, Accounts::AccountCreatedEvent.create(account, user)
     end
 
+    class ShowAlertRelatedEventTest < ProviderAnyTest
+      setup do
+        user.stubs(:has_permission?)
+
+        cinstance = FactoryBot.build_stubbed(:simple_cinstance)
+        alert = FactoryBot.build_stubbed(:limit_violation, cinstance: cinstance)
+        @limit_violation_reached_provider_event = Alerts::LimitViolationReachedProviderEvent.create(alert)
+      end
+
+      attr_reader :limit_violation_reached_provider_event
+
+      test 'cannot show AlertRelatedEvent when user does not have :monitoring' do
+        user.expects(:has_permission?).with(:monitoring).returns(false)
+
+        assert_cannot ability, :show, limit_violation_reached_provider_event
+      end
+
+      # This is related to the :service_permissions rolling update. Users created before it do not have the :services
+      # section included in their member permissions, which grants them access to all services.
+      test 'can show AlertRelatedEvent when user has :monitoring and the user has access to all services through the old permission system' do
+        user.expects(:has_permission?).with(:monitoring).returns(true)
+
+        assert_can ability, :show, limit_violation_reached_provider_event
+      end
+
+      test 'cannot show AlertRelatedEvent when user has :monitoring and does not have access to the service' do
+        user.stubs(:has_permission?).with(:monitoring).returns(true)
+        user.member_permissions.build(admin_section: :services, service_ids: [])
+
+        assert_cannot ability, :show, limit_violation_reached_provider_event
+      end
+
+      test 'can show AlertRelatedEvent when user has :monitoring and has access to the service' do
+        user.stubs(:has_permission?).with(:monitoring).returns(true)
+        user.member_permissions.build(
+          admin_section: :services,
+          service_ids: [limit_violation_reached_provider_event.service.id]
+        )
+
+        assert_can ability, :show, limit_violation_reached_provider_event
+      end
+    end
+
     private
 
     def ability


test/unit/abilities/provider_any_test.rb
```diff
@@ -69,6 +69,47 @@ def test_policies_not_tenant
...
+        user.expects(:has_permission?).with(:monitoring).returns(false)
+
+        assert_cannot ability, :show, limit_violation_reached_provider_event
+      end
+
+      test 'can show AlertRelatedEvent when user has :monitoring and does not have a service' do
+        user.expects(:has_permission?).with(:monitoring).returns(true)
+
+        assert_can ability, :show, limit_violation_reached_provider_event
+      end
```
jlledom, line 93: I don't understand this test. The user can receive alerts for all services when it doesn't have permission over any of them?

thalesmiguel: Oh... That's a bad name. Should be something like "can show AlertRelatedEvent when user has :monitoring and the alert is not related to a specific service".
Updated in [ea018de](https://github.com/3scale/porta/pull/3134/commits/ea018dee8c8642478003730a4bfb9ffe5f48cd97).

thalesmiguel: Oh... That's a bad name.
Should be something like "can show AlertRelatedEvent when user has :monitoring and the user has no services (old permissions)".
Which's true because of [this comment](https://github.com/3scale/porta/blob/master/app/models/user/permissions.rb#L96-L98). Basically, we're supporting an old permission system where users have no services.
Updated in [ea018de](https://github.com/3scale/porta/pull/3134/commits/ea018dee8c8642478003730a4bfb9ffe5f48cd97).

jlledom (approved):

test/unit/abilities/provider_any_test.rb
```diff
@@ -69,6 +69,47 @@ def test_policies_not_tenant
...
+
+      attr_reader :limit_violation_reached_provider_event
+
+      test 'cannot show AlertRelatedEvent when user does not have :monitoring' do
+        user.expects(:has_permission?).with(:monitoring).returns(false)
+
+        assert_cannot ability, :show, limit_violation_reached_provider_event
+      end
+
+      test 'can show AlertRelatedEvent when user has :monitoring and the user has no services (old permissions)' do
```
mayorova, line 89: Do you mean `:service_permissions` rolling update? Maybe it would be nice to add a comment about it.

thalesmiguel: The "old permissions" part came from [this comment](https://github.com/3scale/porta/blob/master/app/models/user/permissions.rb#L96-L98) 😬 . What comment do you think would be good in here?
Something like "Check User::Permissions#has_access_to_all_services?"?

thalesmiguel:
> would it make sense to move the check to provider_member.rb instead, as provider admin (not member) by definition has access to all services.

Good question, @mayorova.
This is the only place where we set permissions for events, so it's not clear to me if we could move all of them to be set on `provider_member.rb`.
By doing so, we'd also need to set those permissions on `provider_admin.rb`, but without checking for services, so the difference between those roles is clear. 
Looks like things work correctly now because `has_access_to_service?` is able to determinate correctly which services a member or admin has access to 🤔
From what I can gather looking at the current code, I think it makes sense to try and set this permission for any provider, regardless of the role but I might be missing something...
All that to say that I'm not 100% sure one option makes more sense over the other. As a guideline, I try to have permissions set in a less places as possible but, of course, sometimes it makes sense to have it spread.

test/unit/abilities/provider_any_test.rb
mayorova: Thanks @thalesmiguel ! 

I checked this further, and it does seem to be related to the `:service_permissions` rolling update.

So, the test description is not really correct then. I think it should say something like (maybe it's too long!):
```
can show AlertRelatedEvent when user has :monitoring and the user has access to all services through old permission system
```

So, it's not that the user does not have service, it's that the member permissions of the user do not include the `services` section, and this happens for users before `:service_permissions` rolling update was introduced.

Maybe you can reference this rolling update in the comment so it is clearer.

mayorova:
Yep, I am not sure whether there would be any benefit in splitting it between the two files (_member and _admin). So let's keep it where it is.

config/abilities/provider_any.rb
```diff
@@ -98,8 +101,7 @@
 
     can :manage, AccessToken, owner_id: user.id
 
-    if can?(:manage, :partners) && can?(:manage, account)
-      can :export, :data
-    end
+    can :export, :data if can?(:manage, :partners) && can?(:manage, account)
   end
+  # rubocop:enable Metrics/BlockLength
```
akostadinov, line 106: IMO it's better to approve issues in code climate before merge than permanently disabling checks in source. In this way by a later refactoring, one may decide to actually fix these.

akostadinov (commented):
Looks nice!

I have one concern though. Do our users have the use case of a monitoring user that is not allowed to use the APIs? Because such use cases would be broken.

I'm not saying it makes sense. Just asking.

test/unit/abilities/provider_any_test.rb
thalesmiguel: Thanks for the explanation, @mayorova!
And nothing wrong with long names if it makes life easier 😬 
Updated in 972db04.

config/abilities/provider_any.rb
```diff
@@ -17,17 +19,15 @@
     cannot :update_permissions, User, &:admin?
 
     # Can't destroy or update role of himself.
-    cannot [:destroy, :update_role], user
+    cannot %i[destroy update_role], user
 
     # Services
-    can [:read, :show, :edit, :update], Service, user.accessible_services.where_values_hash
+    can %i[read show edit update], Service, user.accessible_services.where_values_hash
```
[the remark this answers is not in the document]
thalesmiguel: > I suggest we don't update unrelated lines just for style

I'd argue that, since we have our code linters "enforced" by Code Climate, we should walk in the direction of having more standardised code. The changes in this file, for example, came from Code Climate complaints. I agree this should be discussed with the rest of the team.

> Also it is not easily understandable for me what actually was the fix.

Do you mean in this line 25? If so, this is about the [Style/SymbolArray cop](https://www.rubydoc.info/github/bbatsov/RuboCop/RuboCop/Cop/Style/SymbolArray).

config/abilities/provider_any.rb
```diff
@@ -41,10 +41,13 @@
...
     if user.has_permission?(:monitoring)
-      can [:show], AlertRelatedEvent
+      can [:show], AlertRelatedEvent do |event|
+        user.has_access_to_service?(event.try(:service))
+      end
     end
+
     can [:show], Reports::CsvDataExportEvent do |event|
-      user.admin? && event.recipient.try!(:id) == user.id
+      user.admin? && event.recipient&.id == user.id
```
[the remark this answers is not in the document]
thalesmiguel: They're actually equivalent. `try!` (mind the bang) will raise a `NoMethodError` in case `recipient` doesn't respond to `id`, as the safe navigation operator does.
You may try it out with something like:
```ruby
123.try!(:upcase) #=> NoMethodError
123&.upcase #=> NoMethodError
nil.try!(:update) #=> nil
nil&.update #=> nil
```

config/abilities/provider_any.rb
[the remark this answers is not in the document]
mayorova: > Also it is not easily understandable for me what actually was the fix.

I think @akostadinov means (at least this is my opinion also) that when looking at all of the PR changes and there are a lot of style fixes, it's not easy to see what changes are the actual fix, and what are just style updates.

But I like what @thalesmiguel did, and I do it myself - split these change sets in different commits, although it will only be useful if the PR is merged with merge commit and not "squash and merge".

I've added it as an item for discussion in [our workflow doc](https://docs.google.com/document/d/1dNlcUVV2ozZSpnDXgtL5LTu4uRvQ5ZX_xuNVszPfmqc/edit#heading=h.nfaon3vyid9p), we can chat about it later :wink:

config/abilities/provider_any.rb
mayorova: Or, if we don't care about some rules, we can as well just disable them.

thalesmiguel: I'm used to have the cops disabled directly in the code. The reasoning is that it makes clear for any developer that the file has some style (or worse) issue that was caught but not addressed in the past. It also puts the code as the source of truth instead of a 3rd party application. (I actually had troubles with it in the past once a Company stopped using CodeClimate in favour of another application).

config/abilities/provider_any.rb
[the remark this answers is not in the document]
thalesmiguel: > although it will only be useful if the PR is merged with merge commit and not "squash and merge".

If it's one of my PRs, you can be sure I'll prever to squash and merge things 😬 

> I've added it as an item for discussion in [our workflow doc](https://docs.google.com/document/d/1dNlcUVV2ozZSpnDXgtL5LTu4uRvQ5ZX_xuNVszPfmqc/edit#heading=h.nfaon3vyid9p), we can chat about it later 😉

YOU. ARE. THE. BEST!

thalesmiguel:
> I have one concern though. Do our users have the use case of a monitoring user that is not allowed to use the APIs? Because such use cases would be broken.

This is interesting @akostadinov and MAYBE 3Scale supports that already.
I think this could be accomplished by saying a member has `Access & query analytics of` `All current and future existing API products` (or a set of products) 🤔 
Screenshot 2022-12-20 at 10 14 52

If it's something used at all, I guess Product would be able to answer us.

jlledom (dismissed):

mayorova (dismissed):
LGTM :+1:

config/abilities/provider_any.rb
josemigallas: In general I approve with @akostadinov but this one particularly will never be fixed 🤷🏻‍♂️ `Metrics/BlockLength`.

config/abilities/provider_any.rb
```diff
@@ -1,7 +1,9 @@
-# here we define abilities for all users of provider account, for members use
-# provider_member.rb
+# frozen_string_literal: true
+
+# here we define abilities for all users of provider account, for members use provider_member.rb
+# rubocop:disable Metrics/BlockLength
 Ability.define do |user|
```
josemigallas, line 5: To avoid having to enable the check again, you can disable it for the block:
```suggestion
Ability.define do |user| # rubocop:disable Metrics/BlockLength
```

config/abilities/provider_any.rb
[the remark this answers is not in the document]
josemigallas: I think we do have to do these styling changes at the same time BUT not mixed within the same commit as the fix whatever. To me this PR only lacks 1 proper rebase:
- commit 1: fixes X prblem
- commit 2: 🚨 fixes rubocop warnings

josemigallas (changes requested):
I only request the commits are rebased properly as discussed

thalesmiguel:
> I only request the commits are rebased properly as discussed

@josemigallas  My plan was actually to Squash and merge this PR, as I've been doing. Is it **really** a blocker for this PR not to have the 2 commits as suggested?

config/abilities/provider_any.rb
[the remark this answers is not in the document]
thalesmiguel: I'd like to further discuss this "messes up the history" point outside of this PR. I really don't see it as a shortcoming when compared with the benefits of a standardised code and I'd also argue that the history doesn't really get "messed up". I'd like to understand better the problems that happened here in the past so I can better grasp this issue.
I added a topic to [our workflow doc](https://docs.google.com/document/d/1dNlcUVV2ozZSpnDXgtL5LTu4uRvQ5ZX_xuNVszPfmqc/edit#heading=h.nfaon3vyid9p).

thalesmiguel:
Force pushed in order to have 2 commits: One for the fix and another one for the Rubocop warnings, as per https://github.com/3scale/porta/pull/3134#discussion_r1055452059

josemigallas (approved):

PR #3135, opened 2022-12-19
THREESCALE-8990: CMS access token scope: make  public

jlledom:

This removes the rolling update for the CMS API and makes it public for everyone. This is one of the tasks required to release the tech preview.

[THREESCALE-8990](https://issues.redhat.com/browse/THREESCALE-8990)

**Verification steps** 

In the provider portal:

1. Account Settings -> Personal -> Tokens
2. Create or edit a token
3. The check button for the `Developer portal` scope must appear.
4. When `Developer portal` or `Account` tokens are enabled, the CMS API must accept the requests. It must deny the requests otherwise.

image

### DIFF
--- a/app/models/access_token.rb
+++ b/app/models/access_token.rb
@@ -47,12 +47,6 @@ class AccessToken < ApplicationRecord
       end
     end
 
-    def visible_for(account)
-      select_and_build do |scope|
-        account.provider_can_use?("#{scope.value}_api")
-      end
-    end
-
     def keys
       scopes.map(&:key)
     end
@@ -155,7 +149,7 @@ class AccessToken < ApplicationRecord
   end
 
   def available_scopes
-    owner.allowed_access_token_scopes.visible_for(owner.account)
+    owner.allowed_access_token_scopes
   end
 
   def human_scopes
--- a/app/models/access_token.rb
+++ b/app/models/access_token.rb
@@ -38,7 +38,7 @@ class AccessToken < ApplicationRecord
     private
 
     def non_public_scopes
-      %w(cms).freeze
+      %w[].freeze
     end
   end
 
--- a/app/models/access_token.rb
+++ b/app/models/access_token.rb
@@ -30,16 +30,6 @@ class AccessToken < ApplicationRecord
       else value.to_s.to_sym
       end
     end
-
-    def public?
-      non_public_scopes.exclude?(value.to_s)
-    end
-
-    private
-
-    def non_public_scopes
-      %w[].freeze
-    end
   end
 
   class Scopes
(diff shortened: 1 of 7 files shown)


jlledom:
> What about
> 
> https://github.com/3scale/porta/blob/493f7a87e551c06486956c86ee2240f638912462/app/lib/logic/rolling_updates.rb#L237-L245
> ?
> Do you think it's worth deleting it also?

Absolutely. Done

app/models/access_token.rb
```diff
@@ -38,7 +38,7 @@ def public?
     private
 
     def non_public_scopes
-      %w(cms).freeze
+      %w[].freeze
```
thalesmiguel, line 41: I think we should remove this method as it's just an empty Array now.

jlledom: I'd prefer to leave it empty, so we keep the structure to add future items to it if needed. If we remove it, then the `public?` method above would make no sense neither and should be removed too, also the `visible_for` method should be refactored. Is it worth it?

akostadinov: if we don't have any private APIs, then why have these?

thalesmiguel: > Are we sure we won't have any in the future?

I'm pretty sure no one is really sure 😆
If we're gonna leave this structure here, I'd ask for a comment in this method saying something like "This method exists because we need the flexibility to have private APIs. As we have none now, it returns an empty string".

akostadinov: Why should we have private APIs?

jlledom: I removed the support for the private APIs

app/models/access_token.rb
```diff
@@ -59,11 +49,7 @@ def allowed_for(owner)
 
     def visible_for(account)
       select_and_build do |scope|
-        if scope.public?
-          true
-        else
-          account.provider_can_use?("#{scope.value}_api")
-        end
+        account.provider_can_use?("#{scope.value}_api")
```
thalesmiguel, line 52: The `public?` method would only return `false` if the scope was "cms". Only then we'd check for `account.provider_can_use?`.
By removing the "non_public_scopes", this `visible_for` method should be returning `true` all the time, right?
IF that's the case, maybe we could remove it as well.

jlledom: You're right, I misunderstood this method. Removed!

josemigallas:
Probably need to remove it from https://github.com/3scale/deploy as well.

thalesmiguel (approved):

jlledom:
@josemigallas I created this PR: https://github.com/3scale/deploy/pull/749

mayorova (approved):

PR #3138, opened 2022-12-20
Fix httpclient fork in managed-services

gsaslis:

* Updates the fork used for the `httpclient` ruby gem

same exact code on both forks - the 3scale one having the benefit that it doesn't depend on an external (nowadays) contributor.

* Fixes version of httpclient

the `require_relative` change is necessary for bundler to be able to correctly resolve the dependencies provided by Cachito

(cherry picked from commit f916a10ef03068a119f84f394730e098e24ba6f7)

I would like to promote this fix to `managed-services` stream, so that continuous product builds are functional again. (they'll be broken until this is ported over). 

Perhaps merging all commits from `master` to `managed-services` is preferred, but in case it isn't, we can use this PR in the meantime.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -95,7 +95,7 @@ gem 'ransack', '2.4.1' # we can remove line when stop using ruby 2.4
 gem 'browser', '~> 5.0.0' # we can update to lts when we stop using ruby 2.4
 gem 'diff-lcs', '~> 1.2'
 gem 'hiredis', '~> 0.6.3'
-gem 'httpclient', github: 'mikz/httpclient', branch: 'ssl-env-cert'
+gem 'httpclient', github: '3scale/httpclient', branch: 'ssl-env-cert'
 gem 'json-schema', git: 'https://github.com/3scale/json-schema.git'
 gem 'paperclip', '~> 6.0'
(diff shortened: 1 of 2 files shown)


akostadinov (changes requested):
we should rather push master to managed-services. We can do that any time. Oterwise we break fast-forward between the branches.

gsaslis:
@akostadinov up to you ! 

In this specific case, it's true that a merge would cause a merge conflict but it is a very very straightforward one to resolve, right? If we are ready to merge `master` to `managed-services`, then it would be great if we could do that, otherwise can we merge this in the meantime to unblock the cachito migration ?

jlledom:
I prefer to merge this into `master` and then FF master into `managed-servides`. We got a [github action](https://github.com/3scale/porta/actions/workflows/fast-forward-managed-services.yml) that performs the fast-forward automatically and it can't fix conflicts no matter how easy they are. After merging this to `master` I'll FF to `managed-services` manually myself.

gsaslis:
@jlledom it's already in `master` so fast-forwarding would be great ! ;)

gsaslis:
once you do that, please feel free to close this PR ;)

jlledom:
Done!

PR #3140, opened 2022-12-20
CMS: Sections: Make title a mandatory parameter

jlledom:

For sections, `title` is not required, but it's used to generate `system_name`, which is required. That makes `title` required in practice, since the UI doesn't provide a way to directly set the `system_name`. This PR adds the validation to make the `title` presence required.

**Verification steps** 

**UI:**

- Audience -> Developer Portal -> New section
- Don't fill any field
- Click "Create Section"
- A popup suggesting to fill out the title is shown
- Fill the title with an empty value (one or more spaces)
- Click "Create Section"
- An error message about the lack of title should be returned

**API:**

```
curl --request POST \
  --url 'http://provider-admin.3scale.localhost:3000/admin/api/cms/sections?access_token=secret' \
  --data '{                                                                                                                                                                                                                    
        "system_name": "test"
}'

```

Should return a response including:

`<error>Title can't be blank</error>`


thalesmiguel (commented):
Don't know if this would go out of the PR's scope (and how difficult it'd be to add) but maybe a DB index should be put in place so we make sure not to save faulty objects. Not saying the effort will be worth it 😬

akostadinov (commented):
tbh I'm not sure we should fix UI issues on the backend. Maybe add some js to prevent form without `title` to be submitted?

wrt title, if it is important, we can just add it to validation to cover the API use case. But if there are valid use cases where `title` is blank, then keep as is. Also we need to make sure there are no existing sections with blank `title`. Otherwise existing entries will be rendered invalid after the change.

btw the nice long explanation message IMO better goes to commit message.

app/models/cms/section.rb
```diff
@@ -201,6 +202,10 @@ def not_own_child
     end
   end
 
+  def system_name_presence
+    errors.add(:system_name) if system_name.blank? && title.present?
```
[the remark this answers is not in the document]
jlledom: right!

jlledom:
> Don't know if this would go out of the PR's scope (and how difficult it'd be to add) but maybe a DB index should be put in place so we make sure not to save faulty objects. Not saying the effort will be worth it grimacing

I'd prefer to avoid DB migrations 😱

jlledom:
> tbh I'm not sure we should fix UI issues on the backend. Maybe add some js to prevent form without `title` to be submitted?
> 

Haven't seen it that way, and makes sense. But I think `title` should really be required and validated, it's not logical to compute the system name from the title and then make the system name mandatory and not the title. Besides, this is not only being used from the UI, also from the API.

> wrt title, if it is important, we can just add it to validation to cover the API use case. But if there are valid use cases where `title` is blank, then keep as is. Also we need to make sure there are no existing sections with blank `title`. Otherwise existing entries will be rendered invalid after the change.
> 

After [this line](https://github.com/3scale/porta/blame/master/app/models/cms/section.rb#L29) was added it's not possible to create a section without a title, and that line has been there for a long time. Also I queried the DB and found only one section in SaaS with an empty title. Since it's just one case that can be fixed easily, I would enforce the title requirement.

> btw the nice long explanation message IMO better goes to commit message.

If we merge this I'll add it to the squash commit

akostadinov:
> After this line was added it's not possible to create a section without a title, and that line has been there for a long time. 

Not really. One could use API to set `system_name` but not title.

> I queried the DB and found only one section in SaaS with an empty title. Since it's just one case that can be fixed easily, I would enforce the title requirement.

How do things behave when `title` is blank? Because we may need to create a migration rake task if we add that. So that users can update their DBs before/after upgrade. We don't have access to that.

jlledom:
> Not really. One could use API to set `system_name` but not title.
> 

If you send `system_name` through an API request the param is filtered and ignored. Only keys in [this](https://github.com/3scale/porta/blob/master/app/controllers/admin/api/cms/sections_controller.rb#L5) array are considered.

> How do things behave when `title` is blank? Because we may need to create a migration rake task if we add that. So that users can update their DBs before/after upgrade. We don't have access to that.

It appears like this, and you can't open it because the link is in the title.

image

thalesmiguel (dismissed):

jlledom:
I added a rake task to fill empty section titles. It's based on this solution:

https://github.com/ankane/strong_migrations#backfilling-data

test/unit/tasks/cms/cms_test.rb
```diff
@@ -2,24 +2,82 @@
...
+      def test_fix_section_empty_titles_system_name_empty
+        section = FactoryBot.build(:cms_section, partial_path: '/', parent: @provider.sections.root, title: nil, system_name: '')
+        section.save!(validate: false)
+        expected_title = "Section #{section.id}"
+
+        execute_rake_task 'cms/cms.rake', 'cms:fix:section_empty_titles'
+
+        section.reload
+        assert_equal expected_title, section.title
+      end
```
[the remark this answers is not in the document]
jlledom: Done

test/unit/tasks/cms/cms_test.rb
```diff
@@ -2,24 +2,82 @@
...
+      def test_fix_section_empty_titles_valid
+        section = FactoryBot.create(:cms_section, partial_path: '/', parent: @provider.sections.root)
 
-      @page.reload
+        execute_rake_task 'cms/cms.rake', 'cms:fix:section_empty_titles'
 
-      assert_equal true,  @page.published.include?('{{ part.url }}')
-      assert_equal false, @page.published.include?('{{ part.link }}')
+        section.reload
+        assert_not_equal section.system_name, section.title
```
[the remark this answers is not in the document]
jlledom: Done

app/models/cms/section.rb
```diff
@@ -201,6 +202,10 @@ def not_own_child
     end
   end
 
+  def system_name_presence
+    errors.add(:system_name, :blank) if system_name.blank? && title.present?
+  end
```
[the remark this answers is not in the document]
akostadinov: What I mean is not to allow submitting a form with an empty `title` instead of later reporting the error. It doesn't make sense to me to design the object around the UI. For the AR object, we want to validate `system_name` is not null. But instead we write the validation around a UI deficiency.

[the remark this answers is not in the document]
jlledom: OK, I left the system_name validation as it was, and added a html attribute to make the title required in the UI. Unfortunately, the `required` attribute doesn't prevent sending a title filled with only spaces, but it seems that's how it's done in all the project. @josemigallas, @lvillen is that right?

lib/tasks/cms/cms.rake
```diff
@@ -210,6 +210,17 @@ namespace :cms do
...
     end
+
+    desc 'Fixes empty titles in sections'
+    task :section_empty_titles => :environment do
+      CMS::Section.unscoped.where(title: ['', nil]).in_batches do |sections|
+        sections.each do |section|
+          title = section.system_name.presence || "Section #{section.id}"
+          section.update_column(:title, title)
+        end
+        sleep(0.01) # throttle
```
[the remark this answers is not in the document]
jlledom: IMO it's a bad idea to hardcode SQL queries. Using rails we can write the same code for alls DB engines.

About the throttle, AIUI it's necessary to allow the DB to perform other operations during the backfilling, so I left it.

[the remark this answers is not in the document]
jlledom: I wast thinking: what if, by a unlucky coincidence, the new `title` doesn't meet the uniqueness constraint? That would raise an exception and abort the script. It would be better to always set a UUID for title, so it reduces the chances of `title` being in use.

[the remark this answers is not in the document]
akostadinov: This likely is a part of a transaction. When introducing a delay, that will just keep the transaction locks for longer, thus reducing the parallel operations the database can handle as well occupy database resources for longer.

Even if for a particular RDBMS it is of any advantage, I would classify it as a premature optimization unless you observed positive effects on the preview database.

I think the `update_all` would be the least resource intensive approach for both - database and client.

See https://stackoverflow.com/a/73038493/520567 , we can have something like

```
CMS::Section.unscoped.where(title: ['', nil]).update_all("title = system_name")
CMS::Section.unscoped.where(title: ['', nil]).update_all("title = id")
```

Although given how few elements we update, it is fine to leave current implementation and just remove the artificial delay.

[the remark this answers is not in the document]
josemigallas: Is it possible to create a section without giving it a title? Is this task useful _right now_?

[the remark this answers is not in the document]
jlledom: Since I'm not opening a transaction explicitly, I assumed that would run out of a transaction. Is a transaction created automatically? Anyway I committed some changes.

lib/tasks/cms/cms.rake
```diff
@@ -210,6 +210,11 @@ namespace :cms do
         puts "\t++ Fixing pagination href in #{page.id}/shared/pagination"
       end
     end
+
+    desc 'Fixes empty titles in sections'
+    task :section_empty_titles => :environment do
+      CMS::Section.unscoped.where(title: ['', nil]).update_all("title = id")
```
mayorova, line 216: Why did you chose the `id` for filling the missing value? I think the `system_name` would be a more logical choice? Although, maybe I am missing some scenario where it might not work...

mayorova:
I personally would keep `system_name` as an optional parameter on the Section Create endpoint (this is how other APIs in System tend to work), and for consistency would add a field on the UI form.

So, if `system_name` is provided, it would be used, and otherwise it would be calculated from the `title`. I agree that Title should be a required field.

On the UI the Update form would contain System Name as a read-only field. Again, this is pretty much how almost all other forms work in the admin portal.

The UI changes could obviously go to a different PR. I am just saying that I wouldn't remove `system_name` from the Create endpoint.

doc/active_docs/CMS API.json
```diff
@@ -262,11 +262,6 @@
               "paramType": "query",
               "allowMultiple": false
             },
-            {
-              "name": "system_name",
-              "description": "Human readable and unique identifier",
-              "paramType": "query"
-            },
             {
               "name": "title",
```
josemigallas, line 266: Do we need this as well?
```suggestion
              "name": "title",
              "required": true,
```

test/unit/tasks/cms/cms_test.rb
```diff
@@ -2,24 +2,56 @@
 
 module Tasks
   class CmsTest < ActiveSupport::TestCase
+    class CmsTestFixPaginationHref < CmsTest
+      def setup
+        @page = FactoryBot.build(:cms_partial, system_name: 'shared/pagination',
+                                 published: '<p>Title</p><a href="{{ part.link }}">Supertramp</a>')
```
josemigallas, line 8: This indentation hurts my eyes 😄. However, I just learnt it is the standard for ruby actually.

test/unit/tasks/cms/cms_test.rb
```diff
@@ -2,24 +2,56 @@
...
-    def setup
-      @page = FactoryBot.build(:cms_partial, system_name: 'shared/pagination',
-        published: '<p>Title</p><a href="{{ part.link }}">Supertramp</a>')
+        @page.save(validate: false)
+      end
 
-      @page.save(validate: false)
+      def test_fix_pagination_href
+        assert_equal true,  @page.published.include?('{{ part.link }}')
+        assert_equal false, @page.published.include?('{{ part.url }}')
```
josemigallas, line 15:
```suggestion
        assert @page.published.include?('{{ part.link }}')
        assert_not @page.published.include?('{{ part.url }}')
```
Or maybe this works?
```suggestion
        assert_includes @page.published, '{{ part.link }}'
        assert_not_includes @page.published, '{{ part.url }}'
```

test/unit/tasks/cms/cms_test.rb
```diff
@@ -2,24 +2,56 @@
...
+
+        execute_rake_task 'cms/cms.rake', 'cms:fix:pagination_href'
+
+        @page.reload
+
+        assert_equal true,  @page.published.include?('{{ part.url }}')
+        assert_equal false, @page.published.include?('{{ part.link }}')
```
josemigallas, line 22:
```suggestion
        assert @page.published.include?('{{ part.url }}')
        assert_not @page.published.include?('{{ part.link }}')
```

```suggestion
        assert_includes @page.published '{{ part.url }}'
        assert_not_includes @page.published, '{{ part.link }}'
```

test/unit/tasks/cms/cms_test.rb
```diff
@@ -2,24 +2,56 @@
...
-    def test_fix_pagination_href
-      assert_equal true,  @page.published.include?('{{ part.link }}')
-      assert_equal false, @page.published.include?('{{ part.url }}')
+    class CmsTestFixSectionEmptyTitles < CmsTest
+      def setup
+        @provider = FactoryBot.create(:provider_account)
+      end
+
+      [nil, ""].each do |empty_val|
+        define_method("test_fix_section_empty_titles_title_#{empty_val.nil? ? 'nil' : 'empty'}") do
```
josemigallas, line 32: I think this approach makes it harder to read, and therefore to maintain.

test/unit/tasks/cms/cms_test.rb
```diff
@@ -2,24 +2,56 @@
...
+          section = FactoryBot.build(:cms_section, partial_path: '/', parent: @provider.sections.root, title: empty_val, system_name: 'system-name')
+          section.save!(validate: false)
+          expected_title = section.id.to_s
+
+          execute_rake_task 'cms/cms.rake', 'cms:fix:section_empty_titles'
+
+          section.reload
+          assert_equal expected_title, section.title
```
josemigallas, line 40: Maybe it would be better to assert title is _not_ empty? De we care what the title actually is?
```suggestion
          assert_not_empty expected_title
```

test/unit/tasks/cms/cms_test.rb
```diff
@@ -2,24 +2,56 @@
...
+        end
+      end
 
-      execute_rake_task 'cms/cms.rake', 'cms:fix:pagination_href'
+      def test_fix_section_empty_titles_valid
```
josemigallas, line 44: It is not clear from then title what this test is actually testing. I _guess_ it's that the rake task do not change the titles that are valid. If so, I would name it just like so `section_empty_titles do not change existing titles`.

test/unit/tasks/cms/cms_test.rb
mayorova: Well, the rake task sets the title to something specific. I'd rather we check that it's actually set to the value we intended to, and not something potentially random.

lib/tasks/cms/cms.rake
jlledom: I'm choosing the id for two reasons:

1. I'm not 100% sure that `system_name` is not empty or `nil` in all cases, there could be corner cases
2. The title has some uniqueness restrictions, one good way to ensure `title` is unique is setting it to the `id`, which must be unique.

I don't think it's setting the `id` as title is a big deal. When a section has no title, there's no way the user can click on it to edit/delete it. That's probably the only reason why there are sections with empty titles: they were created somehow, and then they can't be removed. See the attached image:

[image]

For theses cases, having anything to click on will be helpful enough, they can change the title later or remove the section.

doc/active_docs/CMS API.json
jlledom: You are right

test/unit/tasks/cms/cms_test.rb
jlledom: Your eyes are wrong

lib/tasks/cms/cms.rake
mayorova: Well, there are currently to be the following validations:
```
validates :system_name, :provider, presence: true
validates :system_name, uniqueness: { :scope => [:provider_id] }, length: { maximum: 255 }
```

So I guess the system name is something that is expected to be present always (unlike the title for which the presence validation was missing).

Using system name value to fill the title is just more consistent with how title/name and system name are related for most of the models in Porta. Also, some suggestions here: https://github.com/3scale/porta/pull/3140#issuecomment-1400504580

test/unit/tasks/cms/cms_test.rb
jlledom: Done

test/unit/tasks/cms/cms_test.rb
jlledom: Done

test/unit/tasks/cms/cms_test.rb
jlledom: Readability vs. Not repeating code. I wrote separated tests at first and then merged them suggested by @akostadinov. I think the current state is better but I don't have a strong opinion, so if you convince Aleks, I'm fine.

test/unit/tasks/cms/cms_test.rb
jlledom: Agreed with @mayorova

test/unit/tasks/cms/cms_test.rb
jlledom: I renamed it to `test_fix_section_empty_titles_valid_dont_change`. Hope it's better.

lib/tasks/cms/cms.rake
jlledom: OK, I checked all the git history from the beginning in the system repo and it seems the `system_name` presence validation has been there the whole time. So it should be impossible to have any section in the DB without a `system_name`. I also queried `CMS::Sections.where(system_name: [nil, ''])` in SaaS to verify, and it returned 0 records. It updated the task.

jlledom:
> I personally would keep system_name as an optional parameter on the Section

I implemented this and added a few tests. @mayorova

test/unit/tasks/cms/cms_test.rb
akostadinov: And while nitpicking better use the `test "some behavior happens"` IMO than `def test_some_behavior_happens` :)

test/unit/tasks/cms/cms_test.rb
akostadinov: Readability can be improved by using `test "..."` syntax instead of `define_method`. Or by defining a test and then just loop through the two values inside it. 

```rb
[nil, ""].each do |empty_val| do |value|
  test "section titles are fixed for #{value.inspect}" do
    # actual test code
  end
end
```

```rb
  test "section empty titles are fixed" do
    [nil, ""].each do |empty_val|
      # actual test code
    end
  end
```

mayorova: I'd vote for the 2nd one @akostadinov suggested.

jlledom: I pushed the first option, because the second would create only one test and run the rake task twice without a DB rollback between them

test/unit/tasks/cms/cms_test.rb
jlledom: I also prefer the `test "description"` over `def test_description`. I was using the second here because it's the one being used before in this file, for consistency. But I pushed it anyway.

lib/tasks/cms/cms.rake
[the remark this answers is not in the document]
akostadinov: Yes, afaik everything runs by default in a transaction in porta.

lib/tasks/cms/cms.rake
```diff
@@ -210,6 +210,11 @@ namespace :cms do
         puts "\t++ Fixing pagination href in #{page.id}/shared/pagination"
       end
     end
+
+    desc 'Fixes empty titles in sections'
+    task :section_empty_titles => :environment do
+      CMS::Section.unscoped.where(title: ['', nil]).update_all("title = system_name")
```
akostadinov, line 216: Just FYI this probably needs to be added in release notes or upgrade notes so that customers can fix their databases.

test/integration/cms/api/sections_test.rb
```diff
@@ -109,5 +109,56 @@ def create_section
...
+
+        doc = Nokogiri::XML::Document.parse(@response.body)
+        error = doc.xpath('//error').text
+
+        assert_match "Title can't be blank", error
+      end
+
+      test 'create with title but without system_name' do
+        expected_title = 'New Section'
+        expected_sysname = 'New Section'
```
akostadinov, line 132: This doesn't seem like a sensible system name. Is this correct behavior? Most names that I see in actual DB are more like `new-section`

test/unit/tasks/cms/cms_test.rb
akostadinov: I think the `::test` method creates a method under the hood, but it reads better anyway.

akostadinov (approved):
PR looks good on my end. Two small comments I leave on your discretion.

Other participants need to approve as well though.

wrt taking system name from user, provided it is only an immutable identifier, and title is what user works with, then it is somehow confusing to let user specify both. So I'm on the side to generate something with `#parameterize` and not accept parameter from user (as originally PR was intended). But this is not a hard opinion.

If definitive conclusion is hard to reach immediately, maybe makes sense to leave things before, merge the fixes and discuss addin/removing `system_name` in a separate PR.

test/integration/cms/api/sections_test.rb
jlledom: Yes, I have a PR for that: https://github.com/3scale/porta/pull/3163

I'll have to fix this test when both are merged to master

test/integration/cms/api/sections_test.rb
```diff
@@ -109,5 +109,56 @@ def create_section
...
     end
 
+    class SystemNameTest < ActionDispatch::IntegrationTest
+      def setup
+        @provider = FactoryBot.create(:provider_account)
+        host! @provider.external_admin_domain
+      end
+
+      test 'create without title' do
+
```
josemigallas, line 119:
```suggestion
```

app/models/cms/section.rb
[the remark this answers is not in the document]
josemigallas: TL;DR: Yes, that's right.

`require` prevents the field to be submitted _empty_ of characters, and a blank space is still a character. IMHO this should be up to the controller to validate. If we want some form of client validation, I suggest a hint is added to the field, something like "use a valid name" (Whatever that means).
Optionally, we can specify a `pattern` that the input's value must match. However, this is not an ideal UX without a proper hint / warning, it will only say "The value must match the pattern" or something like that.
It is also possible to set a `min` and `max` length to the input, but I don't think this helps.

The last resort is of course JS. Everything can be done with JS 😄.

josemigallas (approved):
Tentative approval

akostadinov:
What does `Tentative approval` mean? @josemigallas

app/models/cms/section.rb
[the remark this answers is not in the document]
jlledom: I think adding the required attribute is enough for this case. It will return a validation error from rails in the case of sending an empty space

test/integration/cms/api/sections_test.rb
jlledom: Is this a mistake? 🙃

jlledom:
In order to make things simpler for us and the clients, I removed the rake task and implemented the backfill as a migration, this way it'll be executed automatically.

This is from staging:

```
irb(main):001:0> CMS::Section.unscoped.where(title: ['', nil]).size
=> 1
```

That's because even when `title` is not mandatory for sections, the title is required to generate `system_name`, which is mandatory, so `title` is mandatory for practical purposes, and the migration should have little or no effect in SaaS or most clients.

thalesmiguel (approved):
🚀

thalesmiguel (dismissed):
🚀[2]

akostadinov (changes requested):
You seem to introduce no change of behavior except adding a validation for `title`. Am I correct?

Besides the few comments, perhaps good to update PR description that looks a little outdated so that one can better understand reading it in the future after the PR was fixed.

app/controllers/admin/api/cms/sections_controller.rb
```diff
@@ -71,6 +72,7 @@ def show
   ##~ op.parameters.add @parameter_access_token
   ##~ op.parameters.add @parameter_section_id
   ##~ op.parameters.add :name => "title", :description => "Title of the section", :paramType => "query"
+  ##~ op.parameters.add :name => "system_name", :description => "Human readable and unique identifier", :paramType => "query"
```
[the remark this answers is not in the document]
mayorova: I think it is to contrast with the ID (numeric one) as a unique identifier.
System name is also unique, but unlike the numeric ID it is (or supposed to be) also meaningful.

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2022_02_28_215614) do
+ActiveRecord::Schema.define(version: 2023_03_08_155529) do
```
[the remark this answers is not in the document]
jlledom: Done

test/unit/tasks/cms/cms_test.rb
```diff
@@ -5,7 +5,7 @@ class CmsTest < ActiveSupport::TestCase
 
     def setup
       @page = FactoryBot.build(:cms_partial, system_name: 'shared/pagination',
-        published: '<p>Title</p><a href="{{ part.link }}">Supertramp</a>')
+                               published: '<p>Title</p><a href="{{ part.link }}">Supertramp</a>')
```
akostadinov, line 8: maybe revert this indentation only change?

akostadinov (approved):
Please make sure to test it on a copy of prod db before merging.

test/unit/tasks/cms/cms_test.rb
jlledom: It's not directly related to the PR but it's a Rubocop fix, I think it's OK.

thalesmiguel (dismissed):

thalesmiguel (approved):

PR #3141, opened 2022-12-21
CMS API: Remove wrong params for DELETE file

jlledom:

The OpenAPI v1 spec for the CMS API contains some wrong parameters for the DELETE File endpoint

This is part of: [THREESCALE-8991](https://issues.redhat.com/browse/THREESCALE-8991)


thalesmiguel (commented):
Should we also remove the source2swagger magic comments in https://github.com/3scale/porta/blob/master/app/controllers/admin/api/cms/files_controller.rb#L105-L109 for consistency?

jlledom:
> Should we also remove the source2swagger magic comments in https://github.com/3scale/porta/blob/master/app/controllers/admin/api/cms/files_controller.rb#L105-L109 for consistency?

You're right! Done!

akostadinov (approved):

PR #3143, opened 2022-12-21
Fix licenses report mambo jambo

josemigallas:


lib/tasks/licenses.rake
```diff
@@ -4,7 +4,7 @@ namespace :licenses do  # rubocop:disable Metrics/BlockLength
   report_path = Rails.root.join('doc/licenses/licenses.xml')
 
   desc 'Generates a report with the dependencies and their licenses'
-  task :report do
+  task report: :environment do
```
akostadinov, line 7: why?

lib/tasks/licenses.rake
```diff
@@ -39,9 +39,9 @@ namespace :licenses do  # rubocop:disable Metrics/BlockLength
 
   def cli
     @cli ||= begin
-               require 'license_finder'
-               LicenseFinder::CLI::Main.new
-             end
+      require 'license_finder'
+      LicenseFinder::CLI::Main.new
+    end
```
akostadinov, line 44: My automatic indentation made it like before. I like more what you have here. But we should enforce single style somehow.

lib/tasks/licenses.rake
josemigallas: Rubocop changed it:
```
Offenses:

lib/tasks/licenses.rake:7:3: C: [Correctable] Rails/RakeEnvironment: Include :environment task as a dependency for all Rake tasks.
  task :report do
  ^^^^^^^^^^^^
```

lib/tasks/licenses.rake
josemigallas: Rubocop again, changed it:
```
lib/tasks/licenses.rake:44:14: W: [Correctable] Layout/BeginEndAlignment: end at 44, 13 is not aligned with @cli ||= begin at 41, 4.
             end
             ^^^
```

thalesmiguel (approved):

lib/tasks/licenses.rake
akostadinov: Disable that check. It is wrong. See https://github.com/3scale/porta/blob/master/Rakefile#L53-L60

lib/tasks/licenses.rake
akostadinov: Ok, I'll check if it is something local for me. I'm fine with the change.

lib/tasks/licenses.rake
akostadinov: For these tasks also we absolutely don't need `environment`. It is very very strange for me that it tries to enforce a dependency that developer can see whether it is needed or not. It is a bad practice IMO to try loading everything for everything. What is the point of a dependency system then?

josemigallas: OK

akostadinov: Moreover rubocop should be bad practices check, while some people maybe think it has to be a reminder. I dunno why I'm so angry at this check.

josemigallas:
@akostadinov According to Circleci, the report is outdated. Any idea why? Here's the diff after running `rake licenses:report` via ssh:
```
git diff doc/licenses/licenses.xml
diff --git a/doc/licenses/licenses.xml b/doc/licenses/licenses.xml
index 183a3ef..546178f 100644
--- a/doc/licenses/licenses.xml
+++ b/doc/licenses/licenses.xml
@@ -5676,8 +5676,8 @@
         <version>1.2.13</version>
         <licenses>
                       <license>
-              <name>MIT</name>
-              <url>http://opensource.org/licenses/mit-license</url>
+              <name>unknown</name>
+              <url></url>
             </license>
                   </licenses>
       </dependency>
@@ -5686,8 +5686,8 @@
         <version>2.3.2</version>
         <licenses>
                       <license>
-              <name>MIT</name>
-              <url>http://opensource.org/licenses/mit-license</url>
+              <name>unknown</name>
+              <url></url>
             </license>
                   </licenses>
       </dependency>
```

lib/tasks/licenses.rake
```diff
@@ -37,14 +55,8 @@ namespace :licenses do  # rubocop:disable Metrics/BlockLength
 
   private
 
-  def cli
-    @cli ||= begin
-               require 'license_finder'
-               LicenseFinder::CLI::Main.new
-             end
-  end
```
jlledom, line 45: Why remove this and create a new instance of `LicenseFinder::CLI::Main.new` every time?

josemigallas: It won't write the temporary report if I don't create a new instance. Anyway I'm shooting in the dark, trying to make it work, expect weird stuff 💩

jlledom: Is it not reviewable yet?

josemigallas: Very reviewable

Gemfile
```diff
@@ -238,7 +238,7 @@ group :development, :test do
   gem 'bootsnap', '~> 1.4'
   gem 'colorize'
   gem 'factory_bot_rails', '~> 6.2'
-  gem 'license_finder', '~> 7.0.1'
+  gem 'license_finder', '~> 7.1.0'
```
akostadinov, line 241: :muscle:

.rubocop.yml
```diff
@@ -253,3 +253,6 @@ Lint/Debugger:
 
 Naming/RescuedExceptionsVariableName:
   PreferredName: exception
+
+Rails/RakeEnvironment:
+  Enabled: false
```
akostadinov, line 258: why?

lib/tasks/licenses.rake
```diff
@@ -14,19 +16,18 @@ namespace :licenses do  # rubocop:disable Metrics/BlockLength
...
   task :compliance do
-    warn 'Checking license action items...'
-    unless cli.action_items
-      warn '*** License compliance test failed  ***'
-      exit 1
-    end
+    require 'license_finder'
+
+    warn 'Checking action items...'
+    exit(1) unless LicenseFinder::CLI::Main.new.action_items
```
akostadinov, line 22: what is this doing?

lib/tasks/licenses.rake
```diff
@@ -14,19 +16,18 @@ namespace :licenses do  # rubocop:disable Metrics/BlockLength
...
 
     warn 'Checking license report is up to date...'
-    Tempfile.new(report_path.to_s) do |temp_file|
+    Tempfile.create do |temp_file|
```
akostadinov, line 25: shouldn't this be `Tempfile.open`?

josemigallas:
`fsevent` is a package exclusive for OSX, therefore when running license_finder on a Linux machine the licenses aren't found:
```
fsevents, 1.2.13, unknown
fsevents, 2.3.2, unknown
```

The solution is simply to approve these versions manually with `license_finder licenses add fsevents MIT`.

Thanks @thalesmiguel for point this out.

lib/tasks/licenses.rake
josemigallas: Action items is supposed to print a list of licenses that are to be approved. If we ever add a library with a weird license I guess this will warn us about it.

.rubocop.yml
josemigallas: Because https://github.com/3scale/porta/pull/3143#discussion_r1054682728

lib/tasks/licenses.rake
josemigallas: I hope not!

josemigallas: I guess I copy pasted it from somwhere, I can't say there's a big difference:
>create
>Creates a temporary file as usual File object [...]
>If a block is given, then a File object will be constructed, and the block is invoked with the object as the argument. >The File object will be automatically closed and the temporary file is removed after the block terminates. The call >returns the value of the block.

>open
>Creates a new Tempfile [...]
>If a block is given, then a [Tempfile](https://ruby-doc.org/stdlib-2.5.3/libdoc/tempfile/rdoc/Tempfile.html) object will >be constructed, and the block is run with said object as argument. The [Tempfile](https://ruby-doc.org/stdlib->2.5.3/libdoc/tempfile/rdoc/Tempfile.html) object will be automatically closed after the block terminates. The call >returns the value of the block.

lib/tasks/licenses.rake
akostadinov: But the shouldn't it be something like `exit(1) if LicenseFinder::CLI::Main.new.action_items.present?`.

Also Perhaps print action items out but this can be for another time.

lib/tasks/licenses.rake
akostadinov: hmm, with a block it seems `#create` is the recommended way. `#open` should also work though. Cool.

lib/tasks/licenses.rake
josemigallas: Unless I've got it wrong, which is highly possible, `#action_items` will print everything and return an IO object. Otherwise it will return nothing. I may be wrong, I am not sure how to reproduce it.

jlledom: To make it fail you just need to remove any approved license (e.g. MIT) from `dependency_decisions.yml`. I removed MIT, ran `rake licenses:compliance` and saw this error:

```
Dependencies that need approval:
3scale_client, 2.11.0, MIT
@babel/code-frame, 7.18.6, MIT
@babel/compat-data, 7.14.7, MIT
@babel/compat-data, 7.19.0, MIT
@babel/core, 7.14.6, MIT
@babel/generator, 7.19.0, MIT
...
```

`LicenseFinder::CLI::Main.new.action_items` exits the task when it finds a not-approved license, it returns nothing otherwise:

https://github.com/pivotal/LicenseFinder/blob/master/lib/license_finder/cli/main.rb#L133-L147

So, between `if ... .present?` and `unless`, the right answer is: it doesn't matter, you can just call it directly
```suggestion
    LicenseFinder::CLI::Main.new.action_items
```
It will `exit(1)` when not complying and continue to the up-to-date test otherwise.

lib/tasks/licenses.rake
```diff
@@ -1,10 +1,12 @@
 # frozen_string_literal: true
 
 namespace :licenses do  # rubocop:disable Metrics/BlockLength
-  report_path = Rails.root.join('doc/licenses/licenses.xml')
+  report_path = Rails.root.join('doc/licenses/licenses.xml').freeze
 
   desc 'Generates a report with the dependencies and their licenses'
   task :report do
+    require 'license_finder'
```
jlledom, line 8: You don't need the `require`, rails should make that job for you. I tried removing it and it works

lib/tasks/licenses.rake
thalesmiguel: Wait... So `LicenseFinder::CLI::Main.new.action_items` will `exit(0)` if there are no dependencies to be approved and `exit(1)` if there's any. I don't think checking the value it returns will work. 
It must be returning an object of `<IO:<STDOUT>>`, which should always be present 🤔 Like `STDOUT.present? #=> true`.
Also, as this `action_items` always exits, I'd expect this rake task to exit as well when `action_items` is called...
Maybe we should call this method with a rescue for `SystemExit`. Then we can check the exit code status with something like like:
```ruby
LicenseFinder::CLI::Main.new.action_items
rescue SystemExit => error
  error.status #=> should be 0 or 1
end
```

lib/tasks/licenses.rake
akostadinov: maybe in the past we had `require: false` in the Gemfile. I think it is good to have `require: false` in the Gemfile actually. But if we don't, then the `require 'license_finder'` is indeed not needed.

lib/tasks/licenses.rake
josemigallas: After further investigation I've found out that `LicenseFinder::CLI::Main.new.action_items` will abort the rake task if there are any action items. IMHO it doesn't make sense to check the returned `IO:STDOUT`, because if the task doesn't stop it means there aren't any pending actions items 😄 .

akostadinov (approved):

jlledom (approved):

thalesmiguel (approved):
🚀

PR #3147, opened 2023-01-03
allow easier caching of oracle gems

akostadinov:
Now if one wants to cache only oracle gems when building images so that gems are available in image but not installed so only free bits are available in the image, then can do

```
BUNDLE_WITHOUT: "assets:test:development:default:preview:production"
bundle package
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -257,9 +257,11 @@ gem 'developer_portal', path: 'lib/developer_portal'
 gem 'unicorn', require: false, group: %i[production preview]
 
 # NOTE: Use ENV['DB'] only to install oracle dependencies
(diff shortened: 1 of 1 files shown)


jlledom (approved):

gsaslis:
@akostadinov with all the changes we introduced and shutdown in the way, I admit I'm a little lost. 

How does this help us / what does it improve?

Is it that we no longer have to run `bundle package` for **all** the gems ? (meaning redundant files / image size) ?

nidhi-soni1104 (approved):
LGTM

akostadinov:
@gsaslis , exactly. To prevent cached gems that were also installed in the image.

gsaslis:
ok, thanks for confirming @akostadinov ! 

Once this is merged, could you open the respective MR against 3scale-amp-2 branch in the porta-midstream repository?

gsaslis (approved):

PR #3157, opened 2023-01-12
CircleCI: Remove 3scale-ninjabot key

jlledom:

There's a reference to the 3scale-ninjabot key in our CircleCI configuration. The key was compromised so we need to rotate it, but I think it's not actually being used and I'd like to remove it. I'm creating this PR to launch CircleCI jobs and check if something fails.


jlledom:
I surrender

PR #3175, opened 2023-01-24
CircleCI: Fix slack notification on failure

jlledom:

The `notify_failure` job in CircleCI is never called, because it's configured to run after a list of required previous jobs, using the command `requires`. The problem with `requires` is that all listed jobs must no only finish, but finish successfully in order to the dependent job to run. When one required job fails, the `requires` condition is not met and the `notify_failure` doesn't start.

The approach to solve this is to manually add a `notify_failure` step to every job and call it the moment any job fails, no need to wait for all other jobs to finish.

When merged, this **_should_** work, but we can't be sure until we test it in the real environment


akostadinov:
Hmm. One issue with this is that we may receive multiple notifications from same workflow.

akostadinov:
In docs I see they recommend using post steps. I don't know if there is any advantage doing that.

https://support.circleci.com/hc/en-us/articles/360047082992-How-to-send-a-slack-notification-at-end-of-workflow

akostadinov:
Another option seems to be using webhooks and slack integration. See https://circleci.com/blog/slack-integration/

This would be better because it will simplify circleci config, as well avoid multiple notifications. But it moves configuration to multiple places. Also I wonder if we can filter by branch in such case.

jlledom:
With this approach there will be multiple notifications if multiple jobs fail. Also, there will be no notification if the job fails during any other step than the test one, for instance, no notification when the test fails in a setup step like launching some container, etc. It's better what we have now, though, which is no notifications at all.

I'll take a look at webhooks to see if they fit our needs

akostadinov:
That is what I was going to say - it's fine to go with this for the time being if other approaches are more complicated. I'm just wondering whether post-steps make any difference with your approach.

I think though that we need notification if anything fails. We will still have to fix it, regardless.

jlledom:
Post-steps are steps to be run at the end of a job, not a workflow, AFAIK, so still multiple notifications for multiple failures. Post-steps might fix the notifications for when a job fails because of a configuration step, not sure.

akostadinov:
We can also make the report job wait others to complete like in https://discuss.circleci.com/t/workaround-run-jobs-sequentially-regardless-of-the-outcome-of-the-required-jobs/40807

But I'm fine with your more simple approach. I would just request that we add this notification for the other jobs as well - `lint`, `assets_precompile`, ` dependencies_bundler`.

akostadinov (changes requested):
I see a number of jobs without notify_failure. I think all should have. Or did I forget something?

jlledom:
> I see a number of jobs without notify_failure. I think all should have. Or did I forget something?

Many jobs reuse the same commands for different contexts (e.g. `rails-test`), so I added the `notify_failure` step to the command to avoid the redundancy of adding it to every job. The problem is some jobs don't use any reusable command, so for those I added the `notify_failure` step to the job definition itself. I understand that can be confusing, because you see some jobs calling `notify_failure` and some others don't. So I pushed some changes and now `notify_failure` is called only from jobs. Redundant but less confusing.

akostadinov (dismissed):
I see. Thank you!

PR #3178, opened 2023-01-25
online webpacker assets compilation fix

akostadinov:
`webpack-dev-server` and running `rails server` in dev mode
need more memory to compile assets. Now this will happen automatically
without the need to set `NODE_OPTIONS` on the command line.

More frustratingly, when running `rails server` the OOM condition is not reported, rahter compilation silently fails and the useless error message is
present that `dashboard` or whatever is missing from `manifest.json`. Which
is reported empty.

`rake assets:precompile` does **not** need this for whatever reason.


INSTALL.md
```diff
@@ -72,17 +72,12 @@ bundle exec rake ts:configure ts:start
 
 ## Asset compilation
 
-On development, assets are compiled on automatically everytime a page is requested to rails. To avoid the uptime we recommend running webpack-dev-server in a second terminal:
+In development, assets are compiled automatically when a page is requested from rails. To allow online updates when an asset file changes, start:
```
josemigallas, line 75: I don't think I get the "online" updates part. According to rails documentation the right term is _live code reloading_. So maybe:

```suggestion
In development, assets are compiled automatically on demand when a Rails page loads. If you want to use [live code reloading](https://guides.rubyonrails.org/webpacker.html#running-webpacker-in-development) you'll need to run:
```

config/initializers/webpacker.rb
```diff
@@ -0,0 +1,6 @@
+# Memory needed presently to compile js/ts is aroung 2.5GB
```
josemigallas, line 1: Thing about this is that `--max-old-space-size` is only a temporary workaround for an old version of node. Do we still want to make an initializer for it? Memory limit has been a very well-known problem with older nodejs but AFAIK it's kinda fixed in the present. At least according to [this article](https://medium.com/geekculture/node-js-default-memory-settings-3c0fe8a9ba1), the default limit in `>= 14.x` seems to be `4.0 GB`.
I haven't found any mention of it but in node 12 everybody is basically complaining about the hard-coded limit of 2GB so I would be surprised if more recent versions kept this limit (which is confirmed by the article not to be the case).

Copying table for reference:
```
Node.js Version   Limit  
----------------- -------
  18.x             4.0 GB
  17.x             4.0 GB
  16.x             4.0 GB
  15.x             4.0 GB  
  14.x             4.0 GB  
  13.x             2.0 GB  
  12.x             2.0 GB  
  11.x             1.4 GB  
  10.x             1.4 GB  
  9.x              1.4 GB
  ```

akostadinov: We can remove this once we upgrade. But also keeping the setting may help us identify big regressions in memory usage that we have.

I think that at least it is worth to make things work automatically without extra variables on command line, as it is very frustrating to see things fail without proper error.

josemigallas (approved):

PR #3191, opened 2023-02-01
🧪 🥒 Make some cucumbers more robust

josemigallas:
By removing an unused step definition, I noticed 2 scenarios failed but I could not figure out why. It turns out these particular scenarios rely on 2 buyers created for 2 providers, but the way our steps find those buyers is by `Account.find_by_org_name` so for any X buyers with the same name, only one of them will be found, regardless of the provider they belong too.

This is precisely what happens in commit 30b0c970ad813f9a81e77a7a9f74456d91a478fd, the 2nd buyer is supposed to be approved but it isn't. Instead, the first `buyer_one` is approved twice, making the sign in fail.

The second commit makes sure all buyers are fetched from the provider they belong to.

### DIFF
--- a/features/step_definitions/developer_portal/admin/account/payment_details_steps.rb
+++ b/features/step_definitions/developer_portal/admin/account/payment_details_steps.rb
@@ -22,13 +22,13 @@ When "the buyer is reviewing their account settings" do
 end
 
 When "the buyer is reviewing their credit card details" do
-  stub_stripe_intent_setup if @buyer.provider_account.payment_gateway_type == :stripe # Only stripe need to be stubbed
+  stub_stripe_intent_setup if @buyer.reload.provider_account.payment_gateway_type == :stripe # Only stripe need to be stubbed
   visit admin_account_path
   credit_card_details_tab.click
 end
 
 When "the buyer enters the generic credit card details URL manually" do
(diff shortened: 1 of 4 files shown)


features/step_definitions/developer_portal/admin/account/payment_details_steps.rb
```diff
@@ -22,7 +22,7 @@
 end
 
 When "the buyer is reviewing their credit card details" do
-  stub_stripe_intent_setup if @buyer.provider_account.payment_gateway_type == :stripe # Only stripe need to be stubbed
+  stub_stripe_intent_setup if @buyer.reload.provider_account.payment_gateway_type == :stripe # Only stripe need to be stubbed
```
josemigallas, line 25: I am not convinced about this, at all. Problem is the order streps are executed. `@provider` and `@buyer` are first created, then `@provider.payment_gateway_type` is changed and finally `@buyer.provider_account.payment_gateway_type` is read, but it still has the old value 😕.

akostadinov: `@buyer.provider_account` and `@provider` might be different Ruby objects depending on how they were created and then operations on them like save, reload, etc. `reload` shouldn't hurt if it makes things working, it is fine with me.

akostadinov (approved):

lvillen (approved):

PR #3193, opened 2023-02-03
CMS API: Implement API Contract

jlledom:

This is the implementation of the API contract we agreed:

https://docs.google.com/document/d/14Lpu5T-ktDVUNl9SINtd7WRdew1x0f5qDzFngThZyV8/edit

[THREESCALE-8992](https://issues.redhat.com/browse/THREESCALE-8992)

**Verification steps** 

Call the endpoints to check they match the format in the document

Finally there will be breaking changes. We decided to not return fields that make no sense for the type being returned, that breaks responses for all builtin types, which were returning non-sensic info. This is the complete list of changes:

#### Sections

- Two tags for sections: `<section>` and `<builtin_section> `. It was always `<section>` before and the client couldn't know which type of section was being returned **[BREAKING]**
- Long responses now return a list of contents in the format: `<type><id/></type>`

#### Files

- Removed: `<tag_list>` **[BREAKING]**
- Added: `<downloadable>`, `<content_type>`

#### Templates

- **All**: Return draft and published contents inside a CDATA wrapper
- **Page**: Replace `<layout>` (containing the name) by `<layout_id>` **[BREAKING]**
- **Builtin page**:
  - Removed: `<system_name>`, `<liquid_enabled>` **[BREAKING]**
  - Added: `<title>`, `<layout_id>` 
- **Layout**:
  - Removed: `<content_type>`, `<handler>` **[BREAKING]**
- **Builtin Partial & Partial**:
  - Removed: `<content_type>`, `<handler>`, `<liquid_enabled>` **[BREAKING]**


josemigallas (approved):
Codewise LGTM, haven't checked the actual responses.

app/controllers/admin/api/cms/sections_controller.rb
```diff
@@ -101,9 +102,19 @@ def destroy
...
-    end
+  def find_section(contents: false)
+    q = current_account.sections
+    q = q.includes(:pages, :files, :builtins, :children) if contents
+    q = q.where(id: params[:id]).or(q.where(system_name: params[:id]))
+    q.load
+
+    raise ActiveRecord::RecordNotFound, "Couldn't find CMS::Section with id or system_name=#{params[:id]}" if q.empty?
+
+    @section = q.first
```
thalesmiguel, line 114: Is `system_name` uniq? I ask because I'm trying to understand if there's a possibility for this query to return multiple rows.
**IF** that's the case and since we only care about the first row, we should add a `limit` to the query. Or even use `find_by` with `OR`... Something like:
```ruby
@section = current_account.sections.tap do |query|
  query.includes(:pages, :files, :builtins, :children) if contents
  query.find_by('id = ? OR system_name = ?', params[:id], params[:id]) # Ouch double params[:id]
end.load
```

thalesmiguel (approved):
Just one comment there. Ready to be ignored.

app/controllers/admin/api/cms/sections_controller.rb
```diff
@@ -101,9 +102,19 @@ def destroy
...
 
   private
-    def find_section
-      @section = current_account.sections.find_by_id(params[:id]) || current_account.sections.find_by_system_name(params[:id])
 
-      raise ActiveRecord::RecordNotFound.new("Couldn't find CMS::Section with id or system_name=#{params[:id]}") if @section.nil?
-    end
+  def find_section(contents: false)
+    q = current_account.sections
+    q = q.includes(:pages, :files, :builtins, :children) if contents
```
akostadinov, line 108: When we discussed we agreed to include IDs as content. But to get these IDs, one has to perform 4 extra queries per request. Is it worth it when one can get pages, files, builtins and children specifying parent? Returning IDs doesn't seem to save on API calls or db queries :(

jlledom: I think it's worth it because:

1. This feature and the search filter are complementary and both are going to be implemented eventually. So if in the future client we decide that the best way to get the whole contents tree is using the search filter and not this, that's ok.
2. The UI shows the contents when opening a particular section, so we should provide the same feature for the API.
3. If we remove this, there won't be a way to get the contents for one section in one single request, we'll need to request all pages for the lists of templates, sections and files filtered by parent_id in order to build the list of childs of one section, which is in the best of cases 3 requests and much more DB queries, since every request queries for the account, permissions, etc.
4. I don't expect a high traffic for this API anyway.

app/models/cms/builtin.rb
```diff
@@ -135,20 +135,17 @@ class Page < CMS::Builtin
     def to_xml(options = {})
       xml = options[:builder] || Nokogiri::XML::Builder.new
 
-      xml.builtin_page do |x|
+      xml.__send__(CMS::XML::CLASS_TAG.call(self.class)) do |x|
```
akostadinov, line 138: I think you can do this without a lookup table like this:
```
self.class.name[5..-1].parameterize(separator: "_")
```

app/models/cms/builtin.rb
```diff
@@ -135,20 +135,17 @@ class Page < CMS::Builtin
...
           xml.created_at created_at.xmlschema
           xml.updated_at updated_at.xmlschema
         end
-
-        # x.title title
-        x.system_name system_name
-        x.liquid_enabled liquid_enabled
-
+        x.title title
+        x.layout_id layout_id
```
akostadinov, line 145: did we discuss this change in API, I'm sorry if I forgot

app/models/cms/builtin.rb
```diff
@@ -135,20 +135,17 @@ class Page < CMS::Builtin
...
         unless options[:short]
-          x.draft draft
-          x.published published
+          x.draft { |node| node.cdata draft }
+          x.published { |node| node.cdata published }
```
akostadinov, line 148: has to be tested if this works with CDATA including HTML content

app/models/cms/layout.rb
```diff
@@ -21,22 +21,18 @@ def content_type
...
       unless new_record?
         xml.id id
         xml.created_at created_at.xmlschema
         xml.updated_at updated_at.xmlschema
       end
-
+      x.title title
       x.system_name system_name
-      x.content_type content_type
-      x.handler handler
```
akostadinov, line 33: did we discuss these two also?

app/models/cms/page.rb
```diff
@@ -91,25 +91,24 @@ def mime_type
...
       end
-
       x.title title
+      x.section_id section_id
+      x.path path
+      x.layout_id layout_id
       x.system_name system_name
-      x.path(path) if respond_to?(:path)
-      x.hidden hidden?
-      x.layout layout_name
```
akostadinov, line 105: Also I don't remember discussing the removal of this one.

akostadinov, line 105: Also I don't remember discussing the removal of this one.

akostadinov, line 105: Also I don't remember discussing the removal of this one.

app/models/cms/partial.rb
```diff
@@ -19,21 +19,16 @@ def search
...
       unless new_record?
         xml.id id
         xml.created_at created_at.xmlschema
         xml.updated_at updated_at.xmlschema
       end
-
       x.system_name system_name
-      x.content_type content_type
-      x.handler handler
-      x.liquid_enabled liquid_enabled
```
akostadinov, line 32: I guess not relevant for partials?

akostadinov (changes requested):

app/controllers/admin/api/cms/sections_controller.rb
akostadinov: Getting IDs is also not content. If one needs the content, then one can request the other items with filter. It is common that UI and API do not translate 1-1 with each other.

akostadinov: Getting IDs is also not content. If one needs the content, then one can request the other items with filter. It is common that UI and API do not translate 1-1 with each other.

akostadinov: Getting IDs is also not content. If one needs the content, then one can request the other items with filter. It is common that UI and API do not translate 1-1 with each other.

mayorova: I also think it's nice to have the contents of the section (the IDs of the elements inside of it). But my only concern is that if we add it, it might be more difficult to remove this in future.
But I'd vote for including it, and maybe if we see that it gets much traffic, we can optimize it by implementing some more efficient queries with joins.

akostadinov: I think this is feature creep, when we don't know if something is useful or not but potentially inefficient.

In current state, it makes sense to me to include child objects summaries as well to the response, because we already get these from database anyway. So if client needs only summaries, then it will not have to perform further queries on children.

akostadinov: So I actually have a pretty strong opinion here to remove this content. When we write a client, we can easily and painlessly add it. What is the benefit to add it now? We don't set the feature in stone and adding it later should not break existing clients. I see no drawbacks to leave it out for now.

mayorova: I agree with the latest statement.
We can have this one feature in our "TODO"  (or "maybe TODO") list for future, and add it at any later point.

app/controllers/admin/api/cms/sections_controller.rb
jlledom: `system_name` is unique, yes. I prefer to not hardcode any SQL in the code that could make a future rails upgrades harder, so I'll stick to the `.where`  + `.first` approach. However, I actually like the `.tap` styling. I tried to adapt the current query to the tap styling but it doesn't work because `tap` always returns `current_account.sections` and the modifications inside the block are lost. Finally I'm leaving this as it was (sad face).

app/controllers/admin/api/cms/sections_controller.rb
jlledom: I removed this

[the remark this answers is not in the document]
jlledom: I removed the contents from the response

app/controllers/admin/api/cms/sections_controller.rb
```diff
@@ -101,9 +102,19 @@ def destroy
   end
 
   private
-    def find_section
-      @section = current_account.sections.find_by_id(params[:id]) || current_account.sections.find_by_system_name(params[:id])
 
-      raise ActiveRecord::RecordNotFound.new("Couldn't find CMS::Section with id or system_name=#{params[:id]}") if @section.nil?
-    end
+  def find_section(contents: false)
+    q = current_account.sections
```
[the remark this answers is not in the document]
jlledom: No need anymore! that code has been removed.

app/controllers/admin/api/cms/sections_controller.rb
```diff
@@ -101,9 +102,19 @@ def destroy
...
+    q = q.includes(:pages, :files, :builtins, :children) if contents
+    q = q.where(id: params[:id]).or(q.where(system_name: params[:id]))
+    q.load
```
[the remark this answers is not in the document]
jlledom: I'm totally for using services for all. Controller should only manage parameters and permissions, everything els is business logic and should be moved to services that can be easily unit tested. Anyway, I removed that query so no need for now.

app/models/cms/builtin.rb
jlledom: I think I prefer to leave the lookup table, because it works as a single source of truth for tags reference and shouldn't be a difference in performance, right?

app/models/cms/builtin.rb
jlledom: Not this in particular but we agreed to remove all non-sensic fields from responses. I realized about this fields after the call.

app/models/cms/layout.rb
jlledom: Same, removed for being fixed, empty or not applicable fields for this type.

app/models/cms/page.rb
jlledom: Same, removed for being fixed, empty or not applicable fields for this type.

jlledom: Same, removed for being fixed, empty or not applicable fields for this type.

app/models/cms/partial.rb
jlledom: Same, removed for being fixed, empty or not applicable fields for this type.

app/models/cms/builtin.rb
jlledom: I added tests for:

- HTML inside CDATA
- CDATA inside CDATA

test/integration/cms/api/templates_test.rb
```diff
@@ -135,6 +135,22 @@ def setup
         assert_equal '/cool', doc.xpath('/page/path').text
       end
 
+      {
+        HTML: '<html><body><p>This is a test</p></body></html>',
+        CDATA: '<![CDATA[<html><body><p>This is a test</p></body></html>]]>'
+      }.each do |name, published|
+        test "show a page with CDATA + #{name} content" do
```
akostadinov, line 142: nitpick
```suggestion
        test "show a page with #{name} inside the content" do
```

app/models/cms/builtin.rb
akostadinov: To me a transformation method would be a better source of truth because it immediately shows that there is a convention without the need to figure this out in the lookup table.

Ideally we would have a method `self.class.xml_tag` or if that is probably also valid for JSON, maybe `::response_tag` or `::data_tag` or something better. And I assume this method can be defined in some parent class.

test/integration/cms/api/templates_test.rb
jlledom: I committed this from the Github UI, first time ever!

app/models/cms/builtin.rb
jlledom: I committed the last idea: a new `self.class.data_tag` attribute

akostadinov (approved):
Thank you.

thalesmiguel (commented):
Just a couple notes:

1. The new (and nice) `CMS::DataTag` is added to classes that do not set a data_tag with `has_data_tag`. This results in classes with a `nil` data_tag class attribute. From the code i saw, this doesn't have an impact but I'm wondering **IF** we should include the module only on the classes that make use of it.
2. `CMS::DataTag` is missing a test and I think we should add a test for each class that uses `has_data_tag` to assert the data_tag value. This way, if someone in the future changes a tag, the test would break.

jlledom:
> 1. The new (and nice) `CMS::DataTag` is added to classes that do not set a data_tag with `has_data_tag`. This results in  classes with a `nil` data_tag class attribute. From the code i saw, this doesn't have an impact but I'm wondering **IF**  we should include the module only on the classes that make use of it.

It could only fail if `class.data_class` is being accessed for a Class without `has_data_tag`. I checked that `data_tag` has a value for all cases where it's being accessed. The only classes where `DataTag` is included but `has_data_tag` is not called are `CMS::EmailTemplate` and the ones inheriting from `CMS::Portlet::Base`. `CMS::EmailTemplate` isn't calling `self.class.data_tag` and `EmailTemplate.data_class` isn't being called from anywhere, so it shouldn't fail. About the `CMS::Portlet::*` classes, those aren't calling `data_tag` neither and they are going to be removed eventually.

> 2. `CMS::DataTag` is missing a test and I think we should add a test for each class that uses `has_data_tag` to assert the data_tag value. This way, if someone in the future changes a tag, the test would break.

I added a test for the module itself, but I'm not testing tags for CMS resources because that'll be part of [THREESCALE-8904](https://issues.redhat.com/browse/THREESCALE-8904)  (GSoC 2023), and I don't think it's worth to add tests to remove them in a few months.

akostadinov:
How is THREESCALE-8904 related to `#data_tag`?

thalesmiguel (dismissed):
🚀

app/models/cms/builtin.rb
```diff
@@ -111,10 +111,12 @@ class StaticPage < CMS::Builtin
       record.errors.add attr, :is_present if value.present?
     end
 
+    has_data_tag :builtin_page
+
     def to_xml(options = {})
       xml = options[:builder] || Nokogiri::XML::Builder.new
 
-      xml.builtin_page do |x|
+      xml.__send__(self.class.data_tag) do |x|
```
mayorova, line 119: Why this change for all classes? What is its benefit?

Not that I am opposing to it, but it makes it more difficult to understand what's going on (you need to go check what the `CMS::DataTag` module does... ).

app/models/cms/builtin.rb
```diff
@@ -132,23 +134,23 @@ def to_xml(options = {})
...
         unless new_record?
           xml.id id
           xml.created_at created_at.xmlschema
           xml.updated_at updated_at.xmlschema
         end
-
-        # x.title title
-        x.system_name system_name
```
mayorova, line 146: I don't remember what I said when we discussed this, but now checking this, I think maybe it makes sense to keep the `system_name` field. The titles of many built-in pages are the same ("New", "Show", "Index"), but the system name does explain what it is for (`applications/new`, `login/new` etc.).

app/models/cms/page.rb
```diff
@@ -91,25 +93,24 @@ def mime_type
...
+    xml.__send__(self.class.data_tag) do |x|
       unless new_record?
         x.id id
         x.created_at created_at.xmlschema
         x.updated_at updated_at.xmlschema
       end
-
       x.title title
+      x.section_id section_id
+      x.path path
```
mayorova, line 104: @jlledom Have you checked why it was `x.path(path)` before? Could it be that this pre-processing is still needed?

app/models/cms/builtin.rb
jlledom: Maybe it's over-engineering. This came from `CMS::Section` and `CMS::Builtin::Section` sharing the same tag `<section>`. That was wrong and I needed to add a different tag for `CMS::Builtin::Section`, but to do so I would've needed to add a new `#to_xml` method to that class in order to call it's father only to change the root tag name, like in [CMS::Builtin::Partial](https://github.com/3scale/porta/blob/769f3623c35554933512dbdbc1b532d4aa70efac/app/models/cms/builtin.rb#L254-L256) and [CMS::Partial](https://github.com/3scale/porta/blob/769f3623c35554933512dbdbc1b532d4aa70efac/app/models/cms/partial.rb#L22). I thought that was too complicated and it would be good to simplify the models. I moved root tag names to a different module `CMS::DataTag` and unified all classes to use it. Thanks to that I could remove `#to_xml` also from `CMS::Builtin::Partial`.

mayorova: OK, thanks for the explanation!

app/models/cms/page.rb
mayorova: Nevermind, sorry, I completely misread it before :facepalm: 

I somehow thought it was calling the `self.path` method https://github.com/3scale/porta/pull/3193/files#diff-9715ba077dc884135c880762dff0260b7b37c6ce7cd310855879c81551e4da35L33-L42

Of course it's not the case, but I am wondering what that method is for :thinking: It doesn't seem that it is used. But anyway, not part of this PR, we can clean up at some later point.

I am still wondering though, why in the original implementation it included `respond_to?(:path)` :thinking:  Any idea?

mayorova (approved):

app/models/cms/page.rb
jlledom: No worries! About `respond_to?(:path)`, I found the commit: https://github.com/3scale/system/commit/2a41b095b5244b21fe3a6d7bd2f2c0f51f7efa0c. It doesn't explain why it calls `respond_to?` but it seems redundant to me, since it's a column name, it must exist.

PR #3196, opened 2023-02-03
free the Sphinx 1/2

akostadinov:
To allow decoupling of Sphinx and porta container requirements, stop bundling Sphinx with Porta.

But also generate Sphinx configuration file to be easily available when building the searchd standalone image.

part of THREESCALE-8729

### DIFF
--- a/config/docker/thinking_sphinx.yml
+++ b/config/docker/thinking_sphinx.yml
@@ -9,6 +9,7 @@ production:
   big_document_ids: true
 
   configuration_file: <%= ENV.fetch('THINKING_SPHINX_CONFIGURATION_FILE') { Rails.root.join("config/#{Rails.env}.sphinx.conf")} %>
+  indices_location: <%= ENV.fetch('THINKING_SPHINX_INDICES_LOCATION', Rails.root.join("db/sphinx/#{Rails.env}")) %>
 
   # Batch size for real-time index processing (via the `ts:index` and `ts:rebuild` tasks), the default is 1000
   batch_size: <%= ENV.fetch('THINKING_SPHINX_BATCH_SIZE', 1000) %>
--- a/lib/tasks/openshift.rake
+++ b/lib/tasks/openshift.rake
@@ -47,6 +47,11 @@ ERROR_MESSAGE
       daemon.start
     end
 
+    desc 'Generate configuration file for Thinking Sphinx engine'
+    task configure: %i[environment] do
+      ThinkingSphinx::RakeInterface.new(nodetach: true).configure
+    end
+
     desc 'Cleanup pid and lock files from unclean shutdown'
     task cleanup: %i[environment] do
       indices_location = ThinkingSphinx::Configuration.instance.indices_location
--- a/openshift/system/Dockerfile
+++ b/openshift/system/Dockerfile
@@ -1,4 +1,4 @@
-FROM quay.io/centos7/ruby-26-centos7
+FROM quay.io/centos7/ruby-26-centos7 AS porta-base
 
 USER root
 
@@ -91,11 +91,31 @@ RUN export ${BUNDLER_ENV} >/dev/null \
     && rm log/*.log \
     && chmod g+w /opt/system/config
 
-USER 1001
 ADD openshift/system/entrypoint.sh /opt/system/entrypoint.sh
+USER 1001
 EXPOSE 3000 9306
 # TODO: dumb-init!
 ENTRYPOINT ["/opt/system/entrypoint.sh"]
 CMD ["unicorn", "-c", "config/unicorn.rb", "-E", "${RAILS_ENV}", "config.ru"]
 
-# vim: set ft=dockerfile:
+
+FROM porta-base AS porta-sphinx-config
+ENV THINKING_SPHINX_ADDRESS=0.0.0.0 \
+    THINKING_SPHINX_PID_FILE=/var/run/sphinx/searchd.pid \
+    THINKING_SPHINX_INDICES_LOCATION=/var/lib/sphinx \
+    THINKING_SPHINX_CONFIGURATION_FILE=/opt/system/config/standalone.sphinx.conf \
+    RAILS_ENV=production \
+    SECRET_KEY_BASE=dummy \
+    DATABASE_URL='mysql2://root:@localhost/porta'
+USER 0
+RUN yum install -y mariadb-server \
+    && setpriv --reuid 27 --regid 27 --clear-groups mysql_install_db \
+    && (mysqld_safe &) \
+    && cp -v config/docker/thinking_sphinx.yml config/ \
+    && ./entrypoint.sh bundle exec rake db:create db:schema:load openshift:thinking_sphinx:configure \
+    && grep -q "rt_field = account_id" "$THINKING_SPHINX_CONFIGURATION_FILE" \
+    && kill $(</var/run/mariadb/mariadb.pid)
+
+
+FROM porta-base AS porta-prod
+COPY --from=porta-sphinx-config /opt/system/config/standalone.sphinx.conf /opt/system/config/


mayorova:
What about using the nulldb DB adapter? https://github.com/nulldb/nulldb

That's just adding to `Gemfile`
```
gem 'activerecord-nulldb-adapter'
```

And then the config can be generated with:
```
DATABASE_URL=nulldb://dummy SECRET_KEY_BASE=dummy THINKING_SPHINX_CONFIGURATION_FILE=/opt/system/config/sphinx.conf RAILS_ENV=production bundle exec rake openshift:thinking_sphinx:configure
```

`THINKING_SPHINX_CONFIGURATION_FILE` is optional, but I think it's nice to make it explicit.

The drawback is that we need to install another gem, and make it available in production environment...

akostadinov:
The `nulldb` adapter looks look and I was quick to celebrate. It works very easily without error but unfortunately I found out that the generated sphinx configs are badly flawed. See attached

```diff
--- mysql.sphinx.conf	2023-02-07 20:54:38.747654752 +0200
+++ nulldb.sphinx.conf	2023-02-07 20:54:55.671708043 +0200
@@ -17,45 +17,17 @@
 {
   type = rt
   path = /home/user/ws/repos/porta/db/sphinx/development/account_core
-  charset_table = 0..9, A..Z->a..z, a..z, U+23, U+25, U+27, U+2A..U+2C, U+2E, U+3A, U+3B, U+3F, U+5F, U+60, U+7B, U+7D
-  min_infix_len = 3
-  html_strip = 1
-  html_index_attrs = img=alt,title; a=title
-  html_remove_elements = style, script
-  rt_field = name
-  rt_field = account_id
-  rt_field = username
-  rt_field = user_full_name
-  rt_field = email
-  rt_field = user_key
-  rt_field = app_id
-  rt_field = app_name
-  rt_field = user_id
   rt_attr_uint = sphinx_deleted
   rt_attr_bigint = sphinx_internal_id
-  rt_attr_bigint = provider_account_id
-  rt_attr_bigint = tenant_id
   rt_attr_string = sphinx_internal_class
-  rt_attr_string = state
 }
 
 index topic_core
 {
   type = rt
   path = /home/user/ws/repos/porta/db/sphinx/development/topic_core
-  charset_table = 0..9, A..Z->a..z, a..z
-  min_infix_len = 3
-  html_strip = 1
-  html_index_attrs = img=alt,title; a=title
-  html_remove_elements = style, script
-  rt_field = title
-  rt_field = post
   rt_attr_uint = sphinx_deleted
-  rt_attr_uint = sticky
-  rt_attr_uint = last_updated_at
   rt_attr_bigint = sphinx_internal_id
-  rt_attr_bigint = tenant_id
-  rt_attr_bigint = forum_id
   rt_attr_string = sphinx_internal_class
 }
 
@@ -63,12 +35,6 @@
 {
   type = rt
   path = /home/user/ws/repos/porta/db/sphinx/development/backend_api_core
-  charset_table = 0..9, A..Z->a..z, a..z
-  min_infix_len = 3
-  html_strip = 1
-  html_index_attrs = img=alt,title; a=title
-  html_remove_elements = style, script
-  rt_field = name
   rt_attr_uint = sphinx_deleted
   rt_attr_bigint = sphinx_internal_id
   rt_attr_string = sphinx_internal_class
@@ -78,17 +44,8 @@
 {
   type = rt
   path = /home/user/ws/repos/porta/db/sphinx/development/cms_page_core
-  charset_table = 0..9, A..Z->a..z, a..z
-  min_infix_len = 3
-  html_strip = 1
-  html_index_attrs = img=alt,title; a=title
-  html_remove_elements = style, script
-  rt_field = sphinx_internal_class_name
-  rt_field = title
-  rt_field = published
   rt_attr_uint = sphinx_deleted
   rt_attr_bigint = sphinx_internal_id
-  rt_attr_bigint = tenant_id
   rt_attr_string = sphinx_internal_class
 }
 
@@ -96,13 +53,6 @@
 {
   type = rt
   path = /home/user/ws/repos/porta/db/sphinx/development/email_configuration_core
-  charset_table = 0..9, A..Z->a..z, a..z
-  min_infix_len = 3
-  html_strip = 1
-  html_index_attrs = img=alt,title; a=title
-  html_remove_elements = style, script
-  rt_field = email
-  rt_field = user_name
   rt_attr_uint = sphinx_deleted
   rt_attr_bigint = sphinx_internal_id
   rt_attr_string = sphinx_internal_class
@@ -112,12 +62,6 @@
 {
   type = rt
   path = /home/user/ws/repos/porta/db/sphinx/development/metric_core
-  charset_table = 0..9, A..Z->a..z, a..z
-  min_infix_len = 3
-  html_strip = 1
-  html_index_attrs = img=alt,title; a=title
-  html_remove_elements = style, script
-  rt_field = friendly_name
   rt_attr_uint = sphinx_deleted
   rt_attr_bigint = sphinx_internal_id
   rt_attr_string = sphinx_internal_class
@@ -127,13 +71,6 @@
 {
   type = rt
   path = /home/user/ws/repos/porta/db/sphinx/development/plan_core
-  charset_table = 0..9, A..Z->a..z, a..z
-  min_infix_len = 3
-  html_strip = 1
-  html_index_attrs = img=alt,title; a=title
-  html_remove_elements = style, script
-  rt_field = sphinx_internal_class_name
-  rt_field = name
   rt_attr_uint = sphinx_deleted
   rt_attr_bigint = sphinx_internal_id
   rt_attr_string = sphinx_internal_class
@@ -143,29 +80,15 @@
 {
   type = rt
   path = /home/user/ws/repos/porta/db/sphinx/development/proxy_rule_core
-  charset_table = 0..9, A..Z->a..z, a..z, U+21, U+27, U+28..U+2F, U+5F
-  min_infix_len = 1
-  html_strip = 1
-  html_index_attrs = img=alt,title; a=title
-  html_remove_elements = style, script
-  rt_field = pattern
   rt_attr_uint = sphinx_deleted
   rt_attr_bigint = sphinx_internal_id
-  rt_attr_bigint = owner_id
   rt_attr_string = sphinx_internal_class
-  rt_attr_string = owner_type
 }
 
 index service_core
 {
   type = rt
   path = /home/user/ws/repos/porta/db/sphinx/development/service_core
-  charset_table = 0..9, A..Z->a..z, a..z
-  min_infix_len = 3
-  html_strip = 1
-  html_index_attrs = img=alt,title; a=title
-  html_remove_elements = style, script
-  rt_field = name
   rt_attr_uint = sphinx_deleted
   rt_attr_bigint = sphinx_internal_id
   rt_attr_string = sphinx_internal_class
```

I think it doesn't really detect any attributes because `nulldb` is ... `null`.

Maybe this would be fixable by generating a *non-documented* `myschema.rb` file but there is IMO too big risk of it becoming outdated some day and this biting us in the a^Mback.

Also the `nulldb` gem has not received any commits since 1 year which is not too bad for the intended usage but still.

So I think that although it's much work to spin up a real database, it is the safest way forward. Unless you tell me other options.

akostadinov:
btw, now I notive that we will need to handle these `path` parameters on the sphinx container side. Probably some `sed` expression. I think they are fine to stay as they are here though.

akostadinov:
But I will make this MySQL thing nicer. Need some time to fiddle with it.

mayorova:
@akostadinov oh, yes, you're totally right :facepalm: 

I just compared the `production.sphinx.conf` generated by the command I pasted above with the existing `development.sphinx.conf` I assumed I had generated earlier with the real DB, and they were equal.
But apparently, I had run the generation with nulldb on the development env earlier, so it was also broken :upside_down_face: So I was comparing two broken configs :grimacing: 

Yes, let's go the real DB way. I'd suggest though maybe we should use real MySQL and not MariaDB. Or is there any problem with that?

akostadinov:
MariaDB is a fork of MySQL. We don't have MySQL server available in in official RHEL repos, I'm not sure about CentOS but wanted to have what we have in RHEL. I think it is fine as long as the MySQL schema is loaded without errors.

akostadinov:
Now ready for review. Doing mysql in another stage not to mess up with the image.

openshift/system/Dockerfile
```diff
@@ -91,11 +82,23 @@ RUN export ${BUNDLER_ENV} >/dev/null \
...
 # TODO: dumb-init!
 ENTRYPOINT ["/opt/system/entrypoint.sh"]
 CMD ["unicorn", "-c", "config/unicorn.rb", "-E", "${RAILS_ENV}", "config.ru"]
 
-# vim: set ft=dockerfile:
+
+FROM porta-base AS porta-sphinx-config
+USER 0
+RUN yum install -y mariadb-server \
+    && setpriv --reuid 27 --regid 27 --clear-groups mysql_install_db \
```
jlledom, line 96: Why this?

openshift/system/Dockerfile
```diff
@@ -91,11 +82,23 @@ RUN export ${BUNDLER_ENV} >/dev/null \
...
+    && (mysqld_safe &) \
+    && DATABASE_URL='mysql2://root:@localhost/3scale_system_production' SECRET_KEY_BASE=dummy RAILS_ENV=production /opt/system/entrypoint.sh bundle exec rake db:create db:schema:load openshift:thinking_sphinx:configure \
```
jlledom, line 98: Is this a disposable database? Why does Sphinx need a mysql database to generate the config?

openshift/system/Dockerfile
```diff
@@ -40,15 +40,6 @@ RUN yum-config-manager --save --setopt=cbs.centos.org_repos_sclo7-$RUBY_SCL-rh-c
...
 
-# Install sphinx search
-RUN if [ "`uname -m`" == "x86_64" ]; then \
-      yum install -y https://sphinxsearch.com/files/sphinx-2.2.11-1.rhel7.x86_64.rpm ; \
-    else \
-      curl -sSL http://sphinxsearch.com/files/sphinx-2.2.11-release.tar.gz | tar xz -C /tmp \
-      && cd /tmp/sphinx-2.2.11-release && ./configure && make install \
-      && rm -rf /tmp/sphinx* ; \
-    fi
-
```
jlledom, line 51: So, Sphinx will be running in another pod with a different image?

jlledom (dismissed):
It looks good to me. I asked some doubts.

openshift/system/Dockerfile
akostadinov: sudo without sudo.

openshift/system/Dockerfile
akostadinov: sphinx doesn't generate the config. Porta does and to start porta, one needs a database. Not sue what  you mean by "disposable" database. Image layer is gone afterwards so state of DB doesn't matter.

jlledom (approved):
LGTM

openshift/system/Dockerfile
```diff
@@ -91,11 +82,33 @@ RUN export ${BUNDLER_ENV} >/dev/null \
...
+FROM porta-base AS porta-sphinx-config
+ENV THINKING_SPHINX_ADDRESS=0.0.0.0 \
+    THINKING_SPHINX_PORT=9306 \
+    THINKING_SPHINX_PID_FILE=/var/run/sphinx/searchd.pid \
+    THINKING_SPHINX_BATCH_SIZE=100 \
+    THINKING_SPHINX_INDICES_LOCATION=/var/lib/sphinx \
+    THINKING_SPHINX_CONFIGURATION_FILE=/opt/system/config/standalone.sphinx.conf \
+    RAILS_ENV=production \
+    SECRET_KEY_BASE=dummy \
+    DATABASE_URL='mysql2://root:@localhost/porta'
```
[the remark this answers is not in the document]
akostadinov: database name doesn't matter because this layer is scrapped. Port is not desirable because we expect connection ove unix socket, what MySQL does with `localhost` host.

lib/tasks/openshift.rake
```diff
@@ -47,6 +47,11 @@ ERROR_MESSAGE
       daemon.start
     end
 
+    desc 'Generate configuration file for Thinking Sphinx engine'
+    task configure: %i(environment) do
```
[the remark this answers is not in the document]
akostadinov: hmm, I'm sure rubocop complained against this at some point but maybe an older version or a different configuration...

config/docker/thinking_sphinx.yml
```diff
@@ -9,6 +9,7 @@ production:
   big_document_ids: true
 
   configuration_file: <%= ENV.fetch('THINKING_SPHINX_CONFIGURATION_FILE') { Rails.root.join("config/#{Rails.env}.sphinx.conf")} %>
+  indices_location: <%= ENV.fetch('THINKING_SPHINX_INDICES_LOCATION') { Rails.root.join("db/sphinx/#{Rails.env}")} %>
```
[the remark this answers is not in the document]
akostadinov: Thank you. Don't want to change existing lines. Makes easier to see what changed.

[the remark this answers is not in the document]
akostadinov: but also one space is needed after `=`

```suggestion
  indices_location: <%= ENV.fetch('THINKING_SPHINX_INDICES_LOCATION', Rails.root.join("db/sphinx/#{Rails.env}")) %>
```

mayorova:
Hmmm, I'm getting
```
Can't connect to local MySQL server through socket '/var/lib/mysql/mysql.sock' (2)
Couldn't create 'porta' database. Please check your configuration.
rake aborted!
Mysql2::Error::ConnectionError: Can't connect to local MySQL server through socket '/var/lib/mysql/mysql.sock' (2)
/opt/system/vendor/bundle/ruby/2.6.0/gems/mysql2-0.5.3/lib/mysql2/client.rb:90:in `connect'
[...]
Tasks: TOP => db:create
Error: error building at STEP "RUN yum install -y mariadb-server     && setpriv --reuid 27 --regid 27 --clear-groups mysql_install_db     && (mysqld_safe &)     && cp -v config/docker/thinking_sphinx.yml config/     && ./entrypoint.sh bundle exec rake db:create db:schema:load openshift:thinking_sphinx:configure     && grep -q "rt_field = account_id" "$THINKING_SPHINX_CONFIGURATION_FILE"     && kill $(</var/run/mariadb/mariadb.pid)": error while running runtime: exit status 1

```

akostadinov:
> Hmmm, I'm getting

You've got how? This is the quay build:
https://quay.io/repository/3scale/porta/build/b35c5966-0f60-4908-b072-3aea8e3a3a94

mayorova:
Well, I was just running
```
podman build -f openshift/system/Dockerfile --tag quay.io/3scale/porta:sphinx-conf .
```

> You've got how? This is the quay build:
> https://quay.io/repository/3scale/porta/build/b35c5966-0f60-4908-b072-3aea8e3a3a94

Hm, from the log it looks like it built and pushed just the first stage (`porta-base`), but the image `quay.io/3scale/porta:free-sphinx` does seem to have the right sphinx config. Is it built through this automated build? :thinking:

akostadinov:
For other lurkers, this is probably because quay cust some of the logs unfortunately https://issues.redhat.com/browse/PROJQUAY-5076 but I have built that locally as well. I'll try again.

Also a tip, to view downloaded Quay logs, one can do use `jq`:
```
jq -r '.logs[].message' < /tmp/41e1de3c-2a65-48db-ad02-720eb0a13d3d.json | less -R
```

openshift/system/Dockerfile
```diff
@@ -91,11 +82,33 @@ RUN export ${BUNDLER_ENV} >/dev/null \
...
 EXPOSE 3000 9306
 # TODO: dumb-init!
 ENTRYPOINT ["/opt/system/entrypoint.sh"]
 CMD ["unicorn", "-c", "config/unicorn.rb", "-E", "${RAILS_ENV}", "config.ru"]
 
-# vim: set ft=dockerfile:
+
+FROM porta-base AS porta-sphinx-config
+ENV THINKING_SPHINX_ADDRESS=0.0.0.0 \
+    THINKING_SPHINX_PORT=9306 \
```
mayorova, line 95: `THINKING_SPHINX_PORT` is currently hardcoded in the `thinking_sphinx.yml`.

I guess it only makes sense to include it here if you also include ENV fetch in the yml.

openshift/system/Dockerfile
```diff
@@ -91,11 +82,33 @@ RUN export ${BUNDLER_ENV} >/dev/null \
...
+    THINKING_SPHINX_PID_FILE=/var/run/sphinx/searchd.pid \
+    THINKING_SPHINX_BATCH_SIZE=100 \
```
mayorova, line 97: `THINKING_SPHINX_BATCH_SIZE` only affects the client side, it does not pass to the Sphinx .conf file

akostadinov:
leaving original sphinx install for now, until we can get operator catch up, please approve @mayorova

mayorova (approved):
Well done!

PR #3198, opened 2023-02-07
👷 Separate lint from webpack compilation

josemigallas:
`lint` does not depends on assets compilation, only on `yarn install`.

On the other hand, assets compilation is already run by mysql_build. This is a relatively heavy process and we can save time by running it only once.

**mysql_build**
Screenshot 2023-02-07 at 16 59 37

**javascript_build**
Screenshot 2023-02-07 at 13 25 32

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,8 +63,7 @@ generate-assets-checksums: &generate-assets-checksums
 use-example-config-files: &use-example-config-files
   run:
     name: Copy example config files into place to be used by tests
-    command: |
-      cp config/examples/*.yml config/
+    command: cp config/examples/*.yml config/
 
 disable-internet-access: &disable-internet-access
   run:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1177,21 +1200,33 @@ workflows:
...
           requires:
+            - licenses
+            - swagger
             - lint
             - jest
           <<: *only-master-filter
       - notify_failure:
           requires:
+            - licenses
+            - swagger
```
akostadinov, line 1229: leave off `notify_failure`. @jlledom is fixing some stuff with that so better avoid conflict.

akostadinov: Looks good otherwise and no additional approval from me needed.

akostadinov:
don't licenses also depend on `yarn install`? How does license_finder know about npm deps licenses?

josemigallas:
> don't licenses also depend on `yarn install`? How does license_finder know about npm deps licenses?

It reads the yarn.lock, not installed modules.

akostadinov:
btw why then we need assets compiled for integration and functional tests? They don't start a browser IIRC.

josemigallas:
> btw why then we need assets compiled for integration and functional tests? They don't start a browser IIRC.

Good catch! I will try removing those, only cucumber should require assets.

josemigallas:
@akostadinov Turns out they do depend on asset compilation! 😅

.circleci/config.yml
```diff
@@ -473,22 +472,44 @@ jobs:
...
+      - *attach-to-workspace
+      - *use-example-config-files
       - run:
           name: Licenses check
-          command: |
-            bundle exec rake licenses:compliance
+          command: bundle exec rake licenses:compliance
+      - store_artifacts:
+          path: doc/licenses
+          destination: licenses
```
akostadinov, line 495: not sure we need this. It is in the repo already

akostadinov (approved):

.circleci/config.yml
akostadinov: it was there already in the past so fine to merge. Just question.

PR #3200, opened 2023-02-07
⬆️ 🥒 Upgrade cucumber 8

josemigallas:
`map_column!` and `map_header!` have been removed.


features/step_definitions/buyers_management/account_steps.rb
```diff
@@ -31,7 +31,7 @@
 
 Then /^(.*) for (?:buyer|provider|account) "([^"]*)"$/ do |action, org_name|
   account = Account.find_by!(org_name: org_name)
-  within('#' + dom_id(account)) { step action }
+  within("##{dom_id(account)}") { step action }
```
akostadinov, line 34: probably intentional duplicate symbol, but just double-checking

features/support/alerts.rb
```diff
@@ -1,23 +1,18 @@
-# -*- coding: utf-8 -*-
-module Alerts
+# frozen_string_literal: true
 
-  def symbolize_headers(table)
-    table.map_headers! { |header| header.downcase.gsub(/\s+/, '_').to_sym }
-  end
```
akostadinov, line 6: Although it has a single usage only I'd suggest moving it under `DataTableTransforms` and still use it in the step.

akostadinov (approved):
Just two small questions. Leaving it to your discretion to decide what to do with them.

features/support/alerts.rb
josemigallas: Move to DataTableTransforms.

PR #3201, opened 2023-02-08
Require the current password on the password update form

lvillen:

[THREESCALE-6648 Require the current password on the password update form](https://issues.redhat.com/browse/THREESCALE-6648)

**Verification steps** 


lib/developer_portal/app/controllers/developer_portal/admin/account/personal_details_controller.rb
```diff
@@ -61,4 +63,17 @@ def redirect_path
     end
   end
 
+  def user_params
+    params.require(:user)
+  end
+
+  def current_password_verification
+    return true unless current_user.using_password?
```
thalesmiguel, line 71: Preemptive review (2 in 1 bargain edition):
1. Changing the name of the method to an action makes its intentions clearer.
2. Since you're not using the `true` value this method return, if can simply return `nil`. Also, if there's a `return true`, I'd also expect the possibility of a `return false`. 

```suggestion
  def verify_current_password
    return unless current_user.using_password?
```

features/developer_portal/admin/account/personal_details.feature
```diff
@@ -0,0 +1,36 @@
...
+    And clicks on update personal details
+    Then they should see current password is incorrect
+
+  Scenario: Buyer wants to update their password adding their current password
+    Given the buyer wants to edit their personal details
+    When the buyer writes their current password
+    And the buyer writes a new password
+    And the buyer confirms their new password
+    And clicks on update personal details
+    Then they should have their password updated
```
josemigallas, line 36: I'm afraid this is all testing the same, IMO there are 2 possible scenarios: with and without current password. You could go the brief way, something like:
```suggestion
  Scenario: Buyer uses the wrong current password
    Given the buyer wants to edit their personal details
    When they change their personal details and password
    But they add the wrong current password
    Then they should not be able to edit their personal details
    
  Scenario: Buyer uses the wrong current password
    Given the buyer wants to edit their personal details
    When they change their personal details and password
    And they add the correct current password
    Then they should be able to edit their personal details
```

Or the long way, reusing the steps available, such as:
```cucumber
And go to the personal details page
And fill in "Username" with "Alfred"
And press on "Update Personal Details" 
```
(See features/step_definitions/web_steps.rb)

features/step_definitions/developer_portal/admin/account/personal_details_steps.rb
```diff
@@ -0,0 +1,47 @@
...
+
+Then "they should see current password is incorrect" do
+  assert_flash 'Current password is incorrect.'
+  assert_current_path admin_account_personal_details_path
+end
+
+Then "they should see their information updated" do
+  assert_flash 'User was successfully updated.'
+  assert_current_path admin_account_personal_details_path
+  assert_equal find('#user_username').value, current_user.username
```
josemigallas, line 39:
```suggestion
  assert has_css?('#user_username', text: current_user.username)
```

features/step_definitions/developer_portal/admin/account/personal_details_steps.rb
```diff
@@ -0,0 +1,47 @@
...
+  assert_equal find('#user_email').value, current_user.email
+end
+
+Then "they should have their password updated" do
+  assert_flash 'User was successfully updated.'
+  assert_current_path admin_account_personal_details_path
+  assert_equal BCrypt::Password.new(current_user.password_digest), 'ultrasecret'
```
josemigallas, line 46: Use an instance variable to refer to the tested value across multiple steps:
```suggestion
  assert_equal BCrypt::Password.new(current_user.password_digest), @new_password
```

lib/developer_portal/app/controllers/developer_portal/admin/account/personal_details_controller.rb
```diff
@@ -61,4 +63,17 @@ def redirect_path
     end
   end
 
+  def user_params
+    params.require(:user)
+  end
+
+  def verify_current_password
+    return unless current_user.using_password?
```
josemigallas, line 71: What user is not using a password? Is there any test for this?

lib/developer_portal/app/controllers/developer_portal/admin/account/personal_details_controller.rb
```diff
@@ -61,4 +63,17 @@ def redirect_path
...
+
+    unless current_user.authenticated?(user_params[:current_password])
+      flash.now[:error] = 'Current password is incorrect.'
+      assign_user_drop
+      render :action => 'show'
```
josemigallas, line 76: This looks like duplicated logic. It should be part of `update` somehow. Maybe add a validator for password if that's even possible or extend `update!` and return an error so that it can be catched with https://github.com/3scale/porta/blob/8612ff9ee902063a0e4c45dc49a4ffa2133e161b/lib/developer_portal/app/controllers/developer_portal/admin/account/personal_details_controller.rb#L28-L31

lib/developer_portal/app/views/developer_portal/user/show.html.liquid
```diff
@@ -11,20 +11,47 @@
...
+              </label>
+              <div class="col-md-6">
+                <input
+                  id="user_current_password"
+                  name="user[current_password]"
+                  type="password"
+                  class="form-control">
+                {{ user.errors.current_password | inline_errors: 'help-block' }}
+              </div>
+            </div>
```
josemigallas, line 38: If user not `.using_password?`, will this be disabled / hidden? How does it work in that scenario?

josemigallas (changes requested):

features/developer_portal/admin/account/personal_details.feature
josemigallas: What should be tested here though is other validations, such as, sending empty form, sending an invalid email, wrong confirmation password.

lib/developer_portal/app/views/developer_portal/user/show.html.liquid
lvillen: I believe it's not being handled, but I'm taking care of this.

lib/developer_portal/app/controllers/developer_portal/admin/account/personal_details_controller.rb
```diff
@@ -10,25 +10,21 @@ class DeveloperPortal::Admin::Account::PersonalDetailsController < ::DeveloperPo
...
-        redirect_to(redirect_path)
-      end
+    if resource.errors.empty? && resource.update(user_params)
+      resource.kill_user_sessions(user_session) if resource.just_changed_password?
 
-      failure.html do
-        assign_user_drop
-        render :action => 'show'
-      end
+      redirect_to redirect_path, notice: 'User was succesfully updated.'
```
mayorova, line 24: `succesfully` => `successfully`

I think this typo was already in the provider controllers, so maybe we need to fix it there also.

lib/developer_portal/app/controllers/developer_portal/admin/account/personal_details_controller.rb
```diff
@@ -50,6 +46,7 @@ def deny_unless_can_update
   end
 
   def redirect_path
+    #TODO: Test this
```
mayorova, line 49: OK, I think that this code is quite outdated...

I am not sure, but I think in the past (like, long long time ago), the admin portal (provider) users could log in to the developer portal. Now it's not the case, in fact the admin portal and developer portal are two different Rails engines (on different domains also), so there is no way to jump between one and another. Also, the users of the two portals are different, provider users can't login into the developer portal.

Also, I think `arams[:origin] == "users"` does not apply here.

In the admin portal it is valid, because there are two places from where you can go to the personal details update form:
1. From the menu **Account Settings > Personal > Personal Details** https://github.com/3scale/porta/blob/3017562449d3275840cae6afe338a01c1dcca899/app/views/shared/provider/navigation/account/_personal.html.slim#L4
2. From the list of the provider users **Account Settings > Users > Listing** 
https://github.com/3scale/porta/blob/3017562449d3275840cae6afe338a01c1dcca899/app/views/provider/admin/account/users/index.html.slim#L27
Which is where it may have `params[:origin] == "users"`

In the developer portal though there is just a single entrypoint, and `origin=users` is not set anywhere.
So this can be simplified by always redirecting to the same path.
Now, there are two options: either stay on the Personal Details update form (`admin_account_personal_details_path`), or to go back to the users list page (`admin_account_users_path`). I am not sure which is the best here.

lib/developer_portal/app/controllers/developer_portal/admin/account/personal_details_controller.rb
```diff
@@ -61,4 +58,16 @@ def redirect_path
...
+    # There are extra fields customized by the provider so we can't add a permit here
+    params.require(:user)
+  end
+
+  def verify_current_password
+    return unless current_user.using_password?
+    return if current_user.authenticated?(user_params[:current_password])
+
+    flash.now[:error] = 'Current password is incorrect.'
+    resource.errors.add(:current_password, 'Current password is incorrect.')
```
mayorova, line 71: Maybe move that error to i18n locale? 
But just a suggestion.

features/step_definitions/developer_portal/admin/account/personal_details_steps.rb
```diff
@@ -0,0 +1,91 @@
...
+  click_on 'Update Personal Details'
+  assert has_css?('.help-block', text: 'Current password is incorrect')
+  assert_flash 'CURRENT PASSWORD IS INCORRECT.'
+  assert 'Current password is incorrect.'
+  assert_current_path admin_account_personal_details_path
+end
+
+Then "they should be able to edit their personal details" do
+  click_on 'Update Personal Details'
+  assert_flash 'USER WAS SUCCESFULLY UPDATED.'
```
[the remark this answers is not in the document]
lvillen: This uppercase is necessary due to:
https://github.com/3scale/porta/blob/5c71501b4402da3ca177838ecf1f5314843d8366/features/step_definitions/flash_steps.rb#L16

lib/developer_portal/app/controllers/developer_portal/admin/account/personal_details_controller.rb
lvillen: Nice catch! 🎯

lib/developer_portal/app/controllers/developer_portal/admin/account/personal_details_controller.rb
lvillen: I would say, from an UI perspective, that it makes more sense going to the users list page (`admin_account_users_path`), but I'm open to feedback.

features/developer_portal/admin/account/personal_details.feature
```diff
@@ -0,0 +1,82 @@
+Feature: Dev Portal Buyer Personal Details
+  As a buyer
+  I want to change my personal details
+
+  Background:
+    Given a buyer logged in to a provider
+
+  # This is needed, otherwise assert_flash doesn't receive the right message.
+  # Check flash-buyer.js
```
josemigallas, line 9: Is this comment really necessary?

features/developer_portal/admin/account/personal_details.feature
```diff
@@ -0,0 +1,82 @@
...
+  @javascript
+  Scenario: Buyer doesn't use current password
+    Given the buyer wants to edit their personal details
+    When the buyer edits their personal details
```
josemigallas, line 13: Maybe use _they_ to reduce repetition and verbosity
```suggestion
    When they edit their personal details
```

features/developer_portal/admin/account/personal_details.feature
```diff
@@ -0,0 +1,82 @@
...
+    Then they should not be able to edit their personal details
```
josemigallas, line 14: I think this scenario is missing a step:
_ Given the buyer wants to edit their personal details, when they do, then should not be able to do it._
It doesn't make a lot of sense to me. IMO there should be 2 scenarios: one when the buyer update their details (best case scenario) and another when the password validation fails. Since the current password is a required field, it can be tested in the "form is incomplete" scenario.

features/developer_portal/admin/account/personal_details.feature
```diff
@@ -0,0 +1,82 @@
...
+
+  @javascript
+  Scenario: Buyer uses wrong current password
+    Given the buyer wants to edit their personal details
+    When the buyer edits their personal details
+    And the buyer writes a wrong current password
```
josemigallas, line 20:
```suggestion
    And the buyer uses the wrong current password
```

features/developer_portal/admin/account/personal_details.feature
```diff
@@ -0,0 +1,82 @@
...
+    Then they should not be able to edit their personal details
+
+  @javascript
+  Scenario: Buyer uses correct current password
+    Given the buyer wants to edit their personal details
+    When the buyer edits their personal details
+    And they change their password
+    And the buyer writes a correct current password
+    Then they should be able to edit their personal details
+    And password must have changed
```
josemigallas, line 30: Why the password is checked as a separate field? I think it should part of `they should/could be able to edit their personal details`.

features/developer_portal/admin/account/personal_details.feature
```diff
@@ -0,0 +1,82 @@
...
+
+    And they should not see the fields:
+      | not present          |
+      | Last name            |
+      | Job role             |
+      | User extra read only |
+      | User extra hidden    |
+
+    When the buyer writes a correct current password
+    And they press "Update Personal Details"
```
josemigallas, line 57: This makes me think that current password is verified separately, whereas to me it would make for a better UX to verify everything at the same time. This way, when the password (or the wrong password) is provided, the form will render all errors at the same time: password wrong and missing fields.

features/developer_portal/admin/account/personal_details.feature
```diff
@@ -0,0 +1,82 @@
...
+    Then they should see error in fields:
+      | errors              |
+      | First name          |
+      | User extra required |
+
+    When the buyer edits their custom personal details
+    And the buyer writes a correct current password
```
josemigallas, line 64: These 2 steps should be merged. The step should refer to the whole form, regardless of its set of fields.

features/developer_portal/admin/account/personal_details.feature
```diff
@@ -0,0 +1,82 @@
...
+    Then they should be able to edit their custom personal details
```
josemigallas, line 65: - the buyer edits their custom personal details
- the buyer is able to edit their custom personal details

These two convey the same idea, I think. I would suggest something like
- the buyer edits their customer personal details
- the buyer personal details' have been updated / have changed

And `the buyer is able to...` is the combination of both.

features/developer_portal/admin/account/personal_details.feature
```diff
@@ -0,0 +1,82 @@
...
+  Scenario: Editing of personal details can be blocked
+    Given provider "foo.3scale.localhost" has "useraccountarea" disabled
+    And the buyer follow "Settings"
+    Then they should not see "Personal Details"
+
+  @javascript
+  Scenario: Buyer sends an empty form
+    Given the buyer wants to edit their personal details
+    When they don't provide any personal details
+    Then they should not be able to edit their personal details
```
josemigallas, line 76: This is already tested in https://github.com/3scale/porta/pull/3201/files#diff-558ebc2ae6cdd5df8fec5bd78497c8250398015740fe302f22dc1e8cbc0dace3R58-R61

features/developer_portal/admin/account/personal_details.feature
```diff
@@ -0,0 +1,82 @@
...
+
+  Scenario: Buyer sends a wrong email
+    Given the buyer wants to edit their personal details
+    When fill in "Email" with "email"
+    And the buyer writes a correct current password
+    Then they should see email errors
```
josemigallas, line 82: This one is a bit weird, when is the form submitted? And why is email tested? Are there other validations?
```suggestion
    Given the buyer wants to edit their personal details
    When they submit an invalid email address # Fill form, click submit
    Then they should see "Email is wrong" # Whatever inline error is rendered
```

features/developer_portal/admin/account/personal_details_sso.feature
```diff
@@ -0,0 +1,16 @@
+Feature: Dev Portal Buyer Personal Details
+  As a buyer
+  I want to change my personal details
+
+  Background:
+    Given a buyer logged in to a provider using SSO
+
+  Scenario: Buyer shouldn't see any password input 
+    Given the buyer wants to edit their personal details
+    Then the buyer shouldn't see any reference to password
```
josemigallas, line 10: This scenario is redundant, try to merge it with _the buyer edits their personal details_.

features/step_definitions/buyer_steps.rb
```diff
@@ -155,6 +155,15 @@ def login_form
   )
 end
 
+Given "a buyer logged in to a provider using SSO" do
+  steps %(
```
josemigallas, line 159: Reusing steps as if they were function is generally a bad pattern. If possible (it may be too hard to be worth it) replace this steps with their actual implementation.

features/step_definitions/developer_portal/admin/account/personal_details_steps.rb
```diff
@@ -0,0 +1,79 @@
+# frozen_string_literal: true
+
+Given "the buyer wants to edit their personal details" do
+  visit admin_account_personal_details_path
+end
+
+When "the buyer edits their personal details" do
+  fill_in('user[username]', with: 'Alfred')
+  fill_in('user[email]', with: 'alfred@batcave.com')
```
josemigallas, line 9: IMO, since the form has a new field and it's required, it should be included in the standard "fill form" step. However, the field is not always present so add a check to it. If we want to assert that a certain form has this field I would rather use `I should see field "Current password"` or similar.

```suggestion
  fill_in('user[email]', with: 'alfred@batcave.com')
  fill_in('user[current_password]', with: current_password) if has_css?('user[current_password]')
```

features/step_definitions/developer_portal/admin/account/personal_details_steps.rb
```diff
@@ -0,0 +1,79 @@
...
+  fill_in('user[password]', with: @new_password)
+  fill_in('user[password_confirmation]', with: @new_password)
+end
+
+And "the buyer writes a wrong current password" do
+  fill_in('user[current_password]', with: 'megasecret')
+end
+
+And "the buyer writes a correct current password" do
+  fill_in('user[current_password]', with: 'supersecret')
```
josemigallas, line 23: How do we know `supersecret` is the current password? Wouldn't it be more robust to get it from `@user`?

features/step_definitions/developer_portal/admin/account/personal_details_steps.rb
```diff
@@ -0,0 +1,79 @@
...
+end
+
+And "clicks on update personal details" do
+  click_on 'Update Personal Details'
+end
```
josemigallas, line 28: Completely redundant 😄

features/step_definitions/developer_portal/admin/account/personal_details_steps.rb
```diff
@@ -0,0 +1,79 @@
...
+
+Then "they should be able to edit their personal details" do
+  click_on 'Update Personal Details'
+  assert_flash 'USER WAS SUCCESSFULLY UPDATED.'
+  assert_current_path admin_account_users_path
+  assert has_css?('tr > td:nth-child(2)', text: current_user.username)
+  assert has_css?('tr > td:nth-child(3)', text: current_user.email)
+end
+
+And "password must have changed" do
```
josemigallas, line 46:
```suggestion
And "password has changed" do
```

features/step_definitions/developer_portal/admin/account/personal_details_steps.rb
```diff
@@ -0,0 +1,79 @@
...
+  click_on 'Update Personal Details'
+  assert has_css?('#user_email.error')
+  assert has_css?('.inline-errors', text: 'is too short (minimum is 6 characters) and should look like an email address')
+end
+
+Then "the buyer shouldn't see any reference to password" do
+  assert has_no_css?("#user_current_password")
+  assert has_no_css?("#user_password")
+  assert has_no_css?("#user_password_confirmation")
+end
```
josemigallas, line 65: This step make it a bit unclear when reading the scenario. The reader does not know what "any reference to password" means. I would imagine it means `I should not see "Password"`. I think this could be asserted this way:
```
And I should not see the fields:
      | hidden fields         |
      | Current password      |
      | New password          |
      | Password confirmation |
      | Account extra hidden  |
```

features/step_definitions/developer_portal/admin/account/personal_details_steps.rb
```diff
@@ -0,0 +1,79 @@
...
+
+When "the buyer edits their custom personal details" do
+  fill_in('user[first_name]', with: 'Alfred')
```
josemigallas, line 68: This is not a custom field, is it?

test/functional/developer_portal/admin/account/personal_details_controller_test.rb
```diff
@@ -49,4 +49,18 @@ def setup
 
     assert_response :success
   end
+
+  test 'put update should succeed with current password' do
```
josemigallas, line 53:
```suggestion
  test 'update should succeed with current password' do
```

features/developer_portal/admin/account/personal_details_sso.feature
```diff
@@ -0,0 +1,16 @@
+Feature: Dev Portal Buyer Personal Details
```
josemigallas, line 1: It has the same name as the other feature. Maybe:
```suggestion
Feature: Dev Portal Buyer Personal Details SSO
```

Also add some clarification in the feature description, like.. what SSO means in this context hehe.

josemigallas (changes requested):
LGTM in general, I commented "nice to have" improvements but it's mainly my opinion.

One thing though, ti see functional tests for the case when current_password is required but there is a feature (personal_details_sso) that does not require it. I think it's missing some functional test for that case.

jlledom (dismissed):

jlledom (commented):
It looks good to me, but I have some comments:

1. I asked before whether we should ask for the current password when the user updates a field other than the password, it seems we shouldn't, according to the PR description: `"The password update form on the developer portal doesn't require entering the previous password thus representing a vulnerability threat."` It doesn't mention any other fields than the password ones.
2. This changes break all old templates not containing the `current_password` field. For those templates, when the user tries to change any field, no only the password ones, they will receive an error `"Current password is incorrect."` for a field that doesn't exist in the form, that's confusing, and they won't have any other way to solve it than contacting support. I think we should at least add some help or description in the error message, for instance a link the users can click and learn what's the problem and how to add the new field to their templates.

josemigallas:
@lvillen Please add some screenshots

josemigallas:
> 1. I asked before whether we should ask for the current password when the user updates a field other than the password, it seems we shouldn't, according to the PR description: `"The password update form on the developer portal doesn't require entering the previous password thus representing a vulnerability threat."` It doesn't mention any other fields than the password ones.

You're right. The mockup is actually a bit misleading in that sense, should the new field be located alongside the other password fields?

> 2. This changes break all old templates not containing the `current_password` field. For those templates, when the user tries to change any field, no only the password ones, they will receive an error `"Current password is incorrect."` for a field that doesn't exist in the form, that's confusing, and they won't have any other way to solve it than contacting support. I think we should at least add some help or description in the error message, for instance a link the users can click and learn what's the problem and how to add the new field to their templates.

lib/developer_portal/app/controllers/developer_portal/admin/account/personal_details_controller.rb
```diff
@@ -10,25 +10,21 @@ class DeveloperPortal::Admin::Account::PersonalDetailsController < ::DeveloperPo
...
+      resource.kill_user_sessions(user_session) if resource.just_changed_password?
 
-      failure.html do
-        assign_user_drop
-        render :action => 'show'
-      end
+      redirect_to admin_account_users_path, notice: 'User was successfully updated.'
+    else
+      assign_user_drop
+      render :action => 'show'
```
mayorova, line 27:
```suggestion
      render action: :show
```

lib/developer_portal/app/controllers/developer_portal/admin/account/personal_details_controller.rb
```diff
@@ -49,16 +45,17 @@ def deny_unless_can_update
...
+                                   resource.special_fields +
+                                   resource.defined_fields.map(&:name))
   end
 
+  def verify_current_password
+    return unless current_user.using_password?
+    return if current_user.authenticated?(user_params[:current_password])
+
+    flash.now[:error] = 'Current password is incorrect.'
+    resource.errors.add(:current_password, 'Current password is incorrect.')
```
mayorova, line 59:
```suggestion
    resource.errors.add(:base, :current_password_incorrect)
    flash.now[:error] = resource.errors.full_messages.to_sentence
```


This requires also adding this to `en.yml` on line `:935`

```
          current_password_incorrect: "Current password is incorrect."
```

(`activerecord.errors.models.user.current_password_incorrect`)

mayorova: But this can be done in another PR, this should be transparent for the user.

features/step_definitions/provider_steps.rb
```diff
@@ -247,8 +247,7 @@ def fill_and_submit_form_to_create_tenant(username:)
 end
 
 When(/I authenticate by Oauth2$/) do
-  # it works for Oauth2, which is for what is being used. In case it wants to be used to Auth0, it needs the state param
-  visit "/auth/#{@authentication_provider.system_name}/callback"
+  try_login_with_sso
```
josemigallas, line 250: 👏🏻

lib/developer_portal/app/controllers/developer_portal/admin/account/personal_details_controller.rb
josemigallas: Assuming `resource` is an instance of `User` you can simply add an entry here:
https://github.com/3scale/porta/blob/master/config/locales/en.yml#L976-L979

With a keyword `invalid` I guess, like it's done here: https://github.com/3scale/porta/pull/3533/files#diff-44438ce218f5287c58d0017f965d888715635d94280669896f75841fbd7b4cd7R922

You will need to run this instead:
```suggestion
    resource.errors.add(:current_password, :invalid)
```

See [ActiveModel/Errors.html](https://api.rubyonrails.org/classes/ActiveModel/Errors.html#method-i-add) for reference.

josemigallas (dismissed):

lib/developer_portal/app/controllers/developer_portal/admin/account/personal_details_controller.rb
mayorova: @josemigallas I've been playing around with this, but it doesn't work like this in this specific case.
The reason is that `:current_password` is not actually a real ActiveRecord attribute on the User model. Instead, this name (`current_password`) is used on the form - both for input, and to visualize the errors.

I will add an alternative suggestion :wink:

lib/developer_portal/app/controllers/developer_portal/admin/account/personal_details_controller.rb
mayorova, line 59:
```suggestion
    resource.errors.add(:current_password, 'is incorrect')
    flash.now[:error] = resource.errors.full_messages.to_sentence
```

This outcome of this is slightly different (see screenshots below), but I think it is actually more standard - showing only a short error message on the input hint (which does not duplicate the field name), and then show the full message on the flash.

While the current templates only check for the specific errors (e.g. for `password` and `current_password` fields), I think we need to make sure that `resource.errors.full_messages.to_sentence` actually returns something meaningful. But with the current approach it would return `"Current password Current password is incorrect."`.

The suggested option still doesn't use I18n, but I think in this case it would have to be:
```
resource.errors.add(:current_password, t('activerecord.errors.models.user.current_password_incorrect'))
```

and adding `current_password_incorrect: 'is incorrect'` on the corresponding path of `en.yml`, because `:current_password` is not an actual field of User, so we can't treat it as a normal "attribute error".
This is not too elegant, but maybe could be done also.

And the screenshots... 

**Before:**
Screenshot from 2023-10-02 15-49-07

**After:**
Screenshot from 2023-10-02 15-48-37

As I said before, the "After" style is more similar to other fields' validation errors:

Screenshot from 2023-10-02 15-47-27
Screenshot from 2023-10-02 15-47-46

mayorova (approved):

PR #3206, opened 2023-02-09
🧠🌴 Remove dependency from braintree-web CDN

josemigallas:

Remove dependency from `braintree-web` sdk loaded via CDN by moving the script into a TS file handled by webpack.

Additionally there are some minor improvements, namely:
- Disable the submit button once click to avoid multiple form submissions.
- Move css to a scss file (some typos have been spotted).
- Move hosted fields styles and options to a separate utils file, which will be reused by same form in dev portal.
- Show potential errors in console to help future debugging.

Possible scenarios:
- Scenarios 1: hosted fields failure
braintree-error

- Scenario 2: wrong client token
- Scenario 3: wrong credit card
- Scenario 4: braintree error after submission

[THREESCALE-9152: Remove double dependency from braintree-web SDK](https://issues.redhat.com/browse/THREESCALE-9152)

**Verification steps** 

You can verify it still works by going to `Account Settings > Billing > Payment Details > edit`
(AKA `/p/admin/account/braintree_blue/edit`)

Then use any of the Braintree test credit cards, such as: `4000000000001000` with any valid exp. date and CVV.

> NOTE: you will need to have your gateway configured in master (see test credentials in Bitwarden).


lvillen (approved):

app/javascript/packs/braintree_provider_form.ts
```diff
@@ -0,0 +1,51 @@
...
+    throw new Error('Required elements not found')
+  }
+
+  const clientToken = document.getElementById('braintree_data')?.dataset.clientToken
+
+  if (!clientToken) {
+    throw new Error('Braintree token not found')
+  }
+
+  void createHostedFields(clientToken)
```
jlledom, line 25: I'm know very little of Typescript, what does this `void` here?

jlledom (dismissed):
I tried locally and seems to work fine

app/javascript/packs/braintree_provider_form.ts
josemigallas: Ah, thanks for pointing it out. It's actually a javascript operator and it is not anymore needed here. It is a way to mark a promise as intentionally not awaited. But in this case it has `then` and `catch` so _it is_ used 😄 , ergo no need for void.

PR #3210, opened 2023-02-14
Fix Oracle version on Docs

thalesmiguel:

As pointed out in [here](https://github.com/3scale/porta/pull/3195#issuecomment-1429499513), Porta currently supports Oracle 19c but we have links related with Oracle 21c in the Docs.
This PR updates those links.

none


mayorova (approved):
Thanks @thalesmiguel !

PR #3217, opened 2023-02-21
graceful unicorn termination in container

akostadinov:
k8s/Openshift sends SIGTERM to containers when terminating them.

This installs same handler for SIGTERM as it is set for SIGQUIT to
make unicorn terminate gracefully instead of killing workers with
active connections.

### DIFF
--- a/config/docker/unicorn.rb
+++ b/config/docker/unicorn.rb
@@ -60,6 +60,14 @@ before_fork do |server, _worker|
       # someone else did our job for us
     end
   end
+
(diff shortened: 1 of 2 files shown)


jlledom (approved):
I've spent some time trying to figure out how to get the current handler for a signal without overriding it, but it doesn't seem to be an easy task, only doable by using a gem, which I don't think is worth it. So approved

mayorova (approved):

PR #3222, opened 2023-02-23
sphinx clean indices

akostadinov:
Need more investigation how it works and how exactly to apply it with our custom background jobs indexing (if we keep that to begin with) but the basic idea is described in pat/thinking-sphinx#1192

So basically this is all to make the PR work. The only thing needed is to wipe all indices of sunning sphinx though. Which should happen automatically at present. Except on SaaS.


mayorova (approved):
Nice one :+1:

PR #3226, opened 2023-02-28
Added Audited to Proxy Class

RobGri001:
Hello all,

we like to extend the proxy class with an audited annotation. So we changed the proxy.rb File. 
This caused in an exception, as we assigned a policy. Please have a look at the screenshot.
This agreed with Alexander Kostatinov for help with the analyses.  

Best Regards

Bildschirmfoto 2023-02-28 um 14 17 28

More Informations: https://issues.redhat.com/browse/THREESCALE-6853

### DIFF
--- a/app/controllers/provider/sessions_controller.rb
+++ b/app/controllers/provider/sessions_controller.rb
@@ -10,6 +10,11 @@ class Provider::SessionsController < FrontendController
   before_action :redirect_if_logged_in, only: %i[new]
   before_action :redirect_to_enforced_sso, only: %i[new]
 
+  def initialize
+    super
+    @auditlog = AuditLogService.new
+  end
+
   def new
     @session = Session.new
     @authentication_providers = published_authentication_providers
@@ -18,13 +23,15 @@ class Provider::SessionsController < FrontendController
   def create
     session_return_to
     logout_keeping_session!
-
     @user, strategy = authenticate_user
 
     if @user
       self.current_user = @user
       create_user_session!(strategy.authentication_provider_id)
       flash[:notice] = 'Signed in successfully'
+
+      @auditlog.log("Signed in: #{self.current_user.username} #{self.current_user.id} #{self.current_user.first_name} #{self.current_user.last_name}")
+
       redirect_back_or_default provider_admin_path
     else
       @session = Session.new
@@ -43,6 +50,9 @@ class Provider::SessionsController < FrontendController
     user = current_user
     logout_killing_session!
     destroy_user_session!
+
+    @auditlog.log("destroy_user_session: #{user.username} #{user.id} #{user.first_name} #{user.last_name}")
+
     if @provider.partner? && (logout_url = @provider.partner.logout_url)
       redirect_to logout_url + {user_id: user.id, provider_id: @provider.id}.to_query
     else
@@ -75,7 +85,7 @@ class Provider::SessionsController < FrontendController
                sso_params
              else
                request.post? ? auth_params : sso_params
-    end
+             end
 
     user = strategy.authenticate(params)
 
--- a/app/controllers/provider/sessions_controller.rb
+++ b/app/controllers/provider/sessions_controller.rb
@@ -10,10 +10,6 @@ class Provider::SessionsController < FrontendController
   before_action :redirect_if_logged_in, only: %i[new]
   before_action :redirect_to_enforced_sso, only: %i[new]
 
-  def initialize
-    super
-    @auditlog = AuditLogService.new
-  end
 
   def new
     @session = Session.new
@@ -30,7 +26,7 @@ class Provider::SessionsController < FrontendController
       create_user_session!(strategy.authentication_provider_id)
       flash[:notice] = 'Signed in successfully'
 
-      @auditlog.call("Signed in: #{self.current_user.username} #{self.current_user.id} #{self.current_user.first_name} #{self.current_user.last_name}")
+      AuditLogService.call("Signed in: #{self.current_user.username} #{self.current_user.id} #{self.current_user.first_name} #{self.current_user.last_name}")
 
       redirect_back_or_default provider_admin_path
     else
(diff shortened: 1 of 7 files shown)


app/services/audit_log_service.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class AuditLogService
+
+  def log(message)
+    Rails.logger.info "[AUDIT]: #{message}"
+  end
+end
```
akostadinov, line 8: This looks good. Please follow a service class pattern as described here:
https://github.com/selleo/pattern#service

You basically need to `include Callable` and then log like `AuditLogService.call(my_message)`.

app/models/proxy.rb
```diff
@@ -64,16 +64,16 @@ class Proxy < ApplicationRecord # rubocop:disable Metrics/ClassLength
...
 
   validates :auth_app_id, :auth_app_key, :auth_user_key, :secret_token,
-                      format: { with: APP_OR_USER_KEY, allow_nil: false }
+            format: { with: APP_OR_USER_KEY, allow_nil: false }
 
   validates :error_headers_auth_failed,
-                      :error_headers_auth_missing, :error_headers_no_match,
-                      :error_auth_failed, :error_auth_missing, :error_no_match,
-                      :error_headers_limits_exceeded, :error_limits_exceeded,
-                      format: { with: HTTP_HEADER }
```
akostadinov, line 76: please avoid just formatting changes even if formatting is not ideal to make review easier and git blame easier to follow

app/models/proxy.rb
```diff
@@ -114,6 +114,8 @@ class Proxy < ApplicationRecord # rubocop:disable Metrics/ClassLength
   delegate :backend_apis, :backend_api_configs, to: :service
   delegate :scheduled_for_deletion?, to: :account, allow_nil: true
 
+  audited allow_mass_assignment: true
```
akostadinov, line 117: I assume this is where trouble comes from. It is probably because we are using a custom type `PoliciesConfig` and this somehow confuses audoted. I need to investigate that tomorrow.

akostadinov (commented):
Added some comments. Need some time to investigate how to help audited work.
Thank you!

app/models/proxy.rb
akostadinov: @RobGri001 , still not sure why this happens in the first place. Seems to be the way rails or psych does serialization or how audited uses that. But there is a workaround that should get you going for the time being.

```diff
--- a/app/models/attributes/policies_config.rb
+++ b/app/models/attributes/policies_config.rb
@@ -13,3 +13,5 @@ class Attributes::PoliciesConfig < ActiveRecord::Type::Text
     new_value != cast(raw_old_value)
   end
 end
+
+PoliciesConfig = Attributes::PoliciesConfig
```

Add this line and it should stop failing. I'll check later whether a better fix will be available.

akostadinov: Issue is easy to reproduce and is IMO some bug in `Psych`
```rb
1] pry(main)> pc = Proxy.last.policies_config
=> #<Proxy::PoliciesConfig:0x000000000d074358
 @policies_config=
  [#<Proxy::PolicyConfig:0x000000000d07be50
    @configuration={},
    @enabled=true,
    @name="apicast",
    @version="builtin">,
   #<Proxy::PolicyConfig:0x000000000d07be00
    @configuration={},
    @enabled=true,
    @name="3scale_referrer",
    @version="builtin">]>
[2] pry(main)> ser = pc.to_yaml
=> "--- !ruby/object:PoliciesConfig\npolicies_config:\n- !ruby/object:Proxy::PolicyConfig\n  name: apicast\n  version: builtin\n  configuration: {}\n  enabled: true\n- !ruby/object:Proxy::PolicyConfig\n  name: 3scale_referrer\n  version: builtin\n  configuration: {}\n  enabled: true\n"
[3] pry(main)> YAML.load ser
ArgumentError: undefined class/module PoliciesConfig
from /home/avalon/.asdf/installs/ruby/2.6.7/lib/ruby/2.6.0/psych/class_loader.rb:54:in `path2class'
Caused by ArgumentError: undefined class/module Struct::PoliciesConfig
from /home/avalon/.asdf/installs/ruby/2.6.7/lib/ruby/2.6.0/psych/class_loader.rb:54:in `path2class'
```

This also shows that my above workaround is not ideal. But need more to investigate.

akostadinov: I think I see where the issue comes from:

https://github.com/3scale/porta/blob/962e945968721331c5f1d5d9211c95286537067e/app/models/proxy.rb#L565

I have zero idea why this was needed. But it obviously is used by Psych as class name
```ruby
[10] pry(main)> pc.class.name
=> "PoliciesConfig"
```

So one option would be to remove this method. But then something in presentation may break. We can see if any tests would fail. A safer approach might be to extract this class not to be a sub-class of `Proxy` so this method will not be needed anymore as the class will be at the root layer anyway.

app/services/audit_log_service.rb
```diff
@@ -0,0 +1,7 @@
+class AuditLogService
+  include Callable
+
+  def self.call(message)
```
akostadinov, line 4:
```suggestion
  def call(message)
```

The `Callable` concern should handle this transparently.

app/controllers/provider/sessions_controller.rb
```diff
@@ -18,13 +19,15 @@ def new
...
     logout_keeping_session!
-
     @user, strategy = authenticate_user
 
     if @user
       self.current_user = @user
       create_user_session!(strategy.authentication_provider_id)
       flash[:notice] = 'Signed in successfully'
+
+      AuditLogService.call("Signed in: #{self.current_user.username} #{self.current_user.id} #{self.current_user.first_name} #{self.current_user.last_name}")
```
akostadinov, line 29: I wonder if it makes sense to log both - successful and unsuccessful login attempts.

akostadinov:
Migrating changes from here to #3342

PR #3233, opened 2023-03-03
dependabot.yml

akostadinov:
Enable also workflow step updates. Easier to customize going forward.


jlledom:
What does this configuration do?

akostadinov:
@jlledom , I guess PR description was too short. 

Presently only security updates are provided. I wanted to enable git workflow step updates. The presented config should enable that. While maintaining only security updates for the rest. At some point we may also enable version updates but for now lets keep it under control.

jlledom (approved):

PR #3235, opened 2023-03-03
CMS API: Implement API Contract for JSON responses

thalesmiguel:

This PR updates the JSON format responses for the CMS API so that they follow the same structure and return the same data as the XML responses, updated recently in https://github.com/3scale/porta/pull/3193.

Fixes [THREESCALE-8993](https://issues.redhat.com/browse/THREESCALE-8993).

**Verification steps** 

Call the following endpoints to check they match the responses in [this document](https://docs.google.com/document/d/14Lpu5T-ktDVUNl9SINtd7WRdew1x0f5qDzFngThZyV8/edit):
```
GET /admin/api/cms/sections/:section_id/files.json
GET /admin/api/cms/sections.json
GET /admin/api/cms/sections/:id.json
GET /admin/api/cms/files.json
GET /admin/api/cms/files/:id.json
GET /admin/api/cms/templates.json
GET /admin/api/cms/templates/:id.json
```

Be happy

### DIFF
--- a/app/models/cms/builtin.rb
+++ b/app/models/cms/builtin.rb
@@ -150,8 +150,8 @@ class CMS::Builtin < CMS::BasePage
         x.system_name system_name
         x.layout_id layout_id
         unless options[:short]
-          x.draft { |node| node.cdata draft }
-          x.published { |node| node.cdata published }
+          x.draft { |node| draft.present? ? node.cdata(draft) : nil }
+          x.published { |node| published.present? ? node.cdata(published) : nil }
         end
       end
 
--- a/app/models/cms/builtin.rb
+++ b/app/models/cms/builtin.rb
@@ -150,8 +150,8 @@ class CMS::Builtin < CMS::BasePage
         x.system_name system_name
         x.layout_id layout_id
         unless options[:short]
-          x.draft { |node| draft.present? ? node.cdata(draft) : nil }
-          x.published { |node| published.present? ? node.cdata(published) : nil }
+          x.draft { |node| node.cdata(draft) unless draft.nil? }
+          x.published { |node| node.cdata(published) unless published.nil? }
         end
       end
 
--- a/app/models/cms/builtin.rb
+++ b/app/models/cms/builtin.rb
@@ -150,8 +150,8 @@ class CMS::Builtin < CMS::BasePage
         x.system_name system_name
         x.layout_id layout_id
         unless options[:short]
-          x.draft { |node| node.cdata(draft) unless draft.nil? }
-          x.published { |node| node.cdata(published) unless published.nil? }
+          x.draft { |node| node.cdata(draft) if draft }
+          x.published { |node| node.cdata(published) if published }
         end
       end
 
--- a/app/representers/cms/file_representer.rb
+++ b/app/representers/cms/file_representer.rb
@@ -1,18 +1,22 @@
-module CMS::FileRepresenter
-  include ThreeScale::JSONRepresenter
+# frozen_string_literal: true
 
-  wraps_resource
+module CMS
+  module FileRepresenter
+    include ThreeScale::JSONRepresenter
 
-  property :id
-  property :created_at
-  property :updated_at
-  property :title
-  property :path
-  property :url
-  property :section_id
-  property :tag_list
+    wraps_resource
 
-  link :section do
-    admin_api_cms_section_path section
+    with_options(unless: :new_record?) do
+      property :id
+      property :created_at
+      property :updated_at
+    end
+
+    property :section_id
+    property :path
+    property :downloadable, getter: ->(*) { downloadable? }
+    property :url
+    property :title
+    property :content_type, getter: ->(*) { attachment.content_type }
   end
 end
--- a/app/representers/cms/file_representer.rb
+++ b/app/representers/cms/file_representer.rb
@@ -4,7 +4,7 @@ module CMS
   module FileRepresenter
     include ThreeScale::JSONRepresenter
 
-    wraps_resource
+    wraps_resource ->(*) { self.class.data_tag }
 
     with_options(unless: :new_record?) do
       property :id
--- a/app/representers/cms/layout_representer.rb
+++ b/app/representers/cms/layout_representer.rb
@@ -3,7 +3,7 @@
 module CMS::LayoutRepresenter
   include ThreeScale::JSONRepresenter
 
-  wraps_resource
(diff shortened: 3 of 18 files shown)


app/representers/cms/file_representer.rb
```diff
@@ -1,18 +1,22 @@
...
 
-  property :id
-  property :created_at
-  property :updated_at
-  property :title
-  property :path
-  property :url
-  property :section_id
-  property :tag_list
+    wraps_resource
```
jlledom, line 7: I think you should use the same formula in all resources:

`wraps_resource ->(*) { self.class.data_tag }`

So the given `data_tag` is always used

app/representers/cms/layout_representer.rb
```diff
@@ -1,20 +1,19 @@
...
   property :system_name
-  property :content_type
-  property :handler, render_nil: true
+  property :liquid_enabled
 
-  with_options(if: ->(options) { options[:short] == false }) do |p|
-    p.property :published, render_nil: true
-    p.property :draft, render_nil: true
+  with_options(if: ->(options) { !options[:short] }) do
+    property :draft, render_nil: true
```
jlledom, line 16: This prints `null` when there's no draft. To match the XML response, it should return empty, in JSON that should be `""`

In the review comment I'm attaching a list of fields that aren't returned when they are empty, and they should be returned empty.

app/representers/cms/page_representer.rb
```diff
@@ -1,32 +1,28 @@
...
-
-  with_options(if: ->(options) { options[:short] == false }) do |p|
-    p.property :draft, render_nil: true
-    p.property :published, render_nil: true
-  end
-
+  property :created_at
+  property :updated_at
+  property :title
   property :system_name
```
jlledom, line 12: For some reason, `system_name` doesn't appear when it's empty, it should appear as `""`.

After https://github.com/3scale/porta/pull/3163, `system_name` is not going to be empty anymore. But I think this should be fixed anyway, it's still not clear what are we going to do with that PR.

app/representers/cms/page_representer.rb
```diff
@@ -1,32 +1,28 @@
...
-  property :liquid_enabled
+  property :layout_id
```
jlledom, line 13: Doesn't appear when empty. The field should be returned, but empty. In JSON the way to return emptiness is `""`, right?

app/representers/cms/page_representer.rb
```diff
@@ -1,32 +1,28 @@
...
 
   with_options(if: ->(*) { is_a?(CMS::Page) }) do |p|
-    # this needs a getter
+    p.property :section_id
     p.property :path, if: ->(*) { respond_to?(:path) }
-    p.property :title
-    p.property :hidden
-    p.property :layout
     p.property :content_type
+    p.property :liquid_enabled
```
jlledom, line 19: The property is not returned when `false`. I think this could work (haven't verified myself)

```suggestion
    p.property :liquid_enabled?
```

app/representers/cms/section_representer.rb
```diff
@@ -1,15 +1,21 @@
...
+    with_options(unless: :new_record?) do
+      property :id
+      property :created_at
+      property :updated_at
+    end
+
+    property :title
+    property :system_name
+    property :public
+    property :parent_id
```
jlledom, line 18: Not being returned when empty.

jlledom (changes requested):
Some fields aren't returned in some circumstances:

- section: `parent_id` is empty in XML but non-existent in JSON
- page:
	- short:
		- `system_name`, `layout_id` don't appear when empty
		- `liquid_enabled` doesn't appear when `false`
	- long
		- `draft`/`published`: are `null` when empty, should be `""`
- builtin_page:
	- short:
		- `system_name`, `layout_id` don't appear when empty
	- long
		- `draft`/`published`: are `null` when empty, should be `""`
- partial / builtin_partial:
	- `draft`/`published`: are `null` when empty, should be `""`
- layout:
	- `draft`/`published`: are `null` when empty, should be `""`

Thanks for adding tests!

app/representers/cms/file_representer.rb
thalesmiguel: This is only needed for the cases where the tag may change. Like for `section` and `builtin_section` that share the same representer.
For this case, data will always be wrapped by `file`.

thalesmiguel:
@jlledom I added a bunch of `render_nil` so we don't miss `nil` valued keys :v:
Also, from our quick talk, I changed the XML response so it returns null when we don't have data on the `draft` and `published` attributes.
Could you please double check?

app/representers/cms/file_representer.rb
jlledom: Yes but it's wrapped by `file` automatically probably because that's the class name, I don't know where that `file` comes from. The point of the `data_tag` thing is for us to explicitly define the tag names in one place and have it used for JSON and XML formats, if we decide to modify `data_tag` in the future, it will have effect on the XML but not in the JSON.

thalesmiguel: Indeed it comes from the class name.
Added the explicit data tag wrap in e3d778b.

app/models/cms/builtin.rb
```diff
@@ -150,8 +150,8 @@ def to_xml(options = {})
         x.system_name system_name
         x.layout_id layout_id
         unless options[:short]
-          x.draft { |node| node.cdata draft }
-          x.published { |node| node.cdata published }
+          x.draft { |node| draft.present? ? node.cdata(draft) : nil }
```
jlledom, line 153: I don't think this works, because `present?` returns false for `""` and `" "`, so the XML response doesn't have a way to distinguish between `""`, `" "` and `nil`. While JSON does.

And now I attach the best cheatsheet ever. You're welcome.
rails_blank cheatsheet

thalesmiguel: Aaaaand there are no validations for `draft` and `published` that would make `""` and `" "` impossible 😓 
Nil check it is... Updated in 0a2d607.

thalesmiguel: > And now I attach the best cheatsheet ever. You're welcome.

In the battle for the best cheatsheet ever, I'd say this one saved me a couple times in the past:
Visual_SQL_JOINS_V2

jlledom: That's for weak people, real programmers use big text file instead of a DB engine.

jlledom (approved):

PR #3238, opened 2023-03-03
query count test for audited

akostadinov:


test/integration/audited_hacks_async_test.rb
```diff
@@ -81,6 +82,19 @@ class AuditedHacksAsyncTest < ActionDispatch::IntegrationTest
...
 
+  # this will hopefully remind us to disable the :touch callback in audited 5.3
+  test "async audit causes no synchronous DB queries" do
+    # just a model without noise producing callbacks
+    line_item = FactoryBot.build(:line_item)
+
+    assert_number_of_queries(2, matching: /./) do
+      LineItem.with_auditing do
+        line_item.save!
+        line_item.update!(quantity: 5)
```
nidhi-soni1104, line 93: why save and update both?

can we remove line_item.save! or its added for a purpose

jlledom, line 93: Don't understand what is this PR about but, AFAIK, N+1 problems should occur when reading from the DB, not when writing. And only when there are many records involved, not only one like in this test. How could a `save` or `update` operation over a single resource cause a N+1?

akostadinov: Nobody said N+1. The test name is `async audit causes no synchronous DB queries`. 

The test basically performs 2 operations that are audited and checks only the expected number of queries are performed.

jlledom (approved):

PR #3244, opened 2023-03-06
dependabot - automate licenses report

akostadinov:
This action should automatically update licenses report when dependabot files a pull request.

Note: It is far fetched that this will work as it is, but since actions doesn't allow us to test this before committing, I would rather commit and correct when failures are seen.

Update: Actually it is more complicated. It requires sources and bundle install and yarn... I'm wondering whether we should better just remove the report and generate it in the production build.


akostadinov:
closed in favor of #3245

PR #3248, opened 2023-03-09
CMS: Unify system names

jlledom:

This is a simpler approach to https://github.com/3scale/porta/pull/3163.

In that PR I added validations to all CMS models so `system_name`:

1. Is always present.
2. Meets the format `%r{\A\w[\w\-/_]+\z}`.

The problem is that there are some sections and pages in the DB that doesn't meet the format, so adding validations would require a backfill migration to fix those two tables.

In this PR, I applied the minimal changes possible to ensure those two conditions above are meet **for all new records**, but it doesn't add the validations, so the existing records not meeting the conditions are allowed in order to avoid breaking anything.

Part of [THREESCALE-8991](https://issues.redhat.com/browse/THREESCALE-8991)

**Verification steps** 

- Creating a new section which title contains spaces should generate a system name without spaces.
- Creating a Pages without a system name should generate one automatically, with the proper format.
- Creating a new tenant should generate valid system names for all the default Developer Portal resources.

There are changes only in `CMS::Page` and `CMS::Section`, all other models were already fixed, this tables summarizes the state of system names before this PR.

|Type|Required|Format enforced|Value|
|--|--|--|--|
|Section|Yes|No|Generated from `title` parameter|
|Page|No|Yes|From `system_name` parameter|
|Builtin Page|Yes|Yes|From `system_name` parameter|
|Layout|Yes|Yes|From `system_name` parameter|
|Partial|Yes|Yes|From `system_name` parameter|
|Builtin Partial|Yes|Yes|From `system_name` parameter|
|File|No|No|There's no  `system_name`  column in the table|

As the table shows, `system_name` is not applicable for **Files**; while **Builtin Page**, **Page**, **Layout**, **Partial** and **Builtin Partial** already meet the two conditions. So only changes in **Page** and **Section** are required.

### DIFF
--- a/test/unit/cms/page_test.rb
+++ b/test/unit/cms/page_test.rb
@@ -97,4 +97,48 @@ class PageTest < ActiveSupport::TestCase
     page.expects(:tag_list).returns([]).once
     page.as_json(include: [:tag_list])
   end
+
+  test 'generates a system_name with proper format from title when creating' do
+    page = CMS::Page.new.tap do |p|
+      p.title = 'New Page'
+      p.provider = @provider
+      p.section = @provider.sections.root
+      p.path = '/new-page'
+    end
+
+    page.save!
+
+    assert_equal 'new-page', page.reload.system_name
+  end
+
+  test 'generates a system_name with proper format from title when updating' do
+    page = CMS::Page.new.tap do |p|
+      p.title = 'New Page'
+      p.system_name = 'sysname-1'
+      p.provider = @provider
+      p.section = @provider.sections.root
+      p.path = '/new-page'
+    end
+    page.save!
+
+    page.system_name = ''
+    page.save!
+
+    assert_equal 'new-page', page.reload.system_name
+  end
+
+  test "doesn't generate a system_name when one is given" do
+    sysname = 'sysname-1'
+    page = CMS::Page.new.tap do |p|
+      p.title = 'New Page'
+      p.system_name = sysname
+      p.provider = @provider
+      p.section = @provider.sections.root
+      p.path = '/new-page'
+    end
+
+    page.save!
+
+    assert_equal sysname, page.reload.system_name
+  end
 end
--- a/test/unit/cms/page_test.rb
+++ b/test/unit/cms/page_test.rb
@@ -141,4 +141,16 @@ class PageTest < ActiveSupport::TestCase
 
     assert_equal sysname, page.reload.system_name
   end
+
+  test "raises an error when an invalid system_name is given" do
+    page = CMS::Page.new.tap do |p|
+      p.title = 'New Page'
+      p.system_name = 'sysname 1'
+      p.provider = @provider
+      p.section = @provider.sections.root
+      p.path = '/new-page'
+    end
+
+    assert_raise(ActiveRecord::RecordInvalid) { page.save! }
+  end
 end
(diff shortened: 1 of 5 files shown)


lvillen (approved):

josemigallas (approved):

josemigallas (changes requested):
It's missing some tests

akostadinov:
Looks good to me. Probably it already has some tests for `create`, right? But it wouldn't for `update` though. Is `system_name` mutable? Do we need the `on :update`?

jlledom:
> Looks good to me. Probably it already has some tests for `create`, right? But it wouldn't for `update` though. Is `system_name` mutable? Do we need the `on :update`?

There were no tests for `system_name` autofill AFAIK. I've added some. `system_name` is mutable, but it will be autofilled only when it's empty.

- For sections: it should never be empty because it's presence is validated. So `#update` with an empty `system_name` should never happen, but it's a little effort to make the autofill work also for `#update` so I added it, just in case is there any invalid section in the DB or the customer sends and empty `system_name` through the API when updating.
- For pages: `system_name` is not required, so it's good to add the autofill also for `#update`. When some old page without `system_name` is updated, it will get one.

test/unit/cms/page_test.rb
```diff
@@ -97,4 +97,48 @@ def setup
...
+    end
+    page.save!
+
+    page.system_name = ''
+    page.save!
+
+    assert_equal 'new-page', page.reload.system_name
+  end
+
+  test "doesn't generate a system_name when one is given" do
```
josemigallas, line 130: What about if given an invalid system name, such as `'sysname 1'`?

test/unit/cms/page_test.rb
```diff
@@ -97,4 +97,48 @@ def setup
...
+      p.provider = @provider
+      p.section = @provider.sections.root
+      p.path = '/new-page'
+    end
+    page.save!
+
+    page.system_name = ''
+    page.save!
+
+    assert_equal 'new-page', page.reload.system_name
```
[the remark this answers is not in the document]
jlledom: No, since the system name is empty, it generates a new one based on the title

test/unit/cms/page_test.rb
jlledom: It will return a validation error because the `system_name` format for pages is enforced in the parent class `CMS::Template` https://github.com/3scale/porta/blob/1ece547ea9568f4ed9c845ea477dfaa7b1e451c7/app/models/cms/template.rb#L19-L20

test/unit/cms/page_test.rb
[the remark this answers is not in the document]
josemigallas: But it's updating an existing record, already with a system_name. Maybe I got it wrong, but updating with an empty string means either erase the field or leave it alone.

test/unit/cms/page_test.rb
josemigallas: Yes, I mean what about writing a test for that scenario.

test/unit/cms/page_test.rb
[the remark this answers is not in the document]
jlledom: It sets `system_name` to an empty string, yes, that's why it works.

test/unit/cms/page_test.rb
jlledom: Done

PR #3256, opened 2023-03-16
CMS API: Forbid extra params

jlledom:

Before this PR, when a client sent a parameter not permitted but also not forbidden, it was ignored. It's a requisite that the API returns an error in this case.

Rails' strong parameters provides a [config option](https://guides.rubyonrails.org/configuring.html#config-action-controller-action-on-unpermitted-parameters) to enable this behavior, the problem is that option acts globally, so if we enable it that would problably break some customers implementation for other APIs.

To avoid that, this PR adds a controller method to do the same, but at controller level:

`controller_action_on_unpermitted_parameters :raise`

Another method to whitelist some particular parameters we don't want to raise the exception for is also provided:

`controller_always_permitted_parameters %i[id page per_page]`

These methods are called from `Admin::Api::CMS::BaseController` so extra params will return an error from all CMS API endpoints.

[THREESCALE-9172](https://issues.redhat.com/browse/THREESCALE-9172)

**Verification steps**

Request:

```
curl --request POST \
  --url 'http://provider-admin.3scale.localhost:3000//admin/api/cms/sections?access_token=secret' \
  --data '{
	"title": "Test section from API",
	"system_name": "test-section-api",
	"extra_param": "value"
}'
```

Should return:

`<?xml version="1.0" encoding="UTF-8"?><error>found unpermitted parameter: :extra_param</error>`


app/lib/api_support/params.rb
```diff
@@ -6,32 +6,85 @@ module ApiSupport::Params
   included do
     wrap_parameters format: %i[url_encoded_form json multipart_form]
     before_action :wrap_parameters_check
+    before_action :unpermitted_parameters_check
+    class_attribute :_class_action_on_unpermitted_parameters
+    class_attribute :_class_always_permitted_parameters, default: []
```
thalesmiguel, line 11: Looks to me that this configuration logic looks mature enough to deserve its own separate module.
Maybe a `ApiSupport::PermittedParams` or `ApiSupport::PermittedParamsConfig`.
We could even include it only in `Admin::Api::BaseController` for now, leaving `Master::Api::BaseController` unchanged.
What do you think?

app/controllers/admin/api/cms/base_controller.rb
```diff
@@ -3,6 +3,10 @@ class Admin::Api::CMS::BaseController < Admin::Api::BaseController
   before_action :deny_on_premises_for_master
   self.access_token_scopes = %i[cms account_management]
 
+  class_action_on_unpermitted_parameters :raise
+  class_always_permitted_parameters %i[id page per_page]
```
thalesmiguel, line 7: `class` makes me wonder if those configurations may be used anywhere. I think we could prepend "controller" for better understanding or even remove it altogether so it looks closed to the Rails config option.

```suggestion
  controller_action_on_unpermitted_parameters :raise
  controller_always_permitted_parameters %i[id page per_page]
```
or even

```suggestion
  action_on_unpermitted_parameters :raise
  always_permitted_parameters %i[id page per_page]
```

**Warning:** This comment smells a little too much of _personal preference_ so feel extra free to discard it.

thalesmiguel (commented):
Interesting approach 🎉

app/lib/api_support/params.rb
jlledom: I'd prefer to use a separate module, yes. In fact, I tried, but I found that I needed to call `flat_params`: look at the `unpermitted_keys` method. Besides, `unpermitted_keys` also calls the new method `wrapped_keys` which makes sense when the controller is calling `wrap_parameters`, but not all controllers do that and I don't know if that could cause problems.

Therefore, this module `ApiSupport::Params` is a strong dependency and I probably would need to include it from the new module in the case I create it. I could do it if you think it's a good idea.

One way or another, the logic for both modules will be included on any controller that includes extra params restriction.

app/lib/api_support/params.rb
```diff
@@ -6,32 +6,85 @@ module ApiSupport::Params
...
-      option_name = _wrapper_options[:name]
-      param_name = params[option_name]
-      return if !param_name || param_name.is_a?(ActionController::Parameters)
+    def class_always_permitted_parameters(permitted_params)
+      self._class_always_permitted_parameters = self._class_always_permitted_parameters.dup
+      self._class_always_permitted_parameters.push(*permitted_params).uniq!
+    end
+  end
+
+  protected
```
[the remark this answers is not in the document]
jlledom: Done

akostadinov:
Is this logic mostly copied from rails or is it custom one? If it is mostly mimicking something from rails, better put a reference explaining that.

But do we need all this logic? Why not implement something simpler that will just always raise when unpermitted params are found. And controllers that don't want it, will simply not include it.

About these always permitted, they very much look like `protected_attributes`. Do we need that?

app/lib/api_support/params.rb
thalesmiguel: > I could do it if you think it's a good idea.

I do not 😅

thalesmiguel (approved):
No more comments from my side 🚀

jlledom:
> Is this logic mostly copied from rails or is it custom one? If it is mostly mimicking something from rails, better put a reference explaining that.
> 

Yes, it mimics the Rails behavior, but at controller level instead of globally. I added a link to the docs in the comment.

> But do we need all this logic? Why not implement something simpler that will just always raise when unpermitted params are found. And controllers that don't want it, will simply not include it.
> 

It could be simpler if we stick to always raise instead of allowing the user to select between `:raise` and `:log`. But there's no improvement for user experience, in both cases they will have to add a line to de controller.

A second call to whitelist always-permitted params would be required anyway, since controllers receive params we want to ignore.

> About these always permitted, they very much look like `protected_attributes`. Do we need that?

Yes, we have the same features but using Strong Parmeters and therefore calling the methods from the controller, not the model (like in PA). Different controllers can have different whitelists, that wasn't possible before.

thalesmiguel (approved):
Second approval is still free of charge.

akostadinov:
> It could be simpler if we stick to always raise instead of allowing the user to select between :raise and :log. But there's no improvement for user experience, in both cases they will have to add a line to de controller.

If global rails behavior is to `:log` then why do we need to add this choice?

> A second call to whitelist always-permitted params would be required anyway, since controllers receive params we want to ignore.

I thought that the point was to require strong parameters, why do we need automatic whitelists?

jlledom:
Closing in favor of https://github.com/3scale/porta/pull/3296

PR #3257, opened 2023-03-16
invoice rendering fixes

akostadinov:


josemigallas (approved):

jlledom (dismissed):

akostadinov:
@jlledom , chec last commit, this test is obsoleted because we don't need to escape chars anymore.

jlledom:
> @jlledom , chec last commit, this test is obsoleted because we don't need to escape chars anymore.

I understand, but since we are not escaping anymore, we are in fact vulnerable to XSS attack, aren't we?. I don't really understand how a XSS attack can be done via PDF, I mean, even if you write javascript code in a PDF, it's not going to be executed, so why escaping? I think we should figure that out before deploying.

akostadinov:
@jlledom , we don not parse HTML anymore so tags appear to be irrelevant. Strings appear literally in the PDF. So if somebody names his company `<sCrIpt>muhahahah</sCrIpt>` this is what will show up in the PDF. And other special characters like `&`, `'`, etc.

akostadinov:
So confirmed. In the past we have been using the now discontinued `prawn-format` to parse HTML snippets and inserting them into the PDF. But now we do not use it and HTML tag processing does not take place.

There is an option to enable some limited HTML parsing using the [Text](https://www.rubydoc.info/github/prawnpdf/prawn/Prawn/Text) option `inline_format`. But we do not use it. So escaping HTML tags makes no sense as these strings are presented literally in the output. If we escape, then strings are improperly presented escaped in the readable PDF output which is wrong and is the primary reason of this PR.

Just for educational purposes I enabled `inline_format` and added `<b>` and `<script>` tags to a PDF to see if it would do any harm. And actually no. The `<b>` tag is recognized and output shows bold characters. But the `script` tag is apparently not recognized and appears in output without opening `<`. Like this:

image

While as I think I stated earlier, without this option, tags are presented literally and again, no script injection appears to be possible:

image

In conclusion, I maintain that the changes are safe. Thank you for making me triple check. I'm merging this.

PR #3263, opened 2023-03-20
THREESCALE-8719 The default metric should be a Product's metric

nidhi-soni1104:
to make  default metric should be a Product's metric

https://issues.redhat.com/browse/THREESCALE-8719

**Note:**

-  Steps to reproduce is clearly mentioned in ticket with scenshots


nidhi-soni1104:
closing this PR because it rebased wrongly. I will create a new one

PR #3271, opened 2023-03-22
THREESCALE-8719 The default metric should be a Product's metric

nidhi-soni1104:
to make default metric should be a Product's metric

https://issues.redhat.com/browse/THREESCALE-8719

**Note:**

Steps to reproduce is clearly mentioned in ticket with scenshots


app/views/stats/usage/index.html.slim
```diff
@@ -27,7 +27,7 @@ javascript:
         methodsTableContainer: '.StatsMethodsTable-container',
         csvLinkContainer: '.StatsCSVLink-container',
         usageMetricsUrl: '#{admin_service_stats_usage_path(@service, format: :json)}',
-        hitsMetricName: '#{@metrics.find(&:hits?).system_name}',
+        hitsMetricName: '#{@service.metrics.where(system_name: "hits").last.system_name}',
```
thalesmiguel, line 30: This will perform an extra query to the DB (`@service.metrics`) that can be avoided since the `@metrics` instance variable already holds all the metrics we need for this view.

To avoid the extra query, we could search the values we need from `@metrics` instead. Something like:
```suggestion
        hitsMetricName: "#{@metrics.find { |metric| metric.hits? && metric.owner_type == 'Service' }.system_name}",
```

Maybe even create a `Metric#service?` or `Metric#owner_type_service?` method that checks if `owner_type == 'Service'` so we may do something like:
```suggestion
        hitsMetricName: '#{@metrics.find { |metric| metric.hits? && metric.service? }.system_name}',
```

nidhi-soni1104:
@thalesmiguel I have done the changes suggested

thalesmiguel:
Cool @nidhi-soni1104!
From the looks of the code and what's described on the ticket, this should be enough to fix the issue.
Now, I do have 2 concerns about the current state of the repo's code (NOT your PR) that we could address:

1. I couldn't find any tests related with this code we're changing here. It's a good idea to add a test so this BUG doesn't come back in the future to haunt us;
2. I see we have the exact same code about `hitsMetricName` in [here](https://github.com/3scale/porta/blob/master/app/views/provider/admin/backend_apis/stats/usage/index.html.slim#L30). Any idea if it should also be changed to only talk about `service` metrics?

nidhi-soni1104:
> AS much as I tried, I couldn't reproduce the bug. I can arrange everything to get a list of metrics like the one the issue mentions, this is from my tests:
> 
> ```
> {
>   "metrics": [
>     {
>       "metric": {
>         "id": 10,
>         "system_name": "hits.4",
>         "description": "Number of API hits",
>         "unit": "hit",
>         "created_at": "2023-03-27T15:44:48Z",
>         "updated_at": "2023-03-27T15:44:48Z",
>         "service_id": null,
>         "friendly_name": "Hits",
>         "parent_id": null,
>         "tenant_id": null,
>         "owner_id": 4,
>         "owner_type": "BackendApi"
>       }
>     },
>     {
>       "metric": {
>         "id": 11,
>         "system_name": "hits",
>         "description": "Number of API hits",
>         "unit": "hit",
>         "created_at": "2023-03-27T15:45:12Z",
>         "updated_at": "2023-03-27T15:45:12Z",
>         "service_id": 4,
>         "friendly_name": "Hits",
>         "parent_id": null,
>         "tenant_id": 2,
>         "owner_id": 4,
>         "owner_type": "Service"
>       }
>     },
>     {
>       "metric": {
>         "id": 16,
>         "system_name": "hits_backend.4",
>         "description": "",
>         "unit": "hit",
>         "created_at": "2023-03-27T15:58:33Z",
>         "updated_at": "2023-03-27T15:58:33Z",
>         "service_id": null,
>         "friendly_name": "Hits_backen",
>         "parent_id": null,
>         "tenant_id": null,
>         "owner_id": 4,
>         "owner_type": "BackendApi"
>       }
>     }
>   ],
>   "methods": []
> }
> ```
> 
> So the first metric in the JSON is a Backend one. But when I open the analytics screen, the scond metric (`"system_name": "hits"`) is open by default. I did a quick research and found why:
> 
> https://github.com/3scale/porta/blob/3ee9596ee2a75f6ef79efef168499cdd200281ca/app/javascript/src/Stats/lib/stats.jsx#L33-L37
> 
> The `groupMetrics` function sorts the given JSON, so metrics with `serviceId != null` (those belonging to a service) are placed first. Then we can rely on showing always the first metric as default, both here:
> 
> https://github.com/3scale/porta/blob/3ee9596ee2a75f6ef79efef168499cdd200281ca/app/views/stats/usage/index.html.slim#L30
> 
> and here:
> https://github.com/3scale/porta/blob/3ee9596ee2a75f6ef79efef168499cdd200281ca/app/javascript/src/Stats/lib/metrics_selector.jsx#L56
> 
> About your changes, I don't think they are necessary because it seems to be already fixed. Adding a test, however, is still a good idea, but your cucumber is testing what happens when the service metric comes first, and it should test precisely the opposite: what happens when the backend metric comes first, something like:
> 
> ```
>  And the following metrics:
>       | Metric   | system_name | owner_type    |
>       | Pizza    |      pizza  |  "BackendApi" |
>       | Hits     |      hits   |   "Service"   |
> ```
> 
> @nidhi-soni1104 @thalesmiguel were you able to reproduce this?

I have updated the sequence, but i was able to produce following same steps mention and n matter how many metrics there it gives me the first was BackendApi in my case 

here is the screenshot before fix
image


Here is the screenshot after fix 
image

nidhi-soni1104:
@jlledom Can you please try creating new products and new back-ends and then reproduce this issue

jlledom:
> @jlledom Can you please try creating new products and new back-ends and then reproduce this issue

I already tried that, could you please paste here the JSON that is received when loading that screen?

nidhi-soni1104:
> > @jlledom Can you please try creating new products and new back-ends and then reproduce this issue
> 
> I already tried that, could you please paste here the JSON that is received when loading that screen?

image

jlledom (approved):
OK I finally was able to reproduce it. My problem was that the default metric is part of the URL (`selectedMetricName` param) and I was refreshing the screen with the same URL after switching to `master`. To reproduce it, you need to remove the params from the URL, then press enter, at least in Firefox.

What I said in my previous comment is wrong because that Javascript code is only used to generate the selector dropdown. The default metric is a part of the React component state, and that's initialized from the metrics given by the controller, which aren't ordered.

The fix looks good to me.

jlledom (changes requested):

features/api/overview.feature
```diff
@@ -14,3 +14,10 @@ Feature: Product > Overview page
       | Backend 1 |
       | Backend 2 |
       | Backend 3 |
+
+  @javascript
+  Scenario: Show product metric as default
```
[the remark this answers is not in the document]
nidhi-soni1104: resolving because removed this and added a new test

features/api/overview.feature
```diff
@@ -14,3 +14,10 @@ Feature: Product > Overview page
...
+      When I go to the overview page of product "My Product"
+      And I follow "Analytics"
+      And I follow "Traffic"
+    Then I should see "Hits"
```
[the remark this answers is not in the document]
nidhi-soni1104: @jlledom  PR is ready to review with updated test

I have written test similar as the steps we are trying to reproduce .
As soon we as  we create product and backend and connect them with each other
we get two  metric by default having system_name "hits" and "hits.55"

image

So when I would test I will se hits and not hits.55

app/models/metric.rb
```diff
@@ -218,6 +218,10 @@ def disabled_for_plan?(plan)
     !enabled_for_plan?(plan)
   end
 
+  def service?
+    owner_type == 'Service'
```
akostadinov, line 222:
```suggestion
    service_id || owner_type == 'Service'
```

I'm not sure whether metrics can have a `service_id` but no owner. I wanted in the past to make metrics always use `owner` instead of the obsoleted `service` but I'm not sure that I came to finalize that.

akostadinov: Actually for consistency, better add a method `service_metric?` to `app/lib/backend_api_logic/metric_extension.rb`. See `#backend_api_metric?`.

features/old/stats/provider_side.feature
```diff
@@ -47,7 +47,17 @@ Feature: Provider stats
...
+    Given provider "foo.3scale.localhost" has multiple applications disabled
+    When I log in as provider "foo.3scale.localhost"
+    Given a product
+    And a backend
+    And the backend is used by this product
+    When I go to the overview page of product "API"
+      And I follow "Analytics"
+      And I follow "Traffic"
+    Then I should see "Hits (hits)"
+    Then I should not see "Hits (hits.55)"
```
jlledom, line 60: About whether the `I should see` step finds hidden items, I did  a quick research and I think it doesn't. Because:

1. The step uses the method should `have_content`
2. `have_content` is an alias for `have_text` [code](https://github.com/teamcapybara/capybara/blob/master/lib/capybara/rspec/matchers.rb#L133)
3. `have_test` ends up calling `assert_text`
4. We are not sending any option to `assert_text` , and also don't set `ignore_hidden_elements`  (AFAIK). In this context the [docs](https://github.com/teamcapybara/capybara/blob/master/lib/capybara/node/matchers.rb#L654) say only visible elements are considered.

About this cucumber, I still think it doesn't assure the issue is solved. If I place the debugger to see in which order the metrics are being received from the controller , I get:

```
0> @metrics.first
=> #<Metric id: 2, system_name: "hits", description: "Number of API hits", unit: "hit", created_at: "2023-04-05 15:05:51", updated_at: "2023-04-05 15:05:51", service_id: 2, friendly_name: "Hits", parent_id: nil, tenant_id: 2, owner_id: 2, owner_type: "Service">
```
So the first metric is for a service and we cannot be sure if the default metric is a service one by coincidence or because of the code. You have to ensure the first metric received is a backend one, so if the default metric is for a service, it's because the code did it.

jlledom (changes requested):

features/old/stats/provider_side.feature
nidhi-soni1104: why do you think that in response it should always  be backend one, suppose you have total of 5 metric and out of which only one is of backend.

so we have added check in the view to filter it by condition  of checking owner_type. And in specs also for a given product and backend by default they have  2 metric on for each owner type and hits and hits.55 are their system name

so if we he Hits (hits)  then definitely its a service one

nidhi-soni1104: I can check the 4 points you mentioned

jlledom: > so if we he Hits (hits) then definitely its a service one

That's the expected result, but you can't know either it's correct because of your patch or because the service one happens to be the first in the list. If you run this test on master it will pass too, because it's just selecting the first in the item in the list, which happens to be `Hits (hits)`.

app/models/metric.rb
mayorova: Agree with @akostadinov 
in fact, I think the following can be used in the view:
```
@metrics.find { |metric| metric.hits? && !metric.backend_api_metric? }.system_name
```
Or, add a `service_metric?` method in that `metric_extension.rb`

features/old/stats/provider_side.feature
nidhi-soni1104: so what kind of test you are expecting. I am not sure at what extent we can make thing check with cucumber

app/models/metric.rb
nidhi-soni1104: @akostadinov @mayorova Changes are done

features/old/stats/provider_side.feature
jlledom: I expect the same test, but receiving a backend metric first in the list

nidhi-soni1104: I don't know how to do that. Can you give example.

features/old/stats/provider_side.feature
```diff
@@ -52,12 +52,16 @@ Feature: Provider stats
     When I log in as provider "foo.3scale.localhost"
     Given a product
     And a backend
+    And the following BackendApi type metrics:
+      | pasta  |
+      | Pizza  |
+    And the following metrics:
+      | Curry |
```
jlledom, line 59: This doesn't work, and it's because the new step `the following BackendApi type metric(s):` is adding the new backend to the existing service. To ensure this works. The backend must be created before the service. So you must arrange everything to:

1. Create the backend
2. Create the service
3. add the backend to that service.

These are the exact steps required to reproduce the issue as they are detailed in the [issue](https://issues.redhat.com/browse/THREESCALE-8719) description.

jlledom (changes requested):

features/step_definitions/service_steps.rb
```diff
@@ -5,7 +5,7 @@
 end
 
 Given "a service {string} of {provider}" do |name, provider|
-  provider.services.create! :name => name, :mandatory_app_key => false
+  @service = provider.services.create! :name => name, :mandatory_app_key => false
```
mayorova, line 8: As this line is changing, I'd suggest to use the syntax `name: name, mandatory_app_key: false`

features/step_definitions/api/metrics_steps.rb
```diff
@@ -6,6 +6,12 @@
   end
 end
 
+Given('the following BackendApi type metric(s):') do |table|
+  table.raw.flatten.each do |name|
+    FactoryBot.create(:metric, owner: @provider.default_service, system_name: name, friendly_name: name, owner_type: "BackendApi")
+  end
+end
+
```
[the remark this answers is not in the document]
nidhi-soni1104: Okay this not needed anymore, I added before but now not needed

features/step_definitions/service_steps.rb
nidhi-soni1104: ok

features/step_definitions/backend_api_steps.rb
```diff
@@ -39,6 +39,10 @@
   @product.backend_api_configs.create!(backend_api: @backend, path: "/my_product")
 end
 
+Given "the backend is used by this service" do
+  @service.backend_api_configs.create!(backend_api: @backend, path: "/my_product")
+end
+
```
josemigallas, line 45: You've duplicated the same step from above. Use that instead.

features/old/stats/provider_side.feature
```diff
@@ -40,6 +40,17 @@ Feature: Provider stats
...
+  Scenario: Default metric is product metric
+    Given provider "foo.3scale.localhost" has multiple applications disabled
+    When I log in as provider "foo.3scale.localhost"
+    Given a backend
+    And a service "My API" of provider "foo.3scale.localhost"
+    And the backend is used by this service
+    When I go to the overview page of product "My API"
+    And I follow "Analytics"
+    And I follow "Traffic"
+    Then I should see "Hits (hits)"
```
josemigallas, line 52: We need to tests, and there's also some room for improvements:
```suggestion
  Scenario: Default metric when creating a backend after product
    Given a service "My API" of the provider
    And a backend
    And the backend is used by this product
    When I go to the overview page of product "My API"
    And I follow "Analytics"
    And I follow "Traffic"
    Then I should see "Hits (hits)"

  Scenario: Default metric when creating a backend before product
    Given a backend
    And a service "My API" of the provider
    And the backend is used by this product
    When I go to the overview page of product "My API"
    And I follow "Analytics"
    And I follow "Traffic"
    Then I should see "Hits (hits)"
```

features/step_definitions/backend_api_steps.rb
jlledom: Not really, because the first modifies `@product` and the second modifies `@service`

nidhi-soni1104:
@mayorova @josemigallas Changes done and test also passed

lvillen (approved):

features/step_definitions/backend_api_steps.rb
josemigallas: Same logic and model exactly, different nomenclature. Moreover, service is outdated and we should use product. Define @product in https://github.com/3scale/porta/pull/3271/files#diff-c9c27f4519be20d65e92d0aad9b8195f25e1f62748ff556399e9c6e04da9e965R8 or do double assignment `@service = @product = ...`

josemigallas (changes requested):

features/step_definitions/backend_api_steps.rb
josemigallas: And if you reeeally want to change the name then you could do a polyvalent step such as:
```
Given "the backend is used by this service/product" do
  (@service || @product) ...
end
```

josemigallas: Also, in no case would make sense to have both `@service` and `@product` defined in the same Scenario, because they refer to the same instance.

nidhi-soni1104: all over the application both service and product are used 
even if you go to product list also the route says service so I think we can use both.

nidhi-soni1104: May be we can use double assignment

josemigallas: TL;DR: don't use service. Use product.

This all come from the times before API as a Product (APIAP).

Back in the day, API and service was the same, that's why you see _service_ all over the tests. However, since the implementation of APIAP we have Product and Backend.
Service and Product are the same: `Service` is still the model but the "concept" was renamed to Product, because technically BackendAPIs are the services. Therefore, we should quit using `service` and use `product` and `backend` instead, specially when we use it in a "public context" such as a cucumber scenario that is part of the specification / documentation.

nidhi-soni1104:
@josemigallas All changes are done. can we merge this now?

josemigallas (approved):

jlledom (approved):

PR #3274, opened 2023-03-22
THREESCALE-8184  Use assert_nil if expecting nil

nidhi-soni1104:

make those field required

**DEPRECATED**: Use assert_nil if expecting nil

 https://issues.redhat.com/browse/THREESCALE-8184

### DIFF
--- a/test/events/service_contracts/service_contract_cancellation_event_test.rb
+++ b/test/events/service_contracts/service_contract_cancellation_event_test.rb
@@ -10,7 +10,7 @@ class ServiceContracts::ServiceContractCancellationEventTest < ActiveSupport::Te
     assert event
     assert_equal event.plan_name, plan.name
     assert_equal event.service_name, service.name
-    assert_equal event.provider, contract.provider_account
+    assert_equal nil, contract.provider_account
     assert_equal event.service, contract.issuer
     assert_equal event.account_id, contract.buyer_account.id
     assert_equal event.account_name, contract.buyer_account.name
--- a/test/events/service_contracts/service_contract_cancellation_event_test.rb
+++ b/test/events/service_contracts/service_contract_cancellation_event_test.rb
@@ -10,7 +10,8 @@ class ServiceContracts::ServiceContractCancellationEventTest < ActiveSupport::Te
     assert event
     assert_equal event.plan_name, plan.name
     assert_equal event.service_name, service.name
-    assert_equal nil, contract.provider_account
+    assert_nil contract.provider_account
+    assert_nil event.provider
     assert_equal event.service, contract.issuer
     assert_equal event.account_id, contract.buyer_account.id
     assert_equal event.account_name, contract.buyer_account.name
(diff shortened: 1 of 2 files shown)


test/unit/liquid/drops/account_drop_test.rb
```diff
@@ -22,7 +22,7 @@ def test_url_helpers
   end
 
   test 'returns vat_rate' do
-    assert_equal(@drop.vat_rate, @buyer.vat_rate)
+    assert_nil(nil, @buyer.vat_rate)
```
akostadinov, line 25: This doesn't seem right.

Why are `event.provider` and `@drop.vat_rate` `nil` though?

nidhi-soni1104: @akostadinov  This always comes nil when we run the test and  so we receive this warning .

 @drop is **Liquid::Drops::Account** not any object and its vat rate is always nil
@buyer is buyers account and to this account also vat rate is nil 
so basically whoever added this test is checking nil to  nil and so it gives a warning

there is no new change with the data of the object they are as it is .
I don't know why someone was checking nil equal values, so this fix actually clear that they are checking nil

nidhi-soni1104: This warning  was keep showing when test are run so I fixed the warning

akostadinov: The issues are two:
1. `assert_nil(nil, @buyer.vat_rate)` is not correct parameters for the assertion
2. the test is called `returns vat_rate`, if it is always `nil`, the test name seems confusing

akostadinov: yes, create object with VAT. Also the `nil` check is good to be performed, just fix the parameters.

test/events/service_contracts/service_contract_cancellation_event_test.rb
```diff
@@ -10,7 +10,7 @@ def test_create
     assert event
     assert_equal event.plan_name, plan.name
     assert_equal event.service_name, service.name
-    assert_equal event.provider, contract.provider_account
+    assert_equal nil, contract.provider_account
```
akostadinov, line 13: Is it ok that we have events without a provider? Maybe test was boken.

test/unit/liquid/drops/account_drop_test.rb
```diff
@@ -21,10 +21,17 @@ def test_url_helpers
...
 
+  test 'returns nil vat_rate for drop if buyer vate rate is nil' do
+    assert_nil @drop.vat_rate
+  end
+
   test 'returns vat_rate' do
-    assert_equal(@drop.vat_rate, @buyer.vat_rate)
+    @buyer.update(vat_rate: '123.4')
+    @drop_account = Drops::Account.new(@buyer)
+    assert_equal(@drop_account.vat_rate, @buyer.vat_rate)
```
akostadinov, line 31: looks good, just a nit pick
```suggestion
    assert_equal @buyer.vat_rate, @drop_account.vat_rate
```

test/events/service_contracts/service_contract_cancellation_event_test.rb
nidhi-soni1104: I think yes, because I cannot see any such association

akostadinov (changes requested):

nidhi-soni1104:
@akostadinov all comments addressed

akostadinov (approved):

PR #3275, opened 2023-03-22
💣🐒 Reduce webpackers' config monkey-patching and use v5 default

josemigallas:
> ⚠️ **IMPORTANT NOTE:**
> The main problem about upgrading webpacker and webpack is that at some point we stopped using the default configuration. Now webpacker 6 diverges too much, specially because it now supports webpack 5 and a lot changes there.
>
> This PR removes some "tech-debt" related to our webpack configuration. There are more to come.

Ran `webpacker install` to "re-install" all default files and configs and then cleaned up everything so that we have a standard webpacker installation. This way the upgrade to v6 and later shakapacker will happen smoothly.


config/webpack/environment.js
```diff
@@ -1,44 +1,33 @@
...
 
-// Add global webpack configs here
-
-environment.loaders.delete('css')
-environment.loaders.delete('moduleCss')
-environment.loaders.delete('sass')
-environment.loaders.delete('moduleSass')
+// TS and TSX transpilation is done with tsc via ts-loader, we don't use babel even though it's the
+// webpacker's default. The only use for babel is to transpile legacy module Stats, which could not
+// be migrated to Typescript due to potential extreme sanity loss.
```
akostadinov, line 6: could you clarify sanity loss? Sanity of the code or sanity of the coder?

config/webpack/environment.js
```diff
@@ -62,4 +51,9 @@ Object.assign(fileLoader.use[0].options, {
   postTransformPublicPath: (p) => `window.rails_asset_host + ${p}`
 });
 
+// Whatever this is, it seems we don't need it.
+environment.loaders.delete('nodeModules')
+
+console.dir(environment.loaders, { depth: null })
```
akostadinov, line 57: what is this doing?

akostadinov (approved):
Some questions and nitpicks that you may or may not ignore, at your discretion. This is a huge improvement!

config/webpack/environment.js
```diff
@@ -62,4 +51,9 @@ Object.assign(fileLoader.use[0].options, {
   postTransformPublicPath: (p) => `window.rails_asset_host + ${p}`
 });
 
+// Whatever this is, it seems we don't need it.
+environment.loaders.delete('nodeModules')
```
akostadinov, line 55: Probably this gives a little explanation:
https://github.com/rails/webpacker/blob/54c3ca9245e9ee330f8ca63b447c202290f7b624/docs/v4-upgrade.md#excluding-node_modules-from-being-transpiled-by-babel-loader

config/webpack/environment.js
josemigallas: > **Sanity.**
> the state of having a healthy mind and not being mentally ill

The coder of course 😆

.babelrc
```diff
@@ -1,3 +1,4 @@
+// Only use to transpile JSX from app/javscript/src/Stats
```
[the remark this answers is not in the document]
josemigallas: I've been tempted of removing this file several times, since it is not used apparently. Anyway, it's not used anymore by webpacker and it's gonna be removed sooner than later in favor of `babel.config.js`. I wouldn't sweat about it.

config/webpack/environment.js
josemigallas: This is for debugging, I left it there by accident 😅

config/webpack/environment.js
josemigallas: You know what, this in fact should not be part of the PR. I removed it at first to "see" what happens but this loader comes from webpacker so why should we delete it. I will restore this line.

.postcssrc.yml
akostadinov, line 1: we need this file?

akostadinov (approved):
another nit pick but looks good!

.postcssrc.yml
josemigallas: Yes, now that we actually include (or rather don't remove) `postcss-loader`, it will complain without a config file. It complains anyway that we aren't using any plugin 😅 but it's gonna be changed to `postcss.config.js` soon and I'll try to make the plugins work as it is indented by webpacker.

config/webpack/environment.js
akostadinov: If it isn't needed, then it can speed up compilation to be removed. So maybe keep that line. But as you decide.

bin/webpack-dev-server
```diff
@@ -4,16 +4,15 @@ ENV["RAILS_ENV"] ||= ENV["RACK_ENV"] || "development"
 ENV["NODE_ENV"]  ||= "development"
 
 require "pathname"
-
-APP_ROOT = File.expand_path("..", __dir__)
-
-ENV["BUNDLE_GEMFILE"] ||= File.expand_path("Gemfile", APP_ROOT)
+ENV["BUNDLE_GEMFILE"] ||= File.expand_path("../../Gemfile", Pathname.new(__FILE__).realpath)
```
jlledom, line 7: Why this change? it's not exactly the same?

babel.config.js
jlledom, line 1: Is this file automatically generated?

config/initializers/content_security_policy.rb
```diff
@@ -23,3 +23,7 @@
 # For further information see the following documentation:
 # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy-Report-Only
 # Rails.application.config.content_security_policy_report_only = true
+
+Rails.application.config.content_security_policy do |policy|
+  policy.connect_src :self, :https, "http://localhost:3035", "ws://localhost:3035" if Rails.env.development?
+end
```
jlledom, line 29: What's this for?

config/webpacker.yml
```diff
@@ -3,60 +3,96 @@
...
+    - .ico
+    - .svg
+    - .eot
+    - .otf
+    - .ttf
+    - .woff
+    - .woff2
+
   extensions:
-    - .coffee
```
jlledom, line 17: what about coffee? is it not packed by webpack?

config/webpacker.yml
```diff
@@ -3,60 +3,96 @@
...
+
   # Compile test packs to a separate directory
   public_output_path: packs-test
 
 production:
   <<: *default
 
   # Production depends on precompilation of packs prior to booting for performance.
-  compile: false
+  compile: true
```
jlledom, line 92: Why change this? I think it was correct before, we precompile on the image build and deliver the assets inside the image.

config/initializers/content_security_policy.rb
akostadinov: Probably this is for `webpack-dev-server`. But please make sure app works well also without it as right now.

app/javascript/src/Stats/buyer/stats_application.jsx
```diff
@@ -1,3 +1,4 @@
+/** @jsxRuntime classic */
```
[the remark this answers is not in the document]
josemigallas: It's a pragma for babel's `plugin-transform-react-jsx`. In summary It's required to transpile `Stats/` files.

bin/webpack-dev-server
josemigallas: Yes, but this is auto generated by the webpacker installer.

babel.config.js
josemigallas: Yes

app/javascript/src/Stats/buyer/stats_application.jsx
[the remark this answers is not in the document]
josemigallas: More context about this line: https://github.com/system-ui/theme-ui/issues/1160#issuecomment-715530924

config/initializers/content_security_policy.rb
josemigallas: it's not needed 👍🏻

jlledom (commented):
Why not upgrading `@types/jest`? Does it do any harm?

josemigallas:
> Why not upgrading `@types/jest`? Does it do any harm?

It's not directly related to webpacker and I'd like to upgrade the whole jest suite in another PR.

jlledom (dismissed):

josemigallas:
This is not looking good. CI won't compile because OOM, the bundle is bigger and styles are not injected causing some tests to fail.
I'm gonna hold this for now and try upgrade directly to shakapacker, which has direct support over their Slack channel and is supposed to fix most of this stuff.

PR #3279, opened 2023-03-27
⬆️ Bump `@typescript-eslint plugin`

josemigallas:
And fixes some new errors caught.

Main reason is to fix the following warning when webpack compies (or simply `yarn lint`):
```
=============

WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: >=3.3.1 <4.5.0

YOUR TYPESCRIPT VERSION: 4.8.4

Please only submit bug reports when using the officially supported version.

=============
```


akostadinov (approved):
Looks good but better somebody that does js also review.

lvillen (approved):

PR #3281, opened 2023-03-28
⬆️ Bump `jest` to v28

josemigallas:
Latest is v29 but it requires node 14 so separating into a different PR.

**NOTE ABOUT CONFIG FILE**
It is unchanged except maybe a couple of redundant options have been removed. As for the apparently useless changes, they are the result of running `jest --init` which uses the latest config. Better to keep this file up to date in case we need setting any of the configs later.


mayorova (dismissed):

.eslintrc
```diff
@@ -10,8 +10,7 @@
     }
   },
   "env": {
-    "browser": true,
-    "jest": true
```
akostadinov, line 14: Why remove from here and put in the other one? Other than that things make sense.

josemigallas: This is a linter config specific to `jest` so I moved it to the jest's specific config file `spec/javascripts/.eslintrc`.

However, I've looked into it and it seems this env is outdated and it belonged back when there wasn't a `eslint-plugin-jest`. It can be removed now.

PR #3283, opened 2023-03-28
db:seed succeed if backend is unavailable

akostadinov:
When doing db:reset usually backend is down. If local configuration
doesn't use `fake_server`, then operation will fail at `db:seed`.

And since recently our example config does disable `fake_server`.

This change makes sure we mock it regardless of configuration when
we detect that backend is not available. So if anybody has a use case
where backend is indeed available it will work as expected. But not fail
for others.

### DIFF
--- a/app/lib/three_scale/socket.rb
+++ b/app/lib/three_scale/socket.rb
@@ -1,0 +1,13 @@
+module ThreeScale
+  module Socket
+    module ClassMethods
+      def accepts_connections?(host, port)
+        ::Socket.tcp(host, port, connect_timeout: 1) { true }
+      rescue
+        false
+      end
+    end
+
+    extend ClassMethods
+  end
+end
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -2,11 +2,10 @@
 
 require 'dotenv/rails-now'
 
-
 ActiveRecord::Base.transaction do
-  if Rails.env.test? || System::Application.config.three_scale.core.fake_server
+  backend_url = URI::parse ThreeScale::Core.url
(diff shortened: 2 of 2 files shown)


app/lib/three_scale/socket.rb
```diff
@@ -0,0 +1,13 @@
...
+  module Socket
+    module ClassMethods
+      def accepts_connections?(host, port)
+        ::Socket.tcp(host, port, connect_timeout: 1) { true }
+      rescue
+        false
+      end
+    end
+
+    extend ClassMethods
```
jlledom, line 11: This design is a bit tricky, are you trying to avoid extending `ActiveSupport::Concern`?

db/seeds.rb
```diff
@@ -2,11 +2,10 @@
 
 require 'dotenv/rails-now'
 
-
 ActiveRecord::Base.transaction do
-  if Rails.env.test? || System::Application.config.three_scale.core.fake_server
+  backend_url = URI::parse ThreeScale::Core.url
+  if Rails.env.test? || System::Application.config.three_scale.core.fake_server || !ThreeScale::Socket.accepts_connections?(backend_url.host, backend_url.port)
     require Rails.root.join('test/test_helpers/backend')
-    TestHelpers::Backend::MockCore.mock_core!
```
jlledom, line 9: Why is this not needed anymore? at this point we don't have a backend, mocking one seems correct.

app/lib/three_scale/socket.rb
jlledom: My bad!, I assumed that you couldn't use `ClassMethods` unless you include `ActiveSupport::Concern`.

jlledom (approved):

PR #3291, opened 2023-03-30
CMS API: Remove DataTag module

jlledom:
The old `CMS::DataTag` module was created originally to set a single point of reference to share a tag between XML and JSON formats. But after removing XML, this is not much useful and it's basically dead code. I created this new `CMS::TypeMap`  class to replace it in a way it's more general and can be used from both API responses and search filters, or any other place in the future.


mayorova (changes requested):
Good stuff!

Although, I would add some protection from wrong input.

e.g. in `cms_class`discard those that don't have `to_sym`
and in `cms_type` only call `[0]` if the element is actually found.

Maybe some other checks... 
Not sure what would be the behavior on wrong input, but I guess returning nil is OK.

Also we need a unit test (where all the above scenarios would be checked).

jlledom:
> I would add some protection from wrong input.
> 
> Also we need a unit test (where all the above scenarios would be checked).

Done, and done.

mayorova (approved):
Great, thank you :+1:

PR #3292, opened 2023-03-30
Fix `bootstrap-sass` warning

josemigallas:
Recently the following warning shows up when running webpack:
```
Deprecation Warning: Using / for division outside of calc() is deprecated and will be removed in Dart Sass 2.0.0.

Recommendation: math.div($line-height-computed, 2) or calc($line-height-computed / 2)

More info and automated migrator: https://sass-lang.com/d/slash-div
```
To remove it, upgrade `bootstrap-sass`.

On the other hand, we're importing **the whole** library to style only a couple components in Integration / Policies. Ideally we should remove bootstrap-sass altogether and use PF4 components. In the meantime, the imported scss has been removed to the minimum.
```
master    8.7M      public/packs/js/policies-df14e7c216a288d657bb.js
fix       8.6M      public/packs/js/policies-8007597a59460b826fb8.js
```

**NOTE**
After a relatively thorough inspection, the looks have not changed in a noticeable way. The advantage of using Sass is that whenever a mixin, variable or whatever is missing, it fails compilation.

**Verification**:
1. There is no deprecation warning when `bin/webpack`.
2. Policies' styles are not broken

**Changelog for bootstrap-sass**: https://github.com/twbs/bootstrap-sass/blob/master/CHANGELOG.md#343-non-ruby-only

### DIFF
--- a/app/javascript/src/Policies/styles/policies.scss
+++ b/app/javascript/src/Policies/styles/policies.scss
@@ -2,7 +2,39 @@
 
 $icon-font-path: "~bootstrap-sass/assets/fonts/bootstrap/";
 .PolicyConfiguration {
-  @import "~bootstrap-sass/assets/stylesheets/_bootstrap.scss";
+  @import "~bootstrap-sass/assets/stylesheets/bootstrap/_variables.scss";
+
+  // Utilities
+  @import "~bootstrap-sass/assets/stylesheets/bootstrap/mixins/_opacity.scss";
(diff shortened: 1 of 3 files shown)


nidhi-soni1104 (approved):
LGTM, just a question why so many imports

lvillen (approved):

josemigallas:
> LGTM, just a question why so many imports?

They're actually fewer now, take a look bootstrap-sass/assets/stylesheets/_bootstrap.scss 😃

PR #3293, opened 2023-03-30
Flatten javscript packs

josemigallas:
Webpacker does not want to have nested directories under `/packs`.


lvillen (approved):

jlledom (approved):
I tried and it works. Two questions:

1. 
> Webpacker does not want to have nested directories under `/packs`

Why?

2. You got rid of the `email_configurations` and `quickstarts` folders, but there was another folder called `PF4Styles` (or something like that, I don't remember) which has also disappeared, how is that possible?

josemigallas:
> I tried and it works. Two questions:
> 
> 
> > Webpacker does not want to have nested directories under `/packs`
> 
> Why?

Webpacker 6 doesn't support it, but it seem they implemented it in shakapacker. From the changelog:
```
 Configuration boolean option `nested_entries` to use nested entries. This was the default prior to v6.0. Because entries maybe generated, it's useful to allow a `generated` subdirectory
 ```
 The default is false and I suggest we don't keep nested entries in order to try keep the packs folder small.

> 2. You got rid of the `email_configurations` and `quickstarts` folders, but there was another folder called `PF4Styles` (or something like that, I don't remember) which has also disappeared, how is that possible?

Magic. Also this PR https://github.com/3scale/porta/pull/3290. But mainly magic 😄

PR #3295, opened 2023-03-31
Clean up `navigation` pack

josemigallas:
It removes a piece of code that I've been not liking for a long time 😬 and that IMO it doesn't make sense anymore. It's a very generic piece of logic that's only used by the 2 overflow menus at the right of the header.

It also fixes a weird kind of UI glitch that makes the menu transparent for a second:
Screenshot 2023-03-31 at 12 34 14

Verification:
pop-navigation


app/javascript/packs/vertical_nav.ts
```diff
@@ -17,4 +17,17 @@ document.addEventListener('DOMContentLoaded', () => {
     activeItem,
     currentApi: safeFromJsonString(currentApi)
   }, containerId)
+
+  // TODO
```
josemigallas, line 21: Oops, removing this, it will come next!

app/javascript/packs/navigation.ts
```diff
@@ -1,42 +1,38 @@
...
+
+    docs.classList.remove(expandedClass)
+    docsMenu.classList.remove(expandedClass)
+  })
 
-    store.isVerticalNavCollapsed = shouldVertNavCollapse
+  document.body.addEventListener('click', () => {
+    [docs, docsMenu, session, sessionMenu].forEach(({ classList }) => {
+      classList.remove(expandedClass)
+    })
```
jlledom, line 36: I don't like the idea of adding a listener for body. Do you think there could be another way? Isn't there some dropdown React component which visibility depends on a reactive boolean variable, that way we could avoid the listener.

josemigallas: Right now we're not using React on the header. We have a hook for handling this but I'm afraid the listener is going to be there one way or another.
https://github.com/3scale/porta/blob/bc7513d8844cc3dd3519c16ccdd1b60750f22b60/app/javascript/src/utilities/useClickOutside.ts#L14

jlledom (approved):

PR #3296, opened 2023-04-03
CMS API: Forbid extra params at controller level

jlledom:

Before this PR, when a client sent a parameter not permitted but also not forbidden, it was ignored. It's a requisite that the API returns an error in this case.

Rails' strong parameters provides a [config option](https://guides.rubyonrails.org/configuring.html#config-action-controller-action-on-unpermitted-parameters) to enable this behavior, the problem is that option acts globally, so if we enable it that would problably break some customers implementation for other APIs.

To avoid that, this PR adds a controller method to do the same, but at controller level:

`forbid_extra_params :raise, whitelist: %i[id page per_page]`

Available actions are `:log` and `:raise`. The `:whitelist` option accepts a list of parameters that will be ignored by the module.

[THREESCALE-9172](https://issues.redhat.com/browse/THREESCALE-9172)

**Verification steps** 

Request:

```
curl --request POST \
  --url 'http://provider-admin.3scale.localhost:3000//admin/api/cms/sections?access_token=secret' \
  --data '{
	"title": "Test section from API",
	"system_name": "test-section-api",
	"extra_param": "value"
}'
```

Should return:

`{"error":"Unpermitted parameters: [\"extra_param\"]"}`

### DIFF
--- a/app/controllers/admin/api/cms/templates_controller.rb
+++ b/app/controllers/admin/api/cms/templates_controller.rb
@@ -14,6 +14,8 @@ class Admin::Api::CMS::TemplatesController < Admin::Api::CMS::BaseController
   wrap_parameters :template, include: AVAILABLE_PARAMS,
                              format: %i[json multipart_form url_encoded_form]
 
+  forbid_extra_params :raise, whitelist: %i[id page per_page type layout_name section_name]
+
   before_action :find_template, except: %i[index create]
 
   before_action :can_destroy, only: :destroy
--- a/app/lib/api_support/forbid_params.rb
+++ b/app/lib/api_support/forbid_params.rb
@@ -1,0 +1,62 @@
+# frozen_string_literal: true
+
+module ApiSupport
+
+  class UnpermittedParameters < StandardError
+    def initialize(unpermitted_keys)
+      msg = "Unpermitted parameters: #{unpermitted_keys}"
+      super msg
+    end
+  end
+
+  module ForbidParams
+    extend ::ActiveSupport::Concern
+    include Params
+
+    included do
+      class_attribute :_forbid_params_action
+      class_attribute :_forbid_params_whitelist, default: []
+
+      rescue_from UnpermittedParameters do |error|
+        render_error error.message, status: :unprocessable_entity
+      end
+    end
+
+    class_methods do
+      def forbid_extra_params(action, options = {})
+        before_action :unpermitted_parameters_check
+        self._forbid_params_action = action
+
+        return unless options[:whitelist]
+
+        self._forbid_params_whitelist = [*options[:whitelist]]
+      end
+    end
+
+    private
+
+    def forbid_params_whitelist
+      self.class._forbid_params_whitelist.map(&:to_s)
+    end
+
+    def wrapped_keys
+      [_wrapper_key, *_wrapper_options[:include]].map(&:to_s)
+    end
+
+    def unpermitted_keys
+      flat_params.keys - wrapped_keys - forbid_params_whitelist
+    end
+
+    def unpermitted_parameters_check
+      return unless _forbid_params_action
+      return if unpermitted_keys.blank?
+
+      case self.class._forbid_params_action
+      when :log
+        Rails.logger.warn("Unpermitted parameters: #{unpermitted_keys}")
+      when :raise
+        raise UnpermittedParameters, unpermitted_keys
+      end
+    end
+  end
+end
--- a/app/lib/api_support/params.rb
+++ b/app/lib/api_support/params.rb
@@ -10,7 +10,7 @@ module ApiSupport::Params
     private
 
     def flat_params
-      params.except(:format, :controller, :action, :provider_key, :api_key)
+      params.except(:format, :controller, :action, :provider_key, :api_key, :access_token)
     end
 
     # This basically checks if the wrapped parameter is passed as a Hash, if not it will go 422
--- a/test/integration/api_support/forbid_params_test.rb
+++ b/test/integration/api_support/forbid_params_test.rb
@@ -1,0 +1,151 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class ApiSupport::ForbidParamsTest < ActionDispatch::IntegrationTest
+
+  class TestController < ApplicationController
+    include ApiSupport::ForbidParams
+    def create
+      params.permit(:permitted)
+    end
+  end
+
+  def with_test_routes
+    Rails.application.routes.draw do
+      post '/test/default' => 'api_support/forbid_params_test/action/default/default#create'
+      post '/test/log' => 'api_support/forbid_params_test/action/log/log#create'
+      post '/test/raise' => 'api_support/forbid_params_test/action/raise/raise#create'
+      post '/test/symbol' => 'api_support/forbid_params_test/whitelist/symbol/symbol#create'
+      post '/test/string' => 'api_support/forbid_params_test/whitelist/string/string#create'
+      post '/test/list' => 'api_support/forbid_params_test/whitelist/list/list#create'
+      post '/test/child' => 'api_support/forbid_params_test/single_callback/child#create'
+    end
+    yield
+  ensure
+    Rails.application.routes_reloader.reload!
+  end
+
+  def capture_log
+    orig_logger = Rails.logger.dup
+    logger_output = StringIO.new
+
+    begin
+      Rails.logger = ActiveSupport::Logger.new(logger_output)
+      yield
+    ensure
+      Rails.logger = orig_logger
+    end
+
+    logger_output.string
+  end
+
+  class Action < ApiSupport::ForbidParamsTest
+    class Default <Action
+      class DefaultController < TestController; end
+
+      def setup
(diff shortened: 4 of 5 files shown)


thalesmiguel (commented):
Cool stuff. Just left a couple comments there :v:

app/lib/api_support/forbid_params.rb
```diff
@@ -0,0 +1,62 @@
...
+      class_attribute :_forbid_params_whitelist, default: []
+
+      rescue_from UnpermittedParameters do |error|
+        render_error error.message, status: :unprocessable_entity
+      end
+    end
+
+    class_methods do
+      def forbid_extra_params(action, options = {})
+        before_action :unpermitted_parameters_check
```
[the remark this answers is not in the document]
thalesmiguel: >This is standard for ActiveSupport::Concern

What is standard?

test/integration/api_support/forbid_params_test.rb
```diff
@@ -0,0 +1,154 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class ApiSupport::ForbidParamsTest < ActionDispatch::IntegrationTest
+
```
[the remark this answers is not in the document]
jlledom: Done

app/lib/api_support/forbid_params.rb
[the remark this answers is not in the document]
akostadinov: `class_methods` is an `ActiveSupport::Concern` thing. But now I realize you're talking about `before_action :unpermitted_parameters_check`. :no_mouth: 

And I understand your concern. Joan demonstrated though that it's not the case and added a test case. Also we may raise in case `self._forbid_params_action` was already set to catch occasions where we call this method twice.

I suggested the approach because it is mostly clean for controllers that don't use it and only things change for controllers that do.

For example in present version, one can include the module in the base controller and then only call `forbid_extra_params` wherever needed.

It is very similar to how it is done in gems like [audited](https://github.com/collectiveidea/audited/blob/900303f7a296884957c27794eb01b4dac43d50d6/lib/audited/auditor.rb#L85-L88). And to my opinion it is nice.

app/lib/api_support/params.rb
```diff
@@ -10,7 +10,7 @@ module ApiSupport::Params
     private
 
     def flat_params
-      params.except(:format, :controller, :action, :provider_key, :api_key)
+      params.except(:format, :controller, :action, :provider_key, :api_key, :access_token)
```
akostadinov, line 13: could you explain this change?

app/lib/api_support/forbid_params.rb
```diff
@@ -0,0 +1,61 @@
...
+
+        return unless options[:whitelist]
+
+        self._forbid_params_whitelist = [*options[:whitelist]]
+      end
+    end
+
+    private
+
+    def forbid_params_whitelist
```
akostadinov, line 37: all these methods, if they are only internal methods, not supposed to be called by anything outside this module, maybe rename them to `_***`?

If they are to be called by code outside this module, then this is alright.

app/lib/api_support/params.rb
jlledom: Yes, `flat_params` goal is to return the keys that actually mater, that is, the parameters belonging to the resource we are working with. It returns all the keys except those that are usually received in requests but we don't care about. I added the `:access_token` to this list, that way it will be always ignored by `ForbidParams`.

app/lib/api_support/forbid_params.rb
jlledom: They are already marked as private methods, why the prefix? I don't see the prefix in any other private method in porta.

akostadinov: To avoid conflicts. I'm always very paranoid about modules. As they can interfere with each other :(
Maybe longer, more specific/clear names?

jlledom: It's OK, I renamed then.

app/lib/api_support/forbid_params.rb
```diff
@@ -0,0 +1,61 @@
+# frozen_string_literal: true
+
+module ApiSupport
+  class UnpermittedParameters < StandardError
```
akostadinov, line 4: Just a nitpick, unless you have special considerations, to be `UnpermittedParametersError`.

app/lib/api_support/forbid_params.rb
```diff
@@ -0,0 +1,61 @@
...
+      flat_params.keys - _wrapped_keys - _forbid_params_whitelist_keys
+    end
+
+    def _unpermitted_parameters_check
+      return unless _forbid_params_action
+      return if _unpermitted_keys.blank?
+
+      case self.class._forbid_params_action
+      when :log
+        Rails.logger.warn("Unpermitted parameters: #{_unpermitted_keys}")
```
akostadinov, line 55: Another nitpick to extract this as a local variable, although we shouldn't see many of these.

test/integration/api_support/forbid_params_test.rb
```diff
@@ -0,0 +1,153 @@
...
+      yield
+    ensure
+      Rails.logger = orig_logger
+    end
+
+    logger_output.string
+  end
+
+  class Action < ApiSupport::ForbidParamsTest
+    class Default <Action
```
akostadinov, line 43: spacing, I wonder why rubocop didn't complain

test/integration/api_support/forbid_params_test.rb
```diff
@@ -0,0 +1,153 @@
...
+      end
+
+      test 'defaults to global action_on_unpermitted_parameters' do
+        with_test_routes do
+          assert_raises(ActionController::UnpermittedParameters) { post '/test/default', params: { unpermitted: true } }
+        end
+      end
+    end
+
+    class Log <Action
```
akostadinov, line 61: spacing

test/integration/api_support/forbid_params_test.rb
```diff
@@ -0,0 +1,153 @@
...
+            post '/test/log', params: { unpermitted: true }
+          end
+
+          assert_response :success
+          assert log_output.include?('Unpermitted parameters: ["unpermitted"]')
+        end
+      end
+    end
+
+    class Reject <Action
```
akostadinov, line 78: spacing

test/integration/api_support/forbid_params_test.rb
```diff
@@ -0,0 +1,153 @@
...
+      class RejectController < TestController
+        forbid_extra_params :reject
+      end
+
+      test 'returns 406 when forbid_extra_params = :reject' do
```
akostadinov, line 83: `unprocessable_entity` coresponds to 422

test/integration/api_support/forbid_params_test.rb
```diff
@@ -0,0 +1,153 @@
...
+      forbid_extra_params :log
+    end
+
+    class ChildController < BaseController
+      forbid_extra_params :log
+    end
+
+    test 'the callback is only called once' do
+      with_test_routes do
+        ChildController.any_instance.expects(:_unpermitted_parameters_check).times(1)
```
akostadinov, line 148:
```suggestion
        ChildController.any_instance.expects(:_unpermitted_parameters_check).once
```

akostadinov (approved):
Looks nice to me. Thank you! Just some nitpicks if you like.

app/controllers/admin/api/cms/templates_controller.rb
```diff
@@ -14,6 +14,8 @@ class Admin::Api::CMS::TemplatesController < Admin::Api::CMS::BaseController
   wrap_parameters :template, include: AVAILABLE_PARAMS,
                              format: %i[json multipart_form url_encoded_form]
 
+  forbid_extra_params :reject, whitelist: %i[id page per_page type layout_name section_name]
```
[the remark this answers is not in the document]
jlledom: I doesn't extend, it overwrites the configuration from the base class

app/lib/api_support/forbid_params.rb
jlledom: Done

app/lib/api_support/forbid_params.rb
jlledom: Fixed

test/integration/api_support/forbid_params_test.rb
jlledom: Done

test/integration/api_support/forbid_params_test.rb
jlledom: Super done

test/integration/api_support/forbid_params_test.rb
jlledom: Mega done

test/integration/api_support/forbid_params_test.rb
jlledom: Right

test/integration/api_support/forbid_params_test.rb
jlledom: Done

akostadinov (approved):
Thank you!

PR #3297, opened 2023-04-05
👷 Fixes bundle --deployment deprecation in in ci

josemigallas:
Fixes the following deprecation warning:
```
[DEPRECATED] The `--deployment` flag is deprecated because it relies on being remembered across bundler invocations, which bundler will no longer do in future versions. Instead please use `bundle config set --local deployment 'true'`, and stop using this flag
```


akostadinov:
please run curcleci from the branch first before merging

PR #3298, opened 2023-04-05
Proposal: disable asset compilation in development

josemigallas:
I think nobody in their right mind never ever works with porta without running `webpack-dev-server`. Loading a page takes forever otherwise. However, sometimes one simply doesn't need to recompile assets so it would be nice to be able to open a page without waiting centuries for the server to finish compiling.

By setting `compile: false` we'll have to compile manually `assets:precompile´ or via webpack-dev-server to watch for changes. This is exactly how tests work by the way.

### DIFF
--- a/config/webpacker.yml
+++ b/config/webpacker.yml
@@ -34,7 +34,7 @@ default: &default
 
 development:
   <<: *default
-  compile: true
+  compile: false
   webpack_compile_output: true
   check_yarn_integrity: false
 


akostadinov:
I use it like this. It compiles once you open a page and then it opens instantaneously. What use case bothers you?

josemigallas:
> I use it like this. It compiles once you open a page and then it opens instantaneously. What use case bothers you?

And does JS break often? This is what I see when not running webpacke-dev-server:
Screenshot 2023-04-10 at 11 57 19

akostadinov:
I don't understand what you mean by this screenshot. Basically on first open it should spend some 2 minutes to compile and then open further pages quick. Or do you see something else?

josemigallas:
> I don't understand what you mean by this screenshot. Basically on first open it should spend some 2 minutes to compile and then open further pages quick. Or do you see something else?

That screenshot is what I see when I don't run webpack-dev-server

josemigallas:
> But with latest master (16ce3210b) compilation actually fails :(

You're missing some upgrades, run `yarn install`.

akostadinov:
Actually I reproduced your issue by removing existing assets and when compilation failed. To get around this I had to:

1.  `rm -r public/packs`
2. update to node 14 in `.tool-versions`
2.  `yarn clean --all`
3.  `yarn install`
4. update NODE memory to 3072MB, see #3301 

Then it works fine. Pls verify it works for you as well.

If I don't tune memory it still fails which means we need to update the default setting after the node 14 update. 
```
[Webpacker] Compilation failed:

<--- Last few GCs --->

[697983:0x618dbc0]   133942 ms: Mark-sweep (reduce) 2556.8 (2580.9) -> 2536.2 (2560.0) MB, 892.8 / 0.0 ms  (+ 0.0 ms in 1 steps since start of marking, biggest step 0.0 ms, walltime since start of marking 894 ms) (average mu = 0.152, current mu = 0.061) a[697983:0x618dbc0]   134787 ms: Mark-sweep (reduce) 2556.8 (2576.7) -> 2536.1 (2556.0) MB, 804.6 / 0.0 ms  (average mu = 0.104, current mu = 0.048) allocation failure scavenge might not succeed
```

PR #3299, opened 2023-04-05
Fix api docs endpoints in the developer portal

jlledom:

The endpoint urls in the Developer portal docs were not being processed, due to a regression caused by https://github.com/3scale/porta/pull/3246. This fixes it.

[THREESCALE-9498](https://issues.redhat.com/browse/THREESCALE-9498)

**Verification steps** 

Just open http://provider.3scale.localhost:3000/docs and check the "Servers" dropdown is set to apicast.

**Notes for the reviewer**

The problem is at:

 https://github.com/3scale/porta/blob/81f9b67687d137206436f5c6dba1a149528ffd19/app/javascript/src/ActiveDocs/OAS3Autocomplete.ts#L135-L137

Here, `specUrl` comes as a relative path but `response.url` is absolute, so it's a false negative and `autocompleteOAS3` doesn't get called when it should. The value ultimately comes from:

https://github.com/3scale/porta/blob/81f9b67687d137206436f5c6dba1a149528ffd19/lib/developer_portal/lib/liquid/drops/api_spec.rb#L10-L12

which calls a `*_path` helper. In my opinion, this issue could be better solved if we could call a `*_url` helper instead, which would return a full url and therefore the check at `OAS3Autocomplete.ts` would work. The problem is we can't call `*_url` helpers as far as I know, they return an error in my local because the host is not set, and seems logical, provided we handle a different domain for each tenant.

This is a workaround that will be definitely fixed when we merge https://github.com/3scale/porta/pull/3103. See the [diff](https://github.com/3scale/porta/pull/3103/files#diff-4f8bdb6882899231dfd4039aec865573cb6a79e61de460f4425e45b5aa966439R148-R150)

### DIFF
--- a/spec/javascripts/ActiveDocs/OAS3Autocomplete.spec.ts
+++ b/spec/javascripts/ActiveDocs/OAS3Autocomplete.spec.ts
@@ -46,10 +46,6 @@ const specResponse = {
   },
   obj: {}
 }
-const specRelativeResponse = {
-  ...specResponse,
-  url: specRelativeUrl
-}
 const apiResponse = {
   ...specResponse,
   url: apiUrl,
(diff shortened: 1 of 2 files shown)


thalesmiguel (dismissed):

spec/javascripts/ActiveDocs/OAS3Autocomplete.spec.ts
```diff
@@ -82,7 +93,6 @@ describe('when the request is fetching OpenAPI spec', () => {
     expect(examplesSecondParam).toBe(undefined)
   })
 })
-
```
josemigallas, line 85: 😱

josemigallas:
> In my opinion, this issue could be better solved if we could call a *_url helper instead

Agree. Maybe there's another way to call `_url`? Some other module or something.

Current solution also LGTM.

josemigallas (dismissed):

spec/javascripts/ActiveDocs/OAS3Autocomplete.spec.ts
jlledom: Fixed!

jlledom:
> Agree. Maybe there's another way to call `_url`? Some other module or something.

I don't know if there's an easy way. But I think it's not worth spending more in this because this won't be an issue anymore after we merge https://github.com/3scale/porta/pull/3103

josemigallas (dismissed):

mayorova (dismissed):

PR #3300, opened 2023-04-10
🥒 Disable failing test

josemigallas:
The test that is suddenly failing in master is for a very stupid functionality. When performing bulk updates on accounts, apps and subscriptions, the modal offer a blank option that does nothing. In fact, the server returns an error because this fields is required:
Screenshot 2023-04-10 at 13 01 25
Screenshot 2023-04-10 at 13 01 31

This is a very clumsy implementation, since the select supports a "no blank option" config that could have been set from then beginning. Why the test fails for subscriptions and not for accounts and apps is still a mystery. But considering how convoluted the whole bulk_operations / jquery / tipsy modal implementation is... this should be reimplemented and cleaned up instead of "fixing" this useless test.

### DIFF
--- a/features/old/services/subscriptions/bulk_operations/change_states.feature
+++ b/features/old/services/subscriptions/bulk_operations/change_states.feature
@@ -20,6 +20,7 @@ Feature: Bulk operations
     Given current domain is the admin domain of provider "foo.3scale.localhost"
     Given I am logged in as provider "foo.3scale.localhost"
 
+  @wip
   Scenario: Do nothing
     And I am on the service contracts admin page
     When I follow "Account" within table


akostadinov:
I didn't get it. So test should verify that error will be shown if no action was selected? But in fact happens what?

Can we set the parameter as `required` while on it?

lvillen (approved):

josemigallas:
> I didn't get it. So test should verify that error will be shown if no action was selected? But in fact happens what? 

The test verifies nothing happens. However, when submitting the form, a browser dialog has to be accepted and for some reason capybara can't find it for this very particular case, hence the test fails. I double check it works and the test even passes on local env 🤷🏻‍♂️. It's super hard to debug capybara in circleci so I think it's not worth it to try to fix the test.

> Can we set the parameter as `required` while on it?

The `required` functionality is disabled for the whole formtastic module, setting it on could affect other forms, although I agree this should be done.

akostadinov (approved):

PR #3301, opened 2023-04-10
😢 assets compilation now requires 3GB

akostadinov:


openshift/system/Dockerfile
```diff
@@ -82,7 +82,7 @@ RUN source /opt/app-root/etc/scl_enable \
     && cd /opt/system \
     && npm install -g yarn \
     && yarn install:safe --no-progress \
-    && SECRET_KEY_BASE=rails/32947 NODE_OPTIONS=--max-old-space-size=2560 bundle exec rake assets:precompile tmp:clear \
+    && SECRET_KEY_BASE=rails/32947 bundle exec rake assets:precompile tmp:clear \
```
akostadinov, line 85: this shouldn't be needed, we do without it in CI as the initializer should do the job either way

josemigallas (approved):
I actually use 4G in my local env 😆

lvillen (approved):

PR #3302, opened 2023-04-11
Task for compiling assets for tests

josemigallas:
It's important to set the correct env vars when compiling assets for tests, so here's a task to be run before running tests.


akostadinov (approved):
Approving but I'm concerned we may be missing something simple. Can anybody else also take a look?

lvillen (approved):

PR #3305, opened 2023-04-12
👷⚡️ Change CI cache strategy

josemigallas:


.circleci/config.yml
```diff
@@ -31,6 +31,7 @@ store-log-artifacts: &store-log-artifacts
 npm-cache-key: &npm-cache-key node-v10-{{ checksum "yarn.lock" }}-5
 bundle-cache-key: &bundle-cache-key v2-bundler-gems-{{ .Environment.DB }}-{{ arch }}-{{ .Branch }}-{{ checksum "Gemfile.lock" }}
 assets-cache-key: &assets-cache-key v1-asset-cache-{{ checksum "tmp/assets_related_checksums" }}
+packs-cache-key: &packs-cache-key packs-cache-{{ checksum "tmp/packs_related_checksums" }}
```
akostadinov, line 34: Good to add `v1-` prefix to the key. We can always add some prefix if we want to invalidate the key. Just for consistency with the others. Not mandatory.

.circleci/config.yml
```diff
@@ -47,19 +48,32 @@ save-assets-cache: &save-assets-cache
     key: *assets-cache-key
     paths:
       - public/assets
-      - public/packs-test
       - tmp/cache/assets
-      - tmp/cache/webpacker
```
akostadinov, line 52: I think we needed this in the past to make webpacker faster to identify that everything was present. Do you think it is not needed anymore? Or possibly was never needed?

config/webpacker.yml
```diff
@@ -4,7 +4,7 @@ default: &default
   source_path: app/javascript
   source_entry_path: packs
   public_output_path: packs
-  cache_path: tmp/cache/webpacker
+  cache_path: node_modules/.cache
```
akostadinov, line 7: ah, so webpacker will use cache `node_modules/.cache` but then I don't see where we cache this directory.

josemigallas: I it will be included in the npm-cache, no?

.circleci/config.yml
josemigallas: No idea to be honest 🤷🏻‍♂️

config/webpacker.yml
akostadinov: ah, got your idea. But this `.cache` is generated by webpacker, not node, correct? So node cache can restore something that is not what webpacker would generate. Or no?

PR #3309, opened 2023-04-13
THREESCALE-4145: Unify CDN backends

jlledom:

[THREESCALE-4145](https://issues.redhat.com/browse/THREESCALE-4145)

**Background**

We provide a liquid tag `cdn_asset` which accepts a limited [list of values](https://github.com/jlledom/porta/blob/THREESCALE-4145-unify-cdn/lib/developer_portal/lib/liquid/tags/cdn_asset.rb#L19-L46). Those refer to assets frequently used from developer portals, mainly jQuery, SwaggerUI and FontAwesome libs.

Thanks to this, the buyers can add something like this to their CMS templates:

`{% cdn_asset /jquery/3.5.0/jquery.min.js %}`

That will set jQuery 3.5.0 ready to use in the developer portal. We make the assets available for buyers just by uploading them to [`/cdn`](https://github.com/3scale/porta/tree/6017b4a916df3b027601a4c455124133b2b4acda/cdn). This allows us to update the assets without requiring providers to update their templates .

This feature is called CDN but it's not really a CDN, since the assets are served from our servers. Besides, this is only available from the developer portal, and completely ignores the Rails standard way to configure a CDN by the `asset_host` config parameter (`RAILS_ASSET_HOST` for us).

However, another environment variable is available: `ASSETS_CDN_HOST`. When set, the `cdn_asset` tag will try to download the assets from the `ASSETS_CDN_HOST` instead of our server... but, that can't be a true CDN, because in practice, a CDN receiving a request for `"#{ASSETS_CDN_HOST}/jquery/3.5.0/jquery.min.js"` will try to fetch `"/jquery/3.5.0/jquery.min.js"` from our servers, but those assets won't be there because setting `ASSETS_CDN_HOST` prevents rails routes to be created. In fact, even if they were created, the routes would look like `"/_cdn_assets_/jquery/3.5.0/jquery.min.js"` and the CDN would fail to fetch the assets due to the `/_cdn_assets_` prefix. Summarizing, the only thing you can do with `ASSETS_CDN_HOST` is to copy the assets somewhere else and serve them from there.

All this logic is implemented in a simple way, at [cdn_asset.rb](https://github.com/3scale/porta/blob/013be400daa96cfef89cc5f8d5cdd28b57f6403a/lib/developer_portal/lib/liquid/tags/cdn_asset.rb#L51-L54). For a path like `/jquery/3.5.0/jquery.min.js`, if `ASSETS_CDN_HOST` is set, it renders `"#{ASSETS_CDN_HOST}/jquery/3.5.0/jquery.min.js"`, otherwise it renders `"/_cdn_assets_/jquery/3.5.0/jquery.min.js"`. All requests sent to `/_cdn_assets_` are [redirected](https://github.com/3scale/porta/blob/013be400daa96cfef89cc5f8d5cdd28b57f6403a/config/routes.rb#L28) to a new [Sprockets environment](https://github.com/3scale/porta/blob/013be400daa96cfef89cc5f8d5cdd28b57f6403a/config/routes.rb#L4-L13) pointing to the `/cdn` folder in our root.

**What this PR does**

The issue asks for 3 things, as I understood it:

1. Remove the additional Sprockets environment and user the Rails default one
2. Serve all assets in porta, including the assets under `/cdn`, using the same CDN.
3. Move all `/cdn` assets to a new directory `/dev-portal`.

This PR implements the 3 points above. They are implemented this way:

1. All assets under `/cdn` have been moved to `/public/dev-portal`. This way the default Sprockets environment will find this assets as statics under the `/dev-portal` path. Since now the path is always the same, it can be reached from our server as well as from a CDN.
2. The additional Sprockets environment and the additional routes have been deleted from `routes.rb`. They are not needed anymore.
3. The `cdn_asset` liquid tag now always renders the `/dev-portal` path, no matter whether a CDN is configured or not. When a CDN is configured, it renders an absolute URL instead of a relative URL.
4. The `ASSETS_CDN_HOST` env variable and the `assets_cdn_host` Rails config have been deleted. Now only one way to set a CDN is allowed: by using `RAILS_ASSET_HOST`.
5. I updated the tests to check the new behavior.

**How to review this PR**

- There are more than 50 changed files, but most of them are the assets being moved to another directory. Those files don't need to be reviewed.
- The point `2` above is implemented in `routes.rb`
- The point `3` above is implemented in `cdn_asset.rb`
- The point `4` above is implemented in the files `settings.yml`

**Breaking changes**

- Since `ASSETS_CDN_HOST` has been removed, anybody using it will have to migrate to `RAILS_ASSET_HOST`.
- The old `ASSETS_CDN_HOST` variable expected a full url with the protocol included: `http[s]://cdn.domain.net`. But `RAILS_ASSET_HOST` only accepts the domain: `cdn.domain.net`. For the later, the protocol is [guessed](https://github.com/3scale/porta/blob/013be400daa96cfef89cc5f8d5cdd28b57f6403a/app/helpers/application_helper.rb#L378) from the received HTTP request, to avoid Mixed Content. Maybe it's not the best thing to do, because the system admin must ensure the CDN server will accept the same protocol as porta, but so far nobody complained. For the `cdn_asset` tag, it does exactly the same now.

**Verification steps** 

1. Find a template where `cdn_asset` is being used. E.g: In a default developer portal, the layout `main_layout` includes `{% cdn_asset /jquery/3.5.0/jquery.min.js %}`
2. Restart the rails server with `RAILS_ASSET_HOST` unset.
3. Enter the developer portal, it should load jquery from `http://provider.3scale.localhost:3000/dev-portal/jquery/3.5.0/jquery.min.js`
4. Restart the rails server with `RAILS_ASSET_HOST=cdn.3scale.localhost:3000`
6. Enter the developer portal, it should load jquery from `http://cdn.3scale.localhost:3000/dev-portal/jquery/3.5.0/jquery.min.js`


mayorova:
Great job @jlledom :+1: 
This is working well.

One thing though - we need to update the documentation and maybe add a release note to SaaS (https://access.redhat.com/articles/3107441)

Mainly because we need to make it clear to the customers that they shouldn't use `/dev-portal` path on their developer portal (or at least don't use any of the full paths of our assets, e.g. `/dev-portal/font-awesome/4.3.0/css/font-awesome.css`)
But well, I guess the chances that they do it are pretty low :upside_down_face:

mayorova:
I would suggest maybe to rename the prefix from `/dev-portal` to `/dev-portal-assets` to make it clear that this is only for assets?... WDYT?

jlledom:
> I would suggest maybe to rename the prefix from `/dev-portal` to `/dev-portal-assets` to make it clear that this is only for assets?... WDYT?

That's fine, I pushed the changes

test/unit/liquid/tags/cdn_asset_test.rb
```diff
@@ -2,23 +2,11 @@
 require 'test_helper'
 
 class Liquid::Tags::CdnAssetTest < ActiveSupport::TestCase
-
   test 'gets the correct file name' do
-    Rails.configuration.three_scale.assets_cdn_host = 'https://cdn.3scale.net'
```
akostadinov, line 7: could you also test this scenario, where we do specify the host?

features/step_definitions/asset_host_steps.rb
```diff
@@ -1,25 +1,42 @@
...
+
 Given /^the asset host is set to "(.*)"$/ do |asset_host|
   cdn_url = "#{asset_host}:#{Capybara.current_session.server.port}"
-  Rails.configuration.asset_host = cdn_url
   Rails.configuration.three_scale.asset_host = cdn_url
 end
 
 Then /^((javascript|font)\s)?assets should be loaded from the asset host$/ do |asset_type|
-  cdn_url = Rails.configuration.asset_host
+  cdn_url = Rails.configuration.three_scale.asset_host
```
akostadinov, line 13: maybe assert here that the value is `#present?`?

features/step_definitions/asset_host_steps.rb
```diff
@@ -1,25 +1,42 @@
...
+end
+
+Then /^provided assets should be loaded from the asset host$/ do
+  cdn_url = Rails.configuration.three_scale.asset_host
+  js_regexp = Regexp.new("https?://#{cdn_url}/dev-portal-assets/.*?\\.js")
+
+  assert_not_nil Capybara.page.source.match js_regexp
+end
+
+Then /^provided assets shouldn't be loaded from the asset host$/ do
```
akostadinov, line 37: `provided` assets is not very clear. Maybe a little bit better ame would be `portal` assets?

features/step_definitions/asset_host_steps.rb
```diff
@@ -1,25 +1,42 @@
...
+  # When no CDN is set, we expect to find relative paths to reusable assets
+  js_regexp = Regexp.new('src="/dev-portal-assets/.*?\.js"')
```
akostadinov, line 39: nitpick ,better style would be to use `r%{...}`

app/views/admin/api_docs/base/swagger.html.erb
```diff
@@ -31,10 +31,10 @@
 <% elsif @api_docs_service.specification.swagger_2_0? -%>
 
   <% content_for :javascripts do -%>
-    <%= javascript_include_tag "/_cdn_assets_/swagger-ui/2.2.10/swagger-ui.js" -%>
+    <%= javascript_include_tag "/dev-portal-assets/swagger-ui/2.2.10/swagger-ui.js" -%>
```
akostadinov, line 34: shouldn't we use `cdn_asset ...` here?

akostadinov (commented):
This is a really nice improvement. If prod/support doesn't find issue with backward compatibility I'm all for it.

added some small comments.

features/step_definitions/asset_host_steps.rb
jlledom: Done

features/step_definitions/asset_host_steps.rb
jlledom: I don't know how to call this. I mean specifically the assets that are provided by us, exclusively to be used by clients through the `cdn_asset` tag in their developer portals, but not being part of porta. I think `portal` assets is also confusing because it's too generic. Any other suggestion?

features/step_definitions/asset_host_steps.rb
jlledom: Done

app/views/admin/api_docs/base/swagger.html.erb
jlledom: No because `cdn_asset` is a liquid tag and can only be used in liquid templates.

jlledom:
> If prod/support doesn't find issue with backward compatibility I'm all for it.

I think I could remove the breaking change with a bit of work, I'll do it when I can.

jlledom:
I made some changes in the code to remove the breaking changes. It should work for a client just using the old `ASSETS_CDN_HOST`, although that only works for the developer portal, which was it's original purpose.

We should add a release note saying `ASSETS_CDN_HOST` is deprecated, and encouraging customers to use `RAILS_ASSET_HOST` instead, which works for both admin and developer portals.

On the other hand, now both env variables admit URLs with or without protocol (`http[s]://`).

lib/developer_portal/lib/liquid/tags/cdn_asset.rb
```diff
@@ -48,9 +61,13 @@ def render(context)
 
       private
 
-      def cdn_host
-        hosted_cdn = Rails.configuration.three_scale.assets_cdn_host
-        hosted_cdn ? hosted_cdn : '/_cdn_assets_'
+      def cdn_host(request)
+        rails_config = Rails.configuration.three_scale
+        asset_host_url = rails_config.asset_host.presence || rails_config.assets_cdn_host.presence
```
mayorova, line 66: Hmm... but the config `assets_cdn_host` was removed... (`ASSETS_CDN_HOST` value used as a fallback option for `asset_host` config)

mayorova: And I think `rails_config.asset_host.presence` points to a lambda function?

jlledom: `assets_cdn_host` was removed from our `settings.yml` example, but customer's `settings.yml` could still have it. We access it as fallback if `three_scale.asset_host` is not set.

And yes, `rails_config.asset_host.presence` is a [lambda function](https://github.com/3scale/porta/blob/master/config/environments/production.rb#L112-L119) in production.

mayorova: > And yes, rails_config.asset_host.presence is a lambda function in production.

I guess I don't fully understand how this lambda thing works... Like at what stage does it end up being transformed into the actual URL.

Would it ever match `asset_host_url.match? %r{^https?://}` ?

akostadinov: I don't think this achieves backward compatibility. If one used `ASSETS_CDN_HOST` in the past, it doesn't mean they used the same for `RAILS_ASSET_HOST` and it doesn't mean that the specified  `ASSETS_CDN_HOST` will be capable of returning any other assets besides the bundled assets for dev portal.

I'm not concerned that users may need to start using another variable to set heir asset host. This is not the backward compatibility issue I'm worried about. Rather I'm worried about users that used a  `ASSETS_CDN_HOST` but a different or no `RAILS_ASSET_HOST`.

It makes little (or no) sense to me for anybody to use such a setup. But we have to ask prod team about it.

Additionally if anybody relied on `ASSETS_CDN_HOST`, they would be serving the assets from the old path `/_cdn_assets_/` rather than the current `/dev-portal-assets/`. So any pre-existing `ASSETS_CDN_HOST` would be invalid either way.

So I don't see any benefit of this commit. My suggestion is to remove this commit. All the rest look fine.

If users had a `RAILS_ASSET_HOST`, then their CDN requirements would be probably still met because that host will be able to cache and serve the built-in portal assets in the same way it does for the rest of the assets. (I assume)

The only group of users that may suffer is if they had  `ASSETS_CDN_HOST` but didn't have `RAILS_ASSET_HOST`. In this case though, if we use that configuration for assets, no assets would be likely to work because of the changed path. But if we ignore the  `ASSETS_CDN_HOST` configuration, then assets will be loaded directly from porta so they should work unless user relied on modifying the assets on their external server.

So IMO if you remove this commit, there would be fewer compatibility issues than if you keep it.

features/asset_host/asset_host.feature
```diff
@@ -9,23 +9,47 @@ Feature: Asset host
...
       Then assets should be loaded from the asset host
 
+  Rule: Developer
+    Background:
+      Given a provider exists
+      And the provider has one buyer
+
+    Scenario: Developer portal with asset host not configured
+      Given the asset host is unset
+      And the buyer logs in to the provider
```
josemigallas, line 41:
```suggestion
      When the buyer logs in to the provider
```

features/asset_host/asset_host.feature
```diff
@@ -9,23 +9,47 @@ Feature: Asset host
...
+      Then javascript assets shouldn't be loaded from the asset host
+      And provided assets shouldn't be loaded from the asset host
+
     Scenario: Developer portal with asset host configured
-      When the buyer logs in to the provider
+      Given the asset host is set to "cdn.3scale.localhost"
+      And the buyer logs in to the provider
```
josemigallas, line 47:
```suggestion
      When the buyer logs in to the provider
```

features/asset_host/asset_host.feature
```diff
@@ -9,23 +9,47 @@ Feature: Asset host
...
+      Then javascript assets should be loaded from the asset host
+      Then provided assets should be loaded from the asset host
```
josemigallas, line 49:
```suggestion
      And provided assets should be loaded from the asset host
```

features/asset_host/asset_host.feature
```diff
@@ -9,23 +9,47 @@ Feature: Asset host
...
+
+    Scenario: Developer portal with asset host configured as full URL
+      Given the asset host is set to "http://cdn.3scale.localhost"
+      And the buyer logs in to the provider
```
josemigallas, line 53:
```suggestion
      When the buyer logs in to the provider
```

test/unit/helpers/application_helper_test.rb
```diff
@@ -26,36 +26,45 @@ class AssetHostTest < ApplicationHelperTest
     setup do
       @request = ActionDispatch::TestRequest.create
       @asset_host = 'cdn.3scale.test.localhost'
+      @full_asset_host = 'https://cdn.3scale.test.localhost'
```
josemigallas, line 29:
```suggestion
      @full_asset_host = "https://#{@asset_host}"
```

josemigallas (commented):
❤️ I like this a lot. Not gonna approve just yet because I've only reviewed the code.

test/unit/helpers/application_helper_test.rb
```diff
@@ -26,36 +26,45 @@ class AssetHostTest < ApplicationHelperTest
...
     end
 
     attr_reader :request
 
     test 'asset host is not configured' do
-      Rails.configuration.expects(:asset_host).returns(nil)
-      Rails.configuration.three_scale.expects(:asset_host).returns(@asset_host)
+      Rails.configuration.stubs(:asset_host).returns(nil)
```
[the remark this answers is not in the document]
jlledom: Because `expects` expects the method to be called once, `stubs` doesn't care about number of calls

lib/developer_portal/lib/liquid/tags/cdn_asset.rb
jlledom: > I guess I don't fully understand how this lambda thing works... Like at what stage does it end up being transformed into the actual URL.

Because it's a hell. There are three ways a CDN url can be generated based on what kind of assets is:

- Sprokets:
  - For `javascript_include_tag`, CSSs, images, etc.
  - It prepends `Rails.configuration.asset_host` to the given value
  - `Rails.configuration.asset_host` (at `production.rb`) instead of being just a hostname, is a lambda function that returns `Rails.configuration.three_scale.asset_host` when the given resource actually exists in our assets public folder (don't ask me why). Otherwise it `breaks`, which I don't know the effects of, but I assume the asset loading is cancelled.
-  Webpacker:
    - For `javascript_pack_tag`
        - It uses the same pipeline as Sprockets, so `Rails.configuration.asset_host`.
    - For fonts:
        - CSS styles for fonts are generated on runtime, it prepends `rails_asset_host_url` (at `application_helper.rb`) to the `url()` attribute.
- (New) For developer portal "CDN" assets: It prepends
  - For `cdn_asset` tag in liquid templates.
  - Prepends `cdn_host` (at `cdn_asset.rb`) to the given value.

> Would it ever match asset_host_url.match? %r{^https?://} ?

The lambda function intervenes in a workflow where this never gets executed.

mayorova: Thanks for the explanation @jlledom !

Actually, my confusion was caused by my inability to read properly :clown_face: 
I thought `rails_config` was `Rails.configuration`, while it's actually `Rails.configuration.three_scale` :upside_down_face: 

So, maybe I'd just suggest to rename the variable to `threescale_config`.

jlledom: @mayorova done!

jlledom: @akostadinov 

> Rather I'm worried about users that used a ASSETS_CDN_HOST but a different or no RAILS_ASSET_HOST.

Yeah, this doesn't really work. I think I'm gonna revert it. Any way here's a summary of the behavior before and after this PR, just for reference.

Before:
|RAILS_ASSET_HOST|ASSETS_CDN_HOST|Result|
|---|---|---|
|Empty|Empty| Admin portal assets from `/public`, Developer portal assets from `/cdn` |
|Empty|Set| Admin portal assets from `/public`, Developer portal assets from `ASSETS_CDN_HOST` |
|Set|Empty| Admin portal assets from `RAILS_ASSET_HOST` as cache for `/public`, Developer portal assets from `/cdn`  |
|Set|Set| Admin portal assets from `RAILS_ASSET_HOST` as cache for `/public`, Developer portal assets from `ASSETS_CDN_HOST` |

After:
|RAILS_ASSET_HOST|ASSETS_CDN_HOST|Result|
|---|---|---|
|Empty|Empty| Admin portal assets from `/public`, Developer portal assets from `/public/dev-portal-assets` |
|Empty|Set| Admin portal assets from `/public`, Developer portal assets from `ASSETS_CDN_HOST/dev-portal-assets` as cache  for `/public/dev-portal-assets` |
|Set|Empty| Admin portal assets from `RAILS_ASSET_HOST` as cache for `/public`, Developer portal assets from `RAILS_ASSET_HOST/dev-portal-assets` as cache  for `/public/dev-portal-assets`  |
|Set|Set| Admin portal assets from `RAILS_ASSET_HOST` as cache for `/public`, Developer portal assets from `RAILS_ASSET_HOST/dev-portal-assets` as cache  for `/public/dev-portal-assets` |

And if we revert this commit (`ASSETS_CDN_HOST` would be ignored):
|RAILS_ASSET_HOST|Result|
|---|---|
|Empty| Admin portal assets from `/public`, Developer portal assets from `/public/dev-portal-assets` |
|Set| Admin portal assets from `RAILS_ASSET_HOST` as cache for `/public`, Developer portal assets from `RAILS_ASSET_HOST/dev-portal-assets` as cache  for `/public/dev-portal-assets` 

> Additionally if anybody relied on ASSETS_CDN_HOST, they would be serving the assets from the old path /_cdn_assets_/ rather than the current /dev-portal-assets/. So any pre-existing ASSETS_CDN_HOST would be invalid either way.

This is not exactly right, if anybody relied on `ASSETS_CDN_HOST` they must have a copy of their files in a server on `ASSETS_CDN_HOST`, and `/_cdn_assets_/` would be completely ignored in that case.

jlledom:
@josemigallas 

> Even though it's not exactly a CDN, I would just keep calling it CDN assets to be consistent with the method. Alternatively I think developer portal assets is still a better name than provided assets.

I renamed it to cdn assets

lib/developer_portal/lib/liquid/tags/cdn_asset.rb
jlledom: Yes, I reverted that commit

jlledom:
I pushed some changes in order to simplify and unify the CDN management a bit more. I think all the CDN issue in porta is overly complicated and difficult to maintain. These are the changes I made in order to simplify:

- Stop loading the assets host only `if config.assets.digest`:
  - This was added only to simplify the transition between rails 3 and 4, it's not required anymore and I removed it.
  - https://github.com/3scale/system/pull/5558
- config.asset_host as a lambda function:
  - The lambda function returns the CDN url to be prepended before the requested resource path, and it returns nothing when the requested resource is not in the file system, that causes the asset to be loaded from the server instead of the CDN. I failed to find the reason behind this. I assume the reason is to allow sprockets or webpack to find the resource even when not in the file system. E.g. sprockets and webpack can translate non-digested file names to digested ones in the filesystem, but AFAIK this kind of situations are already managed by rails, we don't need to implement the hack ourselves. This lambda function is *very* old code and it's probably not required anymore, but it's making the CDN management overly complicated. I prefer to delete it unless I find a true reason to keep it.

lib/developer_portal/lib/liquid/tags/cdn_asset.rb
```diff
@@ -62,9 +62,11 @@ def render(context)
       private
 
       def cdn_host(request)
-        return '' unless Rails.configuration.three_scale.asset_host.presence
+        asset_host_url = Rails.configuration.three_scale.asset_host.presence
+        return '' unless asset_host_url
+        return asset_host_url if asset_host_url.match? %r{^https?://}
 
-        "#{request.protocol}#{Rails.configuration.three_scale.asset_host}"
+        "#{request.protocol}#{asset_host_url}"
```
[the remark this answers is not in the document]
jlledom: I pushed some changes the simplify the whole thing, including unifying this two methods.

config/environments/development.rb
```diff
@@ -17,12 +17,7 @@
   config.assets.digest = false
   config.assets.precompile += %w( spec_helper.js )
 
-  config.asset_host = ->(*args) {
-    _source, request = args
-
-    return unless request
-    request.headers['host']
-  }
+  config.asset_host = config.asset_host = config.three_scale.asset_host.presence
```
mayorova, line 20: Is `config.asset_host = config.asset_host` a typo?

mayorova:
It's great to see a significant simplification of the code.
I am a bit worried if we might miss some scenario that would not be supported with this updatd code... 

Initially, the lambda for asset host was added here it seems https://github.com/3scale/system/pull/4628/files

And the reason was that developer portals tried to load [some JS files through the CDN](https://github.com/3scale/system/issues/4623), but it was not there (because it's a custom JS file in the dev portal's CMS).

But uploading custom JS files works OK, I think (using `{{ '/javascripts/my-custom.js' | javascript_include_tag }}`)

I'm not sure, maybe it's to support some other way to include JS files... hm...

mayorova:
@jlledom I think we need to change the CodeClimate config:
["cdn/*"](https://github.com/3scale/porta/blob/master/.codeclimate.yml#L38) => `"public/dev-portal-assets/*"`

jlledom:
> @jlledom I think we need to change the CodeClimate config: ["cdn/*"](https://github.com/3scale/porta/blob/master/.codeclimate.yml#L38) => `"public/dev-portal-assets/*"`

Good catch! I fixed it

jlledom:
> Initially, the lambda for asset host was added here it seems https://github.com/3scale/system/pull/4628/files
> 
> And the reason was that developer portals tried to load [some JS files through the CDN](https://github.com/3scale/system/issues/4623), but it was not there (because it's a custom JS file in the dev portal's CMS).

Thanks for finding this.

> But uploading custom JS files works OK, I think (using `{{ '/javascripts/my-custom.js' | javascript_include_tag }}`)
> 
> I'm not sure, maybe it's to support some other way to include JS files... hm...

I've been taking a look at it and I see the same results as you, removing the lambda function doesn't make any difference when loading a custom dev portal js, whether is `RAILS_ASSET_HOST` set or not. Neither whether are digests enabled or not. Right now it doesn't come to my mind any other situation where the lambda function could be useful.

When that PR was merged, 3scale was using rails 3, from 2014. Maybe by that time sprockets wasn't good enough to manage our particular case but it is today.

lib/developer_portal/lib/liquid/tags/cdn_asset.rb
```diff
@@ -13,45 +13,51 @@ class CdnAsset < Liquid::Tags::Base
...
+        SwaggerUI 2.1.3
+        - swagger-ui/2.1.3/swagger-ui.js
+
+        SwaggerUI 2.2.10
+        - swagger-ui/2.2.10/swagger-ui.js
+      )
 
       def render(context)
         action_view = context.registers[:controller].view_context
+        file_url = File.join(context.registers[:controller].helpers.rails_asset_host_url, @file)
```
mayorova, line 50: Maybe extract `context.registers[:controller]` to a variable to make CodeClimate happy?

jlledom: Done!

mayorova (approved):

lib/developer_portal/lib/liquid/tags/cdn_asset.rb
```diff
@@ -13,45 +13,52 @@ class CdnAsset < Liquid::Tags::Base
 
       def initialize(tag_name, file, tokens)
         super
-        @file = File.join(cdn_host, file).strip # strip because we are allowing the path not enclosed between quotes
+        @file = File.join('/dev-portal-assets', file).strip # strip because we are allowing the path not enclosed between quotes
```
akostadinov, line 16: Can we set host here as well? 

```suggestion
        @file = File.join(controller.helpers.rails_asset_host_url, '/dev-portal-assets', file).strip # strip because we are allowing the path not enclosed between quotes
```

akostadinov (approved):
Great work!

I added two comments but I think they can be tried as a separate PR if they make sense to begin with.

app/helpers/application_helper.rb
```diff
@@ -370,10 +370,9 @@ def rails_asset_url_tag
...
 
   def rails_asset_host_url
-    asset_host_enabled = Rails.configuration.asset_host.present?
     asset_host_url = Rails.configuration.three_scale.asset_host.presence
-
-    return '' unless asset_host_enabled && asset_host_url
+    return '' unless asset_host_url
+    return asset_host_url if asset_host_url.match? %r{^https?://}
 
     "#{request.protocol}#{asset_host_url}"
```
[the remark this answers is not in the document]
mayorova: In theory, nobody should access over http, but there are some hacks that customers can use to get http working https://access.redhat.com/solutions/3455941

I don't think we need to close that possibility completely.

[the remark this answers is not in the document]
akostadinov: would such customer need to also access assets over `http` and at the same time specify a host without schema? I think that for customers insisting on http for assets, they can just specify a schema as well.

lib/developer_portal/lib/liquid/tags/cdn_asset.rb
jlledom: No, we can't, because at this point we don't have access to the `request`, and the request it's required to automatically append the protocol (`https?:`) when it's not provided.

app/helpers/application_helper.rb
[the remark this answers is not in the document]
jlledom: @akostadinov I don't understand your question, but here's how it works:

- If `RAILS_ASSETS_HOST` specify a protocol, that one is used to generate assets urls.
- Otherwise, the same protocol being used by porta is used for assets.

Does that help?

lib/developer_portal/lib/liquid/tags/cdn_asset.rb
akostadinov: This is the point - to hardcode `https`. If somebody so much needs `http`, they can setup the value to include the desired schema.

mayorova (approved):

mayorova (approved):

app/helpers/application_helper.rb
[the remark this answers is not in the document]
akostadinov: My point was that porta request schema doesn't have to be same as the assets schema. So it makes sense to hardcode https here like we do in porta configuration. And if one wants to use `http`, they can configure it explicitly.

PR #3311, opened 2023-04-14
Update Rails to 6.1

thalesmiguel:

This PR attempts to update Porta's Rails version to 6.1.x. It also tries to update the minimum amount of gems necessary for the Rails update.

**Verification steps** 

When I first attempted to update Rails to 6.0.x, I noticed that there shouldn't be too much of a difference compared with the update to 6.1.x, therefore I pushed for a bigger update.
Also, 6.0.x's EOL will happen in June 1st, 2023.


Rakefile
```diff
@@ -58,8 +58,22 @@ end
 # connection failure and thus failure to create the database.
 # Since we don't use credentials, we can remove that dependency.
 # And then a whole mess to fix other use cases especally db:reset
-warn "Removing :environment prerequisite from db:create"
-Rake::Task['db:load_config'].prerequisites.delete("environment")
+warn "Removing :environment prerequisite from db:load_config"
+Rake::Task["db:load_config"].clear
+# Extracted from https://github.com/rails/rails/blob/6-1-stable/activerecord/lib/active_record/railties/databases.rake#L21-L27
+# As we remove :environment from db:load_config, the database_configuration might be nil.
```
akostadinov, line 64: Seems like in 6.1 they are increasingly relying on `environment` being loaded before everything :/

Maybe it would be simpler to add back what was removed previously in rails/rails@2beeb6fbd30be9522f41d03effe256e22d0445ea

```ruby
ActiveRecord::Tasks::DatabaseTasks.database_configuration = Rails.application.config.database_configuration
```

As a prerequisite of `load_config` task. probably checking if `ActiveRecord::Tasks::DatabaseTasks.database_configuration` is empty first. e.g. something like:

```ruby
  if ActiveRecord::Tasks::DatabaseTasks.database_configuration.blank?
    ActiveRecord::Tasks::DatabaseTasks.database_configuration = Rails.application.config.database_configuration || {}
  end
```

WDYT?

thalesmiguel: Looks better indeed. This way, we don't need to override the `db:load_config` rake task :v:
Updated in 119e313.

jlledom:
Should we close this?

thalesmiguel:
Closed in favour of https://github.com/3scale/porta/pull/3336.

PR #3322, opened 2023-04-21
🥒 Enable javascript in cukes

josemigallas:
**NOTE**: only add `@javascript` to scenarios logging in to a provider

Javascript is disabled for most cucumbers and it creates some false positives. For instance, some forms have HTML5 validations that prevent them from being submitted, but it won't work unless `@javscript` tag is used.

On the other hand the login page has a no-javascript alternative (that nobody uses because porta needs JS to work) that will be removed in #3316, and will be a React rendered component, so @javascript has to be enabled.

### DIFF
--- a/app/inputs/price_input.rb
+++ b/app/inputs/price_input.rb
@@ -16,7 +16,7 @@ class PriceInput < Formtastic::Inputs::NumberInput
   end
 
   def input_html_options
-    super.merge({ value: value, min: 0, step: 0.01 })
+    super.merge({ value: value, step: 0.001 })
   end
 
   private
--- a/features/api/services/api_docs.feature
+++ b/features/api/services/api_docs.feature
@@ -33,7 +33,7 @@ Feature: ActiveDocs
       | swagger version |
       | Swagger 1.2     |
       | Swagger 2       |
-  # | OAS 3.0         | Feature not implemented |
+      # | OAS 3.0         | Feature not implemented |
 
   @javascript
   Scenario Outline: Slashes generated curl command for header values
@@ -45,7 +45,7 @@ Feature: ActiveDocs
       | swagger version |
       | Swagger 1.2     |
       | Swagger 2       |
-  # | OAS 3.0         | Feature not implemented |
+      # | OAS 3.0         | Feature not implemented |
 
   @javascript
   Scenario Outline: Create a second spec
--- a/features/api/services/api_docs.feature
+++ b/features/api/services/api_docs.feature
@@ -1,3 +1,4 @@
+@javascript
 Feature: ActiveDocs
   As a provider
   I want to manage my ActiveDocs
@@ -5,7 +6,6 @@ Feature: ActiveDocs
   Background:
     Given a provider is logged in
 
-  @javascript
   Scenario Outline: Create a spec for the first time
     Given an admin wants to add a spec to a new service
     When they are reviewing the service's active docs
@@ -24,8 +24,7 @@ Feature: ActiveDocs
     When an admin is reviewing the service's active docs
     Then the table should not contain a column for the service
 
-  @javascript 
-  Scenario Outline: Autocomplete 
+  Scenario Outline: Autocomplete
     Given a service with a <swagger version> spec
     When an admin is reviewing the spec
     And the swagger autocomplete should work for "user_key" with "user_keys"
@@ -34,19 +33,19 @@ Feature: ActiveDocs
       | swagger version |
       | Swagger 1.2     |
       | Swagger 2       |
-      # | OAS 3.0         | Feature not implemented |
+  # | OAS 3.0         | Feature not implemented |
 
   @javascript
   Scenario Outline: Slashes generated curl command for header values
     Given a service with a <swagger version> spec
     When an admin is reviewing the spec
-    Then <swagger version> should escape properly the curl string  
+    Then <swagger version> should escape properly the curl string
 
     Examples:
       | swagger version |
       | Swagger 1.2     |
       | Swagger 2       |
-      # | OAS 3.0         | Feature not implemented |
+  # | OAS 3.0         | Feature not implemented |
 
   @javascript
   Scenario Outline: Create a second spec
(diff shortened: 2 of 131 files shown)


lvillen (commented):
Some tests are failing, but overall looks good, when tests are green count with my approval 👍

app/inputs/price_input.rb
```diff
@@ -16,7 +16,7 @@ def to_html
   end
 
   def input_html_options
-    super.merge({ value: value, min: 0, step: 0.01 })
+    super.merge({ value: value, step: 0.001 })
```
akostadinov, line 19: why this change?

josemigallas: _fixes price input so that tests pass_

According to tests, this is what price input should accept.

features/support/invoices_helpers.rb
```diff
@@ -1,11 +1,13 @@
...
+
 module InvoicesHelpers
   def create_invoice(buyer, date, opts = {})
     options = opts.reverse_merge(buyer_account: buyer, period: Month.new(date), creation_type: :background)
     buyer.provider_account.billing_strategy.create_invoice!(options)
   end
 
   def assert_secure_invoice_pdf_url(url, invoice)
-    assert_equal invoice.pdf.expiring_url, url
+    assert url.ends_with?(invoice.pdf.expiring_url)
```
akostadinov, line 10: why no match anymore?

features/step_definitions/security/generic_steps.rb
```diff
@@ -1,5 +1,5 @@
+# frozen_string_literal: true
+
 Then /^I should be denied the access$/ do
-  # assert has_content?('Access Denied')
-  assert_equal 403, page.status_code
+  assert_content 'Access Denied'
```
akostadinov, line 4: Also could you explain this?

features/step_definitions/provider_steps.rb
```diff
@@ -147,6 +147,7 @@ def setup_provider(login)
   step 'I log in as provider "foo.3scale.localhost"' if login
 
   @provider = Account.find_by_domain!('foo.3scale.localhost')
+  @service ||= @provider.default_service
```
akostadinov, line 150: Can service be already set? Maybe just `=`?

features/step_definitions/menu_steps.rb
```diff
@@ -1,21 +1,23 @@
...
-  items.raw.each do |item|
-    within '#mainmenu' do
-      assert has_css? 'li', :text => item[0]
-    end
+  within '#mainmenu' do
+    assert_equal items.raw.flatten, find_all('#mainmenu .pf-c-nav__list > .pf-c-nav__item > .pf-c-nav__link').map(&:text)
   end
 end
 
-Then /^I should not see menu items$/ do |items|
```
akostadinov, line 9: This step not needed anymore?

features/step_definitions/finance/settings_steps.rb
```diff
@@ -19,7 +19,6 @@
 
 Then "they should not be able to review the charging and gateway billing settings" do
   visit admin_finance_settings_path
-  assert_equal 403, status_code
```
akostadinov, line 22: this stopped working somehow?

features/step_definitions/apicast_steps.rb
```diff
@@ -13,5 +13,5 @@
 
 Given(/^the default proxy does not use apicast configuration driven$/) do
   proxy = @provider.default_service.proxy
-  proxy.update!(apicast_configuration_driven: false, sandbox_endpoint: 'https://api-2.staging.apicast.io:4443')
+  proxy.update!(apicast_configuration_driven: false, sandbox_endpoint: 'https://api-2.staging.apicast.io:4443', endpoint: 'https://api-2.staging.io:4443')
```
akostadinov, line 16: would be interesting to know why this change was needed

akostadinov (commented):
Looks nice. besides 2 nitpicks, most comments are just asking to explain. Thank you!

lvillen (approved):

features/step_definitions/finance/settings_steps.rb
josemigallas: Yes, `status_code` is not meant for browser tests so it's not available anymore.

features/step_definitions/menu_steps.rb
josemigallas: I had to fix the other step `I should see menu items` and now it will fail unless _all_ menu items match, so it already assert missing menu items.

akostadinov: maybe call the other step `I should see only menu items`

features/step_definitions/provider_steps.rb
josemigallas: Good point. I don't think it should be set here actually. This `@service` is reserved for steps that refer to a service. This one is about the provider.

features/step_definitions/security/generic_steps.rb
josemigallas: Selenium does not support `status_code`: https://github.com/seleniumhq/selenium-google-code-issue-archive/issues/141

features/api/services/api_docs.feature
```diff
@@ -34,19 +33,19 @@ Feature: ActiveDocs
       | swagger version |
       | Swagger 1.2     |
       | Swagger 2       |
-      # | OAS 3.0         | Feature not implemented |
+  # | OAS 3.0         | Feature not implemented |
```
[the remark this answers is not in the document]
josemigallas: It seems to be a bug in the formatter, it works all right with commented table rows in the middle. I'll fix it manually.

features/step_definitions/apicast_steps.rb
josemigallas: Indeed... the error was "Base URL can't be blank" but I wonder why the validation did not work before.

akostadinov: possible that in the past that attribute was already set by previous steps?

josemigallas: I don't think any step is being removed by this PR 🤔

features/step_definitions/menu_steps.rb
josemigallas: That's `I should see menu items` 😄

PR #3323, opened 2023-04-24
Upgrade Rmagick to 4.3.0

jlledom:

I upgraded to Fedora 38 and I found I can't run porta anymore. The cause is we are using rmagick 2.15.4 which requires ImageMagick 6 and doesn't support ImageMagick 7. But Fedora 38 only provides ImageMagick 7, version 6 is not available.

I had to upgrade rmagick to > 4.1 where support for ImageMagick 7 was added. In my tests, this doesn't seem to cause any issue. Why were we stuck on rmagick 2.x?

I need this to get merged because otherwise I'll have to manually update the Gemfile in all my branches but never commit it, and that's very inconvenient.

This upgrade shouldn't affect others because Rmagick 4 still supports ImageMagick 6. But I'd like you to test this and verify it works fine on Mac and older Fedora releases.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -76,7 +76,7 @@ gem 'cancancan', '~> 2.3.0'
 gem 'formtastic', '~> 2.3.1'
 gem 'gruff', '~>0.3', require: false
 gem 'htmlentities', '~>4.3', '>= 4.3.4'
-gem 'rmagick', '~> 2.15.3', require: false
+gem 'rmagick', '~> 4.1', require: false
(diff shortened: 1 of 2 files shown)


akostadinov:
lets replace this by merging #3065 and #3165

jlledom:
> lets replace this by merging https://github.com/3scale/porta/pull/3065 and https://github.com/3scale/porta/pull/3165

OK. Closing.

PR #3334, opened 2023-04-28
⚡ Remove react-icons chunk

josemigallas:
Import icons individually vs. import the whole module.

Before:
icons-before

After:
icons-after


akostadinov (approved):
Nice! I assume this prevents downloads of hundreds if not thousands of icons (ok 1737 minus the handful of them that we use).

lvillen (approved):

PR #3341, opened 2023-05-02
💄➗ Use calc() for math operations in scss files

lvillen:

This PR deals with:
`Deprecation Warning: Using / for division outside of calc() is deprecated and will be removed in Dart Sass 2.0.0.`
As recommended by sass, now our webpack-handled-css files will be using `calc()` method for divisions, keeping webpack happy and green 🍏. 

**Verification steps** 

Check that when running webpacker there aren't any warnings.


josemigallas (approved):
Well done,🧉!

PR #3352, opened 2023-05-04
❤️🐒 Maintain webpackers' config monkey-patching but use some defaults

josemigallas:
Inspired by #3275 

But basically stripped off every braking change. The default config used by `webpacker 5` seems to be worse than what we have. And what we have works all right, so we're gonna leave it like that until we upgrade to `webpacker 6`.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,8 +1,0 @@
-{
-  "plugins": [
-  ],
-  "presets": [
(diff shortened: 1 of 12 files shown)


jlledom (approved):
I tested this locally and it works fine, also the CDN.

lvillen (approved):

PR #3355, opened 2023-05-08
add a little more auditing

akostadinov:
This is the part of #3342 that is ready to merge.

### DIFF
--- a/app/models/backend_api.rb
+++ b/app/models/backend_api.rb
@@ -5,6 +5,8 @@ class BackendApi < ApplicationRecord
   include SystemName
   include ProxyConfigAffectingChanges::ModelExtension
 
+  audited :allow_mass_assignment => true
+
   define_proxy_config_affecting_attributes :private_endpoint
 
   self.background_deletion = %i[proxy_rules metrics backend_api_configs]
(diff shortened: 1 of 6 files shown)


app/models/backend_api.rb
```diff
@@ -5,6 +5,8 @@ class BackendApi < ApplicationRecord
   include SystemName
   include ProxyConfigAffectingChanges::ModelExtension
 
+  audited :allow_mass_assignment => true
```
jlledom, line 8: what is this for?

app/services/audit_log_service.rb
```diff
@@ -0,0 +1,7 @@
+class AuditLogService
+  include Callable
+
+  def self.call(message)
+    Rails.logger.info "[AUDIT]: #{message}"
+  end
+end
```
jlledom, line 7: No new line at the end of the file!

app/models/backend_api.rb
akostadinov: One audit for all changes, not one per change. We use it everywhere.

app/services/audit_log_service.rb
akostadinov: It's a good practice. I don't know why codeclimate doesn't complain about it. Fixed.

jlledom (approved):

PR #3356, opened 2023-05-09
THREESCALE-9458 Trial period not cleaning when changing plan

nidhi-soni1104:

Customer change the plan from one with trial period to another without trial period, but the trial warning still appears on the application.

https://issues.redhat.com/browse/THREESCALE-9458

Note: Video is attached which shows how to reproduce issue

### DIFF
--- a/app/controllers/provider/admin/applications_controller.rb
+++ b/app/controllers/provider/admin/applications_controller.rb
@@ -86,8 +86,18 @@ class Provider::Admin::ApplicationsController < FrontendController
     # there is no need to query available_application_plans as we already have a validation
     new_plan = accessible_plans.stock.find(plan_id)
     @cinstance.provider_changes_plan!(new_plan)
+    update_trial_period(new_plan)
     flash[:notice] = "Plan changed to '#{new_plan.name}'."
     redirect_to provider_admin_application_url(@cinstance)
+  end
+
+  def update_trial_period(new_plan)
+    if new_plan.trial_period_days.nil?
+      @cinstance.trial_period_expires_at = Time.zone.now
+    else
+      @cinstance.trial_period_expires_at = Time.zone.now + new_plan.trial_period_days.to_i.days
+    end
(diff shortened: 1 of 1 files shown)


akostadinov (changes requested):
JIRA issue only mentions a transition between a plan with a trial period and a plan without.

The changes here also change behavior for switching to another plan with trial period. And now the period will be set to the new plan trial days.

I'm not sure this is correct behavior. Usually one gets a trial period once. Then this is not extended by switching to a new plan.

Or if the provider wishes to grant a customer access to another plan during the trial.

So I think we probably need a separate checkbox, e.g. `reset trial period`, that will instruct porta, depending on what the user chose, to reset the period or keep it.

Hope this makes sense. Also if you have discussed how things should behave with PM, then let me know.

PS Also we need tests for this functionality.

nidhi-soni1104:
> > JIRA issue only mentions a transition between a plan with a trial period and a plan without.
> > The changes here also change behavior for switching to another plan with trial period. And now the period will be set to the new plan trial days.
> > I'm not sure this is correct behavior. Usually one gets a trial period once. Then this is not extended by switching to a new plan.
> > Or if the provider wishes to grant a customer access to another plan during the trial.
> > So I think we probably need a separate checkbox, e.g. `reset trial period`, that will instruct porta, depending on what the user chose, to reset the period or keep it.
> > Hope this makes sense. Also if you have discussed how things should behave with PM, then let me know.
> > PS Also we need tests for this functionality.
> 
> Yes agree your points are true but when I saw video shared by client she is having multiple plan with trial and we can even update trial days of any plan even after assigning it to any application. so I think we can have multiple plan with different trial days, otherwise its a basic validation which would have added at the initial phase of the feature.

I will add a comment on the JIRA mentioning this but what I found while checking app is this behavior only

nidhi-soni1104:
Closing as its not an issue

PR #3359, opened 2023-05-15
Fix CDN cucumbers

jlledom:
Some tests were wrong, somehow they didn't fail in their branch. ¯\_(ツ)_/¯

### DIFF
--- a/features/step_definitions/asset_host_steps.rb
+++ b/features/step_definitions/asset_host_steps.rb
@@ -38,7 +38,7 @@ Then /^cdn assets should be loaded from the asset host$/ do
   assert_not_nil Capybara.page.source.match js_regexp
 end
 
-Then /^provided assets shouldn't be loaded from the asset host$/ do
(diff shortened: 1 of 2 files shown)


josemigallas (approved):

features/step_definitions/asset_host_steps.rb
```diff
@@ -38,7 +38,7 @@
   assert_not_nil Capybara.page.source.match js_regexp
 end
 
-Then /^provided assets shouldn't be loaded from the asset host$/ do
+Then /^cdn assets shouldn't be loaded from the asset host$/ do
   # When no CDN is set, we expect to find relative paths to reusable assets
   js_regexp =  %r{src="/dev-portal-assets/.*?\.js"}
```
akostadinov, line 43: btw in all negative tests I would prefer also to have a positive part. For example make sure that assets are loaded just with a different path/url. Because if we look at the wrong page for some reason, the negative test may pass but because page was wrong, not because functionality was right.

PR #3366, opened 2023-05-18
🦋 Use `isWidthLimited` instead of `max-width`

josemigallas:
Patternfly React 4 introduced this property.

[THREESCALE-9643: Remove hardcoded max-width in favor of isWithLimited](https://issues.redhat.com/browse/THREESCALE-9643)


akostadinov (approved):
Looks good to standardize that.

I guess you checked but why backend form needs padding removed while other do not?

josemigallas:
> Looks good to standardize that.
> 
> I guess you checked but why backend form needs padding removed while other do not?

Ah! Good catch, @akostadinov. That's a completely different issue that I am working on in a different branch, but it seems some of leaked here.

akostadinov (approved):

PR #3389, opened 2023-05-29
⚰️ Remove unused module Settings 🔥 

josemigallas:
This PR was originated after some investigation for [THREESCALE-9636: Replace FormField with Patternfly's component](https://issues.redhat.com/browse/THREESCALE-9636) and its sibling [THREESCALE-9637: Replace FormLegend with Patternfly's component](https://issues.redhat.com/browse/THREESCALE-9637).

This React version of then settings page was implemented a long time ago, it's pre-Patternfly and it doesn't seem we ever had a real need (or the interest) to use it so it is best to remove it from the repo.

In case we implement the Integration Settings page we would use Patternfly React which would probably need another considerable refactor.


akostadinov (approved):
:fire: :heart_on_fire: :fire:

lvillen (approved):

PR #3396, opened 2023-05-31
THREESCALE-8036 fixed traffic analytics chart to show integers on y axis

nidhi-soni1104:

https://issues.redhat.com/browse/THREESCALE-8036

**Note for reviewer** **PR IS NOT READY FOR MERGE**

This JIRA demands fix on there pages of analytics
1. Traffic
2. Daily Averages
3. Hourly Averages

Fix will be similar kind of fix so I fixed one page first if looks good I can move to other pages

Here I am attaching screenshots of before and after for clarity

**Before Screenshot**
image

**After Screenshot**
image

### DIFF
--- a/app/javascript/src/Stats/lib/chart.js
+++ b/app/javascript/src/Stats/lib/chart.js
@@ -29,8 +29,9 @@ export class StatsChart {
             }
           },
           y: {
+            default: [0, 10],
             tick: {
-              format: d => numeral(d).format('0.0a').toUpperCase()
+              format: d => numeral(d).format(d < 1000 ? '0' : '0.0a').toUpperCase()
             }
           }
         },
--- a/app/javascript/src/Stats/lib/chart.js
+++ b/app/javascript/src/Stats/lib/chart.js
@@ -31,7 +31,7 @@ export class StatsChart {
           y: {
             default: [0, 10],
             tick: {
-              format: d => numeral(d).format(d < 1000 ? '0' : '0.0a').toUpperCase()
+              format: d => numeral(d).format('0[.]0a').toUpperCase()
             }
           }
         },
--- a/app/javascript/src/Stats/lib/metrics_selector.js
+++ b/app/javascript/src/Stats/lib/metrics_selector.js
@@ -29,7 +29,7 @@ export class StatsMetricsSelector extends StatsUI {
     return (
       <div className={`StatsSelector ${this.open ? 'is-open' : ''}`}>
         <button onclick={() => this._toggleOpen(!this.open)} className="StatsSelector-item StatsSelector-toggle">
-          {numeral(total).format(total < 1000 ? '0' : '0.0a').toUpperCase()} {selectedMetric.name} <span className="StatsSelector-item-detail">({selectedMetric.systemName})</span>
+          {numeral(total).format('0[.]0a').toUpperCase()} {selectedMetric.name} <span className="StatsSelector-item-detail">({selectedMetric.systemName})</span>
         </button>
         <ul className="StatsSelector-menu">
           {
--- a/app/javascript/src/Stats/lib/metrics_selector.js
+++ b/app/javascript/src/Stats/lib/metrics_selector.js
@@ -29,7 +29,7 @@ export class StatsMetricsSelector extends StatsUI {
     return (
       <div className={`StatsSelector ${this.open ? 'is-open' : ''}`}>
         <button onclick={() => this._toggleOpen(!this.open)} className="StatsSelector-item StatsSelector-toggle">
-          {numeral(total).format('0.0a').toUpperCase()} {selectedMetric.name} <span className="StatsSelector-item-detail">({selectedMetric.systemName})</span>
+          {numeral(total).format(total < 1000 ? '0' : '0.0a').toUpperCase()} {selectedMetric.name} <span className="StatsSelector-item-detail">({selectedMetric.systemName})</span>
         </button>
         <ul className="StatsSelector-menu">
           {
--- a/app/javascript/src/Stats/provider/stats_days_of_week.js
+++ b/app/javascript/src/Stats/provider/stats_days_of_week.js
@@ -23,7 +23,7 @@ class StatsDaysOfWeekChart extends StatsChart {
         y: {
           default: [0, 10],
           tick: {
-            format: d => numeral(d).format(d < 1000 ? '0' : '0.0a').toUpperCase()
+            format: d => numeral(d).format('0[.]0a').toUpperCase()
           }
         }
       },
@@ -64,4 +64,4 @@ let statsDaysOfWeek = (serviceId, options = {}) => {
   })
 }
 
-export { statsDaysOfWeek }
+export { statsDaysOfWeek, StatsDaysOfWeekChart }
--- a/app/javascript/src/Stats/provider/stats_days_of_week.js
+++ b/app/javascript/src/Stats/provider/stats_days_of_week.js
@@ -21,8 +21,9 @@ class StatsDaysOfWeekChart extends StatsChart {
           categories: ['Sunday', 'Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday']
         },
         y: {
+          default: [0, 10],
           tick: {
-            format: d => numeral(d).format('0.0a').toUpperCase()
+            format: d => numeral(d).format(d < 1000 ? '0' : '0.0a').toUpperCase()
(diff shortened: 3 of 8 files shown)


mayorova:
@nidhi-soni1104 please check http://numeraljs.com/#format

The format currently used transforms big (and not very human-readable) values such as `4589850` to something like `4.6M`. We don't want to remove that.

lvillen (commented):
From what I can see here http://numeraljs.com/#format, I guess we were using this format for high number of hits, so I'm not sure about removing the library. Do we know what that 1.0 references?

app/javascript/src/Stats/lib/metrics_selector.jsx
```diff
@@ -29,7 +28,7 @@ export class StatsMetricsSelector extends StatsUI {
     return (
       <div className={`StatsSelector ${this.open ? 'is-open' : ''}`}>
         <button onclick={ev => this._toggleOpen(!this.open)} className='StatsSelector-item StatsSelector-toggle'>
-          {numeral(total).format('0.0a').toUpperCase()} {selectedMetric.name} <span className='StatsSelector-item-detail'>({selectedMetric.systemName})</span>
+          { Math.round(total * 10).toString()} {selectedMetric.name} <span className='StatsSelector-item-detail'>({selectedMetric.systemName})</span>
```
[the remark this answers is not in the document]
nidhi-soni1104: yes I did this, because I saw in every chart its 0.1 0.2 like this

nidhi-soni1104:
> @nidhi-soni1104 please check http://numeraljs.com/#format
> 
> The format currently used transforms big (and not very human-readable) values such as `4589850` to something like `4.6M`. We don't want to remove that.


@mayorova I checked the link you shared and tried multiple fromats, these formats not converting 0.2 to 2, either do convert them into 1 or 0

nidhi-soni1104:
> From what I can see here http://numeraljs.com/#format, I guess we were using this format for high number of hits, so I'm not sure about removing the library. Do we know what that 1.0 references?

@lvillen 1.0 mean in formats

nidhi-soni1104:
> > @nidhi-soni1104 please check http://numeraljs.com/#format
> > The format currently used transforms big (and not very human-readable) values such as `4589850` to something like `4.6M`. We don't want to remove that.
> 
> @mayorova I checked the link you shared and tried multiple fromats, these formats not converting 0.2 to 2, either do convert them into 1 or 0

https://github.com/chartjs/Chart.js/issues/58 , I checked this for reference and then applied this calculation

nidhi-soni1104:
@mayorova 

here is one solution

```
tick: {
  format: d => {
    if (d < 1) {
      return Math.round(d * 10).toString();
    } else {
      return numeral(d).format('0.0a').toUpperCase();
    }
  }
}
```

lvillen:
> @lvillen 1.0 mean in formats
> 
But going back to @mayorova's comment and to http://numeraljs.com/#format, it seems that `1.0` represents a million hits, not a decimal number. Is it possible?

mayorova:
@nidhi-soni1104 No, `return Math.round(d * 10).toString();` is a **NO**, it modifies the actual value. The issue is not that the value is wrong, but that it's formatted in a way that people don't like.

mayorova:
Here is what will do the trick, I guess https://github.com/3scale/porta/pull/3400/files

But of course it would be preferrable to extract this into a utility function rather than duplicate 4 times. I am not sure though where to put it - I don't quite understand how the Stats code is organized.

mayorova:
> But going back to @mayorova's comment and to http://numeraljs.com/#format, it seems that 1.0 represents a million hits, not a decimal number. Is it possible?

`1.0 M`, `1.0 K` represent a million or a thousand hits respectively.
`1.0` with no letter represents 1 hit

lvillen:
> > But going back to @mayorova's comment and to http://numeraljs.com/#format, it seems that 1.0 represents a million hits, not a decimal number. Is it possible?
> 
> `1.0 M`, `1.0 K` represent a million or a thousand hits respectively. `1.0` with no letter represents 1 hit

Oh, right, got confused, sorry 😓

nidhi-soni1104:
> format

@mayorova As I mentioned here, I tied this but this doesn't sound good to  me , because this showing either in 1 or 0 format.
i think it should make 0.2 to 2 and not 0, still not sure so I added this way because I was also confused about this.
https://github.com/3scale/porta/pull/3396#issuecomment-1570272254

mayorova:
> we are showing them as 0 and 1. so your fix is not what we expect if I am right

Well, it is quite strange indeed that the Y axis shows several 0's and several 1's
Screenshot from 2023-05-31 17-49-18

But does it really matter if the values are 0? (I mean, there are no values on the graph at all)

The Y axis tick values are calculated based on the actual values on the graph. So, when there are some real data, the Y axis looks fine:
Screenshot from 2023-05-31 17-49-01

Or for higher values:
Screenshot from 2023-05-31 17-50-42

nidhi-soni1104:
> > we are showing them as 0 and 1. so your fix is not what we expect if I am right
> 
> Well, it is quite strange indeed that the Y axis shows several 0's and several 1's Screenshot from 2023-05-31 17-49-18
> 
> But does it really matter if the values are 0? (I mean, there are no values on the graph at all)
> 
> The Y axis tick values are calculated based on the actual values on the graph. So, when there are some real data, the Y axis looks fine: Screenshot from 2023-05-31 17-49-01
> 
> Or for higher values: Screenshot from 2023-05-31 17-50-42

Oh yes looks fine with real data 

but what about this 0 and 1, I tired this but this made  me confused so multiplied it with 10 and made it integer.

so this trick can be good because this will do only if its less than 1 and not for above 1 and rest thing will works fine as expected and already working

tick: {
  format: d => {
    if (d < 1) {
      return Math.round(d * 10).toString();
    } else {
      return numeral(d).format('0.0a').toUpperCase();
    }
  }
}

app/javascript/src/Stats/provider/stats_days_of_week.jsx
```diff
@@ -22,7 +22,7 @@ class StatsDaysOfWeekChart extends StatsChart {
         },
         y: {
           tick: {
-            format: d => numeral(d).format('0.0a').toUpperCase()
+            format: d => { numeral(d).format(d < 1000 ? '0' : '0.0a').toUpperCase() }
```
mayorova, line 25:
```suggestion
            format: d => numeral(d).format(d < 1000 ? '0' : '0.0a').toUpperCase()
```

mayorova:
Setting `axis.y.default: [0, 10] sets the ticks from 0 to 10 by default, if there is no data:
https://c3js.org/reference.html#axis-y-default

 
Screenshot from 2023-06-01 11-28-23

If the data is low, though, the Y axis will still look a bit weird, because it is calculated automatically based on the data:

Screenshot from 2023-06-01 11-28-11


I think it's fine and we don't need to fix this. If somebody has this kind of usage - I bet they are not real customers :D

app/javascript/src/Stats/lib/chart.jsx
```diff
@@ -29,8 +29,9 @@ export class StatsChart {
             }
           },
           y: {
+            default: [0, 10],
             tick: {
-              format: d => numeral(d).format('0.0a').toUpperCase()
+              format: d => numeral(d).format(d < 1000 ? '0' : '0.0a').toUpperCase()
```
nidhi-soni1104, line 34:
```suggestion
              format: d => numeral(d).format(d <= 10 ? '0' : '0.0a').toUpperCase()
```

nidhi-soni1104: @mayorova what you think of this ?

mayorova:
@nidhi-soni1104 I think the test you've added is not enough. I'd say we need to check the values in three places:
- Y axis ticks
- Number of hits in the metric selector
- On the chart itself (when hovering) - although I think this will use the same format as Y axis ticks

Also, we need to check several scenarios: 
- when the data is empty (usage is 0)
- when the values are < 1000
- when the values are > 1000

You'd need to mock what the backend returns for this.

nidhi-soni1104:
@mayorova Okay I will check above scenarios

Y axis ticks check test is added, I will try to add others.

I dont have real time data with charts, can you let  me know how can I have that on my local?

nidhi-soni1104:
@mayorova what do you think about this, we can merge this now. because till now no graph test are written yet and if  need to write in js then we can create another jira for its test?

mayorova (approved):

PR #3399, opened 2023-05-31
🦋 Secondary nav (formerly known as menu)

josemigallas:
[THREESCALE-9660: Change page header menus with Breadcrumb or secondary nav](https://issues.redhat.com/browse/THREESCALE-9660)

### Before
Mix of breadcrumb and tertiary nav apparently but it's really just a tertiary nav.
Screenshot 2023-06-06 at 14 00 10
Screenshot 2023-06-06 at 14 00 15

### After
The most similar component is the legacy tertiary navigation. An up to date version should maybe use a secondary navigation. But we want it to be as close as it can be to master, so here's the legacy tertiary nav:
Screenshot 2023-06-06 at 13 59 46
Screenshot 2023-06-06 at 13 59 51


lvillen (approved):

app/views/buyers/applications/_menu.html.slim
```diff
@@ -1,13 +0,0 @@
-ul#subsubmenu
```
josemigallas, line 1: This menu isn't rendered anywhere.

PR #3405, opened 2023-06-01
🦋 Standardize page headers using PF4

josemigallas:
[THREESCALE-9709: Implement proper PF4 page headers](https://issues.redhat.com/browse/THREESCALE-9709)

[THREESCALE-9703: Remove force padding by class Main](https://issues.redhat.com/browse/THREESCALE-9703)

Use PF 4 in all page headers. Example:

Screenshot 2023-06-05 at 15 40 08

It needed some adjustments to existing CSS (rules that targetted all h1 and p).

To ensure all pages have the same title, it uses the existing page_header template and the `content_for` method.

However there some special scenarios where the page_header cannot be used as is:
* Some pages are fully implemented with React, that is, they include a header section. These and other pages not using the page_header should use `content_for :main_section_with_react`.
* Some templates are rendered exclusively inside a colorbox modal, therefore the page header is already defined by the "host" template. For this templates, the layout has been disabled from the controller (the layout is already loaded by the host view so it's redundant anyway).

Up until now, `content_for :title` was being use for both the page title (browser window) and the page _header_ title (the one showing inside the page). Now :title is only for the title, and :page_header_title is for the content.

Sublayouts have been removed (most didn't do anything different, with the exception of cms) in favor of using provider layout everywhere.


app/controllers/admin/upgrade_notices_controller.rb
```diff
@@ -5,6 +5,8 @@ class Admin::UpgradeNoticesController < FrontendController
 
   activate_menu :upgrade_notices
 
+  layout false # ???? Where and when is this rendered
```
mayorova, line 8: I figured out that this is used in SaaS, when a tenant has some limitations (some disabled so-called "feature switches").
For example, if the current plan allows only one application per account, clicking on the Create Application button redirects to: `/admin/upgrade_notices/multiple_applications`.

This is how it looks like in production:
Screenshot from 2023-06-06 00-20-31

And this is this branch:
Screenshot from 2023-06-06 00-21-00

(except that in production there is no "upgrade to plan `X" button, and in dev there is one.

mayorova: Just removing this line is good enough IMO:

Screenshot from 2023-06-06 00-27-22

app/lib/simple_layout.rb
```diff
@@ -146,7 +146,6 @@ def create_builtin_pages_and_partials!
     create_builtin_partial!('signup/cas')
 
     # Messages
-    create_builtin_partial!('messages/menu')
```
mayorova, line 149: From what I can tell, this is not related to the deleted `app/views/provider/admin/messages/_menu.html.erb`, this is a template (partial "messages/menu") that is created in the developer portal CMS, so that the user can update it.

So, here and in other places I'll mark it as **revert**

app/models/cms/builtin.rb
```diff
@@ -170,7 +170,6 @@ def self.filesystem_templates
     ORIGINAL_PATHS = {
       'stats/chart' => full_path('stats/_chart.html.liquid'),
       'applications/form' => full_path('applications/_form.html.liquid'),
-      'messages/menu' => full_path('messages/_menu.html.liquid'),
```
mayorova, line 173: revert

test/unit/simple_layout_test.rb
```diff
@@ -50,8 +50,7 @@ def setup
   test '#import - partials' do
     assert SimpleLayout.new(@provider).import!
 
-    %w( messages/menu
```
mayorova, line 53: revert

test/integration/sites/forums_controller_test.rb
```diff
@@ -11,10 +11,6 @@ def setup
   end
 
   def test_edit
-    Account.any_instance.expects(:provider_can_use?).returns(true).at_least_once
```
mayorova, line 14: Why is this removed here?
I mean, I know forums are deprecated and are going to be deleted soon, but why this specific test?... :thinking:

app/controllers/admin/upgrade_notices_controller.rb
josemigallas: 👍🏻

test/integration/sites/forums_controller_test.rb
josemigallas: Because it's the only one breaking CI 🙂.

lvillen (dismissed):

app/views/api/services/edit.html.slim
```diff
@@ -1,4 +1,4 @@
-- content_for :sublayout_title, 'Naming'
+- content_for :page_header_title, 'Naming'
```
mayorova, line 1: Wow, WTF, this is an **extremely** confusing title :flushed: 

It seems that it has always been like that, but I'm kinda shocked :sweat_smile: 

Should be "Edit Product" to me...

josemigallas: I stopped getting porta-shocked a long time ago 🙂.

app/views/api/proxy_rules/edit.html.slim
```diff
@@ -1,4 +1,4 @@
-= content_for :sublayout_title, 'Edit Mapping Rule'
+= content_for :page_header_title, 'Edit Mapping Rule'
```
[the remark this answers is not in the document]
josemigallas: Ah, yes, I've missed this everywhere. Use `=` when rendering (returning) stuff like `= link_to 'habba babba'` and use `-` to execute some login such as defining a variable or calling `content_for`. So you're absolutely right it should be the same everywhere.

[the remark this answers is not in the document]
mayorova: Yeah, that's what I though, thanks for clarification!

PR #3415, opened 2023-06-08
🦋 makes sidebar collapsible like it should be

josemigallas:
[THREESCALE-9697: Make navigation collapsable](https://issues.redhat.com/browse/THREESCALE-9697)

It still does not make it responsive for smaller screen sized, but at least it gives the user more room. Plus it's the default behavior.

NOTE: in Patternfly React the sidebar collapses automagically when the screen shrinks but we don't support that yet.
Screenshot 2023-06-08 at 15 46 20
Screenshot 2023-06-08 at 15 46 12
Screenshot 2023-06-08 at 15 46 16


josemigallas:
> How can one open the sidebar on mobile?

We don't really support mobile browsers but there shouldn't be any difference:
sidebar

Maybe you need to request the desktop site from the mobile.

PR #3419, opened 2023-06-12
⚛️ Turn InlineChart into functional component

josemigallas:
And clean css up.


app/javascript/src/Stats/inlinechart/InlineChart.scss
```diff
@@ -0,0 +1,67 @@
+#mini-charts {
```
josemigallas, line 1: These styles the ones from removed file `app/assets/stylesheets/provider/_charts.scss`.

PR #3423, opened 2023-06-14
Fix missing Patternfly styles

josemigallas:


mayorova (approved):
:+1: 

I don't really understand the implementation - all the hacks and overrides are confusing, as I am not familiar with PF4 classes names, and I don't know what overrides apply to what pages.

But I clicked through the pages, and the ones I checked look good to me - slightly different than current production version, but that's expected.

PR #3425, opened 2023-06-14
remove now redundant billing_locks table

akostadinov:
merge this only once we are confident the new locking strategy (#3403) is superior and actually prevents double execution

### DIFF
--- a/app/models/billing_lock.rb
+++ b/app/models/billing_lock.rb
@@ -1,9 +1,0 @@
-# frozen_string_literal: true
-
-class BillingLock < ApplicationRecord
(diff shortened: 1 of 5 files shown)


jlledom (approved):

akostadinov:
I don't see any references to this model. No need to do in steps in my opinion. But you are right that the model needs to be removed , my bad for not doing.

But lets see if we keep the new locking in the long term first.

jlledom:
Are we giving up on this one?

akostadinov:
No, need to add some stuff to it. Thanks for reminder.

PR #3430, opened 2023-06-16
[WIP]Changes for cachito

jjkiely:
In order to get Porta to build using sources from Cachito, the following changes need to be merged. 
The `name` and `version` fields in the package.json are necessary to prevent a recurring error we had which asked us to add these fields.

The package.json for yarn is necessary in order to install yarn using cachito during the build process

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,4 +1,6 @@
 {
+  "name": "porta",
+  "version": "0.0.1",
   "private": true,
   "repository": {
     "type": "git",
--- a/script/yarn/package.json
+++ b/script/yarn/package.json
@@ -1,0 +1,16 @@
+{
+    "name": "yarn-dependency",
+    "version": "1.0.0",
+    "description": "",
+    "main": "index.js",
+    "scripts": {
+      "test": "echo \"Error: no test specified\" && exit 1"
+    },
+    "keywords": [],
+    "author": "",
+    "license": "ISC",
+    "devDependencies": {
+      "yarn": "^1.22.19"
+    }
+  }
+  


package.json
```diff
@@ -1,4 +1,6 @@
 {
+  "name": "porta",
+  "version": "0.0.1",
```
jlledom, line 3: Why no setting the actual porta version `2.13.2`?

script/yarn/package.json
```diff
@@ -0,0 +1,16 @@
...
+      "test": "echo \"Error: no test specified\" && exit 1"
+    },
+    "keywords": [],
+    "author": "",
+    "license": "ISC",
+    "devDependencies": {
+      "yarn": "^1.22.19"
+    }
+  }
+
```
jlledom, line 16: Better add a new line at the end of file.

jlledom (changes requested):

package.json
gsaslis: @jlledom is that the porta version, or the 3scale version ?

gsaslis: (but, yes, we absolutely need to set the correct version here)

jlledom: @gsaslis is there a porta version different from 3scale version?

gsaslis: well: 

- latest 3scale product build is 2.13.4 (not 2.13.2) 
- the porta component can have a different version number than the whole product does
- we **do** use a different version for the porta (system) build - (see tags in https://catalog.redhat.com/software/containers/3scale-amp2/system-rhel7/5d80b9d7d70cc5521d516a49 )

Having said that, this here is the community porta (so a different versioning policy is 100% fine) **and** (and this is important) I don't think this `name` + `version` actually makes much difference to porta itself _anyway_ (otherwise I'd have expected to find it here already)...

tl;dr 

`42` would also work. 🙃

script/yarn/package.json
akostadinov: or rather remove trainling white spaces at last line

package.json
akostadinov: maybe instead of `0.0.1` use `placeholder` or some other value indicating that it is not supposed to correspond to actual version

jjkiely: The version must be in semantic versioning format according to nodejs docs, so what do you guys think would be a suitable "not a version" version? Personally, I don't see how any other version will be any better than 0.0.1, although none of them are ideal...

akostadinov: idk, maybe `0.0.0`, otherwise `0.0.1` looks second bet option in this case :)

jjkiely: I've set it to 0.0.0 :)

jlledom (approved):

akostadinov (approved):

PR #3432, opened 2023-06-19
💄 Fix [yet even more] css, this time missing legacy styles

josemigallas:
### What
In the past, we tried to apply Patternfly by adding classes like `pf-c-button` and `pf-m-link` to all buttons and links. This only helper because Patternfly stylesheet was being loaded before our legacy stylesheet, therefore the latter had priority.

Now that Patternfly is loaded last (so that it overrides legacy styles) some of these components are broken.

[THREESCALE-9734: Broken UI/layout on multiple pages - missing CSS](https://issues.redhat.com/browse/THREESCALE-9734)

### How
This PR removes Patternfly classes from legacy components. It doesn't make sense to add them to components that should not look different, only to those that are updated.

Example Application Overview:
* Buttons and links **with** Patternfly classes:
Screenshot 2023-06-19 at 15 52 00

* Buttons and links **without** them:
Screenshot 2023-06-19 at 18 23 30

### DIFF
--- a/app/helpers/account_helper.rb
+++ b/app/helpers/account_helper.rb
@@ -79,7 +79,7 @@ module AccountHelper
     alert = t('buyers.accounts.edit.delete.admin_restricted', admin: current_account.first_admin.try(:email))
 
     url = can?(:destroy, account) ? admin_buyers_account_path(account) : javascript_alert_url(alert)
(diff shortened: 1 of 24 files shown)


mayorova (approved):
Nice one! :+1:

PR #3441, opened 2023-06-26
🦋 Update Buyer accounts' table header

josemigallas:
[THREESCALE-9764: PF4 to buyer accounts table header](https://issues.redhat.com/browse/THREESCALE-9764)

Before:
Screenshot 2023-06-27 at 12 22 14

After:
Screenshot 2023-06-26 at 17 47 22
Screenshot 2023-06-26 at 17 47 41
Screenshot 2023-06-26 at 17 47 50

### DIFF
--- a/app/helpers/column_sorting_helper.rb
+++ b/app/helpers/column_sorting_helper.rb
@@ -1,3 +1,4 @@
+# rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/MethodLength
 module ColumnSortingHelper
 
   # table sorting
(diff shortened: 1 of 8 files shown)


app/javascript/packs/provider.scss
```diff
@@ -2,7 +2,7 @@
 // consistent styles throughout the application, PatternFly Base CSS should be used in the project,
 // or some components may diverge in appearance. This import statement should be placed before your
 // main application component is imported.
-@import '~@patternfly/react-core/dist/styles/base.css';
+@import '~@patternfly/patternfly/patternfly-base.css';
```
josemigallas, line 5: This base package also includes the icons, which `patternfly/react-core` gets from `patternfly/react-icons`.

app/helpers/column_sorting_helper.rb
```diff
@@ -30,6 +31,43 @@ def sortable(column, title = nil, path = :url_for)
             { :class => css_class }
   end
 
+  def th_sortable(column, title = nil, path = :url_for)
```
josemigallas, line 34: This will eventually replace the old method `sortable` once all tables have been implemented using PF4.

mayorova: I think this might need a unit test?

mayorova:
The screenshots don't show that, but actually the filters are misplaced:

Screenshot from 2023-06-27 13-20-38

Seems like it's missing an empty `<td>` in the Signup Date column.

Also, I think it would be nice if Create and Activate buttons were aligned.

mayorova (changes requested):
My change requests here: https://github.com/3scale/porta/pull/3441#issuecomment-1609307883

app/helpers/column_sorting_helper.rb
josemigallas: In all seriousness, I don't think it's worth it. The method (and the original) are to fiddly and IMHO it should be enough to test it indirectly (via 🥒)

app/views/buyers/accounts/_search_results.html.slim
```diff
@@ -71,7 +68,7 @@ div class="pf-c-card"
             = account_states_info(account)
           td role="cell" class="pf-c-table__action actions operations"
             ul.Actions
-              - if account.admin_user.try!(:pending?)
+              - if account.admin_user.try!(:pending?) || true
```
mayorova, line 71: :eyes:

josemigallas: 😭

mayorova (approved):

PR #3447, opened 2023-06-29
Prepare constants for Zeitwerk autoloader

jlledom:

In rails 6 we'll use the Zeitwerk autoloader, which founds the constants in a slightly different way:

- The old `:classic` autoloader starts from a missing constant: e.g. `API`.
  1. It runs `"API".underscore` and the result is `"api"`.
  2. It looks for a file or folder called `api` and expects the constant to be defined there.
  
- The new autoloader `:zeitwerk` starts from a file name, e.g. `api.rb`.
  1. It runs `"api".camelize` and the result is `"Api"`.
  2. It expects the constant `Api` to be defined there.

Most of times, `"ConstantName".underscore.camelize == "ConstantName"`, but that's not always the case, like in the example above.

In this PR, some constants are renamed to names which are good for both `:classic` and `:zeitwerk`, so they will work in rails 6 as well.

More info: https://guides.rubyonrails.org/classic_to_zeitwerk_howto.html

**Verification steps** 

Everything should work. The tests should pass.

**How to review**

There are many changes, but all of them belong to one of these categories:

1. **Constants renaming**: I checked all constants one by one but I could have made a mistake. To verify the constants names are right, you should open a rails console and run `"ConstantName".underscore`. If the resulting string matches the name of the file where it's defined, it's correct. Also, writing the constant name on the rails console should return the constant. Most of changes in this PR belong to this category.
2. **Files renamed or moved to other folders**: Some files have been renamed/moved to match the constant name defined inside (e938f6fa or db29f43c). Others have been moved to their own file, as Zeitwerk doesn't like two classes defined in the same file (e79e0ee3).
3. **Avoid sharing namespaces between gems and autoloadable constants**: Everything in the autoload path is unloaded and reloaded again when we make any change locally. When a constant in the autoload path shares namespace with a gem, Zeitwerk will unload the namespace also for the gem, but the gem is not reloaded again, which usually leads to server crashes. A couple of commits fix this problem (ac0e9cf8 and 9a3242be).
4. **Avoid loading autoloaded constants from the initializers**: Like in (11f3527a). Check https://edgeguides.rubyonrails.org/autoloading_and_reloading_constants.html#autoloading-when-the-application-boots


app/presenters/provider_oauth_flow_presenter.rb
```diff
@@ -1,4 +1,4 @@
-class ProviderOauthFlowPresenter < OauthFlowPresenter
+class ProviderOAuthFlowPresenter < OAuthFlowPresenter
```
akostadinov, line 1: Is there a special rule about `OAuth`? Because usually only first letter would be capitalized.

app/indexes/account_index.rb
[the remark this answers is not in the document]
akostadinov: in 4.4.1 [1], there is 
> Automatically remove app/indices from Zeitwerk's autoload paths in Rails 6.0 onwards (if using Zeitwerk as the autoloader).

I think we should do the same. Instead of current workarounds, to exclude the path. After Rails 6 upgrade, we can remove the manual removal.

[1] https://github.com/pat/thinking-sphinx/releases

app/lib/finance/api/response.rb
```diff
@@ -1,6 +0,0 @@
-module Finance
-  module API
-    module Collection
-    end
-  end
-end
```
akostadinov, line 6: what was this?

app/mailers/mail_preview.rb
```diff
@@ -8,7 +8,7 @@ def application_created
   end
 
   def application_plan_change_requested
-    event = Applications::ApplicationPlanChangeRequestedEvent.create(ApplicationContract.last, User.last, Plan.last)
+    event = Applications::ApplicationPlanChangeRequestedEvent.create(Cinstance.last, User.last, Plan.last)
```
akostadinov, line 11: why this change?

app/models/application_contract.rb
```diff
@@ -0,0 +1,3 @@
+# frozen_string_literal: true
+
+ApplicationContract = Cinstance
```
akostadinov, line 3: do you think this will always run after `cinstance`?

app/presenters/provider_oauth_flow_presenter.rb
jlledom: We don't need it.

For Rails 5.2, the autoloader will find `ProviderOAuthFlowPresenter` and will look for a file called `"ProviderOAuthFlowPresenter".underscore`, that is, `provider_oauth_flow_presenter`. Which matches the name of the file.

For Rails 6, it will find a file called `provider_oauth_flow_presenter` and will expect `"provider_oauth_flow_presenter".camelize` => `ProviderOAuthFlowPresenter` to be defined inside. I know it's strange but check this in the rails console:

```
[4] pry(main)> "oauth".camelize
=> "OAuth"
[5] pry(main)> "oauth2".camelize
=> "Oauth2"
```

The problem only exists for "oauth2", for which we have defined some [inflectors](https://github.com/3scale/porta/blob/THREESCALE-7405-update-rails-to-6-0/config/initializers/zeitwerk.rb) in rails 6.

`camelize` makes other exceptions, like:

```
[16] pry(main)> "github".camelize
=> "GitHub"
[17] pry(main)> "sso".camelize
=> "SSO"
```

app/indexes/account_index.rb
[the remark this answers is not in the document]
jlledom: So, if I understood correctly, Sphinx doesn't want the indices in the autoload path? In that case I think it's better to revert db29f43c from this PR and from #3339. For Rails 5 it won't make a difference, for rails 6, Sphinx is already doing it. Am I right?

app/lib/finance/api/response.rb
jlledom: I don't know, check this comment: https://github.com/3scale/porta/pull/3339#discussion_r1242428237

I think it was a mistake of leftover.

app/mailers/mail_preview.rb
jlledom: That because I initially removed `ApplicationContract` completely, so I removed the only reference to it, here. Then I restored it in a way it's actually loaded from both autoloaders, and I never reverted the change in this line. See the discussion here: https://github.com/3scale/porta/pull/3339#discussion_r1238512693

app/models/application_contract.rb
jlledom: That's not a problem in Rails 5, if `Cinstance` is a missing constant, the autoloader will find the file and load it. In Rails 6, I didn't find any mention to load order being an issue in the Zeitwerk docs, I assume that kind of problems are solved.

app/indexes/account_index.rb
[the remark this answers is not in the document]
akostadinov: Correct.

app/presenters/provider_oauth_flow_presenter.rb
akostadinov: Yeah, this is what I meant. I didn't expect `"oauth".camelize` -> `"OAuth"`

It's nice.

app/services/segment_integration/client_error.rb
```diff
@@ -0,0 +1,3 @@
+# frozen_string_literal: true
+
+class SegmentIntegration::ClientError < SegmentIntegration::ResponseError; end
```
akostadinov, line 3: I'm not sure this path is ideal for these files but it's basically making the old approach work so it's fine.

jlledom: Do you suggest another path? as long as it's not `segment/` it'll be valid.

app/indexes/account_index.rb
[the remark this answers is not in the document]
jlledom: Any idea on why they want to remove the `indices` folder from the autoload path? If I revert the commit, the indices aren't found, precisely because the gem removes it from the autoload path. The server doesn't start.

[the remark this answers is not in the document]
akostadinov: You mean on on 6.0?

app/services/segment_integration/client_error.rb
akostadinov: what you chose looks good to me

app/indexes/account_index.rb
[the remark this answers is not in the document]
jlledom: yes

config/initializers/money.rb
```diff
@@ -1,4 +1,8 @@
 ActiveSupport.on_load(:active_record) do
   ActiveRecord::Base.send(:include, ThreeScale::HasMoney)
 end
-ActionController::Base.helper(ThreeScale::MoneyHelper) if defined?(ActionController)
+
+Rails.application.config.to_prepare do
+  require 'three_scale/money'
```
akostadinov, line 6: why is require needed here? Shouldn't it be auto-loaded?

app/lib/active_merchant_hacks.rb
```diff
@@ -2,6 +2,9 @@
 
 # ActiveMerchant monkey patches
 
+module ActiveMerchantHacks
+end
```
akostadinov, line 6: I would rather move this file out of autoload path. Or better do same as with `HashHacks`.

config/initializers/money.rb
jlledom: You're right. I removed this.

config/initializers/money.rb
```diff
@@ -3,6 +3,5 @@
 end
 
 Rails.application.config.to_prepare do
-  require 'three_scale/money'
   ActionController::Base.helper(ThreeScale::MoneyHelper) if defined?(ActionController)
```
akostadinov, line 6: One more remark. Perhaps inside `to_prepare`, `ActionController` is already available so this condition could be removed?

config/initializers/money.rb
jlledom: After removing it, I found a problem: the required class included a required monkey patch. So I moved the monkey patch to it's own class out of the autoload path, and left the `ThreeScale::Money` in the autoload path. See 5c3101a838d9d7d4333b970eede969df74802fc0

config/initializers/money.rb
jlledom: I'll check this in the rails 6 branch, but I prefer to ignore it here since it's not directly related to the PR

app/presenters/provider_oauth_flow_presenter.rb
mayorova: This is because of https://github.com/3scale/porta/blob/d11cf2129b7fe952b5073dd364e3ea3f187543ac/config/initializers/inflections.rb#L22

By default, Rails would camelize "oauth" as "Oauth", but porta has a special rule for OAuth (as well as some other acronyms)

jlledom: Thanks for the info!

app/mailers/mail_preview.rb
mayorova: I think we can keep it like this. Maybe at some point we'll get rid of ApplicationContract alias...

config/application.rb
```diff
@@ -281,11 +276,12 @@ def try_config_for(*args)
...
     }.merge(try_config_for(:paperclip) || {})
 
-    initializer :paperclip_defaults, after: :load_config_initializers do
-      Paperclip::Attachment.default_options.merge!(s3_options: CMS::S3.options) # Paperclip does not accept s3_options set as a Proc
-    end
+    config.to_prepare do
+      Rails.application.config.log_formatter = System::ErrorReporting::LogFormatter.new
+      (Rails.application.config.logger || Rails.logger).formatter = Rails.application.config.log_formatter
+
+      Paperclip::Attachment.default_options[:s3_options] = CMS::S3.options # Paperclip does not accept s3_options set as a Proc
```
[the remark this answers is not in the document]
jlledom: Rubocop warning:

image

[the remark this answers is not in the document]
mayorova: Great, thanks

lib/thinking_sphinx/active_record/database_adapters/oracle_adapter.rb
mayorova, line 1: Why this change?

lib/three_scale/conversions.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module ThreeScale
+  module Conversions
```
mayorova, line 4: I would prefer to namespace it within Money, as it's about money, and not just any conversion.

But I guess that would clash with the class ThreeScale::Money ... so, well, not strongly objecting this naming.

lib/thinking_sphinx/active_record/database_adapters/oracle_adapter.rb
jlledom: It was causing a crash server on Zeitwerk, if I recall correctly, because it shares namespace with the sphinx gem. It's explained in the `How to review section` of the description, point 3.

lib/three_scale/conversions.rb
mayorova: Yeah, that's an option too (and should be quite easy).
Up to you, I don't mind either way (keeping it like this, or renaming)

mayorova (approved):
Nice

lib/three_scale/conversions.rb
jlledom: I'll modify this in https://github.com/3scale/porta/pull/3339, since this is a minimal change, so I don't dismiss this approval.

akostadinov (approved):
Very cool, thank you!

The only remaining thing is whether we really need:
```
ActionController::Base.helper(ThreeScale::MoneyHelper) if defined?(ActionController)
```

or can go without condition or using `on_load` but since it is no change from before, I think we are fine to close this PR.

PR #3449, opened 2023-07-05
Fix some missing icons

josemigallas:
These and other icons have disappeared... who knows when! This PR replace such icons with "updated" ones, according to cms-toolbar-menu-right. Note that they now look flat whereas before they were outlined.

Before:
Screenshot 2023-07-05 at 14 58 33

After:
Screenshot 2023-07-05 at 14 55 03

### DIFF
--- a/app/assets/javascripts/provider/cms/sidebar.js.coffee
+++ b/app/assets/javascripts/provider/cms/sidebar.js.coffee
@@ -24,7 +24,7 @@ jQuery.expr[':'].like = (el, i, selector) ->
 
 ## Sidebar class
 class Sidebar
-  FOLDER_ICONS = ['fa-folder-o', 'fa-folder-open-o']
+  FOLDER_ICONS = ['fa-folder', 'fa-folder-open']
   ICON_SETS = [FOLDER_ICONS]
   HIGHLIGHT_CLASS = 'current'
   TOGGLE_SELECTOR = '[data-behavior~=toggle]'
@@ -84,7 +84,7 @@ class Sidebar
 
   update_expand_collapse_button: (all_packed = @all_items_packed()) ->
     @expand_collapse_all_button().
-      html(Sidebar.icon(if all_packed then 'plus-square-o' else 'minus-square-o'))
+      html(Sidebar.icon(if all_packed then 'plus-square' else 'minus-square'))
 
   fire_ajax: ->
     @ajax.abort() if @ajax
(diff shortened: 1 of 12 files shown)


mayorova:
> Note that they now look flat whereas before they were outlined.

Why this change?... I personally prefer them the old way

akostadinov:
What's wrong with me. What is flat?

mayorova:
@akostadinov 

Current (prod):
Screenshot from 2023-07-05 15-56-02

This PR:
Screenshot from 2023-07-05 15-55-25

akostadinov:
Ok, agreed. Outlined is more stylish. Flat is so nineties.

lvillen:
The change is related to PF guidelines, you can check them here: https://www.patternfly.org/v4/guidelines/icons

akostadinov:
@lvillen , can you clarify which part you are referring to?

lvillen:
@akostadinov sure! In the previous link you can check https://www.patternfly.org/v4/guidelines/icons/#font-awesome-solid-fas-vs-font-awesome-regular-far, where it says that "Font Awesome icons included with PatternFly core are FAS (Font Awesome solid) webfont icons", which is, as I understand, the guideline to follow. 

In the same link below you can check the icons @josemigallas's been using here, and their style is `fas`.

mayorova:
OK, thanks @lvillen

akostadinov:
>  If a FAR icon is needed, developers must include the FAR icon(s) themselves. There are a few ways of doing so:
> ...

It doesn't sound like a requirement. I think it would be fine to use whatever we like more.

jlledom (approved):
I think flat is OK, if that's what the guidelines say.

josemigallas:
> > If a FAR icon is needed, developers must include the FAR icon(s) themselves. There are a few ways of doing so:
> > ...
> 
> It doesn't sound like a requirement. I think it would be fine to use whatever we like more.

You may be right, but also PF do not include these icons so we require yet another 3rd party library just to use a couple of "illegal" 😆 outlined icons. I think it's better to just stick to Patternfly 🤷🏻.

josemigallas:
Moreover, some icons are outlined but most are flat and I think we should use the same "format" everywhere for consistency.

akostadinov:
Ok, only if you add a link to my favorite 90th band.

jlledom:
> Ok, only if you add a link to my favorite 90th band.

This one, maybe? https://backstreetboys.com/home/

akostadinov:
> This one, maybe? https://backstreetboys.com/home/

:exploding_head:

PR #3450, opened 2023-07-05
Remove deprecation warnings for Rails 6

jlledom:

Move to 5.2 all changes that fix deprecation warnings in 6 but are also valid for 5.2

**Verification steps** 

Everything should work. The tests should pass.

### DIFF
--- a/app/presenters/dashboard/notifications_presenter.rb
+++ b/app/presenters/dashboard/notifications_presenter.rb
@@ -2,7 +2,7 @@ class Dashboard::NotificationsPresenter < Dashboard::MessagesPresenter
   IGNORED_NOTIFICATIONS = %w(csv_data_export daily_report weekly_report).freeze
 
   def initialize(notifications)
-    super notifications.where.not(system_name: IGNORED_NOTIFICATIONS, title: nil).order(created_at: :desc)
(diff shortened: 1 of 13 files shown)


app/presenters/dashboard/notifications_presenter.rb
```diff
@@ -2,7 +2,7 @@ class Dashboard::NotificationsPresenter < Dashboard::MessagesPresenter
   IGNORED_NOTIFICATIONS = %w(csv_data_export daily_report weekly_report).freeze
 
   def initialize(notifications)
-    super notifications.where.not(system_name: IGNORED_NOTIFICATIONS, title: nil).order(created_at: :desc)
+    super notifications.where.not(system_name: IGNORED_NOTIFICATIONS).where.not(title: nil).order(created_at: :desc)
```
akostadinov, line 5: doesn't `not` revert the condition?

jlledom: I think it's right, check https://www.bigbinary.com/blog/rails-6-deprecates-where-not-working-as-nor-and-will-change-to-nand-in-rails-6-1

akostadinov: You are right, I missed the first `not` :(

mayorova (approved):

akostadinov (approved):

PR #3458, opened 2023-07-10
Test capybara timeout

jlledom:

Some random Oracle tests always fail in rails 6, the error is always the same:  Capybara not finding an element in the page. I suspect this is because of the Oracle slowness we discussed a few weeks ago, probably the screen takes too much time to load and Capybara times out.

In this PR I increase the timeout for Capybara on the Rails 6 branch. Let's see if that fixes the tests.

No issue.

**Verification steps** 

All tests should pass for Oracle.


jlledom:
Oracle tests passed after increasing the timeout. I'm closing this.

PR #3467, opened 2023-07-14
🦋 Site Usage Rules with PF4

josemigallas:
[THREESCALE-9949: Site Usare Rules](https://issues.redhat.com/browse/THREESCALE-9949)

Also fixes [THREESCALE-9833: Misaligned text for changing account and service plans](https://issues.redhat.com/browse/THREESCALE-9833)

### Before

Screenshot 2023-07-14 at 12 40 44-
Screenshot 2023-07-14 at 12 40 49

### After
Screenshot 2023-08-01 at 17 13 19
Screenshot 2023-08-01 at 17 11 58

### DIFF
--- a/app/views/shared/_plan_change_settings.html.slim
+++ b/app/views/shared/_plan_change_settings.html.slim
@@ -6,8 +6,7 @@ div class="pf-c-form__group-control pf-m-stack"
                                 checked: current_setting == 'request'
   = render 'plan_change_radio', permission: permission,
                                 value: 'direct',
-                                checked: current_setting == 'direct',
-                                hint: can?(:see, :finance) ? t('formtastic.hints.settings.change_plan_finance') : nil
+                                checked: current_setting == 'direct'
   - if can?(:see, :finance)
     = render 'plan_change_radio', permission: permission,
                                   value: 'credit_card',
@@ -17,5 +16,4 @@ div class="pf-c-form__group-control pf-m-stack"
                                   checked: current_setting == 'request_credit_card'
     = render 'plan_change_radio', permission: permission,
                                   value: 'none',
-                                  checked: current_setting == 'none',
-                                  hint: t("formtastic.hints.settings.#{permission}")
+                                  checked: current_setting == 'none'
--- a/app/views/shared/_plan_change_settings.html.slim
+++ b/app/views/shared/_plan_change_settings.html.slim
@@ -1,21 +1,21 @@
-= form.input permission,
-             label: label,
-             hint: '',
-             as: :radio,
-             collection: Logic::PlanChanges::PERMISSIONS[0..1]
+- current_setting = settings.send(permission)
 
-- if can?(:see, :finance)
-  li.inline-hints
-    p.inline-hints Change plan directly if developer has credit card, otherwise:
-
-  = form.input  permission,
-                label: '',
-                hint: '',
-                as: :radio,
-                collection: Logic::PlanChanges::PERMISSIONS[2..3]
-
-  = form.input  permission,
-                label: '',
-                hint: "When 'Request plan change' is selected you will be informed about users' requests by email - you can then execute it in administration interface.",
-                as: :radio,
-                collection: Logic::PlanChanges::PERMISSIONS[-1..-1]
+div class="pf-c-form__group-control pf-m-stack"
+  = render 'plan_change_radio', permission: permission,
+                                value: 'request',
+                                checked: current_setting == 'request'
+  = render 'plan_change_radio', permission: permission,
+                                value: 'direct',
+                                checked: current_setting == 'direct',
+                                hint: can?(:see, :finance) ? t('formtastic.hints.settings.change_plan_finance') : nil
+  - if can?(:see, :finance)
+    = render 'plan_change_radio', permission: permission,
+                                  value: 'credit_card',
+                                  checked: current_setting == 'credit_card'
+    = render 'plan_change_radio', permission: permission,
+                                  value: 'request_credit_card',
+                                  checked: current_setting == 'request_credit_card'
+    = render 'plan_change_radio', permission: permission,
+                                  value: 'none',
+                                  checked: current_setting == 'none',
+                                  hint: t("formtastic.hints.settings.#{permission}")
--- a/app/views/sites/usage_rules/edit.html.slim
+++ b/app/views/sites/usage_rules/edit.html.slim
@@ -48,28 +48,46 @@ div class="pf-c-card"
           div class="pf-c-form__group-control"
             = render 'settings_check', setting: 'public_search', checked: settings.public_search
 
-      - if can?(:see, :account_plans) || can?(:see, :service_plans)
-        section class="pf-c-form__section" role="group" aria-labelledby="settings-section-advanced-plans"
-          h2 class="pf-c-form__section-title" id="settings-section-advanced-plans" aria-hidden="true" Advanced Plans
-          - if can?(:see, :account_plans)
-            div class="pf-c-form__group"
-              div class="pf-c-form__group-control"
-                = render 'settings_check', setting: 'account_plans_ui_visible', checked: settings.account_plans_ui_visible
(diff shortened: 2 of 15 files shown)


mayorova:
I've seen a couple of issues so far:
usage_rules

1. weird hint on "Developers are allowed sign up themselves"
(LOL, also "to" is missing here)
Can be fixed by adding `formtastic.hints.settings.signups_enabled: ""` to i18n

2. link is broken for "Account Approval Required" hint
(you need to have some SSO configured to make this part visible)
Also... this label is not taken from i18n (hence, this camel case, I guess), it shows as 
```<span class="translation_missing" title="translation missing: en.formtastic.labels.settings.account_approval_required">Account Approval Required</span>``` in the HTML

File: app/views/sites/usage_rules/edit.html.slim [hunk 1]
```diff
@@ -1,60 +1,81 @@
...
+            div class="pf-c-form__group-control"
+              = render 'settings_check', setting: 'account_approval_required',
+                                         checked: settings.account_approval_required,
+                                         hint: approval_required_editable_hint(current_account)
+        - elsif @settings.approval_required_disabled?
+          div class="pf-c-form__group"
+            div class="pf-c-form__group-control"
+              = render 'settings_check', setting: 'account_approval_required',
+                                         checked: settings.account_approval_required,
+                                         disabled: true,
```
mayorova, line 40: missing handing of the `disabled` property in the `settings_check` partial?...

app/views/sites/usage_rules/edit.html.slim
```diff
@@ -1,60 +1,81 @@
...
-    = form.inputs 'Users' do
-      = form.input :strong_passwords_enabled, hint: t("formtastic.hints.settings.strong_passwords_enabled", strong_password_definition: strong_password_definition)
-
-    = form.inputs "Search" do
-      = form.input :public_search
+div class="pf-c-card"
+  div class="pf-c-card__body"
+    = semantic_form_for @settings, url: admin_site_usage_rules_path, html: { class: "pf-c-form pf-m-limit-width" } do |form|
+      section class="pf-c-form__section" role="group" aria-labelledby="settings-section-user-account"
+        div class="pf-c-form__section-title" id="settings-section-user-account" aria-hidden="true" User account management zone
```
mayorova, line 8: I think `h2` would look nicer here (well, for all sections), and semantically better?...

mayorova:
I don't really like the **Advanced Plans** section, as I think the hierarchy is really mixed up.

Like "Account plans changing" should be really "under" Account Plans section, and same for "Service Plan changing".

Also, 

"Change plan directly if developer has credit card, otherwise:" title is specifically for the two checkboxes:
- Only request a plan change
- Request Credit Card entry for paid plans

app/views/sites/usage_rules/edit.html.slim
josemigallas: Oops, I forgot

app/views/sites/usage_rules/edit.html.slim
josemigallas: I see what you mean but Patternfly disagrees: https://www.patternfly.org/v4/components/form/html#form-sections 🤷🏻

mayorova: yeah, but https://www.patternfly.org/v4/components/form/html/#usage says that `.pf-c-form__section-title` is applicable to multiple elements, `h2` included :innocent: 

also: https://github.com/patternfly/patternfly/issues/3972

So, I think it's at our discretion...

josemigallas: Fair enough, let's use `h2` then. And also `label` instead of `div`.

josemigallas:
I noticed the radiogroups are supposed to hid when the checkbox is unchecked, so I updated it. I also removed the "Advanced plans" section:

Screenshot 2023-07-18 at 15 08 02

mayorova:
> I noticed the radiogroups are supposed to hid when the checkbox is unchecked, so I updated it. I also removed the "Advanced plans" section:

hmmmm, really? it's not like it works in production :shrug:

app/views/shared/_plan_change_settings.html.slim
```diff
@@ -1,21 +1,21 @@
...
-                as: :radio,
-                collection: Logic::PlanChanges::PERMISSIONS[-1..-1]
+div class="pf-c-form__group-control pf-m-stack"
+  = render 'plan_change_radio', permission: permission,
+                                value: 'request',
+                                checked: current_setting == 'request'
+  = render 'plan_change_radio', permission: permission,
+                                value: 'direct',
+                                checked: current_setting == 'direct',
+                                hint: can?(:see, :finance) ? t('formtastic.hints.settings.change_plan_finance') : nil
```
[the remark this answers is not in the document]
josemigallas: This is the lesser evil. Either we tweak PF4 in our favor or we start a css-wrestling battle agains it that eventually could break... OR we use a different design for this radiogroup.

mayorova:
@josemigallas 

I've tried these two options:

Account Plans are similar to what we have now (using `div class="pf-c-form__group-label-main"` for extra text)

Service Plans are with rewording, that hopefully makes the behavior more clear (and the structure more flat)

Basically, it's translating the logic that is defined [here](https://github.com/3scale/porta/blob/8bc4ef1d5ebf8e0794b2d5c684cccd3abd7b746c/app/lib/logic/plan_changes.rb#L77-L101) into human language.

josemigallas:
> Which one do you like more? (or maybe none? 😅 )

I like the second one a lot. Docs could take a look at the whole page's microcopy and these new radios to ensure the have the right wording.

config/locales/en.yml
```diff
@@ -481,6 +481,12 @@ en:
         captcha_hint_true: "When 'Suspicious only' is selected then reCAPTCHA is shown only if the automated checks detect a possible spammer. For 'Always', reCAPTCHA will always appear when a form is presented to a user who is not logged in."
     usage_rules:
       edit:
+        plans_label: 'Developers are allowed to:'
+        plans_info: |
+          The following options define whether the developers are allowed to change the plan
+          directly, or only request a plan change. When plan change is requested,
+          notifications about the requests will be sent to the admin users by email. Th plan
```
mayorova, line 488: `The plan`

config/locales/en.yml
```diff
@@ -1327,6 +1333,12 @@ en:
         public_search: 'Enable public search on Developer Portal'
         cms_escape_draft_html: 'Use Automatic Escaping for draft content'
         cms_escape_published_html: 'Use Automatic Escaping for published content'
+        change_plan_permission:
+          request: Request a plan change
+          direct: Change the plan directly
+          credit_card: Only request a plan change
+          request_credit_card: Request Credit Card entry for paid plans
+          none: Don't allow plan changes
```
jlledom, line 1341: After removing the hints (`"Change plan directly if developer has credit card, otherwise:"`), these options are not clear for me now. For instance, what's the difference between `"Request a plan change"` and `"Only request a plan change"`?

I think it needs some rewording. Just a suggestion:

```suggestion
          none: Don't allow plan changes
          request: Request a plan change
          direct: Change the plan directly
          credit_card: Change the plan if developer has CC, request a plan change otherwise
          request_credit_card: Change the plan if developer has CC, request CC details otherwise
```

mayorova: Agreed, it's not clear at all :upside_down_face: 

I suggested this wording: https://github.com/3scale/porta/pull/3467#issuecomment-1641979386 (in Service Plans section). It is probably too long... but I think it's the only way to clarify the selected behavior.

mayorova:
Apart from the comments, it looks really nice!

config/locales/en.yml
jlledom: Oh yes, true 👍.

config/locales/en.yml
josemigallas, line 488:
```suggestion
          notifications about the requests will be sent to the admin users by email. The plan
```

config/locales/en.yml
josemigallas: I missed this 😬. Here is the new wording: https://github.com/3scale/porta/pull/3509

PR #3469, opened 2023-07-14
🦋 Use PF4 titles in CMS

josemigallas:
[THREESCALE-9769: Styles in the Dev Portal CMS are a bit broken](https://issues.redhat.com/browse/THREESCALE-9769)

### DIFF
--- a/app/views/provider/admin/cms/builtin_static_pages/edit.html.erb
+++ b/app/views/provider/admin/cms/builtin_static_pages/edit.html.erb
@@ -1,4 +1,6 @@
-<h1>Built-in page '<%= @page.title %>'</h1>
+<div class="pf-c-content">
+  <h1>Built-in page '<%= @page.title %>'</h1>
+</div>
 
 <%= cms_new_button_dropdown(CMS::Layout) %>
 
--- a/app/assets/stylesheets/provider/_commons.scss
+++ b/app/assets/stylesheets/provider/_commons.scss
@@ -10,23 +10,18 @@ body {
 }
 
 main > h1,
-main >  h2,
-#tab-content > h1,
-#tab-content > h2 {
+main > h2 {
   font-weight: $font-weight-light;
 }
 
 main > h1:first-child,
-main >  h2:first-child,
-#tab-content > h1:first-child,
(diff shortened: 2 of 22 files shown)


mayorova:
I think alignment-wise, it might be better to wrap the header's div into `<section class="pf-c-page__main-section pf-m-light">`, WDYT?

Compare without section:
Screenshot from 2023-07-14 16-00-30

with section:
Screenshot from 2023-07-14 16-00-06


Although it feels like the page has too much empty space, but at least the page is "consistent" in this, and the header doesn't stand out.

josemigallas:
> <section class="pf-c-page__main-section pf-m-light">

OK I added the section around these titles, however that means now all pages have to have this extra blank space for consistency.

app/views/provider/admin/cms/_tab_content_title.html.slim
```diff
@@ -0,0 +1,5 @@
+section class="pf-c-page__main-section pf-m-light"
+  div class="pf-c-content"
+    h1 = header
+    - if body = local_assigns[:body]
```
jlledom, line 4: Shouldn't it be like this?

```suggestion
    - if body == local_assigns[:body]
```

josemigallas: No, it's an assignment inside a condition. It basically defines `body` and then checks its value. Alternatively I'd do `if local_assigns.key?(:body)`.

app/views/provider/admin/cms/builtin_static_pages/edit.html.erb
```diff
@@ -1,4 +1,4 @@
-<h1>Built-in page '<%= @page.title %>'</h1>
+<%= render locals: 'provider/admin/cms/tab_content_title', { header: "Built-in page '#{@page.title}'" } %>
```
mayorova, line 1:
```suggestion
<%= render partial: 'provider/admin/cms/tab_content_title', locals: { header: "Built-in page '#{@page.title}'" } %>
```

mayorova (approved):
:+1:

PR #3471, opened 2023-07-17
THREESCALE-9952 completed feature test

nidhi-soni1104:
/features/old/api/features.feature

Jira issue
https://issues.redhat.com/browse/THREESCALE-9952

Notes:
Scenario: Delete a feature
This feature is failing at last line. I am unable to figure out the reason because I can see link when I run test in browser

```
Problem is 
Unable to find link "Delete" (Capybara::ElementNotFound)

```
see this screenshot and let me know if any suggestions
image


features/old/api/features.feature
```diff
@@ -8,59 +8,60 @@ Feature: Features :)
     And an application plan "Basic" of provider "foo.3scale.localhost"
     Given current domain is the admin domain of provider "foo.3scale.localhost"
 
-  @javascript @wip
+  @javascript
   Scenario: Create a feature
+    Given a service plan "Basic" of provider "foo.3scale.localhost"
```
mayorova, line 13: I think this is not needed here. This does not affect the functionality of this specific scenario, as `I go to the edit page for plan "Basic"` will pick the **first** plan, which in this case will be the application plan

features/old/api/features.feature
```diff
@@ -8,59 +8,60 @@ Feature: Features :)
...
+    And I fill in "System name" with "tee"
     And I fill in "Description" with "T-shirt with logo of our company for free."
-    And I press "Create Feature"
-    Then I should see enabled feature "Free T-shirt"
-    And provider "foo.3scale.localhost" should have feature "Free T-shirt"
-    And feature "Free T-shirt" should be enabled for plan "Basic"
+    And I press "Save"
+    Then I should see "T-shirt with logo of our company for free."
 
   # TODO: Test that the "No features yet" notice dissapears when first feature is created,
```
mayorova, line 23: I think it would be nice to implement this TODO, as we are on this test.

For reference, here is how it looks like:
Screenshot from 2023-07-31 14-47-52

features/old/api/features.feature
```diff
@@ -8,59 +8,60 @@ Feature: Features :)
...
-    And feature "Light-speed connection" is disabled for plan "Basic"
+    Given a service plan "Basic" of provider "foo.3scale.localhost"
+    And a feature "50% less bugs" of provider "foo.3scale.localhost"
+
     When I log in as provider "foo.3scale.localhost"
-    And I go to the edit page for plan "Basic"
-    And I press the enable button for feature "Light-speed connection"
-    Then I should see enabled feature "Light-speed connection"
-    And feature "Light-speed connection" should be enabled for plan "Basic"
+    Then I go to the edit page for plan "Basic"
```
mayorova, line 41: This goes to the Application plan Basic page, but then the other steps are for service plans... this is not right.

features/old/api/features.feature
```diff
@@ -8,59 +8,60 @@ Feature: Features :)
...
+    And I enable feature "50% less bugs"
```
mayorova, line 42: These steps "I enable feature", "should be enabled" etc. only test the logic on the model level, they have nothing to do with the UI (which is the main purpose of the cucumber tests), and as I mentioned earlier these steps for some reason are for the service plans, while in the UI the page that is loaded is the application plan.

The steps should interact with the UI elements - in case of enabling/disabling the feature it should be clicking on the icon in the **Enabled?** column of the features table and verifying that the icon has changed (and additionally we can check if on the DB level the attribute has changed accordingly - to prove that it's not only a visual thing).

features/old/api/features.feature
```diff
@@ -8,59 +8,60 @@ Feature: Features :)
...
-    Then I should not see feature "Invulnerability"
-    And provider "foo.3scale.localhost" should not have feature "Invulnerability"
+    Given the provider has "service_plans" switch allowed
+    Given a service plan "Basic" for service "API" exists
+    And a feature "Invulnerability" of provider "foo.3scale.localhost"
 
+    When I log in as provider "foo.3scale.localhost"
+    And I go to the service plans admin page
+    Then I go to the edit page for admin service plan "Basic"
+    And I follow "Delete" and I confirm dialog box
```
mayorova, line 67: The issue here is that it says `follow "Delete"` but the "Delete" is a button, not just a link, so it should be `I press "Delete"`

mayorova (changes requested):
See inline comments

features/old/api/features.feature
nidhi-soni1104: correct

features/old/api/features.feature
nidhi-soni1104: Okay  let me recheck this

features/old/api/features.feature
nidhi-soni1104: Actually this is changed if you see in step and I forgot to update here but good you pointed it

features/step_definitions/features_steps.rb
```diff
@@ -0,0 +1,45 @@
...
+  feature_name = @service_plan.service.features.last.name
+  assert_not_equal feature_name, string
+end
+
+Then('there {is} feature named {string}') do |is, string|
+  feature_name = @service_plan.service.features.last.name
+  assert_equal feature_name, string
+end
+
+And('I disable feature {string}') do |string|
```
mayorova, line 25: 1. The `string` is not used anywhere, you should use it, because otherwise you might be clicking the wrong feature

features/step_definitions/features_steps.rb
```diff
@@ -0,0 +1,45 @@
...
+  element = find('.fa-check-circle')
```
mayorova, line 26: 2. This selects the icon `<i>`, I guess clicking on the icon does nothing.
The actual action is triggered by clicking on `<a>` which is the parent of this icon.

Taking into account point 1., I'd say you should actually first find the `tr` element - as the row that contains the name of the feature (that you capture in `{string}`), and from there find the `<a>` that corresponds to the icon.

features/old/api/features.feature
```diff
@@ -8,59 +8,63 @@ Feature: Features :)
...
-    And provider "foo.3scale.localhost" should have feature "Free T-shirt"
-    And feature "Free T-shirt" should be enabled for plan "Basic"
+    And I press "Save"
+    Then I should see "T-shirt with logo of our company for free."
 
   # TODO: Test that the "No features yet" notice dissapears when first feature is created,
   #       and appears when last one is deleted.
-
-  @javascript @wip
+  @javascript @chrome
```
mayorova, line 24: We can use `@chrome` for local testing, but shouldn't commit it.

features/step_definitions/features_steps.rb
```diff
@@ -0,0 +1,45 @@
...
+  element.click
+end
+
+Then('feature {string} should be enabled') do |string|
+  feature = @service_plan.service.features.find_by(name: string)
+  assert_equal 1, feature.features_plans.count
+end
+
+Then('feature {string} should be disabled') do |string|
+  using_wait_time(10) do
```
mayorova, line 41: So, I checked some other features with a similar behavior (AJAX calls that result in updating the UI), this is one example: https://github.com/3scale/porta/blob/master/features/old/api/metrics/visibility.feature#L72-L80

You see, its steps use `wait_for_requests` helper method https://github.com/3scale/porta/blob/master/features/step_definitions/metric_steps.rb#L144-L147

Also, you can use this feature's (metrics enable/disable) steps as an example to improve this step - rather than checking the models directly we need to check how enabling/disabling the metric gets reflected in the UI (by changing the icon).

Also, I found some interesting information here: https://github.com/teamcapybara/capybara#asynchronous-javascript-ajax-and-friends 
So, when we are querying with Capybara we also need to take care of the expressions we use to make sure the retry logic is triggered.

And finally, Rails 6 upgrade PR changes the Capybara's default wait time from 2 to 10, so maybe it will also help going forward: https://github.com/3scale/porta/blob/THREESCALE-7405-rails-6/features/support/capybara.rb#L31

features/step_definitions/features_steps.rb
mayorova: For example, you can check how such a step is implemented here using `xpath`: https://github.com/3scale/porta/blob/master/features/step_definitions/metric_steps.rb#L57-L59

features/old/api/features.feature
nidhi-soni1104: yes I pushed it by mistake not intentionally

features/old/api/features.feature
nidhi-soni1104: Added

features/old/api/features.feature
```diff
@@ -8,59 +8,80 @@ Feature: Features :)
...
-    Then I should see enabled feature "Free T-shirt"
-    And provider "foo.3scale.localhost" should have feature "Free T-shirt"
-    And feature "Free T-shirt" should be enabled for plan "Basic"
+    And I press "Save"
+    Then I should see "T-shirt with logo of our company for free."
 
-  # TODO: Test that the "No features yet" notice dissapears when first feature is created,
-  #       and appears when last one is deleted.
+  @javascript
+  Scenario: "No features yet" notice dissapears when first feature is created
```
jlledom, line 23: Typo:
```suggestion
  Scenario: "No features yet" notice disappears when first feature is created
```

nidhi-soni1104:
> I think these tests were intended to test the application plans screen, not the service plans screen.



> I think these tests were intended to test the application plans screen, not the service plans screen.

Umm we have same feature for both but we added test for service plan.

nidhi-soni1104:
@mayorova can we merge this??

features/old/api/features.feature
```diff
@@ -8,59 +8,80 @@ Feature: Features :)
     And an application plan "Basic" of provider "foo.3scale.localhost"
     Given current domain is the admin domain of provider "foo.3scale.localhost"
 
-  @javascript @wip
+  @javascript
   Scenario: Create a feature
     When I log in as provider "foo.3scale.localhost"
-    And I go to the edit page for plan "Basic"
+    Then I go to the edit page for plan "Basic"
```
mayorova, line 14: Why this change? I think it's not needed, as it still describes conditions, not the expected outcome.

features/old/api/features.feature
```diff
@@ -8,59 +8,80 @@ Feature: Features :)
...
-  #       and appears when last one is deleted.
+  @javascript
+  Scenario: "No features yet" notice disappears when first feature is created
+    Given a service plan "Basic" of provider "foo.3scale.localhost"
+    Given the provider has "service_plans" switch allowed
+    And a feature "50% less bugs" of provider "foo.3scale.localhost"
+    When I log in as provider "foo.3scale.localhost"
+    And I go to the service plans admin page
+    And I follow "Basic"
+    Then I should not see "No features yet"
```
mayorova, line 30: The exact message is "This plan has no features yet.". I think we need to have the exact message.

mayorova:
> I think these tests were intended to test the application plans screen, not the service plans screen.

> Umm we have same feature for both but we added test for service plan.

I agree with @jlledom that it's a bit confusing that we have both application plans and service plans in the same test. The features indeed work in the same way in both service plans and application plans (and account plans too). And to be honest I don't see the need to deal with service plans here, as it requires more steps (enabling the switch, adding more feature steps, etc.)

I would rather stick with the application plan, so the test is consistent.

nidhi-soni1104:
> > I think these tests were intended to test the application plans screen, not the service plans screen.
> 
> > Umm we have same feature for both but we added test for service plan.
> 
> I agree with @jlledom that it's a bit confusing that we have both application plans and service plans in the same test. The features indeed work in the same way in both service plans and application plans (and account plans too). And to be honest I don't see the need to deal with service plans here, as it requires more steps (enabling the switch, adding more feature steps, etc.)
> 
> I would rather stick with the application plan, so the test is consistent.

@mayorova @jlledom You guys want me to fix all these test for Application Plan now, It would be waste of time working on all test again, as if behavior is already same

features/old/api/features.feature
nidhi-soni1104: done

mayorova:
> @mayorova @jlledom You guys want me to fix all these test for Application Plan now, It would be waste of time working on all test again, as if behavior is already same

1. It is worth it if it makes the tests more consistent, actually more correct, and more readable. It is a problem when we find tests written a while ago that are flaky, or that are hard to understand. So, for new tests we need to avoid this.

2. Most of the steps are already there, so we don't need to do it all from scratch. Just reusing and tweaking a bit what we already have, we can get much better tests that are more coherent.

I've opened a PR with my proposal for the tests improvements: https://github.com/3scale/porta/pull/3508/files

I will leave some inline comments on this PR to explain why I thought that some previous steps needed an improvement.

features/old/api/features.feature
```diff
@@ -8,59 +8,80 @@ Feature: Features :)
...
-    And provider "foo.3scale.localhost" should have feature "Free T-shirt"
-    And feature "Free T-shirt" should be enabled for plan "Basic"
+    And I press "Save"
+    Then I should see "T-shirt with logo of our company for free."
 
-  # TODO: Test that the "No features yet" notice dissapears when first feature is created,
-  #       and appears when last one is deleted.
+  @javascript
+  Scenario: "No features yet" notice disappears when first feature is created
+    Given a service plan "Basic" of provider "foo.3scale.localhost"
```
mayorova, line 24: As I mentioned previously, we don't need a service plan. The features behavior is pretty much the same. We already have an **application** plan called "Basic", defined in the scenario background, so there is absolutely no need to create another **service** plan (which also requires some switches enabling), because not only it does not help, but it also can make the things more confusing:

- A question may rise: why are we suddenly switching to service plans, if there is an application plan in the background and in the first test
- Two plans with the same name "Basic" may also get confused if some steps do not take the type of plan into account

features/old/api/features.feature
```diff
@@ -8,59 +8,80 @@ Feature: Features :)
...
+    Given the provider has "service_plans" switch allowed
+    And a feature "50% less bugs" of provider "foo.3scale.localhost"
```
mayorova, line 26: This step is very confusing as the description is very generic (doesn't say what plan type does the feature belong to - and there are 3 different types), but the implementation is very specific - it assumes not only the plan type - Service - but also searches for service plan name "Basic", and also for some reason assumes that the plan that we are interested is the last one.

```
  @service_plan = provider.service_plans.where(name: "Basic").last
  feature = @service_plan.service.features.find_or_create_by(name: feature_name, scope: "ServicePlan")
```

features/old/api/features.feature
```diff
@@ -8,59 +8,80 @@ Feature: Features :)
...
+    When I log in as provider "foo.3scale.localhost"
+    And I go to the service plans admin page
+    And I follow "Basic"
+    Then I should not see "This plan has no features yet."
+
+  @javascript
+  Scenario: "No features yet" notice appears when last one is deleted or no feature present.
```
mayorova, line 33: "when last one is deleted" flow is not covered, which is OK, but then the scenario title shouldn't say so :)

features/old/api/features.feature
```diff
@@ -8,59 +8,80 @@ Feature: Features :)
...
+    Then I should see "This plan has no features yet."
 
-  @javascript @wip
+  @javascript
   Scenario: Disable a feature
-    Given a feature "50% less bugs" of provider "foo.3scale.localhost"
-    And feature "50% less bugs" is enabled for plan "Basic"
+    Given a service plan "Basic" of provider "foo.3scale.localhost"
+    Given the provider has "service_plans" switch allowed
+    And an enabled feature "50% more bugs" of provider "foo.3scale.localhost"
```
mayorova, line 45: Same issue as the other step I referenced [here](https://github.com/3scale/porta/pull/3471/files#r1291236981)

features/support/paths.rb
```diff
@@ -247,6 +247,11 @@ def path_to(page_name, *args) # rubocop:disable Metrics/AbcSize, Metrics/Cycloma
       service = provider.services.find_by_name! $2
       polymorphic_path [$1.to_sym, :admin, service]
 
+    when /^the edit page for feature "([^"]+)"$/
+      plan = @provider.service_plans.where(name: "Basic").last
```
mayorova, line 251: This steps is not correct.

1. It assumes that the feature belongs to Service plans, and that there is a plan "Basic". That's not right.

2. This leads to a page that does actually have feature edit functionality, but it is NOT how we update features in the admin portal.

This page looks like this:

image

While in the product we update features thought this modal window:

image

features/support/paths.rb
```diff
@@ -269,6 +274,10 @@ def path_to(page_name, *args) # rubocop:disable Metrics/AbcSize, Metrics/Cycloma
       plan = Plan.find_by_name!($1)
       edit_polymorphic_path([:admin, plan])
 
+    when /^the edit page for admin service plan "([^"]*)"$/
```
mayorova, line 277: The previous step with the expression `/^the edit page for plan "([^"]*)"$/` can cover this scenario, so a new step is not necessary. **BUT!** the trick is that it will find the **first** plan with the defined name - the reason why we should avoid creating different plans (even of the same type) with the same names.

features/step_definitions/features_steps.rb
```diff
@@ -0,0 +1,51 @@
...
+  assert_not_equal feature_name, string
+end
+
+Then('there {is} feature named {string}') do |is, string|
+  feature_name = @service_plan.service.features.last.name
+  assert_equal feature_name, string
+end
+
+When("I {string} the feature {string}") do |enable_disable, feature_name|
+  feature = @service_plan.service.features.find_by(name: feature_name)
```
mayorova, line 26: This step should not assume that the feature belongs to service, and we should not know its ID. We are trying to replicate the users steps. The user will just see a table with different features in it, will find the row with the desired feature and click on the icon in that row. They will never care about the ID - it is all transparent.

So, the step should just do that - find the feature in the table by the name and click the icon in that row.

features/step_definitions/features_steps.rb
```diff
@@ -0,0 +1,51 @@
...
+    when 'disable'
+      find('i.included').click
+    end
+  end
+end
+
+Then('I see feature {string} is {enabled_or_disabled}') do |string, enabled_or_disabled|
+  wait_for_requests
+  feature = @service_plan.service.features.find_by(name: string)
+  within(:xpath, "//tr[@id='feature_#{feature.id}']") do
```
mayorova, line 41: The same thing I mentioned [here](https://github.com/3scale/porta/pull/3471/files#r1291256058) applies to this step.

features/step_definitions/features_steps.rb
```diff
@@ -0,0 +1,51 @@
...
+
+Given('an enabled feature {string} of {provider}') do |feature_name, provider|
+  @service_plan = provider.service_plans.where(name: "Basic").last
+  feature = @service_plan.service.features.find_or_create_by(name: feature_name, scope: "ServicePlan")
+  @service_plan.features << feature
+  @service_plan.save!
+end
+
+Then('there {is} no feature named {string}') do |is, string|  
+  feature_name = @service_plan.service.features.last.name
```
mayorova, line 16: Why "last" here? What happens if there are more than 1 features on the service?

Also, "is" argument is not used.

And if you wanted to use `{is}` parameter type, you could actually merge the two steps, no need to separate them, because it would match both "there is no feature" and "there is feature", and the variable `is` would be a boolean with `true` or `false`

features/step_definitions/features_steps.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+Given('a feature {string} of {provider}') do |feature_name, provider|
+  @service_plan = provider.service_plans.where(name: "Basic").last
+  feature = @service_plan.service.features.find_or_create_by(name: feature_name, scope: "ServicePlan")
+end
+
+Given('an enabled feature {string} of {provider}') do |feature_name, provider|
```
mayorova, line 8: This step is very similar to the one above, but just enables the feature additionally. So they could be merged, making the "enabled" optional. You could also use the same step for "disabled"

mayorova (approved):

mayorova (changes requested):
I made a mistake in https://github.com/3scale/porta/pull/3471/commits/d0bda86c4979f0a841e9e48f3893136c366483f6 but https://github.com/3scale/porta/pull/3471/commits/c6f8556d25dba51b30eae879f612ad307884c37c is not the right fix either 😬 

The name should be `plan_features_steps.rb`

mayorova (approved):

features/api/plan_features.feature
```diff
@@ -0,0 +1,75 @@
...
+    And I should see the flash message "Feature has been updated."
+
+  @javascript
+  Scenario: Delete a feature
+    Given application plan "Basic" has "Some Feature" enabled
+    When I log in as provider "foo.3scale.localhost"
+    And I go to the edit page for plan "Basic"
+    And I click "Delete" for the feature "Some Feature" and I confirm dialog box
+    Then I should not see "Some Feature"
+    And I should see the flash message "Feature has been deleted."
```
akostadinov, line 75: This test seems to mostly duplicate what we have in the first test except that in first test we don't check the flash message. This is fine in general but not sure we need to be such purists.

features/step_definitions/api/plan_features_steps.rb
```diff
@@ -0,0 +1,29 @@
...
+  within(:xpath, "//tr[td[text() = '#{feature_name}' and @class='title']]") do
+    if enable
+      find('i.excluded').click
+    else
+      find('i.included').click
+    end
+  end
+end
+
+When('I click {string} for the feature {string}') do |action, feature_name|
```
akostadinov, line 13: Do you think this step would be useful for anything but plans? Maybe this would be a little better?

```suggestion
When('I click {string} for the plan feature {string}') do |action, feature_name|
```

Same question applies to all new steps here.

akostadinov (approved):

features/api/plan_features.feature
mayorova: Hm... The first test is for creating, and this one is for deleting a feature. Why do you consider them duplicate?

features/step_definitions/api/plan_features_steps.rb
mayorova: I think features only apply to plans, so yeah, feature == plan feature.
No problem, we can add this to make it less confusing.

features/api/plan_features.feature
akostadinov: Right, I actually meant
```
"No features yet" notice disappears when the first feature is created, and appears when last one is deleted.
```

But all is fine. Separation of tested items.

features/step_definitions/api/plan_features_steps.rb
mayorova: Done

features/api/plan_features.feature
```diff
@@ -0,0 +1,75 @@
...
+    And I go to the edit page for plan "Basic"
+    Then I should see "This plan has no features yet."
+
+    When application plan "Basic" has "Some Feature" enabled
+    And I go to the edit page for plan "Basic"
+    Then I should not see "This plan has no features yet."
+
+    When application plan "Basic" does not have any features
+    And I go to the edit page for plan "Basic"
+    Then I should see "This plan has no features yet."
```
josemigallas, line 35: Given, when, then is supposed to be used once not thrice. Spaces also suggests this Scenario should split.

features/api/plan_features.feature
```diff
@@ -0,0 +1,75 @@
...
+
+  @javascript
+  Scenario: Disable a feature
+    Given application plan "Basic" has "Some Feature" enabled
+    And I log in as provider "foo.3scale.localhost"
+    And I go to the edit page for plan "Basic"
```
josemigallas, line 41: Half steps are repeated and could be moved into a Background. The feature being enabled/disabled can be handled by using 2 different Rule.

features/step_definitions/api/plan_features_steps.rb
```diff
@@ -0,0 +1,29 @@
...
+
+When('I {enable} the plan feature {string}') do |enable, feature_name|
+  within(:xpath, "//tr[td[text() = '#{feature_name}' and @class='title']]") do
+    if enable
+      find('i.excluded').click
+    else
+      find('i.included').click
+    end
+  end
+end
```
josemigallas, line 11: This step is unnecessary and prone tu error (enabling a feature twice will raise an error). Moreover, if the step says "I enable feature X" then it is redundant to check "feature is enabled" right after.

It should be possible to use `click feature_x within features_selector`.

features/api/plan_features.feature
mayorova: Ugh... then it would be just too much repetition... :unamused: 
I'll try smth...

features/api/plan_features.feature
mayorova: Sorry, I'm not sure what you mean here.

features/step_definitions/api/plan_features_steps.rb
mayorova: > This step is unnecessary and prone tu error (enabling a feature twice will raise an error).

Well, I think that's fine. The user's action is specific - either enable or disable. If the step is "enable the feature", but the "enable" button is not visible, that means that the background was not set up properly, and I would expect this step to fail.

> Moreover, if the step says "I enable feature X" then it is redundant to check "feature is enabled" right after.

I don't think it's redundant. `I {enable} the plan feature {string}` triggers a server action, which may work as expected, or may not. The `I see the plan feature {string} is {enabled}` verifies that after executing the previous step achieved the desired result.

> It should be possible to use `click feature_x within features_selector`.

But what element do you suggest to click on? The clickable item that is tested in this scenario is this cross/check (toggable) icon in the **Enabled?** column:
image

I don't think the action should be generic. For example, if the feature is already disabled, I will actually only see the "enabled" (checkbox) icon. Clicking on "icon in the Enabled? column" IMO is confusing.

features/api/plan_features.feature
mayorova: Done

features/api/plan_features.feature
jlledom: @josemigallas `I log in as provider "foo.3scale.localhost"` and `I go to the edit page for plan "Basic"` appear in all scenarios, but only the "I login" step could be moved to the background. The "go to edit" step only makes sense after the plan has been properly configured, which is different in every scenario, so that step can't be moved to the background. About using 2 different Rule, I don't know what you mean neither.

features/step_definitions/api/plan_features_steps.rb
jlledom: You could add the assertion `assert_selector('i.included[title="Feature is enabled for this plan"]', visible: true)` inside the `I {enable} the plan feature {string}` step. I saw that in many steps before ([example](https://github.com/3scale/porta/blob/36cf0b2b1a633d358ddb7d4ca57b6bac3323a282/features/step_definitions/backend_api_steps.rb#L64-L68), [example](https://github.com/3scale/porta/blob/a788c5bb5aa02890e98ce90b6ea468a3bffe202d/features/step_definitions/session_steps.rb#L48-L55)), but not sure that's a good practice or the expected way to do it.

You should check for the flash message anyway.

features/step_definitions/api/plan_features_steps.rb
```diff
@@ -0,0 +1,29 @@
...
+
+Then('I see the plan feature {string} is {enabled}') do |feature_name, enabled|
+  wait_for_requests
+  within(:xpath, "//tr[td[text() = '#{feature_name}' and @class='title']]") do
+    if enabled
+      assert_selector('i.included[title="Feature is enabled for this plan"]', visible: true)
+    else
+      assert_selector('i.excluded[title="Metric is disabled for this plan"]', visible: true)
```
jlledom, line 26: `Metric`? 😅 Maybe this is a good opportunity to fix this message.

jlledom (approved):

features/step_definitions/api/plan_features_steps.rb
mayorova: To me this practice seems a bit weird, I prefer to separate the action and checking the outcome. But I am not a Cucumber expert in any way :grimacing:

features/step_definitions/api/plan_features_steps.rb
mayorova: Oh... :sweat_smile: :hankey: 
That's a nice catch @jlledom thank you! :detective:

features/api/plan_features.feature
mayorova: I moved both `log in` and `has "Some Feature" enabled` steps to the Background.
`I go to the edit page for plan "Basic"` is still repeated in all steps, but in 2 (out of 7) they are not the 1st step.

I guess that by Rule you might mean [Gherkin Rules](https://cucumber.io/docs/gherkin/reference/#rule). To be honest, I don't think it's semantically applicable in this case - there is no generic "business rule" that would be grouping here. I think adding them here just for extracting one step to background would make the scenarios more confusing, and IMO is not worth it.

jlledom (approved):

mayorova:
@josemigallas I think I addressed your comments now, can you please recheck? :pray:

PR #3473, opened 2023-07-18
THREESCALE-9960 removed old feature

nidhi-soni1104:
**This PR fix failure of wip feature of file**
/features/old/public_site_access.feature

**Jira issue**
https://issues.redhat.com/browse/THREESCALE-9960

I removed this test because they no longer exist, I tried reproducing them but no such links are seen  .
If anyone knows existence of this feature let me know how to reproduce.

### DIFF
--- a/features/old/public_site_access.feature
+++ b/features/old/public_site_access.feature
@@ -8,7 +8,7 @@ Feature: Public site access
 
   @javascript
   Scenario: View site when site access code is set
-    And I follow "Developer Portal"
+    When I follow "Developer Portal"
     And I follow "Visit Portal"
     Then the current domain in a new window should be foo.3scale.localhost
     And I should not see field "Access code"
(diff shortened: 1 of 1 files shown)


jlledom (changes requested):
About the deleted tests. I don't really know what are they suppose to test. There are no links to the Developer Portal in the Dashboard AFAIK. Also no "User widget".

The last one is more clear, it should test that the developer portal is accessible from a port different from 80. But I don't know if we support that. If we do, then it would be good to have a test for that. But I assume writing such test is going to be far from easy.

So I'm okay with removing these tests.

nidhi-soni1104:
@jlledom Can we merge this?

jlledom:
> @jlledom Can we merge this?

Could you ask for another review? I'm not confident about these tests. I don't know what were they supposed to test so I'm not sure whether we should delete them or fix them.

akostadinov:
@josemigallas , do you remember these widgets or whatever? Is this something removed or changed?

jlledom (approved):

PR #3475, opened 2023-07-19
enable bullet in testing

akostadinov:
Enable N+1 detection in testing. We need to empty the safelist over time, see THREESCALE 9973

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -163,7 +163,6 @@ gem 'slim-rails', '~> 3.2'
 gem 'draper', '~> 3.0'
 
 group :development do
-  gem 'bullet', '~> 6.1.5'
   gem 'listen'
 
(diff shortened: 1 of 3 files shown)


jlledom (approved):
It looks good to me. The script is very cool.

Don't you think we could remove some tests that check for N+1 problems? like this one:

https://github.com/3scale/porta/blob/master/test/integration/admin/api/application_plan_limits_controller_test.rb#L39-L41

Those are redundant now, right?

akostadinov:
> Don't you think we could remove some tests that check for N+1 problems?

Some might be redundant. On the other hand how do we make sure that conditions are triggered for N+1 without them. I agree it is highly likely that this one you linked to is redundant. But is it worth evaluating them?

PR #3476, opened 2023-07-19
THREESCALE-9655: Add primary keys to all tables

jlledom:

Adding primary keys to all tables is required to use the Group Replication feature in MySQL.

[THREESCALE-9655](https://issues.redhat.com/browse/THREESCALE-9655)

**Verification steps** 

Query the tables DDL and check the primary keys are there.

I ran these queries to find tables without a PK:

MySQL:
```
select tab.table_schema as database_name, tab.table_name, tco.constraint_name as pk_name, group_concat(kcu.column_name order by kcu.ordinal_position separator ', ') as columns from information_schema.tables tab left join information_schema.table_constraints tco on tab.table_schema = tco.table_schema and tab.table_name = tco.table_name and tco.constraint_type = 'PRIMARY KEY' left join information_schema.key_column_usage kcu on tco.constraint_schema = kcu.constraint_schema and tco.constraint_name = kcu.constraint_name and tco.table_name = kcu.table_name where tab.table_schema not in ('mysql', 'information_schema', 'performance_schema', 'sys') and tab.table_schema = '3scale_system_development' and tco.CONSTRAINT_NAME is NULL group by tab.table_schema, tab.table_name, tco.constraint_name order by tab.table_schema, tab.table_name;
```

PostgreSQL:
```
select tbl.table_schema, tbl.table_name from information_schema.tables tbl where table_type = 'BASE TABLE' and table_schema not in ('pg_catalog', 'information_schema') and not exists (select 1 from information_schema.key_column_usage kcu where kcu.table_name = tbl.table_name and kcu.table_schema = tbl.table_schema)
```

Oracle:
```
select ut.table_name from user_tables ut where not exists (select 1 from user_constraints ac where ac.table_name = ut.table_name and ac.constraint_type = 'P' );
```

The queries only return two tables: `backend_events` and `features_plans`.

For `backend_events`, the migration reuses the already existing `id` column and unique index as primary key. The only difference between `backend_events` and other tables is the primary key is not autonumeric, and `BackendEvent.create!` raises and exception.

`features_plans`, on the other hand, is a many-to-many table between `features` and `plans` and doesn't have any id column that could be used as PK. I thought on two approaches to add a PK to it:

1. Add a new column `:id`:
   - Pros: Rails standard way
   - Cons: The migration should backfill the id column with values for all existing records in the table. Currently more than 200k records in SaaS.
   
2. Use `plan_id + feature_id` as PK:
   - Pros: Doesn't need a backfill
   - Cons:
     - Currently there's no unique index over `plan_id + feature_id`, so there could be repeated records that would made the migration fail on production. I checked our SaaS DB and we don't have any, but we don't know about clients on premises.
     - `plan_id` and `feature_id` were non nullable columns on MySQL and Postgres, but not on Oracle, so there could be some client having nulls on any of those columns. In that case, the migration would fail for them. After the migration, these columns are not-nullable is Oracle as well.
     - I would have liked to add `tenant_id` to the PK, but that column is nullable and we have more than 3K records in SaaS with `tenant_id = NULL`. So it's not an option. Anyway, I think it's OK because plans belong to tenants, so there should never be a same `plan_id` with different `tenant_id`. As I understand it.

I chose the second approach, to avoid the backfill.


db/migrate/20230718082519_add_primary_key_to_backend_events.rb
```diff
@@ -0,0 +1,25 @@
...
+    if System::Database.mysql?
+      execute 'ALTER TABLE backend_events DROP PRIMARY KEY'
+    elsif System::Database.postgres?
+      execute 'ALTER TABLE backend_events DROP CONSTRAINT backend_events_pk'
+    else
+      execute 'ALTER TABLE BACKEND_EVENTS DROP PRIMARY KEY'
+    end
+
+    add_index :backend_events, :id, unique: true
+  end
```
akostadinov, line 24: As last time, I don't think we should provide `down` here. It would be very unsafe to revert. And revert will only succeed if migration already succeeded. At which point we don't want to allow going back.

db/migrate/20230719112703_add_primary_key_to_features_plans.rb
```diff
@@ -0,0 +1,25 @@
...
+    if System::Database.mysql?
+      execute 'ALTER TABLE features_plans DROP PRIMARY KEY'
+    elsif System::Database.postgres?
+      execute 'ALTER TABLE features_plans DROP CONSTRAINT features_plans_pk'
+    else
+      execute 'ALTER TABLE FEATURES_PLANS DROP PRIMARY KEY'
+    end
+
+    add_index :features_plans, [:plan_id, :feature_id]
+  end
```
akostadinov, line 24: same

akostadinov (approved):
Looks great except optional comment.

db/migrate/20230718082519_add_primary_key_to_backend_events.rb
jlledom: Reverting is highly convenient for development. I'll remove the down method just before merging when I'm sure I'm not making more changes to the migrations.

app/models/backend_event.rb
```diff
@@ -1,7 +1,5 @@
 class BackendEvent < ApplicationRecord
 
-  self.primary_key = :id
```
mayorova, line 3: This is quite confusing :upside_down_face: 

I am wondering what was the intention here, and why the default `id` (as created by Rails) was not used here.

test/integration/user-management-api/application_plan_features_test.rb
```diff
@@ -105,10 +105,12 @@ def teardown
 
     test 'associate existing feature' do
       feat = FactoryBot.create(:feature, featurable: @provider.default_service)
+      feat2 = FactoryBot.create(:feature, featurable: @provider.default_service)
 
       assert_difference @app_plan.features.method(:count), 2 do
         post admin_api_application_plan_features_path(@app_plan, format: :xml), params: params.merge({ feature_id: feat.id })
-        post admin_api_application_plan_features_path(@app_plan, format: :xml), params: params.merge({ feature_id: feat.id })
+        result = post admin_api_application_plan_features_path(@app_plan, format: :xml), params: params.merge({ feature_id: feat2.id })
```
mayorova, line 112: is the `result` variable needed?

test/integration/user-management-api/service_plan_features_test.rb
```diff
@@ -44,15 +44,6 @@ def setup
     assert @service_plan.features.reload.include?(feat)
   end
 
-  test 'associate feature to service plan twice' do
```
mayorova, line 47: I guess this test is related to this controller?

https://github.com/3scale/porta/blob/1be51d3b3ff6d5f5e7f836d334c4b4a0082fc3cf/app/controllers/admin/api/featurings_base_controller.rb#L9-L12

So, what would happen after adding a PK? Maybe we need to add another test that would verify the behavior?

app/models/backend_event.rb
jlledom: I think they wanted a primary key, but not an autonumeric `id`, which is the default in Rails. So they did a hack to have something equivalent to a PK without the autonumeric. It seems to work that way now, probably because of the `default: nil` added to `create_table` [here](https://github.com/3scale/porta/pull/3476/files#diff-cbf8b25d6d853acf58fa9057841f407d29ffe90649c75cf9e3f51b2d6e3aa1d3R235)

test/integration/user-management-api/application_plan_features_test.rb
jlledom: No, :(

I added it for debugging purposes and forgot to remove it.

test/integration/user-management-api/service_plan_features_test.rb
mayorova: Yeah, I think the new behavior is correct (I mean - I doubt someone really relied on the old behavior returning a 200 OK to adding the same feature twice)

But yes, it's just the matter of adding a test to know what is actually returned now (and having the test would confirm that the behavior is expected).

jlledom: Done!

test/integration/user-management-api/application_plan_features_test.rb
jlledom: Fixed!

test/integration/user-management-api/service_plan_features_test.rb
```diff
@@ -44,13 +44,13 @@ def setup
...
+  test "can't associate feature to service plan twice'" do
```
mayorova, line 47: Trailing `'` can be removed :)

test/integration/user-management-api/application_plan_features_test.rb
```diff
@@ -105,10 +105,11 @@ def teardown
 
     test 'associate existing feature' do
       feat = FactoryBot.create(:feature, featurable: @provider.default_service)
+      feat2 = FactoryBot.create(:feature, featurable: @provider.default_service)
 
       assert_difference @app_plan.features.method(:count), 2 do
         post admin_api_application_plan_features_path(@app_plan, format: :xml), params: params.merge({ feature_id: feat.id })
-        post admin_api_application_plan_features_path(@app_plan, format: :xml), params: params.merge({ feature_id: feat.id })
+        post admin_api_application_plan_features_path(@app_plan, format: :xml), params: params.merge({ feature_id: feat2.id })
```
mayorova, line 112: Looking at the original test - does it mean that you could "attach" the same feature to the plan twice, and it would look like two features?... :thinking: 

So, this would mean 2 entries in the table with the exact same columns?

jlledom: I don't know how would it look like from UI, but yes, two repeated rows in the table. That's why the test failed in this branch, because of the PK.

test/integration/user-management-api/service_plan_features_test.rb
jlledom: ouch!

jlledom: Fixed!

test/integration/user-management-api/application_plan_features_test.rb
mayorova: I checked this, and the UI is immune to this duplication, but API does return repeated entries. This certainly looks like a bug to me... 

I think we probably need to include this change in the release notes for on-premises, and also make support aware of this change, so they can assist customers whose migrations might fail because of this (I hope there will be no issues, but who knows... )

db/migrate/20230718082519_add_primary_key_to_backend_events.rb
```diff
@@ -0,0 +1,25 @@
+class AddPrimaryKeyToBackendEvents < ActiveRecord::Migration[5.2]
+  def up
```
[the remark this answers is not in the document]
mayorova: OK, it works with `SAFETY_ASSURED=1`, just the error message is misleading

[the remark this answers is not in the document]
jlledom: Yes, it's the `strong_migrations` gem alerting of a potentially dangerous operation. You can bypass the control by running:

`SAFETY_ASSURED=1 bundle exec rails db:migrate`

However, it already removed the index so it will fail anyway if you try to run migration again. So you need to manually add the index before trying. Or resetting the DB to before the migration.

That's why I was complaining about not having DDL transactions.

test/integration/user-management-api/application_plan_features_test.rb
mayorova: Example error on migration:
```
$ SAFETY_ASSURED=1 bundle exec rails db:migrate:up VERSION=20230719112703
== 20230719112703 AddPrimaryKeyToFeaturesPlans: migrating =====================
-- remove_index(:features_plans, [:plan_id, :feature_id])
   -> 0.0163s
-- execute("ALTER TABLE features_plans ADD PRIMARY KEY (plan_id, feature_id)")
rails aborted!
StandardError: An error has occurred, all later migrations canceled:

Mysql2::Error: Duplicate entry '8-23' for key 'PRIMARY': ALTER TABLE features_plans ADD PRIMARY KEY (plan_id, feature_id)
/home/dmayorov/Projects/3scale/porta/db/migrate/20230719112703_add_primary_key_to_features_plans.rb:6:in `up'
bin/rails:4:in `require'
bin/rails:4:in `<main>'

Caused by:
ActiveRecord::RecordNotUnique: Mysql2::Error: Duplicate entry '8-23' for key 'PRIMARY': ALTER TABLE features_plans ADD PRIMARY KEY (plan_id, feature_id)
/home/dmayorov/Projects/3scale/porta/db/migrate/20230719112703_add_primary_key_to_features_plans.rb:6:in `up'
bin/rails:4:in `require'
bin/rails:4:in `<main>'

Caused by:
Mysql2::Error: Duplicate entry '8-23' for key 'PRIMARY'
/home/dmayorov/Projects/3scale/porta/db/migrate/20230719112703_add_primary_key_to_features_plans.rb:6:in `up'
bin/rails:4:in `require'
bin/rails:4:in `<main>'
Tasks: TOP => db:migrate:up
```

test/integration/user-management-api/application_plan_features_test.rb
```diff
@@ -105,10 +105,11 @@ def teardown
 
     test 'associate existing feature' do
```
mayorova, line 106: Actually, after changing the test the test name is not correct anymore, because the test is not associating the existing feature.

I guess it should probably try to post the same `feat`, and assert that count returns 1. Or additionally assert that response is 409, like we did with service plan features test.

db/migrate/20230718082519_add_primary_key_to_backend_events.rb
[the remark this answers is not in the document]
mayorova: Yep, I see :upside_down_face:

test/integration/user-management-api/application_plan_features_test.rb
jlledom: Agreed. I'll work on a query or script to hand to clients if they run into this problem. Fortunately, there are no repeated records in SaaS, which is a good indicator the code is working properly.

test/integration/user-management-api/application_plan_features_test.rb
jlledom: Done!

test/integration/user-management-api/application_plan_features_test.rb
mayorova: Great, thank you!

jlledom: I'm adding this to the release notes, let me know if you have some comments:

----
The migration could fail due to duplicate entries in the `features_plans` table. These are the errors raised by the DB engine in this case:

MySQL:
`Mysql2::Error: Duplicate entry '8-23' for key 'PRIMARY': ALTER TABLE features_plans ADD PRIMARY KEY (plan_id, feature_id)`

Postgres:
```
PG::UniqueViolation: ERROR:  could not create unique index "features_plans_pk"
DETAIL:  Key (plan_id, feature_id)=(8, 3) is duplicated.
: ALTER TABLE features_plans ADD CONSTRAINT features_plans_pk PRIMARY KEY (plan_id, feature_id)
```

Oracle:
`OCIError: ORA-02437: cannot validate (RAILS.SYS_C0010658) - primary key violated: ALTER TABLE FEATURES_PLANS ADD PRIMARY KEY (PLAN_ID, FEATURE_ID)`

The duplicated rows must be removed from the table before retrying to run the migration. The next query returns a list of duplicated rows (Valid for three DBs):
	
```
SELECT plan_id, feature_id, count(*)
FROM features_plans
GROUP BY plan_id, feature_id
HAVING count(*) > 1
```

When duplicated rows are found, we must keep one row for each of the results returned by the query above, and delete it's duplicates. The process to delete the duplicated rows is different depending on the DB engine and version. For instance, one can use "**ROW_NUM**" method for oracle or MySQL 8+, but only the "**Intermediate table**" method is available for MySQL 5.7. The next links summarize all possible situations:

MySQL:
https://www.mysqltutorial.org/mysql-delete-duplicate-rows/

PostgreSQL:
https://www.postgresqltutorial.com/postgresql-tutorial/how-to-delete-duplicate-rows-in-postgresql/

Oracle:
https://www.oracletutorial.com/advanced-oracle-sql/how-to-delete-duplicate-records-in-oracle/

In the worst scenario when removing duplicates isn't possible, this migration can be safely ignored unless the client wants to use "**Group Replication**" feature in MySQL.


Additionally, in Oracle the next error can happen when retrying the migration after failing because of duplicates:

`No indexes found on features_plans with the options provided.`

This can be solved by running the next SQL query and trying the migration again

`CREATE INDEX FEATURES_PLANS_PLAN_ID_IDX ON FEATURES_PLANS (PLAN_ID,FEATURE_ID);`

akostadinov (approved):
One minor question but things look alright to me.

app/models/features_plan.rb
```diff
@@ -6,9 +6,6 @@ class FeaturesPlan < ApplicationRecord
 
   validate :feature_scope_matches_plan_class?
 
-  # TODO: enable this on rails 3.1
-  # validates :plan_id, :plan_type, :feature_id, :presence => true
```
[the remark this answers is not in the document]
jlledom: The validation itself seems correct for me, but it's commented out so no effect currently, and since the comment says "enable on rails 3.1" and we are on rails 5, I considered that TODO is expired at this point.

Do you think we should keep the comment? or even uncomment the validation?

test/integration/user-management-api/application_plan_features_test.rb
jlledom: done!

akostadinov (dismissed):

mayorova (approved):
Great work! :+1:

PR #3485, opened 2023-07-24
THREESCALE-9592 show provide invoice years

nidhi-soni1104:

Reason mentioned in JIRA

https://issues.redhat.com/browse/THREESCALE-9592

We can remove this method as we are not using it any more 

**Pending** : test are pending if query looks good I can check for test

**Notes**
This is how it will look now, will show only current year if no invoices are present

image

this is how it was looking before

image

I don't have data on my local, so if any one have please past the screenshot how it will looks with data for reference but IMO query will work fine

### DIFF
--- a/app/controllers/finance/provider/invoices_controller.rb
+++ b/app/controllers/finance/provider/invoices_controller.rb
@@ -12,6 +12,7 @@ class Finance::Provider::InvoicesController < Finance::Provider::BaseController
   def index
     @search = ThreeScale::Search.new(params[:search] || params)
     @invoices = collection.scope_search(@search).order_by(params[:sort], params[:direction]).paginate(paginate_params).decorate
+    @years = Invoice.years_by_provider(current_account.id).presence || [(ActiveSupport::TimeZone.new(current_account.timezone) || Time.zone).now.year]
   end
 
   def create
--- a/features/provider/finance/invoices.feature
+++ b/features/provider/finance/invoices.feature
@@ -1,0 +1,38 @@
+@ignore-backend @javascript
+Feature: Provider's invoices
+  In order to manage invoices
+  As a provider
+  I want to be able to see all invoices issued by me
+
+  Background:
+    Given a provider is logged in on 1st January 2011
+    And the provider is charging its buyers in prepaid mode
+    And a buyer "zoidberg" signed up to provider "foo.3scale.localhost"
+
+  Scenario: Display years of invoices when there are invoices
+    Given an invoice of buyer "zoidberg" for January, 2011
+    And an invoice of buyer "zoidberg" for January, 2014 on 1st January 2014 (without scheduled jobs)
+    When the provider is at all provider's invoices page
+    Then invoices can be filtered by the following years:
+      |      |
+      | 2014 |
+      | 2013 |
+      | 2012 |
+      | 2011 |
+
+  Scenario: Display the current year when there are no invoices
+    When the provider is at all provider's invoices page
+    Then invoices can be filtered by the following years:
+      |      |
+      | 2011 |
+
+  Scenario: Display only years belonging to the current provider
+    Given an invoice of buyer "zoidberg" for January, 2011
+    And a provider "bar.3scale.localhost"
+    And provider "bar.3scale.localhost" is charging its buyers
+    And a buyer "leela" signed up to provider "bar.3scale.localhost"
+    And an invoice of buyer "leela" for January, 2012
+    When the provider is at all provider's invoices page
+    Then invoices can be filtered by the following years:
+      |      |
+      | 2011 |
--- a/features/step_definitions/finance/invoicing_steps.rb
+++ b/features/step_definitions/finance/invoicing_steps.rb
@@ -165,3 +165,9 @@ Then(/there is only one invoice for "([^"]*)"/) do |date|
   nodes = page.find_all(:xpath, ".//tr[contains(@class,'invoice')]/td[contains(text(), '#{date}')]")
   assert_equal 1, nodes.count
 end
+
+Then "invoices can be filtered by the following years:" do |table|
+  actual_years = find('#search_year').find_all('option').map(&:value).map(&:to_s)
+  expected_years = table.raw.flatten.map(&:to_s)
+  assert_same_elements expected_years, actual_years
+end
--- a/features/step_definitions/timehelpers_steps.rb
+++ b/features/step_definitions/timehelpers_steps.rb
@@ -41,11 +41,9 @@ When /^(?:the )?time flies to (.*)$/ do |date|
 end
 
 # Suffix 'on 5th July 2009'
-#
-Then /^(.+) on (\d+(?:th|st|nd|rd) \S* \d{4}(?: .*)?)$/ do |original, date|
-  # this ensures billing actions are run
-  step %(time flies to #{date})
-  # and then we freeze the time
+# When '(without scheduled jobs)' is present, scheduled jobs will be skipped when travelling in time
+Then /^(.+) on (\d+(?:th|st|nd|rd) \S* \d{4}(?: [^\(]*)?)( \(without scheduled jobs\))?$/ do |original, date, skip_jobs|
+  step %(time flies to #{date}) unless skip_jobs
   safe_travel_to(Time.zone.parse(date)) do
     step original.strip
   end
--- a/app/models/invoice.rb
+++ b/app/models/invoice.rb
@@ -587,9 +587,13 @@ class Invoice < ApplicationRecord
     (buyer_account || provider_account.buyer_accounts.build).field_label(name)
   end
 
-  # Returns years which have invoice
-  def self.years
-    self.connection.select_values(selecting { sift(:year, period).as('year') }.distinct.reorder('year DESC').to_sql).map(&:to_i)
+  # Returns years which have invoice, scoped by provider
+  def self.years_by_provider(provider_id)
+    connection.select_values(
+      selecting { sift(:year, period).as('year') }
+        .where.has { provider_account_id == provider_id }
+        .distinct.reorder('year DESC').to_sql
+    ).map(&:to_i)
   end
 
   protected
--- a/app/views/finance/provider/invoices/index.html.erb
+++ b/app/views/finance/provider/invoices/index.html.erb
@@ -24,8 +24,7 @@
         <td><%= s.text_field :buyer_query, :size => 20, :disabled => @buyer.present? %></td>
         <td>
           <%= select_month @search.month_number.try(:to_i), :prefix => :search, :field_name => :month_number, :include_blank => true %>
-          <%- years = Invoice.years %>
-          <%= select_year @search.year.try(:to_i), :prefix => :search, :field_name => :year, :include_blank => true, :start_year => years.first, :end_year => years.last %>
+          <%= select_year @search.year.try(:to_i), :prefix => :search, :field_name => :year, :include_blank => true, :start_year => @years.first, :end_year => @years.last %>
         <td></td>
         <td><%= s.select :state, Invoice.state_machine.states.keys.collect(&:to_s).sort, :include_blank => true, :selected => @search.state %></td>
         <td><%= s.button 'Search', :name => nil, :class => "pf-c-button pf-m-primary" %></td>
--- a/features/old/finance/invoices/deleted_account.feature
+++ b/features/old/finance/invoices/deleted_account.feature
@@ -18,7 +18,7 @@ Feature: Invoices of deleted account
       | Custom | 42   |
     And I issue the invoice number "2011-01-00000001"
     And account "bob" is deleted
-    And I go to the invoices issued by me
+    And I go to all provider's invoices page
     Then I should see 1 invoice
     When I follow "2011-01-00000001"
     Then I should see "2011-01-00000001"
--- a/features/step_definitions/finance/navigation_steps.rb
+++ b/features/step_definitions/finance/navigation_steps.rb
@@ -23,7 +23,7 @@ end
 
 # TODO: remove this legacy step
 When /^I navigate to invoices issued by me$/ do
-  step %(I go to the invoices issued by me)
+  step %(I go to all provider's invoices page)
 end
 
 When /^I navigate to my (?:earnings|revenue)$/ do
--- a/features/step_definitions/provider_steps.rb
+++ b/features/step_definitions/provider_steps.rb
@@ -284,3 +284,7 @@ Then /^(?:|I |they )should see inline error "([^"]*)" for ([^"]*)$/ do |text, in
   inline_error_selector = "li##{input_name.parameterize.underscore} p.inline-errors"
   assert_selector(inline_error_selector, text: text)
 end
+
+When "the provider is at {}" do |page_name|
+  visit path_to(page_name)
+end
--- a/features/support/paths.rb
+++ b/features/support/paths.rb
@@ -584,7 +584,7 @@ World(Module.new do
       account = Account.find_by!(org_name: $1)
       admin_buyers_account_invoices_path(account)
 
-    when /^the invoices issued by me$/
+    when /^all provider's invoices page$/
       admin_finance_invoices_path
 
     when /^the invoice "(.+?)" page$/
--- a/test/integration/provider/invoices_controller_test.rb
+++ b/test/integration/provider/invoices_controller_test.rb
@@ -10,8 +10,7 @@ class Finance::Provider::InvoicesControllerTest < ActionDispatch::IntegrationTes
     @provider_account.create_billing_strategy
     @provider_account.settings.allow_finance!
 
-    @invoice = FactoryBot.create(:invoice, buyer_account: @cinstance.buyer_account, provider_account: @cinstance.provider_account)
-    Invoice.any_instance.stubs(:find).returns(@invoice)
+    @invoice = FactoryBot.create(:invoice, buyer_account: @cinstance.buyer_account, provider_account: @provider_account)
 
     @line_item = FactoryBot.create(:line_item, invoice: @invoice, cost: 2000)
 
@@ -24,6 +23,17 @@ class Finance::Provider::InvoicesControllerTest < ActionDispatch::IntegrationTes
     assert_template 'finance/provider/invoices/index'
     assert_not_nil assigns(:invoices)
     assert_active_menus({main_menu: :audience, submenu: :finance, sidebar: :invoices, topmenu: :dashboard})
+  end
+
+  test 'default year when there are no invoices' do
+    Invoice.where(provider_account: @provider_account).destroy_all
+    @provider_account.update(timezone: 'Pacific Time (US & Canada)')
+
+    travel_to(Time.utc(2023, 1, 1))
+    get admin_finance_account_invoices_path @buyer
+
+    assert_empty assigns(:invoices)
+    assert_equal [2022], assigns(:years)
   end
 
   test 'list invoices by month' do


app/controllers/finance/provider/invoices_controller.rb
```diff
@@ -12,6 +12,11 @@ class Finance::Provider::InvoicesController < Finance::Provider::BaseController
   def index
     @search = ThreeScale::Search.new(params[:search] || params)
     @invoices = collection.scope_search(@search).order_by(params[:sort], params[:direction]).paginate(paginate_params).decorate
+    @years = Invoice.where(provider_account_id: current_account.id)
```
mayorova, line 15: Some notes:

1. I think `YEAR(period)` only works for MySQL, PostgreSQL and Oracle have other functions to extract the year. I suggest using BabySqueel, that same way that `Invoice.years` is currently using - in fact, I think the easiest would be to modify that method to incorporate `provider_account_id` in there, as this method would only be used by this controller (if I am not mistaken)

2. Well, that's also included in 1 - I suggest to move this logic to the model, and not the controller.

3. The query should return the years in order, because the view queries for `first` and `last`, and if the years are not ordered, the result will not be correct.

app/controllers/finance/provider/invoices_controller.rb
```diff
@@ -12,6 +12,11 @@ class Finance::Provider::InvoicesController < Finance::Provider::BaseController
...
+               .pluck("YEAR(period)")
+               .compact
+               .map(&:to_i)
+               .presence || [Time.now.year]
```
mayorova, line 19: better use `Time.zone.now.year`, with Time Zone

mayorova (changes requested):
This might work:

```
  # Returns years which have invoice, scoped by provider
  def self.years_by_provider(provider_id)
    connection.select_values(
      selecting { sift(:year, period).as('year') }
        .where.has { provider_account_id == provider_id }
        .distinct.reorder('year DESC').to_sql
    ).map(&:to_i)
  end
```
in `invoice.rb`

and
```
@years = Invoice.years_by_provider(current_account.id).presence || [Time.zone.now.year]
```
in the invoices controller.

Also, tests are needed

nidhi-soni1104:
@mayorova Test are added and requested changes are also adressed

features/old/accounts/invoices.feature
```diff
@@ -36,3 +36,18 @@ Feature: Show invoices from account's page (#16015909)
     And I follow "2011-01-00000001"
     # TODO: Properly check if order is opened
     Then I should see "Add"
+
+  Scenario: Display years of invoices when there are invoices
+    Given a provider is logged in on 1st January 2011
+    And the provider is charging its buyers in prepaid mode
+    And an application plan "Fixed" of provider "foo.3scale.localhost" for 0 monthly
```
mayorova, line 43: Subscribing to applications are not needed for these tests. So, this and the following steps should not be needed.

features/old/accounts/invoices.feature
```diff
@@ -36,3 +36,18 @@ Feature: Show invoices from account's page (#16015909)
...
+    And a buyer "zoidberg" signed up to application plan "Fixed"
+    And an invoice of buyer "zoidberg" for January, 2011
+    And I go to the invoices issued by me
+    Then I should see the list of years with invoices
```
mayorova, line 47: This PR specifically fixes the issue where the provider is seeing the years of invoices that this specific provider doesn't have.
So, I think this is what is needed here - add some invoices for another provider, and make sure that the current provider doesn't see the years of that another one.

features/step_definitions/finance/invoicing_steps.rb
jlledom, line 1: I think that, instead of these steps, it'll be easier if we had one single step to check the elements of the year selector. For instance, something like `The year dropdown should have the following elements`. Like in `header.feature`

https://github.com/3scale/porta/blob/3scale-2.13.2-GA/features/provider/header.feature#L8-L12

features/old/accounts/invoices.feature
```diff
@@ -36,3 +36,32 @@ Feature: Show invoices from account's page (#16015909)
...
+    And the provider is charging its buyers in prepaid mode
+    And an invoice of buyer "zoidberg" for January, 2011
+    And I go to the invoices issued by me
+    Then I should see the list of years with invoices
+
+  Scenario: Display the current year when there are no invoices
+    Given a provider is logged in on 1st January 2011
+    And the provider is charging its buyers in prepaid mode
+    And I go to the invoices issued by me
+    Then I should see the current year
```
jlledom, line 52: I think you should check explicitly only the current year is on the list

features/old/accounts/invoices.feature
nidhi-soni1104: Added

nidhi-soni1104:
@jlledom comments are addressed.

jlledom (changes requested):

features/old/accounts/invoices.feature
```diff
@@ -36,3 +36,38 @@ Feature: Show invoices from account's page (#16015909)
...
+      | 0  |
+      | 2011 |
+
+  Scenario: Display the current year when there are no invoices
+    Given a provider is logged in on 1st January 2011
+    And the provider is charging its buyers in prepaid mode
+    And I go to the invoices issued by me
+    Then I should see the list of years with invoices have the following years:
+      | 0  |
+      | current_year |
```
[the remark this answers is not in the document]
nidhi-soni1104: No No this will be 2023 not 2011.

if user don't have any invoice then we show current year i.e 2023 so this is correct .

features/step_definitions/finance/invoicing_steps.rb
```diff
@@ -165,3 +165,17 @@
   nodes = page.find_all(:xpath, ".//tr[contains(@class,'invoice')]/td[contains(text(), '#{date}')]")
   assert_equal 1, nodes.count
 end
+
+Then "I should see the list of years with invoices have the following years:" do |table|
+  actual_years = find('#search_year').find_all('option').map(&:value).map(&:to_i)
+  expected_years = table.raw.flatten.map do |value|
+    value == 'current_year' ? Time.now.year : value.to_i
+  end
```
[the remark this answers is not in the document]
nidhi-soni1104: check the above reason

features/old/accounts/invoices.feature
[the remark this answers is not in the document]
mayorova: The steps ideally should describe exactly what the user sees. So, having `current_year` is confusing. It should be an actual year. Here @jlledom suggested `2011` because this is what specified in the first step: Given a provider is logged in **on 1st January 2011** - in bold is the time helper, that sets the current date to the specified one.

Also, `0` should not be there either, because this is not what the user sees, and is not the actual year.

features/old/accounts/invoices.feature
```diff
@@ -36,3 +36,37 @@ Feature: Show invoices from account's page (#16015909)
...
+
+  Scenario: Display years of invoices when there are invoices
+    Given a provider is logged in on 1st January 2011
+    Given a buyer "zoidberg" signed up to provider "foo.3scale.localhost"
+    And the provider is charging its buyers in prepaid mode
+    And an invoice of buyer "zoidberg" for January, 2011
+    And I go to the invoices issued by me
+    Then I should see the list of years with invoices have the following years:
+      | 0  |
+      | 2011 |
```
mayorova, line 48: For this step I would add another invoice for a different year, e.g. 2016 (you may need to do some time machine trick to move to a different year, because there is a validation on invoice period that does not allow creating invoices too far into the future), and confirm that the list of the years is the range between those, in descending order, e.g. 

```
2016
2015
2014
2013
2012
2011
```

features/old/accounts/invoices.feature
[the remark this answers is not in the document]
nidhi-soni1104: so user is seeing 2023 as screenshot say and if user is logged in on 2011 but he don't have any invoice which is a correct case. 

And 0 is shown if you see the screenshot first blank is shown in UI see this screenshot

current year here used because of a reason, because in 2024 also the test will be valid if we have user with no invoice.

image

features/old/accounts/invoices.feature
nidhi-soni1104: yes 
` Billing Period must be between the provider account creation date and 12 months from now `

and so I have added 2 invoices which seem to test what we want

features/old/accounts/invoices.feature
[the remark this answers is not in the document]
jlledom: As @mayorova mentioned, the first step is freezing time at 2011, so `Time.now` will always be `1st January 2011`.

About the `0`, you can add it manually to `expected_years` in the step.

features/step_definitions/finance/invoicing_steps.rb
```diff
@@ -165,3 +165,15 @@
...
+  actual_years = find('#search_year').find_all('option').map(&:value).map(&:to_i)
+  expected_years = table.raw.flatten.map(&:to_i)
+  assert_same_elements expected_years, actual_years
+end
+
+Then("I should not see the list of years with invoices belonging to other providers") do
+  select_element = find('#search_year')
+  displayed_years = select_element.all('option').map(&:value).map(&:to_i)
+  expect(displayed_years).not_to include(*[2011])
+end
```
jlledom, line 179: This could be deleted now.

jlledom (changes requested):

features/step_definitions/finance/invoicing_steps.rb
nidhi-soni1104: removed

features/support/time_machine_helpers.rb
```diff
@@ -9,12 +9,15 @@ module TimeMachineHelpers
...
   # 3. Assert the user has been effectively removed
-  def time_machine(till)
-    freeze_time
-
-    while Time.zone.now < till
-      travel_to(1.day.from_now)
-      run_jobs
+  def time_machine(till, opts = {})
+    opts ||= {}
+    unless opts[:skip_jobs]
```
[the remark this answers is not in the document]
mayorova: I commented on it [here](https://github.com/3scale/porta/pull/3515/files#r1293389796)

Basically, when we do "time travelling" all jobs are executed (all daily jobs). I noticed, that it takes too long when travelling within several years, so added this options to skip jobs, as it's not needed for this test.

In fact, I am thinking we may switch to "no jobs" by default, and only enable it (by "(with jobs)" wherever needed. I think this might reduce the time for cucumbers significantly

features/old/accounts/invoices.feature
```diff
@@ -36,3 +36,41 @@ Feature: Show invoices from account's page (#16015909)
...
+    And the provider is charging its buyers in prepaid mode
+    And an invoice of buyer "zoidberg" for January, 2010
+
+    Given a provider "boo.3scale.localhost"
+    Given a buyer "zudio" signed up to provider "boo.3scale.localhost"
+    And the provider is charging its buyers in postpaid mode
+    And an invoice of buyer "zudio" for January, 2011
+
+    And I go to the invoices issued by me
+    Then the years selector should have the following options:
```
josemigallas, line 74: Empty lines inside scenarios are discouraged: it usually means is has too many step or it is written in a complex way. Some common steps could be extracted into a Background and others could be combined into a new step.

This scenario for should communicate better what it tests. This is a suggestion of what I would do IF NOT under /old

```suggestion
  Scenario: Display only years belonging to the current provider
    Given a buyer "zoidberg" of a provider "foo.3scale.localhost"
    And an invoice of buyer "zoidberg" for January, 2010
    And a buyer "zudio" of another provider "boo.3scale.localhost"
    And an invoice of buyer "zudio" for January, 2011
    When an admin logs in to provider "foo.3scale.localhost"
    And go to the invoices
    Then invoices can be selected from the following years:
```

features/old/accounts/invoices.feature
```diff
@@ -36,3 +36,41 @@ Feature: Show invoices from account's page (#16015909)
     And I follow "2011-01-00000001"
     # TODO: Properly check if order is opened
     Then I should see "Add"
+
```
josemigallas, line 39: We should avoid at all cost adding new features under `features/old` and instead create a new feature located at a proper location and with proper steps.

josemigallas (changes requested):

features/step_definitions/timehelpers_steps.rb
```diff
@@ -33,18 +33,19 @@ def access_user_sessions
...
+  time_machine(Time.zone.parse(date), time_machine_opts)
   step %(the date should be #{date})
   access_user_sessions
 end
 
 # Suffix 'on 5th July 2009'
 #
-Then /^(.+) on (\d+(?:th|st|nd|rd) \S* \d{4}(?: .*)?)$/ do |original, date|
+Then /^(.+) on (\d+(?:th|st|nd|rd) \S* \d{4}(?: [^\(]*)?)( \(without jobs\))?$/ do |original, date, skip_jobs|
   # this ensures billing actions are run
```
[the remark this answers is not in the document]
mayorova: What are "billing actions"?

This is to avoid all daily jobs running for every day from the initial date to the target date (a defined in `time_machine` method)

features/old/accounts/invoices.feature
mayorova: What would be the proper location and proper steps?

features/step_definitions/finance/invoicing_steps.rb
```diff
@@ -165,3 +165,9 @@
   nodes = page.find_all(:xpath, ".//tr[contains(@class,'invoice')]/td[contains(text(), '#{date}')]")
   assert_equal 1, nodes.count
 end
+
+Then "the years selector should have the following options:" do |table|
```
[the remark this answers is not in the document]
mayorova: > There's no such thing as a selector, it's either a select or a dropdown.

I disagree: https://dictionary.cambridge.org/dictionary/english/selector

In this context, selector is just that - some element on the page that allows the user to select something, years in this case. Indeed, it could be implemented as radio buttons, for example.

Although, to be honest, I don't see anything bad in even calling it a "drop-down list". Is there some rule that prohibits referencing UX/UI decisions in cucumber scenarios?

"invoices can be filtered by the following years:" to me it seems a big too vague :shrug:

features/old/accounts/invoices.feature
nidhi-soni1104: then do we need to fix cucumber under this features/old?

mayorova: Actually, here we 're not adding new features, we're just adding a couple of scenarios to the existing feature file.
But I can move it to another place that might be suitable.

mayorova (dismissed):

features/support/time_machine_helpers.rb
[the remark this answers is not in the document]
jlledom: I don't think we need to modify `time_machine` because if we want to travel in time without running jobs we can use `travel_to`. The point of `time_machine` is precisely to simulate the actual pass of time with jobs updating invoices, etc.

jlledom (changes requested):
I think the changes over `time_machine` should be reverted. Check my comment here: https://github.com/3scale/porta/pull/3485/files#r1301367529

app/controllers/finance/provider/invoices_controller.rb
```diff
@@ -12,6 +12,7 @@ class Finance::Provider::InvoicesController < Finance::Provider::BaseController
   def index
     @search = ThreeScale::Search.new(params[:search] || params)
     @invoices = collection.scope_search(@search).order_by(params[:sort], params[:direction]).paginate(paginate_params).decorate
+    @years = Invoice.years_by_provider(current_account.id).presence || [Time.zone.now.year]
```
akostadinov, line 15: It is a very minor thing, but can we use `Account#timezone` for the default value?

akostadinov (dismissed):

features/provider/finance/invoices.feature
```diff
@@ -0,0 +1,43 @@
...
+  In order to manage invoices
+  As a provider
+  I want to be able to see all invoices issued by me
+
+  Scenario: Display years of invoices when there are invoices
+    Given a provider is logged in on 1st January 2011
+    And a buyer "zoidberg" signed up to provider "foo.3scale.localhost"
+    And the provider is charging its buyers in prepaid mode
+    And an invoice of buyer "zoidberg" for January, 2011
+    And an invoice of buyer "zoidberg" for January, 2014 on 1st January 2014 (without jobs)
```
akostadinov, line 12: I don't understand this what the steps above should be doing? Why is the second step called `on 1st January 2014 (without jobs)`?

features/provider/finance/invoices.feature
```diff
@@ -0,0 +1,43 @@
...
+    When I go to the invoices issued by me
+    Then the years selector should have the following options:
+      |      |
+      | 2014 |
```
akostadinov, line 16: Better parse all table elements so that we don't have a header element. 
```
| 2014 |
| 2013 |
| 2012 |
| 2011 |
```

But while on it, why don't we only see `2014` and `2011`?

app/controllers/finance/provider/invoices_controller.rb
mayorova: Yeah, no problem

features/provider/finance/invoices.feature
mayorova: The step creates an invoice for the specific period (month+year). This `on 1st January 2014 (without jobs)` is needed so that this creation is executed on a certain date.

This is because of this validation for the invoice period :upside_down_face:  https://github.com/3scale/porta/pull/3073/files#diff-c7710675b799dc693083ef5075a5bb1fee822ddae19fe8a64a1be2f5a87bb6edR621-R628

mayorova: Ah, and "(without jobs)" is to avoid that all cron jobs execute from 1st Jan 2021 and 1st Jan 2014, which was taking too long :grimacing:

features/provider/finance/invoices.feature
mayorova: It is not a header, it is an empty element, which is what is seen in the UI:

Screenshot from 2023-09-12 14-46-50

Yes, having only `2011` and `2014` would be more logical, but this is how this [select_year](https://github.com/3scale/porta/blob/THREESCALE-9592-invoice-years/app/views/finance/provider/invoices/index.html.erb#L27) helper works, see https://github.com/rails/rails/blob/v6.0.6.1/actionview/lib/action_view/helpers/date_helper.rb#L638-L665

jlledom (approved):

features/step_definitions/finance/invoicing_steps.rb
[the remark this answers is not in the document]
josemigallas: I counter-disagree: https://dictionary.cambridge.org/dictionary/english/drop-down-menu

This is either a dropdown or a select, not a selector. I didn't mean selector does not exist as a word, but as a front-end/web term.

features/step_definitions/timehelpers_steps.rb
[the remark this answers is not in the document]
josemigallas: I don't know, it's what the comment says. But I assume billing actions are some kind of "scheduled job" and this step skips them so I figured this comment is no longer accurate because the step skips jobs, including billing ones. Does this make sense?

features/provider/finance/invoices.feature
```diff
@@ -0,0 +1,43 @@
+@ignore-backend @javascript
+Feature: Show all provider's invoices
+  In order to manage invoices
+  As a provider
+  I want to be able to see all invoices issued by me
+
+  Scenario: Display years of invoices when there are invoices
+    Given a provider is logged in on 1st January 2011
+    And a buyer "zoidberg" signed up to provider "foo.3scale.localhost"
```
josemigallas, line 9:
```suggestion
    And a buyer signed up to the provider
```

features/provider/finance/invoices.feature
```diff
@@ -0,0 +1,43 @@
...
+    And the provider is charging its buyers in prepaid mode
+    And an invoice of buyer "zoidberg" for January, 2011
+    And an invoice of buyer "zoidberg" for January, 2014 on 1st January 2014 (without scheduled jobs)
```
josemigallas, line 12:
```suggestion
    And an invoice of the buyer for January, 2011
    And an invoice of the buyer for January, 2014 on 1st January 2014 (without scheduled jobs)
```

features/provider/finance/invoices.feature
```diff
@@ -0,0 +1,43 @@
...
+    When I go to the invoices issued by me
```
josemigallas, line 13: We talked about not using `I...` and `me` steps anymore in new cucumbers. I suggest simply adding a new option in paths.rb:
https://github.com/3scale/porta/blob/f2df57d44306cefb9aac9aea92ae78e35a3c46a4/features/support/paths.rb#L587

Since the provider (not me) is the one logged in, I would say "the provider/admin is at the invoices page".

features/step_definitions/finance/invoicing_steps.rb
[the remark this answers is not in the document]
josemigallas: It's also worth mentioning that using "should" is discouraged. The steps should be imperative: should vs. is. Moreover, the purpose of this dropdown is to filter invoices by year, therefore I proposed `invoices can be filtered by the following years`.

features/step_definitions/timehelpers_steps.rb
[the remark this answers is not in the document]
mayorova: Aaaah, OK, thanks for the clarification, I didn't see that comment :eyes: :sweat_smile: 

Yes, this comment is confusing, I will remove it and will add another one clarifying the behavior of the step.

features/step_definitions/finance/invoicing_steps.rb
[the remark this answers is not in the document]
mayorova: > This is either a dropdown or a select, not a selector. I didn't mean selector does not exist as a word, but as a front-end/web term.

Yeah, OK, but my point is - I used it as a generic word - an element that is used to select some options, not a UI term.

I guess as a team we need to agree whether in our cucumbers we need to be more specific (referencing specific UI elements), or more generic. 

> It's also worth mentioning that using "should" is discouraged. 

When you reference statements like this, can you please also add some references to the documentation or whatever, so we all learn? Who discourages "should"? E.g. the docs here provide quite a lot of examples with "should": https://cucumber.io/docs/gherkin/reference/

features/provider/finance/invoices.feature
mayorova: @josemigallas if we are not having a "cucumber 101" session any time soon, can you please just pop all of your "cucumber guidelines" in a Google doc or something, so we can discuss/agree/use it going forward?

I know that it's the same comments in every PR, but it is very hard to remember such things, if they mainly just sit in your head :wink:

jlledom (approved):

features/provider/finance/invoices.feature
```diff
@@ -0,0 +1,43 @@
+@ignore-backend @javascript
+Feature: Show all provider's invoices
```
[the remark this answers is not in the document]
mayorova: Yeah, it's the index page showing all provider's invoices (as opposed to the pages with a single buyer's invoices).

If you think "Provider invoices" is clearer, that's OK with me.

features/provider/finance/invoices.feature
mayorova: There is an inconvenience in the `a buyer signed up to the provider`, and it's that it creates an application (which we don't need here), and actually this step fails with a:
```
Provider has no default application plan
```

We would need to add a default application plan to make it work, but again - we don't need an application in this scenario.

Also, for the last (and the most important) scenario `Display only years belonging to the current provider` it is better to have the buyer's names explicit, because it may be confusing, as the different buyers belong to different providers.

features/provider/finance/invoices.feature
mayorova: Explained above: https://github.com/3scale/porta/pull/3485#discussion_r1326541747

features/provider/finance/invoices.feature
mayorova: > I suggest simply adding a new option in paths.rb:

Hmm... not sure what you mean. The option in `paths.rb` was already there, that's why I just used the standard "I go to" approach. Or do you mean to change the expression for this path?

mayorova: I made some related changes in https://github.com/3scale/porta/pull/3485/commits/922896c5d2fc5b49308c006d1b9d8939a858c035
What do you think?

jlledom (approved):

akostadinov (approved):

mayorova:
@josemigallas I think I addressed your comments now, can you please recheck? :pray:

nidhi-soni1104:
@josemigallas can we merge this?

features/step_definitions/finance/invoicing_steps.rb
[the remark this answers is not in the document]
josemigallas: Of course, sorry. A while ago we talked about "how to write better cucumbers" according to them: https://cucumber.io/docs/bdd/better-gherkin/ and I try to follow these guidelines as much as possible.

josemigallas (approved):
Tentatively approved 😂

PR #3488, opened 2023-07-26
🦋 Implement Toolbar for Accounts index

josemigallas:
[THREESCALE-9766: Buyer accounts table toolbar](https://issues.redhat.com/browse/THREESCALE-9766)

Looking good baby!

### Old:
Screenshot 2023-11-09 at 08 43 34

### New:
Filtering and stuff:
Screenshot 2023-10-27 at 14 04 37
toolbar

Bulk operations working:
Screenshot 2023-11-09 at 14 39 18

Export CSV is hidden in the overflow dropdown:
Screenshot 2023-10-31 at 13 59 55


akostadinov:
Can we also show an error when a word with less than 2 characters was used?

mayorova:
Haven't completed the review yet, but I am already missing some things:
image

There seems to be no way to reset filtering by Plan and State

mayorova:
The current version has a note:

> Search for Group/Org., User login, First name, Last name, email, user_key, app_id or Application Name.

A terrible copy, but I think it's useful to understand where the specified keyword is searched in - otherwise, the results may be very confusing.

Maybe we could use a [tooltip](https://v4-archive.patternfly.org/v4/components/tooltip/design-guidelines/) or a [popover](https://v4-archive.patternfly.org/v4/components/popover/design-guidelines/) on the search input to add this info?

josemigallas:
@mayorova
Your concerns are very valid but it would require a re-design involving UX, docs and product. It is out of the scope of this JIRA/Epic. The objective is to remove old styles and clean up the hacks that "fake" Patternfly 4. A refactorization rather than a remake.

Some of your suggestions, like adding a tooltip to the searchbar or changing the behavior of dropdowns would have to be applied to all other tables for consistency. However, I am working on leveling the UI out, making all tables and common components equal, and adding specific changes now would disrupt this effort.
I would rather maintain everything everywhere as is and then do a generic "improve toolbar v2" that will affect _all_ tables at once, not just this one.

WDYT?

mayorova:
@josemigallas I really don't like these dropdowns (Plan and Status) and the fact that they can't be reset (by the way, I noticed now the "Clear all filters" in your screenshots, but I can't seem to have the showing up for me locally).
Maybe at least this option would be possible:

> This example includes the "titles": https://v4-archive.patternfly.org/v4/components/toolbar#component-managed-toggle-groups
> But they appear as if they were just elements to select, which I think is confusing too.

If you consider that this is also hard to do of bad-looking (it's definitely not ideal), that's fine. You just need to find another approver :grimacing: 

It's OK to have an "improvements" JIRA and improve that later.

josemigallas:
> I really don't like these dropdowns (Plan and Status) and the fact that they can't be reset

But they can be reset: click on the checked option and it will be un-selected.

However, I've found a better solution: Select supports a clear button (see last gif in description)

> by the way, I noticed now the "Clear all filters" in your screenshots, but I can't seem to have the showing up for me locally

It could be a bug. The clear all button should be rendered whenever the search returns nothing: https://github.com/3scale/porta/blob/f27412289fd919aa30ae7cbb55ac44eb291d4244/app/presenters/buyers/accounts_index_presenter.rb#L47-L49

mayorova:
> However, I've found a better solution: Select supports a clear button (see last gif in description)

That's really nice! Thanks!

> It could be a bug. The clear all button should be rendered whenever the search returns nothing:

Yeah, the Empty State section is showing OK when nothing is found. I was referring to the link inside the toolbar (like on the second screenshot).
I guess you might have added it, and then removed?

Also, I like how "Export to CSV" is collapsed into kebab. Can we always have it like that? :grimacing:

josemigallas:
> Yeah, the Empty State section is showing OK when nothing is found. I was referring to the link inside the toolbar (like on the second screenshot). I guess you might have added it, and then removed?
 
 Ah I see! Yes I removed it from the toolbar.

> Also, I like how "Export to CSV" is collapsed into kebab. Can we always have it like that? 😬

It can be done yes but I will have to think of a clean way of parameterize it.

mayorova:
Looks good @josemigallas ! I knew you could do better :wink: :trophy: 
I'll still review the code tomorrow, there are some things that I am not quite sure about....

app/javascript/packs/table_toolbar.ts
```diff
@@ -1,8 +1,8 @@
 // TODO: combine with Messages toolbar
```
mayorova, line 1: Looks like this is now combined?...

josemigallas: Right!

spec/javascripts/Common/components/TableToolbar.spec.tsx
```diff
@@ -0,0 +1,32 @@
+import { mount } from 'enzyme'
+
+import { TableToolbar } from 'Common/components/TableToolbar'
+
+import type { Props } from 'Common/components/TableToolbar'
+
+const defaultProps = {
+  totalEntries: 50,
+  leftActions: undefined,
+  rightActions: undefined,
```
mayorova, line 10:
```suggestion
  actions: undefined,
  overflow: undefined,
```

I realize that this is TBD, but just to get rid of confusion on the props names.

app/presenters/buyers/accounts_index_presenter.rb
```diff
@@ -1,24 +1,33 @@
...
+                          .order_by(*sorting_params)
+                          .paginate(pagination_params)
+  end
+
+  def account_plans
+    @account_plans ||= provider.account_plans.stock
+  end
+
+  def account_plans_size
+    @account_plans_size ||= account_plans.length
```
mayorova, line 30: I think `.size` is more common for Ruby, but it's OK, they are doing the same thing for arrays.

mayorova:
So, the outstanding "improvements request" that could go to a different JIRA would be:

1. Bring back the note that helps users to understand where the keywords will be searched.  The original text:
> Search for Group/Org., User login, First name, Last name, email, user_key, app_id or Application Name.

2. Consider whether another filter would be more intuitive: https://v4-archive.patternfly.org/v4/components/toolbar#stacked-example Basically, the difference is that here you see what the values belong to (because when you see just a selected value "Default" you may be confused by what column it belongs to)

josemigallas:
I added this JIRA: [THREESCALE-10374: Improve Toolbars](https://issues.redhat.com/browse/THREESCALE-10374)

mayorova (approved):
Looks good :+1:

lvillen (approved):

PR #3492, opened 2023-08-01
use ThinkingSphinx public API

akostadinov:
### main change

This is upgrading thinking-sphinx and updating `SphinxIndexationWorker` and `SphinxAccountIndexationWorker`.

In the past we used some internal APIs to call the proper thinking-sphinx callbacks for indexing and deinxing stuff. But  in 5.5.0 the `ThinkingSphinx::Processor` was introduced.

Unfortunately not getting access to the index, we also don't get the index `scope`, that's why `SphinxAccountIndexationWorker` hardcodes the searching scope. And the index `scope` is extracted into a model `scope` for easier access (see `account.rb`).

<del>### the error reporting</del>

<del>Improved error reporting of index worker errors. Switched `ErrorReporting` to use `add_metadata` because `add_tab` is deprecated.</del>

<del>https://www.rubydoc.info/gems/bugsnag/6.24.2/Bugsnag/Report#add_metadata-instance_method</del>

<del>Also tried to avoid double-wrapping of parameters into a hash. Need to test this though.</del>

### monkey patching to use proper fields for indices

This is `config/initializers/sphinx.rb`. You can leave that alone, I tested the config generated is same as previously. Or if you like to dig into it, you'd have to understand how previous monkey patching worked and then the pull requests linked in the comment. We can have a call to drive you though this if so much want.

### test suite changes

The main issue was that we didn't have a global way to turn on/off sphinx (previously we could disable the callback but no such API to turn off `Processor`) and it broke a lot of tests.

So the main part for unit/integration/functional tests is `test/test_helpers/sphinx.rb`. It adds methods to stub and unstub the workers as well automatically doing the right thing inside a `real_time_run/rt_run` blocks. Also it adds a test helper to disable sphinx workers before each test. So that we only enable them when we want.

For cucumbers, the main part is `features/support/sphinx.rb`. There hooks are added to disable indexation workers for all tests but these tagged with `@search`.

Then the rest is about removing redundant calls to disable sphinx indexing workers. Occasionally a call to enable index workers where `rt_run` was not used.

### DIFF
--- a/app/workers/sphinx_account_indexation_worker.rb
+++ b/app/workers/sphinx_account_indexation_worker.rb
@@ -8,18 +8,14 @@ require_relative 'sphinx_indexation_worker'
 # - User gets created, updated, deleted
 # - Cinstance gets created, updated, deleted
 class SphinxAccountIndexationWorker < SphinxIndexationWorker
-  def perform(model, id)
-    pk = model.primary_key
-    account = model.find_by(pk => id)
+  def perform(_model, id)
+    account = Account.searchable.find_by(id: id)
 
-    indices_for_model(model).each do |index|
-      if account && index.scope.find_by(pk => id)
-        reindex(index, account)
-      elsif account && !account.master?
-        delete_from_index(index, id, *account.buyers.pluck(:id))
-      else
-        delete_from_index(index, id)
-      end
+    if account
+      reindex(account)
+    else
+      buyers = Account.buyers.where(provider_account_id: id).pluck(:id)
+      delete_from_index(Account, id, *buyers)
     end
   end
 end
--- a/features/support/initial_data.rb
+++ b/features/support/initial_data.rb
@@ -12,11 +12,7 @@ Before do |scenario|
 
   ThreeScale.config.stubs(superdomain: '3scale.localhost')
 
-  SphinxIndexationWorker.stubs(:perform_later)
-
   FieldsDefinition.create_defaults! master_account
-
-  SphinxIndexationWorker.unstub(:perform_later) if scenario.source_tag_names.include?('@search')
 
   ThreeScale.config.stubs(onpremises: false)
   ThreeScale.config.sandbox_proxy.stubs(apicast_registry_url: 'http://apicast.alaska/policies')
--- a/test/integration/master/api/finance/billing_jobs_controller_test.rb
+++ b/test/integration/master/api/finance/billing_jobs_controller_test.rb
@@ -28,8 +28,6 @@ class Master::Api::Finance::BillingJobsControllerTest < ActionDispatch::Integrat
   end
 
   test 'trigger billing to all buyers of a provider' do
-    SphinxIndexationWorker.stubs(:perform_later)
-
     Sidekiq::Testing.inline! do
       FactoryBot.create_list(:buyer_account, 4, provider_account: @provider)
 
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -67,10 +67,6 @@ class ActiveSupport::TestCase
   def assert_cannot(ability, *args)
     assert ability.cannot?(*args), "User can #{args.join(' ')} but should not"
   end
-
-  teardown do
-    User.current = nil
-  end
 end
 
 # Load test helpers
--- a/test/test_helpers/simple_mini_test.rb
+++ b/test/test_helpers/simple_mini_test.rb
@@ -1,8 +1,3 @@
 # frozen_string_literal: true
 
-class SimpleMiniTest < ActiveSupport::TestCase
-  def teardown
-    super
-    User.current = nil
-  end
-end
+class SimpleMiniTest < ActiveSupport::TestCase; end
--- a/test/unit/account/search_test.rb
+++ b/test/unit/account/search_test.rb
@@ -161,6 +161,7 @@ class Account::SearchTest < ActiveSupport::TestCase
       ThinkingSphinx::Test.clear
       ThinkingSphinx::Test.init
       ThinkingSphinx::Test.start index: false
+      ThinkingSphinx::Test.enable_search_jobs!
       perform_enqueued_jobs(only: SphinxAccountIndexationWorker) do
         @provider = FactoryBot.create(:simple_provider)
         @buyers = FactoryBot.create_list(:simple_buyer, 2, provider_account: provider)
--- a/test/unit/indices/indexing_test.rb
+++ b/test/unit/indices/indexing_test.rb
@@ -86,14 +86,6 @@ class IndexingTest < ActiveSupport::TestCase
 
   private
 
-  def indexed_models
-    ThinkingSphinx::Test.indexed_models
-  end
-
-  def indexed_ids(model)
-    model.search(middleware: ThinkingSphinx::Middlewares::IDS_ONLY)
-  end
-
   def factory_for(model)
     overrides = {
       account: :simple_provider,
--- a/test/unit/indices/indexing_test.rb
+++ b/test/unit/indices/indexing_test.rb
@@ -7,6 +7,7 @@ class IndexingTest < ActiveSupport::TestCase
     ThinkingSphinx::Test.clear
     ThinkingSphinx::Test.init
     ThinkingSphinx::Test.start index: false
+    ThinkingSphinx::Test.enable_search_jobs!
   end
 
   teardown do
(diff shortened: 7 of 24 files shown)


test/test_helpers/simple_mini_test.rb
```diff
@@ -1,8 +1,3 @@
 # frozen_string_literal: true
 
-class SimpleMiniTest < ActiveSupport::TestCase
-  def teardown
-    super
-    User.current = nil
-  end
-end
+class SimpleMiniTest < ActiveSupport::TestCase; end
```
jlledom, line 3: Why keep this class anymore?

jlledom (commented):
I don't really understand 90% of this PR. But I tried the Account Indexer and it works.

Could you add a brief explanation in the PR description?

test/unit/account/search_test.rb
```diff
@@ -161,6 +161,7 @@ class IncrementalIndexingTest < ActiveSupport::TestCase
       ThinkingSphinx::Test.clear
       ThinkingSphinx::Test.init
       ThinkingSphinx::Test.start index: false
+      ThinkingSphinx::Test.enable_search_jobs!
```
jlledom, line 164: Don't you need to disable on `teardown`?

test/unit/indices/indexing_test.rb
```diff
@@ -7,6 +7,7 @@ class IndexingTest < ActiveSupport::TestCase
     ThinkingSphinx::Test.clear
     ThinkingSphinx::Test.init
     ThinkingSphinx::Test.start index: false
+    ThinkingSphinx::Test.enable_search_jobs!
```
jlledom, line 10: Same here

test/workers/sphinx_indexation_worker_test.rb
```diff
@@ -6,30 +6,8 @@ class SphinxIndexationWorkerTest < ActiveSupport::TestCase
...
-    worker = SphinxIndexationWorker.new
-    indices = ThinkingSphinx::Test.indexed_models.select do |model|
-      assert_match SizeMatcher.new(1), worker.send(:indices_for_model, model).to_a
-    end
-
-    assert_not_empty indices.map(&:name)
-  end
-
   test 'it does not raises if id does not exist' do
+    enable_search_jobs!
```
jlledom, line 10: Same, don't need to disable them after the test?

test/test_helper.rb
```diff
@@ -67,10 +67,6 @@ def assert_can(ability, *args)
   def assert_cannot(ability, *args)
     assert ability.cannot?(*args), "User can #{args.join(' ')} but should not"
   end
-
-  teardown do
-    User.current = nil
-  end
```
jlledom, line 73: Why was this required before but not now?

features/support/initial_data.rb
```diff
@@ -12,12 +12,8 @@
 
   ThreeScale.config.stubs(superdomain: '3scale.localhost')
 
-  SphinxIndexationWorker.stubs(:perform_later)
```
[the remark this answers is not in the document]
akostadinov: Now disabling and enabling indexation workers is done in `support/sphinx.rb`. I guess these were needed before to avoid these workers being run later and failing. Presently this should be redundant though.

test/integration/master/api/finance/billing_jobs_controller_test.rb
```diff
@@ -28,8 +28,6 @@ class Master::Api::Finance::BillingJobsControllerTest < ActionDispatch::Integrat
   end
 
   test 'trigger billing to all buyers of a provider' do
-    SphinxIndexationWorker.stubs(:perform_later)
-
```
akostadinov, line 32: Here and at the places where `perform_enqueued_jobs(except:` was used, now this is not needed anymore because sphinx workers are globally disabled (stubbed) and enabled only where needed.

test/test_helper.rb
akostadinov: This is redundant. It is also done in `test/test_helpers/user.rb`

test/test_helpers/simple_mini_test.rb
akostadinov: Multiple tests inherit from it, didn't want to clutter the PR even more than it already is. But can easily be removed.

test/unit/account/search_test.rb
akostadinov: In `setup` these are disabled. So it is redundant to disable also in `teardown`. Moreover mocks are automatically removed on teardown anyway. So if we disable in teardown, then they will be automatically enabled by the mocha `teardown`.

test/unit/indices/indexing_test.rb
akostadinov: same answer :)

test/workers/sphinx_indexation_worker_test.rb
akostadinov: Enabling and disabling should be automatic in most cases. This is of the rare cases that we need to alter it manually. It will be disabled again in next test case `setup`. And after `tierdown` actually the expected state is enabled :) And because we use mocha, it unstubs in teardown.

akostadinov:
@jlledom , now should be ready to review, I will only need to add a couple of tests (hope oracle passes without issues). I didn't realize you reviewed it so I was all squashing. But will add further changes in new commits.

test/workers/sphinx_indexation_worker_test.rb
akostadinov: If I stub `perform_later` as in the past though, this line will not be needed here. WDYT? Should we stub `#perform` or `::perform_later`?

test/unit/account/search_test.rb
jlledom: which `setup` do you mean?

test/workers/sphinx_indexation_worker_test.rb
jlledom: Honestly, no idea.

akostadinov: I think this is fine. Disable means disable, not disable only background. Shouldn't make a big difference.

test/unit/account/search_test.rb
akostadinov: https://github.com/3scale/porta/pull/3492/files#diff-33d4a34306c39c3dc095690ac70832e3a35f67a38cc820ceeae333006b330513R55

app/workers/sphinx_account_indexation_worker.rb
```diff
@@ -8,18 +8,14 @@
...
 
-    indices_for_model(model).each do |index|
-      if account && index.scope.find_by(pk => id)
-        reindex(index, account)
-      elsif account && !account.master?
-        delete_from_index(index, id, *account.buyers.pluck(:id))
-      else
-        delete_from_index(index, id)
-      end
+    if account
```
jlledom, line 14: Why can't you access the index? It seems I have access. When I place a breakpoint in this line:

```
    14:   binding.pry
 => 15:   if account
    16:     reindex(account)
    17:   else
```

```rb
[2] pry(#<SphinxAccountIndexationWorker>)> idxs = ThinkingSphinx::Configuration.instance.index_set_class.new(classes: [_model])
=> #<ThinkingSphinx::IndexSet:0x000000000e141e88
 @configuration=
  #<ThinkingSphinx::Configuration:0x00007fc0cc722b78
...
```

```rb
[4] pry(#<SphinxAccountIndexationWorker>)> idxs.count
=> 1
[5] pry(#<SphinxAccountIndexationWorker>)> idxs.first
=> #<ThinkingSphinx::RealTime::Index:0x00007fc0cc7353b8
 @attributes=
  [#<ThinkingSphinx::RealTime::Attribute:0x00007fc0cc734a30
    @column=
     #<ThinkingSphinx::ActiveRecord::Column:0x00007fc0cc734ad0
...
```

```rb
[6] pry(#<SphinxAccountIndexationWorker>)> idxs.first.scope
=> [#<Account:0x00007fc0cc6ff498
  id: 2,
  org_name: "Provider Name",
  org_legaladdress: "",
  created_at: Thu, 03 Aug 2023 15:28:30 UTC +00:00,
  updated_at: Thu, 03 Aug 2023 15:28:30 UTC +00:00,
...
```

```rb
[7] pry(#<SphinxAccountIndexationWorker>)> idxs.first.scope.size
=> 2
```

Is this not the same you see? Or is there another problem?

akostadinov: From old version, what I see is that probably no index was found. Or no callbacks were available. I don't know why in some situations this happened because I couldn't locally or even in place reproduce with another image. So trying to be using supported APIs.

jlledom (approved):

PR #3494, opened 2023-08-01
⬆️ Bump React to 17.0.2

lvillen:
[THREESCALE-10015: Upgrade React to 17.0.2](https://issues.redhat.com/browse/THREESCALE-10015)

Quickstart guides require React 17, and, furthermore, a potential upgrade to PF5 will also require React 17. Here there's an upgrade for React and different dependencies it requires.


josemigallas (approved):

spec/javascripts/utilities/useClickOutside.spec.tsx
```diff
@@ -12,15 +12,15 @@ const HookedComponent: FunctionComponent<{ callback: jest.Mock }> = ({ callback
...
+it('should add and remove a mousedown event listener', () => {
   const addEventListener = jest.fn()
   const removeEventListener = jest.fn()
 
   document.addEventListener = addEventListener
   document.removeEventListener = removeEventListener
 
   const wrapper = mount(<HookedComponent callback={jest.fn()} />)
-  expect(addEventListener).toHaveBeenCalledTimes(1)
+  expect(addEventListener).toHaveBeenCalledWith('mousedown', expect.any(Function))
```
jlledom, line 23: Not checking the number of calls?

lvillen: The important thing is to check the `mousedown` event has been registered, since there are other calls which might not be related with the useClickOutside hook. Hence the change.

jlledom: But it's useful to check the mouse event has been registered once, right?

lvillen: There's not an specific way of filtering those calls, and though we've checked it manually, we consider it isn't worth the fuzz.

PR #3287, opened 2023-03-29
THREESCALE-8131 show inline errors on product create

nidhi-soni1104:
**What this PR does / why we need it:**

When we create new product it doesn't show inline error message as backend page

The inline error in New Product page should like like the one in New Backend page. 
New Backend is correct, New Product is not. So fixed that

**Steps to reproduce**
Go to Products
Create new product 
try creating product with name already used or not in the expected format which is mentioned in the form

Bakend page is a rails page and New prodcut is typescript. 

**Which issue(s) this PR fixes**
https://issues.redhat.com/browse/THREESCALE-8131

This is how a new product form looks
image

For reference

image

image

PR #3495, opened 2023-08-02
THREESCALE-8131 fix misleading error

nidhi-soni1104:

Updated Flash errors while creating  new product
updated screenshots is added in JIRA comments

**Steps to reproduce**
Go to Products
Create new product
try creating product with name already used or not in the expected format which is mentioned in the form

Bakend page is a rails page and New prodcut is typescript.

https://issues.redhat.com/browse/THREESCALE-8131

### DIFF
--- a/features/old/services/multiservice.feature
+++ b/features/old/services/multiservice.feature
@@ -42,8 +42,8 @@ Feature: Multiservice feature
     And I fill in "Name" with "Less fancy API"
     And I fill in "System name" with "SystemName@123"
     And I press "Create Product"
-    Then I should see "Invalid."
-    Then I should see the flash message "System name Invalid."
+    Then I should see "invalid"
+    And I should see the flash message "System name invalid"
 
   @javascript
   Scenario: Create new product: with already existing System name
@@ -55,6 +55,7 @@ Feature: Multiservice feature
     And I fill in "System name" with "api"
     And I press "Create Product"
     Then I should see "Has already been taken"
+    And I should see the flash message "System name has already been taken"
 
   @wip
   Scenario: Create new backend
--- a/features/old/services/multiservice.feature
+++ b/features/old/services/multiservice.feature
@@ -43,7 +43,7 @@ Feature: Multiservice feature
     And I fill in "System name" with "SystemName@123"
     And I press "Create Product"
     Then I should see "Invalid."
-    Then I should see the flash message "Couldn't create Product. Check your Plan limits"
+    Then I should see the flash message "System name Invalid."
 
   @javascript
   Scenario: Create new product: with already existing System name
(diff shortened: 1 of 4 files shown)


mayorova:
Good, but the tests need to be updated also:

https://github.com/3scale/porta/pull/3287/files#diff-f2bb6ae2aeee8f3af29ad9ec6fc37b3eee2050917980e86d44d611491c186ddeR57

Also, here the check for the flash message was removed, it needs to be brought back.

nidhi-soni1104:
@mayorova Test updated. 

the line you pointed is not necessarily need to be updated because we are showing this as inline  message when error come so that will also work

here is screenshot for reference

image

mayorova:
I think this block needs to be removed:
https://github.com/3scale/porta/blob/THREESCALE-8131-inline-errors/config/locales/en.yml#L898-L902

The text for `invalid` error is already "invalid", but not the lowercase. So, it gets capitalized when needed, and remains lowercase when needed (e.g. in the flash message), compare:
Screenshot from 2023-08-07 13-23-52

Screenshot from 2023-08-07 13-25-43

On the 1st screenshot it says "System name Invalid" which looks weird.
There is another issue with separation of the error message and the hint, but I think it needs to be handled with CSS (probably in another PR).

features/old/services/multiservice.feature
```diff
@@ -43,7 +43,7 @@ Feature: Multiservice feature
     And I fill in "System name" with "SystemName@123"
     And I press "Create Product"
     Then I should see "Invalid."
-    Then I should see the flash message "Couldn't create Product. Check your Plan limits"
+    Then I should see the flash message "System name Invalid."
```
mayorova, line 46: After removing the `Invalid.` translation, this will need to be changed to "System name invalid"

mayorova:
```suggestion
    And I should see the flash message "System name invalid"
```

test/integration/api/services_controller_test.rb
```diff
@@ -319,10 +319,10 @@ class ServiceCreateTest < self
       @provider.settings.allow_multiple_services!
 
       post admin_services_path, params: { service: { name: '' } }
-      assert_equal 'Couldn\'t create Product. Check your Plan limits', flash[:error]
+      assert_equal 'Name can\'t be blank', flash[:error]
 
       post admin_services_path, params: { service: { name: 'example-service', system_name: '###' } }
-      assert_equal 'Couldn\'t create Product. Check your Plan limits', flash[:error]
+      assert_equal 'System name Invalid.', flash[:error]
```
mayorova, line 325:
```suggestion
      assert_equal 'System name invalid', flash[:error]
```

features/old/services/multiservice.feature
mayorova: And the line `:45` should also change: `Then I should see "Invalid"`

features/old/services/multiservice.feature
```diff
@@ -43,7 +43,7 @@ Feature: Multiservice feature
...
 
   @javascript
   Scenario: Create new product: with already existing System name
```
mayorova, line 49: It would be nice to add
```
And I should see the flash message "System name has already been taken"
```

as the last step of this scenario.

nidhi-soni1104:
@mayorova PR is completed

mayorova (approved):

PR #3504, opened 2023-08-08
🦋 updates ActiveDocs empty state and alerts

josemigallas:
[THREESCALE-9773: Unattached specs info box in ActiveDocs](https://issues.redhat.com/browse/THREESCALE-9773)

Unattached specs info box is now an alert:

Screenshot 2023-08-08 at 17 49 59

And Swagger upgrade box is now an alert:

Screenshot 2023-08-08 at 18 03 26

### DIFF
--- a/app/views/admin/api_docs/base/index.html.slim
+++ b/app/views/admin/api_docs/base/index.html.slim
@@ -1,27 +1,12 @@
 - content_for :page_header_title, 'ActiveDocs'
 
-- if @api_docs_services.any? { |service| service.needs_swagger_update? }
-  - content_for :page_header_alert do
-    br
-    div class="pf-c-alert pf-m-info pf-m-inline" aria-label="Success alert"
-      div class="pf-c-alert__icon"
-        i class="fas fa-fw fa-info-circle" aria-hidden="true"
-      p class="pf-c-alert__title"
-        span class="pf-screen-reader" Info alert:
-        | ActiveDocs now supports Swagger 2.0
-      div class="pf-c-alert__description"
-        p
-          ' Read
-          = link_to 'the documentation', t("api.integrations.proxy.documentation_create_active_doc_spec", docs_base_url: docs_base_url)
-          '  to know why it makes sense to upgrade your service specs to Swagger 2.0
-
 - if @api_docs_services.any?
   = render partial: 'active_docs_table', locals: {service: @service, api_docs_services: @api_docs_services}
 - else
   - if current_account.api_docs_services.without_service.any? && can?(:manage, :portal)
     - content_for :page_header_alert do
       br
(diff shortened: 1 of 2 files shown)


app/views/admin/api_docs/base/index.html.slim
```diff
@@ -1,27 +1,44 @@
 - content_for :page_header_title, 'ActiveDocs'
 
 - if @api_docs_services.any? { |service| service.needs_swagger_update? }
-  = link_to t("api.integrations.proxy.documentation_create_active_doc_spec", docs_base_url: docs_base_url), class: 'SwaggerUpgradeBox' do
-    p.SwaggerUpgradeBox-message
-      span.SwaggerUpgradeBox-logo
-      | ActiveDocs now supports Swagger 2.0. Read here why it makes sense to upgrade your service specs to Swagger 2.0
+  - content_for :page_header_alert do
+    br
+    div class="pf-c-alert pf-m-info pf-m-inline" aria-label="Success alert"
```
mayorova, line 6: Why "Success alert" ? (here and below also)

app/views/admin/api_docs/base/index.html.slim
```diff
@@ -1,27 +1,44 @@
...
+      div class="pf-c-alert__icon"
+        i class="fas fa-fw fa-info-circle" aria-hidden="true"
+      p class="pf-c-alert__title"
+        span class="pf-screen-reader" Info alert:
+        | ActiveDocs now supports Swagger 2.0
+      div class="pf-c-alert__description"
+        p
+          ' Read
+          = link_to 'the documentation', t("api.integrations.proxy.documentation_create_active_doc_spec", docs_base_url: docs_base_url)
```
mayorova, line 15: 1. This link seems to be broken

2. Even if we update it to something that seems relevant, like [Update To ActiveDocs 2.0](https://access.redhat.com/documentation/en-us/red_hat_3scale/2-saas/html/providing_apis_in_the_developer_portal/activedocs-migrate), the content also seems outdated :upside_down_face: And indeed, since a long time ago we also support OpenAPI v3, so I think this message is quite outdated.

3. Also, while we are at it, maybe we can update the link referenced in "Learn how to upgrade to Swagger 2" (inside the table)

So, I would vote for removing the alert for this message.

app/views/admin/api_docs/base/index.html.slim
josemigallas: Typo

app/views/admin/api_docs/base/index.html.slim
josemigallas: Yes, it's probably super outdated and I also vote to remove it altogether.

josemigallas: Changed my mind again:

mayorova (approved):

PR #3511, opened 2023-08-11
fix double billing of providers

akostadinov:
The main fix is adding a scope to the `buyer_accounts` association that prevents returning self as a buyer. The changes in `BillingWorker::enqueue_for_buyer`, `BillingService` and `BillingStrategy` are extra safety layers.

<del>The main fix is in `BillingWorker::enqueue_for_buyer`. The changes in
`BillingService` and `BillingStrategy` are extra safety layers.</del>

And of course some tests.

### DIFF
--- a/app/models/account/provider_methods.rb
+++ b/app/models/account/provider_methods.rb
@@ -68,7 +68,7 @@ module Account::ProviderMethods
       end
     end
 
-    has_many :buyer_accounts, ->(this_account) { where.not(id: this_account.id) }, class_name: 'Account', foreign_key: 'provider_account_id', dependent: :destroy, inverse_of: :provider_account do
+    has_many :buyer_accounts, ->(provider) { where.not(id: provider.id) }, class_name: 'Account', foreign_key: 'provider_account_id', dependent: :destroy, inverse_of: :provider_account do
       def latest
         order(created_at: :desc).includes([:admin_users]).limit(5)
       end
--- a/app/models/account/provider_methods.rb
+++ b/app/models/account/provider_methods.rb
@@ -68,7 +68,7 @@ module Account::ProviderMethods
       end
     end
 
-    has_many :buyer_accounts, class_name: 'Account', foreign_key: 'provider_account_id', dependent: :destroy, inverse_of: :provider_account do
+    has_many :buyer_accounts, -> { where(Account.arel_table[:id].not_eq(Account.arel_table[:provider_account_id])) }, class_name: 'Account', foreign_key: 'provider_account_id', dependent: :destroy, inverse_of: :provider_account do
       def latest
         order(created_at: :desc).includes([:admin_users]).limit(5)
       end
--- a/app/models/account/provider_methods.rb
+++ b/app/models/account/provider_methods.rb
@@ -68,7 +68,7 @@ module Account::ProviderMethods
       end
     end
 
-    has_many :buyer_accounts, -> { where(Account.arel_table[:id].not_eq(Account.arel_table[:provider_account_id])) }, class_name: 'Account', foreign_key: 'provider_account_id', dependent: :destroy, inverse_of: :provider_account do
+    has_many :buyer_accounts, ->(this_account) { where.not(id: this_account.id) }, class_name: 'Account', foreign_key: 'provider_account_id', dependent: :destroy, inverse_of: :provider_account do
       def latest
         order(created_at: :desc).includes([:admin_users]).limit(5)
       end
--- a/app/services/finance/billing_service.rb
+++ b/app/services/finance/billing_service.rb
@@ -2,8 +2,6 @@
 
 module Finance
   class BillingService
-    class LockBillingError < StandardError; end
-
     def self.async_call(provider_account, now = Time.zone.now, buyers_scope = nil)
       BillingWorker.enqueue(provider_account, now, buyers_scope)
     end
@@ -30,13 +28,13 @@ module Finance
 
     def call!
       with_lock { call }
-    rescue LockBillingError => error
+    rescue LockBillingError, SpuriousBillingError => error
       report_error(error)
       nil
     end
 
     def call
-      Finance::BillingStrategy.daily(billing_options)
+      Finance::BillingStrategy.daily(validate_options(billing_options))
     end
 
     def notify_billing_finished(_status, billing_results)
@@ -64,9 +62,16 @@ module Finance
       options
     end
 
+    def validate_options(billing_options)
+      unless billing_options[:only].presence&.size == 1 && billing_options[:buyer_ids].presence&.size == 1
+        raise SpuriousBillingError, "Expected to bill individual buyers separately but got providers: #{billing_options[:only].inspect}, buyers: #{billing_options[:buyer_ids].inspect}"
+      end
+      billing_options
+    end
+
     def with_lock
-      # intentionally skip unlocking, no further billing of account within 3 hours allowed
-      raise LockBillingError, "Concurrent billing job already running for account #{account_id}" unless Synchronization::NowaitLockService.call("billing:#{account_id}", timeout: 1.hours.to_i * 1000).result
+      # intentionally skip unlocking, no further billing of account within 1 hour allowed
+      raise LockBillingError, "Concurrent billing job already running for account #{account_id}" unless Synchronization::NowaitLockService.call("billing:#{account_id}", timeout: 1.hour.in_milliseconds).result
 
       yield
     end
--- a/app/services/finance/billing_service.rb
+++ b/app/services/finance/billing_service.rb
@@ -57,9 +57,9 @@ module Finance
     private
 
     def billing_options
-      options = { only: [provider_account_id], now: now, skip_notifications: skip_notifications }
-      raise SpuriousBillingError, "whole provider will be billed when buyer id is the same: #{account_id}" if provider_account_id == account_id
+      raise SpuriousBillingError, "provider and buyer ids are the same: #{account_id}" if provider_account_id == account_id
 
+      options = { only: [provider_account_id], now: now, skip_notifications: skip_notifications }
       options[:buyer_ids] = [account_id]
       options
     end
--- a/app/services/finance/billing_service.rb
+++ b/app/services/finance/billing_service.rb
@@ -34,7 +34,7 @@ module Finance
     end
 
     def call
-      Finance::BillingStrategy.daily(validate_options(billing_options))
+      Finance::BillingStrategy.daily(billing_options)
     end
 
     def notify_billing_finished(_status, billing_results)
@@ -58,15 +58,10 @@ module Finance
 
     def billing_options
       options = { only: [provider_account_id], now: now, skip_notifications: skip_notifications }
-      options[:buyer_ids] = [account_id] if provider_account_id != account_id
-      options
-    end
+      raise SpuriousBillingError, "whole provider will be billed when buyer id is the same: #{account_id}" if provider_account_id == account_id
 
-    def validate_options(billing_options)
-      unless billing_options[:only].presence&.size == 1 && billing_options[:buyer_ids].presence&.size == 1
-        raise SpuriousBillingError, "Expected to bill individual buyers separately but got providers: #{billing_options[:only].inspect}, buyers: #{billing_options[:buyer_ids].inspect}"
-      end
-      billing_options
+      options[:buyer_ids] = [account_id]
+      options
     end
 
     def with_lock
--- a/app/services/finance/spurious_billing_error.rb
+++ b/app/services/finance/spurious_billing_error.rb
@@ -1,0 +1,5 @@
+# frozen_string_literal: true
+
+module Finance
+  class SpuriousBillingError < BillingError; end
+end
--- a/app/workers/billing_worker.rb
+++ b/app/workers/billing_worker.rb
@@ -45,6 +45,8 @@ class BillingWorker
   end
 
   def self.enqueue_for_buyer(buyer, billing_date)
+    return if buyer.id == buyer.provider_account_id # this should filter out master and any broken accounts
+
     time = billing_date.to_s(:iso8601)
     perform_async(buyer.id, buyer.provider_account_id, time)
   end
--- a/app/workers/billing_worker.rb
+++ b/app/workers/billing_worker.rb
@@ -45,7 +45,12 @@ class BillingWorker
   end
 
   def self.enqueue_for_buyer(buyer, billing_date)
-    return if buyer.id == buyer.provider_account_id # this should filter out master and any broken accounts
+    if buyer.id == buyer.provider_account_id
+      System::ErrorReporting.report_error(ArgumentError.new("invalid buyer #{buyer.id}, has self as provider"),
+                                          buyer_id: buyer.id,
+                                          billing_date: billing_date)
+      return
+    end
 
     time = billing_date.to_s(:iso8601)
     perform_async(buyer.id, buyer.provider_account_id, time)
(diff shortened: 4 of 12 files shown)


app/models/account/provider_methods.rb
```diff
@@ -68,7 +68,7 @@ def build_kind(kind:, **attributes)
       end
     end
 
-    has_many :buyer_accounts, class_name: 'Account', foreign_key: 'provider_account_id', dependent: :destroy, inverse_of: :provider_account do
+    has_many :buyer_accounts, -> { where(Account.arel_table[:id].not_eq(Account.arel_table[:provider_account_id])) }, class_name: 'Account', foreign_key: 'provider_account_id', dependent: :destroy, inverse_of: :provider_account do
```
mayorova, line 71:
```suggestion
    has_many :buyer_accounts, ->(provider) { where.not(id: provider.id) }, class_name: 'Account', foreign_key: 'provider_account_id', dependent: :destroy, inverse_of: :provider_account do
```

mayorova: How about this?

I think it's more concise and more readable/understandable. And produces the same result.

I actually thought that it might be more efficient, because it search in the DB using indexes and primary keys directly, while the `id <> provider_account_id` would actually compare the values for each record, but the DB is smarter and results in the very similar query (from what I can tell).

app/services/finance/billing_service.rb
```diff
@@ -30,13 +28,13 @@ def initialize(account_id, options = {})
...
       with_lock { call }
-    rescue LockBillingError => error
+    rescue LockBillingError, SpuriousBillingError => error
       report_error(error)
       nil
     end
 
     def call
-      Finance::BillingStrategy.daily(billing_options)
+      Finance::BillingStrategy.daily(validate_options(billing_options))
```
mayorova, line 37: I don't know, to me it seems unnecessary to "wrap" it, I would just throw within `def billing_options` directly

app/services/finance/spurious_billing_error.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+module Finance
+  class SpuriousBillingError < BillingError; end
```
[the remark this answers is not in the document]
akostadinov: image

I think the third one match well.

app/models/account/provider_methods.rb
akostadinov: In the Arel version the query is:
```
[34] pry(main)> account.buyer_accounts.count
   (2.3ms)  SELECT COUNT(*) FROM `accounts` WHERE `accounts`.`provider_account_id` = 2 AND `accounts`.`id` != `accounts`.`provider_account_id`
```

In your suggested version the query is:
```
[2] pry(main)> account.buyer_accounts.count
   (1.8ms)  SELECT COUNT(*) FROM `accounts` WHERE `accounts`.`provider_account_id` = 2 AND `accounts`.`id` != 2
```

Searching by primary index is not possible nor advisable because the huge work is getting the buyer ids by the index on `provider_account_id`. Then the additional filtering cannot use an index. Either case, I assume the engine is smart enough to treat the same negation and "not equal" the same.

But comparing a constant should apparently be slightly better than comparing to another field.

It is very different semantics though - filtering out self vs filtering out id/provider_id matches. I like your suggstion much better but it initially confused me of why it worked. So I accept change but rename `provider` to `this_account`. 

Does it look good to you that way?

app/services/finance/spurious_billing_error.rb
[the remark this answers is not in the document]
jlledom: What about `SelfProviderBillingError`?

app/workers/billing_worker.rb
```diff
@@ -45,6 +45,13 @@ def self.enqueue(provider, billing_date, buyers_scope = nil)
   end
 
   def self.enqueue_for_buyer(buyer, billing_date)
+    if buyer.id == buyer.provider_account_id
+      System::ErrorReporting.report_error(ArgumentError.new("invalid buyer #{buyer.id}, has self as provider"),
```
jlledom, line 49: Why not `SpuriousBillingError` here?

test/unit/account_test.rb
```diff
@@ -72,11 +72,9 @@ def setup
...
     buyer = FactoryBot.create(:simple_buyer, provider_account: master)
 
-    assert_includes master.buyer_account_ids, buyer.id
-    assert_includes master.buyer_account_ids, master.id
-
-    assert_includes master.buyer_accounts.not_master.ids, buyer.id
-    assert_not_includes master.buyer_accounts.not_master.ids, master.id
+    assert_includes Account.all.ids, master.id
+    assert_includes Account.all.not_master.ids, buyer.id
+    assert_not_includes Account.all.not_master.ids, master.id
```
jlledom, line 77: I understand the old `assert_includes master.buyer_account_ids, master.id` isn't valid anymore, but why replacing `buyer_accounts` by `Account.all` in the other assertions? It seems to me we should test `buyer_accounts` since it's precisely what changed.

test/unit/services/finance/billing_service_test.rb
```diff
@@ -28,23 +28,23 @@ class Finance::BillingServiceTest < ActionDispatch::IntegrationTest
     end
   end
 
-  test 'triggers billing' do
+  test 'does not trigger billing without a buyer' do
     Sidekiq::Testing.inline! do
       now = Time.utc(2018, 1, 16, 8)
       billing_options = { only: [@provider.id], now: now, skip_notifications: true }
```
jlledom, line 34: This is unused now.

test/unit/services/finance/billing_service_test.rb
```diff
@@ -80,6 +80,35 @@ class Finance::BillingServiceTest < ActionDispatch::IntegrationTest
...
+      [nil], # neither provider nor a buyer specified
+    ]
+
+    opts.each do |args|
+      bs = Finance::BillingService.new(*args)
+      assert_raise(Finance::SpuriousBillingError) do
+        bs.call
+      end
+    end
+  end
```
jlledom, line 96: I think it's better to split this one into three, so if one of them fails, we can see it easily in Circle CI
```suggestion
  opts = [
    [5], # account == provider which means to bill a whole provider
    [6, { provider_account_id: 6 }], # again account == provider
    [nil], # neither provider nor a buyer specified
  ]

  opts.each do |args|
    test 'raises unless a single buyer is billed and charged with args: #{args}' do
      bs = Finance::BillingService.new(*args)
      assert_raise(Finance::SpuriousBillingError) do
        bs.call
      end
    end
  end
```

test/unit/services/finance/billing_service_test.rb
```diff
@@ -80,6 +80,35 @@ class Finance::BillingServiceTest < ActionDispatch::IntegrationTest
...
+  test 'reports error unless a single buyer is billed and charged' do
+    opts = [
+      [5], # account == provider which means to bill a whole provider
+      [6, { provider_account_id: 6 }], # again account == provider
+      [nil], # neither provider nor a buyer specified
+    ]
+
+    Finance::BillingService.any_instance.expects(:report_error).with { |error| error.is_a? Finance::SpuriousBillingError }.times(opts.size)
+
+    opts.each do |args|
```
jlledom, line 107: Same as above

app/models/account/provider_methods.rb
```diff
@@ -68,7 +68,7 @@ def build_kind(kind:, **attributes)
       end
     end
 
-    has_many :buyer_accounts, class_name: 'Account', foreign_key: 'provider_account_id', dependent: :destroy, inverse_of: :provider_account do
+    has_many :buyer_accounts, ->(this_account) { where.not(id: this_account.id) }, class_name: 'Account', foreign_key: 'provider_account_id', dependent: :destroy, inverse_of: :provider_account do
```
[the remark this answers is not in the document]
mayorova: It is caused by the fact that the master account is considered its own provider :upside_down_face: 

You didn't find anything in staging, because in staging/production master's account ID is `6`.
So if you search by ID 6 you will surely find it, the master account's `id` and `provider_account_id` are both equal to `6`.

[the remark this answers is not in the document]
jlledom: Yes, but my query filters by the `provider` column, not `id`. The query doesn't return master because its `provider` column is `0`.

How was this caused by master being considered its own provider? Could you briefly explain, please?

test/unit/services/finance/billing_service_test.rb
```diff
@@ -28,23 +28,23 @@ class Finance::BillingServiceTest < ActionDispatch::IntegrationTest
...
+      Finance::BillingService.call!(@provider.id, now: now, skip_notifications: true)
     end
   end
 
   test 'lock remains if execution fails' do
     now = Time.utc(2018, 1, 16, 8)
     Finance::BillingStrategy.expects(:daily).raises RuntimeError, 'random failure'
     assert_raises RuntimeError do
-      Finance::BillingService.call!(@provider.id, now: now)
+      Finance::BillingService.call!(@provider.id, now: now, provider_account_id: master_account.id)
```
mayorova, line 44:
```suggestion
      Finance::BillingService.call!(@provider.id, { now: now, provider_account_id: master_account.id })
```

In preparation for Ruby 2.7 / Ruby 3 upgrade

test/unit/services/finance/billing_service_test.rb
```diff
@@ -28,23 +28,23 @@ class Finance::BillingServiceTest < ActionDispatch::IntegrationTest
...
     end
-    Finance::BillingService.any_instance.expects(:report_error).with { |error| error.is_a? Finance::BillingService::LockBillingError }
-    Finance::BillingService.call!(@provider.id, now: now)
+    Finance::BillingService.any_instance.expects(:report_error).with { |error| error.is_a? Finance::LockBillingError }
+    Finance::BillingService.call!(@provider.id, now: now, provider_account_id: master_account.id)
```
mayorova, line 47:
```suggestion
    Finance::BillingService.call!(@provider.id, { now: now, provider_account_id: master_account.id })
```

test/workers/billing_worker_test.rb
```diff
@@ -64,4 +64,21 @@ class BillingWorkerTest < ActiveSupport::TestCase
...
+
+    System::ErrorReporting.expects(:report_error).with { |*args| args[0].is_a?(ArgumentError) }
+    BillingWorker.enqueue_for_buyer(master_account, Time.zone.now)
+    assert_empty BillingWorker.jobs
+
+    BillingWorker.enqueue(master_account, Time.zone.now)
+    assert_equal 1, BillingWorker.jobs.size
+
+    provider_id = BillingWorker.jobs.first["args"][1]
+    buyer_id = BillingWorker.jobs.first["args"][0]
```
mayorova, line 79:
```suggestion
    buyer_id, provider_id = BillingWorker.jobs.first["args"]
```

app/models/account/provider_methods.rb
akostadinov: The logic behind `this_account` is that rails seems to pass current object to the scope lambda. So we filter out buyers that have an `id` same as current object id. I'm not sure anymore what makes more sense. I'd rather update it to `_1` numbered parameter once on Ruby 2.7. WDYT?

app/services/finance/spurious_billing_error.rb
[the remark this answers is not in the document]
akostadinov: Doesn't speak to me :) Could also be `UnexpectedBillingError` although it doesn't sound cool to me :sunglasses:

app/models/account/provider_methods.rb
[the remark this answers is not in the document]
akostadinov: @jlledom , run:
```
select * from accounts where provider_account_id = id
```

master account doesn't seem to be marked as a provider

app/models/account/provider_methods.rb
mayorova: Well, yeah, sorry, not a full table scan definitely.... :facepalm:  
Just got confused for a bit.

app/models/account/provider_methods.rb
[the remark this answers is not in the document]
akostadinov: This is where the scope is defined:
https://github.com/3scale/porta/blob/master/app/models/finance/billing_strategy.rb#L46

It doesn't seem to filter based on buyer/provider columns. Maybe that's an inefficiency.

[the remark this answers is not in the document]
akostadinov: The buyers are filtered out later in `BillingWorker::enqueue` https://github.com/3scale/porta/blob/master/app/workers/billing_worker.rb#L39

But yeah, it may make sense to filter them out earlier if we know that buyer accounts will not have buyers of their own anyway.

app/workers/billing_worker.rb
akostadinov: My line of thinking was that for `::enqueue_for_buyer` method it is an invalid argument to supply not a buyer. Also Also will differentiate better where we get the error from if that matters.

test/unit/account_test.rb
akostadinov: I believe the test intention is to test the `not master` scope. To be a good test, the originally used scope should include the master account and once `not_master` is applied, then it should not include master account anymore.

In that sense `buyer_account_ids` is not a good scope to test `not_master` with. Where we test `buyer_accounts` in another test case specifically for it.

test/unit/services/finance/billing_service_test.rb
akostadinov: I know this. But to me it's more clear to read in this way. Once we run locally, I expect it to be easy to reproduce in case ever this fails.

I'm very disappointed by minitest's lack of `assert_all` and perhaps general development of the project. You can see my discussion here: minitest/minitest#948

I even have the idea to try make rails switch to test-unit or at least optionally when I have time. Although that might also be never.

test/unit/services/finance/billing_service_test.rb
akostadinov:
```suggestion
```

app/models/account/provider_methods.rb
jlledom: @akostadinov What do you mean by `_1` for Ruby 2.7? I don't understand that part

app/services/finance/spurious_billing_error.rb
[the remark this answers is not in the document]
jlledom: If it's not cool, it's against 3scale values.

app/services/finance/billing_service.rb
```diff
@@ -60,13 +58,15 @@ def provider_account
 
     def billing_options
       options = { only: [provider_account_id], now: now, skip_notifications: skip_notifications }
-      options[:buyer_ids] = [account_id] if provider_account_id != account_id
+      raise SpuriousBillingError, "whole provider will be billed when buyer id is the same: #{account_id}" if provider_account_id == account_id
```
akostadinov, line 61: > provider and buyer ids are the same

app/models/account/provider_methods.rb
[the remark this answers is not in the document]
akostadinov: > provider

mayorova:
Just wanted to point out that we have this: https://github.com/3scale/porta/blob/master/app/controllers/master/api/finance/billing_jobs_controller.rb#L14

It's an API endpoint that can trigger billing for a single provider.
But I think this implementation is not breaking it.

mayorova:
Looks good to me! :+1:

app/models/account/provider_methods.rb
jlledom: Better don't use that `_1` pattern IMO. It's harder to understand than `provider`.

akostadinov: I think it is very useful when there is one thing that is passed by convention or by usage and especially when a good short term is unavailable to describe it. But it's easy not to touch after merge, so your wish will be granted :)

jlledom (dismissed):

jlledom (dismissed):

PR #3519, opened 2023-08-16
Filter Policy chain

josemigallas:
[THREESCALE-10060: Filter chain policy](https://issues.redhat.com/browse/THREESCALE-10060)

Before:
Screenshot 2023-08-16 at 15 02 43

After:
Screenshot 2023-08-16 at 17 10 45


jlledom (dismissed):

app/javascript/src/Policies/components/PolicyChain.tsx
```diff
@@ -21,77 +34,90 @@ interface Props {
...
-        editPolicy={actions.editPolicy}
-        helperClass="Policy--sortable"
-        items={chain}
-        onSortEnd={onSortEnd}
-      />
+    <section>
+      <Toolbar>
+        <ToolbarContent>
+          <ToolbarItem variant="search-filter">
+            <SearchInput aria-label="Items example search input" onChange={handleOnSearch} />
```
lvillen, line 81: Maybe we could add here a `placeholder` and an `onClear`?

lvillen (approved):

mayorova:
There is a weird glitch related to dragging when search filter is not empty.
While the "drag button" is disabled, you can actually start dragging, and it remains in this dragging state, even if you delete the search query.
So, you need to actually drag something back and forth to reset this.

Maybe it can be improved somehow?

josemigallas:
> There is a weird glitch related to dragging when search filter is not empty. While the "drag button" is disabled, you can actually start dragging, and it remains in this dragging state, even if you delete the search query. So, you need to actually drag something back and forth to reset this.
> 
> Maybe it can be improved somehow?
> 
> Peek 2023-09-18 13-15

Good catch @mayorova ... It looks like a Patternfly bug. The way I see it there are 2 options here:
-  Leave with it.
- Allow dragging but dropping fails (the item returns to its original place)

mayorova:
> Allow dragging but dropping fails (the item returns to its original place)

Sounds reasonable

jlledom (dismissed):

mayorova (dismissed):
> Allow dragging but dropping fails (the item returns to its original place)

I find this behavior more user-friendly.

We'll see - if it gets confusing for customers, maybe we can add an information banner whenever search is enabled (e.g. "Clear the filter to reorder policies")

PR #3528, opened 2023-08-23
THREESCALE-9752: Set :exception as the default forgery protection strategy

jlledom:

Check the description and comments at [THREESCALE-9752](https://issues.redhat.com/browse/THREESCALE-9752)

[THREESCALE-9752](https://issues.redhat.com/browse/THREESCALE-9752)

**Verification steps** 

See the issue description to reproduce the bug.

### DIFF
--- a/app/lib/action_controller/request_forgery_protection/exception_and_reset_strategy.rb
+++ b/app/lib/action_controller/request_forgery_protection/exception_and_reset_strategy.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+module ActionController
+  module RequestForgeryProtection
+    class ExceptionAndResetStrategy
+      def initialize(controller)
+        @controller = controller
+      end
+
+      def handle_unverified_request
+        @controller.reset_session
+        exception.handle_unverified_request
+      end
+
+      private
+
+      def exception
+        @exception || ActionController::RequestForgeryProtection::ProtectionMethods::Exception.new(@controller)
+      end
+    end
+  end
+end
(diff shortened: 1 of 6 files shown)


app/lib/action_controller/request_forgery_protection/exception_and_reset_strategy.rb
```diff
@@ -0,0 +1,22 @@
...
+
+      def handle_unverified_request
+        @controller.reset_session
+        exception.handle_unverified_request
+      end
+
+      private
+
+      def exception
+        @exception || ActionController::RequestForgeryProtection::ProtectionMethods::Exception.new(@controller)
```
akostadinov, line 18:
```suggestion
        @exception ||= ActionController::RequestForgeryProtection::ProtectionMethods::Exception.new(@controller)
```

Or was `||` intentional?

app/lib/action_controller/request_forgery_protection/exception_and_reset_strategy.rb
```diff
@@ -0,0 +1,22 @@
...
+
+module ActionController
+  module RequestForgeryProtection
+    class ExceptionAndResetStrategy
+      def initialize(controller)
+        @controller = controller
+      end
+
+      def handle_unverified_request
+        @controller.reset_session
```
akostadinov, line 11: should we also update the comment on https://github.com/3scale/porta/blob/5c9335b7bf1a347070524dc4250ddf5ac9563d33/app/lib/authenticated_system.rb#L25

jlledom:
> How does the error page presently look? Can you show a screenshot?

In my local environment, like this. Not sure how it looks in production.

image

app/lib/action_controller/request_forgery_protection/exception_and_reset_strategy.rb
jlledom: Done

akostadinov (approved):

PR #3529, opened 2023-08-23
🔥🥒 Delete old scenario and selector

josemigallas:
🔥🔥🔥 

Back in the day, there was a widget in the Dashboard showing details about the first API (product). With APIaP it was replaced by the current widgets Products and Backends.

These 2 widgets are tested in `features/provider/dashboard/widgets.feature`.

Screenshot 2023-08-23 at 13 14 30

### DIFF
--- a/features/old/menu/dashboard.feature
+++ b/features/old/menu/dashboard.feature
@@ -16,9 +16,6 @@ Feature: Dashboard
     And I should see the link "0 DRAFTS" in the audience dashboard widget
     And I should see the link "0 MESSAGES" in the audience dashboard widget
 
-  Scenario: first API widget
-    And I should see "API" in the first api dashboard widget
-
   Scenario: Audience widget with Finance enabled
     Given the provider is charging its buyers
     And I go to the provider dashboard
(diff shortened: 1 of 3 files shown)


features/old/menu/dashboard.feature
```diff
@@ -16,9 +16,6 @@ Feature: Dashboard
     And I should see the link "0 DRAFTS" in the audience dashboard widget
     And I should see the link "0 MESSAGES" in the audience dashboard widget
 
-  Scenario: first API widget
-    And I should see "API" in the first api dashboard widget
```
jlledom, line 20: How could this work before if the old widget doesn't exist?

josemigallas: This happens when tests are 💩. This scenario is asserting that "API" is on the screen inside the APIs widget:
Screenshot 2023-08-24 at 14 33 22

Instead of removing this in the past, the selector _the first api dashboard widget_ was updated so that it target the 2 widgets Products and Backends...

jlledom (approved):

PR #3531, opened 2023-08-24
🦋 Compose Message form

josemigallas:
[THREESCALE-10079: Messages > Outbox > New](https://issues.redhat.com/browse/THREESCALE-10079)

Heads up ⚠️ To see the new required star `"*"` you need to restart rails.

### Before
Screenshot 2023-08-23 at 17 11 11

### After (`p/admin/messages/outbox/new`):
Screenshot 2023-09-19 at 08 55 12
Screenshot 2023-09-19 at 08 55 20
Screenshot 2023-09-19 at 08 55 37

### After (`buyers/accounts/:id`):
Screenshot 2023-09-19 at 08 56 10
Screenshot 2023-09-19 at 08 56 14
Screenshot 2023-09-19 at 09 09 31


config/initializers/formtastic.rb
```diff
@@ -24,7 +24,7 @@
   # '<abbr title="required">*</abbr>'. In other words, if you configure formtastic.required
   # in your locale, it will replace the abbr title properly. But if you don't want to use
   # abbr tag, you can simply give a string as below.
-  # Formtastic::FormBuilder.required_string = "(required)"
+  Formtastic::FormBuilder.required_string = '&nbsp;<span class="pf-c-form__label-required" aria-hidden="true">&#42;</span>'
```
jlledom, line 27: This will affect all forms in porta, right? is that OK?

Don't we have many forms including the `*` manually? woudln't this add an additional `*` to them?

josemigallas: Setting this config won't add anything new but it will replace the default `<abbr title="required">*</abbr>` that is added to any input with `required: true`.
I think we only add this manually in React and Liquid templates, where we don't use formtastic.

jlledom: If you're sure it's right. OK. Approved.

jlledom (dismissed):

app/views/provider/admin/messages/outbox/new.html.slim
```diff
@@ -0,0 +1,36 @@
...
+      - if sending_to_all
+        div class="pf-c-alert pf-m-info pf-m-inline" aria-label="Inline information alert"
+          div class="pf-c-alert__icon"
+            i class="fas fa-fw fa-info-circle" aria-hidden="true"
+          p class="pf-c-alert__title"
+            ' If you want to send a message to a particular user, go to the
+            = link_to 'accounts page', admin_buyers_accounts_path
+            ' , click on the user and then click on the send message icon.
+
+      = form.input :to, as: :patternfly_input, label: 'To:', readonly: true, input_html: { readonly: true, value: sending_to_all ? 'All Users' : @message.to.map(&:org_name).join(', '), name: nil }
```
mayorova, line 21: Maybe we don't need a colon? "To:" => "To"

mayorova:
I think this can be improved. Currently, there is this this error message when submitting the form with missing required inputs:
Screenshot from 2023-09-07 16-01-22

However, I believe in Patternfly:
- the submit button should be disabled until all required fields are filled in (I might be wrong though)
- when there is an error, it should be shown in patterfly way, i.e. 
Screenshot from 2023-09-07 16-12-23

josemigallas:
> I think this can be improved. Currently, there is this this error message when submitting the form with missing required inputs: Screenshot from 2023-09-07 16-01-22
> 
> However, I believe in Patternfly:
> 
> * the submit button should be disabled until all required fields are filled in (I might be wrong though)
> * when there is an error, it should be shown in patterfly way, i.e.
>   Screenshot from 2023-09-07 16-12-23

Ideally, yes. However this is just a paint job. Making the form fully interactive requires either using React (which is a bigger effort) or adding scripts that mutate the DOM manually (a bigger effort probably).

IMHO we should focus first on replacing legacy styles with Patternfly. It's a relatively easy task do to and review and everything will look consistent.

mayorova:
Now the "required" asterisks look like this, is it expected?
image

mayorova:
Not related to this PR, but it's a quick fix:
on Sent Messages index table, the column "From" should in fact be "To". It's a regression introduced here https://github.com/3scale/porta/commit/8013b4a2caeec6417fd21c6a801cda5559b0bec1

mayorova:
A couple of more things:

> Ideally, yes. However this is just a paint job. Making the form fully interactive requires either using React (which is a bigger effort) or adding scripts that mutate the DOM manually (a bigger effort probably).
> IMHO we should focus first on replacing legacy styles with Patternfly. It's a relatively easy task do to and review and everything will look consistent.

Well, also, when there is a server-side error (e.g. subject or body are submitted empty, or maybe some other validation fails), these errors are not shown on the inputs themselves. But well, it can be considered a future improvement.

Also, I see that the bulk message form is still old - this is expected, I assume, and will be handed separately?

josemigallas:
> Well, also, when there is a server-side error (e.g. subject or body are submitted empty, or maybe some other validation fails), these errors are not shown on the inputs themselves. But well, it can be considered a future improvement.

This is very easy to do in the case of p/admin/messages/outbox/new because the form is submitted via html but in the moda in buyers/accounts/ is AJAX.. and it means mutating the DOM manually with "complex" scripts.. I will see what I can do.

> Also, I see that the bulk message form is still old - this is expected, I assume, and will be handed separately?

Yes the bulk actions will be taken care all at once eventually.

mayorova:
> @mayorova I implemented the inline errors BUT they only appear after submitting the form. That means, HTML validation kicks in first (the ugly popover). A solution would be to rely on server validation only and make subject and body inputs not "required".

@josemigallas not sure... This "required" was added explicitly in https://github.com/3scale/porta/pull/2983/files and removing it would kind of cause a regression - although since then a server-side validation was added also, so in theory "this is fine". It's OK for me to keep it as it is.

mayorova (approved):
Reviewing the functionality rather than the code (as I found it a bit too complicated :grimacing:).

Works well and looks nice (except the minor "html-required" thing which I can happily ignore)

features/step_definitions/fields_steps.rb
```diff
@@ -1,4 +1,16 @@
...
+# FIXME: this step should assert that input has required=true
 Then /^fields should be required:$/ do |table|
+  table.rows.each do |row|
+    field = row.first
+    assert find('label', text: field).has_css?('span.pf-c-form__label-required'),
+           "Field #{field} is not required"
+  end
+end
+
+# FIXME: this used in some liquid templates, it's wrong, remove it.
```
jlledom, line 12: Are these FIXMEs work in progress?

app/controllers/provider/admin/messages/outbox_controller.rb
```diff
@@ -20,12 +19,14 @@ def destroy
   end
 
   def create
-    @message = messages.build(message_params)
+    @message = build_message(message_params)
     if @message.valid?
       enqueue_message_and_respond
     else
-      flash[:error] = @message.errors.full_messages.to_sentence
```
jlledom, line 27: Is the flash message not used anymore?

features/step_definitions/fields_steps.rb
josemigallas: This FIXME in particular not because it would affect many scenarios and maybe we don't even want to do it. But we shouldn't lose sigh of it.

app/controllers/provider/admin/messages/outbox_controller.rb
josemigallas: Not anymore, now the errors are rendered via JS as inline errors for each input (in https://github.com/3scale/porta/pull/3531/files#diff-d0c02a70007d734a61efaae1acb175ce7890d010bed1d37d1524f37838edfc1f)

jlledom (approved):

PR #3533, opened 2023-08-24
🚧 Application custom user key form

josemigallas:
[THREESCALE-10082: Update Custom User Key form](https://issues.redhat.com/browse/THREESCALE-10082)


app/views/provider/admin/keys/edit.html.erb
```diff
@@ -1,14 +1,19 @@
...
+  &nbsp;
+</div>
+<div class="pf-c-card__body">
+  <p id="invalid-key-warning" style="display:none;">&nbsp;</p>
 
-<p id="invalid-key-warning" style="display:none;">&nbsp;</p>
+  <%= semantic_form_for @cinstance, url: provider_admin_application_key_path(@cinstance, @cinstance.user_key),
+                                    remote: true,
+                                    html: { class: 'pf-c-form pf-m-limit-width' }  do |form| %>
+    <%= form.input :user_key, as: :patternfly_input, required: true, input_html: { size: 100, value: nil, required: true } %>
```
jlledom, line 10: `required: true` appears twice here, is that correct?

jlledom (commented):
The hint format is different now, it was grayer and smaller before. IS the new format the Patternfly standard?

features/step_definitions/applications/keys_steps.rb
```diff
@@ -85,6 +85,7 @@ def limit_warning
 
 Then /^I should see application keys limit reached error$/ do
   wait_for_requests
+  binding.pry
```
mayorova, line 88: 👀

features/support/capybara.rb
```diff
@@ -96,7 +96,7 @@ def is_a?(klass)
 Capybara.register_driver :headless_chrome do |app|
   options = Selenium::WebDriver::Chrome::Options.new
 
-  options.add_argument('--headless')
+  # options.add_argument('--headless')
```
mayorova, line 99: :eyes:

app/controllers/provider/admin/keys_controller.rb
```diff
@@ -14,9 +14,13 @@ def edit
 
   # user key
   def update
-    @cinstance.user_key = params[:cinstance][:user_key]
-    unless @cinstance.save
-      flash.now[:error] = "Invalid key. Please review and try submitting again."
+    user_key = params[:cinstance][:user_key]
```
mayorova, line 17:
```suggestion
    if user_key.empty?
      @cinstance.errors.add(:user_key, :blank)
    else
      @cinstance.user_key = user_key
      @cinstance.save
    end
    flash.now[:error] = @cinstance.errors.full_messages.to_sentence if @cinstance.errors.present?
```

The suggestion is not really accurate, because there the final `end` is not included.

But basically the whole point is - rely on standard "invalid" or "blank" messages.

app/controllers/provider/admin/keys_controller.rb
```diff
@@ -14,9 +14,13 @@ def edit
...
+    if user_key.empty?
+      @cinstance.errors.add(:base, "Key can't be blank")
+      flash.now[:error] = "Key can't be blank"
+    else
+      @cinstance.user_key = user_key
+      flash.now[:error] = "Enter a valid key" unless @cinstance.save
     end
```
mayorova, line 24:
```suggestion
    if user_key.empty?
      @cinstance.errors.add(:user_key, :blank)
    else
      @cinstance.user_key = user_key
      @cinstance.save
    end
    flash.now[:error] = @cinstance.errors.full_messages.to_sentence if @cinstance.errors.present?
```

app/views/provider/admin/keys/edit.html.erb
```diff
@@ -1,14 +1,24 @@
...
+                                    html: { class: 'pf-c-form pf-m-limit-width' }  do |form| %>
+    <div id="invalid-key-warning" class="pf-c-alert pf-m-danger pf-m-inline" style="display:none;">
+      <div class="pf-c-alert__icon">
+        <i class="fas fa-fw fa-exclamation-circle" aria-hidden="true"></i>
+      </div>
+      <p class="pf-c-alert__title"></p>
+    </div>
 
-<p id="invalid-key-warning" style="display:none;">&nbsp;</p>
+    <%= form.input :user_key, as: :patternfly_input, required: true, input_html: { size: 100, value: nil, required: false } %>
```
mayorova, line 15: Hm, this `required: true` and `required: false` thing is quite confusing :upside_down_face:

features/provider/admin/applications/keys.feature
```diff
@@ -0,0 +1,49 @@
...
+
+  Rule: Backend v2
+    Background:
+      Given the provider uses backend v2 in his default service
+      And they are reviewing the buyer's application details
+
+    Scenario: Remove and add keys
+      Given the key limit for the application is reached
+      Then I should see application keys limit reached error
+      # When I follow "Delete" for first application key
```
mayorova, line 44: Do these steps need to be completed?

mayorova:
This might not be relevant to this PR (as it's talking about the User Key), but just in case wanted to point out here that this modal might need some styling too: `app/views/provider/admin/keys/new.html.erb`

Screenshot from 2023-09-07 22-35-21

Also mentioning it because both "custom user key" and "custom app key" were mentioned here https://issues.redhat.com/browse/THREESCALE-10000 and maybe we could fix both in the same PR.

app/views/provider/admin/keys/edit.html.erb
josemigallas: No, I'll fix it

app/controllers/provider/admin/keys_controller.rb
mayorova: Resolved with no comment?...

app/views/provider/admin/keys/edit.html.erb
mayorova: So it is expected to have it like this? Wouldn't one override another?

features/step_definitions/applications/keys_steps.rb
mayorova: :eyes: :eyes:

app/views/provider/admin/keys/edit.html.erb
josemigallas: No, it's not expected. One is for formtastic's internal logic, another is for the html tag.

config/locales/en.yml
```diff
@@ -910,6 +918,10 @@ en:
           attributes:
             plan:
               not_allowed: "not allowed in this context"
+            keys:
+              invalid: Invalid. Use only alphanumeric characters [0-9, a-z, A-Z], hyphen-minus (-), no spaces and 40 characters at most.
```
mayorova, line 922: Where is `40` coming from?

The validation is for [5..255 characters](https://github.com/3scale/porta/blob/b122bced37550ea23661cee1e7a13c4ccd50d105/app/models/application_key.rb#L15) and it used to say "256 characters at most".

Also, `/\A[\x20-\x7E]+\Z/` includes a bunch of "special" characters, so we need to clarify that als. And in fact - we need to restrict the validation rules because some characters should not be allowed - but there is another JIRA for this, I think.

config/locales/en.yml
```diff
@@ -910,6 +918,10 @@ en:
...
+            user_key:
+              invalid: Invalid. Use only alphanumeric characters [0-9, a-z, A-Z], hyphen-minus (-), no spaces and 100 characters at most.
```
mayorova, line 924: similar to my other comment, where is `100` coming from? :sweat_smile: 

[here](https://github.com/3scale/porta/blob/b122bced37550ea23661cee1e7a13c4ccd50d105/app/models/cinstance.rb#L122-L123) it says 256

app/controllers/provider/admin/keys_controller.rb
```diff
@@ -26,8 +28,9 @@ def create
 
     if @key.persisted?
       @keys = @cinstance.application_keys.pluck_values
+      @notice = t('.create.success')
     else
-      flash.now[:error] = "Invalid key. Please review and try submitting again."
+      @cinstance.errors.add(:keys, :invalid)
```
mayorova, line 33: I think this is not ideal, because it doesn't give enough info...
The actual error would be coming from `@key.errors`, which could include:

`"Value has already been taken"` or `"Limit reached"` (as only 5 keys are allowed).

However, value can also be e.g. `["Value is too short (minimum is 5 characters)"]`, which would be covered by the current `cinstance.keys.invalid` message.

app/views/provider/admin/keys/new.html.erb
```diff
@@ -1,28 +1,37 @@
-<h2>Add key</h2>
+<div class="pf-c-card__title">
+  &nbsp;
+</div>
 
-<p id="invalid-key-warning" style="display:none;">&nbsp;</p>
+<div class="pf-c-card__body">
+  <%= form_tag provider_admin_application_keys_path(@cinstance), class: 'pf-c-form pf-m-limit-width',
+                                                                 data: { remote: 'true' } do |form| %>
```
mayorova, line 7:
```suggestion
                                                                 remote: true do |form| %>
```

config/locales/en.yml
mayorova: The other JIRA that I was thinking of seems to be this one: https://issues.redhat.com/browse/THREESCALE-10002

app/controllers/provider/admin/keys_controller.rb
akostadinov: it is marked `outdated`, so I think github close these

app/views/provider/admin/messages/outbox/create.js.erb
```diff
@@ -2,19 +2,16 @@
...
       const input = document.getElementById('<%= "message_#{field.to_s}" %>')
-      console.log(input, input.getAttribute('aria-invalid'))
 
       if (input.getAttribute('aria-invalid') === 'true') {
         // HACK: Remove helperText and add it again, since replacing the text also removes the icon.
         input.closest('.pf-c-form__group-control').children[1].remove()
       }
 
-      input.insertAdjacentHTML('afterend', '<%= j render partial: "shared/pf_error_helper_text", locals: { error: errors[field].first } %>')
+      input.insertAdjacentHTML('afterend', '<%= j render partial: "shared/pf_error_helper_text", locals: { error: errors.first } %>')
```
akostadinov, line 14: Excuse my ignorance, above we check `@message.errors[field].presence` and here we use `errors.first`. Can you explain that?

app/controllers/provider/admin/keys_controller.rb
akostadinov: maybe merge errors then? I believe I've seen this done somewhere else.

mayorova: > I believe I've seen this done somewhere else.

yeah, that rings a bell, but I don't remember where I saw it either :sweat_smile:

jlledom (commented):
I found a problem: since the backend validation for keys is essentially broken (See https://issues.redhat.com/browse/THREESCALE-10002), what the hint says about the key format is incorrect, and keys with e.g. spaces pass the ruby validation, and the call to `save` returns true.

When using App ID + Key, and a new custom key is added, and the key includes spaces, the key is sent to the backend, and stored correctly in the database.

Is there a javascript validation for this? If so, the form shouldn't be sent when the client validation fails.

mayorova:
yeah @jlledom I mentioned this here also: https://github.com/3scale/porta/pull/3533#discussion_r1338734624

But I don't think adding a JS validation (duplicating the server-side one) is a good approach.
I'd rather fix the validation on the server-side (restrict the characters in the existing regex - because some characters should probably not be supported). I think this needs to be done in another PR though.

app/views/provider/admin/messages/outbox/create.js.erb
josemigallas: `@message.errors` returns an array of errors but we only render the first one. There is not supposed to be more errors anyway.

app/controllers/provider/admin/keys_controller.rb
josemigallas: I'll try to improve it but I would want to block this PR on something that's been in the code since forever. Another PR for improving error handling would probably better.

config/locales/en.yml
josemigallas: It was like this already, who knows. I updated it.

josemigallas: The validation is bogus, the following key is actually valid `%%%%%` 🤷🏻 but nobody seems to care

josemigallas:
@mayorova @jlledom 

About the server validation: I agree we should fix it in another PR.

About updating the input hint: It doesn't make sense to me to change it now and then again. Let's come up with a definitive one in the second PR.

jlledom (dismissed):

mayorova:
> About updating the input hint: It doesn't make sense to me to change it now and then again. Let's come up with a definitive one in the second PR.

Yeah, that's OK.
What I don't like is this block https://github.com/3scale/porta/pull/3574/files#diff-44438ce218f5287c58d0017f965d888715635d94280669896f75841fbd7b4cd7L921-L931

mayorova (approved):
:rocket:

jlledom (approved):

PR #3537, opened 2023-09-04
Display percentage at Current utilization

lvillen:
[THREESCALE-10133](https://issues.redhat.com/browse/THREESCALE-10133): Display percentage at Current utilization

This PR fixes an error at Current utilization, where it was duplicating the value metrics instead of displaying value and percentage. 

**Verification steps** 

**Before**
Captura de pantalla 2023-09-04 a las 14 10 14

**After**
Captura de pantalla 2023-09-04 a las 14 10 34

from @mayorova:

Just for reference, the bug was introduced here: https://github.com/3scale/porta/commit/ee6afab4bc84146c6327f16e09a88bf2eb906cab#diff-c03110c10f92df9dcba83b75e405de05ac811ef040b854a1692ad71c6479cdecR45

### DIFF
--- a/app/views/provider/admin/applications/_utilization.html.erb
+++ b/app/views/provider/admin/applications/_utilization.html.erb
@@ -42,7 +42,7 @@
               <%= content_tag(:span, utilization, class: utilization_type) %>
             </td>
             <td role="cell" data-label="%" class="utilization">
-              <%= content_tag(:span, utilization, class: utilization_type) %>
+              <%= content_tag(:span, percentage, class: utilization_type) %>
             </td>
           </tr>
         <% end %>


akostadinov (approved):

app/views/provider/admin/applications/_utilization.html.erb
```diff
@@ -42,7 +42,7 @@
               <%= content_tag(:span, utilization, class: utilization_type) %>
             </td>
             <td role="cell" data-label="%" class="utilization">
```
akostadinov, line 44: should class also change or style is fine? I guess it is fine

mayorova (approved):
Looks good, but it seems that we are missing a regression test, that would have allowed us to catch this in the first place.

Do you think it would be hard to add? (we can potentially add it in another PR)

PR #3545, opened 2023-09-13
🔒️ Set cache headers to no-cache

lvillen:

[THREESCALE-10076: Content caching not set for personal tokens page. Allows page retrieval when logged out.
](https://issues.redhat.com/browse/THREESCALE-10076)

When logging out and clicking back button on the browser page is rendered from the browser cache.

**Verification steps** 

Though the PR says it's in the personal tokens page, it happens anywhere so it could be verified there and also wherever in the application. Just log out and confirm that when clicking back you won't be redirected to the previous page.

**Update**

As @jlledom proposed, finally we'll be applying this to the following screens: 

Admin portal:
- [x] /p/admin/api_docs
- [x] /p/admin/account
- [x] /p/admin/user/access_tokens
- [x] /p/admin/account/authentication_providers <- Maybe? not sure about this one
- [x] /p/admin/webhooks/ <- Could include tokens in the URL as paremeters
- [x] /apiconfig/services/<ID>/settings
- [x] /apiconfig/services/<ID>/integration
- [x] /apiconfig/services/<ID>/api_docs/<ID>/preview
- [x] /p/admin/applications/<ID>
- [x] /site/dns
- [x] /p/admin/authentication_providers (applied everywhere except index)
- [x] /p/admin/api_docs/account_data // Couldn't test it but the method is there
- [x] /finance/settings

Developer portal:
- [x] /
- [x] /admin/applications/<ID>
- [x] /docs

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -18,9 +18,9 @@ class ApplicationController < ActionController::Base
   skip_forgery_protection if: -> { api_controller? }
 
   before_action :set_timezone
-
   before_action :enable_analytics
   before_action :check_browser
+  before_action :set_cache_headers
 
   def status
     begin
@@ -229,5 +229,13 @@ class ApplicationController < ActionController::Base
   def other_modern_browsers?(browser)
     (browser.firefox? && browser.device.tablet? && browser.platform.android?('>= 14')) ||
       (!browser.compatibility_view? && (browser.ie?('>= 9') || browser.edge?))
+  end
+
+  def set_cache_headers
+    response.headers.merge!(
+      'Cache-Control' => 'no-cache, no-store',
+      'Pragma' => 'no-cache',
+      'Expires' => 'Mon, 01 Jan 1990 00:00:00 GMT'
+    )
   end
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -79,6 +79,14 @@ class ApplicationController < ActionController::Base
     end
   end
 
+  def disable_client_cache
+    response.headers.merge!(
+      'Cache-Control' => 'no-cache, no-store',
+      'Pragma' => 'no-cache',
+      'Expires' => 'Mon, 01 Jan 1990 00:00:00 GMT'
+    )
+  end
+
   protected
 
   def check_browser
@@ -228,13 +236,5 @@ class ApplicationController < ActionController::Base
   def other_modern_browsers?(browser)
     (browser.firefox? && browser.device.tablet? && browser.platform.android?('>= 14')) ||
       (!browser.compatibility_view? && (browser.ie?('>= 9') || browser.edge?))
-  end
-
-  def disable_client_cache
-    response.headers.merge!(
-      'Cache-Control' => 'no-cache, no-store',
-      'Pragma' => 'no-cache',
-      'Expires' => 'Mon, 01 Jan 1990 00:00:00 GMT'
-    )
   end
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -20,7 +20,6 @@ class ApplicationController < ActionController::Base
   before_action :set_timezone
   before_action :enable_analytics
   before_action :check_browser
-  before_action :set_cache_headers
 
   def status
     begin
@@ -231,7 +230,7 @@ class ApplicationController < ActionController::Base
       (!browser.compatibility_view? && (browser.ie?('>= 9') || browser.edge?))
   end
 
-  def set_cache_headers
+  def disable_client_cache
     response.headers.merge!(
       'Cache-Control' => 'no-cache, no-store',
       'Pragma' => 'no-cache',
--- a/doc/liquid/tags.md
+++ b/doc/liquid/tags.md
@@ -63,6 +63,18 @@ We recommend __to remove this tag__ from public templates.
 ```liquid
 {% debug:help %}
 ```
+# Tag 'disable_client_cache'
+
+Adds HTTP headers to disable the browser cache for the current screen.
+
+__Example:__ Disable browser cache for this screen
+```liquid
+<html>
+  <head>
+    {% disable_client_cache %}
+  </head>
+</html>
+```
 # Tag 'email'
 
 
--- a/lib/developer_portal/lib/liquid/tags/disable_client_cache.rb
+++ b/lib/developer_portal/lib/liquid/tags/disable_client_cache.rb
@@ -12,8 +12,7 @@ module Liquid
 
       desc "Adds HTTP headers to disable the browser cache for the current screen."
       def render(context)
-        controller = context.registers[:controller].dup
-        controller.disable_client_cache
+        context.registers[:controller].disable_client_cache
         nil
       end
     end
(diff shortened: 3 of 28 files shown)


app/controllers/application_controller.rb
```diff
@@ -20,9 +20,9 @@ class ApplicationController < ActionController::Base
   end
 
   before_action :set_timezone
-
   before_action :enable_analytics
   before_action :check_browser
+  before_action :set_cache_headers
```
akostadinov, line 25: Shouldn't we apply this only in certain situations where sensitive data might be preserved locally?

jlledom: I agree with @akostadinov. Disabling the cache everywhere has side effects:

- The overall performance will be affected since all requests must be always sent to the server.
- Costs will increase for providers, because of the increase of requests received.

I think we should keep the cache enabled by default and disable it only for a white list of screens that include auth data, like tokens. I would do it this way:

1. Add a new method to `ApplicationController`: `disable_client_cache?` which returns `false`
2. Rename `set_cache_hesders` to `disable_client_cache`, it's more clear
3. Call `disable_client_cache` only when `disable_client_cache?` is `true`. Like this:
```
before_action :disable_client_cache, if: :disable_client_cache?
```
4. Define the method `disable_client_cache?` to return `true` in those particular controllers that include auth data.

The only problem is to find that list of controllers. The next come to my mind:

**Admin portal:**
- `/p/admin/api_docs`
- `/p/admin/account`
- `/p/admin/user/access_tokens`
- `/p/admin/account/authentication_providers` <- Maybe? not sure about this one
- `/p/admin/webhooks/` <- Could include tokens in the URL as paremeters
- `/apiconfig/services/<ID>/settings`
- `/apiconfig/services/<ID>/integration`
- `/apiconfig/services/<ID>/api_docs/<ID>/preview`
- `/p/admin/applications/<ID>`
- `/site/dns`
- `/p/admin/authentication_providers/<ID>/edit`
- `/p/admin/authentication_providers/new`

**Developer portal:**
- `/`
- `/admin/applications/<ID>`
- `/docs`

The developer portal is tricky because URLs can change, and we can't control what users show in their templates anyway, so the whole developer portal should have cache disabled probably.

Since this should be merged soon, I think it's OK to merge this now and work on the fine-grain filtering of controllers later.

jlledom (dismissed):

app/controllers/application_controller.rb
mayorova: > /p/admin/account/authentication_providers <- Maybe? not sure about this one

Yeah, while `#index` seems safe, but `#show` has the provider's client ID and secret.
So, I think we can add it.

> /p/admin/webhooks/ <- Could include tokens in the URL as paremeters

Agreed


Also,

`/p/admin/api_docs/account_data` - it returns user's credentials

`/finance/settings` - shows the payment gateway settings

jlledom:
> @mayorova and I where thinking if we need to add the `disable_client_cache` here or not.

The change would affect all the developer portal. The problem is the developer portal receives many more visits than the admin portal. So disabling cache on it would have a bigger impact on the customer resources and site performance perceived by users. For that reason, we should keep using the cache I think.

I've been looking for other ways to disable cache and I found that HTML `<meta>` tags could work. Like these:

```
<meta http-equiv="Cache-Control" content="no-cache, no-store, must-revalidate">
<meta http-equiv="Pragma" content="no-cache">
<meta http-equiv="Expires" content="0">
```

The idea is to create a new liquid method to add this tags, and add a call to that method in the default templates we provide and consider vulnerable. But we need to somehow place them inside the `<head>` and I'm not sure we can do that with liquid. Also, this would only work for new clients. Old clients should find this in a Release Note and add the call to the new liquid method manually for the templates they think are vulnerable.

Not sure this could work but couldn't investigate further.

mayorova:
I agree with @jlledom 

And great find that it can be achieved with `<meta>` tags.

This means that the customers can add these tags themselves - normally in the layout. E.g. the default Main layout that we provide:
image

So I even think it's not worth adding it as Liquids.

But it might be good to add this tip to the documentation.

jlledom:
> So I even think it's not worth adding it as Liquids.
> 
> But it might be good to add this tip to the documentation.

It doesn't work 😥

I tried adding the meta tags to the layout but Firefox ignores them. Also they are not officially documented:

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta#http-equiv

jlledom:
Now another idea came to my mind. Just thinking aloud. Maybe we can add a liquid method that calls `set_cache_headers`? Then add that method to our vulnerable templates and write a release note so clients can do the same for their portals?

WDYT @mayorova could that work?

akostadinov:
Sounds reasonable, just figure out a more clear name maybe.

app/controllers/application_controller.rb
```diff
@@ -80,6 +80,14 @@ def sublayout
     end
   end
 
+  def set_cache_headers
```
mayorova, line 83: Maybe `set_no_cache_headers` ? 
I don't know what's better, but probably it better reflects the meaning :grimacing:

mayorova (dismissed):

lib/developer_portal/lib/liquid/tags/disable_client_cache.rb
```diff
@@ -0,0 +1,21 @@
...
+          <head>
+            {% disable_client_cache %}
+          </head>
+        </html>
+      }
+
+      desc "Adds HTTP headers to disable the browser cache for the current screen."
+      def render(context)
+        controller = context.registers[:controller].dup
+        controller.disable_client_cache
```
akostadinov, line 16: why dup instead of `context.registers[:controller].disable_client_cache`?

lvillen: @jlledom I think this one is for you ⬆️

doc/liquid/tags.md
```diff
@@ -63,6 +63,18 @@ We recommend __to remove this tag__ from public templates.
 ```liquid
 {% debug:help %}
 ```
+# Tag 'disable_client_cache'
+
+Adds HTTP headers to disable the browser cache for the current screen.
```
jlledom, line 68: I would add a comment here explaining why they would want to do this. They should add this tag in all their templates containing sensitive data.

lib/developer_portal/lib/liquid/tags/disable_client_cache.rb
jlledom: Because I copied it from: https://github.com/3scale/porta/blob/25fa9bc53bdd3eeb8f2a9c8e882f771c85c3c3cb/lib/developer_portal/lib/liquid/tags/csrf.rb#L15

But you're right, `.dup` is not needed here.

jlledom (dismissed):

doc/liquid/tags.md
mayorova: Yeah, makes sense. Except that it would need to be updated here: https://github.com/3scale/porta/pull/3545/files#diff-656368b7147747316683f3193e866e3530e6636908a1fd6cef8a2379eba12ac3R13

mayorova (approved):

jlledom (approved):

PR #3558, opened 2023-09-19
THREESCALE-10133: Add a cucumber for Current Utilization section

jlledom:

Add a test to ensure `Current Utilization` section in `/p/admin/applications/[ID]` works properly.

[THREESCALE-10133](https://issues.redhat.com/browse/THREESCALE-10133)

### DIFF
--- a/features/step_definitions/stats/application_steps.rb
+++ b/features/step_definitions/stats/application_steps.rb
@@ -7,46 +7,10 @@ Given(/^a provider has a developer "(.*?)" with an application name "(.*?)"$/) d
   }
 end
 
-When(/^the provider is logged in and visits the "(.*?)" application stats$/) do |developer_app_name|
-  steps %q{
-    Given current domain is the admin domain of provider "foo.3scale.localhost"
-    And I log in as provider "foo.3scale.localhost"
-  }
-
-  click_on(text: /\d Applications?/i, match: :first)
-
-  within '.cinstances' do
-    click_on developer_app_name, match: :one
-  end
-
-  assert_text "Application '#{developer_app_name}'"
-
-  within(selector_for('the secondary nav')) do
-    click_on 'Analytics', match: :one
-  end
-
-  assert_text "Traffic statistics for #{developer_app_name}"
-end
-
-When(/^the provider is logged in and visits the "(.*?)" application page$/) do |developer_app_name|
-  steps %q{
-    Given current domain is the admin domain of provider "foo.3scale.localhost"
-    And I log in as provider "foo.3scale.localhost"
-  }
-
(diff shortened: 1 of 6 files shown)


features/step_definitions/transaction_steps.rb
```diff
@@ -22,3 +22,10 @@
 When /^the buyer makes (\d+|a) service transactions? with:$/ do |count, table|
   step %(buyer "#{@buyer.name}" makes #{count} service transactions with:), table
 end
+
+Given "the backend will respond to a utilization request for application {string} with:" do |application_name, json|
+  application = Cinstance.by_name(application_name).first
+  TestHelpers::Backend::MockCore.stubs.get("/internal/services/#{application.service_id}/applications/#{application.application_id}/utilization/") do
+    [ 200, {'content-type'=>'application/json'}, json ]
```
jlledom, line 29: After a hell of hours of pure suffering, I found this seems to be the one and only way to stub requests to Backend from tests, so I wonder why [this](https://github.com/3scale/porta/blob/f6e98d1d8a66f28d13a91ebba8767e12814509ae/features/old/applications/buyers/referrer_filters/buyer_side_multiple_applications.feature#L42) works...

jlledom: That makes sense

features/step_definitions/stats/application_steps.rb
```diff
@@ -28,8 +28,31 @@
   assert_text "Traffic statistics for #{developer_app_name}"
 end
 
+When(/^the provider is logged in and visits the "(.*?)" application page$/) do |developer_app_name|
```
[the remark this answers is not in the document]
jlledom: We are verifying this section shows the data properly:

image

Ok, I'll remove them

features/step_definitions/transaction_steps.rb
```diff
@@ -22,3 +22,8 @@
 When /^the buyer makes (\d+|a) service transactions? with:$/ do |count, table|
   step %(buyer "#{@buyer.name}" makes #{count} service transactions with:), table
 end
+
+Given "the backend will respond to a utilization request for the application with:" do |json|
```
jlledom, line 26: I think I addressed all your comments. Now I have a question, I tried defining this step like this:

```suggestion
Given "the backend will respond to a utilization request for {application} with:" do |application, json|
```
To follow the same formula as in [here](https://github.com/3scale/porta/pull/3558/files#diff-235c89ea8624bdbdda33252cc58951529ef4bc59a43e7b079df8b7b1a2425008R69).

But it failed to take the application correctly. Do you know why it works for the buyer case above but not for this case with the application?

PR #3572, opened 2023-09-27
THREESCALE-8404: Add TLS and ACL support for Redis

jlledom:

This implements support for ACL and TLS credentials for Redis from Porta.

After Sidekiq 7 upgrade, this PR has become pretty trivial: It only requires some minimal changes in the Redis config parser class to be able to read SSL params from yaml config files. It also loads a default CA cert from the `/config` folder if one is needed and no other certificate is provided.

All other changes are on tests. Also, all yaml example files have been updated.

[THREESCALE-8404](https://issues.redhat.com/browse/THREESCALE-8404)

**Verification steps** 

This requires three steps:

1. Generate TLS certificates and keys for Redis.
2. Configure Redis to enable ACL and TLS, and use the generated keys.
3. Configure Porta.
   3.1. TLS Client
   3.2. Mutual TLS Client

#### Generate keys and certificates:

The first thing we need is a certification authority. For that, we need to create its key and certificate.

- key: `openssl genrsa -out ca-root-key.pem 4096`
- cert: `openssl req -new -x509 -days 365 -key ca-root-key.pem -out ca-root-cert.pem`

Then we'll have to create a key and certificate for the server, and sign it with our CA. Ensure setting `localhost` as `CN`, if that's the domain you're going to use to connect to Redis:

- key: `openssl genrsa -out redis.key 4096`
- cert request: `openssl req -new -key redis.key -out redis.csr`
- signed cert: `openssl x509 -req -days 365 -in redis.csr -CA ca-root-cert.pem -CAkey ca-root-key.pem -CAcreateserial -out redis.crt`

#### Configure Redis:

First, we need a Redis configuration file which enables ACL and TLS, defines the users and sets the certificate the server will use and the CA the server will trust. This is the minimal `redis.conf` for that purpose:

```
port 6379
tls-port 26379
tls-cert-file /etc/redis.crt
tls-key-file /etc/redis.key
tls-ca-cert-file /etc/ca-root-cert.pem
tls-auth-clients optional
user default off
user porta on >sup3rS3cre1! ~* &* +@all
user apisonator on >secret#Passw0rd ~* &* +@all
```

Next step is to launch a container that has access to the certificates, keys and Redis config files created above. We create a volume for each file and modify the container start command to reference the configuration file we created. This is how I configured the redis pod in my `docker-compose.yml`. But the container can be launched by other ways:

```yml
  redis:
    image: redis:6.2-alpine
    container_name: redis-compose-ssl
    ports:
      - "26379:26379"
    volumes:
      - /home/jlledom/redis.conf:/etc/redis.conf:Z
      - /home/jlledom/redis.crt:/etc/redis.crt:z
      - /home/jlledom/redis.key:/etc/redis.key:z
      - /home/jlledom/ca-root-cert.pem:/etc/ca-root-cert.pem:z
    command: [redis-server, /etc/redis.conf]
```

The `:z` at the end of the volume definitions is required when using Selinux, for instance in Fedora.

#### Configure Porta:

Porta can be configured as a regular TLS client, when only the server needs a certificate, or a Mutual TLS Client, where both client and server must provide a certificate.

##### As TLS Client:

Update our config files to add a new base configuration:

**`redis-yml`:**
```yml
ssl: &ssl
  url: "<%= ENV.fetch('REDIS_URL', 'rediss://localhost:26379/5') %>"
  pool_size: <%= ENV.fetch('RAILS_MAX_THREADS', 5) %>
  pool_timeout: 5 # this is in seconds
  namespace: "<%= ENV['REDIS_NAMESPACE'] %>"
  sentinels: "<%= ENV['REDIS_SENTINEL_HOSTS'] %>"
  role: <%= ENV['REDIS_SENTINEL_ROLE'] %>
  username: porta
  password: sup3rS3cre1!
  ssl_params:
    ca_file: /home/jlledom/ca-root-cert.pem
```

**`backend-redis-yml`:**
```yml
ssl: &ssl
  url: "<%= ENV.fetch('BACKEND_REDIS_URL', 'rediss://localhost:26379/6') %>"
  pool_size: <%= ENV.fetch('RAILS_MAX_THREADS', 5) %>
  pool_timeout: 5 # this is in seconds
  sentinels: "<%= ENV['BACKEND_REDIS_SENTINEL_HOSTS'] %>"
  role: <%= ENV['BACKEND_REDIS_SENTINEL_ROLE'] %>
  username: porta
  password: sup3rS3cre1!
  ssl_params:
    ca_file: /home/jlledom/ca-root-cert.pem
```

Note how the `ssl_params` section is not always required. We are using it here because the server is using a certificate signed by an unknown authority we've just created, so we need to explicitly tell Porta to trust that authority. If the server were using a certificate signed by any of the well-known CAs, the `ssl_params` section could be omitted.

Another possible situation is when the server is using a self-signed certificate. When this happens, there's no trusted CA to add to `ssl_params`, so the only way to go is to skip  certificate validation. Unfortunately we don't support this yet because the `redis-client` gem still haven't added support for verify modes (see: https://github.com/redis-rb/redis-client/issues/133) so we can't set `SSL_VERIFY_NONE` to the client. Anyway this is only useful for development purposes, we'll have to use our own CA for development for the moment.

##### As Mutual TLS Cliet:

Generate a key and a singed certificate for Porta:

- key: `openssl genrsa -out porta.key 4096`
- cert request: `openssl req -new -key porta.key -out porta.csr`
- signed cert: `openssl x509 -req -days 365 -in porta.csr -CA ca-root-cert.pem -CAkey ca-root-key.pem -CAcreateserial -out porta.crt`

Update our config files to add a new base configuration:

**`redis-yml`:**
```yml
ssl: &ssl
  url: "<%= ENV.fetch('REDIS_URL', 'rediss://localhost:26379/5') %>"
  pool_size: <%= ENV.fetch('RAILS_MAX_THREADS', 5) %>
  pool_timeout: 5 # this is in seconds
  namespace: "<%= ENV['REDIS_NAMESPACE'] %>"
  sentinels: "<%= ENV['REDIS_SENTINEL_HOSTS'] %>"
  role: <%= ENV['REDIS_SENTINEL_ROLE'] %>
  username: porta
  password: sup3rS3cre1!
  ssl_params:
    ca_file: /home/jlledom/ca-root-cert.pem
    cert: /home/jlledom/porta.crt
    key: /home/jlledom/porta.key
```

**`backend-redis-yml`:**
```yml
ssl: &ssl
  url: "<%= ENV.fetch('BACKEND_REDIS_URL', 'rediss://localhost:26379/6') %>"
  pool_size: <%= ENV.fetch('RAILS_MAX_THREADS', 5) %>
  pool_timeout: 5 # this is in seconds
  sentinels: "<%= ENV['BACKEND_REDIS_SENTINEL_HOSTS'] %>"
  role: <%= ENV['BACKEND_REDIS_SENTINEL_ROLE'] %>
  username: porta
  password: sup3rS3cre1!
  ssl_params:
    ca_file: /home/jlledom/ca-root-cert.pem
    cert: /home/jlledom/porta.crt
    key: /home/jlledom/porta.key
```

As mentioned above, the `ca_file` field is only required when the server isn't using a certificate signed by one of the well-known CAs.

#### Enable the TLS Client:

You can switch between TLS and non-TLS client by updating the `development` environment to use the new base, on both files:

```yml
development:
  <<: *ssl
```

At this point it _**should**_ work. You can verify the connection using RedisInsight or trying with invalid certs or keys from the porta side, to verify it works fine.


config/initializers/sidekiq_hacks.rb
```diff
@@ -0,0 +1,85 @@
...
+
+# This monkey path can be deleted after migrating to Sidekiq >= 7
+module OptionsPatch
+  def client_opts(options)
+    valid_keys = %i[url host port namespace sentinels role path username password ssl_params]
+    super(options.slice(*valid_keys))
+  end
+end
+
+Sidekiq::RedisConnection::RedisAdapter.prepend OptionsPatch
```
jlledom, line 11: I'm not a fan of monkey patches, but this is required to fix a bug in Sidekiq, due to the new `redis-client` gem we use not being officially supported until `sidekiq >= 7`

config/initializers/sidekiq_hacks.rb
```diff
@@ -0,0 +1,85 @@
...
+# Delete this after solving https://github.com/breamware/sidekiq-batch/issues/86
+module CorrectArgumentBatch
+  def enqueue_callbacks(event, bid)
+    event_name = event.to_s
+    batch_key = "BID-#{bid}"
+    callback_key = "#{batch_key}-callbacks-#{event_name}"
+    already_processed, _, callbacks, queue, parent_bid, callback_batch = Sidekiq.redis do |r|
+      r.multi do |pipeline|
+        pipeline.hget(batch_key, event_name)
+        pipeline.hset(batch_key, event_name, true.to_s)
```
jlledom, line 24: `sidekiq-batch` doesn't support the `redis-client` gem either, but we need the last to add support for TLS, so I had to add this patch and overwrite the whole method only to change this line, to include the call to `.to_s`.

I created an issue but I don't think `sidekiq-batch` is maintained anymore. It's a problem because Sidekiq 7 only supports `redis-client` so `sidekiq-batch` is not really compatible with that version.

Also, I can't know in which other places is `sidekiq-batch` sending wrong parameters to redis. So there will be potential issues when deploying.

akostadinov: You can also create a PR in `sidekiq-batch` should be easy enough and more likely to attract any surviving maintainer's  attention.

config/initializers/sidekiq_hacks.rb
akostadinov: A comment will help here too.

jlledom:
@akostadinov @mayorova I added some comments

config/initializers/sidekiq_hacks.rb
jlledom: I removed this monkey patch and created a fork which includes this and another change. I'm not sure how many other changes will I have to make on `sidekiq-batch` to make it support `redis-client`, so I prefer to not keep adding patches.

https://github.com/3scale/sidekiq-batch

app/lib/three_scale/redis_config.rb
```diff
@@ -3,7 +3,7 @@
 module ThreeScale
   class RedisConfig
     def initialize(redis_config = {})
-      raw_config = (redis_config || {}).symbolize_keys
+      raw_config = (redis_config || {}).deep_symbolize_keys
```
jlledom, line 6: We need this because we now accept the new `:ssl_params` config parameter for redis, which is a hash

lib/tasks/boot.rake
```diff
@@ -18,7 +18,6 @@ namespace :boot do
   desc 'Tries to connect to Redis'
   task :redis do
     require Rails.root.join('app', 'lib', 'three_scale', 'redis_config')
-    require Rails.root.join('config', 'initializers', 'redis_hacks')
```
jlledom, line 21: This is a cleanup. We should have removed this in the Sidekiq 7 PR

config/examples/redis.yml
```diff
@@ -5,6 +5,14 @@ base: &default
   sentinels: "<%= ENV['REDIS_SENTINEL_HOSTS'] %>"
   name: <%= ENV['REDIS_SENTINEL_NAME'] %>
   role: <%= ENV['REDIS_SENTINEL_ROLE'] %>
+  # == ACL Params ==
+  username: <%= ENV['REDIS_ACL_USERNAME'] %>
```
mayorova, line 9: Hm, I would call the env vars just `REDIS_USERNAME` and `REDIS_PASSWORD` (with `BACKEND_` prefix accordingly), I don't think we need to mention in the env var itself that it's about ACL (it's good to have it in the comment though)

app/lib/three_scale/redis_config.rb
akostadinov: Thanks for the explanation!

openshift/system/config/redis.yml
```diff
@@ -5,3 +5,12 @@ production:
   sentinels: "<%= ENV['REDIS_SENTINEL_HOSTS'] %>"
   name: <%= ENV['REDIS_SENTINEL_NAME'] %>
   role: <%= ENV['REDIS_SENTINEL_ROLE'] %>
+  # == ACL Params ==
+  username: <%= ENV['REDIS_ACL_USERNAME'] %>
+  password: <%= ENV['REDIS_ACL_PASSWORD'] %>
+  # == TLS Params ==
+  ssl_params:
+    ca_file: <%= ENV['REDIS_TLS_CA_FILE'] %>
```
akostadinov, line 13: I thought I commented somewhere on this. I would like first to know how will customers be providing their CA (perhaps depends on operator) before finalizing this variable. To make sure that this is the expected way it should work. At the very least I think we should have a default location to search for this so that users can only mount a secret and then the files are present, porta to use them.

The point about default location applies to the cert and public key variables as well.

openshift/system/config/redis.yml
```diff
@@ -5,3 +5,12 @@ production:
...
+    cert: <%= ENV['REDIS_TLS_CERT'] %>
+    key: <%= ENV['REDIS_TLS_KEY'] %>
```
akostadinov, line 15:
```suggestion
    cert: <%= ENV['REDIS_USER_CERT'] %>
    key: <%= ENV['REDIS_PRIVATE_KEY'] %>
```

This naming makes it more clear about the intended purpose of these variables. `TLS` is redundant in this case I believe. What else can certificate be about? I'm not super much against the `TLS` part but it's nicer without I believe.

mayorova: I like the original names (i.e. `REDIS_TLS_CERT` and `REDIS_TLS_KEY`), because they match how Redis itself calls them, see https://redis.io/docs/management/security/encryption/#running-manually

Also, isn't it called "client certificate" rather than "user"?... But maybe I'm misinterpreting what `ssl_params.cert` represents :grimacing:

openshift/system/config/redis.yml
jlledom: Yes, you mentioned that in the Apisonator PR. If not provided, Redis will verify the server certificate against all trusted certificates in the OS default ssl certs folder. In Fedora that's `/etc/ssl/certs`.

I would say the `ca_path` param would work out of the box (See [`hiredis-client`](https://github.com/redis-rb/redis-client/blob/v0.17.0/hiredis-client/lib/redis_client/hiredis_connection.rb#L24-L30)) to provide an alternative directory to `/etc/ssl/certs`. I can add a new env variable for it, so let the clients use `ca_path` or `ca_file` as they want.

AFAIK there's no equivalent config param for client certs and keys folder, but I don't think it's really needed, in OpenShift you can mount folders as well as single files.

openshift/system/config/redis.yml
jlledom: > they match how Redis itself calls them, see https://redis.io/docs/management/security/encryption/#running-manually

They match the server parameters names, but we are the client here.

> Also, isn't it called "client certificate" rather than "user"?... But maybe I'm misinterpreting what `ssl_params.cert` represents 😬

As Porta is the client, `cert` is the client certificate.

What about these options?
- `REDIS_CA_FILE`, `REDIS_CERT` and `REDIS_KEY`
  - To match params names
- `REDIS_CLIENT_CA_FILE`, `REDIS_CLIENT_CERT` and `REDIS_CLIENT_KEY`
  - To note it's the client, and match params names

mayorova: > What about these options?
> 
> * `REDIS_CA_FILE`, `REDIS_CERT` and `REDIS_KEY`
>   
>   * To match params names
> * `REDIS_CLIENT_CA_FILE`, `REDIS_CLIENT_CERT` and `REDIS_CLIENT_KEY`
>   
>   * To note it's the client, and match params names

I think `CLIENT_CA_FILE` is not necessarily accurate, it's not client or server, it's just CA, isn't it? :upside_down_face: 

`REDIS_CA_FILE` and `REDIS_CERT` look good, but `REDIS_KEY` IMO is too vague.

So, I'd be sticking with the original options. Optionally, `_TLS_` could be replaced with `_SSL_`, but probably not good, because TLS term kind of supersedes SSL.

openshift/system/config/redis.yml
jlledom: I pushed some changes to allow looking for trusted CA certs through `ca_path`. However, you can't just store your cert file under `ca_path`, that doesn't work (check https://github.com/redis/redis/issues/7392). You need to manually install your certificate in your OS using the official way. For Fedora, that's [update-ca-trust](https://docs.fedoraproject.org/en-US/quick-docs/using-shared-system-certificates/). And, since this mechanism essentially stores the cert in the proper format in the default folder (`/etc/ssl/certs` for Fedora), there's no difference between providing or not the `ca_path` param.

I also tried setting `ca_path` to a directory under my `$HOME` which is a symbolic link to `/etc/ssl/certs`, and it worked fine.

Summarizing, I don't think `ca_path` is much useful but since it's an officially accepted parameter and I already added the code, I think it's OK to provide it, It might be useful to somebody.

openshift/system/config/redis.yml
jlledom: > I think `CLIENT_CA_FILE` is not necessarily accurate, it's not client or server, it's just CA, isn't it? 🙃

Right!! 😅

> `REDIS_CA_FILE` and `REDIS_CERT` look good, but `REDIS_KEY` IMO is too vague.
> 
> So, I'd be sticking with the original options.

What do you mean? Do you have an example of variable names?

mayorova: > What do you mean? Do you have an example of variable names?

I mean that I wouldn't change what you have now :)

```
  ssl_params:
   ca_file: <%= ENV['REDIS_TLS_CA_FILE'] %>
   ca_path: <%= ENV['REDIS_TLS_CA_PATH'] %>
   cert: <%= ENV['REDIS_TLS_CERT'] %>
   key: <%= ENV['REDIS_TLS_KEY'] %>
```

mayorova:
Except for the comment I left here: https://github.com/3scale/porta/pull/3572#discussion_r1372985920 everything seems to be working well! Nice :clap:

config/examples/redis.yml
```diff
@@ -5,6 +5,15 @@ base: &default
   sentinels: "<%= ENV['REDIS_SENTINEL_HOSTS'] %>"
   name: <%= ENV['REDIS_SENTINEL_NAME'] %>
   role: <%= ENV['REDIS_SENTINEL_ROLE'] %>
+  # == ACL Params ==
+  username: <%= ENV['REDIS_USERNAME'] %>
+  password: <%= ENV['REDIS_PASSWORD'] %>
```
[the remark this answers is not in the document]
akostadinov: Actually it might be even better to `#to_json` them. Otherwise a password like `my nice password with " sign` will break the quoting. While `#to_json` should handle all cases.

Then `#presence` should be used on the values to check them as they will never be `nil` I assume.

openshift/system/config/redis.yml
akostadinov: True, "client certificate". For me `TLS` is redundant as I said but I can live with it. More important for me is to have `CLIENT_CERT` and `PRIVATE_KEY`. This makes things clear beyond doubt.

We don't have to follow redis naming conventions rather we better establish a clear configuration naming conventions for porta.

jlledom: I finally set them to

```yaml
 ssl_params:
   ca_file: <%= ENV['REDIS_CA_FILE'] %>
   ca_path: <%= ENV['REDIS_CA_PATH'] %>
   cert: <%= ENV['REDIS_CLIENT_CERT'] %>
   key: <%= ENV['REDIS_PRIVATE_KEY'] %>
```
I hope you both agree on these.

config/examples/redis.yml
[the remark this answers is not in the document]
jlledom: Fixed:

```yaml
# == ACL Params ==
username: <%= ENV['REDIS_USERNAME']&.to_json %>
password: <%= ENV['REDIS_PASSWORD']&.to_json %>
```

This works fine for `nil`, `>` and `"`, and I assume any other character.

openshift/system/config/redis.yml
akostadinov: I think we are not on the same page here. I'm asking what will be the documented way to mount the certs/key in OpenShift. I know how to create a ca path (sorry for writing ca dir previously) with `openssl-rehash`. But probably that wont be the easiest option to mange as an openshift secret. Also the documented way will show what default location to use.

config/examples/redis.yml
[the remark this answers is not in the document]
akostadinov: Actually `nil.to_json` results in `null` so I think the `&` is redundant.

openshift/system/config/redis.yml
akostadinov: I'm fine with these.

openshift/system/config/redis.yml
jlledom: IMO if we provide both `ca_file` and `ca_path` then we are providing all available options for the operator team to define where to place the files or how to mount the secrets.

config/examples/redis.yml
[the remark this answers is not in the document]
jlledom: I tried removing the `&` and it still works. So you're right. But I'm still not sure, because:

```shell
$ bundle exec irb
irb(main):001:0> nil.to_json
Traceback (most recent call last):
...
NoMethodError (undefined method `to_json' for nil:NilClass)

irb(main):002:0> nil&.to_json
=> nil
```

But

```shell
$ bundle exec rails c
Loading development environment (Rails 6.0.6.1)
[1] pry(main)> nil.to_json
=> "null"
```

So the `&` is only redundant assuming Rails is loaded the moment the yaml is read., which I think we can assume it'll be always true. But I don't like Rails returning a string `"null"`, and sending that string to Redis, rather than a plain Ruby `nil` which we know for sure it won't cause problems on the Redis gems.

So I'd prefer to keep the `&`.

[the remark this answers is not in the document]
akostadinov: This is a template file which should result in valid YAML which in turn would be loaded by Rails. One option would result in:
```yaml
field:
```

The other:
```yaml
field: null
```

Both result in equivalent YAML, nothing to do with rails.

```rb
🐚 pry
[1] pry(main)> require 'yaml'
=> true
[2] pry(main)> YAML.load "field: "
=> {"field"=>nil}
[3] pry(main)> YAML.load "field: null"
=> {"field"=>nil}
```

One other concern though is that either way (using `&` or not) an empty environment variable will result in the empty string `""`. While your original version would treat and empty value for the variable as `nil`. I can't immediately say whether this is a problem or not.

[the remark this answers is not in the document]
jlledom: Fair enough, I removed the `&`

openshift/system/config/redis.yml
akostadinov: I was thinking whether it might be decided to provide CA file as content in an env variable... unlikely.

jlledom: I'm not sure the `redis-client` gem would accept that.

config/examples/redis.yml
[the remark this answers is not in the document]
akostadinov: FYI in case it matters, to have the original behavior where unset variable is equivalent to the empty one, we could use `ENV['REDIS_PASSWORD'].presence.to_json`

openshift/system/config/redis.yml
akostadinov: Yeah, this is fine. I think we need to check a default location though. Otherwise customers will need to both - mount the file somehow and then set the variable. It makes sense to let them only mount the file and things to start working.

jlledom: Do you suggest a path?

akostadinov: I don't know off the top of my head. Need to check other secrets that we mount.

jlledom: I've checked SaaS, also our dev cluster, and it seems secrets for porta are mounted at `EXTRA_CONFIGS_DIR`, which defaults to `/opt/system-extra-configs`. See https://github.com/3scale/porta/blob/master/openshift/system/entrypoint.sh#L5

What if we use the same path `$EXTRA_CONFIGS_DIR/ca_cert.pem` ?

So if no `ca_file` nor `ca_path` are provided, we can look for `$EXTRA_CONFIGS_DIR/ca_cert.pem` and use it if present.

akostadinov: Sounds good to. Probably filename should be `ca_redis.pem` or something like that. 

Do we have documented that users can edit this secret?

jlledom: > Sounds good to. Probably filename should be ca_redis.pem or something like that.

I think `ca_cert.pem` is better because this doesn't need to be exclusive for Redis. We provide a way for the user to define in which CAs they want porta to trust, it can be for Redis or anything else in the future. `ca_cert.pem` doesn't need to be a single certificate, a certificate bundle is also accepted.

> Do we have documented that users can edit this secret?

This is a new feature, not documented.

jlledom: I added this in https://github.com/3scale/porta/pull/3572/commits/81762db80a0b4e4cbfa0c5d2402b6836bcd825d2.

I finally took `config/` as the default path for the certificate. It turns out than `EXTRA_CONFIGS_DIR` is just a workaround to create symbolic links in `config/` (check the [code](https://github.com/3scale/porta/blob/master/openshift/system/entrypoint.sh)). So whatever secret the user mounts on `EXTRA_CONFIGS_DIR` will be available from `config/`.

akostadinov: ok, nice. Maybe test that this works, will also help write instructions how to do by customers.

app/lib/three_scale/redis_config.rb
```diff
@@ -56,5 +58,16 @@ def parse_sentinels(sentinels)
         parsed_sentinel
       end
     end
+
+    def load_default_ca_cert(conf)
+      return if conf[:ssl_params]&.key?(:ca_file) || conf[:ssl_params]&.key?(:ca_path)
```
akostadinov, line 63: Presently we have in config
```
ca_file: <%= ENV['BACKEND_REDIS_CA_FILE'] %>
```

This (I believe) means that the keys will always be set and the default will never be used. At least it requires testing. I expect that we would need something like
```suggestion
      return if conf.dig(:ssl_params, :ca_file).presence? || conf.dig(:ssl_params, :ca_path).presence?
```
It will still require some manual testing though to make sure behavior is as we expect.

openshift/system/config/redis.yml
jlledom: I updated the Jira issue with some comments

app/lib/three_scale/redis_config.rb
jlledom: Good catch, it was actually failing in that situation, I fixed it and added a new test for this.

app/lib/three_scale/redis_config.rb
```diff
@@ -56,5 +58,16 @@ def parse_sentinels(sentinels)
         parsed_sentinel
       end
     end
+
+    def load_default_ca_cert(conf)
+      return if conf.dig(:ssl_params, :ca_file).present? || conf.dig(:ssl_params, :ca_path).present?
+
+      cert_path = Rails.root.join('config/ca_cert.pem').to_s
```
akostadinov, line 65: I forgot if we discussed this already. But shouldn't default name be redis specific? This config foleder seems to be global for porta. But this is redis specific config. Maybe `redis_ca.pem`?

btw `ca_file` is more correct than `ca_cert` because it can be a single cert, a cert chain and I believe a whole bunch of CA certs. If the idea is that this fine be a global CA file for porta, then also better to rename.

akostadinov (approved):
Except for the renaming comment, all looks great!

app/lib/three_scale/redis_config.rb
jlledom: Yes, you mentioned that [here](https://github.com/3scale/porta/pull/3572/commits/1f518574c054da0a41725d507b7e5407a1ecc69c#r1397103552)

I don't think `ca_cert.pem` is so bad name. `cert.pem` t's the default name for the system certificate bundle in Debian, and also in RHEL for compatibility with Debian. Check your `/etc/ssl` folder.

akostadinov:
> Another possible situation is when the server is using a self-signed certificate. When this happens, there's no trusted CA to add to ssl_params, so the only way to go is to skip certificate validation.

With self-signed certificate, you should be able to use the certificate itself as CA too.   Should work OOB or maybe you need also to mark it as CA when creating.

app/lib/three_scale/redis_config.rb
akostadinov, line 65: Another question. If customer has put this file into the OCP secret, will it end up and be found here automatically?

test/unit/three_scale/redis_config_test.rb
```diff
@@ -43,5 +43,51 @@ class RedisConfigTest < ActiveSupport::TestCase
...
+        assert result[:ssl_params].key? :ca_file
+        assert_equal Rails.root.join('config/ca_cert.pem').to_s, result[:ssl_params][:ca_file]
+      end
+    end
+
+    test 'it trusts on CAs in config/ca_cert.pem if ca_file or ca_path are provided empty' do
+      FakeFS.with_fresh do
+        FakeFS::FileSystem.clone(file_fixture_path)
+        FakeFS::FileSystem.clone(Rails.root.join('config'))
+        FileUtils.cp file_fixture('ca_cert.pem'), Rails.root.join('config')
```
akostadinov, line 79: you can just create an empty file instead of keeping a fixture in the repo
```rb
FileUtils.touch
```

test/unit/three_scale/redis_config_test.rb
```diff
@@ -43,5 +43,51 @@ class RedisConfigTest < ActiveSupport::TestCase
...
+        assert result[:ssl_params].key? param
+        assert_equal value, result[:ssl_params][param]
+      end
+    end
+
+    test 'it trusts on CAs in config/ca_cert.pem if no ca_file or ca_path are provided' do
+      FakeFS.with_fresh do
+        FakeFS::FileSystem.clone(file_fixture_path)
+        FakeFS::FileSystem.clone(Rails.root.join('config'))
+        FileUtils.cp file_fixture('ca_cert.pem'), Rails.root.join('config')
```
akostadinov, line 65: you can just create an empty file instead of keeping a fixture in the repo
```rb
FileUtils.touch
```

akostadinov (commented):
Looks great! Just a few minor comments.

app/lib/three_scale/redis_config.rb
jlledom: It's not automatic unless implemented by the operator. Otherwise the user must edit the yamls to mount the certificate at `/opt/system/config` or `/opt/system-extra-configs`.

test/unit/three_scale/redis_config_test.rb
jlledom: Fixed https://github.com/3scale/porta/pull/3572/commits/fcc87b20bd83512cece9f93bb9de3f642a6dcf36

test/unit/three_scale/redis_config_test.rb
jlledom: Fixed https://github.com/3scale/porta/pull/3572/commits/fcc87b20bd83512cece9f93bb9de3f642a6dcf36

app/lib/three_scale/redis_config.rb
akostadinov: For me that's the whole point of he earlier discussion - when the file is added to the secret, to be used automatically without further configuration. Otherwise it is unnecessary complicated for the users.

jlledom: What could we do from porta?

akostadinov: We need to coordinate with operator folks how these can be mounted. For CA and cert files, we can put them in config maps I guess and have them mounted to default locations. For private keys though we need to use secrets and idk what is the best way to handle these.

config/examples/redis.yml
```diff
@@ -3,8 +3,20 @@ base: &default
...
   sentinels: "<%= ENV['REDIS_SENTINEL_HOSTS'] %>"
+  sentinel_username: <%= ENV['REDIS_SENTINEL_USERNAME'].to_json %>
+  sentinel_password: <%= ENV['REDIS_SENTINEL_PASSWORD'].to_json %>
   name: <%= ENV['REDIS_SENTINEL_NAME'] %>
   role: <%= ENV['REDIS_SENTINEL_ROLE'] %>
+  # == ACL Params ==
+  username: <%= ENV['REDIS_USERNAME'].to_json %>
+  password: <%= ENV['REDIS_PASSWORD'].to_json %>
+  # == TLS Params ==
+  ssl: <%= ENV['REDIS_SSL'] %>
```
mayorova, line 14: I'm a bit confused by this setting... `redis-client` docs say:

> `ssl`: Whether to connect using SSL or not.

But I am not sure how it is actually used :thinking: 

If I set it to `false` but use `rediss://` in the URL, it still connects via SSL. So, is it really needed? :upside_down_face:  Or does it only apply when the URL is provided in a different format? e.g. `host:port` with no protocol.

Also, if this value is expected to be a boolean, I think we need to cast the env var value (which is considered a string).

I think for most of the boolean values we are using this pattern:
```
ssl: <%= ENV.fetch('REDIS_SSL', '0') == '1' %>
```

See [settings.yml](https://github.com/3scale/porta/blob/eea43d58f4e1636cb14561cfa4940de06d07c817/config/examples/settings.yml#L11)

So, it needs to be set as `REDIS_SSL=1` to be `true`.

I don't like it too much, but we have it everywhere. An alternative would be to set `REDIS_SSL=true` but then change the config to
```
ssl: <%= ActiveModel::Type::Boolean.new.cast(ENV['REDIS_SSL']) %>
```

We discussed this with @akostadinov here: https://github.com/3scale/porta/pull/3107#discussion_r1021452236

app/lib/three_scale/redis_config.rb
mayorova: Why do we need to load a default CA cert in the first place? :thinking: 

I'd say - if nothing is provided in `:ca_file` or `:ca_path` explicitly, so be it? Does checking `config/ca_cert.pem` help in any way?...

config/examples/redis.yml
```diff
@@ -3,8 +3,20 @@ base: &default
   size: <%= ENV.fetch('RAILS_MAX_THREADS', 5) %>
   pool_timeout: 5 # this is in seconds
   sentinels: "<%= ENV['REDIS_SENTINEL_HOSTS'] %>"
+  sentinel_username: <%= ENV['REDIS_SENTINEL_USERNAME'].to_json %>
```
mayorova, line 6: I haven't tested these, but I guess it should just work, as we just pass the config to the redis client.

app/lib/three_scale/redis_config.rb
jlledom: If there's no `:ca_file` or `:ca_path` it checks `config/ca_cert.pem` to load that one. If it doesn't exist, then it loads the system cert bundle in `/etc`.

The idea behind `config/ca_cert.pem` is there to provide a way for the user to use its own certificate effortlessly.

config/examples/redis.yml
jlledom: `REDIS_SSL` is optional, because the gem is able to infer it correctly in most cases. But it doesn't infer it correctly for the sentinels + tls scenario. That scenario requires setting it manually.

I'll check the boolean format issue.

app/lib/three_scale/redis_config.rb
mayorova: But I think the effort is pretty much the same in both cases, just that the user will always need to set an env var for `:ca_file` or `:ca_path`. But at least it's clear where the cert should be mounted (wherever these env vars point to).

I am not sure we need to define some porta-specific CA cert location :shrug: 
But whatever, if you think it's useful, it's OK for me.

config/examples/redis.yml
jlledom: I tried locally and it works

config/examples/redis.yml
jlledom: Solved parsing `REDIS_SSL` https://github.com/3scale/porta/pull/3572/commits/3e537ba116edb95e2333a5f9f4f90bd5b632ceab

app/lib/three_scale/redis_config.rb
jlledom: I've been discussing with Eguz on how to integrate this TLS anc ACL with the operator, and we came to these two Jira issues:

https://issues.redhat.com/browse/THREESCALE-11020
https://issues.redhat.com/browse/THREESCALE-11021

Basically, porta and apisonator should limit to read the environment variables and it would be the operator who mounts the certificates and set the environment variables according to some new YAML keywords.

I made some changes to adapt to the new approach.

mayorova (approved):
Works for me with the following configs:

```
REDIS_SSL=1

REDIS_USERNAME=porta
REDIS_PASSWORD=portapassword
REDIS_CA_FILE=/path/to/ca-root-cert.pem
REDIS_CLIENT_CERT=/path/to/porta.crt
REDIS_PRIVATE_KEY=/path/to/porta.key

REDIS_URL=redis://redis-master
REDIS_SENTINEL_HOSTS=localhost:56380,localhost:56381,localhost:56382
REDIS_SENTINEL_NAME=redis-master
REDIS_SENTINEL_ROLE=master
REDIS_SENTINEL_USERNAME=sentuser
REDIS_SENTINEL_PASSWORD=sentpass


BACKEND_REDIS_SSL=1

BACKEND_REDIS_USERNAME=apisonator
BACKEND_REDIS_PASSWORD=apisonatorpassword
BACKEND_REDIS_CA_FILE=/path/to/ca-root-cert.pem
BACKEND_REDIS_CLIENT_CERT=/path/to/apisonator.crt
BACKEND_REDIS_PRIVATE_KEY=/path/to/apisonator.key

BACKEND_REDIS_URL=redis://redis-master
BACKEND_REDIS_SENTINEL_HOSTS=localhost:56380,localhost:56381,localhost:56382
BACKEND_REDIS_SENTINEL_NAME=redis-master
BACKEND_REDIS_SENTINEL_ROLE=master
BACKEND_REDIS_SENTINEL_USERNAME=sentuser
BACKEND_REDIS_SENTINEL_PASSWORD=sentpass
```

Great job :+1:

config/examples/backend_redis.yml
```diff
@@ -4,8 +4,19 @@ base: &default
...
   role: <%= ENV['BACKEND_REDIS_SENTINEL_ROLE'] %>
+  # == ACL Params ==
+  username: <%= ENV['BACKEND_REDIS_USERNAME'].to_json %>
+  password: <%= ENV['BACKEND_REDIS_PASSWORD'].to_json %>
+  # == TLS Params ==
+  ssl: <%= ENV.fetch('BACKEND_REDIS_SSL', '0') == '1' %>
+  ssl_params:
+    ca_file: <%= ENV['BACKEND_REDIS_CA_FILE'] %>
+    cert: <%= ENV['BACKEND_REDIS_CLIENT_CERT'] %>
+    key: <%= ENV['BACKEND_REDIS_PRIVATE_KEY'] %>
```
akostadinov, line 19: If my assumption is correct, one can enable ssl with just CA when client auth is not needed. So we should still allow cert and key to be omitted.

Also I believe we should allow CA_FILE to be `nil` when the default operating system CA is suitable for the connection.

```suggestion
    ca_file: <%= ENV['BACKEND_REDIS_CA_FILE'].to_json %>
    cert: <%= ENV['BACKEND_REDIS_CLIENT_CERT'].to_json %>
    key: <%= ENV['BACKEND_REDIS_PRIVATE_KEY'].to_json %>
```

openshift/system/config/backend_redis.yml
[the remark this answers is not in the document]
mayorova: > it's fine if you leave it for another time, but would be really nice to get rid of these openshift specific files, is there anything different than the default config files?

@akostadinov let's cover this in https://github.com/3scale/porta/pull/3766

Once this PR is merged, I'll incorporate the changes there.

test/unit/three_scale/redis_config_test.rb
```diff
@@ -43,5 +43,23 @@ class RedisConfigTest < ActiveSupport::TestCase
...
+      raw_config[:ssl_params][:ca_file] = value
+
+      result = RedisConfig.new(raw_config)
+
+      assert result.key? :ssl_params
+      assert result[:ssl_params].key? :ca_file
+      assert_equal value, result[:ssl_params][:ca_file]
+    end
+
+    test "it doesn't trust any CA if no ca_file is provided" do
```
akostadinov, line 59:
```suggestion
    test "it doesn't set CA if no ca_file is provided in config" do
```

config/examples/redis.yml
```diff
@@ -3,8 +3,19 @@ base: &default
...
   role: <%= ENV['REDIS_SENTINEL_ROLE'] %>
+  # == ACL Params ==
+  username: <%= ENV['REDIS_USERNAME'].to_json %>
+  password: <%= ENV['REDIS_PASSWORD'].to_json %>
+  # == TLS Params ==
+  ssl: <%= ENV.fetch('REDIS_SSL', '0') == '1' %>
+  ssl_params:
+    ca_file: <%= ENV['REDIS_CA_FILE'] %>
+    cert: <%= ENV['REDIS_CLIENT_CERT'] %>
+    key: <%= ENV['REDIS_PRIVATE_KEY'] %>
```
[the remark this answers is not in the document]
akostadinov: wait, ignore this, it will result in `nil` anyway, correct? So no need to apply this. Will it handle spaces in path without `to_json`?

openshift/system/config/backend_redis.yml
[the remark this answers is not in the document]
jlledom: Agreed

config/examples/redis.yml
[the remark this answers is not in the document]
jlledom: What's the call to `to_json` for anyway?

test/unit/three_scale/redis_config_test.rb
jlledom: Right, it'll trust the SO cert bundle in this situation

config/examples/redis.yml
```diff
@@ -3,8 +3,19 @@ base: &default
...
   sentinels: "<%= ENV['REDIS_SENTINEL_HOSTS'] %>"
+  sentinel_username: <%= ENV['REDIS_SENTINEL_USERNAME'].to_json %>
+  sentinel_password: <%= ENV['REDIS_SENTINEL_PASSWORD'].to_json %>
   name: <%= ENV['REDIS_SENTINEL_NAME'] %>
   role: <%= ENV['REDIS_SENTINEL_ROLE'] %>
+  # == ACL Params ==
+  username: <%= ENV['REDIS_USERNAME'].to_json %>
+  password: <%= ENV['REDIS_PASSWORD'].to_json %>
+  # == TLS Params ==
+  ssl: <%= ENV.fetch('REDIS_SSL', '0') == '1' %>
```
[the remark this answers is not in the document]
jlledom: In that situation the Redis client will infer `ssl: true`. This variable is optional in almost all cases, it's only mandatory when connecting to TLS protected sentinels.

akostadinov (approved):

config/examples/redis.yml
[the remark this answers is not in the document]
akostadinov: `to_json` will escape the values if needed

config/examples/redis.yml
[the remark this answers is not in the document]
akostadinov: I'm just wondering, when this is set to `false` hopefully it wont enforce no SSL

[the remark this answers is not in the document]
mayorova: > I'm just wondering, when this is set to `false` hopefully it wont enforce no SSL

I believe it will ONLY enable SSL when `REDIS_SSL=1`, it will be disabled with any other value.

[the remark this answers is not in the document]
jlledom: > I believe it will ONLY enable SSL when REDIS_SSL=1, it will be disabled with any other value.

True

> I'm just wondering, when this is set to false hopefully it wont enforce no SSL

No, you can ignore the value or set it to `0`, `false` or whatever and it still connects to redis if you provide a certificate. The scenario you mention would only fail for sentinels + TLS, where `REDIS_SSL=1` is mandatory.

PR #3596, opened 2023-10-20
🦋 Update Billing Charging & Gateway settings

josemigallas:
[THREESCALE-10237: Billing > Settings > Charging & Gateway](https://issues.redhat.com/browse/THREESCALE-10237)

New features:
- The deprecation warning is shown both in the page header and the gateway form section

Before:
Screenshot 2023-10-18 at 15 11 16

After:
Screenshot 2023-10-20 at 11 43 35
Screenshot 2023-10-20 at 11 43 48

### DIFF
--- a/app/views/finance/provider/settings/show.html.slim
+++ b/app/views/finance/provider/settings/show.html.slim
@@ -1,135 +1,113 @@
-- content_for(:title)
-  = "Charging & Gateway"
-
+- content_for :title, 'Charging & Gateway'
 - content_for :page_header_title, 'Charging & Gateway'
 
-= form_for @billing_strategy, url: admin_finance_billing_strategy_path,
-           method: :put, html: { class: 'formtastic' } do |f|
-  fieldset class="inputs"
-    ol
-      li
-        label Mode:
-        input type='text' value=(@billing_strategy.name.capitalize) disabled='disabled'
-        p class="inline-hints"
-          = @billing_strategy.description
+- if @current_gateway&.deprecated?
+  - content_for :page_header_alert do
+    br
+    = pf_inline_alert t('.deprecation_warning', gateway: @current_gateway.type.capitalize), variant: :warning
 
-      li
-        - new_mode = @billing_strategy.is_a?(Finance::PostpaidBillingStrategy) ? 'prepaid' : 'postpaid'
-        = button_tag "Switch to #{new_mode.upcase}",
-                     onclick: "return confirm('Are you sure you want to switch to #{new_mode} mode?');",
-                     class: "pf-c-button pf-m-danger button next"
-        = hidden_field_tag 'type', new_mode
+div class="pf-l-flex pf-m-column"
+  div class="pf-l-flex__item"
+    div class="pf-c-card"
+      div class="pf-c-card__title" Billing mode
+      div class="pf-c-card__body"
+        = semantic_form_for @billing_strategy, url: admin_finance_billing_strategy_path,
+                                               method: :put,
+                                               html: { class: 'pf-c-form pf-m-limit-width' } do |form|
+          div class="pf-c-content"
+            = t(".#{@billing_strategy.name}_description_html")
 
-= semantic_form_for @billing_strategy.becomes(Finance::BillingStrategy),
-                    url: admin_finance_settings_path do |f|
-  = f.inputs do
-    = f.input :charging_enabled,
-              hint: "Enables credit card transactions. When this setting is on,
-                    all due invoices will be charged using the selected payment \
-                    gateway. If you leave this setting off, billing will take \
-                    place and invoices will issued, but no real transaction will \
-                    take place."
+          - new_mode = @billing_strategy.is_a?(Finance::PostpaidBillingStrategy) ? 'prepaid' : 'postpaid'
+          = hidden_field_tag 'type', new_mode
+          div class="pf-c-form__group pf-m-action"
+            div class="pf-c-form__actions"
+              button class="pf-c-button pf-m-danger" type="submit" data={ confirm: t('.switch_confirm', mode: new_mode) }
+                = t('.switch_button', mode: new_mode.upcase)
 
-    = f.input :currency,
-              as: :select,
-              collection: Finance::BillingStrategy::CURRENCIES,
-              hint: "Please make sure that the selected currency is supported by \
-                    your payment gateway. If you are missing currency options,
-                    don't hesitate to contact us."
+  div class="pf-l-flex__item"
+    div class="pf-c-card"
+      div class="pf-c-card__title" Billing settings
+      div class="pf-c-card__body"
+        = semantic_form_for @billing_strategy.becomes(Finance::BillingStrategy),
+                            url: admin_finance_settings_path,
+                            html: { class: 'pf-c-form pf-m-limit-width' } do |form|
+          = form.input :charging_enabled, as: :patternfly_checkbox
+          = form.input :currency, as: :patternfly_select,
+                                  collection: Finance::BillingStrategy::CURRENCIES,
+                                  include_blank: false
 
+          = form.semantic_fields_for :account do |account|
+            = account.input :invoice_footnote, as: :patternfly_textarea, input_html: { rows: 2 }
+            = account.input :vat_zero_text, as: :patternfly_textarea, input_html: { rows: 2 }
 
-    = f.semantic_fields_for :account do |account|
-      = account.input :invoice_footnote, input_html: {rows: 2}
-      = account.input :vat_zero_text,
-                      label: 'Text to show if VAT/Sales Tax is 0%',
-                      input_html: {rows: 2}
+          = form.input :numbering_period, as: :patternfly_select,
+                                          collection: [['Monthly (YYYY-MM-XXXXXXXX)', 'monthly'], ['Yearly (YYYY-XXXXXXXX)', 'yearly']],
+                                          include_blank: false
 
-    = f.input :numbering_period,
-              label: 'Billing periods for invoice ids',
-              collection: ['monthly', 'yearly'],
-              include_blank: false
+          div class="pf-c-form__group pf-m-action"
+            div class="pf-c-form__actions"
+              button class="pf-c-button pf-m-primary" type="submit" Save
 
-    = f.actions do
-      = f.commit_button 'Save'
+  - if can?(:manage, :finance) && @account.billing_strategy.try(:needs_credit_card?)
+    div class="pf-l-flex__item"
+    div class="pf-c-card"
+      div class="pf-c-card__title" Credit card gateway
+      div class="pf-c-card__body"
+        = semantic_form_for @account, url: admin_account_payment_gateway_path,
+                                      html: { id: 'payment-gateway-form', class: 'pf-c-form pf-m-limit-width' } do |form|
+          = form.input :payment_gateway_type, as: :patternfly_select,
+                                              collection: @payment_gateways.map { |gateway| [gateway.display_name, gateway.type] },
+                                              include_blank: 'Choose one'
 
-- if can?(:manage, :finance)
-  - if @account.billing_strategy.try(:needs_credit_card?)
-    = semantic_form_for @account, url: admin_account_payment_gateway_path,
-                        html: { id: 'payment-gateway-form' } do |form|
-      h2 Credit card gateway
-      p Select your credit card gateway and configure it with the details of your merchant account.
-
-      = form.inputs "Basics" do
-        li class="select"
-          = form.label :payment_gateway_type, 'Gateway'
-          = form.collection_select :payment_gateway_type, @payment_gateways,
-                                   :type, :display_name, include_blank: "Choose one"
-          - if @current_gateway&.deprecated?
-            p class="InfoBox InfoBox--notice"
-              ' The
-              => @account.payment_gateway_type.capitalize
-              | gateway has been deprecated. It can still be used but you won't be able to switch back to it once you confirmed the switch to another gateway.
-
-      = form.fields_for :payment_gateway_options do |fields|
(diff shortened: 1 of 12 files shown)


config/locales/en.yml
```diff
@@ -1259,6 +1294,19 @@ en:
       rest: 'REST API'
   formtastic:
     hints:
+      charging_enabled:
+        Enables credit card transactions. When this setting is on, all due invoices will be
+        charged using the selected payment gateway. If you leave this setting off, billing will
+        take place and invoices will issued, but no real transaction will take place.
```
mayorova, line 1300:
```suggestion
        take place and invoices will be issued, but no real transaction will take place.
```

app/views/finance/provider/settings/show.html.slim
```diff
@@ -1,135 +1,114 @@
...
+          - new_mode = @billing_strategy.is_a?(Finance::PostpaidBillingStrategy) ? 'prepaid' : 'postpaid'
+          = hidden_field_tag 'type', new_mode
+          div class="pf-c-form__group pf-m-action"
+            div class="pf-c-form__actions"
+              button class="pf-c-button pf-m-danger" type="submit" data={ confirm: t('.switch_confirm', mode: new_mode) }
+                = t('.switch_button', mode: new_mode.upcase)
+
+  div class="pf-l-flex__item"
+    div class="pf-c-card"
+      div class="pf-c-card__title" Billing Strategy
```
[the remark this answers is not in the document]
mayorova: I think "Strategy" is more of an internal term, which is not necessarily understandable to end users.
Maybe even Billing and Charging Settings, as it covers both.

config/locales/en.yml
```diff
@@ -1375,6 +1423,12 @@ en:
       policies:
         update: "Update Policy Chain"
     labels:
+      account:
+        payment_gateway_type: Gateway
+
+      billing_strategy:
+        numbering_period: Billing periods for invoice ids
```
mayorova, line 1430: This translation doesn't seem to be kicking in?...
image


Also, we could:
- say IDs instead
- maybe add a hint for this field to help users understand what it means (an example of generated IDs)

config/locales/en.yml
```diff
@@ -595,6 +595,40 @@ en:
...
+            </p>
+          deprecation_warning:
+            The %{gateway} gateway has been deprecated. It can still be used but you won't be able
+            to switch back to it once you confirmed the switch to another gateway.
+          save_gateway_confirm_1:
+            Changing your payment gateway will make all your buyers' credit card details invalid, so
+            they will have to enter the credit card information again to be billed next month.\n\n
+            Are you sure you want to change your payment gateway?
+          save_gateway_confirm_2:
+            Are you sure? If you click OK, all credit card details from all your customers will be
```
mayorova, line 627: Some weird characters in here:
image

mayorova: And the 1st confirm also

app/views/finance/provider/settings/show.html.slim
```diff
@@ -1,135 +1,114 @@
...
+                                             label: label,
+                                             checked: false,
+                                             id: ['account', 'payment', payment_gateway.type, name].join('_')
                     - else
-                      = fields.label name, label
-                      - if enabled
-                        = fields.text_field name, value: @account.payment_gateway_options[name]
+                      - if is_current
+                        = fields.input name, as: :patternfly_input,
+                                             input_html: { value: @account.payment_gateway_options[name] }
```
mayorova, line 81:
```suggestion
                        = fields.input name, as: :patternfly_input,
                                             label: label,
                                             input_html: { value: @account.payment_gateway_options[name] }
```

app/views/finance/provider/settings/show.html.slim
```diff
@@ -1,135 +1,114 @@
...
-        var second_question = "Are you sure? If you click OK, all credit card details from all your customers will be erased and won't be recoverable.\n"+
-        "You'll need to ask your customers to enter their credit card details again to be able to bill them.\n\n" +
-        "Do you want to remove all credit card details from your customers?";
-
-
-        return confirm(first_question) && confirm(second_question);
-      });
-    });
+                        = fields.input name, as: :patternfly_input,
+                                             disabled: true
```
mayorova, line 84:
```suggestion
                        = fields.input name, as: :patternfly_input,
                                             label: label,
                                             disabled: true
```

app/views/finance/provider/settings/show.html.slim
mayorova: Otherwise, proper label is not applied:
image

vs 
image

app/views/finance/provider/settings/show.html.slim
mayorova, line 84: Also, `disabled: true` doesn't seem to be handled by the patternfly input :thinking:

app/views/finance/provider/settings/show.html.slim
```diff
@@ -1,135 +1,114 @@
...
-                          = fields.check_box name, checked: @account.payment_gateway_options[name].present?, id: ['account', 'payment', payment_gateway.type, name].join('_')
-                        - else
-                          = fields.check_box name, checked: false, disabled: true, id: ['account', 'payment', payment_gateway.type, name].join('_')
-
-                        = label
-                      p.inline-hints Make sure your Braintree account has 3D Secure enabled before you enable it here as otherwise payments initiated from the developer portal will fail.
-
+                      - if is_current
+                        = fields.input name, as: :patternfly_checkbox,
+                                             id: ['account', 'payment', payment_gateway.type, name].join('_'),
```
mayorova, line 71: A proper label is missing when current :upside_down_face: 

image

app/views/finance/provider/settings/show.html.slim
```diff
@@ -1,135 +1,114 @@
...
+        const form = document.forms['payment-gateway-form']
+        const gatewaySelect = form.querySelector('#account_payment_gateway_type')
+        const gatewaySections = form.querySelectorAll('.payment_gateway_options')
+
+        gatewaySelect.addEventListener('change', event => {
+          const selectedGatewayId = `payment_gateway_${event.target.value}`
+          gatewaySections.forEach(div => div.classList.toggle('hidden', div.id !== selectedGatewayId))
+        })
+
+        // TODO: Don't use confirm as it can be bloqued by browser. Create custom modal.
```
mayorova, line 103: "Bloqued"? :)

app/views/finance/provider/settings/show.html.slim
mayorova: Actually, how about this instead of multiple branches:

```
                    - if payment_gateway.boolean_field_keys.include?(name)
                      = fields.input name, as: :patternfly_checkbox,
                              id: ['account', 'payment', payment_gateway.type, name].join('_'),
                              label: label,
                              input_html: { checked: is_current ? @account.payment_gateway_options[name].present? : false, disabled: !is_current }
                    - else
                      = fields.input name, as: :patternfly_input,
                              label: label,
                              input_html: { value: is_current ? @account.payment_gateway_options[name] : '', disabled: !is_current }
```

Except that the JS handing of `disabled` has been removed, so it doesn't work. So, I'd even say that we can remove the `disabled` piece, because the part of the form is hidden anyway, so if it's not visible, why disable it?

mayorova (dismissed):
Comments inline

app/views/finance/provider/settings/show.html.slim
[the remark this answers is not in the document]
josemigallas: I like _Billing Settings_ more, although there is a bit of inconsistency between Billing and Charging in the same page.

An alternative is to merge both upper cards Mode and Strategy into a single card _Charging settings_ which coincides with the page name "Charging & Gateway". On the other hand, 2 submit buttons in the same card is odd.

app/views/finance/provider/settings/show.html.slim
josemigallas: AFAIK this `disabled` doesn't do anything and it's probably a typo. Moreover, disabling the inputs would prevent the user from changing to another gateway.

NOTE: Current means this is the gateway currently in use, not the one currently selected in the dropdown.

app/views/finance/provider/settings/show.html.slim
josemigallas: Spaniards gonna spaniard

config/locales/en.yml
josemigallas: Good catch 👍🏻

app/views/finance/provider/settings/show.html.slim
mayorova: Hm, well, I guess initially "current" is for setting "currently in use", but also the toggling enabled/disabled was done for "currently selected" https://github.com/3scale/porta/pull/3596/files#diff-8cc2bd457556805766347846af6a80e7635d681acd9fa5fb0a4cd3c7a05751e3L109-L117

I agree that there is no need to enable/disable the inputs, as they are already visible/hidden.

app/views/finance/provider/settings/show.html.slim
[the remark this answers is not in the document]
mayorova: Yeah, I also thought about merging the form into one, but indeed, it might be even more confusing in the end :upside_down_face:

app/views/finance/provider/settings/show.html.slim
josemigallas: OK I see! 👍🏻  I forgot about that JS callback.

config/locales/en.yml
josemigallas: In addition to the hint: _It will only change the friendly ID in invoices, not the billing period._

What do you thing about using the following labels?
Screenshot 2023-10-23 at 13 55 03

josemigallas: Ideally I would add a popover with some info taken from [the docs](https://access.redhat.com/documentation/en-us/red_hat_3scale_api_management/2.13/html/admin_portal_guide/configure-billing#billing_periods_for_invoice_ids_charging_gateway).
 
Screenshot 2023-10-23 at 13 47 50

config/locales/en.yml
```diff
@@ -1259,6 +1293,20 @@ en:
...
   formtastic:
     hints:
+      numbering_period: It will only change the friendly ID in invoices, not the billing period.
+      charging_enabled:
+        Enables credit card transactions. When this setting is on, all due invoices will be
+        charged using the selected payment gateway. If you leave this setting off, billing will
+        take place and invoices will be issued, but no real transaction will take place.
+      currency:
+        Please make sure that the selected currency is supported by your payment gateway. If you are
+        missing currency options, don't hesitate to contact us.
```
mayorova, line 1303:
```suggestion
      finance/billing_strategy:
        numbering_period: It will only change the friendly ID in invoices, not the billing period.
        charging_enabled:
          Enables credit card transactions. When this setting is on, all due invoices will be
          charged using the selected payment gateway. If you leave this setting off, billing will
          take place and invoices will be issued, but no real transaction will take place.
        currency:
          Please make sure that the selected currency is supported by your payment gateway. If you are
          missing currency options, don't hesitate to contact us.
```

config/locales/en.yml
```diff
@@ -1259,6 +1293,20 @@ en:
...
+      account:
+        payment_gateway_type: Select your credit card gateway and configure it with the details of your merchant account.
+        payment_gateway_options:
+          three_ds_enabled: Make sure your Braintree account has 3D Secure enabled before you enable it here as otherwise payments initiated from the developer portal will fail.
+      finance_billing_strategy:
```
mayorova, line 1308:
```suggestion
```

config/locales/en.yml
```diff
@@ -1375,6 +1423,14 @@ en:
       policies:
         update: "Update Policy Chain"
     labels:
+      account:
+        payment_gateway_type: Gateway
+
+      numbering_period: Billing periods for invoice IDs
+      finance_billing_strategy:
```
mayorova, line 1430:
```suggestion
      finance/billing_strategy:
```

mayorova: Ooops, actually it breaks it!

I have no idea why `hints` and `labels` do not behave the same way :smiling_face_with_tear:

config/locales/en.yml
josemigallas: Awesome

config/locales/en.yml
josemigallas: Wow nice finding. I thought it was always made by underscores.

lvillen (dismissed):

app/views/finance/provider/settings/show.html.slim
```diff
@@ -1,135 +1,113 @@
...
+          - new_mode = @billing_strategy.is_a?(Finance::PostpaidBillingStrategy) ? 'prepaid' : 'postpaid'
+          = hidden_field_tag 'type', new_mode
+          div class="pf-c-form__group pf-m-action"
+            div class="pf-c-form__actions"
+              button class="pf-c-button pf-m-danger" type="submit" data={ confirm: t('.switch_confirm', mode: new_mode) }
+                = t('.switch_button', mode: new_mode.upcase)
+
+  div class="pf-l-flex__item"
+    div class="pf-c-card"
+      div class="pf-c-card__title" Billing Settings
```
mayorova, line 29: Just a nitpick :grimacing: 

"Billing Settings" uses title case, while "Billing mode" and "Credit card gateway" use sentence case.
I'm not quite sure what we use in other pages, probably a mix also, but within the same page I think we need to use the same style :grimacing: I personally think sentence case looks better here.

mayorova: Apart from this, looks great!

mayorova (approved):
Approving under assumption that the only change was the "Billing settings" title https://github.com/3scale/porta/pull/3596#pullrequestreview-1694727713

PR #3619, opened 2023-11-14
update mocha

akostadinov:
mocha 1.1.0 seems broken, see #3618

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -463,7 +463,6 @@ GEM
       ntlm-http (~> 0.1, >= 0.1.1)
       webrick (~> 1.7)
       webrobots (>= 0.0.9, < 0.2)
-    metaclass (0.0.4)
     method_source (1.0.0)
     middleware (0.1.0)
     mime-types (3.4.1)
@@ -483,8 +482,8 @@ GEM
       minitest (>= 5.0)
       ruby-progressbar
     minitest-stub-const (0.6)
-    mocha (1.1.0)
-      metaclass (~> 0.0.1)
+    mocha (2.1.0)
+      ruby2_keywords (>= 0.0.5)
     msgpack (1.7.1)
     multi_json (1.15.0)
     multi_test (0.1.2)
@@ -730,6 +729,7 @@ GEM
     ruby-prof (0.17.0)
     ruby-progressbar (1.11.0)
     ruby-rc4 (0.1.5)
+    ruby2_keywords (0.0.5)
     ruby_event_store (0.9.0)
     rubyzip (1.3.0)
     sass (3.4.25)
@@ -984,7 +984,7 @@ DEPENDENCIES
   minitest-ci
   minitest-reporters
   minitest-stub-const
-  mocha (~> 1.1.0)
+  mocha
   mysql2 (~> 0.5.3)
   n_plus_one_control
   nokogiri (~> 1.13.10)
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -124,9 +124,31 @@ RSpec.configure do |config|
 
   end
 
+  config.after(:each) do |ex|
+    # I hope this would be easy to maintain in the long run, but it is optional.
+    # Just to remind anybody trying to use Mocha expectations in specs that this is not mandated.
+    has_expectations = Mocha::Mockery.instance.mocks.any? do |mock|
+      mock.__expectations__.to_a.any? do |expectation|
+        !expectation.instance_variable_get(:@cardinality).allowed_any_number_of_times?
+      end
+    end
+
+    raise "use rspec mocking instead of Mocha" if has_expectations
+  end
+
   config.before(:suite) do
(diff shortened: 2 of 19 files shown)


jlledom (commented):
There are many tests failing. I'll approve when they pass.

jlledom (approved):
> Everybody would approve if tests passed. I thought you were a friend.

Oh, what the hell, approved.

akostadinov:
@jlledom , a little bit more work now :grimacing:

Gemfile.lock
```diff
@@ -1070,4 +1070,4 @@ DEPENDENCIES
   zip-zip
 
 BUNDLED WITH
-   2.2.25
+   2.3.25
```
jlledom, line 1073: Better don't change this. We can upgrade bundler in another PR.

spec/rails_helper.rb
```diff
@@ -124,11 +124,33 @@
...
   end
 
+  config.after(:each) do |ex|
+    # I hope this would be easy to maintain in the long run, but it is optional.
+    # Just to remind anybody trying to use Mocha expectations in specs that this is not mandated.
+    has_expectations = Mocha::Mockery.instance.mocks.any? do |mock|
+      mock.__expectations__.to_a.any? do |expectation|
+        !expectation.instance_variable_get(:@cardinality).allowed_any_number_of_times?
+      end
+    end
```
jlledom, line 134: This seems to access Mocha internal variables and methods. If they change any of these in a next release, this will stop working. We'll have to be aware of this every time we want to update mocha, which I think we'll forget. I would look for another way to implement this check, or remove it if it's optional.

test/unit/logic/provider_upgrade_test.rb
```diff
@@ -6,19 +6,21 @@ def setup
...
+    rule.stubs(:switches).returns(%i[finance multiple_applications branding require_cc_on_signup
       account_plans multiple_users groups]
     )
-    @power1M.plan_rule.stubs(:limits).returns(PlanRule::Limit.new(max_services: 1, max_users: 1))
-    @power1M.plan_rule.stubs(:rank).returns(10)
+    rule.stubs(:limits).returns(PlanRule::Limit.new(max_services: 1, max_users: 1))
+    rule.stubs(:rank).returns(10)
     @pro = FactoryBot.create(:published_plan, :system_name => 'pro3M', :issuer => service)
-    @pro.plan_rule.stubs(:switches).returns(%i[finance multiple_applications branding require_cc_on_signup account_plans
+    rule = @pro.send :plan_rule
```
jlledom, line 16: why reusing the name for this variable? Why not use a different one? it would be less confusing.

test/unit/mailers/account_mailer_test.rb
```diff
@@ -74,7 +74,7 @@ def test_support_entitlements_assigned
 
     def test_support_entitlements_trial_plan
       trial_plan = FactoryBot.create(:application_plan, issuer: @service)
-      trial_plan.plan_rule.metadata = {trial: true}
+      trial_plan.send(:plan_rule).metadata = {trial: true}
```
jlledom, line 77: why this change? Is it not equivalent?

jlledom (approved):

test/unit/logic/provider_upgrade_test.rb
akostadinov: Just saving in a local variable to avoid duplicate calls. Suggestions for naming?

jlledom: `pro_rule` ?

Gemfile.lock
akostadinov: mistake, will fix it, thanks!

spec/rails_helper.rb
akostadinov: It is optional. If api changes, then test suite would fail. And we will know. So my line of thinking is that we can keep it until it becomes a problem and remove if we want later. Maybe comment not clear enough?

test/unit/mailers/account_mailer_test.rb
akostadinov: mocha 2.x stubs private methods as private. Previously all was stubbed public. So we can't call directly anymore.

Gemfile.lock
akostadinov: fixed

test/unit/logic/provider_upgrade_test.rb
akostadinov: fixed

test/unit/mailers/account_mailer_test.rb
jlledom: I wonder why ruby implements private methods if private protection is so easy to bypass.

jlledom (approved):

test/unit/mailers/account_mailer_test.rb
akostadinov: It's to notify the user more than enforce restrictions. Still somehow useful.

